Issue #16090, opened 2022-02-18
tensorflow.keras.utils.plot_model not working correctly with show_shapes=True


RicardFos:
Hi, I've just used the tf.keras.utils.plot_model with tensorflow 2.8.0 and the following diagram appeared:
![image](https://user-images.githubusercontent.com/11573780/154634034-8863e81e-be61-42fb-a6b5-d13f8255bf51.png)

As you can see, the right part of the diagram shows the input and output shapes with a vertical split, but it should be an horizontal split to fit with the "intput:" and  "output:" cells at its left.

RicardFos:
Hi, I could work on it, no problem.

mattdangerw:
Assigning this to you @RicardFos as it looks like you are actively working on this and the conversation has moved to the PR https://github.com/keras-team/keras/pull/16096. Thanks for contributing!

jvishnuvardhan:
@RicardFos This was resolved in recent `tf-nightly` by your PR as it got merged. [Here](https://colab.research.google.com/gist/jvishnuvardhan/a00d1fea601fefae2835d87a8fb84ed2/untitled234.ipynb) is a gist for reference. 

Please verify once and close the issue if this was resolved for you. Thanks!

RicardFos:
Thanks, I've checked the notebook and it works correctly, I'll mark this issue as resolved.

PR #16096, opened 2022-02-19
Issue #16090: Split input_shapes horizontally in utils.vis_utils.plot_model

RicardFos:
From Issue #16090 

Split input_shapes horizontally and not vertically in keras.utils.vis_utils.plot_model when show_shapes=True

Currently, the plot_model function with show_shapes=True returns a table like this:
image

With my contribution, the table will show the last cell, the one with input shapes, splitted horizontally so that the input shape is on  top of the output shape, making it fit with the previus cell (input:/output:)

Here's an example of the execution with the fix:

image

### DIFF
--- a/keras/utils/vis_utils.py
+++ b/keras/utils/vis_utils.py
@@ -290,8 +290,7 @@ def model_to_dot(model,
             [format_shape(ishape) for ishape in layer.input_shapes])
       else:
         inputlabels = '?'
-      label = '{%s}|{input:|output:}|{{%s}|{%s}}' % (label, inputlabels,
-                                                       outputlabels)
+      label = '{%s}|{input:|output:}|{{%s}|{%s}}' % (label, inputlabels, outputlabels)
     if not expand_nested or not isinstance(
         layer, functional.Functional):
       node = pydot.Node(layer_id, label=label)
--- a/keras/utils/vis_utils.py
+++ b/keras/utils/vis_utils.py
@@ -290,7 +290,8 @@ def model_to_dot(model,
             [format_shape(ishape) for ishape in layer.input_shapes])
       else:
         inputlabels = '?'
-      label = '{%s}|{input:|output:}|{{%s}|{%s}}' % (label, inputlabels, outputlabels)
+      label = '{%s}|{input:|output:}|{{%s}|{%s}}' % (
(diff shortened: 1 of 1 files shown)


fchollet:
Thanks for the PR! How does it look like in horizontal orientation? (`rankdir="LR"`)

RicardFos:
> Thanks for the PR! How does it look like in horizontal orientation? (`rankdir="LR"`)

Hi, the horizontal version in the current version looks like this:

image

And in my Pull Request it looks like this:

image

As you can see, the cell with the input and outpiut shapes matches the previous cell in horizontal mode as well.

keras/utils/vis_utils.py
```diff
@@ -290,7 +290,7 @@ def format_shape(shape):
             [format_shape(ishape) for ishape in layer.input_shapes])
       else:
         inputlabels = '?'
-      label = '{%s}|{input:|output:}|{{%s}}|{{%s}}' % (label, inputlabels,
+      label = '{%s}|{input:|output:}|{{%s}|{%s}}' % (label, inputlabels,
                                                        outputlabels)
```
fchollet, line 294: Please fix indent here

RicardFos: I understand you mean the EOL after inputlables, I fixed it:

image

keras/utils/vis_utils.py
```diff
@@ -290,8 +290,7 @@ def format_shape(shape):
             [format_shape(ishape) for ishape in layer.input_shapes])
       else:
         inputlabels = '?'
-      label = '{%s}|{input:|output:}|{{%s}}|{{%s}}' % (label, inputlabels,
-                                                       outputlabels)
+      label = '{%s}|{input:|output:}|{{%s}|{%s}}' % (label, inputlabels, outputlabels)
```
fchollet, line 293: Now the line is too long (86 char > 80). Let's break it down

```
label = '{%s}|{input:|output:}|{{%s}|{%s}}' % (
    label, inputlabels, outputlabels)
```

RicardFos: Sorry about that, fixed it already.

fchollet (approved):
LGTM, thank you

PR #16145, opened 2022-03-01
Fixed EfficientNetV2 b parameter not increasing with each block.

### LINKED ISSUE #16144: EfficientNetV2 does not match google implementation
Hi,
I may be wrong, but checking the efficientnetv2 implementation I think there is a difference with the [google one](https://github.com/google/automl/blob/387d5ddb92bb8fbbec4b012e5636a81ea65fffda/efficientnetv2/effnetv2_model.py)
The  ["survival_probability"](https://github.com/keras-team/keras/blob/d8fcb9d4d4dad45080ecfdd575483653028f8eda/keras/applications/efficientnet_v2.py#L990) is defined as `survival_probability=drop_connect_rate * b / blocks` but b is set to zero, while according to the [google implementation](https://github.com/google/automl/blob/387d5ddb92bb8fbbec4b012e5636a81ea65fffda/efficientnetv2/effnetv2_model.py#L619) it should increase with the "number" of the block.

I think that line should be replaced with:
`survival_probability=drop_connect_rate * i / blocks` 
where i is the counter of the for loop

sebastian-sz:
Closes #16144 

It was found that the `survival_probability` of Efficientnet V2 networks was always 0.0, where it should be slowly increasing with each block size. The reason was I forgot to increase `b` parameter in those networks.

Thanks @SergioG-M for the find.

### DIFF
--- a/keras/applications/efficientnet_v2.py
+++ b/keras/applications/efficientnet_v2.py
@@ -991,6 +991,7 @@ def EfficientNetV2(
           name="block{}{}_".format(i + 1, chr(j + 97)),
           **args,
       )(x)
+      b += 1
 
   # Build top
   top_filters = round_filters(


fchollet (commented):
Thanks for the PR. Weight files are not affected by the change?

sebastian-sz:
@fchollet I just checked and the weights seem to be loading correctly. The only difference in the models is that, there are more Dropout layers, no new variables have been introduced.

fchollet (approved):
LGTM, thanks.

Issue #16146, opened 2022-03-01
`train_on_batch` allows different data types for input and target


jvishnuvardhan:
As mentioned in Keras webpage https://keras.io/api/models/model_training_apis/, keras shouldn't allow different data types for input (`x`) and target (`y`).

```
import tensorflow as tf
import numpy as np
x_np = np.random.uniform(size=(32, 10))
y_np = np.random.uniform(size=(32, 1))
x_tf = tf.convert_to_tensor(x_np)
y_tf = tf.convert_to_tensor(y_np)
data={'x_dict':np.random.uniform(size=(32, 10)),'y_dict':np.random.uniform(size=(32, 1))}

from tensorflow.keras import Input, Model
from tensorflow.keras.layers import Dense

input = Input(shape=(10,), name='input')
output = Dense(1, name='dense')(input)
model = Model(inputs=[input], outputs=[output])
model.compile(optimizer='adam', loss='mse')
# model.summary()
```

After defining the model, i ran `train_on_batch` as follows

```
model.train_on_batch(x_np, y_np)  # works
model.train_on_batch(x_tf, y_tf)  # works
model.train_on_batch(data['x_dict'], data['y_dict']) # works

model.train_on_batch(x_np, y_tf) # this should break but works
model.train_on_batch(x_tf, y_np) # this should break but works
model.train_on_batch(data['x_dict'], y_tf) # this should break but works
```
Last three cases shouldn't work but all the above cases works.

This could be either update the code or update the docs to reflect updated functionality.

[Here](https://colab.research.google.com/gist/jvishnuvardhan/398846ebe6d2f583dd6045638bffdd9d/untitled1176.ipynb) is a gist for reference.

PR #16147, opened 2022-03-01
Updated args of train_on_batch method

jvishnuvardhan:
As mentioned in Keras webpage https://keras.io/api/models/model_training_apis/, keras shouldn't allow different data types for input (x) and target (y). But, currently it takes different types for input and target. For example, input can be a numpy array and target can be a tensor.

Fixes https://github.com/keras-team/keras/issues/16146


jvishnuvardhan (commented):
Updated based on your suggestion. Thanks @fchollet

fchollet (approved):
LGTM, thanks!

PR #16165, opened 2022-03-04
Fix typo in docstring for `DenseFeatures`

gadagashwini:
Replace with the appropriate syntax in a given example tf.keras.layers.DenseFeatures


fchollet (approved):
LGTM, thanks

PR #16221, opened 2022-03-10
fix-grammar

Kyrpel:
Replace "or" with "and" for better readability


gbaned:
We will not be encouraging one liner grammatical changes as this is expensive process, thank you for your interest.
CC @fchollet @qlzh727

Issue #16201, opened 2022-03-09
exclude_from_weight_decay for AdamW


markub3327:
Hello, 

  please can you add an `exclude_from_weight_decay` parameter to AdamW? Some parameters during weight decay must be excluded.

For more details please look [here](https://github.com/tensorflow/addons/issues/2681).

Thanks a lot.
Have a nice day.

chunduriv:
@markub3327,

Can you please elaborate about your feature and please specify the use cases for this feature. Thanks!

markub3327:
I need to exclude some parameters from the weight decay process. Useful during training the Transformer model. List of regex patterns of variables excluded from weight decay. Variables whose names contain a substring matching the pattern will be excluded.

[Here](https://github.com/microsoft/unilm/blob/master/beit/semantic_segmentation/backbone/beit.py#L415) is the use case.

The code behind exclude_from_weight_decay from tensorflow/addons is [here](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/optimizers/weight_decay_optimizers.py#L245).

jvishnuvardhan:
@markub3327 What is the difference between this issue when compared to https://github.com/keras-team/keras/issues/16195?

Can you please state the differences clearly or close this issue this is a duplicate of https://github.com/keras-team/keras/issues/16195. Thanks!

bhack:
This is a proposal to add `exclude_from_weight_decay` in AdamW that was added in Keras 30 days ago (it was already in TFA).

https://github.com/keras-team/keras/issues/16195 is related to port  RectifiedAdam (a new optimizer) with `exclude_from_weight_decay`  in Keras from TFA

markub3327:
@bhack  

Thanks a lot ... Very good explanation!

haifeng-jin:
@chenmoneygithub Any comments on this?

chenmoneygithub:
@markub3327 Thanks for reporting the issue! Are you willing to open a PR to add `exclude_from_weight_decay` to AdamW? Thanks!

markub3327:
Yes. Thanks.

PR #16274, opened 2022-03-20
Add `exclude_from_weight_decay` to AdamW

markub3327:
Hello,
  
  @bhack

  there are changes for applying `exclude_from_weight_decay` to AdamW (#16201).

Thanks.


keras/optimizers/optimizer_experimental/adamw.py
```diff
@@ -173,7 +181,7 @@ def update_step(self, gradient, variable):
     alpha = (lr * tf.sqrt(1 - beta_2_power) / (1 - beta_1_power))
 
     # Apply step weight decay
-    if self.weight_decay != 0:
+    if self.weight_decay and self._do_use_weight_decay(variable):
```
markub3327, line 184: The better than used `if self.weight_decay != 0:` is `if self.weight_decay:` becouse user can set weight_decay=None and it's not a number (float) like 0, but it's empty object.

fchollet (commented):
Thanks for the PR!

No variable name should ever be used in the public API surface, because variable names are not stable.

Instead, this configuration parameter should be an actual list of variables (e.g. `[layer.bias for layer in my_dense_layers]`), which will be checked by identity.

Such a configuration parameter cannot be serialized and is specific to the set of weights that the optimizer will train, which is not defined at optimizer creation time, but only in `build()`. As such **it cannot be a constructor argument**. Instead, this should be configured via a method.

Let's add it to `build()`:

`def build(self, var_list, exclude_from_weight_decay=None)`

Then you can do:

```
model = ...
optimizer = AdamW()
optimizer.build(model.trainable_weights, exclude_from_weight_decay=get_bias_terms(model))
```

We could also have a separate method:

```
optimizer.exclude_from_weight_decay(var_list)
```

Does that make sense?

markub3327:
@fchollet 

Yes. I'll make a changes. Names of variables can be changed dynamicaly, but variables have static pointers during existing Model object....

Thanks.

keras/optimizers/optimizer_experimental/adamw.py
```diff
@@ -136,6 +136,7 @@ def build(self, var_list):
     if hasattr(self, '_built') and self._built:
       return
     self._built = True
+    self._exclude_from_weight_decay = exclude_from_weight_decay
```
fchollet, line 139: `exclude_from_weight_decay or []` (need to support the default `None` case)

keras/optimizers/optimizer_experimental/adamw.py
```diff
@@ -215,6 +216,8 @@ def get_config(self):
     })
     return config
 
+  def exclude_from_weight_decay(self, var_list):
+    self._exclude_from_weight_decay = var_list
```
fchollet, line 220: Need to check that the optimizer has not already been built, and raise an error otherwise. Otherwise calling this might not have any effect.

Also need to cast to list.

fchollet (commented):
Thanks for the update! Please add a correctness test for the change.

keras/optimizers/optimizer_experimental/adamw.py
markub3327: For the future you can define it as `exclude_from_weight_decay: Optional[List[str]] = None` and I added the None case solution.

keras/optimizers/optimizer_experimental/adamw.py
markub3327: Ok... You have a true. I created a raise an error.

Thanks.

keras/optimizers/optimizer_experimental/adamw.py
```diff
@@ -215,6 +221,15 @@ def get_config(self):
     })
     return config
 
+  def exclude_from_weight_decay(self, var_list):
+    if self._built:
+      if var_list == None:
+        self._exclude_from_weight_decay = []
+      else:
+        self._exclude_from_weight_decay = var_list
```
fchollet, line 229: You can replace these 4 lines with

```
self._exclude_from_weight_decay = var_list or []
```

keras/optimizers/optimizer_experimental/adamw.py
```diff
@@ -215,6 +221,15 @@ def get_config(self):
...
+    else:
+      raise ValueError('This optimizer has not yet been built. '
```
fchollet, line 231: The check should work the other way around: the user need to call `exclude_from_weight_decay` before the optimizer has been built.

fchollet (changes requested):
Thanks for the update!

keras/optimizers/optimizer_experimental/adamw.py
markub3327: Ok.

keras/optimizers/optimizer_experimental/adamw.py
markub3327: Sorry.

fchollet (approved):
LGTM, thanks!

keras/optimizers/optimizer_experimental/adamw.py
```diff
@@ -173,7 +174,7 @@ def update_step(self, gradient, variable):
     alpha = (lr * tf.sqrt(1 - beta_2_power) / (1 - beta_1_power))
 
     # Apply step weight decay
-    if self.weight_decay != 0:
+    if self.weight_decay and variable not in self._exclude_from_weight_decay:
```
old-school-kid, line 177: self.weight_decay can not be used as a pythonic bool. This raises the error 

> OperatorNotAllowedInGraphError: Using a symbolic `tf.Tensor` as a Python `bool` is not allowed: AutoGraph did convert this function. This might indicate you are trying to use an unsupported feature

`self.weight_decay != 0` could be a solution.

fchollet: Also, please add a correctness test -- this sort of issue should be caught in unit tests.

keras/optimizers/optimizer_experimental/optimizer_test.py
```diff
@@ -281,6 +281,8 @@ def testSaveAndLoadOptimizerWithModel(self, optimizer_fn):
         [keras.layers.Input(shape=(1,)),
          keras.layers.Dense(1)])
     optimizer = optimizer_fn()
+    if type(optimizer) is adamw_new.AdamW:
```
fchollet, line 284: - Use `isinstance` instead of `type(...) is`
- This test checks that the code runs but does not check that the code is correct. Please add a correctness test.

fchollet (changes requested):

markub3327:
@fchollet 

I cannot run test:
```shell
martin@server-01:~/Desktop/Projects/keras$ bazel test keras/optimizers/optimizer_experimental:optimizer_test
Starting local Bazel server and connecting to it...
WARNING: The following configs were expanded more than once: [v2]. For repeatable flags, repeats are counted twice and may lead to unexpected behavior.
INFO: Options provided by the client:
  Inherited 'common' options: --isatty=1 --terminal_columns=111
INFO: Reading rc options for 'test' from /home/martin/Desktop/Projects/keras/.bazelrc:
  Inherited 'build' options: --apple_platform_type=macos --define open_source_build=true --define=use_fast_cpp_protos=false --define=tensorflow_enable_mlir_generated_gpu_kernels=0 --define=allow_oversize_protos=true --spawn_strategy=standalone -c opt --announce_rc --define=grpc_no_ares=true --config=short_logs --config=v2
INFO: Reading rc options for 'test' from /home/martin/Desktop/Projects/keras/.bazelrc:
  'test' options: --define open_source_build=true --define=use_fast_cpp_protos=false --config=v2
INFO: Found applicable config definition build:short_logs in file /home/martin/Desktop/Projects/keras/.bazelrc: --output_filter=DONT_MATCH_ANYTHING
INFO: Found applicable config definition build:v2 in file /home/martin/Desktop/Projects/keras/.bazelrc: --define=tf_api_version=2 --action_env=TF2_BEHAVIOR=1
INFO: Found applicable config definition build:v2 in file /home/martin/Desktop/Projects/keras/.bazelrc: --define=tf_api_version=2 --action_env=TF2_BEHAVIOR=1
INFO: Repository zlib instantiated at:
  /home/martin/Desktop/Projects/keras/WORKSPACE:39:14: in <toplevel>
  /home/martin/.cache/bazel/_bazel_martin/b46d382d358df92d258999df19e91943/external/com_google_protobuf/protobuf_deps.bzl:9:21: in protobuf_deps
Repository rule http_archive defined at:
  /home/martin/.cache/bazel/_bazel_martin/b46d382d358df92d258999df19e91943/external/bazel_tools/tools/build_defs/repo/http.bzl:336:31: in <toplevel>
WARNING: Download from https://zlib.net/zlib-1.2.11.tar.gz failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 404 Not Found
ERROR: An error occurred during the fetch of repository 'zlib':
	File "/home/martin/.cache/bazel/_bazel_martin/b46d382d358df92d258999df19e91943/external/bazel_tools/tools/build_defs/repo/http.bzl", line 111, column 45, in _http_archive_impl
		download_info = ctx.download_and_extract(
Error in download_and_extract: java.io.IOException: Error downloading [https://zlib.net/zlib-1.2.11.tar.gz] to /home/martin/.cache/bazel/_bazel_martin/b46d382d358df92d258999df19e91943/external/zlib/temp14338963762586419969/zlib-1.2.11.tar.gz: GET returned 404 Not Found
ERROR: Error fetching repository: Traceback (most recent call last):
	File "/home/martin/.cache/bazel/_bazel_martin/b46d382d358df92d258999df19e91943/external/bazel_tools/tools/build_defs/repo/http.bzl", line 111, column 45, in _http_archive_impl
		download_info = ctx.download_and_extract(
Error in download_and_extract: java.io.IOException: Error downloading [https://zlib.net/zlib-1.2.11.tar.gz] to /home/martin/.cache/bazel/_bazel_martin/b46d382d358df92d258999df19e91943/external/zlib/temp14338963762586419969/zlib-1.2.11.tar.gz: GET returned 404 Not Found
INFO: Repository remote_coverage_tools instantiated at:
  /DEFAULT.WORKSPACE.SUFFIX:11:13: in <toplevel>
Repository rule http_archive defined at:
  /home/martin/.cache/bazel/_bazel_martin/b46d382d358df92d258999df19e91943/external/bazel_tools/tools/build_defs/repo/http.bzl:336:31: in <toplevel>
ERROR: /home/martin/.cache/bazel/_bazel_martin/b46d382d358df92d258999df19e91943/external/com_google_protobuf/BUILD:154:11: @com_google_protobuf//:protobuf depends on @zlib//:zlib in repository @zlib which failed to fetch. no such package '@zlib//': java.io.IOException: Error downloading [https://zlib.net/zlib-1.2.11.tar.gz] to /home/martin/.cache/bazel/_bazel_martin/b46d382d358df92d258999df19e91943/external/zlib/temp14338963762586419969/zlib-1.2.11.tar.gz: GET returned 404 Not Found
ERROR: Analysis of target '//keras/optimizers/optimizer_experimental:optimizer_test' failed; build aborted: Analysis failed
INFO: Elapsed time: 15.409s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (64 packages loaded, 687 targets configured)
FAILED: Build did NOT complete successfully (64 packages loaded, 687 targets configured)
martin@server-01:~/Desktop/Projects/keras$
```

markub3327:
@fchollet

1. The lower value of `self.weight_decay` the higher impact of weight decay on the variable - inverse proportion !!!
The intuitivly way is the higher value of `self.weight_decay` the higher impact of weight decay on the variable.

2. Assigning of `self._exclude_from_weight_decay` at `build()` overwrite what was assigned at `exclude_from_weight_decay()`.

fchollet:
> The lower value of self.weight_decay the higher impact of weight decay on the variable - inverse proportion !!! The intuitivly way is the higher value of self.weight_decay the higher impact of weight decay on the variable.

What do you mean exactly? Is there something you would change in the current implementation?

markub3327:
@fchollet 

Yes, look here.

The result with `AdamW(learning_rate=0.001, weight_decay=0.001)`:
> <tf.Variable 'Variable:0' shape=() dtype=float32, numpy=2.026558e-06>

The result with `AdamW(learning_rate=0.001, weight_decay=1.0)`:
> <tf.Variable 'Variable:0' shape=() dtype=float32, numpy=0.0019999743>

The above you can see the higher weight_decay the less impact of weight decay.

In the [original paper](https://arxiv.org/pdf/1711.05101.pdf) the equation is:
Snímka obrazovky 2022-03-31 o 8 01 25

But in the code the equation is:
`variable.assign_sub(variable * (1 - lr * wd))`

I think the better is (like in the original paper):
`variable.assign_sub(variable * wd)`

fchollet:
We should decide whether we want to go with equation A or B:

Equation A: `variable.assign_sub(variable * (1 - lr * wd))`

Equation B: `variable.assign_sub(variable * wd)`

What are the arguments for each?

Apparently A is what PyTorch uses (and our current implementation). B is what the paper specifies and also what TF Addons uses. @markub3327 you seem to be saying that B is more intuitive.

This code was authored  by @yarri-oss -- Douglas, what do you think?

yarri-oss:
@markub3327 thank you again for your feedback. As @fchollet mentioned, we've seen arguments for each of the above equations. A) seems to be in use with other frameworks, and works with the assumption of smaller `weight_decay` on the same order as the `learning_rate`. B) is used with TF Addons, which pre-dates this Keras AdamW implementation and, as you say, is more intuitive with larger (ie., 10x ~ 100x the `lr`) values of `weight_decay`.

Are you experienced with the TF Addons AdamW implementation? Is that why you prefer the larger values of `weight_decay`?

Thanks!

bhack:
https://github.com/tensorflow/addons/issues/2282#issuecomment-742896830

bhack:
See also the author of the paper (Q2) https://github.com/loshchil/AdamW-and-SGDW/issues/1#issuecomment-345880370

chenmoneygithub:
@markub3327 This really depends on how we define `decay`. We are hitting this issue because the paper uses decay in an opposite way of learning rate decay. In this case, I would prefer sticking to the paper, which is larger weight decay means larger impact on model variables.

markub3327:
@chenmoneygithub

Okey. Did You say the preferred is option B): `variable.assign_sub(variable * wd)`?

I can make a change in the code.

fchollet:
In light of the evidence so far I personally find option B `variable.assign_sub(variable * wd)` more convincing (especially given that the general behavior is more consistent with how we define weight decay or learning rate decay elsewhere in the API). Is this the consensus view?

chenmoneygithub:
Yea, following option B sounds good to me.

markub3327:
All is done.

namas191297:
Nnnznn zzz zzz b kk n BB. Fy in nu BB

fchollet:
@yarri-oss does the change sound good?

chenmoneygithub (approved):
Thanks for the PR!

chenmoneygithub:
The implementation looks correct to me.

fchollet (approved):

fchollet:
Re-running tests. Previous run appeared to be flaky.

edwardyehuang:
Any chance to include this in 2.9 rc0? This is a very important feature.
@fchollet @qlzh727

PR #16398, opened 2022-04-11
adds split_dataset utility 

prakashsellathurai:
mentioned in #16394

### DIFF
--- a/keras/utils/dataset_utils.py
+++ b/keras/utils/dataset_utils.py
@@ -280,6 +280,8 @@ def _restore_dataset_from_list(dataset_as_list,dataset_type_spec,
           else:
             restored_dataset[k].append(np.array(v))
       return restored_dataset
+    else:
+      return tuple(np.array(sample) for sample in zip(*dataset_as_list))
   return dataset_as_list
 
 def _rescale_dataset_split_sizes(left_size,right_size,total_length):
--- a/keras/utils/dataset_utils.py
+++ b/keras/utils/dataset_utils.py
@@ -78,13 +78,21 @@ def split_dataset(dataset,
                                                       total_length)
   left_split = list(dataset_as_list[:left_size])
   right_split = list(dataset_as_list[-right_size:])
-
-  left_split = _restore_dataset_from_list(left_split,dataset_type_spec)
-  right_split = _restore_dataset_from_list(right_split,dataset_type_spec)
+  
+  left_split = _restore_dataset_from_list(left_split,dataset_type_spec,dataset)
+  right_split = _restore_dataset_from_list(right_split,dataset_type_spec,
+                                           dataset)
 
   left_split = tf.data.Dataset.from_tensor_slices(left_split)
   right_split = tf.data.Dataset.from_tensor_slices(right_split)
 
+  # Batch the splits if the `dataset` is batched
+  if dataset_type_spec is tf.data.Dataset and is_batched(dataset):   
+    batch_size = get_batch_size(dataset)
+    if batch_size is not None:
+      left_split = left_split.batch(batch_size)
+      right_split = right_split.batch(batch_size)
+  
   left_split = left_split.prefetch(tf.data.AUTOTUNE)
   right_split = right_split.prefetch(tf.data.AUTOTUNE)
 
@@ -112,8 +120,6 @@ def _convert_dataset_to_list(dataset,
   Returns:
       List: A list of tuples/numpy arrays.
   """
-  # TODO (prakashsellathurai): add support for Batched  and unbatched dict tf datasets
-  
   dataset_iterator = _get_data_iterator_from_dataset(dataset,dataset_type_spec)
   dataset_as_list = []
 
@@ -259,12 +265,22 @@ def _get_next_sample(dataset_iterator,
           data_size_warning_flag = False
     yield sample
 
-def _restore_dataset_from_list(dataset,dataset_type_spec):
-  """Restore the dataset from the list of arrays.
-  """
+def _restore_dataset_from_list(dataset_as_list,dataset_type_spec,
+                               original_dataset):
+  """Restore the dataset from the list of arrays."""  
   if dataset_type_spec  in [tuple,list]:
-    dataset = tuple(np.array(sample) for sample in zip(*dataset))
-  return dataset
+    return tuple(np.array(sample) for sample in zip(*dataset_as_list))
+  elif dataset_type_spec == tf.data.Dataset:
+    if type(original_dataset.element_spec) is dict:
+      restored_dataset = dict()
+      for d in dataset_as_list:
+        for k, v in d.items():
+          if k not in restored_dataset:
+            restored_dataset[k] = [np.array(v)]
+          else:
+            restored_dataset[k].append(np.array(v))
+      return restored_dataset
+  return dataset_as_list
 
 def _rescale_dataset_split_sizes(left_size,right_size,total_length):
   """Rescale the dataset split sizes to ensure that the sum of 
@@ -383,6 +399,13 @@ def is_batched(tf_dataset):
     return tf_dataset.__class__.__name__ == 'BatchDataset'
   except :
     return False
+  
+def get_batch_size(tf_dataset):
+  """Get the batch size of the dataset."""
+  if is_batched(tf_dataset):
+    return tf_dataset._batch_size
+  else:
+    return None
 
 def index_directory(directory,
                     labels,
--- a/keras/utils/dataset_utils.py
+++ b/keras/utils/dataset_utils.py
@@ -58,7 +58,7 @@ def split_dataset(dataset,
       A tuple of two `tf.data.Dataset` objects: the left and right splits.
   """
   
-  if not isinstance(dataset,(tf.data.Dataset,list,tuple)):
+  if not isinstance(dataset,(tf.data.Dataset,list,tuple,np.ndarray)):
     raise TypeError('`dataset` must be either a tf.data.Dataset object'
                    f' or a list/tuple of arrays. Received : {type(dataset)}')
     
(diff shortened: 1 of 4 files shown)


keras/utils/dataset_utils.py
```diff
@@ -14,14 +14,137 @@
...
+  
+  if dataset and not isinstance(dataset,(tf.data.Dataset,list)):
+    raise TypeError('`dataset` must be either a tf.data.Dataset object'
+                     'or a list/tuple of arrays',f'Got {type(dataset)}')
+    
+  dataset_as_list = []
+  data_size_warning_flag = False
+  start_time = time.time()
+  
+  for datum in list(dataset):
```
fchollet: I'm noticing that we should handle the case where the dataset is a `tf.data.Dataset` that is already batched. In such cases we should iterate over the batches, split the dataset at the batch level, then merge the batches into new datasets by concatenating on the 0th axis. That way you get two batched datasets with the same batch size as the original one.

It is not possible (I think?) to tell in advance whether a dataset is already batched or not, so we would have to add an argument for it I think, e.g. `batched=False`.

keras/utils/dataset_utils.py
```diff
@@ -14,14 +14,137 @@
...
+    cur_time = time.time()
+    # warns user if the dataset is too large to iterate within 10s
+    if int(cur_time - start_time) > 10 and not data_size_warning_flag:
```
fchollet: - `data_size_warning_flag` should come first so we have less work to do in the loop.
- We should only check every ~10 steps to minimize overhead, e.g.

```python
if data_size_warning_flag:
    if i % 10 == 0:
        current_time = time.time()
        if ...
```

keras/utils/dataset_utils.py
```diff
@@ -14,14 +14,137 @@
...
+      Warning('Takes too long to process the dataset,'
+              'the utility is only meant for small datasets that fit in memory')
+      data_size_warning_flag = True
+    dataset_as_list.append(datum)
+  
+  if shuffle: 
+    if seed:
+      np.random.seed(seed)
```
fchollet, line 91: Don't modify the global random seed, use a random_state for seeding (and reuse it if we have multiple arrays)

keras/utils/dataset_utils.py
fchollet: This `for` loop will not work in case where we have a list of arrays as `dataset`.

There should be a different loop for the array / list of arrays case and for the `Dataset` case.

keras/utils/dataset_utils.py
```diff
@@ -14,14 +14,137 @@
...
+    np.random.shuffle(dataset_as_list)
+    
+  total_size = len(dataset_as_list)
+   
+  # if both left_size and right_size are None, raise error
+  if right_size is None and left_size is None:
+    raise ValueError('Both `left_size` and `right_size`cannot be `None`'
```
fchollet: Fix typos

keras/utils/dataset_utils.py
```diff
@@ -14,14 +14,137 @@
...
+                     ' either one of them must specified')
+    
+  # if left_size is None, it defaults to the complement to right_size
+  # raises error if right_size not in range [0, 1] or [0, total_size]
+  elif left_size is None and right_size:
+    if type(right_size) == float:
+      if right_size < 0 or right_size > 1:
+        raise ValueError('`right_size` must be in range `[0, 1]`')
```
fchollet: All error messages should specify what the value specified by the user was (for easier debugging)

keras/utils/dataset_utils.py
fchollet: Note: adding unit test cases will help you get this right. Test for:

- `dataset=np.ones(shape=(200, 32))`
- `dataset=[np.ones(shape=(200, 32)), np.zeros(shape=(200, 32))]`
- `dataset=[np.ones(shape=(200, 32)), np.zeros(shape=(100, 32))]` -> should raise an error
- `dataset=(np.ones(shape=(200, 32)), np.zeros(shape=(200, 32)))`
- Batched tf.data.Dataset that yields batches of vectors
- Batched tf.data.Dataset that yields batches of tuples of vectors
- Batched tf.data.Dataset that yields batches of dicts of vectors
- Unbatched tf.data.Dataset that yields vectors
- Unbatched tf.data.Dataset that yields tuples of vectors
- Unbatched tf.data.Dataset that yields dicts of vectors

fchollet (commented):
Thanks for the PR!

Adding unit tests will help you make the use cases more concrete.

keras/utils/dataset_utils.py
```diff
@@ -14,14 +14,328 @@
...
+  left_split = list(dataset_as_list[:left_size])
+  right_split = list(dataset_as_list[-right_size:])
+
+  if _get_type_spec(dataset) is tuple:
+    left_split = tuple(np.array(sample) for sample in zip(*left_split))
+    right_split = tuple(np.array(sample) for sample in zip(*right_split))
+  
+  try:
+    left_split = tf.data.Dataset.from_tensor_slices(left_split)
+  except Exception as e:
```
fchollet, line 92: Please except a specific exception -- not a generic Exception. What's the exception that will be raised here?

keras/utils/dataset_utils.py
```diff
@@ -14,14 +14,328 @@
...
+    raise ValueError(f' with `left_size`={left_size} and '
+                     f' `right_size`={right_size}'
+                     f' unable to create the dataset from '
+                     f' left_split of shape {np.array(left_split).shape}. '
+                     f' Received error: {e}')
+  
+  try:
+    right_split = tf.data.Dataset.from_tensor_slices(right_split)
+  except Exception as e:
```
fchollet, line 101: Likewise.

keras/utils/dataset_utils_test.py
```diff
@@ -0,0 +1,357 @@
...
+
+from keras.utils import dataset_utils
+from keras.datasets import mnist
+
+
+class SplitDatasetTest(tf.test.TestCase):
+  
+  def test_numpy_array(self):
+    dataset=np.ones(shape=(200, 32))
+    res = dataset_utils.split_dataset(dataset, left_size=0.8,right_size=0.2)
```
fchollet, line 15: There are various code formatting issues, please use a linter. See: https://pylint.pycqa.org/en/latest/user_guide/run.html

keras/utils/dataset_utils_test.py
```diff
@@ -0,0 +1,357 @@
...
+
+    left_split,right_split=dataset_utils.split_dataset(dataset,left_size=5)
+    
+    self.assertAllEqual(np.array(list(left_split)).shape,(5,2,32,32,1))
+    self.assertAllEqual(np.array(list(right_split)).shape,(5,2,32,32,1))
+     
+  
+    self.assertAllEqual(list(dataset),list(left_split)+list(right_split)) 
+    
+  # def test_mnist_dataset(self):
```
fchollet, line 153: Please remove commented-out code

keras/utils/dataset_utils_test.py
```diff
@@ -0,0 +1,357 @@
...
+    
+    self.assertLen(left_split, 160)
+    self.assertLen(right_split, 40)
+    self.assertAllEqual(dataset[:160] ,list(left_split))
+    self.assertAllEqual(dataset[-40:] ,list(right_split))
+    
+    
+  def test_list_of_numpy_arrays(self):
+    dataset=[np.ones(shape=(200, 32)), np.zeros(shape=(200, 32))]
+    res = dataset_utils.split_dataset(dataset, left_size=4)
```
fchollet, line 29: Please also add a case for fractional sizes.

keras/utils/dataset_utils.py
```diff
@@ -14,14 +14,328 @@
...
+  right_split = right_split.prefetch(tf.data.AUTOTUNE)
+  
+  return left_split, right_split
+
+def _convert_dataset_to_list(dataset,
+                             data_size_warning_flag= True,
+                             ensure_shape_similarity = True):
+  """Helper function to convert a tf.data.Dataset  object or a list/tuple of numpy.ndarrays to a list
+  """
+  # TODO (prakashsellathurai): fix issue with tuple of numpy arrays dataset=(np.ones(shape=(3, 2)), np.zeros(shape=(3,)))
```
fchollet, line 118: What the issue?

keras/utils/dataset_utils.py
```diff
@@ -14,14 +14,328 @@
...
+  # TODO (prakashsellathurai): add support for Batched  and unbatched dict tf datasets
```
fchollet, line 119: Please add dict support.

keras/utils/dataset_utils_test.py
```diff
@@ -0,0 +1,357 @@
...
+    left_split,right_split = dataset_utils.split_dataset(dataset,
+                                                             left_size=0.35,
+                                                             right_size=0.65)
+    self.assertLen(left_split, 4)
+    self.assertLen(right_split, 6)
+    self.assertIsInstance(left_split, tf.data.Dataset)
+    self.assertIsInstance(right_split, tf.data.Dataset)
+    
+  def test_invalid_dataset(self):
+    with self.assertRaises(TypeError):
```
fchollet, line 217: Any time you test an error, use `assertRaisesRegex` in order to make sure the exception being raised in the one you were expecting. Otherwise it may be a different one!

keras/utils/dataset_utils.py
```diff
@@ -14,14 +14,328 @@
...
+  right_split = list(dataset_as_list[-right_size:])
+
+  if _get_type_spec(dataset) is tuple:
+    left_split = tuple(np.array(sample) for sample in zip(*left_split))
+    right_split = tuple(np.array(sample) for sample in zip(*right_split))
+  
+  try:
+    left_split = tf.data.Dataset.from_tensor_slices(left_split)
+  except Exception as e:
+    raise ValueError(f' with `left_size`={left_size} and '
```
fchollet, line 93: Please reformat the error messages:

- Should start with a capital letter
- No need for a space in the beginning
- Should be a grammatically correct sentence
- When using "name=something", don't use backticks

fchollet (commented):
Thanks for the update!

fchollet:
Please let me know when you'd like me to review (and you can also graduate the PR from draft to ready).

prakashsellathurai:
@fchollet I Incorporated the requested changes. The code passes both Pylint and Bazel tests. The pull request is now ready for review.

keras/utils/dataset_utils.py
```diff
@@ -19,9 +19,395 @@
...
+  """Restore the dataset from the list of arrays."""
+  if dataset_type_spec  in [tuple,list]:
+    return tuple(np.array(sample) for sample in zip(*dataset_as_list))
+  elif dataset_type_spec == tf.data.Dataset:
+    if type(original_dataset.element_spec) is dict:
+      restored_dataset = dict()
+      for d in dataset_as_list:
+        for k, v in d.items():
+          if k not in restored_dataset:
+            restored_dataset[k] = [np.array(v)]
```
fchollet, line 279: Why `np.array`? Would that work with string types?

keras/utils/dataset_utils_test.py
```diff
@@ -0,0 +1,440 @@
...
+      dataset_utils.split_dataset(np.array([1,2,3]),left_size=0,
+                                  right_size='1')
+
+    expected_regex = (r'^.*?(\bInvalid `left_size` Type\b)')
+    with self.assertRaisesRegex(TypeError,expected_regex):
+      dataset_utils.split_dataset(np.array([1,2,3]),left_size='100',
+                                  right_size=None)
+
+    expected_regex = (r'^.*?(\bInvalid `right_size` Type\b)')
+    with self.assertRaisesRegex(TypeError,expected_regex):
```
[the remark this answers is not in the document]
prakashsellathurai: for some reasons `pylint` didn't raise any warning for this issue.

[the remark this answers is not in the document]
prakashsellathurai: It seems pylint removed `bad whitespace` rule mentioned [here](https://github.com/PyCQA/pylint/issues/3572).

keras/utils/dataset_utils.py
prakashsellathurai: Thanks for pointing it out. `np.array()` will lead to unnecessary type conversion. I removed it

keras/utils/dataset_utils_test.py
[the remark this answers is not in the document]
prakashsellathurai: `flake8` resolved thsi issue for me

keras/utils/dataset_utils.py
```diff
@@ -19,9 +19,396 @@
...
+    if data_size_warning_flag:
+      if i % 10 == 0:
+        cur_time = time.time()
+        # warns user if the dataset is too large to iterate within 10s
+        if int(cur_time - start_time) > 10 and data_size_warning_flag:
+          warnings.warn('The dataset is taking longer than 10 seconds to '
+                        'iterate. This may be due to the size of the dataset. '
+                        'Please consider using a smaller dataset'
+                        '(e.g. < 10,000 samples). \nTo hide this '
+                        'warning message, set `data_size_warning_flag=False`.',
```
fchollet, line 262: This argument is not public-facing, so we should not reference it in the warning (which is public-facing)

fchollet (approved):
Excellent, I think we're good! Thanks for the update.

I added a comment but we can probably just fix it in post.

fchollet (approved):

PR #16447, opened 2022-04-21
Fix Syntax error for combined_model.compile of WideDeepModel

gadagashwini:
combined_model.compile() missing keyword arguments such as `loss` and `metrics`. 
This PR Fixes `SyntaxError: positional argument follows keyword argument`


fchollet (approved):
LGTM, thanks

PR #16448, opened 2022-04-21
Cherrypick image KPL related fix for labels to r2.9

qlzh727:
Piper cl:443122383
Piper cl:443446864


qlzh727:
It seems that the test failed with random_zoom layer, which was updated in cl/438675482 and not being cherrypicked. I think it will be a bit risky for cl/438675482 to be included.

qlzh727:
Actually only the randomBrightness is safe for cherrypick. I am going to close this PR and create a new one.

PR #16454, opened 2022-04-22
Cherrypick ImageAugmetation related change to R2.9

qlzh727:
Piper cl: 443122383


qlzh727:
The presubmit failure is expected since the test was configured with run with tf-nightly, whereas it should run with tf 2.9 rc. The same test has been run when we merge the original change. 

We will run the same test again with tf 2.9 rc when we bake the keras 2.9 rc binary.

Issue #15239, opened 2021-08-24
3 Possible Essential Features for `model.summary()`


innat:
**System information**.

TensorFlow version (you are using): 2.4.1
Are you willing to contribute it (Yes/No): No

---

**Describe the feature and the current behavior/state**.

Total 3 Feature Request, illustrated below. 

## [Feature Request 1]: Add Model Memory Usage
Many times it's needed to know about the **consumed size of GPU memory** for DL models to avoid potential **out-of-memory** errors either at the beginning of training or in the middle. 

So in the `model. summary`, (we think) it would be better to have such functionalities that would give a **rough estimation** of memory usage by the DL model.

We've found a nice solution from @ZFTurbo on [here](https://stackoverflow.com/a/46216013/9215780), that addresses such a request. And I think it would be nice to have such a mechanism inside `model.summary(memory_usage_info = True / False)`. Here is the starting [gist](https://colab.research.google.com/drive/1FePCsRdutXNyTCiMGuXdXaB8QDj453R0?usp=sharing) for the contributor. 

### ~~[Feature Request 2]: Layer Range to Display Model Summary~~ It has been addressed. HERE https://github.com/keras-team/keras/pull/16458
<details>
<summary>details of feature request 2</summary>
<br>

Let's say, we want to **print the summary** of the following model and are interested to observe the information of trainable and non-trainable parameter counts only, maybe we simply want to print the model summary within a specific range ([like](https://stackoverflow.com/q/52651749/9215780)); in that case, it's not intuitive to get such **LONG output**. 

```python
from tensorflow.keras import Input, Model, Sequential
from tensorflow.keras import layers 

input = Input((224, 224, 3))
base =  keras.applications.MobileNetV3Small(weights=None,
                                             include_top = False, 
                                             input_tensor = input)
features     = layers.GlobalAveragePooling2D() (base.output)
classifier   = layers.Dense(10, activation="sigmoid")(features)
model        = Model(base.input, classifier)
model.summary()
# lots of layer gets print 
# .... 
```

Instead, a new parameters can be introduced such as `model.summary(layer_range=['layer_a', 'layer_z'])` to print sub-graph (like we have now for [`plot_model`](https://github.com/tensorflow/tensorflow/issues/48641)). Also, a parameter that only also `.summary` function to print the following output only:

```
# No Long Layer Prints ...
Total params: 1,540,218
Trainable params: 1,528,106
Non-trainable params: 12,112
```

[...]

PR #16458, opened 2022-04-24
Added layer_range parameter in model_summary

federicopisanu:
Hi, i worked on this PR that solves the 2nd suggestion in the issue #15239 (the last remaining). This is my first contribution to keras so i'm happy to have any feedback or suggestion. It's been quite easy to setup the development environment to run tests even if i am not an expert in OS contributions.

To solve this issue i took inspiration from [this](https://github.com/tensorflow/tensorflow/issues/48641) issue that implements the same feature for the `plot_model` function. I kept the same interface and a similar parameter description in docstrings to be consistent with the work that has been already done.
I also reused the function `get_layer_index_bound_by_layer_name` by importing it from `keras.utils.vis_utils`, im not sure if it's ok, maybe mantainers prefer to move this function to a separate file where both modules can import from. In that case i'm happy to follow up with the requested change.

I also added some test, just tell me if it's better to add some more test case.


keras/utils/layer_utils.py
```diff
@@ -21,6 +21,7 @@
 
 from keras.utils import io_utils
 from keras.utils import tf_inspect
+from keras.utils.vis_utils import get_layer_index_bound_by_layer_name
```
qlzh727, line 24: Please import the module only, rather than import the function directly.

keras/utils/layer_utils.py
```diff
@@ -140,6 +142,14 @@ def print_summary(model,
           If not provided, defaults to `False`.
       show_trainable: Whether to show if a layer is trainable.
           If not provided, defaults to `False`.
+      layer_range: input of type`list` containing two `str` items, which is the
```
qlzh727, line 145: space between type and `list`

keras/utils/layer_utils.py
```diff
@@ -334,7 +359,9 @@ def print_layer(layer, nested_level=0, is_nested_last=False):
                '|' * nested_level)
 
   layers = model.layers
-  for layer in layers:
+  for i, layer in enumerate(layers):
+    if (layer_range) and (i < layer_range[0] or i > layer_range[1]):
```
qlzh727, line 363: the bracket for layer_range is not needed.

keras/utils/layer_utils_test.py
```diff
@@ -370,6 +370,104 @@ def print_to_file(text):
...
+      print(text, file=writer)
+
+    try:
+      layer_utils.print_summary(model, print_fn=print_to_file, layer_range=["conv", "flat"])
+      self.assertTrue(tf.io.gfile.exists(fpath))
+      writer.close()
+      reader = open(fpath, 'r')
+      lines = reader.readlines()
+      reader.close()
+      self.assertEqual(len(lines), 13)
```
qlzh727, line 397: Could u add some comments about the 13 here? the test case below is quite clear that how many lines of output is expected.

keras/utils/layer_utils_test.py
```diff
@@ -370,6 +370,104 @@ def print_to_file(text):
...
+          'Trainable params: 18\n'
+          'Non-trainable params: 6\n'
+          '_________________________________________________________________\n')
+
+      fin_str = ''
+      for line in lines:
+        fin_str += line
+      print(fin_str)
+      self.assertIn(fin_str, check_str)
+      self.assertEqual(len(lines), 23)
```
qlzh727, line 467: Can we use len(check_str) here instead of 23?

qlzh727 (changes requested):

keras/utils/layer_utils.py
federicopisanu: This import was breaking many tests for other functionalities because they didn't find `vis_utils` in `keras.utils` so i decided to move the `get_layer_index_bound_by_layer_name` function to 'layer_utils'. This solved the import error but i had to add some change to the `keras.utils.vis_utils` module.

keras/utils/layer_utils.py
federicopisanu: Done, thanks.

keras/utils/layer_utils.py
federicopisanu: Done, thanks.

keras/utils/layer_utils_test.py
federicopisanu: Done, thanks.

keras/utils/layer_utils_test.py
federicopisanu: That was not working but i hope my new solution goes in the right direction.

keras/utils/layer_utils.py
```diff
@@ -200,6 +236,22 @@ def print_summary(model,
     positions.append(line_length)
     to_display.append('Trainable')
 
+  if layer_range is not None:
+    if len(layer_range) != 2:
```
qlzh727, line 240: I think we probably want to push those validation logic into get_layer_index_bound_by_layer_name as well. They are duplicated here and in vis_util.

keras/utils/layer_utils.py
```diff
@@ -334,7 +386,9 @@ def print_layer(layer, nested_level=0, is_nested_last=False):
                '|' * nested_level)
 
   layers = model.layers
-  for layer in layers:
+  for i, layer in enumerate(layers):
```
qlzh727, line 389: Shouldn't we just get `layer = model.layers[layer_range[0], layer_range[1]]`?

qlzh727 (changes requested):
Thanks for the change, and sorry for the late reply.

keras/utils/layer_utils.py
federicopisanu: Sure, having this freedom i also tried to simplify and remove some unnecessary validation logic by rewriting some lines in `get_layer_index_bound_by_layer_name` function.

keras/utils/layer_utils.py
federicopisanu: It'd be great. With last commit i changed the function output to allow this list slicing (adding 1 to the second layer_range index). I also had to change vis_utils.py accordingly. Thanks!

keras/utils/layer_utils_test.py
```diff
@@ -262,23 +262,32 @@ def print_to_file(text):
       reader.close()
       check_str = (
           'Model: '
-          '"trainable"\n____________________________________________________________________________\n'
+          '"trainable"\n'
```
qlzh727, line 265: Please change this part back (the existing format was more readable.)

keras/utils/layer_utils_test.py
```diff
@@ -330,35 +339,52 @@ def print_to_file(text):
       reader.close()
       check_str = (
           'Model: '
-          '"model_2"\n____________________________________________________________________________\n'
+          '"model_2"\n'
```
qlzh727, line 342: same here.

qlzh727 (approved):
Thanks for the change

qlzh727 (approved):

qlzh727 (changes requested):
I think you forget to update the model.summary() funciton.

federicopisanu:
> I think you forget to update the model.summary() funciton.

thanks!

qlzh727 (approved):

federicopisanu:
Just merged `master` back into my branch, it solves a bunch of tests that were failing. Hope it works!

federicopisanu:
Solved conflicts caused by `black` that has been performed on whole code base in `master`.

gbaned:
@federicopisanu  Can you please resolve conflicts? Thank you!

federicopisanu:
> @federicopisanu Can you please resolve conflicts? Thank you!

Done, Thank you!

federicopisanu:
Hi @qlzh727, thanks for the great support in this PR.
I was wondering if its correct that this change has not been added to the 2.10 pre-release changelog [here](https://github.com/keras-team/keras/releases/tag/v2.10.0-rc1).
There is something different from other features requested in #15239 or its just ok?

PR #16460, opened 2022-04-25
Update index_lookup.py

### LINKED ISSUE #16453: The vocabulary_size method of preprocessing layers does not work in graph mode
**System information**.
- Have I written custom code (as opposed to using a stock example script provided in Keras): **yes**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): **Colab**
- TensorFlow installed from (source or binary): **binary**
- TensorFlow version (use command below): **TF 2.8, nightly 2.10.0-dev20220422**

**Describe the problem**.

Using the `vocabulary_size()` method of preprocessing layers (like `tf.keras.layers.StringLookup`) fails in graph mode, because the implementation looks like this: https://github.com/keras-team/keras/blob/aea9728313bcaa8262774699c21976288171b209/keras/layers/preprocessing/index_lookup.py#L346-L352

Calling `vocabulary_size()` in graph mode fails.

Calling `vocabulary_size()` in graph mode succeeds and returns a `tf.Tensor` with the size.

Colab notebook showing the issue is at https://colab.research.google.com/drive/1Mq9G8eUvNLw6ykk4ARKf6jurLREAvRhu?usp=sharing both for TF 2.8 and TF nightly.

**Source code / logs**.

An obvious fix is to avoid using the `numpy()` method when in graph mode, i.e., instead of
```python
return int(self.lookup_table.size().numpy()) + self._token_start_index()
```
use just
```python
return self.lookup_table.size() + self._token_start_index()
```
Note that the above notebook also shows that this implementation works in the graph mode.

tilakrayal:
Update to `return self.lookup_table.size() + self._token_start_index()` from 
`return int(self.lookup_table.size().numpy()) + self._token_start_index()`. Fixes #16453.

### DIFF
--- a/keras/layers/preprocessing/index_lookup.py
+++ b/keras/layers/preprocessing/index_lookup.py
@@ -349,7 +349,10 @@ class IndexLookup(base_preprocessing_layer.PreprocessingLayer):
     Returns:
       The integer size of the voculary, including optional mask and oov indices.
     """
-    return self.lookup_table.size() + self._token_start_index()
+    if tf.executing_eagerly():
+      return int(self.lookup_table.size().numpy()) + self._token_start_index()
+    else:
+      return self.lookup_table.size() + self._token_start_index()
 
   def vocab_size(self):
     logging.warning("vocab_size is deprecated, please use vocabulary_size.")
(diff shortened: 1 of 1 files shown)


foxik:
Oh, I suggested this fix in the original issue, but I thought it would be used only in the graph mode -- in eager mode the original `.numpy()` should still be used, I believe.

LukeWood:
> Oh, I suggested this fix in the original issue, but I thought it would be used only in the graph mode -- in eager mode the original `.numpy()` should still be used, I believe.

Yeah this is a great idea to preserve backwards compatibility.  @tilakrayal can you add this?  Thanks!

tilakrayal:
As suggested, included `tf.executing_eagerly()`.

fchollet (approved):
LGTM, thank you

PR #16488, opened 2022-05-03
Fix EarlyStopping for patience = 0

### LINKED ISSUE #16393: Early stopping halts training after second epoch with default patience
The `EarlyStopping` callback monitors the number of epochs without improvement through the `wait` attribute.
Whenever there is improvement, this value is set to zero; when there is not, it is incremented by one.
However, the `patience` attribute of `EarlyStopping` objects is set to zero by default.
This means that the condition `self.wait >= self.patience` in [this line](https://github.com/keras-team/keras/blob/d8fcb9d4d4dad45080ecfdd575483653028f8eda/keras/callbacks.py#L1866) is always verified.
This triggers `EarlyStopping` after two epochs regardless of improvement when using the default patience.

gowthamkpr:
Fixed EarlyStopping problem for patience = 0 if there is improvement in accuracy or reduction in loss.

Fixes #16393

### DIFF
--- a/keras/callbacks.py
+++ b/keras/callbacks.py
@@ -1865,7 +1865,7 @@ class EarlyStopping(Callback):
         self.wait = 0
 
     # Only check after the first epoch.
-    if self.wait >= self.patience and epoch > 0:
+    if self.wait >= self.patience and self.wait != 0 and epoch > 0:
       self.stopped_epoch = epoch
       self.model.stop_training = True
       if self.restore_best_weights and self.best_weights is not None:


hertschuh:
@gowthamkpr Can you add a unit test exercising this new condition? Thanks!

gbaned:
@gowthamkpr Can you please resolve conflicts? Thank you!

gbaned:
@gowthamkpr Can you please resolve conflicts? Thank you!

valentas-kurauskas:
I propose to change the documentation
[here](https://github.com/keras-team/keras/blob/b80dd12da9c0bc3f569eca3455e77762cf2ee8ef/keras/callbacks.py#L1916)
instead. Where a more experienced contributor could add as short comment on the special behaviour of the default value is 0.

gbaned:
@gowthamkpr Can you please resolve conflicts? Thank you!

PR #16499, opened 2022-05-05
Fix missed parameter from AUC config

weipeilun:
AUC metrics missed `from_logits` parameter in `get_config` method, which will cause a error when load from a saved model and `from_logits` is True.

### DIFF
--- a/keras/metrics/confusion_matrix_test.py
+++ b/keras/metrics/confusion_matrix_test.py
@@ -1296,15 +1296,15 @@ class RecallAtPrecisionTest(tf.test.TestCase, parameterized.TestCase):
 
 @test_combinations.generate(test_combinations.combine(mode=["graph", "eager"]))
 class AUCTest(tf.test.TestCase, parameterized.TestCase):
-  def setup(self):
-    self.num_thresholds = 3
-    self.y_pred = tf.constant([0, 0.5, 0.3, 0.9], dtype=tf.float32)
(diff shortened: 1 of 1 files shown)


qlzh727 (changes requested):
Thanks for the fix. Could u add a unit test to cover this case.

fchollet (commented):
Thanks for the PR! Could you add a unit test (saving and loading the metric and checking that the behavior is the same)?

weipeilun:
@qlzh727 @fchollet Thanks for the comment. A unit test is added and another parameter missed is found, thanks again.

keras/metrics/metrics_test.py
```diff
@@ -1755,6 +1755,37 @@ def test_axis(self):
     self.assertAllClose(self.evaluate(result), 1.176, atol=1e-3)
 
 
+@test_combinations.generate(test_combinations.combine(mode=['graph', 'eager']))
+class AUCTest(tf.test.TestCase):
```
qlzh727, line 1759: You can move this to the confusion_matrix_test.py, which has a AUCTest class.

qlzh727 (changes requested):
Thanks for the change.

weipeilun:
@qlzh727 I moved the test case to confusion_matrix_test.py, again found a error caused by dtype mismatch of weights and labels when calling update_state(). Please review that too, thanks.

qlzh727 (approved):
Thanks for the PR and sorry for the wait.

PR #16505, opened 2022-05-06
New parameter start_save for ModelCheckPoint callback

Fab16BSB:
Hi everyone a propose to add a new parameters to ModelCheckPoint callback.
This new parameter is called start_save and permited to start the backup process at a certain epoch.
In my project a train lot of model all of them are converging between 100 and 300 epochs.
With this parameters i can specified that i would like a backup on each epoch starting on 100 that permit to stored minus backup file and generaly first epochs have verry bad performance so they are not interesting to save.
To default this parameter are initiate to 0 to start backup at the first epoch.
In my case is working correctly but a don't adapt it this solution for batch_save because a don't use this possibilty.

### DIFF
--- a/keras/callbacks.py
+++ b/keras/callbacks.py
@@ -1253,6 +1253,8 @@ class ModelCheckpoint(Callback):
         epochs, the monitored metric may potentially be less reliable (it
         could reflect as little as 1 batch, since the metrics get reset every
         epoch). Defaults to `'epoch'`.
+      start_save: integer that represent the epoch on which we want to start 
+        the backups
       options: Optional `tf.train.CheckpointOptions` object if
         `save_weights_only` is true or optional `tf.saved_model.SaveOptions`
         object if `save_weights_only` is false.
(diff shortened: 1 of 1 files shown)


rchao:
Hello @Fab16BSB, thanks for the PR! This doesn't appear to be a common use case, so I would recommend that you subclass the ModelCheckpoint callback for your use to save only after certain epochs, instead of merging this into the standard ModelCheckpoint callback.

Fab16BSB:
hello @rchao thank you for your feedback.
I have another question, I'm currently working on a prototype of real-time learning curve and I wanted to know if this type of push will be more appropriate ?

rchao:
Hello @Fab16BSB, thanks for the question. One approach I would take is to implement a subclass of existing callbacks or the util in [tensorflow_addons](https://github.com/tensorflow/addons) for your needs, and we can assess whether that's a use case commonly used enough to be merged into Keras main repo.

gbaned:
@Fab16BSB Can you please resolve conflicts? Thank you!

Fab16BSB:
@gbaned i am new on github so i don't know how i need to do that so i make a new pull request available on https://github.com/keras-team/keras/pull/16677

PR #16525, opened 2022-05-11
Add unittest for supporting VariablePolicy in saved_model.load

wenscarl:
This PR adds unittest for adding VariablePolicy to saved_model.load when soft placement is disabled. A related PR is https://github.com/tensorflow/tensorflow/pull/54319

### DIFF
--- a/keras/tests/saved_model_test.py
+++ b/keras/tests/saved_model_test.py
@@ -1,4 +1,4 @@
-# Copyright 2018 The TensorFlow Authors. All Rights Reserved.
+# Copyright 2022 The TensorFlow Authors. All Rights Reserved.
 #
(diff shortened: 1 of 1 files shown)


qlzh727:
Assign to Kathy since she is also reviewing the other PR on tf side.

k-w-w:
since this is testing something in TensorFlow, could I ask that you move this test to the TensorFlow codebase (and change the Keras-related part of the code)? The low-level equivalent class of a Keras model is `tf.Module`.

k-w-w:
I would put it here: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/saved_model/load_test.py

wenscarl:
Add the unittest there. Thanks and closing.

PR #16528, opened 2022-05-12
fix: weight keys so that imagenet init works

sayakpaul:
This PR fixes the keys of the `WEIGHTS_HASHES`. Otherwise, if someone does `tiny = keras.applications.ConvNeXtTiny()` they would run into:

```
    488     if include_top:
    489       file_suffix = ".h5"
--> 490       file_hash = WEIGHTS_HASHES[model_name][0]
    491     else:
    492       file_suffix = "_notop.h5"

KeyError: 'convnext_tiny'
```

@LukeWood @fchollet

### DIFF
--- a/keras/applications/convnext.py
+++ b/keras/applications/convnext.py
@@ -38,19 +38,19 @@ from tensorflow.python.util.tf_export import keras_export
 BASE_WEIGHTS_PATH = "https://storage.googleapis.com/tensorflow/keras-applications/convnext/"
 
 WEIGHTS_HASHES = {
-  "tiny":
+  "convnext_tiny":
     ("8ae6e78ce2933352b1ef4008e6dd2f17bc40771563877d156bc6426c7cf503ff",
       "d547c096cabd03329d7be5562c5e14798aa39ed24b474157cef5e85ab9e49ef1"),
-  "small":
+  "convnext_small":
     ("ce1277d8f1ee5a0ef0e171469089c18f5233860ceaf9b168049cb9263fd7483c",
       "6fc8009faa2f00c1c1dfce59feea9b0745eb260a7dd11bee65c8e20843da6eab"),
-  "base":
+  "convnext_base":
     ("52cbb006d3dadd03f6e095a8ca1aca47aecdd75acb4bc74bce1f5c695d0086e6",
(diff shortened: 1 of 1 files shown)


qlzh727:
Triage notes: we should have proper test coverage for this.

sayakpaul:
> Triage notes: we should have proper test coverage for this.

Yes! I would have expected this to be caught under the applications weight load test.

LukeWood:
> > Triage notes: we should have proper test coverage for this.
> 
> Yes! I would have expected this to be caught under the applications weight load test.

I'm really surprised this wasn't caught as well.  @sayakpaul any chance we can update applications test to fail when the change isn't made?

sayakpaul:
@LukeWood my observations:

* In `keras/applications/applications_test.py`, only MobileNetV3 seems to have a test for loading weights: 

https://github.com/sayakpaul/keras/blob/2f54af1c18c0b46f3ab0d29cd96427352e023b09/keras/applications/applications_test.py#L183

* In `keras/applications/applications_load_weight_test.py`, in case there's an exception when loading the application with `imagenet` weights the test seems to be skipped:

https://github.com/sayakpaul/keras/blob/patch-1/keras/applications/applications_load_weight_test.py#L135-#L139

  Maybe that is why the current issue wasn't caught?

LukeWood (approved):

haifeng-jin:
Thanks for the PR!
@sayakpaul @LukeWood Should we add test to this PR for loading the weights?
Before this PR, were pretrained convnext not usable? or it works in some cases? Just making sure the change would not break any existing case.

sayakpaul:
> Before this PR, were pretrained convnext not usable? or it works in some cases? Just making sure the change would not break any existing case.

@haifeng-jin I had tested the weights here: 
https://github.com/sayakpaul/keras-convnext-conversion/blob/master/in1k-eval/eval.ipynb

I think we should correct the tests, yes. I have mentioned some points in https://github.com/keras-team/keras/pull/16528#issuecomment-1125557290. 

Let us know your thoughts.

LukeWood:
My thoughts on this topic:

- the test being disabled is separate from this PR, it will actually be easier to fix via a separate commit by a team member than having @sayakpaul do it here.
- we should get this fix in, then try to fix the tests and re-enable them to prevent us from letting issues like this slip through the cracks

Thoughts?

sayakpaul:
Agreed 🙌

PR #16537, opened 2022-05-14
Optimize L2 Regularizer (use tf.nn.l2_loss)

szutenberg:
This change is performance-related because we get a much simpler graph while using L2 Regularizer. I see a difference on a custom accelerator (it's not spectacular because regularization is not compute-expensive but IMHO we should optimize the graph whenever possible).

Also, my change may decrease memory peak usage on some accelerators (no need to use "Tile").

image

I'm sending another PR with added comments in the code (it addresses comment from @fchollet   _IMO 2 * tf.nn.l2_loss(x) hinders readability compared to tf.reduce_sum(tf.square(x)). You don't understand where the 2 comes from and you have to know about an obscure TF API. Basically you cannot understand the code without reading the TF docs._). Previous PR ( https://github.com/keras-team/keras/pull/16520 ) was closed and I have no option to reopen it.


gbaned:
@szutenberg Can you please resolve conflicts? Thanks!

szutenberg:
@gbaned Done

chenmoneygithub (approved):

PR #16603, opened 2022-05-26
fix(v1):  avoid calling training_v1.Model.metrics during PREDICT

s22chan:
training_v1.Model.metrics adds about 10% overhead to my v1 model predictions.

During inference (with no callbacks), `make_logs` and `set_callback_parameters` is repeatedly called, which eventually calls a very expensive `_flatten_layers`.

This change only generates the metrics names when required.

An alternative is to make the metrics a cached property but that can be tricky with memory leak management. Let me know what you prefer.

pstats

### DIFF
--- a/keras/callbacks.py
+++ b/keras/callbacks.py
@@ -611,6 +611,14 @@ class CallbackList:
                 )
         # pylint: enable=protected-access
 
+    def make_logs(self, model, logs, outputs, mode, prefix=''):
+        """Computes logs for sending to `on_batch_end` methods."""
+        if not self.callbacks:
+            return logs
+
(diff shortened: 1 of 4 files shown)


rchao:
Thanks for the PR. This appears to be a safe change. I'd keep this open for a few days to see if there's any concern from the team or the community. Feel free to follow up in a few days and I can approve this.

s22chan:
Haha looks like it has more impact than I expected... I'll look into the failed tests later when I'm free

gbaned:
@s22chan  Any update on this PR? Please. Thank you!

haifeng-jin:
@s22chan Would you please take a look at the failed tests?
Also remember to use `sh shell/format.sh` to format your code.

Thanks!

PR #16726, opened 2022-06-28
Fix batchnorm momentum in ResNetRS

shkarupa-alex:
As mentioned in discussion https://github.com/keras-team/keras/pull/16001#issuecomment-1167067659 BatchNormalization momentum should not equals to 0. There is no difference in inference or training, but harms training from scratch and in finetuning mode.


fchollet (approved):
LGTM, thanks!

PR #16772, opened 2022-07-08
Update standard name

chunduriv:
Update alias to standard name.
From `tf.keras.preprocessing.timeseries_dataset_from_array` to `tf.keras.utils.timeseries_dataset_from_array`


qlzh727 (approved):
Thanks for the update.

PR #16818, opened 2022-07-20
Fix NASNet input shape computation

### LINKED ISSUE #16791: NASNet error using custom input shape without top layer
https://github.com/keras-team/keras/blob/07e13740fd181fc3ddec7d9a594d8a08666645f6/keras/applications/nasnet.py#L174-L180

Shouldn't the parameter `require_flatten` uses `include_top` value instead of constant `True`? 
Using `True` makes custom shaped input invalid when using size that isn't the `default_size` with `include_top = False`
Checked several other models such as efficientnet_v2 and xception, they all use `include_top` as value instead of `True`

ianjjohnson:
Fixes #16791.

Thank you to [ryukinlika](https://github.com/ryukinlika) who found the issue and identified an appropriate fix.

### DIFF
--- a/keras/applications/nasnet.py
+++ b/keras/applications/nasnet.py
@@ -188,7 +188,7 @@ def NASNet(
         default_size=default_size,
         min_size=32,
         data_format=backend.image_data_format(),
-        require_flatten=True,
+        require_flatten=include_top,
         weights=weights,
     )
 
--- a/keras/applications/applications_test.py
+++ b/keras/applications/applications_test.py
@@ -180,7 +180,7 @@ class ApplicationsTest(tf.test.TestCase, parameterized.TestCase):
     def test_application_notop_custom_input_shape(self, app, last_dim):
         output_shape = _get_output_shape(
             lambda: app(
-                weights=None, include_top=False, input_shape=(224, 224, 3)
+                weights='imagenet', include_top=False, input_shape=(224, 224, 3)
             )
         )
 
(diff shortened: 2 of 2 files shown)


ianjjohnson:
@[qlzh727](https://github.com/qlzh727)

qlzh727:
Can u update the unit test Ian?

keras/applications/nasnet.py
```diff
@@ -188,7 +188,7 @@ def NASNet(
         default_size=default_size,
         min_size=32,
         data_format=backend.image_data_format(),
-        require_flatten=True,
+        require_flatten=include_top,
```
qlzh727, line 191: Please add/update unit test for this.

qlzh727 (changes requested):

keras/applications/nasnet.py
ianjjohnson: Added a unit test for this case and verified that test passes locally (and fails without the change).

qlzh727:
Ian, it seems that the presubmit test is not auto triggering since you didn't accept the invite we sent to join the keras-team github org. Could you check that? (which should allow u more freely test the PR.)

ianjjohnson:
> Ian, it seems that the presubmit test is not auto triggering since you didn't accept the invite we sent to join the keras-team github org. Could you check that? (which should allow u more freely test the PR.)

I've accepted the invite -- it looks like this didn't retroactively trigger the presubmits, but hopefully it will next time?

qlzh727 (approved):
Thanks for the fix.

PR #16830, opened 2022-07-25
Added function get_reversed_word_index in imdb dataset

FrancescoBoi:
The rationale for this PR is to provide an easy way to obtain the reversed word index. 

Without it the user should first get the word index with the function `get_word_index` and then iterate inverting the keys and the values. However the conversion is not straightforward due to an existing offset of the index given by special words such as `oov` (out of vocabulary) and `start` (which marks the beginning of a sentence).

With this modification the user does not have to know about the offset of 3, which is very tricky considering alsoe that even the [official tensorflow documentation](https://www.tensorflow.org/api_docs/python/tf/keras/datasets/imdb/get_word_index) does not take into consideration this aspect.

### DIFF
--- a/keras/datasets/imdb.py
+++ b/keras/datasets/imdb.py
@@ -191,14 +191,11 @@ def get_word_index(path="imdb_word_index.json"):
     )
     # Retrieve the word index file mapping words to indices
     word_index = keras.datasets.imdb.get_word_index()
-    # Reverse the word index to obtain a dict mapping indices to words
-    # And add `index_from` to indices to sync with `x_train`
-    inverted_word_index = dict(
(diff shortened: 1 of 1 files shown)


FrancescoBoi:
@gbaned If I would like write some unit tests for the datasets/imdb, where shall I write it? Is there already a test for the function `get_index`?If so, where?

fchollet (commented):
Thanks for the PR!

At this time we are not adding any new functionality to the `keras.dataset` module, which exists purely for debugging and education purposes and should largely be considered legacy. While the feature you propose may be useful to some, it would be against our policy to add it.

Issue #16815, opened 2022-07-20
kernel dimensions in Depthwiseconv1d()


dario-passos:
Hi, 
I noticed that in the Depthwiseconv1d() documentation the definition of kernel_size says "_An integer, specifying the height and width of the 1D convolution window". Since this is a 1d conv windows, isn't the height of the kernel always 1?

sushreebarsa:
@dario-passos Could you please provide more information on the issue and share the documentation site that you are referring? Thank you!

dario-passos:
Hi

I'm refereing to the comments (line 52) on "Args:" of the the file "[https://github.com/keras-team/keras/blob/master/keras/layers/convolutional/depthwise_conv1d.py](https://github.com/keras-team/keras/blob/master/keras/layers/convolutional/depthwise_conv1d.py)" which has been uploaded to this repo 2 months ago but still does not appears as an available layer in keras.io. The same text is also displayed in the latest tensorflow.keras layers documentation. My question is related with the height of the kernel used in this layer... Is the definition of the variable "kernel_size" correct or is there a typo due to a possible text copy from another function "depthwise_conv2d.py"?
Cheers

PR #16854, opened 2022-08-01
Fix typo of kernel_size definition

gadagashwini:
Fixes [#16815](https://github.com/keras-team/keras/issues/16815).


gbaned:
Hi @gadagashwini,  It is better to submit multiple typo fixes in a single PR as the CPU/GPU hours are wasted on CI. Hence, we do not encourage one liner grammatical changes as it is an expensive process. Thank you for your contribution!
@fchollet, @qlzh727

PR #16875, opened 2022-08-08
Respecting KERAS_HOME for get_file()

adrianjhpc:
This is a small feature update to keras/utils/data_utils.py to respect the KERAS_HOME environment variable if set by a user. In this way it's possible to allow a user to download datasets required by Keras applications and put them into a directory to be picked up by application runs on systems where /home (or at least the default ~/.keras directory) is not available when running applications.

The majority of the changes in this pull request are actually from running `sh shell/format.sh` on the code base as outlined in  the contributing guidelines.


adrianjhpc:
Removing to revise with one with commits that adhere to the CLA agreement

PR #16877, opened 2022-08-08
Updating get_file() to respect KERAS_HOME environment variable

adrianjhpc:
This is a small feature update to enable the utils/data_utils.py get_file() function to respect the KERAS_HOME environment variable if set by a user. This would allow any datasets required by applications to be download to be cached in directories other than ~/.keras, which is especially useful for people running on systems where the HOME directory is not accessible from the place applications are being run.


fchollet (approved):
Thanks for the PR! We use the KERAS_HOME variable in the backend already, this is a reasonable change.

PR #16893, opened 2022-08-13
Fix docs of `metrics` parameter in `compile`

matangover:
Fix https://github.com/tensorflow/tensorflow/issues/41361. The existing documentation refers to an old behavior which was replaced when MetricsContainer was added (see commit https://github.com/tensorflow/tensorflow/commit/fb30b12e101e55bf008588a9cd72d4150ba67ad8). The [current code](https://github.com/keras-team/keras/blob/master/keras/engine/compile_utils.py#L647) does not use the loss function for deciding which accuracy metric to use, it uses only the shapes of the targets and of the model output.


fchollet (approved):
LGTM, thanks!

Issue #16902, opened 2022-08-16
tf.keras.layers.Dense / tf.keras.layers.GRU / tf.keras.layers.SimpleRNN documentation wrong


cheyennee:
The parameter 'units' is a positive integer on documentation. However, when 'unit' is 0, code works. And when 'unit' is -1, the error message is "Error:Dimension -1 must be >= 0". So 'units' is not a negative integer whose value can be set to positive integer and zero.

```
import tensorflow as tf
results={}
try:
  arg_0 = 0
  activation = "softmax"
  arg_class = tf.keras.layers.Dense(arg_0,activation=activation,)
  arg_input = tf.random.uniform([3, 50, 1], dtype=tf.float32)
  results["res"] = arg_class(arg_input)
except Exception as e:
  results["err"] = "Error:"+str(e)
print(results)
```
'''
results = {'res': <tf.Tensor: shape=(3, 50, 0), dtype=float32, numpy=array([], shape=(3, 50, 0), dtype=float32)>}
'''

version:
TF2.4
os: win11

jbischof:
Are the docs wrong or is our input validation wrong @cheyennee? It doesn't seem like a `Dense` layer with zero units has no utility and thus is unlikely to be the user intention. My inclination would be to add a `ValueError` for `units=0`.

cheyennee:
Zero units may not really be useful, but I think the user should be allowed to input 0 units, just as the user can input 0-batch dim tensors. @jbischof  Maybe users really have this need? 😂

tilakrayal:
@cheyennee,
Could you please feel free to submit a PR for the requested change to be made. Thank you!

jbischof:
@cheyennee we would prefer a PR with input validation for `units=0` as we cannot determine a use case for this configuration. Please let us know if you're willing to contribute. (cc @fchollet)

cheyennee:
@tilakrayal @jbischof I've been busy doing testing work lately, so I probably don't have free time to fix bugs😭I'm so sorry🥺

gadagashwini:
Hi @cheyennee, Raised PR [#16921](https://github.com/keras-team/keras/pull/16921). Thank you!

PR #16921, opened 2022-08-22
Fix value error for Units of tf.keras.layers.Dense

gadagashwini:
The parameter `units=0` of `tf.keras.layers.Dense` should raise ValueError. Since `Units` expects Positive integer. 
Fixes [#16902](https://github.com/keras-team/keras/issues/16902). Thank you!


jbischof:
Should we fix GRU and RNN as well as mentioned in the issue? Are there other classes with similar issues?

fchollet (approved):
LGTM, thanks!

gadagashwini:
Hi @jbischof, I will raise fix for GRU and RNN.

PR #16926, opened 2022-08-23
Fix Value error for Units of tf.keras.layers.SimpleRNN

gadagashwini:
Fixes [#16902](https://github.com/keras-team/keras/issues/16902). Thank you!


fchollet (approved):
LGTM, thanks!

PR #16929, opened 2022-08-23
Fix Value Error for Units of tf.keras.layers.LSTM

gadagashwini:
Similar [#16921](https://github.com/keras-team/keras/pull/16921).


fchollet (approved):
LGTM, thanks!

PR #16942, opened 2022-08-25
Added an optional argument

jordan663994:
I added an optional argument for the Dense layer that sets a factor, the Dense layer's output is multiplied by this Factor. It is to avoid needing to do this with multiple Lambda layers in larger models.


jordan663994:
oops. forgot to verify my account with google

PR #16945, opened 2022-08-25
Proposal to add an optional constant that scales the output of the Dense layer

jordan663994:
added an optional property to the Dense layer to my draft, it multiplies the layer's output by a factor, this is useful for larger networks with more layers. This Lambda layers are a solution, but blocks of multiplication functions can get hard to keep track of fast. This makes it easier for the user to control the scale of the layer's output.


rthadur:
Please submit your changes to master. Thank you

PR #16955, opened 2022-08-29
Update typo in `compute_output_shape`

chunduriv:
Since the method name is `compute_output_shape` returns should be `an output shape tuple` not `an input shape tuple`.

### DIFF
--- a/keras/engine/base_layer.py
+++ b/keras/engine/base_layer.py
@@ -841,7 +841,7 @@ class Layer(tf.Module, version_utils.LayerVersionSelector):
                 instead of an integer.
 
         Returns:
-            An input shape tuple.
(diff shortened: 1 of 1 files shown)


fchollet (commented):
LGTM, thanks.

Actually, can you specify that it should be a "shape tuple or `tf.TensorShape` instance"?

chunduriv:
> LGTM, thanks.
> 
> Actually, can you specify that it should be a "shape tuple or `tf.TensorShape` instance"?

@fchollet, Thank you for the review. Updated Returns section as per your suggestion.

fchollet (approved):
LGTM. Note that I pushed some changes.

Issue #16917, opened 2022-08-19
The type of parameter 'prefix' in tf.keras.backend.get_uid documentation wrong


cheyennee:
Parameter 'prefix' must be a string on documentation. However, I find that when type of prefix is bool/tuple/integer, the code works. But when type of prefix is list, code throws an exception: "Error:unhashable type: 'list'". I think supported type like integer should be added to documentation.
```
import tensorflow as tf
results={}
try:
  arg_0 = 5
  results["res"] = tf.keras.backend.get_uid(arg_0,)
except Exception as e:
  results["err"] = "Error:"+str(e)
print(results)
# results={'res': 1}
```

tf version: 2.4
os: win11

PR #16974, opened 2022-09-01
Fix documentation of prefix support types

gadagashwini:
Fixes [#16917](https://github.com/keras-team/keras/issues/16917). 

`tf.keras.backend.get_uid(prefix='')`, prefix supports bool/tuple/integer along with string. Thank you!


mattdangerw:
This utility is really meant for strings. It does a cast to string (so would probably support a wide set of types than what you are listing), but that's not something we plan on documenting.

PR #16975, opened 2022-09-01
Suggested fixes by iCR, OpenRefactory, Inc.

openrefactory:
This issue was detected in branch `master` of `keras` project on the version with commit hash `062073`. This is an instance of an inappropriate logic.

**Fixes for inappropriate logic:**
In file: `training_distributed_v1.py`, the comparison of Collection length creates a logical short circuit. **iCR** suggested that the Collection length comparison should be done without creating a logical short circuit.

This issue was detected by **OpenRefactory's Intelligent Code Repair (iCR)**. We are running **iCR** on libraries in the `PyPI` repository to identify issues and fix them. More info at: [pypi.openrefactory.com](https://pypi.openrefactory.com/)

### DIFF
--- a/keras/engine/training_distributed_v1.py
+++ b/keras/engine/training_distributed_v1.py
@@ -460,8 +460,11 @@ def experimental_tpu_test_loop(
     callbacks._call_end_hook(mode)
 
     scope.__exit__(None, None, None)
-    if len(outs) >= 0:
-        outs[0] /= target_steps
+    # OpenRefactory Warning: Collection length comparison should be meaningful.
+    # The length of a collection is always greater than or equal to zero.
+    # So testing that a length is greater than or equal to zero is always true.
+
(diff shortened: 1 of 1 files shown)


gbaned:
Hi @openrefactory Can you please sign CLA. Thank you!

fchollet (commented):
The correct fix would be to do `if len(outs) > 0:`

Anyway this is in `keras/engine/training_distributed_v1.py`, which is legacy and no longer maintained. Does not seem worth fixing.

ramil4icr:
The fix was auto-generated by a tool (iCR). It identified that the test expression is redundant because it is always true and suggested a fix to remove the test altogether. In future, we may improve Intelligent Code Repair (iCR) to look into the context and distinguish the case for non-zero length check. Thanks for your input.

PR #17038, opened 2022-09-16
Fixed Broken link of paper jozefowicz15  et al

mohantym:
Fixed Broken link of paper jozefowicz15  et al at line 92


fchollet (approved):
LGTM, thanks

PR #17050, opened 2022-09-19
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -3,8 +3,13 @@ name: Format the code
 on:
   workflow_dispatch:
 
+permissions: {}
 jobs:
   createPullRequest:
+    permissions:
+      contents: write # to create branch (peter-evans/create-pull-request)
+      pull-requests: write # to create a PR (peter-evans/create-pull-request)
+
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,6 +4,9 @@ on:
   push:
   pull_request:
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   lint:
     name: Check the code format


.github/workflows/format.yml
```diff
@@ -3,8 +3,13 @@ name: Format the code
 on:
   workflow_dispatch:
 
+permissions: {}
 jobs:
   createPullRequest:
+    permissions:
+      contents: write # to create branch (peter-evans/create-pull-request)
```
fchollet, line 10: What's this peter-evans comment?

fchollet (commented):
Thanks for the PR. Can you clarify what risk/threat this is supposed to address?

sashashura:
The threat is a potential compromise of a build tool or action. It is about running CI/CD with the least privilege required to do the task. For example, the lint workflow doesn't need any write permission, but currently runs with the following permissions:
image

.github/workflows/format.yml
sashashura: It is an identifier of the action [that is used in the workflow](https://github.com/keras-team/keras/blob/2eda0600e24c48363a8eb07ad2c3f39436430df7/.github/workflows/format.yml#L32).

fchollet (approved):
Ok, sounds reasonable. Thanks

Issue #16173, opened 2022-03-04
TensorBoard callback does not collect data per batch (accidentally removed)


foxik:
**System information**.
- Have I written custom code (as opposed to using a stock example script provided in Keras): **No**
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): **Linux Debian Stable**
- TensorFlow installed from (source or binary): **binary**
- TensorFlow version (use command below): **2.8.0**
- Python version: **3.9**

**Describe the problem**.

According to `https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/TensorBoard`, the value `update_freq=N` should collect logs every `N` batches. However, no logs are generated after batches, only after an epoch.

**Describe the current behavior**.

No logs are generated after `N` training batches.

**Describe the expected behavior**.

Logs should be generated after `N` training batches.

**Details**

The feature for collecting `batch_*` summaries was removed in https://github.com/keras-team/keras/commit/7d06227f3ef653a451fffe65d1d7dace7f7945df -- see how `write_scalar_summaries` was removed from `keras/engine/training`.

It should be enough to just revert the given commit.

foxik:
Adding the author of the commit removing documented functionality: @jonycgn .

foxik:
Alternatively, some variant of `write_scalar_summaries` could be moved to `on_train_batch_end` method of `TensorBoard` callback.

jonycgn:
The mentioned commit didn't remove functionality from the `TensorBoard` callback, but from `Model`. The problem with these summaries was that they would be written unconditionally after each training step, regardless of whether the `TensorBoard` callback was even used. So reverting wouldn't make the code behave correctly.

The functionality should be corrected in the callback. It is unclear to me why it doesn't exist. It seems the `update_freq` parameter is simply ignored unless it is set to `'epoch'`.

Adding @rchao who might be more knowledgeable on this.

foxik:
Hi,

thanks for answer! The commit actually did remove functionality from `TensorBoard` callback -- the callback just sets correct `tf.summary.record_if` in https://github.com/keras-team/keras/blob/46121eed08d0feef743eacef3b66206df45cf656/keras/callbacks.py#L2362-L2373
but it relied on the `write_scalar_summaries` executed in the `keras/engine/training.py`, the one you removed. So once this "generic summary emitter" was removed, it broke the batch-level logs in `TensorBoard` callback.

[...]

PR #17052, opened 2022-09-20
Restore TensorBoard update_freq behavior

myaaaaaaaaa:
Adaptation of the temporary fix by @kyamagu from #16173.

### DIFF
--- a/keras/callbacks.py
+++ b/keras/callbacks.py
@@ -2767,6 +2767,18 @@ class TensorBoard(Callback, version_utils.TensorBoardVersionSelector):
                 1.0 / batch_run_time,
                 step=self._train_step,
             )
+
+        should_record = False
+        if type(self.update_freq) == int:
+            should_record = tf.equal(self._train_step % self.update_freq, 0)
+        with tf.summary.record_if(should_record):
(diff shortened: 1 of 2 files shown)


keras/callbacks.py
```diff
@@ -2767,6 +2767,18 @@ def on_train_batch_end(self, batch, logs=None):
...
                 step=self._train_step,
             )
+
+        should_record = False
+        if type(self.update_freq) == int:
+            should_record = tf.equal(self._train_step % self.update_freq, 0)
+        with tf.summary.record_if(should_record):
+            if logs:
+                with self._train_writer.as_default():
+                    for name, value in logs.items():
```
rchao, line 2777: Thanks for the PR! Unfortunately this would have a performance hit with asynchronous strategies such as `TPUStrategy` and `ParameterServerStrategy`, and users would need to add the summary logging ops within custom `train_step()` or `call()`. @nkovela1 is actively working on an example workflow users could use in the documentation.

On the other hand, feel free to do this in a custom callback inheriting `TensorBoard`, if your workflow doesn't involve asynchronous tf.distribute strategies.

Issue #17047, opened 2022-09-19
keras Normalization layer doesn't denormalize properly


Vincent-SV:
I wanted to use keras Normalization layer to "denormalize" my output. 
The doc for this object says the argument "invert=True" does exactly that, but it doesn't behave as I thought at all...

After searching a bit, I found that in "keras/keras/layers/preprocessing/normalization.py", the "call" method adds the mean and then multiply by standard deviation instead of multiplying by standard deviation first.

```
    def call(self, inputs):
        inputs = self._standardize_inputs(inputs)
        # The base layer automatically casts floating-point inputs, but we
        # explicitly cast here to also allow integer inputs to be passed
        inputs = tf.cast(inputs, self.compute_dtype)
        if self.invert:
            return (inputs + self.mean) * tf.maximum(
                tf.sqrt(self.variance), backend.epsilon()
            )
        else:
            return (inputs - self.mean) / tf.maximum(
                tf.sqrt(self.variance), backend.epsilon()
            )
```

As a result, 0 will not be denormalized to self.mean but to self.mean * sqrt(self.variance)

Vincent-SV:
I opened a PR : Update normalization.py to fix a bug when "denormalizing" #17054
Maybe I should have waited here, because I am new to github and I have the impression I am skipping some steps

Vincent-SV:
The PR has been merged 3 days ago so it should be fixed, but I don't find it even when I pip install tf-nightly. I don't find the bugfix in the python files, I'm not sure how this works.

Vincent-SV:
Update : the bugfix is part of tf-nightly now, so the issue is solved

PR #17054, opened 2022-09-20
Update normalization.py to fix a bug when "denormalizing"

Vincent-SV:
I created an issue yesterday to talk about this : keras Normalization layer doesn't denormalize properly, https://github.com/keras-team/keras/issues/17047

The denormalization (the .call method of the Normalization layer with invert = True) doesn't work properly for now (if you normalize and denormalize, you don't get the inputs back) and the problem is pretty obvious : you have to multiply by standard deviation first and then add the mean in order to invert the normalization process.
This change looks like an obvious improvement to me, BUT
- I am new to github and not totally sure I am doing things right
- I didn't test this change in any way (because it looks too complicated for me to test it)
- I don't really understand the role of backend.epsilon, it is possible something should be changed (I think it is needed only when dividing ?) 
- I searched a bit and this is the only place I found where this problem, but I don't fully understand how this change will fully fix this problem, maybe another similar change is needed elsewhere ?
- There is no test for "invert = True" in normalization_test.py, maybe it is an issue ?


SalmanshariffSR:
How to update

Vincent-SV:
Oops, I stated that there is no test for invert=True in normalization_test.py but that is false. There are 2 tests but they are only working because the variance is 1 (for both of them) so the wrong denormalization works in this case. I'll propose a slight modification of these tests.

fchollet (approved):
LGTM, thanks for the catch

Vincent-SV:
I fixed a small oversight in a test I modified, and this time I actually tested my modifications (using docker), sorry for not doing it before.

fchollet (approved):
LGTM

PR #17142, opened 2022-10-12
Restore update_freq functionality

myaaaaaaaaa:
Revision of #17052 , adapted from code by @foxik 

@rchao It turns out that all the logging checks in the previous version were unnecessary, since `_push_writer()` already sets a writer that logs at the rate set by update_freq. Can you take a look to see if this version also has performance problems with asynchronous strategies?

### DIFF
--- a/keras/callbacks.py
+++ b/keras/callbacks.py
@@ -2733,6 +2733,13 @@ class TensorBoard(Callback, version_utils.TensorBoardVersionSelector):
                 1.0 / batch_run_time,
                 step=self._train_step,
             )
+
+        # `logs` is a `RemoteValue` when using asynchronous strategies, for now
+        # we just disable `update_freq` entirely in those cases.
+        if isinstance(logs, dict):
+            for name, value in logs.items():
+                tf.summary.scalar("batch_" + name, value, step=self._train_step)
+
         if not self._should_trace:
             return
 
--- a/keras/callbacks.py
+++ b/keras/callbacks.py
@@ -2354,7 +2354,8 @@ class TensorBoard(Callback, version_utils.TensorBoardVersionSelector):
           same applies for `'epoch'`. If using an integer, let's say `1000`, the
           callback will write the metrics and losses to TensorBoard every 1000
           batches. Note that writing too frequently to TensorBoard can slow down
-          your training.
+          your training. May not work when doing distributed training, as
+          currently only a subset of `tf.distribute.Strategy`s are supported.
         profile_batch: Profile the batch(es) to sample compute characteristics.
           profile_batch must be a non-negative integer or a tuple of integers.
           A pair of positive integers signify a range of batches to profile.
@@ -2775,8 +2776,9 @@ class TensorBoard(Callback, version_utils.TensorBoardVersionSelector):
                 step=self._train_step,
             )
 
-        # `logs` is a `RemoteValue` when using asynchronous strategies, for now
-        # we just disable `update_freq` entirely in those cases.
+        # `logs` is a `tf.distribute.experimental.coordinator.RemoteValue` when
+        # using asynchronous strategies, for now we just disable `update_freq`
+        # entirely in those cases.
         if isinstance(logs, dict):
             for name, value in logs.items():
                 tf.summary.scalar("batch_" + name, value, step=self._train_step)
--- a/keras/callbacks.py
+++ b/keras/callbacks.py
@@ -2354,8 +2354,9 @@ class TensorBoard(Callback, version_utils.TensorBoardVersionSelector):
           same applies for `'epoch'`. If using an integer, let's say `1000`, the
           callback will write the metrics and losses to TensorBoard every 1000
           batches. Note that writing too frequently to TensorBoard can slow down
-          your training. May not work when doing distributed training, as
-          currently only a subset of `tf.distribute.Strategy`s are supported.
+          your training, especially when used with `tf.distribute.Strategy` as
+          it will incur additional synchronization overhead.
+          Use with `ParameterServerStrategy` is not supported.
         profile_batch: Profile the batch(es) to sample compute characteristics.
           profile_batch must be a non-negative integer or a tuple of integers.
           A pair of positive integers signify a range of batches to profile.
--- a/keras/callbacks.py
+++ b/keras/callbacks.py
@@ -2349,11 +2349,13 @@ class TensorBoard(Callback, version_utils.TensorBoardVersionSelector):
         write_steps_per_second: whether to log the training steps per second
           into Tensorboard. This supports both epoch and batch frequency
           logging.
-        update_freq: `'batch'` or `'epoch'` or integer. When using `'batch'`,
-          writes the losses and metrics to TensorBoard after each batch. The
-          same applies for `'epoch'`. If using an integer, let's say `1000`, the
-          callback will write the metrics and losses to TensorBoard every 1000
-          batches. Note that writing too frequently to TensorBoard can slow down
+        update_freq: `'batch'` or `'epoch'` or integer. When using `'epoch'`,
+          writes the losses and metrics to TensorBoard after every epoch.
+          If using an integer, let's say `1000`, all metrics and losses
+          (including custom ones added by `Model.compile`) will be logged to
+          TensorBoard every 1000 batches. `'batch'` is a synonym for `1`,
+          meaning that they will be written every batch.
+          Note however that writing too frequently to TensorBoard can slow down
           your training, especially when used with `tf.distribute.Strategy` as
           it will incur additional synchronization overhead.
           Use with `ParameterServerStrategy` is not supported.
@@ -2777,9 +2779,10 @@ class TensorBoard(Callback, version_utils.TensorBoardVersionSelector):
                 step=self._train_step,
             )
 
-        # `logs` is a `tf.distribute.experimental.coordinator.RemoteValue` when
-        # using asynchronous strategies, for now we just disable `update_freq`
-        # entirely in those cases.
+        # `logs` isn't necessarily always a dict. For example, when using
+        # `tf.distribute.experimental.ParameterServerStrategy`, a
+        # `tf.distribute.experimental.coordinator.RemoteValue` will be passed.
+        # For now, we just disable `update_freq` in those cases.
         if isinstance(logs, dict):
             for name, value in logs.items():
                 tf.summary.scalar("batch_" + name, value, step=self._train_step)
--- a/keras/callbacks.py
+++ b/keras/callbacks.py
@@ -1,4 +1,3 @@
-# flake8: noqa
 # Copyright 2015 The TensorFlow Authors. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -2350,12 +2349,12 @@ class TensorBoard(Callback, version_utils.TensorBoardVersionSelector):
         write_steps_per_second: whether to log the training steps per second
           into Tensorboard. This supports both epoch and batch frequency
           logging.
-        update_freq: **disabled**
-
-          Warning: Batch-level summary writing using `update_freq` is
-          currently unsupported. A suggested workaround is shown in the
-          [TensorBoard Scalars tutorial](https://www.tensorflow.org/tensorboard/scalars_and_keras#batch-level_logging). # pylint: disable=protected-access
-
+        update_freq: `'batch'` or `'epoch'` or integer. When using `'batch'`,
+          writes the losses and metrics to TensorBoard after each batch. The
+          same applies for `'epoch'`. If using an integer, let's say `1000`, the
+          callback will write the metrics and losses to TensorBoard every 1000
+          batches. Note that writing too frequently to TensorBoard can slow down
+          your training.
         profile_batch: Profile the batch(es) to sample compute characteristics.
           profile_batch must be a non-negative integer or a tuple of integers.
           A pair of positive integers signify a range of batches to profile.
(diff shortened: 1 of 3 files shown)


keras/callbacks.py
```diff
@@ -2716,6 +2716,11 @@ def on_train_batch_end(self, batch, logs=None):
                 1.0 / batch_run_time,
                 step=self._train_step,
             )
+
+        if logs:
+            for name, value in logs.items():
```
rchao, line 2721: Thank you for you PR! I suspect this may not work with `ParameterServerStrategy`: can you check?

myaaaaaaaaa: I couldn't figure out how to get `ParameterServerStrategy` running, sorry about that. If this is an acceptable workaround, I turned off this block by default so that any errors (if they exist) should only show up if `ParameterServerStrategy` and `update_freq` are being used at the same time.

rchao: [This](https://github.com/keras-team/keras/blob/master/keras/integration_test/distributed_training_test.py) and [this](https://github.com/keras-team/keras/blob/master/keras/integration_test/parameter_server_custom_training_loop_test.py) are two tests you can refer to to create a locally-run unit test that uses `ParameterServerStrategy`. Mind checking with that?

keras/callbacks.py
```diff
@@ -2717,7 +2717,7 @@ def on_train_batch_end(self, batch, logs=None):
                 step=self._train_step,
             )
 
-        if logs:
+        if logs and self.update_freq != "epoch":
```
myaaaaaaaaa, line 2720: > [This](https://github.com/keras-team/keras/blob/master/keras/integration_test/distributed_training_test.py) and [this](https://github.com/keras-team/keras/blob/master/keras/integration_test/parameter_server_custom_training_loop_test.py) are two tests you can refer to to create a locally-run unit test that uses `ParameterServerStrategy`. Mind checking with that?

Yes, those were the tests I tried extending before attempting to enter a `ParameterServerStrategy` scope in `callbacks.py`, but I kept getting build failures when using bazel, and getting `Cannot determine binary path` RuntimeErrors when trying to run `python distributed_training_test.py` directly.

Maybe I'll take another stab at it later if I can find the time, but in the meantime, hopefully a patch like this can be more useful than not having `update_freq` at all.

rchao: I think we would need to make sure it runs with all strategies before we can merge it to avoid breakages. In the meantime, would it work for you if you subclass the TensorBoard callback and log at batch boundaries for your use case?

myaaaaaaaaa:
I've managed to work around my bazel problems and add a basic integration test. `update_freq` should now be disabled when used with asynchronous strategies instead of throwing an error, is this acceptable?

keras/callbacks.py
```diff
@@ -2716,6 +2716,14 @@ def on_train_batch_end(self, batch, logs=None):
                 1.0 / batch_run_time,
                 step=self._train_step,
             )
+
+        # logs is a RemoteValue when using asynchronous strategies, for now we
+        # just disable update_freq entirely in those cases.
+        get_log_items = getattr(logs, "items", None)
```
rchao, line 2722: suggest: use isinstance() to be more explicit

rchao (commented):
Thanks! I think it's reasonable to only log the batch level metrics when they are readily available. Can you also update the docstring for TensorBoard to accompany this change as well?

myaaaaaaaaa:
All done!

keras/callbacks.py
```diff
@@ -2733,6 +2774,13 @@ def on_train_batch_end(self, batch, logs=None):
                 1.0 / batch_run_time,
                 step=self._train_step,
             )
+
+        # `logs` is a `RemoteValue` when using asynchronous strategies, for now
```
rchao, line 2778: nit: use fully qualified name for `RemoteValue`

keras/integration_test/distributed_training_test.py
```diff
@@ -77,7 +77,17 @@ def dataset_fn(input_context):
...
+            x,
+            epochs=2,
+            steps_per_epoch=10,
+            callbacks=[
+                tf.keras.callbacks.TensorBoard(
+                    update_freq=5,
+                    write_steps_per_second=True,
+                )
+            ],
+        )
```
rchao, line 90: Let's also add code to verify that the summaries are indeed written every 5 batches - you can refer to [1] for an example of how to verify this. Thanks!

[1] https://github.com/keras-team/keras/blob/4a670a3a5c4632f8f3ec8805fa2f49ccb1c89b13/keras/utils/sidecar_evaluator_test.py#L88-L95

keras/callbacks.py
```diff
@@ -2733,6 +2774,13 @@ def on_train_batch_end(self, batch, logs=None):
...
+        # we just disable `update_freq` entirely in those cases.
+        if isinstance(logs, dict):
```
rchao, line 2780: Sorry, there's also a performance implication for TPUStrategy as that's asynchronous and reading the metrics at batch interval forces a sync. Can you also exclude that here (by raising an error if using TPUStrategy)?

keras/callbacks.py
```diff
@@ -2350,12 +2349,12 @@ class TensorBoard(Callback, version_utils.TensorBoardVersionSelector):
         write_steps_per_second: whether to log the training steps per second
           into Tensorboard. This supports both epoch and batch frequency
           logging.
-        update_freq: **disabled**
-
-          Warning: Batch-level summary writing using `update_freq` is
-          currently unsupported. A suggested workaround is shown in the
-          [TensorBoard Scalars tutorial](https://www.tensorflow.org/tensorboard/scalars_and_keras#batch-level_logging). # pylint: disable=protected-access
-
+        update_freq: `'batch'` or `'epoch'` or integer. When using `'batch'`,
```
rchao, line 2352: Can you make it explicit that this only works for a subset of `tf.distribute.Strategy`s? And, when user is using batch or integer, there may be a performance implication and users must be cautious?

keras/callbacks.py
rchao, line 2780: Do we need to only do this when it's at the right batch number if `update_freq` is an integer?

rchao (commented):
Thanks! There are still a couple of things we need to be extra careful about. See the comments for more details.

keras/callbacks.py
myaaaaaaaaa: `_push_writer()` already makes it so that `update_freq` applies to all batch logs (including the `write_steps_per_second` block above), see here: https://github.com/keras-team/keras/blob/86e9ed51317872e1dfc75b03c1e257638bd7b083/keras/callbacks.py#L2633-L2646

myaaaaaaaaa: I'd argue that the performance degradation is already implied by the concept of `update_freq` and can be avoided by setting a larger value so that syncs don't happen as often, although I'm not 100% sure how tensorflow works here: does `tf.summary.scalar` only read the values for logging if `tf.summary.record_if` is true?

keras/integration_test/distributed_training_test.py
myaaaaaaaaa: Do you want the tests here or in `callbacks_test.py`? I meant for the test in `distributed_training_test.py` to be more of a crash detector, especially since `update_freq` currently isn't working for all strategies

rchao: `callbacks_test.py` may be more appropriate but the key is we'll need to test it on all strategies (and make sure that the unsupported strategies raise an error).

keras/callbacks.py
rchao: Yes, that's my understanding.

keras/integration_test/distributed_training_test.py
```diff
@@ -73,22 +76,41 @@ def dataset_fn(input_context):
...
+                    if value.tag != "batch_loss":
+                        continue
+                    events += [event.step]
+        events.sort()
+
+        if not isinstance(
+            strategy, tf.distribute.experimental.ParameterServerStrategy
+        ):
+            # total steps = epochs * steps_per_epoch
+            self.assertEqual(events, [5, 10, 15, 20, 25, 30, 35, 40])
```
myaaaaaaaaa, line 112: This is currently failing for `MultiWorkerMirrored2x1CPU`, what's the best way to run this test only for the chief and not the workers?

rchao: fyi, the way to tell whether it's chief is described [here](https://www.tensorflow.org/tutorials/distribute/multi_worker_with_keras#model_saving_and_loading) (look for `def _is_chief`)

myaaaaaaaaa: The assertion appears to be running on both the worker and the chief, it passes for the chief but fails for the worker as it generates no logs.

This could also be a problem with temp directories, `self.get_temp_dir()` is returning a different location for the chief and the worker. How did other tests deal with testing OS level functionality for distributed strategies?

rchao: Ah, I see, for assertion of summary file it indeed is fine to check whether it's chief and only assert if it's chief.

For a test like this that's using `tf.__internal__.test.combinations.combine`, there is not an easy way to inject a consistent directory for chief/non-chief. For a test that's using `MultiProcessRunner`, such as [this](https://github.com/keras-team/keras/blob/a8467f50815a1773be1a14ce89716c3d12c500f6/keras/integration_test/mwms_multi_process_runner_test.py#L66), it's possible to provide it as an argument to the chief and workers.

So in this test, let's just assert if it's chief and not worry much. Thanks!

keras/integration_test/distributed_training_test.py
```diff
@@ -73,11 +76,40 @@ def dataset_fn(input_context):
...
+                if not event.summary:
+                    continue
+                for value in event.summary.value:
+                    if value.tag != "batch_loss":
+                        continue
+                    events += [event.step]
+        events.sort()
+
+        if not isinstance(
+            strategy, tf.distribute.experimental.ParameterServerStrategy
```
rchao, line 109: Can we also assert the case for PSS (which I assume has an empty `events`) and add a comment for why?

keras/integration_test/distributed_training_test.py
```diff
@@ -73,11 +76,40 @@ def dataset_fn(input_context):
...
+        events = []
+        for event_file in glob.glob(logdir + "/train/events.out.*"):
+            for event in tf.compat.v1.train.summary_iterator(event_file):
+                if not event.summary:
+                    continue
+                for value in event.summary.value:
+                    if value.tag != "batch_loss":
+                        continue
+                    events += [event.step]
+        events.sort()
```
rchao, line 106: Is the sort necessary?

keras/callbacks.py
```diff
@@ -2350,12 +2349,13 @@ class TensorBoard(Callback, version_utils.TensorBoardVersionSelector):
...
+          writes the losses and metrics to TensorBoard after each batch. The
+          same applies for `'epoch'`. If using an integer, let's say `1000`, the
+          callback will write the metrics and losses to TensorBoard every 1000
+          batches. Note that writing too frequently to TensorBoard can slow down
+          your training. May not work when doing distributed training, as
+          currently only a subset of `tf.distribute.Strategy`s are supported.
```
rchao, line 2358: Can we make it clear that this is not supported in ParameterServerStrategy, and may have performance implication when using `tf.distribute`?

rchao (commented):
Thanks!

myaaaaaaaaa:
All done

keras/callbacks.py
```diff
@@ -2377,6 +2378,48 @@ class TensorBoard(Callback, version_utils.TensorBoardVersionSelector):
...
+        outputs = self.dense(x)
+        tf.summary.histogram('outputs', outputs)
+        return outputs
+
+    model = MyModel()
+    model.compile('sgd', 'mse')
+
+    # Make sure to set `update_freq=N` to log a batch-level summary every N
+    # batches.  In addition to any `tf.summary` contained in `Model.call`,
+    # metrics added in `Model.compile` will be logged every N batches.
```
rchao, line 2399: Is it N batches or 1?

rchao: (same below)

keras/callbacks.py
```diff
@@ -2733,6 +2776,14 @@ def on_train_batch_end(self, batch, logs=None):
                 1.0 / batch_run_time,
                 step=self._train_step,
             )
+
+        # `logs` is a `tf.distribute.experimental.coordinator.RemoteValue` when
+        # using asynchronous strategies, for now we just disable `update_freq`
```
rchao, line 2781: Let's make this clearer - RemoteValue is particular to ParameterServerStrategy, and not any asynchronous strategies

keras/integration_test/distributed_training_test.py
```diff
@@ -73,11 +76,52 @@ def dataset_fn(input_context):
...
+            strategy, tf.distribute.experimental.ParameterServerStrategy
+        ):
+            # Metrics are not logged with this strategy as they are not
+            # immediately available on batch end
+            events_expected = []
+        if (
+            strategy.cluster_resolver
+            and strategy.cluster_resolver.task_type == "worker"
+        ):
+            # Workaround for an issue with
```
rchao, line 120: Let's probably just say that workers won't have the event files kept.

rchao:
> All done

Thanks! Almost there. Just a couple of last quick things.

keras/callbacks.py
myaaaaaaaaa: N is correct, I'll add a note to `update_freq` documentation about `Model.compile`

myaaaaaaaaa:
All done

PR #17145, opened 2022-10-13
Adds show_trainable option to plot_model()

mfidabel:
Fixes missing part of issue #15425 where `model.summary(show_trainable=True)` was implemented but `plot_model(model, show_trainable=True)` wasn't.

Also fixes a failing test case when having pydot & graphviz installed.

### DIFF
--- a/keras/utils/vis_utils.py
+++ b/keras/utils/vis_utils.py
@@ -86,6 +86,7 @@ def model_to_dot(
     subgraph=False,
     layer_range=None,
     show_layer_activations=False,
+    show_trainable=False,
 ):
     """Convert a Keras model to dot format.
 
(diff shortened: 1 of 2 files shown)


fchollet:
Thanks for the PR! Can you provide a few screenshots of what the plots look like?

mfidabel:
> Thanks for the PR! Can you provide a few screenshots of what the plots look like?

Sure, It would look like this:

195469866-dbb40180-f2b7-4b82-8536-32eedeaa59d4

For something like this:

```python
import keras

model = keras.Sequential(
    [
        keras.Input(shape=(230, 230, 3)),
        keras.layers.Conv2D(3, 1),
        keras.layers.MaxPooling2D(2),
        keras.layers.Dense(10),
    ]
)

model.get_layer(index=0).trainable = False

keras.utils.plot_model(
    model,
    to_file="model_trainable.png",
    show_trainable=True,
)
```

Where the Convolutional Layer is the "not trainable" layer.

gbaned:
@fchollet Can you please take a look on the above comment from @mfidabel. Thank you!

fchollet (approved):
LGTM, thanks!

PR #17150, opened 2022-10-13
Updated Function _create_seed() in keras.backend

### LINKED ISSUE #17149: Function _create_seed() in keras.backend causes DeprecationWarning in Python 3.10
In [line 2032 of keras.backend.py](https://github.com/keras-team/keras/blob/af1408d3255e3db9067522762e22a6c454c56654/keras/backend.py#L2032) we call random.randint(1, 1e9), which causes DeprecationWarning in Python 3.10 and will raise an error in some future version of Python. I suggest to replace it with `random.randint(1, int(1e9))` or `random.randint(1, 1000000000)`.

sushreebarsa:
random.randint(1, 1e9) is replaced with random.randint(1, int(1e9)) in Function _create_seed() which was causing deprecation warning. 
Fixes #17149

### DIFF
--- a/keras/backend.py
+++ b/keras/backend.py
@@ -2029,7 +2029,7 @@ class RandomGenerator(tf.__internal__.tracking.AutoTrackable):
         elif getattr(_SEED_GENERATOR, "generator", None):
             return _SEED_GENERATOR.generator.randint(1, 1e9)
         else:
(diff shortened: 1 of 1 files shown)


fchollet (approved):
LGTM, thanks

fchollet (approved):

PR #17170, opened 2022-10-19
Changed .format to f-string for better readability

AdityaPunetha:
Changed .format to f-string for better readability and performance as per PEP 498


fchollet (approved):
LGTM

Issue #17187, opened 2022-10-26
New optimizers are incompatible with `jit_compile` and `MirroredStrategy`


lgeiger:
**System information**.
- Have I written custom code (as opposed to using a stock example script provided in Keras): No
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Debian
- TensorFlow installed from (source or binary): binary
- TensorFlow version (use command below): 2.11.0-rc1
- Python version: 3.9.9
- GPU model and memory: 4 x NVIDIA T4 on GCP VM

**Describe the problem**.

The new optimizers in Keras 2.11 seem to be incompatible with multi GPU `MirroredStrategy` training and jit compilation.

**Describe the current behavior**.

The following code example, taken from the [docs](https://www.tensorflow.org/guide/distributed_training#use_tfdistributestrategy_with_keras_modelfit), will fail to execute in a multi GPU environment with the mentioned runtime error:
```python
import tensorflow as tf

dataset = tf.data.Dataset.from_tensors(([1.0], [1.0])).repeat(100).batch(10)

with tf.distribute.MirroredStrategy().scope():
    model = tf.keras.Sequential([tf.keras.layers.Dense(1, input_shape=(1,))])

model.compile(loss="mse", jit_compile=True, optimizer=tf.keras.optimizers.SGD())
model.fit(dataset, epochs=2)
```
```
RuntimeError: `merge_call` called while defining a new graph or a tf.function. This can often happen if the
  function `fn` passed to `strategy.run()` contains a nested `@tf.function`, and the nested `@tf.function` contains
  a synchronization point, such as aggregating gradients (e.g, optimizer.apply_gradients), or if the function `fn`
  uses a control flow statement which contains a synchronization point in the body. Such behaviors are not yet
  supported. Instead, please avoid nested `tf.function`s or control flow statements that may potentially cross a
  synchronization boundary, for example, wrap the `fn` passed to `strategy.run` or the entire `strategy.run` inside
  a `tf.function` or move the control flow out of `fn`. If you are subclassing a `tf.keras.Model`, please avoid
  decorating overridden methods `test_step` and `train_step` in `tf.function`.
```
The same example correctly works when either running on a single GPU or switching back to the legacy optimizers.

When compiling the model without jit compilation
```python
model.compile(loss="mse", optimizer=tf.keras.optimizers.SGD())
```
execution also fails due to XLA not being able to access resources on different devices:
```
[...]

PR #17197, opened 2022-10-28
Remove unused `amsgrad` argument in SGD

lgeiger:
This PR removes the unused `amsgrad` argument from `keras.optimizers.SGD`. This argument should've only been part of the Adam optimizer and probably was added to SGD by accident.

Noticed while trying to debug #17187.


fchollet (approved):
LGTM, thank you

PR #17239, opened 2022-11-14
support to compat python 3.11

hongxu-jia:
The python 3.11 has removed long-deprecated inspect methods [1], use collections.namedtuple to instead

[1] https://github.com/python/cpython/commit/d89fb9a5a610a257014d112bdceef73d7df14082

Signed-off-by: Hongxu Jia <hongxu.jia@windriver.com>


fchollet (approved):
LGTM, thank you

Issue #14030, opened 2020-05-07
Disable printing in flow_from_directory


MichalWeisman:
Is there a way to disable printing when using ImageDataGenerator's flow_from_directory?
I'm trying to use a custom generator that calls flow_from_directory for each batch (for random sampling with uniform distribution of classes in each batch). 
So when I call fit_generator, this ruins the standard epochs printing - a line is printed for each batch and not for each epoch.

This is my code:
```lang-python
    def __getitem__(self, idx):
       # create a batch dataframe with one random sample for each class
        batch_df = pd.DataFrame()
        for class_name in batch_class_names:
            rand_row = self.df_dict[class_name].sample(n = 1)
            batch_df = batch_df.append(rand_row, ignore_index=True)
            
       # create generator
        batch_gen = self.generator.flow_from_dataframe(
            dataframe=batch_df, directory=self.directory, x_col="filename",
            y_col="brand", classes=self.class_names,
            class_mode="categorical",
            target_size=self.image_dims, color_mode="rgb", batch_size=batch_df.shape[0], shuffle=self.shuffle)

       # return batch
        return (next(batch_gen))
```

And this is the output (a row for each batch instead of a row for each epoch):
```
1/100 [..............................] - ETA: 20:17 - loss: 3.4675 - tpr_metric: 0.0000e+00Found 9 validated image filenames belonging to 28 classes.
  2/100 [..............................] - ETA: 11:14 - loss: 3.4366 - tpr_metric: 0.0000e+00Found 9 validated image filenames belonging to 28 classes.
  3/100 [..............................] - ETA: 8:13 - loss: 3.3988 - tpr_metric: 0.0000e+00 Found 9 validated image filenames belonging to 28 classes.
  4/100 [>.............................] - ETA: 6:41 - loss: 3.4289 - tpr_metric: 0.0000e+00Found 9 validated image filenames belonging to 28 classes.
  5/100 [>.............................] - ETA: 5:46 - loss: 3.4379 - tpr_metric: 0.0000e+00Found 9 validated image filenames belonging to 28 classes.
  6/100 [>.............................] - ETA: 5:08 - loss: 3.4358 - tpr_metric: 0.0000e+00Found 9 validated image filenames belonging to 28 classes.
  7/100 [=>............................] - ETA: 4:41 - loss: 3.4181 - tpr_metric: 0.0000e+00Found 9 validated image filenames belonging to 28 classes.
```

Mirandatz:
Guys? Please?

[...]

PR #17346, opened 2022-12-14
Tzm/optional verbosity

Mirandatz:
The `xxx_dataset_from_directory` family of functions unconditionally write to stdout, which is undesirable [[1](https://stackoverflow.com/questions/63650807/disable-the-keras-found-images-belonging-to-classes-message/65351797#65351797), [2](https://github.com/keras-team/keras/issues/14030)].

This patch adds a new `verbosity` parameter to them, allowing the user to silence. The default value of the new parameter preserves the current behavior.

### DIFF
--- a/keras/backend.py
+++ b/keras/backend.py
@@ -31,6 +31,7 @@ import tensorflow.compat.v2 as tf
 
 from keras import backend_config
 from keras.distribute import distribute_coordinator_utils as dc
+from keras.dtensor import dtensor_api as dtensor
 from keras.engine import keras_tensor
 from keras.utils import control_flow_util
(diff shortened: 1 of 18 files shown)


fchollet (commented):
Thanks for the PR. Instead of adding a new argument, just switch from `print()` to `io_utils.print_msg()`. Then in your project you can disable interactive logging to avoid printing to stdout (via `keras.utils.disable_interactive_logging()`).

Mirandatz:
Reverted the changes and replaced the function calls, as suggested.

fchollet (approved):
LGTM, thanks!

PR #17374, opened 2022-12-25
Update README.md to fix wrong links

NourEldin-Osama:
Fix the 404 page not found for the Keras functional API

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,9 +52,9 @@ and you can export your Keras models to run in the browser or on a mobile device
 ## First contact with Keras
 
 The core data structures of Keras are __layers__ and __models__.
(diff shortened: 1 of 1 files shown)


hertschuh:
@NourEldin-Osama ,
Thank you for taking the time to do this. Can you also update the other 2 links? The link right above and the the link right below the one you updated.
Thanks!

NourEldin-Osama:
> @NourEldin-Osama , Thank you for taking the time to do this. Can you also update the other 2 links? The link right above and the link right below the one you updated. Thanks!

I just updated the other 2 links

fchollet (approved):
LGTM, thanks

PR #17384, opened 2022-12-30
TerminateOnNaN epoch level callback

abinthomasonline:
`epoch` frequency support for TerminateOnNaN callback. 

Batch level hook fails when used with distributed strategies, as `logs` is not of type `dict` (tensorflow.python.distribute.coordinator.values.RemoteValueImpl)


abinthomasonline:
fixed formatting issue

rchao (changes requested):
Thanks for the PR! Can you add a test to make sure the newly added paths are covered and the test is passing?

keras/callbacks.py
```diff
@@ -985,23 +985,73 @@ def on_epoch_end(self, epoch, logs=None):
...
+            `'batch'` is a synonym for `1`, meaning that check will happen 
+            every batch. Defaults to `'batch'`.
+    """
+
+    def __init__(self, check_freq="batch"):
         super().__init__()
+        self.check_freq = 1 if check_freq == "batch" else check_freq
+        if self.check_freq != "epoch" and not isinstance(self.check_freq, int):
+            raise ValueError(
+                "Unrecognized check_freq: {}".format(self.check_freq)
```
hertschuh, line 1003: It is helpful to additionally mention what was expected.
Also, f-strings are a bit more elegant. How about:
```python
f"Unrecognized check_freq: {self.save_freq}"
"Expected 'batch', 'epoch' or integer"
```

keras/callbacks.py
```diff
@@ -985,23 +985,73 @@ def on_epoch_end(self, epoch, logs=None):
...
+            )
+        self._batches_seen_since_last_saving = 0
```
hertschuh, line 1005: While this variable name made sense for the checkpoint callback, it should have a more appropriate name for this use case. Please rename to `_batches_seen_since_last_check`.

hertschuh (changes requested):
Thanks for the Pull Request!
A couple of suggestions:

abinthomasonline:
> Thanks for the PR! Can you add a test to make sure the newly added paths are covered and the test is passing?

added tests and they are passing.

keras/callbacks.py
abinthomasonline: updated

keras/callbacks.py
abinthomasonline: renamed

abinthomasonline:
@hertschuh @rchao 
anything else needed here?

keras/callbacks.py
```diff
@@ -985,23 +985,74 @@ def on_epoch_end(self, epoch, logs=None):
...
+
         if loss is not None:
             loss = tf_utils.sync_to_numpy_or_python_type(loss)
             if np.isnan(loss) or np.isinf(loss):
-                io_utils.print_msg(
-                    f"Batch {batch}: Invalid loss, terminating training"
-                )
+                if batch is not None:
+                    io_utils.print_msg(
+                        f"Batch {batch}: Invalid loss, terminating training"
```
rchao, line 1019: Suggest: also print the epoch number to be more clear

keras/callbacks.py
```diff
@@ -985,23 +985,74 @@ def on_epoch_end(self, epoch, logs=None):
...
+                    )
+                else:
+                    io_utils.print_msg(
+                        f"Epoch {epoch}: Invalid loss, terminating training"
+                    )
                 self.model.stop_training = True
 
+    def _should_check_on_batch(self, batch):
```
rchao, line 1027: Since the logic is also used by ModelCheckpoint callback, can this be made a method of Callback so the subclasses can share it?

rchao (changes requested):
Sorry for the delay! It's been a busy week. Largely looks good, just a couple quick suggestions.

keras/callbacks.py
abinthomasonline: The epoch number is not passed down to the `on_batch_end()` hooks. The change has to happen deep inside the training loop.

keras/callbacks.py
abinthomasonline: Yes, this can be used across TerminateOnNaN, ModelCheckpoint, BackupAndRestore and TensorBoard. I'll add this.

hmc-cs-mdrissi:
Looking at original issue,

> Batch level hook fails when used with distributed strategies, as logs is not of type dict (tensorflow.python.distribute.coordinator.values.RemoteValueImpl)

why not use [RemoveValue.get](https://www.tensorflow.org/api_docs/python/tf/distribute/experimental/coordinator/RemoteValue#get) to convert removevalue to concrete one in batch hook? Why can't Parameter server/other distributed strategies not support batch level terminate on nan?

abinthomasonline:
> Looking at original issue,
> 
> > Batch level hook fails when used with distributed strategies, as logs is not of type dict (tensorflow.python.distribute.coordinator.values.RemoteValueImpl)
> 
> why not use [RemoveValue.get](https://www.tensorflow.org/api_docs/python/tf/distribute/experimental/coordinator/RemoteValue#get) to convert removevalue to concrete one in batch hook? Why can't Parameter server/other distributed strategies not support batch level terminate on nan?

I can include this as well, but calling `.get()` on remote value comes with sync cost and can slow down distributed training.

hmc-cs-mdrissi:
Yes but that is controllable by check_freq. If you use check_freq of 1 and call .get() every batch then training will probably become slow. If you use check_freq of 100-1000 batches then it’s more likely sync effect will be minor. Exact impact will depend a lot on your model/dataset but you can adjust it to minimize performance impact.

keras/callbacks.py
abinthomasonline: @rchao ignore my above comment, I included epoch number by following the design of ModelCheckpoint.

keras/callbacks.py
abinthomasonline: @rchao Updated!

abinthomasonline:
@rchao @hertschuh 
anything remaining?

keras/callbacks.py
```diff
@@ -985,23 +1005,62 @@ def on_epoch_end(self, epoch, logs=None):
 
 @keras_export("keras.callbacks.TerminateOnNaN")
 class TerminateOnNaN(Callback):
-    """Callback that terminates training when a NaN loss is encountered."""
+    """Callback that terminates training when a NaN loss is encountered.
 
-    def __init__(self):
+    Args:
```
rchao, line 1010: Let's be consistent with other classes and move the Args section to class doc

keras/callbacks.py
```diff
@@ -932,6 +934,24 @@ def _implements_predict_batch_hooks(self):
             self.on_predict_batch_begin
         ) or not generic_utils.is_default(self.on_predict_batch_end)
 
+    def _should_trigger_on_batch(self, trigger_freq, batch):
+        """Determines if the callback should be called for a given batch,
```
rchao, line 938: Let's keep the summary of the method within one line

rchao (changes requested):
Largely looks good to me, just a couple suggestion on the docs. Thanks and sorry about the delay.

gbaned:
Hi @abinthomasonline Can you please check @rchao's comments and keep us posted ? Thank you!

keras/callbacks.py
abinthomasonline: it is already in the class doc

keras/callbacks.py
abinthomasonline: done.

abinthomasonline:
@gbaned @rchao Updated. 
Sorry, I missed the updates on this thread.

rchao:
Update: pradeepkuppla@ is helping with some internal testing, and we'll circle back once we have more updates.

sachinprasadhs:
@rchao , this is failing TAP test with below error.
Is it something actionable from user side?
```
third_party/py/keras/api/tests/api_compatibility_test.py", [line 293](https://cs.corp.google.com/piper///depot/google3/third_party/py/keras/api/tests/api_compatibility_test.py?l=293&ws=tap-presubmit-server/54210978&snapshot=2), in _AssertProtoDictEquals
    self.fail(
AssertionError: 1 differences found between API and golden.
```

rchao:
> @rchao , this is failing TAP test with below error. Is it something actionable from user side?
> 
> ```
> third_party/py/keras/api/tests/api_compatibility_test.py", [line 293](https://cs.corp.google.com/piper///depot/google3/third_party/py/keras/api/tests/api_compatibility_test.py?l=293&ws=tap-presubmit-server/54210978&snapshot=2), in _AssertProtoDictEquals
>     self.fail(
> AssertionError: 1 differences found between API and golden.
> ```

@abinthomasonline can you check if following the [instruction here](https://github.com/keras-team/keras/blob/master/keras/api/tests/README.txt) helps with resolving this?

gbaned:
Hi @abinthomasonline Any update on this PR? Please. Thank you!

sachinprasadhs:

Issue #17392, opened 2023-01-04
`timeseries_dataset_from_array` gives empty output on edge case


basjacobs93:
**Describe the problem**.

The `keras.utils.timeseries_dataset_from_array` function does not give the correct output when `sampling_rate > 1` and the size of `data`  is exactly big enough to extract 1 dataset.

**Describe the current behavior**.
The result is empty

**Describe the expected behavior**.
The result contains 1 element

**[Contributing](https://github.com/keras-team/keras/blob/master/CONTRIBUTING.md)**.

- Do you want to contribute a PR? (yes/no): yes
- Briefly describe your candidate solution(if contributing): replace `+ 1` with `+ sampling_rate` in https://github.com/keras-team/keras/blob/16bc9f8b4b3cfc4d20f8ad683a8c400dcac5b35e/keras/utils/timeseries_dataset.py#L212.

**Standalone code to reproduce the issue**.
```
from keras.utils import timeseries_dataset_from_array
ds = timeseries_dataset_from_array(
    data=np.arange(99),
    targets=None,
    sequence_length=50,
    sampling_rate=2,
    batch_size=None
)

len([a for a in ds])
```
gives `0` but should give `1` (corresponding to the 50 indices `[0, 2, 4, ..., 96, 98]`)

tilakrayal:
@basjacobs93,
We would prefer a PR for the validation, Could you please feel free to submit a PR for the requested change to be made. Thank you!

tilakrayal:
The pr #17396 has been assigned for reviewing and once it is merged this issue will move to closed status.

PR #17396, opened 2023-01-05
Fix timeseries_dataset_from_array counts when sequence_stride > 1

basjacobs93:
Fixes https://github.com/keras-team/keras/issues/17392

### DIFF
--- a/keras/utils/timeseries_dataset.py
+++ b/keras/utils/timeseries_dataset.py
@@ -209,7 +209,7 @@ def timeseries_dataset_from_array(
 
     # Determine the lowest dtype to store start positions (to lower memory
     # usage).
-    num_seqs = end_index - start_index - ((sequence_length - 1) * sampling_rate)
+    num_seqs = end_index - start_index - (sequence_length - 1) * sampling_rate
     if targets is not None:
         num_seqs = min(num_seqs, len(targets))
     if num_seqs < 2147483647:
--- a/keras/utils/timeseries_dataset.py
+++ b/keras/utils/timeseries_dataset.py
@@ -84,7 +84,7 @@ def timeseries_dataset_from_array(
 
     Example 1:
 
-    Consider indices `[0, 1, ... 99]`.
+    Consider indices `[0, 1, ... 98]`.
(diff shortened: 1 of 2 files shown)


basjacobs93:
I believe the test `test_sampling_rate` (that I edited) was indeed also faulty, because if `data = np.arange(100)` (i.e. `0, 1, ..., 99`), then there are 84 such sequences of length 9 with sampling rate 2 possible, not 83:
```
0, 2, 4, 6, 8, 10, 12, 14, 16
1, 3, 5, 7, 9, 11, 13, 15, 17
2, 4, 6, 8, 10, 12, 14, 16, 18
...
81, 83, 85, 87, 89, 91, 93, 95, 97
82, 84, 86, 88, 90, 92, 94, 96, 98
83, 85, 87, 89, 91, 93, 95, 97, 99
```
which results in 17 batches of size 5, of which the last has 4 elements, not 3.
But please verify that my reasoning is correct.

qlzh727 (changes requested):
Thanks for the fix. Can u also update the example above at https://github.com/keras-team/keras/blob/16bc9f8b4b3cfc4d20f8ad683a8c400dcac5b35e/keras/utils/timeseries_dataset.py#L102. I think the last sequence of that example should be 
[81, 83, .... 99], and I believe the error of this example is also caused by the current issue. 

Thanks.

f1nn1711 (approved):

keras/utils/timeseries_dataset.py
```diff
@@ -209,7 +209,7 @@ def timeseries_dataset_from_array(
 
     # Determine the lowest dtype to store start positions (to lower memory
     # usage).
-    num_seqs = end_index - start_index - (sequence_length * sampling_rate) + 1
+    num_seqs = end_index - start_index - ((sequence_length - 1) * sampling_rate)
```
qlzh727, line 212: the extra bracket for `(sequence_length - 1) * sampling_rate` can be skipped.

basjacobs93: Okay, removed them in the new commit

qlzh727 (approved):
Thanks for the fix.

PR #17399, opened 2023-01-06
Avoid resetting validation dataset after every epoch

abinthomasonline:
In the current implementation of `model.fit()`, the validation dataset is reset from epoch to epoch.  This behaviour is undesirable when dealing with large datasets.

This PR addresses this issue by caching the epoch iterator.


ianstenbit:
This approach seems reasonable to me but someone with more expertise on this should probably take a look. Maybe @rchao or @qlzh727 ?

rchao:
Thanks for the PR! It is intended that the validation dataset is reset at new epochs as that gives a fair comparison of evaluation results across the epochs, so we do not intend to have this change at this time.

PR #17403, opened 2023-01-08
Fix index_directory directory structure ignoring when passing labels explicitly

wwossname:
If you pass explicit `labels` (list/tuple) to `index_directory`, the directory structure should be ignored and indexing of the files start from the parent `directory`. After all, if the files were already placed in class-specific subdirectories, you could pass `labels="inferred"`.

The change I made simply allows placing the files directly in the parent `directory`, when passing `labels` explicitly, while using subdirectories will continue to work (passing `labels=None` already allows this). Currently files in the parent `directory` are not found. I also updated the docstring to match both the change in behaviour and the docstrings of all its callers such as `image_dataset_from_directory` (from which I copied the wording).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -294,7 +294,7 @@ mind.
 -   You should add any new applications to the unit tests defined in
     `applications_test.py` and `applications_load_weight_test.py`.
 -   For backwards compatibility, all applications should provide a
-    `preprocess_input()` function. For new applciations, you should leave the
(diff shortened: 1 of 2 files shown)


fchollet (commented):
Thank you for the PR! It looks good. Please add a unit test for this change (a test that would have previously failed and that passes after this PR).

wwossname:
Thank you, I added the unit test

fchollet (approved):
LGTM, thank you

PR #17431, opened 2023-01-15
Handle dict of metrics in distributed training

froody:
When training with a model with a subclass `tf.keras.metrics.Metric` which which returns a dict, e.g.

```
class CustomMetric(tf.keras.metrics.Metric):
    def result(self):
        return {"good", tf.convert_to_tensor([1.0]), "bad": tf.convert_to_tensor([0.0])}
```

I got an exception in `merge_fn_wrapper` when trying to call `array_ops.identity` on a `Dict[str, tf.Tensor]`. This fix allows me to do distributed training with the above metric.

### DIFF
--- a/keras/utils/metrics_utils.py
+++ b/keras/utils/metrics_utils.py
@@ -174,7 +174,13 @@ def result_wrapper(result_fn):
                 # Wrapping result in identity so that control dependency between
                 # update_op from `update_state` and result works in case result
                 # returns a tensor.
-                return tf.identity(result)
+                if isinstance(result, dict):
+                    return {
+                        key: array_ops.identity(value)
+                        for key, value in result.items()
+                    }
+                else:
+                    return array_ops.identity(result)
 
             # Wrapping result in merge_call. merge_call is used when we want to
             # leave replica mode and compute a value in cross replica mode.


keras/utils/metrics_utils.py
```diff
@@ -174,7 +174,13 @@ def merge_fn_wrapper(distribution, merge_fn, *args):
                 # Wrapping result in identity so that control dependency between
                 # update_op from `update_state` and result works in case result
                 # returns a tensor.
-                return tf.identity(result)
+                if isinstance(result, dict):
```
qlzh727, line 177: I think the logic between L140 and L157 can be reuse here, how about extract that into a function and reuse it?

keras/utils/metrics_utils.py
```diff
@@ -174,7 +174,13 @@ def merge_fn_wrapper(distribution, merge_fn, *args):
...
+                    return {
+                        key: array_ops.identity(value)
```
qlzh727, line 179: Note that keras use the public TF API. This should be tf.identity(), and array_ops is a private TF API.

qlzh727 (changes requested):

gbaned:
Hi @froody Can you please check @qlzh727's comments and resolve conflicts?. Thank you!

gbaned:
Hi @froody Any update on this PR? Please. Thank you!

PR #17448, opened 2023-01-19
Support Parameter Server strategy Batch Tensorboard callback

hmc-cs-mdrissi:
# Summary
Recently there was a [pr](https://github.com/keras-team/keras/pull/17142) to support batch frequency metrics in Tensorboard callback. That PR supported most strategies, but did not cover Parameter Server training. This just extends it to handle parameter server. I think this should finish resolving this [issue](https://github.com/keras-team/keras/issues/16173).

# Testing
I'm running couple parameter server training on google cloud ai training and tensorboard metric curves are being produced middle epoch as expected. I've also tested locally with in process cluster that metrics get plotted as expected.

Screenshot 2023-01-18 at 10 17 42 PM

Performance wise this does add 1 sync point, but user can adjust update_freq accordingly to how often they want it to flush/sync. I've tested on baseline job and a job with this change that large cluster (~60 workers, ~10 PS) parameter server training with update_freq=500 has minimal effect on overall job completion time. update_freq=1 does have large slowdown so if using PS training you'll need to set update_freq appropriately.

### DIFF
--- a/keras/callbacks.py
+++ b/keras/callbacks.py
@@ -2366,7 +2366,6 @@ class TensorBoard(Callback, version_utils.TensorBoardVersionSelector):
           Note however that writing too frequently to TensorBoard can slow down
           your training, especially when used with `tf.distribute.Strategy` as
           it will incur additional synchronization overhead.
-          Use with `ParameterServerStrategy` is not supported.
           Batch-level summary writing is also available via `train_step`
           override. Please see
           [TensorBoard Scalars tutorial](https://www.tensorflow.org/tensorboard/scalars_and_keras#batch-level_logging)  # noqa: E501
@@ -2791,10 +2790,9 @@ class TensorBoard(Callback, version_utils.TensorBoardVersionSelector):
                 step=self._train_step,
             )
 
-        # `logs` isn't necessarily always a dict. For example, when using
-        # `tf.distribute.experimental.ParameterServerStrategy`, a
-        # `tf.distribute.experimental.coordinator.RemoteValue` will be passed.
-        # For now, we just disable `update_freq` in those cases.
+        if isinstance(logs, tf.distribute.experimental.coordinator.RemoteValue):
+            logs = logs.get()
+
         if isinstance(logs, dict):
             for name, value in logs.items():
                 tf.summary.scalar("batch_" + name, value, step=self._train_step)
--- a/keras/callbacks.py
+++ b/keras/callbacks.py
@@ -2790,11 +2790,9 @@ class TensorBoard(Callback, version_utils.TensorBoardVersionSelector):
                 step=self._train_step,
             )
 
-        should_record = tf.equal(self._train_step % self.update_freq, 0)
-        if (
-            isinstance(logs, tf.distribute.experimental.coordinator.RemoteValue)
-            and should_record
-        ):
+        if isinstance(
+            logs, tf.distribute.experimental.coordinator.RemoteValue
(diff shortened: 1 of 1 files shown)


keras/callbacks.py
```diff
@@ -2791,10 +2790,13 @@ def on_train_batch_end(self, batch, logs=None):
                 step=self._train_step,
             )
 
-        # `logs` isn't necessarily always a dict. For example, when using
-        # `tf.distribute.experimental.ParameterServerStrategy`, a
-        # `tf.distribute.experimental.coordinator.RemoteValue` will be passed.
-        # For now, we just disable `update_freq` in those cases.
+        should_record = tf.equal(self._train_step % self.update_freq, 0)
```
rchao, line 2793: Unfortunately doing this with ParameterServerStrategy will result in unnecessary syncs across the cluster within an epoch. Using a thread to periodically log the metrics manually would work better for PSS. Or, alternatively, we've planned on time-based (asynchronous) callbacks but they're not available yet.

Thanks for the PR!

hmc-cs-mdrissi: ? I’m measured the performance of this solution with actual large training jobs and if update frequency is kept high (100) I don’t see any speed impact at all. I also checked in debugger and while logs is a remote value, update step and update freq are not remote values and do not appear to trigger a sync.

hmc-cs-mdrissi: Screenshot 2023-01-19 at 9 33 45 AM
Succeeded (1 hr 3 min)

vs

Screenshot 2023-01-19 at 9 35 28 AM
Succeeded (59 min)

The first is training with Parameter Server strategy without this patch. The second is with this patch. The performance impact is less then noise in training speed.

rchao: Understood - I think it's expected that with a high number of `update_freq`, we do see minimal performance impact because it's far enough between syncs, but they do sync when `logs.get()` is called.

An issue here is that with the asynchronous training with `ParameterServerStrategy`, the time at which the `on_batch_end` is called is not really when a batch is finished executing, but when a batch is finished being scheduled to be executed on a worker. That is, if you have `update_freq=500`, the callback is called when the 500th batch has been just "scheduled", at which point very few train functions may have been executed if the function is taking a long time, so it could be a misleading metric. And, I felt having a time-based treatment of this may be less confusing. What do you think?

hmc-cs-mdrissi: I think time based async callback is better, but it's also question of when is that likely to come? This solution has weaknesses, but is small change and works reasonably if you tune update_freq. There are other callbacks usable in parameter server with similar need to tune update_freq like ModelCheckpoint callback. Ups and downs in cpu utilization from the job I showed mainly come from model checkpointing.

I can update documentation to be clearer that with PS you should pick high update_freq for performance.

Right now without this pr, tensorboard callback feels unusable on Parameter server with long epochs. I commonly train models for only 1 or 2 epochs, so epoch level metrics give me almost no visualization of model metrics.

rchao: Thanks for the comment and your patience! Let me bring this to the team and get back to you with an eta on time-based callback. We're worried calling the batch-level calls between scheduling can be misleading to users, so for now I think the best solution would be for you to have a subclassed TensorBoard callback for your specific use case. Thanks for the understanding!

hmc-cs-mdrissi: I'll go ahead and close this given response. If there is desire to re-visit this/add batch metric support to tensorboard callback happy to re-open then (or have news about time based callbacks).

PR #17461, opened 2023-01-23
Update MHA docstrings

haifeng-jin:
Update the MHA docstrings to use 4-space indents.
No changes except for spaces and line breaks.


qlzh727 (commented):
Is it possible to check the public doc site and see if this change will affect any of that? TF itself was always using 2 space indent, and I am not sure the impact of this change.,

haifeng-jin:
@qlzh727 Thanks! I will check that.

haifeng-jin:
Just checked, the doc website generator can correctly parse 4 spaces. The generated page looks good.

Issue #17433, opened 2023-01-16
keras.models.clone_model creates different weight names for RNN layers


mergian:
**System information**.
- Have I written custom code (as opposed to using a stock example script provided in Keras): no
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): CentOS
- TensorFlow installed from (source or binary): PYPI
- TensorFlow version (use command below): v2.11.0-rc2-17-gd5b57ca93e5 2.11.0
- Python version: 3.7.12

**Describe the problem**.

When cloning a Keras RNN model using `keras.models.clone_model(...)` the weight names of the inner RNN cells get incremented instead of being cloned. The problem does not occur when using `save_model(...)` and then `load_model(...)`.

**Describe the current behavior**.

Here an example:
```python
import tensorflow as tf

inputs = tf.keras.Input(batch_shape=[1, 2, 3])
layer = tf.keras.layers.SimpleRNN(units=5)(inputs)
model = tf.keras.Model(inputs, layer)

print("Original:")
for w in model.weights: print(w.name)
print()

cloned = tf.keras.models.clone_model(model)

print("Cloned:")
for w in cloned.weights: print(w.name)
print()

tf.keras.models.save_model(model, 'save')
saved = tf.keras.models.load_model('save')

print("Saved:")
for w in saved.weights: print(w.name)
```

This produces:
```
Original:
simple_rnn/simple_rnn_cell/kernel:0
simple_rnn/simple_rnn_cell/recurrent_kernel:0
simple_rnn/simple_rnn_cell/bias:0

Cloned:
simple_rnn/simple_rnn_cell_1/kernel:0
simple_rnn/simple_rnn_cell_1/recurrent_kernel:0
simple_rnn/simple_rnn_cell_1/bias:0

Saved:
simple_rnn/simple_rnn_cell/kernel:0
simple_rnn/simple_rnn_cell/recurrent_kernel:0
simple_rnn/simple_rnn_cell/bias:0
```

As you can see, the cloned model has `simple_rnn_cell_1` instead as `simple_rnn_cell` as name. When using other layers, i.e., Conv the weight names are correct. The problem occurs with all types of RNN layers (SimpleRNN, LSTM and GRU).

**Describe the expected behavior**.
The names are set correctly ;)

**[Contributing](https://github.com/keras-team/keras/blob/master/CONTRIBUTING.md)**.

- Do you want to contribute a PR? (yes/no): no -> I don't think I know enough of Keras internal to contribute here.
[...]

PR #17465, opened 2023-01-24
keras.models.clone_model creates different weight names for RNN layers

mergian:
This pull request fixes: #17433

It's the easiest solution I came up with.

Another solution would be to add `cell_name` argument to all RNN layers, but that changes the API.

I'm not sure if 'cell' should be used as name, or if we should stick to 'simple_rnn_cell', 'lstm_cell', 'gru_cell' instead.

It also works with saving models AND also when loading old saved models. As in this case, the RNN cell's name is stored in the saved model, so it will correctly be loaded.


qlzh727 (changes requested):
Thanks for the PR. Please format you PR as well.

keras/layers/rnn/base_conv_lstm.py
```diff
@@ -489,6 +489,7 @@ def __init__(
             bias_constraint=bias_constraint,
             dropout=dropout,
             recurrent_dropout=recurrent_dropout,
+            name="lstm_cell",
```
qlzh727, line 492: This should be conv_lstm_cell I think.

qlzh727 (changes requested):

keras/layers/rnn/base_conv_lstm.py
mergian: Yes you are right 🤦🏻‍♂️

qlzh727 (approved):

qlzh727 (approved):

PR #17499, opened 2023-01-30
Validate keras.optimizers.schedules.CosineDecay Arg

SuryanarayanaY:
At present the code for `keras.optimizers.schedules.CosineDecay` not validating value for the argument `decay_steps`.As per document `decay_steps` should be : A scalar int32 or int64 Tensor or a Python number.Even if users passes Zero or Negative number it is not throwing any error and in the github issue #53284 user passing negative value to this argument but it was not raised any error and training happened without updating model weights.Hence proposed changes in code to validate arguments and raise suitable error if not valid argument. Fixes github issue #53284 .

### DIFF
--- a/keras/optimizers/schedules/learning_rate_schedule.py
+++ b/keras/optimizers/schedules/learning_rate_schedule.py
@@ -641,6 +641,12 @@ class CosineDecay(LearningRateSchedule):
         self.decay_steps = decay_steps
         self.alpha = alpha
         self.name = name
+        
+        #validating decay_steps argument to raise error if passed
+        #negative values or zero or float values
+        if not isinstance(self.decay_steps,int) or self.decay_steps<=0:
+            raise ValueError("'decay_steps' should be a nonzero "+
+                             "positive integer")
 
     def __call__(self, step):
         with tf.name_scope(self.name or "CosineDecay"):


keras/optimizers/schedules/learning_rate_schedule.py
```diff
@@ -641,6 +641,12 @@ def __init__(
         self.decay_steps = decay_steps
         self.alpha = alpha
         self.name = name
+        
+        #validating decay_steps argument to raise error if passed
+        #negative values or zero or float values
+        if not isinstance(self.decay_steps,int) or self.decay_steps<=0:
```
qlzh727, line 647: Note that the decay_steps could be a tensor (from the docstring), so the current check will fail.

qlzh727 (changes requested):
Please also fix the format.

keras/optimizers/schedules/learning_rate_schedule.py
```diff
@@ -641,6 +641,14 @@ def __init__(
         self.decay_steps = decay_steps
         self.alpha = alpha
         self.name = name
+        
+        #validating decay_steps argument to raise error if passed
+        #negative values or zero or float values or float Tensors
+        if isinstance(self.decay_steps,tf.Tensor):
+            self.decay_steps = self.decay_steps.numpy()
```
qlzh727, line 648: Note that the tensor here could be a graph tensor, so the numpy() method might raise an error. Also there is no point to convert a tensor back to numpy and then convert back to tensor again in __call__.

keras/optimizers/schedules/learning_rate_schedule.py
qlzh727: I think it should be fine to only check the value when it is int, and you can skip the tensor value here.

qlzh727 (changes requested):
Please fix the format of the change.

PR #17540, opened 2023-02-08
Added a link to the TextVectorization API Reference Page.

Neeshamraghav012:
I have added a URL of the TextVectorization API in the StringLookup Layer page so that users can easily look at and experiment with the API.

### DIFF
--- a/keras/layers/preprocessing/string_lookup.py
+++ b/keras/layers/preprocessing/string_lookup.py
@@ -35,7 +35,7 @@ class StringLookup(index_lookup.IndexLookup):
     This layer translates a set of arbitrary strings into integer output via a
     table-based vocabulary lookup. This layer will perform no splitting or
     transformation of input strings. For a layer than can split and tokenize
-    natural language, see the `TextVectorization` layer.
+    natural language, see the [TextVectorization](https://keras.io/api/layers/preprocessing_layers/text/text_vectorization/#textvectorization-class) layer.
 
     The vocabulary for the layer must be either supplied on construction or
     learned via `adapt()`. During `adapt()`, the layer will analyze a data set,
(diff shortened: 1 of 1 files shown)


keras/layers/preprocessing/string_lookup.py
```diff
@@ -35,7 +35,7 @@ class StringLookup(index_lookup.IndexLookup):
     This layer translates a set of arbitrary strings into integer output via a
     table-based vocabulary lookup. This layer will perform no splitting or
     transformation of input strings. For a layer than can split and tokenize
-    natural language, see the `TextVectorization` layer.
+    natural language, see the [TextVectorization](https://keras.io/api/layers/preprocessing_layers/text/text_vectorization/#textvectorization-class) layer.
```
qlzh727, line 38: I think you only need to change this to `tf.keras.layers.TextVectorization`, and our doc parse will convert it to a link to the TextVectorization API page. It doesn't necessary to put the explicit URL here.

qlzh727 (changes requested):

keras/layers/preprocessing/string_lookup.py
Neeshamraghav012: I tried that but it is not behaving like a like. That's why I added the explicit URL.

qlzh727: If you check https://www.tensorflow.org/api_docs/python/tf/keras/layers/TextVectorization and the corresponding source code at https://github.com/keras-team/keras/blob/e6784e4302c7b8cd116b74a784f4b78d60e83c26/keras/layers/preprocessing/text_vectorization.py#L64. The `tf.keras.layers.StringLookup` is rendered as a link.

Neeshamraghav012:
I have done the required changes.

qlzh727 (approved):

Issue #17386, opened 2022-12-31
ConvNeXt `classifier_activation=softmax` returns linear layer


Frightera:
**Describe the problem**.

Specifying `classifier_activation=softmax` in ConvNeXt models does not work, instead it returns the model with a linear classifier layer.

Inspecting source code:
https://github.com/keras-team/keras/blob/e6784e4302c7b8cd116b74a784f4b78d60e83c26/keras/applications/convnext.py#L524-L526

Definition of `Head` function:
https://github.com/keras-team/keras/blob/e6784e4302c7b8cd116b74a784f4b78d60e83c26/keras/applications/convnext.py#L327-L348

Find the Colab [gist here](https://colab.research.google.com/gist/Frightera/31d1c09bc6c882c35f15afa1310abb6b/convnext-classifier_activation.ipynb).

If OK I can open a PR and fix it.

innat:
FYI, all models are moving to keras-cv. https://github.com/keras-team/keras-cv/pull/678

Frightera:
I see, in keras-cv the implementation is correct. Shall I close this issue?

innat:
Yes, you may. All the models from application will be moved to kcv completely soon.

innat:
@Frightera please re-open this ticket. 

> Shall I close this issue?
> yes.

I was wrong. This should be discussed and fixed right away.

Frightera:
@innat Sure, re-opened it. Should be a small and easy fix.

Frightera:
Closing as PR was merged.

PR #17544, opened 2023-02-09
Fix ConvNeXt classifier activation bug

Frightera:
Fixes the issue #17386.

Please find the bug, fix, and test [gist here](https://colab.research.google.com/gist/Frightera/93a784fc0e03a19f5804e7e14bfbecba/-17386.ipynb).

### DIFF
--- a/keras/applications/applications_test.py
+++ b/keras/applications/applications_test.py
@@ -193,7 +193,9 @@ class ApplicationsTest(tf.test.TestCase, parameterized.TestCase):
 
     @parameterized.parameters(MODEL_LIST)
     def test_application_classifier_activation(self, app):
-        model = app(weights=None, include_top=True, classifier_activation="softmax")
+        model = app(
+            weights=None, include_top=True, classifier_activation="softmax"
+        )
         last_layer_act = model.layers[-1].activation.__name__
         self.assertEqual(last_layer_act, "softmax")
 
@@ -225,7 +227,7 @@ class ApplicationsTest(tf.test.TestCase, parameterized.TestCase):
 
     @parameterized.parameters(*MOBILENET_V3_FOR_WEIGHTS)
     def test_mobilenet_v3_load_weights(
-            self, mobilenet_class, alpha, minimalistic, include_top
+        self, mobilenet_class, alpha, minimalistic, include_top
     ):
         mobilenet_class(
             input_shape=(224, 224, 3),
--- a/keras/applications/applications_test.py
+++ b/keras/applications/applications_test.py
@@ -193,10 +193,9 @@ class ApplicationsTest(tf.test.TestCase, parameterized.TestCase):
 
     @parameterized.parameters(MODEL_LIST)
     def test_application_classifier_activation(self, app):
-        last_activation = _get_last_layer_activation(
-            lambda: app(weights=None, include_top=True, classifier_activation="softmax")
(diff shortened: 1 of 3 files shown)


keras/applications/convnext.py
```diff
@@ -522,7 +525,9 @@ def ConvNeXt(
         cur += depths[i]
 
     if include_top:
-        x = Head(num_classes=classes, name=model_name)(x)
+        x = Head(num_classes=classes,
+                 classifier_activation=classifier_activation,
+                 name=model_name)(x)
         imagenet_utils.validate_activation(classifier_activation, weights)
```
qlzh727, line 531: Thanks for the fix, let's put the validation before building the head.

qlzh727 (changes requested):
Can you also add a unit test for this?

keras/applications/convnext.py
Frightera: Thanks, I thought I moved it yesterday. Also added a unit test for this case.

Frightera: As a second thought, we can also check the case where it is "linear", let me know if you want me to include this in unit test also.

keras/applications/applications_test.py
```diff
@@ -235,5 +242,11 @@ def _get_output_shape(model_fn):
     return model.output_shape
 
 
+def _get_last_layer_activation(model_fn):
```
qlzh727, line 245: This doesn't necessary to be a separate function, just make it part of test_application_classifier_activation

qlzh727 (changes requested):
Please also fix the format.

Frightera:
Should be fine now.

keras/applications/applications_test.py
```diff
@@ -191,6 +191,14 @@ def test_application_pooling(self, app, last_dim):
         )
         self.assertShapeEqual(output_shape, (None, last_dim))
 
+    @parameterized.parameters(MODEL_LIST)
+    def test_application_classifier_activation(self, app):
```
qlzh727, line 195: The unit tests are failing, since MODEL_LIST produce two items.

```
ERROR: test_application_classifier_activation53 (<function RegNetY008 at 0x7f760cb75310>, 768) (__main__.ApplicationsTest)
ApplicationsTest.test_application_classifier_activation53 (<function RegNetY008 at 0x7f760cb75310>, 768)
test_application_classifier_activation(<function RegNetY008 at 0x7f760cb75310>, 768)
----------------------------------------------------------------------
  File "/tmpfs/venv/lib/python3.9/site-packages/absl/testing/parameterized.py", line 320, in bound_param_test
    return test_method(self, *testcase_params)
TypeError: test_application_classifier_activation() takes 2 positional arguments but 3 were given

----------------------------------------------------------------------
Ran 11 tests in 62.691s

FAILED (errors=2)
```

qlzh727 (changes requested):

keras/applications/applications_test.py
Frightera: This has been rectified now by adding an extra/unused param `_`

Frightera:
Regnet also had the same issue, pushed a fix for that aswell.

Frightera:
@qlzh727 It looks like this is ready to pull, if there is any other thing please let me know. Thanks!

PR #17545, opened 2023-02-09
Activations clean up

joesho112358:
Noticed some things that might improve the readability in the `activations.py` file


qlzh727 (approved):
Thanks for the update.

Issue #16907, opened 2022-08-17
ConvNeXt not compatible with mixed precision


andreped:
**System information**.
- Have I written custom code (as opposed to using a stock example script provided in Keras): Yes
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Ubuntu 20.04
- TensorFlow installed from (source or binary): nightly
- TensorFlow version (use command below): 2.11.0a20220816
- Python version: 3.8.10
- GPU model and memory: NVIDIA Quadro RTX 6000 (24 GB VRAM)
- Do you want to contribute a PR? (yes/no): No

**Describe the problem**.
Starting from [this issue](https://github.com/keras-team/keras/issues/16656), I observed that ConvNeXt was not compatible with TimeDistributed, this was then fixed in the nightly release (see [here](https://github.com/keras-team/keras/issues/16656#issuecomment-1216889267)). As it was working I then tried to use mixed precision, where I got a new error. Note that MobileNetV3 works seemlessly with mixed precision. Hence, I think only ConvNeXt might be affected, but not sure.

I believe the model itself is working fine with mixed precision, but it contains the layer `LayerScale`, which may not be (see logs below for more details).

**Describe the expected behavior**.
Mixed precision should work seemlessly with ConvNeXt.

**Standalone code to reproduce the issue**.
It failed when initializing the ConvNeXt model, after mixed precision was enabled. Hence, I believe running this might reproduce the issue (note that source logs are not directly from this script, but I believe you will get the same error):
```
import tensorflow as tf
from tensorflow.keras.applications import ConvNeXtSmall

tf.keras.mixed_precision.set_global_policy('mixed_float16')
model = ConvNeXtSmall(include_top=False, weights="imagenet", pooling="none")
```

**Source code / logs**.
```
Traceback (most recent call last):
  File "source/main.py", line 454, in <module>
    main()
  File "source/main.py", line 216, in main
    model = get_classifier_architecture(MODEL_ARCH=ret.arch, ret=ret, instance_size=instance_size,
  File "/home/andrep/workspace/bcgrade/source/models/classifiers.py", line 371, in get_classifier_architecture
    shared_base_model = ConvNeXtSmall(include_top=False, weights="imagenet", pooling="none", input_shape=instance_size[1:])
  File "/home/andrep/workspace/bcgrade/venv/lib/python3.8/site-packages/keras/applications/convnext.py", line 610, in ConvNeXtSmall
    return ConvNeXt(
[...]

PR #17547, opened 2023-02-10
LayerScale fix to support pretrained ConvNeXt models with mixed precision

andreped:
Fix for the issue https://github.com/keras-team/keras/issues/16907. The proposed solution was proposed by @zibbini, and we both tested and verified that it works.

This PR fixes a bug in the [LayerScale](https://github.com/keras-team/keras/blob/master/keras/applications/convnext.py#L199) layer, which hindered ConvNeXt pretrained models to work with mixed precision. The solution was simply to set the appropriate dtype in the `gamma` variable of the LayerScale layer.

See [original gist](https://colab.research.google.com/gist/tilakrayal/30257452eebf9cc45e590f84ed8da4bd/untitled503.ipynb) to reproduce the bug.

See [new gist](https://colab.research.google.com/gist/andreped/5bc4393b64dbfcfed4c87e5c6863f3df/layerscale-mixed-precision-fix.ipynb) for verifying that the proposed fix works.

### DIFF
--- a/keras/applications/convnext.py
+++ b/keras/applications/convnext.py
@@ -26,6 +26,7 @@ import numpy as np
 import tensorflow.compat.v2 as tf
 
 from keras import backend
+from keras import initializers
 from keras import layers
 from keras import utils
 from keras.applications import imagenet_utils
@@ -217,9 +218,11 @@ class LayerScale(layers.Layer):
         self.projection_dim = projection_dim
 
     def build(self, input_shape):
-        self.gamma = tf.Variable(
-            self.init_values * tf.ones((self.projection_dim,)),
+        self.gamma = self.add_weight(
+            shape=(self.projection_dim,),
             dtype=self._compute_dtype_object,
+            initializer=initializers.Constant(self.init_values),
+            trainable=True,
         )
 
     def call(self, x):
--- a/keras/applications/convnext.py
+++ b/keras/applications/convnext.py
@@ -219,7 +219,7 @@ class LayerScale(layers.Layer):
     def build(self, input_shape):
         self.gamma = tf.Variable(
             self.init_values * tf.ones((self.projection_dim,)),
-            dtype=self._compute_dtype_object
+            dtype=self._compute_dtype_object,
         )
 
     def call(self, x):
--- a/keras/applications/convnext.py
+++ b/keras/applications/convnext.py
@@ -218,7 +218,8 @@ class LayerScale(layers.Layer):
 
     def build(self, input_shape):
         self.gamma = tf.Variable(
-            self.init_values * tf.ones((self.projection_dim,))
+            self.init_values * tf.ones((self.projection_dim,)),
+            dtype=self._compute_dtype_object
         )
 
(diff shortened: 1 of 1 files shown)


Frightera:
You'll have an issue with formatting as there is no `,` at the end:

You have:

```
tf.Variable(
            self.init_values * tf.ones((self.projection_dim,)),
            dtype=self._compute_dtype_object
        )
```

Should have been:

```
tf.Variable(
            self.init_values * tf.ones((self.projection_dim,)),
            dtype=self._compute_dtype_object,
        )
```

Please check the [gist here](https://colab.research.google.com/gist/Frightera/8147f5d4d22b60025b784801da65a47c/17547-lint.ipynb). Thanks!

andreped:
> You'll have an issue with formatting as there is no `,` at the end

Good catch! Did not think about that. Made the necessary change in new commit and now there are no issues with black linting.

keras/applications/convnext.py
```diff
@@ -218,7 +218,8 @@ def __init__(self, init_values, projection_dim, **kwargs):
 
     def build(self, input_shape):
         self.gamma = tf.Variable(
```
qlzh727, line 220: Can you replace this with self.add_weight(), which the more standard way to create weights for layer? (it has more logics for handling other aspects.)

qlzh727 (changes requested):
Thanks for the update.

keras/applications/convnext.py
andreped: Sure. Made a new commit with the suggested change.

keras/applications/convnext.py
```diff
@@ -217,8 +218,11 @@ def __init__(self, init_values, projection_dim, **kwargs):
         self.projection_dim = projection_dim
 
     def build(self, input_shape):
-        self.gamma = tf.Variable(
-            self.init_values * tf.ones((self.projection_dim,))
+        self.gamma = self.add_weight(
+            shape=(self.projection_dim,),
+            dtype=self._compute_dtype_object,
```
qlzh727, line 223: The dtype is not needed since it will be inferred from the initializer.

qlzh727 (approved):

keras/applications/convnext.py
andreped: Ok, no worries. Made a new commit removing the dtype.

qlzh727 (approved):

andreped:
@qlzh727 Seems like the batchnorm float16 unit test failed for the Ubuntu-GPU workflow. Do not see why this PR should be the cause.

Does not look like an assertion error, but rather a CUDA launch issue. I assume you have seen this before. Is this a random error during launch, or is it maybe due to a memory leak when running different tests sequentially on the same GPU?

Does this impact this PR?

qlzh727:
I think it is probably an test infra flaky when running the GPU test. We can ignore it here.

fchollet (approved):
Yep, the `tf.ones` call will not follow the Keras defaults for MP. Your fix looks good, thanks!

PR #17550, opened 2023-02-10
Perform all ops in apply_gradient in a given tf.name_scope.

foxik:
Hi,

today when looking in Tensorboard, I found out that the new optimizers do not create operations in a `tf.name_scope` of the `name` given to the `Optimizer` constructor.

That was the case previously -- see https://github.com/keras-team/keras/blob/bdbca4fb823b65f4dc5a9bb2acc0cf55e1276303/keras/optimizers/legacy/optimizer_v2.py#L701-L754

while the new `Optimizer` runs only `build` in the name scope, not the update operations, see https://github.com/keras-team/keras/blob/bdbca4fb823b65f4dc5a9bb2acc0cf55e1276303/keras/optimizers/optimizer.py#L633-L656.

Therefore, in this pull request, I moved the update operations in `apply_gradient` also inside the `tf.name_scope`.

The tests seem to run fine, code is correctly reformatted, and I verified in Tensorboard that after the change the update operations are indeed in a name scope.

### DIFF
--- a/keras/optimizers/optimizer.py
+++ b/keras/optimizers/optimizer.py
@@ -635,25 +635,27 @@ class _BaseOptimizer(tf.__internal__.tracking.AutoTrackable):
                 # Lift variable creation to init scope to avoid environment
                 # issues.
                 self.build(trainable_variables)
-        grads_and_vars = list(zip(grads, trainable_variables))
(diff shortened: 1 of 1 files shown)


qlzh727 (approved):
Thanks for the fix. This must be a oversight when we create the new optimizer.

PR #17568, opened 2023-02-15
Fix spelling error in Initializer warning

edknv:
This PR fixes a minor spelling error `initializer` -> `initializer` in the warning message. Spelling errors could possibly block CI pipelines e.g., if it has a spell check set up to check even warning messages that come from notebooks. We've run into this in https://github.com/NVIDIA-Merlin/models when we ran `codespell` on our notebooks which usually include all outputs.


qlzh727 (approved):
Thanks for the fix.

PR #17591, opened 2023-02-22
Less restrictive fallback logic

jayfurmanek:
A previous PR https://github.com/keras-team/keras/pull/17111 added some logic to use fallback implementations of GRU and LSTM on ROCm in situations where padded i/o is needed (since ROCm does not support padded i/o).

That logic turns out to be too restrictive - it chooses the fallback path in cases where it is not really needed, which may result in significant performance degradations.

This is a cherry-pick of https://github.com/keras-team/keras/pull/17587


fchollet (approved):
LGTM

CC @qlzh727 -- this would be a 2.12 cherrypick.

PR #17597, opened 2023-02-24
Unique state names for stateful RNN layers

### LINKED ISSUE #17477: Keras generates non-unique names for stateful LSTM's states
**System information**.
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): CentOS
- TensorFlow installed from (source or binary): PYPI
- TensorFlow version (use command below): 2.11.0
- Python version: 3.7

**Describe the problem**.
A stateful LSTM has 2 state values, both with the identical name. As you can see in the reproducing example, we have two different variables, but both have the identical name.

LSTM states have the identical name.

LSTM states should have unique names.

I think the problem comes from here:
https://github.com/keras-team/keras/blob/master/keras/layers/rnn/base_rnn.py#L909

1. I would do something like `lambda x: backend.variable(x, name='state')`. This will change the name to `lstm/state:0` which is already more readable.
2. To me it seems it's not possible to give `tf.Variable` a unique id, so I assume the naming would need to be done manually with something like `lambda x: backend.varaible(x, name=f'state_{unique_id}'`.

```python
import tensorflow as tf
import tensorflow.keras as keras
import numpy as np

inp = keras.Input(batch_shape=[1, 2, 3])
rnn = keras.layers.LSTM(stateful=True, units=3)
model = keras.Model(inp, rnn(inp))

print('Before:')
print(*rnn.states, sep='\n')
print()

data = np.random.rand(1, 2, 3).astype(np.float32)
model(data)

print('After:')
print(*rnn.states, sep='\n')
```

Output:
```
Before:
<tf.Variable 'lstm/Variable:0' shape=(1, 3) dtype=float32, numpy=array([[0., 0., 0.]], dtype=float32)>
<tf.Variable 'lstm/Variable:0' shape=(1, 3) dtype=float32, numpy=array([[0., 0., 0.]], dtype=float32)>

After:
<tf.Variable 'lstm/Variable:0' shape=(1, 3) dtype=float32, numpy=array([[ 0.11526701, -0.02300704, -0.1668661 ]], dtype=float32)>
<tf.Variable 'lstm/Variable:0' shape=(1, 3) dtype=float32, numpy=array([[ 0.17723909, -0.06333476, -0.31777197]], dtype=float32)>
```

mergian:
Fixes #17477

It generates `{layer_name}/state` for GRU and SimpleRNN layers, and `{layer_name}/state_{idx}` for LSTM layers.

### DIFF
--- a/keras/layers/rnn/base_rnn.py
+++ b/keras/layers/rnn/base_rnn.py
@@ -905,8 +905,22 @@ class RNN(base_layer.Layer):
                         self.variable_dtype or backend.floatx(),
                     )
                 )
+            if len(flat_init_state_values) == 1:
+
+                def assign_unique_var(x):
+                    return backend.variable(x, name="state")
+
+            else:
+                unique_id = 0
+
+                def assign_unique_var(x):
+                    nonlocal unique_id
+                    var = backend.variable(x, name=f"state_{unique_id}")
+                    unique_id += 1
+                    return var
+
             flat_states_variables = tf.nest.map_structure(
-                backend.variable, flat_init_state_values
+                assign_unique_var, flat_init_state_values
             )
             self.states = tf.nest.pack_sequence_as(
                 self.cell.state_size, flat_states_variables
--- a/keras/layers/rnn/gru_test.py
+++ b/keras/layers/rnn/gru_test.py
@@ -993,6 +993,21 @@ class GRULayerTest(test_combinations.TestCase):
         clone_names = [x.name for x in clone.weights]
         self.assertEqual(model_names, clone_names)
 
+    def test_unique_state_names(self):
+        inp = keras.Input(batch_shape=[3, None, 3])
+        a = keras.layers.GRU(units=3, stateful=True)
+        b = keras.layers.GRU(units=3)  # no states
+        c = keras.layers.GRU(units=3, stateful=True)
+        _ = keras.Model(inp, [a(inp), b(inp), c(inp)])
+
+        names = set()
+
+        for l in [a, b, c]:
+            for s in l.states:
+                if s is not None:
+                    self.assertNotIn(s.name, names)
+                    names.add(s.name)
+
 
 @test_combinations.generate(test_combinations.combine(mode=["graph", "eager"]))
 class GRULayerGenericTest(tf.test.TestCase):
--- a/keras/layers/rnn/lstm_test.py
+++ b/keras/layers/rnn/lstm_test.py
@@ -1424,6 +1424,21 @@ class LSTMLayerTest(test_combinations.TestCase):
         clone_names = [x.name for x in clone.weights]
         self.assertEqual(model_names, clone_names)
 
+    def test_unique_state_names(self):
+        inp = keras.Input(batch_shape=[3, None, 3])
+        a = keras.layers.LSTM(units=3, stateful=True)
+        b = keras.layers.LSTM(units=3)  # no states
+        c = keras.layers.LSTM(units=3, stateful=True)
+        _ = keras.Model(inp, [a(inp), b(inp), c(inp)])
+
+        names = set()
+
+        for l in [a, b, c]:
+            for s in l.states:
+                if s is not None:
+                    self.assertNotIn(s.name, names)
+                    names.add(s.name)
+
 
 if __name__ == "__main__":
     tf.test.main()
--- a/keras/layers/rnn/simple_rnn_test.py
+++ b/keras/layers/rnn/simple_rnn_test.py
@@ -249,6 +249,21 @@ class SimpleRNNLayerTest(tf.test.TestCase, parameterized.TestCase):
         clone_names = [x.name for x in clone.weights]
         self.assertEqual(model_names, clone_names)
 
+    def test_unique_state_names(self):
+        inp = keras.Input(batch_shape=[3, None, 3])
+        a = keras.layers.SimpleRNN(units=3, stateful=True)
+        b = keras.layers.SimpleRNN(units=3)  # no states
+        c = keras.layers.SimpleRNN(units=3, stateful=True)
+        _ = keras.Model(inp, [a(inp), b(inp), c(inp)])
+
+        names = set()
+
+        for l in [a, b, c]:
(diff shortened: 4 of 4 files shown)


keras/layers/rnn/base_rnn.py
```diff
@@ -905,8 +905,22 @@ def reset_states(self, states=None):
...
+            if len(flat_init_state_values) == 1:
+
+                def assign_unique_var(x):
+                    return backend.variable(x, name="state")
+
+            else:
+                unique_id = 0
+
+                def assign_unique_var(x):
+                    nonlocal unique_id
```
qlzh727, line 917: I think calling them state_0/1 is also bit weird, since the states in LSTM actually carry meaningful information (hidden state and output). So if  you could like to properly name the variable, then let's do that instead.

Also if you have 2 stateful lstm (eg bi-directional), what will be the state variable names? (state_0/1 and state_0/1 or state_0/1/2/3)?

qlzh727 (changes requested):

keras/layers/rnn/base_rnn.py
```diff
@@ -905,8 +905,20 @@ def reset_states(self, states=None):
                         self.variable_dtype or backend.floatx(),
                     )
                 )
+            if len(flat_init_state_values) == 1:
+
+                def assign_unique_var(x):
+                    return backend.variable(x, name="state")
+
+            else:
+                unique_names = iter(["hidden_state", "output_state"])
```
qlzh727, line 914: Note that this is the base RNN layer, which could support all different type of RNN cell, eg (stack multiple RNN cell together or customized RNN cell with different states), so this hard code here is not correct.

qlzh727 (changes requested):

keras/layers/rnn/base_rnn.py
mergian: Valid point. I added a `_gen_state_names` into the LSTM cell. By default, we use `state`, `state_1`, `state_2`, ...

If a RNN cell provides `_gen_state_names`, it is used to generate the names.

mergian: As there is no base `RNNBaseCell` (at least LSTM is not inheriting from any) I implemented it using `getattr(cell, '_gen_state_names`, fallback_implementation)`

keras/layers/rnn/base_rnn.py
```diff
@@ -905,8 +905,23 @@ def reset_states(self, states=None):
                         self.variable_dtype or backend.floatx(),
                     )
                 )
+
+            def gen_state_names():
```
qlzh727, line 909: Since we are adding _gen_state_names() to the LSTM, how about put this as a private method on the base_rnn as _gen_state_names() base implementation?

keras/layers/rnn/gru_test.py
```diff
@@ -993,6 +993,21 @@ def test_cloned_weight_names(self):
...
+    def test_unique_state_names(self):
+        inp = keras.Input(batch_shape=[3, None, 3])
+        a = keras.layers.GRU(units=3, stateful=True)
+        b = keras.layers.GRU(units=3)  # no states
+        c = keras.layers.GRU(units=3, stateful=True)
+        _ = keras.Model(inp, [a(inp), b(inp), c(inp)])
+
+        names = set()
+
+        for l in [a, b, c]:
```
qlzh727, line 1005: For a unit test, let's put clear assertion in the body, and not have any looping/state merge.

How about:

self.assertEqual(a.states[0].name, '....')
self.assertEqual(b.states[0].name, '....')
....

Same for all other tests.

qlzh727 (changes requested):
Sorry for the long wait. I was occupied by something else last week.

Also, can we add a unit test with stateful layer that is wrapped with bi-directional wrapper?

mergian:
I added the `_gen_state_names` to the base `RNN` class.

keras/layers/rnn/base_rnn.py
```diff
@@ -625,6 +625,13 @@ def __call__(self, inputs, initial_state=None, constants=None, **kwargs):
                 kwargs["constants"] = constants
             return super().__call__(inputs, **kwargs)
 
+    def _gen_state_names(self):
```
qlzh727, line 628: Looking at the test output, in case we have a cell that has nested states, the state name will become:

state:0, state_1:0, state_2:0, ...

which is bit weird, how about just use the name state, and let the tf.Variable to increment the count by itself?

qlzh727 (changes requested):

keras/layers/rnn/base_rnn.py
mergian: I'm not sure how to achieve that. For testing I removed the `LSTM._gen_state_names`, and changed the `RNN._gen_state_names` to:

```python
def _gen_state_names(self):
   while True:
       yield "state"
```

And the testcase to:
```python
def test_unique_state_names(self):
        inp = keras.Input(batch_shape=[3, None, 3])
        a = keras.layers.LSTM(units=3, stateful=True)
        b = keras.layers.LSTM(units=3)  # no states
        c = keras.layers.LSTM(units=3, stateful=True)
        _ = keras.Model(inp, [a(inp), b(inp), c(inp)])

        self.assertEqual(a.states[0].name, f"{a.name}/state:0")
        self.assertEqual(a.states[1].name, f"{a.name}/state:1")
        self.assertEqual(b.states[0], None)
        self.assertEqual(b.states[1], None)
        self.assertEqual(c.states[0].name, f"{c.name}/state:0")
        self.assertEqual(c.states[1].name, f"{c.name}/state:1")
```

Which then fails with:
```
in test_unique_state_names
    self.assertEqual(a.states[1].name, f"{a.name}/state:1")
AssertionError:
- lstm/state:0
?            ^
+ lstm/state:1
?
```

To me it seems that `tf.Variable` doesn't increase the `:x` count automatically, but I also don't see any parameter to enforce this.

mergian: @qlzh727 : Any idea how to achieve this increase of the variable index?

sachinprasadhs:

PR #17638, opened 2023-03-07
Optimize mixed-precision finite check for sparse tensors.

nluehr:
Mixed precision can be very slow and memory inefficient with sparse tensors, common, for example, in recommender-type models.

The largest inefficiencies come from checking sparse tensors for NaN values when dynamic loss scaling is enabled.

This PR implements a simple optimization to use the dense values of IndexedSlices directly during NaN checks.

### DIFF
--- a/keras/mixed_precision/loss_scale_optimizer.py
+++ b/keras/mixed_precision/loss_scale_optimizer.py
@@ -49,12 +49,14 @@ class _UnwrapPreventer:
 def _is_all_finite(grads):
     """Returns a scalar boolean tensor indicating if all gradients are
     finite."""
+
     def raw_values(g):
         return g.values if isinstance(g, tf.IndexedSlices) else g
 
     is_finite_per_grad = [
         tf.reduce_all(tf.math.is_finite(raw_values(g)))
-        for g in grads if g is not None
+        for g in grads
+        if g is not None
     ]
     return tf.reduce_all(is_finite_per_grad)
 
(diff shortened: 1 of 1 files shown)


haifeng-jin (changes requested):
Thanks for the PR! Would you please run `sh shell/format.sh` to format your files? Thanks!

qlzh727:
Adding Reed for the visibility.

qlzh727:
Please fix the lint error.

reedwm (approved):

gbaned:
Hi @nluehr Can you please fix Lint failures? Thank you!

gbaned:
Hi @nluehr Can you please fix Lint failures? Thank you!

keras/mixed_precision/loss_scale_optimizer.py
```diff
@@ -307,7 +313,7 @@ def update_if_not_finite_grads():
 
 
 # See LossScaleOptimizer docstring for why this is so big
-_DEFAULT_INITIAL_SCALE = 2**15
+_DEFAULT_INITIAL_SCALE = 2 ** 15
```
reedwm, line 316: The linter doesn't like this change. It wants no spaces around the `**`.

reedwm (changes requested):

keras/mixed_precision/loss_scale_optimizer.py
nluehr: Thanks @reedwm . Running shell/format.sh on my machine actually added those spaces in the first place. It seems the linter scripts are a bit finicky about the environment in which they run.

reedwm (approved):

PR #17667, opened 2023-03-13
Correct the docstring of Head func in convnext.py

Frightera:
Small fixes on `def Head()` function's docstring as it was a copy from `RegNet`.

### DIFF
--- a/keras/applications/convnext.py
+++ b/keras/applications/convnext.py
@@ -328,7 +328,7 @@ def PreStem(name=None):
 
 
 def Head(num_classes=1000, classifier_activation=None, name=None):
-    """Implementation of classification head of RegNet.
+    """Implementation of classification head of ConvNeXt.
 
     Args:
       num_classes: number of classes for Dense layer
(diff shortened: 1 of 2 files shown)


keras/applications/convnext.py
```diff
@@ -328,15 +328,15 @@ def apply(x):
...
+    """Implementation of classification head of ConvNeXt.
 
     Args:
       num_classes: number of classes for Dense layer
       classifier_activation: activation function for the Dense layer
       name: name prefix
 
     Returns:
-      Classification head function.
+      Tensor of logits or softmax values as the output.
```
qlzh727, line 339: I think the existing return string is still correct, since it only returns a function that produce a logits tensor, but not the logits tensor itself.

qlzh727 (changes requested):

keras/applications/convnext.py
Frightera: Hmm I see, maybe I can also edit `regnet.py` to make them consistent, right now regnet head function has this:
https://github.com/keras-team/keras/blob/0ce152e06a8b467a4071ff3d1324406374f007e9/keras/applications/regnet.py#L837-L844

What do you think? @qlzh727

Frightera: I think `Classification head function` would make more sense.

qlzh727 (approved):

PR #17743, opened 2023-03-29
Fix loss object name inference

srasmussenvl:
A callable loss object `obj` with `obj.name == None` would trigger a misleading exception, telling that the loss object is not callable.


gbaned:
Hi @srasmussenvl Can you please sign CLA. Thank you!

qlzh727 (changes requested):
Thanks for the change. Could u also add a unit test for this?

haifeng-jin:
Pending on the author's change and signing the CLA.

PR #17651, opened 2023-03-09
Add CategoricalFocalCrossentropy to Losses API

Frightera:
Implements the `CategoricalFocalCrossentropy()` loss based on the paper [Focal Loss for Dense Object Detection](https://arxiv.org/pdf/1708.02002.pdf) (Lin et al., 2018).

Feature request was made in #17583.

PR #17746, opened 2023-03-30
Minor improvements and code refactoring in backend.py

Frightera:
Small changes in backend.py, some of were discussed in the PR #17651

### DIFF
--- a/keras/backend.py
+++ b/keras/backend.py
@@ -5857,7 +5857,8 @@ def binary_focal_crossentropy(
         gamma: A focusing parameter, default is `2.0` as mentioned in the
             reference.
         from_logits: Whether `output` is expected to be a logits tensor. By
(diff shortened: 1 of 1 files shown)


keras/backend.py
```diff
@@ -5647,7 +5651,7 @@ def categorical_focal_crossentropy(
     )
 
     if from_logits:
-        output = tf.nn.softmax(output, axis=axis)
+        output = softmax(output, axis=axis)
```
qlzh727, line 5654: I think redirect to an extra level of internal function is unnecessary. Please keep the existing code.

qlzh727 (changes requested):

qlzh727 (approved):

haifeng-jin:
The copybara created CL needs to be synced before submission. What do we do in this case? @qlzh727

PR #17748, opened 2023-03-31
Move `Defaults to` to end of arg docstring and standardise values

SamuelMarks:
I've written a Python compiler that can do a bunch of thing. Relevant to Keras I am turning TensorFlow into SQL tables by way of SQLalchemy:
```sh
# TensorFlow 2.12.0
$ python3 -m cdd exmod --module 'tensorflow.keras.optimizers' \
                       --emit 'sqlalchemy_table' \
                       -o '/tmp/tf_optimizers'
```

After deduplicating and running `black` here are the contents of `sgd.py`:
```py
"""
Generated from keras.optimizers.sgd.
"""

SGD = Table(
    "SGD",
    metadata,
    Column(
        "learning_rate",
        Float,
        comment="A `Tensor`, floating point value, or a schedule that is a `tf.keras.optimizers.schedules.LearningRateSchedule`, or a callable that takes no arguments and returns the actual value to use. The learning rate",
        default=0.001,
        nullable=False,
    ),
    Column(
        "momentum",
        String,
        comment="float hyperparameter >= 0 that accelerates gradient descent in the relevant direction and dampens oscillations.e., vanilla gradient descent",
        default="0, i",
        nullable=False,
    ),
    Column(
        "nesterov",
        Boolean,
        comment="boolean. Whether to apply Nesterov momentum",
        default=False,
        nullable=False,
    ),
    Column("clipnorm", default=None, nullable=True),
    Column("clipvalue", default=None, nullable=True),
    Column("use_ema", Boolean, default=False, nullable=False),
    Column("ema_overwrite_frequency", default=None, nullable=True),
    Column("weight_decay", default=None, nullable=True),
    Column("global_clipnorm", default=None, nullable=True),
    Column("ema_momentum", Float, default=0.99, nullable=False),
    Column("jit_compile", Boolean, default=True, nullable=False),
    Column("name", String, default="SGD", nullable=False),
    Column("id", Integer, primary_key=True, server_default=Identity()),
)

__all__ = ["SGD"]
```

As you can see, `momentum` has an incorrect value because of how you've written your docstrings:
https://github.com/keras-team/keras/blob/0f8e81f/keras/optimizers/sgd.py#L62

Some commentary on the `Defaults to` syntax as understood by Sphinx:
- https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format
- https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

---

[...]

PR #17758, opened 2023-04-04
[keras/applications/efficientnet.py]  Standardise docstring usage of "Default to"

SamuelMarks:

### DIFF
--- a/keras/applications/convnext.py
+++ b/keras/applications/convnext.py
@@ -124,7 +124,7 @@ BASE_DOCSTRING = """Instantiates the {name} architecture.
 
   Args:
     include_top: Whether to include the fully-connected
-      layer at the top of the network. Defaults to True.
+      layer at the top of the network. Defaults to `True`.
     weights: One of `None` (random initialization),
       `"imagenet"` (pre-training on ImageNet-1k), or the path to the weights
       file to be loaded. Defaults to `"imagenet"`.
@@ -135,7 +135,7 @@ BASE_DOCSTRING = """Instantiates the {name} architecture.
       if `include_top` is False.
       It should have exactly 3 inputs channels.
     pooling: Optional pooling mode for feature extraction
-      when `include_top` is `False`. Defaults to None.
+      when `include_top` is `False`.
       - `None` means that the output of the model will be
         the 4D tensor output of the last convolutional layer.
       - `avg` means that global average pooling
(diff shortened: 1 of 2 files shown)


keras/applications/convnext.py
```diff
@@ -754,10 +754,10 @@ def preprocess_input(x, data_format=None):
 
     Args:
       x: A floating point `numpy.array` or a `tf.Tensor`.
-      data_format: Optional data format of the image tensor/array. Defaults to
-        None, in which case the global setting
-        `tf.keras.backend.image_data_format()` is used (unless you changed it,
-        it defaults to "channels_last").{mode}
+      data_format: Optional data format of the image tensor/array. `None` means
+        the global setting `tf.keras.backend.image_data_format()` is used
+        (unless you changed it, it uses "channels_last").{mode}.
```
qlzh727, line 759: I think the {mode} here is a place holder, which need to be appear at the end of the line.

keras/applications/efficientnet.py
```diff
@@ -852,10 +852,10 @@ def preprocess_input(x, data_format=None):
 
     Args:
       x: A floating point `numpy.array` or a `tf.Tensor`.
-      data_format: Optional data format of the image tensor/array. Defaults to
-        None, in which case the global setting
-        `tf.keras.backend.image_data_format()` is used (unless you changed it,
-        it defaults to "channels_last").{mode}
+      data_format: Optional data format of the image tensor/array. `None` means
+        the global setting `tf.keras.backend.image_data_format()` is used
+        (unless you changed it, it uses "channels_last").{mode}.
```
qlzh727, line 857: Same here.

qlzh727 (changes requested):

keras/applications/efficientnet.py
SamuelMarks: @qlzh727 The issue is that `Defaults to` also needs to appear at the end of the string. Actually are you sure `{mode}` is only replaced when it appears at the end of the string? - Where does this happen? - Happy to modify that interpolation code.

qlzh727: Well. I think they are a typo here then. We can remove the {mode}.

https://www.tensorflow.org/api_docs/python/tf/keras/applications/convnext/preprocess_input shows that the {mode} is not replaced properly.

It was replaced in some other application models, eg https://github.com/keras-team/keras/blob/901950201d867c85ec34f4d0c9201aea2c15a65d/keras/applications/imagenet_utils.py#L126

haifeng-jin:
Pending on author's change.

SamuelMarks:
@haifeng-jin Fixed

keras/applications/convnext.py
qlzh727: I think the {mode} here can be removed.

haifeng-jin (commented):
LGTM. No more comments from me.

PR #17782, opened 2023-04-04
[keras/estimator/__init__.py] Standardise docstring usage of "Default to"

SamuelMarks:

### DIFF
--- a/keras/estimator/__init__.py
+++ b/keras/estimator/__init__.py
@@ -120,12 +120,11 @@ def model_to_estimator(
       checkpoint_format: Sets the format of the checkpoint saved by the
         estimator when training. May be `saver` or `checkpoint`, depending on
         whether to save checkpoints from `tf.train.Saver` or
-        `tf.train.Checkpoint`. Before 2.0 the argument is `saver`. When
-        2.0 is released, this will be `checkpoint`. Estimators use
-        name-based `tf.train.Saver` checkpoints, while Keras models use
-        object-based checkpoints from `tf.train.Checkpoint`. Currently, saving
-        object-based checkpoints from `model_to_estimator` is only supported by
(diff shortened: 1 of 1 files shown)


haifeng-jin (commented):
I am not familiar with this part. There are 2 questions to be answered.
1. Is 2.0 referring TF 2.0? If so, it is already released and the description can be further changed.
2. The file is wrapped in `# LINT.IfChange`. We may need more changes to pass the test.

@qlzh727 may help answer them.

qlzh727:
> I am not familiar with this part. There are 2 questions to be answered.
> 
> 1. Is 2.0 referring TF 2.0? If so, it is already released and the description can be further changed.
yes, the 2.0 here refers to TF 2.0 release.

> 2. The file is wrapped in `# LINT.IfChange`. We may need more changes to pass the test.
The change here should be fine since it only touch the comment/docstring, and not the code.
> 
> @qlzh727 may help answer them.

haifeng-jin:
Thanks @qlzh727 for the answer!
Then we need to change the description of the saver completely.

@k-w-w may know what would be the most accurate description here.

k-w-w:
Looks like the default was never changed to `checkpoint`, and at this point probably won't be...

Lets remove the 2.0 part. If we want to roll this out we can always add it back.

haifeng-jin:
Pending on author's change.

SamuelMarks:
Sure thing, made the edit

SamuelMarks:
Sure thing, made the edit. Yeah I noticed that issue when I was going through the codebase and had planned a different issue on this point but happy to merge it all into this one PR.

haifeng-jin (approved):

haifeng-jin:
This is a duplicate for #17957.

PR #17787, opened 2023-04-04
[keras/layers/activation/relu.py] Standardise docstring usage of "Default to"

SamuelMarks:

### DIFF
--- a/keras/layers/activation/relu.py
+++ b/keras/layers/activation/relu.py
@@ -65,11 +65,11 @@ class ReLU(Layer):
       Same shape as the input.
 
     Args:
-      max_value: Float >= 0. Maximum activation value. None
-        means unlimited. Defaults to `None`.
+      max_value: Float >= 0. Maximum activation value. None means unlimited.
+        Defaults to `None`.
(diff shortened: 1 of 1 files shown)


keras/layers/activation/relu.py
```diff
@@ -65,9 +65,9 @@ class ReLU(Layer):
       Same shape as the input.
 
     Args:
-      max_value: Float >= 0. Maximum activation value. Default to None, which
-        means unlimited.
-      negative_slope: Float >= 0. Negative slope coefficient. Default to 0.
+      max_value: Float >= 0. Maximum activation value. None
+        means unlimited. Defaults to `None`.
```
qlzh727, line 69: Please move the "means unlimited" to the previous line.

qlzh727 (changes requested):

keras/layers/activation/relu.py
SamuelMarks: What should it be replaced with? `max_value` of `None` doesn't seem to make sense and needs a docstring to explain it. Is it `sys.float_info.max`? - Or `float("inf")`?

qlzh727 (approved):

PR #17794, opened 2023-04-04
[keras/layers/convolutional/conv2d.py] Standardise docstring usage of "Default to"

SamuelMarks:

### DIFF
--- a/keras/layers/convolutional/conv2d.py
+++ b/keras/layers/convolutional/conv2d.py
@@ -101,8 +101,8 @@ class Conv2D(Conv):
         `channels_first`.  The ordering of the dimensions in the inputs.
         `channels_last` corresponds to inputs with shape `(batch_size, height,
         width, channels)` while `channels_first` corresponds to inputs with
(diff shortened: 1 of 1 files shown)


keras/layers/convolutional/conv2d.py
```diff
@@ -101,11 +101,11 @@ class Conv2D(Conv):
         `channels_first`.  The ordering of the dimensions in the inputs.
         `channels_last` corresponds to inputs with shape `(batch_size, height,
         width, channels)` while `channels_first` corresponds to inputs with
-        shape `(batch_size, channels, height, width)`. It defaults to the
+        shape `(batch_size, channels, height, width)`. When unspecified, uses
```
haifeng-jin, line 104: Change "When unspecified, " to "If left unspecified, it "

haifeng-jin (changes requested):
Thanks for the PR!

keras/layers/convolutional/conv2d.py
SamuelMarks: Fixed

haifeng-jin (approved):

PR #17800, opened 2023-04-04
[keras/layers/convolutional/separable_conv2d.py] Standardise docstring usage of "Default to"

SamuelMarks:

### DIFF
--- a/keras/layers/convolutional/separable_conv2d.py
+++ b/keras/layers/convolutional/separable_conv2d.py
@@ -70,9 +70,10 @@ class SeparableConv2D(SeparableConv):
         `(batch_size, height, width, channels)` while `channels_first`
         corresponds to inputs with shape
         `(batch_size, channels, height, width)`.
-        It defaults to the `image_data_format` value found in your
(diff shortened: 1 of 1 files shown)


keras/layers/convolutional/separable_conv2d.py
```diff
@@ -70,9 +70,10 @@ class SeparableConv2D(SeparableConv):
         `(batch_size, height, width, channels)` while `channels_first`
         corresponds to inputs with shape
         `(batch_size, channels, height, width)`.
-        It defaults to the `image_data_format` value found in your
-        Keras config file at `~/.keras/keras.json`.
-        If you never set it, then it will be "channels_last".
+        When unspecified, uses
```
qlzh727, line 73: This line seems to be short, and the content from next line seems to be able to fit here.

qlzh727 (changes requested):

keras/layers/convolutional/separable_conv2d.py
SamuelMarks: Fixed

qlzh727 (approved):

PR #17980, opened 2023-04-13
[keras/benchmarks/benchmark_util.py] Use var rather than string literal for `is None` checks on `measure_performance`

SamuelMarks:
Found by running `pydocstyle` on entire codebase.

### DIFF
--- a/keras/benchmarks/benchmark_util.py
+++ b/keras/benchmarks/benchmark_util.py
@@ -142,13 +142,13 @@ def measure_performance(
       ValueError: If `x` is none or if `optimizer` is not provided or
       if `loss` is not provided or if `num_gpus` is negative.
     """
-    if "x" is None:
+    if x is None:
         raise ValueError("Input data is required.")
-    if "optimizer" is None:
+    elif optimizer is None:
         raise ValueError("Optimizer is required.")
-    if "loss" is None:
+    elif loss is None:
         raise ValueError("Loss function is required.")
-    if num_gpus < 0:
+    elif num_gpus < 0:
         raise ValueError("`num_gpus` cannot be negative")
 
     # TODO(xingyulong): we will add tfds support later and


keras/benchmarks/benchmark_util.py
```diff
@@ -142,13 +142,13 @@ def measure_performance(
       ValueError: If `x` is none or if `optimizer` is not provided or
       if `loss` is not provided or if `num_gpus` is negative.
     """
-    if "x" is None:
+    if x is None:
         raise ValueError("Input data is required.")
-    if "optimizer" is None:
+    elif optimizer is None:
```
Frightera, line 147: Hi @SamuelMarks,

Just curiosity, why did you switch into `elif` instead of `if`? I think keeping them independent might be a better practice, i.e maintain the `if` statements.

SamuelMarks: @Frightera Because it's more efficient:
```sh
$ python -m timeit 'a="foo"' 'if a == "foo":' '   a += "bar"' 'if a == "can":' '   a += "can"'
5000000 loops, best of 5: 51.9 nsec per loop
$ python -m timeit 'a="foo"' 'if a == "foo":' '   a += "bar"' 'elif a == "can":' '   a += "can"'
10000000 loops, best of 5: 39.6 nsec per loop
```

Also it's more logical. The compiler probably optimises the branches better, and it's clearer to read.

Frightera: Thanks, I think the speed difference is because of every `if` statement is executed in your example whereas in here there are `ValueError`s which will cause not to execute other statements, probably the performance difference will be negligible.

fchollet (approved):
LGTM, thanks

haifeng-jin (approved):
Thanks for the PR!

PR #18045, opened 2023-04-22
Update example losses to bce- metrics/confusion_metrics.py

Frightera:
I think it is better to use "binary_crossentropy" instead of "mse" as those metrics are for classification purposes.


rchao:
Gabriel, would you mind taking a look at this change? Thanks.

grasskin (approved):
Thank you for the PR @Frightera, I agree that the example documentation should be using a classification loss instead - they make more sense for confusion metrics.

grasskin:
Hi @Frightera, would you mind rebasing/updating your branch so we can pass CI checks?

Frightera:
Hi @grasskin, should be fine now.

grasskin (approved):

PR #18050, opened 2023-04-24
Removing unnecessary conversion to tf.constant in documentation

Giovanni-Alzetta:
In the documentation of some functions in losses.py there are some examples where a list is converted to a tf constant and then converted to a numpy array. Wouldn't it make more sense to convert it directly to a numpy array?

---

Note: this is my first PR here, sorry for not opening an issue but this really is a minor, cosmetic change.

I tried to see if the tf.constant follows some "smart" rule for conversion (which would justify the extra step), but I couldn't find anything, and I made a quick search and this is the only place in the whole codebase where this double conversion is made. So I really couldn't find any reason to justify the double conversion...

### DIFF
--- a/keras/losses.py
+++ b/keras/losses.py
@@ -858,7 +858,7 @@ class CategoricalCrossentropy(LossFunctionWrapper):
     1.177
 
     >>> # Calling with 'sample_weight'.
-    >>> cce(y_true, y_pred, sample_weight=tf.constant([0.3, 0.7])).numpy()
+    >>> cce(y_true, y_pred, sample_weight=np.array([0.3, 0.7]))
     0.814
 
     >>> # Using 'sum' reduction type.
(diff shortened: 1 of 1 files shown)


rchao (changes requested):
Thank you for the PR! The conversion to numpy would still be needed, because otherwise the output is left as a tf.Tensor. I don't think we need to make this change.

Giovanni-Alzetta:
> Thank you for the PR! The conversion to numpy would still be needed, because otherwise the output is left as a tf.Tensor. I don't think we need to make this change.

I'm not sure I understand your comment: the current code hardcodes a list, loads it into a tf.Tensor which is then converted to numpy array:
`tf.constant([0.3, 0.7])).numpy()`
My proposal is to load the list directly to a numpy array:
`np.array([0.3, 0.7]))`

I mean, there could be some philosophical reason to prefer the former, but I don't see it.

rchao:
Can you give the example code a run and see what the output is? I believe it would be a tf.Tensor instead of an int:

```
import tensorflow as tf
import numpy as np

y_true = [[0, 1, 0], [0, 0, 1]]
y_pred = [[0.05, 0.95, 0], [0.1, 0.8, 0.1]]
cce = tf.keras.losses.CategoricalCrossentropy()
cce(y_true, y_pred, sample_weight=np.array([0.3, 0.7]))
```

PR #18052, opened 2023-04-24
Add ignore nan metric loss

sagi-ezri:
Create classes for metrics and losses in the Keras TensorFlow project that can ignore NaN values in the labels by wrapping another metric or loss and computing it.


haifeng-jin:
Hi @sagi-ezri , Thanks for the PR!
Ror any API changes, we will need to go through a discussion before we can decide whether to accept the change.

Would you please create an issue to initiate the discussion?
Mainly just describe the issues you faced if the labels or predictions contains NaN.

Thanks

PR #18105, opened 2023-05-08
[Add Feature] - Throw an error if softmax is used with 1 neuron

Frightera:
This is a utility function to check if the usage of softmax makes sense (new users make this mistake a lot). Applying softmax on a single neuron will make the model output ones everytime, there are too many Stackoverflow posts about this. 

In order to see this in action, please check [the gist](https://colab.research.google.com/gist/Frightera/fdcec020fff6ee9521ae2fd3eaed774d/checksoftmaxlastlayer.ipynb).

This applies for any other layers (Conv2D etc.) where the applied axis (axis=-1 default) of softmax has only one unit.


rchao:
Thanks for the PR! @Frightera can you confirm that the newly added test would fail without the fix in the library changes?

@pradeepkuppla, would it be possible to give this a run across all internal tests before we proceed with an approval?

Frightera:
Hi @rchao I am not sure if I understood you correctly - do you want me to write a test to fail it on purpose?

sachinprasadhs:
@Frightera ,

It is failing internal test with below error, could you please make the necessary changes.

```
    "softmax" in str(output.name.lower())
AttributeError: 'NoneType' object has no attribute 'lower'
```

Frightera:
@sachinprasadhs Should be fine now, I re-run the tests manually.

That was an edge case I didn't think of 😮

rchao:
> Hi @rchao I am not sure if I understood you correctly - do you want me to write a test to fail it on purpose?

To make sure we properly cover the fix, the idea is to have a unit test that would have failed without your library changes. Can you confirm this is the case here?

Frightera:
@rchao To make it clear - This is not actually a fix, there was no error in Keras. That's just a warning to catch a logical error made by new users.

So you really can not catch this with a regular test as the model will be compiled successfully.

keras/engine/training.py
```diff
@@ -723,6 +720,13 @@ def compile(
               Defaults to `0`.
             **kwargs: Arguments supported for backwards compatibility only.
         """
+
+        validate_softmax_activation = kwargs.pop(
+            "experimental_validate_softmax_activation", True
```
rchao, line 725: Is there any reason why we won't want to always validate the softmax activation?

Frightera: Actually no, however I thought it is better to have this something optional. The users who don't apply softmax on the last axis may trigger this warning. Maybe we can reflect this in warning message, something like: `You can disable this check by setting ...`

So it is something tricky to decide, what do you think?

rchao: In general warnings can go unnoticed or ignored very easily; if we are confident based on the check that the user is making some mistake, I would simply error out. Are there known cases where the check fails but it's valid that the user would want to proceed?

Frightera: There should be no case(s) for users to proceed with that mistake. I'll extend this code to check the axis, and raising an error makes sense in that case. I'll update this by the end of the week.

Frightera: Hi @rchao, sorry for the delay.

Now I updated the logic, we make sure that the user is making a mistake in this case and this throws an error rather than showing a warning.

I think it is better to add @fchollet as a reviewer since this can touch the model API interface.

Frightera:
I'll check the test fails when I have time. cc/@sachinprasadhs

Frightera:
@sachinprasadhs I made the necessary changes and ran the CI pipeline in local enviroment, all seems fine. Can you trigger the tests once more?

rchao:
Thanks @Frightera!

@sachinprasadhs, is using the tag "pending internal tests" all is needed to notify a need of a internal test run, or there's something else we should do? Thanks!

sachinprasadhs:
> Thanks @Frightera!
> 
> @sachinprasadhs, is using the tag "pending internal tests" all is needed to notify a need of an internal test run, or there's something else we should do? Thanks!

This will help the gTech team who is handling the PRs to filter the PRs to test internally, I will look into it if we have missed anything. Thanks!

sachinprasadhs:
@Frightera , Could you please fix the below error.

```
ValueError: The model cannot be compiled because it has no loss to optimize.

During handling of the above exception, another exception occurred:

  File "/build/work/8730373dec43609e9e8498e8212791ddc67c/google3/runfiles/google3/third_party/py/keras/engine/training_test.py", [line 5110], in test_functional_model_output
    with self.assertRaisesRegex(
AssertionError: "has a single unit output, but the activation is softmax.*" does not match "The model cannot be compiled because it has no loss to optimize."
```

Frightera:
@sachinprasadhs That's interesting, this error was not present in my local CI. Let's try again.

sachinprasadhs:
@Frightera , Could you please check the below error.

```
  File "/build/work/45a231adaff5e6784bf9c607c668de34477f/google3/runfiles/google3/third_party/py/keras/engine/training_test.py", [line 5110](https://cs.corp.google.com/piper///depot/google3/third_party/py/keras/engine/training_test.py?l=5110&ws=tap-presubmit-server/67803804&snapshot=2), in test_functional_model_output
    with self.assertRaisesRegex(
AssertionError: ValueError not raised
```

Frightera:
Hi @sachinprasadhs,

I don't know why this is happening for the second time, I don't get any errors in Colab and my local environment when I run the CI cycle. 

Also the PR #18264 has now finished its build without any errors. Why is there a contradiction like this?

sachinprasadhs:
Now the previous error is not showing anymore. 

The internal test result for the above linked PR by copybara is what I'm providing. 

Here is the new result output.

```
File "/build/work/918d6c4f2fd469de06938bb711d72b99a84f/google3/runfiles/google3/third_party/py/keras/engine/training.py", [line 4523](https://cs.corp.google.com/piper///depot/google3/third_party/py/keras/engine/training.py?l=4523&ws=tap-presubmit-server/67837954&snapshot=2), in _check_output_activation_softmax
    raise ValueError(
ValueError: Output layer dense_2 has a single unit output, but the activation is softmax. This is most likely an error because softmax outputs sum to 1 therefore single unit outputs with softmax will only output 1.0. If you think that the error is raised due to an incorrect check, please file an issue on [https://github.com/keras-team/keras/issues](https://www.google.com/url?q=https://github.com/keras-team/keras/issues&sa=D). You can disable this check by setting `experimental_validate_softmax_activation=False` when calling `compile()` on the model.
```

Frightera:
@sachinprasadhs Thanks for the quick response.

Maybe I am missing something but PR opened by copybara `https://github.com/keras-team/keras/pull/18264` https://github.com/keras-team/keras/pull/18264 does not include any errors in the internal results.

image


Also the error you sent:
```
File "/build/work/918d6c4f2fd469de06938bb711d72b99a84f/google3/runfiles/google3/third_party/py/keras/engine/training.py", [line 4523](https://cs.corp.google.com/piper///depot/google3/third_party/py/keras/engine/training.py?l=4523&ws=tap-presubmit-server/67837954&snapshot=2), in _check_output_activation_softmax
    raise ValueError(
ValueError: Output layer dense_2 has a single unit output, but the activation is softmax. This is most likely an error because softmax outputs sum to 1 therefore single unit outputs with softmax will only output 1.0. If you think that the error is raised due to an incorrect check, please file an issue on [https://github.com/keras-team/keras/issues](https://www.google.com/url?q=https://github.com/keras-team/keras/issues&sa=D). You can disable this check by setting `experimental_validate_softmax_activation=False` when calling `compile()` on the model.
```

This part does not exist in my PR: 🤔 
```
please file an issue on [https://github.com/keras-team/keras/issues](https://www.google.com/url?q=https://github.com/keras-team/keras/issues&sa=D).
```

This is slightly different from what I have in my PR:
```python
f"Output layer {layer_name} has a single unit output, "
                    "but the activation is softmax. This is most likely an "
                    "error because softmax outputs sum to 1 therefore single "
                    "unit outputs with softmax will only output 1.0. If you "
                    "think that the error is raised due to an incorrect check, "
                    "please file an issue on "
                    "https://github.com/keras-team/keras/issues. You can "
                    "disable this check by setting "
                    "`experimental_validate_softmax_activation=False` when "
                    "calling `compile()` on the model."
                )
```

So the problem is I can't see the error in the internal tests, and the provided value error is slightly different from what I have in my PR.

sachinprasadhs:
The internal Tests will be run against different hardware configurations, different OS environments, this error seems to be happening in `//third_party/py/keras/integration_test:dragonfish_preprocessing_applied_in_dataset_creator_test_tpu`

Frightera:
So after tracing this I found:

https://github.com/keras-team/keras/blob/4829ddfc5fdcb84efe19b5e571f4be8dda89a3e1/keras/integration_test/preprocessing_test_utils.py#L88-L113

In line 112, we can see: `outputs = tf.keras.layers.Dense(1, activation="softmax")(x)`. So this was triggering my check in the PR and was raising an error. Obviously this is a wrong model setup in the internal test, I can fix the setup in this PR or can open a seperate PR for it. 

Which action should we take in this case?

sachinprasadhs:
@rchao , Can you please suggest in the above comment.

Frightera:
Hi @rchao, any update on this? A different internal test needs to be changed.

Frightera:
Hi, any update on this PR by any chance?

sachinprasadhs:
@qlzh727 , Could you please take a look into this PR. Thanks!

sachinprasadhs:

PR #18113, opened 2023-05-11
[keras/layers/preprocessing] fix comments in RandomWidth, change to 'horizontally' instead of 'vertically'

qibolee:
[keras/layers/preprocessing] fix comments in RandomWidth, change to 'horizontally' instead of 'vertically'

Change-Id: Ib9b3093391ef6bda66b347c4dbb34a9d1fb42cf8

PR #18114, opened 2023-05-11
[keras/layers/preprocessing] fix comments in RandomWidth, change to 'horizontally' instead of 'vertically'

qibolee:
[keras/layers/preprocessing] fix comments in RandomWidth, change to 'horizontally' instead of 'vertically'

Change-Id: Ib9b3093391ef6bda66b347c4dbb34a9d1fb42cf8


ianstenbit:
This seems to be a duplicate of #18113 so I'll close this one and review that one.

PR #18123, opened 2023-05-15
Fix compute_output_signature dtype estimation for model with multiple outputs

shkarupa-alex:
Issue https://github.com/keras-team/keras/issues/17420 was partly fixed with https://github.com/keras-team/keras/pull/17703 but only for models with single output.

This PR fixes it completely.


rchao:
Thanks for the PR! Would it be possible to cover the behavior with a test that would have failed without this fix?

keras/engine/base_layer_test.py
```diff
@@ -1116,6 +1117,33 @@ def __init__(self, var1, var2, var3=None, **kwargs):
         with self.assertRaises(NotImplementedError):
             config = layer.get_config()
 
+    def test_model_compute_output_signature_dtype_single_output(self):
+        with policy_scope("mixed_float16"):
```
qlzh727, line 1121: Is there anything to do with mixed_precision here? or it can be removed here in this test.

qlzh727 (approved):

qlzh727 (approved):

sachinprasadhs:

PR #18144, opened 2023-05-20
Fix CategoricalFocalCE documentation

Frightera:
Looks like we need some spaces/newlines in the docstring, documentation looks like this:

image


haifeng-jin (approved):
Thanks for the PR!

Issue #18152, opened 2023-05-24
Use tf.math.divide_no_nan here


ehtshamelahiwm:
https://github.com/keras-team/keras/blob/f9336cc5114b4a9429a242deb264b707379646b7/keras/layers/pooling/global_average_pooling1d.py#L93

tilakrayal:
@ehtshamelahiwm,
Could you please provide context/use-case for the issue, why it is needed to change/update the **tf.reduce_sum** to **tf.math.divide_no_nan.**

`tf.math.divide_no_nan` computes a safe divide which returns 0 if y (denominator) is zero whereas `tf.reduce_sum` tries to Compute the sum of elements across dimensions of a tensor. Both are absolutely different in computation.

Thank you!

ehtshamelahiwm:
I meant that the division which is taking place in that step should be a safe division. If the `tf.reduce_sum(mask, axis=steps_axis, keepdims=self.keepdims) ` is zero (which can happen), the global average pooling becomes nan

tilakrayal:
@ehtshamelahiwm,
Could you please provide the test case for the above requested change. So that it will be easy to analyse the modification mentioned. Also using  tf.math.divide_no_nan doesn't have an impact in the code and tf.reduce_sum provides the output which was expected. Thank you!

PR #18159, opened 2023-05-26
PR #18152: Changed regular divide to tf.math.divide_no_nan - global_average_pooling1d.py

Daanish1:
Replaced the divide with the safer: tf.math.divide_no_nan, as denominator has the potential to be 0.


qlzh727 (changes requested):
Thanks for the contribution. Can u add a unit test for this change? eg a input is fully masked?

gbaned:
Hi @Daanish1 Can you please check @qlzh727's comments and keep us posted ? Thank you!

PR #18234, opened 2023-06-15
Update example losses @ probabilistic_metrics.py

Frightera:
It's better to demonstrate those metrics using classification losses in the docs.


fchollet (approved):
LGTM, thanks

PR #18251, opened 2023-06-23
Fixed "reset_state" of R2Score metric

pdyakov:
"reset_state" method of R2Score metric throws an error:
`ValueError: Tensor conversion requested dtype int32 for Tensor with dtype float32: <tf.Tensor: shape=(), dtype=float32, numpy=0.0>`
Quick fix is to specify original "dtype" of variable when assigning it to zeros.


fchollet (approved):
LGTM, thanks

PR #18258, opened 2023-06-26
Cherrypick Sequential serialization bug fix for r2.13

nkovela1:
This PR is a fix based on the serialization bug reported by @seanpmorgan.

It fixes the serialization format of Sequential models saved with HDF5 and is important for ensuring backwards compatibility with the serialization format switch in TF 2.13.

### DIFF
--- a/keras/engine/sequential.py
+++ b/keras/engine/sequential.py
@@ -26,6 +26,7 @@ from keras.engine import input_layer
 from keras.engine import training
 from keras.engine import training_utils
 from keras.saving import serialization_lib
+from keras.saving.legacy import serialization as legacy_serialization
 from keras.saving.legacy.saved_model import model_serialization
 from keras.utils import generic_utils
 from keras.utils import layer_utils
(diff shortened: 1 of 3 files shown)


qlzh727:
FYI, original change in https://github.com/keras-team/keras/commit/a78f714973a83fa3502962624c49a6c9f714b31a.

qlzh727 (approved):

qlzh727:
Will let the test run again for the release branch before merge. Thanks for the quick fix.

seanpmorgan:
Will fix https://github.com/tensorflow/tensorflow/issues/61085

qlzh727:
@nkovela1, seems that the presubmit test failed, but it was passing on head. Could it be any change that is between the current head and the release branch also need to be cherrypick into the release branch?

To reproduce the test failure, you can checkout the rc branch and this pr and run the blaze test locally.

qlzh727:
Tested locally with Neel for this PR. The test error is related caused by a inconsistent version of TF used in the config. We suppose to test the RC build against the tf 2.13 RC version, but the test was using the latest tf-nightly. The delta between them was causing these 4 tests to fail. 

We have verified those 4 tests locally, and they are passing with tf.2.13 rc2. Merging this PR now.

PR #18274, opened 2023-07-03
refactor:  _log_epoch_metrics()

arjun-234:
In this refactored version, I have changed the code to consolidate two loops into one loop for the same method, improving efficiency and readability.


rchao:
Thanks for the PR -- now running tests to check this doesn't break anything.

PR #18336, opened 2023-08-10
Validate alpha argument leaky_relu.py

SuryanarayanaY:
The value of alpha should be a float >= `0`. But if given negative value still there is no exception raised and providing output which is not correct. Hence added validation code for same.

Also if alpha is given a `int` value then it is converted into `float` internally.But documentation states alpha should be a Float >= 0 . Hence also adding a note that if given int values then it will be converted into float internally.


gbaned:
Hi @mattdangerw Can you please review this PR ? Thank you!

gbaned:
Hi @mattdangerw Can you please review this PR ? Thank you!

sachinprasadhs:

PR #18394, opened 2023-09-22
Changes to add mixed bfloat16 policy support for RandomFourierFeatures

jojivk73:
These changes adds mixed bfloat16 precision to be enabled for RandomFourierFeatures layer. A few models in our suite, use this layer and mixed bfloat16 policy support improves performance and reduces the number of cast ops in the model by default.


sachinprasadhs:

PR #18476, opened 2023-09-22
Adding erf function

sqali:
Hi, 

This pull request ([clone of PR](https://github.com/keras-team/keras-core/pull/908#event-10450603098) ) is raised to add the ERF function implementation along with support for various backends.

Thanks & Regards

### DIFF
--- a/keras/ops/math_test.py
+++ b/keras/ops/math_test.py
@@ -841,9 +841,9 @@ class MathOpsCorrectnessTest(testing.TestCase, parameterized.TestCase):
         sample_values = np.array([-3.0, -2.0, -1.0, 0.0, 1.0, 2.0, 3.0])
 
         # Expected output using numpy's approximation of the error function
-        expected_output = (2 / np.sqrt(np.pi)) * np.vectorize(math.erf)(
-            sample_values
-        )
+        expected_output = (2 / np.sqrt(np.pi)) * np.vectorize(
+            scipy.special.erf
+        )(sample_values)
 
         # Output from the erf operation in keras_core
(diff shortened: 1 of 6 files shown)


keras/ops/math_test.py
```diff
@@ -835,3 +835,41 @@ def test_rsqrt(self):
...
+        self.assertAllClose(expected_output, output_from_erf_op, atol=1e-4)
+
+    def test_erf_operation_dtype(self):
+        # Test for float32 and float64 data types
+        for dtype in ("float32", "float64"):
+            sample_values = np.array(
+                [-3.0, -2.0, -1.0, 0.0, 1.0, 2.0, 3.0], dtype=dtype
+            )
+            expected_output = (2 / np.sqrt(np.pi)) * np.vectorize(math.erf)(
+                sample_values
```
fchollet, line 861: You can just use `scipy.special.erf(x)` as the reference.

fchollet (commented):
I ran some checks, and while there are minor differences between backends (torch is especially divergent, as is often the case), they should stay below 1e-5. Just use `scipy.special.erf(x)` as the reference.

keras/ops/math_test.py
sqali: Implemented it but still the checks are failing. Kindly let me know if you want me to do some extra checks, so I can rigorously test and find the reasons behind these failures. The difference seems to be the same as in the previous repository (keras-core) errors.

image

fchollet (approved):
Thanks for the contribution! I'll just take it over.

PR #18477, opened 2023-09-22
Increase-test-coverage for `keras_tensor` + Fix `__invert__` + Fix `FloorDivide`

Faisal-Alsrheed:
1. keras/backend/common/keras_tensor.py:

- The method `__floordiv__` now uses `ops.FloorDivide().symbolic_call()` instead of `ops.FloorDiv().symbolic_call()`
- The method `__rfloordiv__ `now uses `ops.FloorDivide().symbolic_call()` instead of `ops.FloorDiv().symbolic_call()`
- Fixed the `__invert__` method to not take the `other` parameter.

3. keras/backend/common/keras_tensor_test.py:
Added 21 new tests covering various binary operations.


fchollet (approved):
It's great that you're able to catch bugs like these. Thank you for the contribution, as usual!

PR #18491, opened 2023-09-25
Add integration test for backend-specific imports

nkovela1:
This PR adds an integration test to ensure that backend-specific imports are not used.

It works by creating a virtual environment where only one backend's packages are installed and then testing a basic subclassed model.fit() flow.

We can integrate this with CI in the next PR or this one.


integration_tests/import_test.py
```diff
@@ -0,0 +1,118 @@
...
+    )
+    run_commands_local(env_setup)
+
+
+def manage_venv_installs(whl_path):
+    other_backends = list(set(BACKEND_REQ.keys()) - {backend.backend()})
+    install_setup = [
+        # Installs the backend's package and common requirements
+        "pip install " + BACKEND_REQ[backend.backend()],
+        "pip install -r requirements-common.txt",
```
fchollet, line 51: Do we actually needed those? One we have the wheel we should be able to just install the wheel + pytest.

integration_tests/import_test.py
```diff
@@ -0,0 +1,118 @@
...
+        "pip install pytest",
+        # Ensure other backends are uninstalled
+        "pip uninstall -y "
```
fchollet, line 54: If you don't install `requirements` you also don't have to do this, which is a plus

integration_tests/import_test.py
```diff
@@ -0,0 +1,118 @@
...
+        # Install `.whl` package
+        "pip install " + whl_path + " --force-reinstall --no-dependencies",
+    ]
+    run_commands_venv(install_setup)
+
+
+def run_keras_flow():
+    test_script = [
+        # Runs the example script
+        "python -m pytest integration_tests/basic_full_flow.py",
```
fchollet, line 67: How do we check it actually works? I think we need some kind of output canary to check the script has successfully completed

fchollet (commented):
Thanks for the PR!

integration_tests/import_test.py
nkovela1: Actually `subprocess.run(command)` prints the output of the command directly to the console, so no output canary is needed. Here's the console output, for example:

image

nkovela1: And here's an example output when it failed (because I added `import jax` in saving_lib.py):

image

integration_tests/import_test.py
nkovela1: All these commands are run in the fresh virtual env, which does not have any of the common packages necessary like absl, h5py, etc.

integration_tests/import_test.py
nkovela1: This is just an extra check (the virtual environment is fresh, but this ensures there is absolutely no chance of the other backends being installed).

integration_tests/basic_full_flow.py
```diff
@@ -0,0 +1,51 @@
...
+        x = self.dense2(x)
+        return self.dense3(x)
+
+
+@pytest.mark.requires_trainable_backend
+class BasicFlowTest(testing.TestCase):
+    def test_basic_fit(self):
+        model = MyModel(hidden_dim=256, output_dim=16)
+
+        x = np.random.random((50000, 128))
```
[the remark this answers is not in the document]
nkovela1: Got it, I've simplified both the data size and dimensions. Thanks!

integration_tests/import_test.py
fchollet: The wheel should install anyway required by Keras. `requirements` is actually dev requirements not user-facing requirements.

fchollet (commented):
One last thing: this is an integration test, but those aren't yet run on CI. We should be running this check on CI.

We could start running integration tests on CI as a separate job, but it would be somewhat complicated. What do you think?

nkovela1:
> One last thing: this is an integration test, but those aren't yet run on CI. We should be running this check on CI.

I think we should merge this PR first, and then I will submit a separate PR to activate the integration tests in CI. Are there any remaining blockers on this PR itself?

fchollet (approved):
LGTM, thank you!

PR #18494, opened 2023-09-25
Fix TF optimizer with distribution strategy.

qlzh727:
The unit test has been updated to not unwrap the raw tf.Variable to mimic the production code behavior.

Also fix distributed_apply_weight_decay which was discovered during unit test fix.


fchollet (approved):
LGTM, thanks!

PR #18497, opened 2023-09-25
Disable the stateless apply for the tf based optimizer

qlzh727:
The main limitation is caused by the tf.distribute.Strategy API, which requires the input to be tf.Variable or mirrored variables. This cause us  to loss the chance to intercept the KerasVariable.assign(), which is the key API to do stateless update.

This should fix the unit test failure on HEAD.


qlzh727:
The jax failure doesn't seem to be related to my change.

fchollet (approved):
LGTM, thanks!

PR #18518, opened 2023-09-27
Make Dense and EinsumDense agree more closely

mattdangerw:
Minor annoyance when doing LoRA stuff. EinsumDense always has a bias attr, it is just sometimes `None`. Dense only has a bias attr if `use_bias` is True. I see no reason for the divergence.


fchollet (approved):
LGTM

PR #18519, opened 2023-09-28
Add variable.trainable setter so torch can set requires_grad

mattdangerw:
This looks like it has major impacts on torch efficiency when a layer or model is set to non trainable.

I ran a quick test on KerasNLP:

```python
classifier = BertClassifier()
classifier.backbone.trainable = False
classifier.fit(..)
```

Train step time went from 500ms before this change, to 150ms after.


haifeng-jin (commented):
Amazing! Thanks for the PR!

PR #18530, opened 2023-09-29
Remove redundant  call in ops.numpy.matmul

kiukchung:
`keras.backend.jax.numpy.matmul` already calls `convert_to_tensor()` on both operands so we can safely remove this redundant call in `keras.ops.numpy.matmul`.

see: https://github.com/keras-team/keras/blob/master/keras/backend/jax/numpy.py#L48-L50


fchollet (approved):
LGTM, thanks!

PR #18534, opened 2023-10-02
Apply `backend.result_type` to `bincount`, `substract`, `matmul`, `multiply`, `mean` and `max`

james77777778:
Followed by #18482 

This PR has applied `result_type` to the following ops:
- `bincount`
- `substract`
- `matmul`
- `multiply`
- `mean`
- `max`

### DIFF
--- a/keras/backend/jax/numpy.py
+++ b/keras/backend/jax/numpy.py
@@ -64,7 +64,7 @@ def mean(x, axis=None, keepdims=False):
     # correctly, so we compute with float32 and cast back to the original type.
     compute_dtype = dtypes.result_type(x.dtype, "float32")
     if "int" in ori_dtype or ori_dtype == "bool":
-        result_dtype = "float64" if ori_dtype == "int64" else compute_dtype
+        result_dtype = compute_dtype
     else:
         result_dtype = ori_dtype
     outputs = jnp.mean(x, axis=axis, keepdims=keepdims, dtype=compute_dtype)
--- a/keras/backend/jax/numpy.py
+++ b/keras/backend/jax/numpy.py
@@ -2,6 +2,7 @@ import jax.numpy as jnp
 
 from keras.backend import config
 from keras.backend.common import dtypes
+from keras.backend.common.variables import standardize_dtype
 from keras.backend.jax.core import cast
 from keras.backend.jax.core import convert_to_tensor
 
@@ -57,17 +58,21 @@ def multiply(x1, x2):
 
 
 def mean(x, axis=None, keepdims=False):
+    x = convert_to_tensor(x)
+    ori_dtype = standardize_dtype(x.dtype)
     # `jnp.mean` does not handle low precision (e.g., float16) overflow
     # correctly, so we compute with float32 and cast back to the original type.
-    outputs = jnp.mean(x, axis=axis, keepdims=keepdims, dtype=jnp.float32)
-    dtype = getattr(x, "dtype", None)
-    if hasattr(dtype, "name") and "float" in dtype.name:
-        return cast(outputs, dtype)
+    compute_dtype = dtypes.result_type(x.dtype, "float32")
+    if "int" in ori_dtype or ori_dtype == "bool":
+        result_dtype = "float64" if ori_dtype == "int64" else compute_dtype
     else:
-        return cast(outputs, config.floatx())
+        result_dtype = ori_dtype
+    outputs = jnp.mean(x, axis=axis, keepdims=keepdims, dtype=compute_dtype)
+    return cast(outputs, result_dtype)
 
 
 def max(x, axis=None, keepdims=False, initial=None):
+    x = convert_to_tensor(x)
     return jnp.max(x, axis=axis, keepdims=keepdims, initial=initial)
 
 
--- a/keras/backend/numpy/numpy.py
+++ b/keras/backend/numpy/numpy.py
@@ -44,12 +44,10 @@ def mean(x, axis=None, keepdims=False):
     axis = tuple(axis) if isinstance(axis, list) else axis
     x = convert_to_tensor(x)
     ori_dtype = standardize_dtype(x.dtype)
-    compute_dtype = dtypes.result_type(x.dtype, "float32")
     if "int" in ori_dtype or ori_dtype == "bool":
-        result_dtype = compute_dtype
+        result_dtype = dtypes.result_type(x.dtype, "float32")
     else:
         result_dtype = ori_dtype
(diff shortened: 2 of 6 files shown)


james77777778:
All tests passed except the failure in Codecov.

fchollet (commented):
Thank you for the PR!

keras/backend/jax/numpy.py
```diff
@@ -57,17 +58,21 @@ def multiply(x1, x2):
...
+    ori_dtype = standardize_dtype(x.dtype)
     # `jnp.mean` does not handle low precision (e.g., float16) overflow
     # correctly, so we compute with float32 and cast back to the original type.
-    outputs = jnp.mean(x, axis=axis, keepdims=keepdims, dtype=jnp.float32)
-    dtype = getattr(x, "dtype", None)
-    if hasattr(dtype, "name") and "float" in dtype.name:
-        return cast(outputs, dtype)
+    compute_dtype = dtypes.result_type(x.dtype, "float32")
+    if "int" in ori_dtype or ori_dtype == "bool":
+        result_dtype = "float64" if ori_dtype == "int64" else compute_dtype
```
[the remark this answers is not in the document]
james77777778:
```bash
~/workspace/keras $ JAX_DEFAULT_DTYPE_BITS=32 JAX_ENABLE_X64=true python3
>>> import jax.numpy as jnp
>>> x = jnp.ones((1,), "int64")
>>> print(jnp.mean(x).dtype)
float64
>>>
```

`jnp.mean` outputs `float64` when the input is `int64`. Should we consider changing this behavior?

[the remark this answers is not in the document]
fchollet: Yes, we should standardize on `float32`.

Generally speaking ops should behave like layers, and layers return `self.compute_dtype` as their return dtype, which for almost all layers is `float32` (unless using mixed precision in which case it's `float16`).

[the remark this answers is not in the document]
james77777778: Got it. I think it's reasonable to maintain consistency with the layers.

Fixed.

keras/backend/tensorflow/numpy.py
```diff
@@ -24,16 +24,37 @@ def add(x1, x2):
...
+            "float64",
+        ]:
+            if "int" in standardize_dtype(weights.dtype):
+                weights = tf.cast(weights, tf.int32)
+            else:
+                weights = tf.cast(weights, tf.float32)
+    else:
+        # TODO: should we upcast to int64 if the actual reaulting dtype is
+        # int64? Currently we follow the type rules of JAX.
+        dtype = "int32"
```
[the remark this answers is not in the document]
james77777778: The TODO has been removed.

keras/backend/jax/numpy.py
```diff
@@ -57,17 +58,21 @@ def multiply(x1, x2):
...
 def mean(x, axis=None, keepdims=False):
+    x = convert_to_tensor(x)
+    ori_dtype = standardize_dtype(x.dtype)
     # `jnp.mean` does not handle low precision (e.g., float16) overflow
     # correctly, so we compute with float32 and cast back to the original type.
-    outputs = jnp.mean(x, axis=axis, keepdims=keepdims, dtype=jnp.float32)
-    dtype = getattr(x, "dtype", None)
-    if hasattr(dtype, "name") and "float" in dtype.name:
-        return cast(outputs, dtype)
+    compute_dtype = dtypes.result_type(x.dtype, "float32")
```
fchollet, line 65: Shouldn't this be `result_type(x.dtype, config.floatx())` rather than hardcoding float32?

keras/backend/numpy/numpy.py
```diff
@@ -20,20 +20,37 @@ def einsum(subscripts, *operands, **kwargs):
...
+    dtype = dtypes.result_type(x1.dtype, x2.dtype)
+    return np.multiply(x1, x2).astype(dtype)
 
 
 def mean(x, axis=None, keepdims=False):
     axis = tuple(axis) if isinstance(axis, list) else axis
-    return np.mean(x, axis=axis, keepdims=keepdims)
+    x = convert_to_tensor(x)
+    ori_dtype = standardize_dtype(x.dtype)
+    compute_dtype = dtypes.result_type(x.dtype, "float32")
```
fchollet, line 47: Likewise here

keras/backend/tensorflow/numpy.py
```diff
@@ -182,7 +222,16 @@ def multiply(x1, x2):
 
 
 def mean(x, axis=None, keepdims=False):
-    return tfnp.mean(x, axis=axis, keepdims=keepdims)
+    x = convert_to_tensor(x)
+    ori_dtype = standardize_dtype(x.dtype)
+    compute_dtype = dtypes.result_type(x.dtype, "float32")
```
fchollet, line 227: Likewise here

keras/backend/torch/numpy.py
```diff
@@ -50,9 +63,15 @@ def mean(x, axis=None, keepdims=False):
     if axis == () or axis == []:
         # Torch handles the empty axis case differently from numpy.
         return x
-    # Conversion to float necessary for `torch.mean`
-    x = cast(x, "float32") if x.dtype in TORCH_INT_TYPES else x
-    return torch.mean(x, axis=axis, keepdims=keepdims)
+    ori_dtype = standardize_dtype(x.dtype)
+    compute_dtype = dtypes.result_type(x.dtype, "float32")
```
fchollet, line 67: And here

fchollet (commented):
👍

keras/backend/jax/numpy.py
james77777778: There is a note:
```python
    # `jnp.mean` does not handle low precision (e.g., float16) overflow
    # correctly, so we compute with float32 and cast back to the original type.
```
it should come from this PR: https://github.com/keras-team/keras-core/pull/410

I have added a test to verify the overflow behavior:
```python
# test overflow
x = np.array([[65500, 65500, 65500]], dtype="float16")
self.assertAllClose(knp.mean(x), np.mean(x))
```
|`jnp.mean(x)`|`jnp.mean(x, dtype="float32")`|`np.mean(x)`|`tfnp.mean(x)`|`tfnp.mean(x, dtype="float32")`|
|-|-|-|-|-|
|inf|65500|65500|inf|65500|

As a result, we should use float32 for jax and tensorflow to compute mean.
Since `torch.mean` only accepts floating points, casting is necessary.

fchollet (approved):
Thank you for the great contribution! LGTM.

Issue #18521, opened 2023-09-28
[Distribution] Support sharding for intermediate tensor within the model


qlzh727:
As a requirement for fine-grained distribution control, Keras distribution API should support specifying layout/sharding for the intermediate tensor in the model. A typical use case for this is specifying the layout for the model activation, which could potentially be very large.

# Potential API change

1. `keras.distribute.relayout(tensor, tensor_layout)`
Change the layout for a intermediate tensor in the keras model. This suppose to only work within the jitted function. For any eager use case, user might want to use `keras.backend.distribute_value(value, tensor_layout)`

2. `keras.distribute.Distribution.get_tensor_layout(path)`
Return the TensorLayout for the corresponding string path. This will be used for model parallel distribution to fetch the layout of a intermediate tensor in the model. Could return None if no layout is found.

PR #18536, opened 2023-10-02
Add sharding constraint for the output tensor in the model

qlzh727:
This will be used for sharding the intermediate state (eg activations).

Updates to the API:
1. `keras.distribute.relayout` for relayout/set sharding constraint for a tensor value.
2. LayoutMap are used to support layer name as the key for the output layout of the layer.

Unit test has been updated for JAX as a demonstration.

This should fix https://github.com/keras-team/keras/issues/18521


keras/layers/layer.py
```diff
@@ -808,6 +810,16 @@ def maybe_convert(x):
                         outputs = super().__call__(*args, **kwargs)
                 else:
                     outputs = super().__call__(*args, **kwargs)
+                # Change the layout for the layer output if needed.
+                # This is useful for relayout intermediate tensor in the model
+                # to achieve the optimal performance.
+                distribution = distribution_lib.distribution()
+                if distribution is not None:
+                    current_layer_path = current_path()
+                    layout = distribution.get_tensor_layout(current_layer_path)
```
fchollet, line 819: So for setting the layout of the output of a `Dense` layer in a subclassed model, you'd do like `layout_map["model/layers/dense_1"] = (...)`?

1. Is there a risk of confusing variable layouts and intermediate tensor layouts?
2. Should we be more specific, e.g. `layout_map["model/layers/dense_1/output"] = (...)` ? This could also leave the door open for `input` if ever needed.

fchollet (commented):
Thanks for the PR! Two points of discussion to finalize the API.

keras/backend/jax/distribution_lib.py
```diff
@@ -43,6 +43,13 @@ def distribute_value(value, tensor_layout):
     return jax.device_put(value, tensor_layout)
 
 
+def relayout(value, tensor_layout):
```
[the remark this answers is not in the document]
qlzh727: Indeed. I think the major difference here is one of them is suppose to work in jitted function, and the other one works in eager.  Currently I don't think we have a good way to detect what mode is user code in, and auto choose the proper API for them.

I think the jax `with_sharding_constraint` has a good indication that it applies the constraint to a intermediate tensor/state within the function. Maybe we should just aligned with that?

qlzh727:
@mattdangerw

keras/distribution/distribution_lib.py
```diff
@@ -507,6 +529,28 @@ def _maybe_populate_device_mesh(self, layout):
...
+        value: a Tensor to change the layout.
+        tensor_layout: TensorLayout to be applied on the value.
+
+    Returns:
+        a new value with the specified tensor layout.
+    """
+    if isinstance(value, KerasTensor):
+        # keras tensor is only used for building functional model, and can't be
+        # used to alter layout/sharding.
+        return value
```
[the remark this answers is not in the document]
qlzh727: Great question. The issue I hit when using kerasTensor with the jax sharding API is that it will always try to convert the KerasTensor to jax array, which result into error. It might make sense to add layout only when KerasTensor is a subclass or jax array or tf.Tensor.

keras/layers/layer.py
qlzh727: > So for setting the layout of the output of a `Dense` layer in a subclassed model, you'd do like `layout_map["model/layers/dense_1"] = (...)`?
> 
> 1. Is there a risk of confusing variable layouts and intermediate tensor layouts?
It does. And my original intent was actually same as option 2.

> 2. Should we be more specific, e.g. `layout_map["model/layers/dense_1/output"] = (...)` ? This could also leave the door open for `input` if ever needed.
That will definitely make the it more explicit. And also, this open the option for mapping to any intermediate keras operations within the layer body.

> 3. Is the full path too much information? What about `layout_map["dense_1/output"] = (...)`? Is that confusing?
Matt has the same question, and he was proposing the use regex.search instead of regex.match so that user can skip the prefix. My original implementation was trying to be a bit strict, so that the layout won't accidentally map to unwanted weights. In the case that there are overlapping rule that apply to same weights, currently the first one wins. Maybe we can take the regex.match approach, and raise an error when multiple rules is mapped to the same weights/tensor. (Probably I will do this in a separate PR.)

fchollet: > Maybe we can take the regex.match approach, and raise an error when multiple rules is mapped to the same weights/tensor

I think that's a good idea, we can use `search` and then make sure that each variable matches at most 1 rule.

keras/backend/jax/distribution_lib.py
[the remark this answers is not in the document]
fchollet: Q: what's the difference between `with_sharding_constraint` applied to the input tensor (i.e. the data array), vs calling `device_put` on the input tensor? Right now we use `device_put` for data distribution. Does `with_sharding_constraint` work for that?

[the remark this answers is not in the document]
qlzh727: So I have a some chat in the jax user group, and the conclusion is that the `with_sharding_constraint` is designed to be used only in jitted function. It might not work in a lot of cases outside of jax.jit.

I think we should rely on the device_put in the pure eager context, for input data, as well as variable initialization.

keras/layers/layer.py
qlzh727: Ack, let me do this in a separate PR.

keras/backend/jax/distribution_lib.py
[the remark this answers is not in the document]
fchollet: Let's do this, I think:

- Have a separate `distribute_variable` and `distribute_tensor` API. Apparently this may be needed for TF?
- In JAX `distribute_tensor`, check if we're in a tracing context. If so, use `sharding_constraint`. If not, use `device_put`.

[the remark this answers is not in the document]
qlzh727: Done. I have added at check for distribute_tensor for whether its in the jitted context, and I think it might not be a cheap check if we have to do it very often. I also didn't find a proper way to do this kind of check https://github.com/google/jax/discussions/9241. We might want to check with Jax team for this.

qlzh727:
PTAL again.

keras/layers/layer.py
qlzh727: @mattdangerw .

keras/backend/jax/distribution_lib.py
```diff
@@ -40,7 +67,13 @@ def distribute_value(value, tensor_layout):
     """
     if not isinstance(tensor_layout, jax.sharding.Sharding):
         tensor_layout = _to_jax_layout(tensor_layout)
-    return jax.device_put(value, tensor_layout)
+
+    # TODO(scottzhu): This might not be a cheap check, we should consider
+    # have some proper JAX API for doing this check.
```
fchollet, line 72: I consulted mattjj and that is not something they are considering.

keras/backend/jax/distribution_lib.py
```diff
@@ -40,7 +67,13 @@ def distribute_value(value, tensor_layout):
...
+    if jax_utils.is_in_jax_tracing_scope():
+        return jax.lax.with_sharding_constraint(tensor, tensor_layout)
+    else:
```
fchollet, line 75: You can remove the indent block and just do `return`

keras/backend/jax/distribution_lib.py
```diff
@@ -27,8 +29,33 @@ def list_devices(device_type=None):
     return [f"{device.device_kind}:{device.id}" for device in jax_devices]
 
 
-def distribute_value(value, tensor_layout):
-    """Distribute the value based on the layout.
+def distribute_variable(value, tensor_layout):
```
fchollet, line 32: it's a bit weird to have `distribute_variable(value, tensor_layout)` if we also have `distribute_tensor(value, tensor_layout)`. I suggest switching to `distribute_variable(value, layout)` and `distribute_tensor(value, layout)`

keras/backend/jax/distribution_lib.py
```diff
@@ -27,8 +29,33 @@ def list_devices(device_type=None):
...
+    """Create a distributed variable for JAX.
+
+    Since JAX doesn't have variable class, this will just return a jax.Array
```
fchollet, line 35: a variable class

fchollet, line 35: Use backticks for code keywords.

keras/distribution/distribution_lib.py
```diff
@@ -507,6 +529,28 @@ def _maybe_populate_device_mesh(self, layout):
 LayoutMap.get.__doc__ = LayoutMap.__getitem__.__doc__
 
 
+@keras_export("keras.distribution.distribute_tensor")
+def distribute_tensor(tensor, tensor_layout):
+    """Change the layout of a Tensor value in the jit function execution.
+
+    Note that this might not work outside of the jitted function for certain
+    backend. To change the layout of a value eagerly, please use
```
fchollet, line 537: It should work in both situations in JAX, right?

fchollet (commented):
Thanks for the updates!

keras/distribution/distribution_lib.py
qlzh727: It works for JAX, but might not for all the other backend, eg tf.dtensor.relayout() will only work with dtensor instance and also in a tf.function, and pytorch is also unknown for now. 

I think we will update the docstring once we have all the backend implemented, and reflect the final behavior.

keras/backend/jax/distribution_lib.py
qlzh727: Done.

qlzh727: Done.

keras/backend/jax/distribution_lib.py
qlzh727: Done.

keras/backend/jax/distribution_lib.py
qlzh727: Done.

keras/backend/jax/distribution_lib.py
qlzh727: Ack. Thanks for the confirmation.

fchollet (approved):
LGTM -- Thank you!

PR #18547, opened 2023-10-04
Fix issues with tensordot on torch backend

mattdangerw:


mattdangerw:
these fixes all look unrelated, but lmk if i am missing something here!

fchollet (approved):
LGTM, thank you

PR #18548, opened 2023-10-05
Apply `backend.result_type` to `absolute`, `argmax`, `argmin`, `argsort`, `ceil`, `clip` and `dot`

james77777778:
Followed by #18482 and #18534 

I choose these ops that are more commonly used in the codebase.
In `ceil`, int64 input will become float64 output but I have changed it to `backend.floatx()`.

The principles in my mind, which are derived from the previous discussion:
- match the type inference of JAX
- if the integer input becomes a floating point output, use `backend.floatx()`
- try to support all dtypes in all backends (even if it introduces some overhead for checking)

### DIFF
--- a/keras/backend/jax/numpy.py
+++ b/keras/backend/jax/numpy.py
@@ -185,7 +185,12 @@ def broadcast_to(x, shape):
 
 
 def ceil(x):
-    return jnp.ceil(x)
+    x = convert_to_tensor(x)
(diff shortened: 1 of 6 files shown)


james77777778:
Some tests have failed and I believe the root cause is the new release of torch 2.1.0.
I'm working on the fix.

fchollet (approved):
Thanks for the PR. Great work!

PR #18550, opened 2023-10-05
Fix for torch2.1.0

james77777778:
Known issues:
- `torchvision.transforms.functional.resize`
TypeError: expected size to be one of int or Tuple[int] or Tuple[int, int] or Tuple[int, int, int], but got size with types [<class 'numpy.int64'>, <class 'numpy.int64'>]

- `keras.backend.torch.optimizer`
RuntimeError: a leaf Variable that requires grad is being used in an in-place operation.

This PR:
- Convert np.ndarray to list of ints
- Add `@torch.no_grad` to torch optimizer


fchollet (approved):
Thanks for the fix.

PR #18551, opened 2023-10-05
Add multi-process/worker support for JAX distribution.

qlzh727:
The unit test for this is quite hard since it will require to setup multiple processes locally, and it doesn't work well with CPU testing.

I am trying to setup some testing infra internally in g3, with TPU backend, as well as leverage JAX team multi-process testing infra for verification.  See cl/570533324 for more details.


qlzh727:
This also need to be rebased upon #18536

fchollet (approved):
LGTM, thank you

Should we consider testing it or is there an obstacle with testing? I would worry that we may break it later if left untested.

qlzh727:
Having the test in OSS will be tricky since the current internal test rely on TPU, as well as the JAX internal testing framework for multi-process testing on one host. I share the same worry as you have, let me check with Jax team and see how they have the proper test coverage if they are also github first.

qlzh727:
> Having the test in OSS will be tricky since the current internal test rely on TPU, as well as the JAX internal testing framework for multi-process testing on one host. I share the same worry as you have, let me check with Jax team and see how they have the proper test coverage if they are also github first.

Checked with Yash from JAX team on cl/570533324. It seems that jax is still running with internal repo as source of truth, and rely on internal testing infra as the final guard before submission.

I guess for now I will setup the internal test, and rely on the copybara port change as a detection for any breakages.

fchollet (approved):
LGTM

PR #18552, opened 2023-10-05
Generalize tf and torch dataset check in case path is not strict prefix

grasskin:
Fails on internal imports.


haifeng-jin (approved):
Thanks! LGTM

fchollet (approved):
LGTM

PR #18573, opened 2023-10-08
Fix format and uncomment TF GPU embedding test

sampathweb:
* Now that Embedding Test is fixed, uncomment TF GPU Test
* Fix code format


fchollet (approved):
Thank you!

PR #18584, opened 2023-10-10
Replace `Tensorflow` with `numpy` for `nn_test`

sampathweb:
Replaces TensorFlow with Numpy expected results.
Test to verify it fixes Torch GPU `nn_test.py` correctness.


fchollet (approved):
LGTM, thank you!

PR #18595, opened 2023-10-11
Extend test coverage for TF SavedModel

nkovela1:
This PR extends test coverage for `tf.saved_model.save` to include fine tuning, passing signatures, and both fixed and non-fixed signatures of numerous dtypes.


fchollet (approved):
LGTM -- thank you!

PR #18603, opened 2023-10-13
Fixed clone_model with no inputs

jackd:
Fixes an issue where `clone_model` without inputs created inputs with an additional leading dimension. This does not affect MLP models (`Dense` layers can have any number of dimensions) but causes issues with layers that depend on inputs with a specific number of dimensions (e.g. Conv1D).


fchollet (approved):
LGTM, thanks for the fix.

PR #18617, opened 2023-10-16
Fix nightly pip

sampathweb:
Nightly pip job failed because it was missing the `checkout` step. Also added trigger via `workflow_dispatch` so we can test it without having to wait for nightly job to run.


fchollet (approved):
LGTM, thanks

PR #18636, opened 2023-10-17
Apply `backend.result_type` to `equal`, `exp`, `expm1`, `full`, `greater*` and `less*`

james77777778:
Followed by #18482 

The principles in my mind, which are derived from the previous discussion:
- match the type inference of JAX
- if the integer input becomes a floating point output, use backend.floatx()
- try to support all dtypes in all backends (even if it introduces some overhead for checking)


fchollet (approved):
LGTM -- thanks for the great contribution!

PR #18641, opened 2023-10-17
Fix torch reshape

mattdangerw:
ops.reshape(tensor, -1) wasn't working on torch backend. It's a fairly common numpy-ism.


fchollet (approved):
LGTM

PR #18644, opened 2023-10-18
Update line 543 in metrics_utils.py

PatReis:
Changed `y_pred.shape` to `ops.shape(y_pred)` for broadcasting of sample weights in `update_confusion_matrix_variables` to avoid shape error in tensorflow. Fix for issue #18620 .

```python
    if sample_weight is not None:
        sample_weight = ops.broadcast_to(
            ops.cast(sample_weight, dtype=y_pred.dtype), ops.shape(y_pred)  # instead of y_pred.shape
        )
```


fchollet (approved):
LGTM, thanks

PR #18656, opened 2023-10-19
Manually pad the tensor only when needed

haifeng-jin:
Torch does not support padding when `padding="same"` and `strides!=1`.
For other cases, use the built-in padding arg of the conv function in torch.

This PR saves memory for torch backend.
It increases the batch size of ResNet50 models from 64 to 128, which is the same as other backends now.

This PR is related to #18657.


fchollet (approved):
Nice fix! 👍

PR #18661, opened 2023-10-20
use cpu for backends when not testing on gpu

haifeng-jin:
This will prevent the GPU versions of different backends to mess up the GPU memory of each other.


qlzh727 (approved):

qlzh727:
Thanks for the fix.

fchollet (approved):

PR #18663, opened 2023-10-20
ensure the main process raise an exception when the subprocess fails

haifeng-jin:
When the subprocess of testing imports and basic workflow runs and errors out, the main process still exit with code 0, which would not fail the CI.

This PR will make the CI fail when the subprocess fails.

This PR should be merged after #18655, which fixes the current breakages.


fchollet (approved):
Nice -- now it does fail as expected. Thanks!

fchollet (approved):
LGTM

PR #18677, opened 2023-10-23
Add unit test for sync batch norm under distribution strategy.

qlzh727:
This should address the test coverage in https://github.com/keras-team/keras/pull/18671#issuecomment-1775712652.


fchollet (approved):
LGTM, thank you!

PR #18687, opened 2023-10-25
Fix tracking of special characters in the progress bar.

foxik:
Currently, if `prog_width == 0` (which does happen if an epoch has enough batches), we do not track the 8 special characters in `bar`, causing the progress logs to "jump" a few lines above.

### DIFF
--- a/keras/utils/progbar.py
+++ b/keras/utils/progbar.py
@@ -119,12 +119,13 @@ class Progbar:
                 numdigits = int(math.log10(self.target)) + 1
                 bar = ("%" + str(numdigits) + "d/%d") % (current, self.target)
                 bar = f"\x1b[1m{bar}\x1b[0m "
+                special_char_len += 8
                 prog = float(current) / self.target
                 prog_width = int(self.width * prog)
 
                 if prog_width > 0:
                     bar += "\33[32m" + "━" * prog_width + "\x1b[0m"
-                    special_char_len += 17
+                    special_char_len += 9
                 bar += "\33[37m" + "━" * (self.width - prog_width) + "\x1b[0m"
                 special_char_len += 9
 


fchollet:
Thanks for the PR. Can you demonstrate the correctness of the fix?

foxik:
> Thanks for the PR. Can you demonstrate the correctness of the fix?

First I note that the new counting of special characters can be trivially checked to be correct (just by counting special characters on the line above counting them), while it was not correct before (because the special characters on line 121 might not be added to `special_char_len`).

Then I include two videos of running the following simple script:
```python
import keras

(x_train, y_train), (x_test, y_test) = keras.datasets.mnist.load_data(path="mnist.npz")

model = keras.Sequential([
    keras.layers.Input([28, 28, 1]),
    keras.layers.Flatten(),
    keras.layers.Dense(100, activation="relu"),
    keras.layers.Dense(10, activation="softmax"),
])
model.summary()

model.compile(
    optimizer=keras.optimizers.Adam(),
    loss=keras.losses.SparseCategoricalCrossentropy(),
    metrics=[keras.metrics.SparseCategoricalAccuracy("accuracy")],
)

model.fit(
    x_train, y_train, validation_data=(x_test, y_test),
    batch_size=4, epochs=1,
)
```

Here is the result with the current `keras-nightly==3.0.0.dev2023102503`:
keras_head2

Here is the result with the patch applied:
keras_patch2

foxik:
BTW, I was wondering why this was not caught earlier -- I just realized the manifestation is terminal-specific.

The non-patched behavior is undoubtedly incorrect, because it generates more `\b` characters than regular characters that were printed. However, different terminals behave differently when a `\b` is emitted at the beginning of a line.
- My default `rxvt-unicode` "skips" to the end of the previous line. Because progbar also prints `\r` after the sequence of `\b`, the cursor ends at the beginning of the previous line
- However, `xterm` just ignores the `\b` at the beginning of a line and does not go to the previous line -- so the behavior seems correct.
- Inside `screen` (even when started in `xterm`), the behavior is the same as in `rxvt-unicode`, i.e., the `\b` goes to the previous line.

So the problem manifests only on some terminals, which might be the reason why it was not encountered sooner.

fchollet (approved):
LGTM

Issue #18666, opened 2023-10-22
[tensorflow] MirroredStrategy, LossScaledOptimizer - merge_call failed


crohkohl:
Hi,

Using the `LossScaledOptimizer` fails for `MirroredStrategy` with the following exception:

```
Exception has occurred: RuntimeError       (note: full exception trace is shown but execution is paused at: _run_module_as_main)
in user code:

    File "/usr/local/lib/python3.10/dist-packages/keras/src/backend/tensorflow/trainer.py", line 105, in one_step_on_data  **
        return self.train_step(data)
    File "/usr/local/lib/python3.10/dist-packages/keras/src/backend/tensorflow/trainer.py", line 72, in train_step
        self.optimizer.apply_gradients(zip(gradients, trainable_weights))
    File "/usr/local/lib/python3.10/dist-packages/keras/src/optimizers/base_optimizer.py", line 206, in apply_gradients
        self.apply(grads, trainable_variables)
    File "/usr/local/lib/python3.10/dist-packages/keras/src/optimizers/loss_scale_optimizer.py", line 183, in apply
        ops.cond(finite, handle_finite_grads, handle_non_finite_grads)
    File "/usr/local/lib/python3.10/dist-packages/keras/src/ops/core.py", line 594, in cond
        return Cond()(pred, true_fn, false_fn)
    File "/usr/local/lib/python3.10/dist-packages/keras/src/utils/traceback_utils.py", line 123, in error_handler
        raise e.with_traceback(filtered_tb) from None
    File "/usr/local/lib/python3.10/dist-packages/keras/src/backend/tensorflow/optimizer.py", line 82, in _internal_apply_gradients
        tf.__internal__.distribute.interim.maybe_merge_call(

    RuntimeError: Exception encountered when calling Cond.call().
    
[...]

PR #18691, opened 2023-10-25
Fix the loss scale optimizer for tf.distribute.

qlzh727:
This is trying  to address https://github.com/keras-team/keras/issues/18666

Please note that this fix put backend specific logic into common loss scale optimizer. Let me know if you would like me to create a backend subclass for tensorflow only loss scale optimizer.

### DIFF
--- a/integration_tests/tf_distribute_training_test.py
+++ b/integration_tests/tf_distribute_training_test.py
@@ -44,7 +44,6 @@ def test_model_fit():
     # Fit from numpy arrays:
     with strategy.scope():
         model.compile(
-            # optimizer=optimizers.SGD(learning_rate=0.001, momentum=0.01),
             optimizer=optimizers.LossScaleOptimizer(
                 optimizers.SGD(learning_rate=0.001, momentum=0.01)
             ),
--- a/integration_tests/tf_distribute_training_test.py
+++ b/integration_tests/tf_distribute_training_test.py
@@ -44,7 +44,10 @@ def test_model_fit():
     # Fit from numpy arrays:
     with strategy.scope():
         model.compile(
-            optimizer=optimizers.SGD(learning_rate=0.001, momentum=0.01),
+            # optimizer=optimizers.SGD(learning_rate=0.001, momentum=0.01),
+            optimizer=optimizers.LossScaleOptimizer(
+                optimizers.SGD(learning_rate=0.001, momentum=0.01)
+            ),
             loss=losses.MeanSquaredError(),
             metrics=[metrics.MeanSquaredError()],
             # TODO(scottzhu): Find out where is the variable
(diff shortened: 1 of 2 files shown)


integration_tests/tf_distribute_training_test.py
```diff
@@ -44,7 +44,10 @@ def test_model_fit():
     # Fit from numpy arrays:
     with strategy.scope():
         model.compile(
-            optimizer=optimizers.SGD(learning_rate=0.001, momentum=0.01),
+            # optimizer=optimizers.SGD(learning_rate=0.001, momentum=0.01),
```
mattdangerw, line 47: should we just delete this commented out line?

keras/optimizers/loss_scale_optimizer.py
```diff
@@ -150,37 +150,87 @@ def apply(self, grads, trainable_variables=None):
...
-            ]
-            self.inner_optimizer.apply(
-                unscaled_grads, trainable_variables=trainable_variables
-            )
+        if backend.backend() == "tensorflow":
+            self._tf_apply(grads, trainable_variables)
+        else:
+            self._common_apply(grads, trainable_variables)
+
+    def handle_finite_grads(self, grads, trainable_variables):
```
mattdangerw, line 158: Why are we pulling these out to toplevels? It is slightly confusing because we also have stateless versions for jax with the same name declared inline.

Probably we should either keep them all inlined, or pull them all out, with a `stateless_` prefix for the stateless ones?

mattdangerw (approved):
Looks good, just a couple comments.

keras/optimizers/loss_scale_optimizer.py
qlzh727: Thanks for the feedback. The root cause here for tf is that when running with distribution strategy, the cond and the actual apply need to happen in different context (cross-replica vs per-replica). This force us to have those method available at top level so that it can be used by distribution and non-distribution code path.

For the stateless path, I can also lift the handle_x function out, and we probably should add _ to make them private methods.

integration_tests/tf_distribute_training_test.py
qlzh727: Done. Thanks for catching this.

keras/optimizers/loss_scale_optimizer.py
qlzh727: Done. Updated.

mattdangerw: don't see the update yet, did you forget to push?

qlzh727: ops, just pushed now.

mattdangerw (approved):
LGTM!

PR #18692, opened 2023-10-25
Add support for Tensorflow SparseTensors: optimizers.

hertschuh:
All optimizers now handle sparse gradients in the form of `tf.IndexedSlices`. Where applicable, this results in partial updates of variables using `scatter_add`/`scatter_sub`.

This was done by performing the following changes in optimizers code:
- replacing all negative operations (e.g. `+`) with the equivalent Keras op (e.g. `ops.add`). Note that an exception was made for scalars.
- replacing `variable.assign/assign_add/assign_sub` with `self.assign/assign_add/assign_sub`.

### DIFF
--- a/keras/backend/tensorflow/optimizer.py
+++ b/keras/backend/tensorflow/optimizer.py
@@ -1,6 +1,7 @@
 import tensorflow as tf
 
 from keras import backend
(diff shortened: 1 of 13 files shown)


fchollet (approved):
LGTM -- great work!

PR #18710, opened 2023-10-31
Apply `backend.result_type` to `all`, `amax`, `amin`, `any`, `logical*`, `maximum`, `min`, `minimum`, `mod`, `nan_to_num` and `not_equal`

james77777778:
The dtype promotion and compatibility rules:
- Match the dtype inference from JAX.
- When an integer input becomes a floating-point output, use `backend.floatx()` for the best attempt.
- Make the best effort to support all dtypes for all operators in all backends (although this might introduce some overhead for checking).

Additionally, this PR has refactored the dtype tests involving two inputs. We can simply use the combinations of `ALL_DTYPES` instead of their product to verify the dtype inference of the operator. This has saved a significant amount of testing time.

Currently, the status of ops applied `backend.result_type` is as follows:

<details>

- [x] abs
- [x] absolute
- [x] add
- [x] all
- [x] amax
- [x] amin
- [ ] append
- [x] arange
- [ ] arccos
- [ ] arccosh
- [ ] arcsin
- [ ] arcsinh
- [ ] arctan
- [ ] arctan2
- [ ] arctanh
- [x] argmax
- [x] argmin
- [x] argsort
- [ ] array
- [ ] average
- [x] bincount
- [ ] broadcast_to
- [x] ceil
- [x] clip
- [ ] concatenate
- [ ] conj
- [ ] conjugate
- [ ] copy
- [ ] cos
- [ ] cosh
- [ ] count_nonzero
- [ ] cross
- [ ] cumprod
- [ ] cumsum
- [ ] diag
- [ ] diagonal
- [ ] diff
- [ ] digitize
- [ ] divide
- [x] dot
- [ ] einsum
- [x] empty
- [x] equal
- [x] exp
- [ ] expand_dims
- [x] expm1
- [x] eye
- [ ] flip
- [ ] floor
- [x] full
- [ ] full_like
- [x] greater
- [x] greater_equal
- [ ] hstack
- [x] identity
- [ ] imag
- [ ] interp
- [ ] isclose
- [ ] isfinite
- [ ] isinf
- [ ] isnan
- [x] less
- [x] less_equal
- [x] linspace
- [x] log
- [x] log10
- [x] log1p
- [x] log2
- [x] logaddexp
- [x] logical_and
- [x] logical_not
- [x] logical_or
- [x] logspace
- [x] matmul
- [x] max
- [x] maximum
- [x] mean
- [x] median
- [ ] meshgrid
- [ ] mgrid
- [x] min
- [x] minimum
- [x] mod
- [ ] moveaxis
- [x] multiply
- [x] nan_to_num
- [ ] ndim
- [ ] nonzero
- [x] not_equal
- [x] ones
- [ ] ones_like
- [ ] outer
- [ ] pad
- [ ] percentile
- [ ] power
- [ ] prod
- [x] quantile
- [ ] ravel
- [ ] real
- [ ] reciprocal
- [ ] repeat
- [ ] reshape
- [ ] roll
- [ ] round
- [ ] sign
- [ ] sin
- [ ] sinh
- [ ] size
- [ ] sort
- [ ] split
- [x] sqrt
- [ ] square
- [ ] squeeze
- [ ] stack
- [ ] std
- [x] subtract
- [ ] sum
- [ ] swapaxes
- [ ] take
- [ ] take_along_axis
- [ ] tan
- [ ] tanh
- [ ] tensordot
- [ ] tile
- [ ] trace
- [ ] transpose
- [x] tri
- [ ] tril
- [ ] triu
- [ ] true_divide
- [ ] vdot
- [ ] vstack
- [ ] where
- [x] zeros
- [ ] zeros_like

</details>

### DIFF
--- a/keras/backend/numpy/numpy.py
+++ b/keras/backend/numpy/numpy.py
@@ -492,7 +492,15 @@ def logspace(start, stop, num=50, endpoint=True, base=10, dtype=None, axis=0):
 
 
 def maximum(x1, x2):
-    return np.maximum(x1, x2)
+    x1 = convert_to_tensor(x1)
+    x2 = convert_to_tensor(x2)
+    dtype = dtypes.result_type(x1.dtype, x2.dtype)
+    # TODO: np.maximum doesn't support dtype=bfloat16, so we need to cast it
+    # first
+    if dtype == "bfloat16":
+        x1 = x1.astype(dtype)
+        x2 = x2.astype(dtype)
+    return np.maximum(x1, x2, dtype=dtype)
 
 
 def median(x, axis=None, keepdims=False):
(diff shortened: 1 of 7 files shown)


james77777778:
I have found that in some layers we need to add `ops.cast` after applying `backend.result_type`.

Let's take `ops.maximum` as example:
```python
# in normalization layer
                ops.maximum(
                    ops.sqrt(self.variance),
                    backend.epsilon(),  # <-- a python float will convert to float32 in ops.maximum
                )

# it breaks when using mixed precision because we expect the return to be a float16 or bfloat16
```
We could explicitly cast `backend.epsilon()` to `self.compute_dtype` to fix the mismatch of the output dtype.
And the above solution has been applied by this PR.

PR #18715, opened 2023-11-01
Fix page rendering issue for Adam Optimizer docs

SuryanarayanaY:
For the Adam Optimizer classes, It observed the inline arguments descriptions all merged in one single tab where as inserted arguments rendering in nice tabular form. Please refer the snapshot below.

As suggested by @MarkDaoust , It might be due to difference in indentation levels of inline arguments that defined in the source code w.r.t inserted arguments through the variable `base_optimizer_keyword_args` from the file below.

https://github.com/keras-team/keras/blob/e237f083784579e5dcbe578c78888951b95bf016/keras/optimizers/base_optimizer.py#L698


fchollet (approved):
LGTM, thank you

PR #18717, opened 2023-11-01
Multi-worker dataset distribution for model+data parallel distribution.

qlzh727:
See the unit test in cl/578605766


fchollet (approved):
LGTM, thank you!

PR #18724, opened 2023-11-03
fix `class_names` arg checking thinko

t-kalinowski:
Without this fix, passing valid values to `class_names` would raise a ValueError. E.g.,
```
image_dataset_from_directory(dir, class_names = ['one', 'two', 'three'])
```
would error, even when the subdirectories are `['one', 'two', 'three']`


fchollet (approved):
LGTM

PR #18731, opened 2023-11-06
added diff op

jackd:
Added wrapper for `numpy.diff` (no `prepend` / `append` args) and other backend implementations.

### DIFF
--- a/keras/ops/numpy.py
+++ b/keras/ops/numpy.py
@@ -140,6 +140,7 @@ zeros_like
 
 
 """
+import builtins
 import re
 
 import numpy as np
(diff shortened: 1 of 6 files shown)


jackd:
Formatting failure is in a file unrelated to this PR

keras/ops/numpy.py
```diff
@@ -153,6 +153,9 @@
 from keras.ops.operation import Operation
 from keras.ops.operation_utils import reduce_shape
 
+# save reference to built in max operation before redefinition
+_max = max  # pylint:disable=used-before-assignment
```
james77777778, line 157: Hi @jackd 
I believe you can use `import builtins` to avoid redefinition

For example:
https://github.com/keras-team/keras/blob/b6df549577a67c98ab1ce4df5c74681fcc98efa1/keras/backend/torch/numpy.py#L1322

jackd: Much cleaner, thanks!

fchollet (approved):

PR #18734, opened 2023-11-06
Pass dtype to cumsum and cumprod

### LINKED ISSUE #18730: cumulative sum/prod dtypes inconsistent with integer inputs
`ops.cumsum` and `ops.cumprod` call backend implementations that return tensors with inconsistent dtypes when the input array is an integer dtype.

Additionally, the no-axis `compute_output_spec` implementation for both `Operation`s returns `int32`, regardless of the input dtype.

Backends all have a `dtype` argument which can be used, but when not provided, the following behaviour is observed:

- numpy: promote signed ints to `int64`, unsigned ints to `uint64`
- torch / tensorflow: promote to `int64`
- jax: same as input array (though this is inconsistent with documentation - see [issue](https://github.com/google/jax/issues/18399)

Happy to put in a PR, just need to know what policy to implement.

SuryanarayanaY:
Currently the APIs [keras.ops.cumsum](https://github.com/keras-team/keras/blob/master/keras/ops/numpy.py#L1825) and [keras.ops.cumprod](https://github.com/keras-team/keras/blob/master/keras/ops/numpy.py#L1790) are calling numpy backend where in all `int` dtypes are being converted into `int64` if specific `dtype` is not provided explicitly for `numpy.cumsum` and `numpy.cumprod` as by default numpy will be converting this unspecified `int` dtype to `int64`.

Hence proposing to infer `dtype` and the pass it explicitly to numpy backend.

#Fixes #18730

### DIFF
--- a/keras/ops/numpy_test.py
+++ b/keras/ops/numpy_test.py
@@ -4733,3 +4733,26 @@ class NumpyDtypeTest(testing.TestCase, parameterized.TestCase):
             ),
             standardize_dtype(jnp.eye(3, 4, 1, dtype=dtype).dtype),
         )
+
+
+class NumpyOutputDtypesTest(testing.TestCase, parameterized.TestCase):
+    """Test the dtypes of numpy outputs to match with inputs dtype"""
+
+    INT_DTYPES = [x for x in ALLOWED_DTYPES if "int" in x]
+
+    @parameterized.named_parameters(named_product(dtype=INT_DTYPES))
+    def test_cumsum_with_int(self, dtype):
+        x = np.array([1, 1, 2, 3, 2, 4, 4, 5], dtype=dtype)
+        output = knp.cumsum(x)
+        self.assertEqual(
+            standardize_dtype(output.dtype), standardize_dtype(x.dtype)
+        )
+
+    @parameterized.named_parameters(named_product(dtype=INT_DTYPES))
+    def test_cumprod_with_int(self, dtype):
+        x = np.array([1, 1, 2, 3, 2, 4, 4, 5], dtype=dtype)
+
+        output = knp.cumprod(x)
+        self.assertEqual(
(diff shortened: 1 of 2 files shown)


fchollet (commented):
Thanks for the PR. Can you add a unit test to validate this fix?

keras/ops/numpy_test.py
```diff
@@ -4733,3 +4733,26 @@ def test_eye(self, dtype):
             ),
             standardize_dtype(jnp.eye(3, 4, 1, dtype=dtype).dtype),
         )
+
+
+class NumpyOutputDtypesTest(testing.TestCase, parameterized.TestCase):
```
james77777778, line 4738: Hi @SuryanarayanaY 

Just came across this PR :)

Instead of creating new class, I believe you can add these tests to `NumpyDtypeTest` which is designed to verify both input and output dtypes

https://github.com/keras-team/keras/blob/21a3c0b7ce2bf738d81e3737aedba1a1ebd93c8b/keras/ops/numpy_test.py#L4306

SuryanarayanaY: Hi @james77777778 , Yeah it can be done there. But the Doc string states this class is to verify Jax behaviour.Hence I thought of creating a new class.

gbaned:
Hi @SuryanarayanaY Can you please resolve conflicts? Thank you!

gbaned:
Hi @SuryanarayanaY Can you please resolve conflicts? Thank you!

james77777778:
Hi @gbaned 
It seems like this PR has already been addressed by #18813

gbaned:
Hi @SuryanarayanaY Can you please take a look on above [commen](https://github.com/keras-team/keras/pull/18734#issuecomment-1858847237)t from @james77777778. Thank you!

sachinprasadhs:
Closing this PR, since the above linked PR addresses the concern for all the backends. Thanks!

PR #18744, opened 2023-11-07
Added solve() function to all backends.

hazemessamm:
* Added Solve class to keras/ops/math.py
* Added solve function to keras/ops/math.py
* Added solve function to all backends.
* Added test cases.


fchollet (approved):

PR #18745, opened 2023-11-07
Add support for Tensorflow SparseTensors: end to end flow.

hertschuh:
Training, evaluation and inference now support sparse tensors end to end.

- Fixed an issue in graph mode with IndexedSlices.
- Added support for sparse tensors in the input conversion logic.
- Added end to end tests


fchollet (approved):
LGTM, thank you!

PR #18746, opened 2023-11-08
Update nightly.yml

toshinakatsu:
Added write access rights to resolve workflow configuration errors.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -20,6 +20,8 @@ jobs:
     env:
       PYTHON: ${{ matrix.python-version }}
       KERAS_BACKEND: ${{ matrix.backend }}
+    permissions:
+      id-token: write
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python


toshinakatsu:
I tried to solve the error in the image below on the issue of building and uploading the Wheel file, and added the code for write access rights, but it did not work,
I would like to know what to do.

スクリーンショット 2023-11-09 133527

sampathweb:
> I tried to solve the error in the image below on the issue of building and uploading the Wheel file, and added the code for write access rights, but it did not work, I would like to know what to do.
> 
> スクリーンショット 2023-11-09 133527

Nightly is getting uploaded daily - https://github.com/keras-team/keras/actions/runs/6820253296/job/18549094402
I am not sure what error you are referring here.  Can you post a link to the error?

haifeng-jin:
I believe the contributor is running into this error in their fork of the repo, which is not suppose to work. Only the `keras-team/keras` suppose to run nightly successfully and upload the package. All forks should fail this workflow.

All contributors should disable the workflow manually in their forks if they don't want to see the error.

fchollet:
Makes sense. Thanks for the clarification.

PR #18749, opened 2023-11-08
Implement missing pieces from Trackable interface for TF-based variable.

3lim:
Some tf.Variable implementations (such as `TPUDistributedVariable`) don't override _restore_from_tensors/_serialize_to_tensors since they rely on the `_gather_saveables_for_checkpoint` interface, which causes Keras to fail (with `NotImplementedError`) when creating a tf.Checkpoint in a distributed environment for example. Here, we add a simple workaround in case this interface is not implemented.

Additonally, we also add the _copy_trackable_to_cpu interface to support async checkpointing on TF.

### DIFF
--- a/keras/backend/tensorflow/core.py
+++ b/keras/backend/tensorflow/core.py
@@ -70,10 +70,21 @@ class Variable(
         return self.value._shared_name
 
     def _serialize_to_tensors(self):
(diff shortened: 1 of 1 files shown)


fchollet (approved):
LGTM

PR #18751, opened 2023-11-08
Add support for Tensorflow SparseTensors: public API to densify.

hertschuh:
The `sparse` argument of `backend.convert_to_tensor` is now public. This allows densifying sparse tensors by passing `False`.


fchollet (approved):
LGTM

PR #18755, opened 2023-11-09
Export `is_tensor` to the public API

tirthasheshpatel:
`is_tensor` was imported in the `keras/ops/__init__.py` file but not exported in the public API. This omission doesn't seem intentional. This PR fixes this by exporting the `keras.ops.is_tensor` function so it's available in the released public API.

@grasskin @fchollet


tirthasheshpatel:
Closed this in favor of #18756

PR #18757, opened 2023-11-10
Remove duplicate in nightly github action

sampathweb:
References the `actions.yml` in `nightly` as a dependency and removes the copy of the format, tests in `nightly.yml`.  This will align with how we do in `keras-nlp`.


fchollet (approved):
LGTM, thanks

PR #18760, opened 2023-11-10
Fix metric name for metrics with custom metric function.

hertschuh:
Metrics specified via a custom metric function were always called "mean_metric_wrapper". They take the function name, which is what tf-keras does.


fchollet (approved):
LGTM, thanks for the fix!

PR #18765, opened 2023-11-12
Added missing get_config on select LossFunctionWrappers

jackd:
Some `LossFunctionWrapper` implementations were still calling `LossFunctionWrapper.get_config` rather than `Loss.get_config`, making deserialization fail. This PR addresses these.

Note the existing `test_config` tests seem misleadingly named, as they don't test the config returned by `get_config`.


fchollet (approved):
LGTM, thanks for the fix!

PR #18773, opened 2023-11-13
Fix torch breakage for convert_to_tensor([])

mattdangerw:
Broke with https://github.com/keras-team/keras/pull/18752

Alternately we could handle this case in `torch.convert_to_tensor`, if we want to preserve the `result_type` error.


fchollet (approved):
LGTM

PR #18774, opened 2023-11-13
Fix tensorflow convert_to_tensor for boolean types

mattdangerw:
This appears to be a recent breakage, though I am not exactly sure what chaged. `ops.convert_to_tensor([1, 0, 1], dtype="bool")` works on all backends but tf. We should allow it to work on tf.


fchollet (approved):
LGTM

PR #18780, opened 2023-11-14
Improve speedup test in `py_dataset_adapter`

james77777778:
Instead of hardcoded 8 secs, we can verify the speed-up by comparing identical dataset with different settings in `PyDataset`
This might reduce false alarms in case the server accidentally runs slowly.


fchollet (approved):
LGTM, thanks!

PR #18785, opened 2023-11-15
Remove unnecessary `print` statement

dfalbel:
Removes print statement that unnecessarily prints intermediate computations to stdout.


fchollet (approved):
LGTM, thanks!

PR #18787, opened 2023-11-16
Revisit tf's numpy api

james77777778:
When ensuring consistent dtype inference across backends, I found that several tf's numpy ops, such as `cos`, `exp`, `log` and others, incorrectly promote bfloat16 to float64 or uint* to int*.

Currently, our approach involves casting back to the expected dtype. However, this is suboptimal, particularly for `mixed_bfloat16` training, as the intermediates might consume a large amount of memory.

To address this issue, we can replace the tfnp ops with native ops which are more bfloat16-friendly.

### DIFF
--- a/keras/backend/tensorflow/numpy.py
+++ b/keras/backend/tensorflow/numpy.py
@@ -556,7 +556,7 @@ def cosh(x):
 
 
 def count_nonzero(x, axis=None):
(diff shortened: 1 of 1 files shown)


fchollet (approved):
Thanks for the fixes! We have no particular preference for using tfnp over plain TF APIs, the main reason we did it in the first place was for development convenience (we had less to worry about when it comes to matching the NumPy API).

LGTM

PR #18789, opened 2023-11-16
Fix speedup test

james77777778:
I still saw the random failures of speed-up test in `py_dataset_adapter_test.py` after #18780 

The root cause might be the limited resouces (only 2 cpu for linux machine) of the runner to spawn multiple process in `PyDataset`:
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

```python
        ExamplePyDataset(
            x,
            y,
            batch_size=4,
            workers=4,
            use_multiprocessing=True,  # <---
            max_queue_size=8,
            delay=0.5,
        )
```

EDITED:
I have conducted 3 experiments on the runner and the results indicate that this issue should be fixed by this PR.
||backend|speed-up (sec)|no speed-up (sec)|
|-|-|-|-|
|1st|numpy|1.56|5.01|
|2nd|numpy|1.55|5.01|
|3rd|numpy|1.56|5.01|
|1st|jax|1.54|5.01|
|2nd|jax|1.57|5.01|
|3rd|jax|1.57|5.01|
|1st|tensorflow|1.58|5.01|
|2nd|tensorflow|1.54|5.01|
|3rd|tensorflow|1.58|5.01|
|1st|torch|1.53|5.01|
|2nd|torch|1.54|5.01|
|3rd|torch|1.52|5.01|

Runs:
- https://github.com/keras-team/keras/actions/runs/6886842509
- https://github.com/keras-team/keras/actions/runs/6886869800
- https://github.com/keras-team/keras/actions/runs/6886889355

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -81,7 +81,7 @@ jobs:
       #     coverage xml --omit='keras/applications/*' -o core-coverage.xml
       - name: Test with pytest
         run: |
-          pytest keras --ignore keras/applications --cov=keras -k test_speedup -s
+          pytest keras --ignore keras/applications -k test_speedup -s
           coverage xml --omit='keras/applications/*' -o core-coverage.xml
       # - name: Codecov keras
       #   uses: codecov/codecov-action@v3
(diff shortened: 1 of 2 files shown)


fchollet (approved):
Thanks for the fix -- that looks better.

PR #18800, opened 2023-11-20
Fix the device conversion logic for keras.distribute.list_devices()

qlzh727:
The https://github.com/keras-team/keras/commit/3d147bed1e7e25ff84b129374784aaf2aabe1ae8 was using the jax.Device.device_kind incorrectly. The jax.Device.device_kind will contain extra information like "tpu v3" or "Tesla V100-SXM2-16GB", which can't be used to parse the type. We should use `platform` instead.

Also adding tests to cover the device conversion logic.

Please also see internal cl in cl/584064211  for tpu in single and multi-process test.


fchollet (approved):
LGTM, thanks for the fix!

PR #18803, opened 2023-11-20
Use set comparisons for serialization errors

mattdangerw:
This is unimportant, just made this change while debugging a KerasNLP failure so though I would port here as well.

Makes error messages much more readable when things go wrong.


fchollet (approved):
LGTM, thanks!

PR #18809, opened 2023-11-21
Enable channels_first for torch CI tests

haifeng-jin:
DO NOT MERGE

pending on other pr to fix all torch channels first tests.


fchollet (approved):
LGTM

PR #18820, opened 2023-11-23
Improve dtype inference of `add`, `divide`, `maximum`, `minimum`, `multiply` and `subtract` when using python native dtypes

james77777778:
This PR has addressed the dtype inference for these ops with python `int` and `float` numbers.
The manually casting in the layer implementation from the previous PR can now be removed.

This should enhance UX when implementing new layers to support mixed precision training.


fchollet (approved):
LGTM, thank you!

PR #18816, opened 2023-11-23
Apply `backend.result_type` to `ones_like`, `outer`, `pad`, `prod`, `ravel`, `repeat`, `reshape`, `roll`, `round`, `sign`, `sort`, `split`, `square`, `squeeze`, `stack`, `std`, `swapaxes` and `zeros_like`

james77777778:
The dtype inference is complicated for `ops.power` so I left it as a todo item.

I might finish all ops for consistent dtype inference in 1 or 2 PR.

Currently, the status of ops applied `backend.result_type` is as follows (A-S):

<details>

- [x] abs
- [x] absolute
- [x] add
- [x] all
- [x] amax
- [x] amin
- [x] append
- [x] arange
- [x] arccos
- [x] arccosh
- [x] arcsin
- [x] arcsinh
- [x] arctan
- [x] arctan2
- [x] arctanh
- [x] argmax
- [x] argmin
- [x] argsort
- [x] array
- [x] average
- [x] bincount
- [x] broadcast_to
- [x] ceil
- [x] clip
- [x] concatenate
- [ ] conj  (Keras does not directly support complex data)
- [ ] conjugate  (Keras does not directly support complex data)
- [x] copy
- [x] cos
- [x] cosh
- [x] count_nonzero
- [x] cross
- [ ] cumprod  (left for #18734 #18813)
- [ ] cumsum (left for #18734 #18813)
- [x] diag
- [x] diagonal
- [x] diff
- [x] digitize
- [x] divide
- [x] dot
- [x] einsum
- [x] empty
- [x] equal
- [x] exp
- [x] expand_dims
- [x] expm1
- [x] eye
- [x] flip
- [x] floor
- [x] full
- [x] full_like
- [x] greater
- [x] greater_equal
- [x] hstack
- [x] identity
- [ ] imag (Keras does not directly support complex data)
- [ ] interp (Keras lacks this op)
- [x] isclose
- [x] isfinite
- [x] isinf
- [x] isnan
- [x] less
- [x] less_equal
- [x] linspace
- [x] log
- [x] log10
- [x] log1p
- [x] log2
- [x] logaddexp
- [x] logical_and
- [x] logical_not
- [x] logical_or
- [x] logspace
- [x] matmul
- [x] max
- [x] maximum
- [x] mean
- [x] median
- [x] meshgrid
- [ ] mgrid (Keras lacks this op)
- [x] min
- [x] minimum
- [x] mod
- [x] moveaxis
- [x] multiply
- [x] nan_to_num
[...]

PR #18824, opened 2023-11-24
Fix torch's `sort` with cuda

james77777778:
This PR has fixed the GPU CI breakage introduced by #18816


fchollet (approved):
Thanks for the fix!

PR #15771, opened 2021-12-09
Correcting documentation of padding="same" for conv layers

sanatmpa1:
Updating the doc string of padding='same' for a list of convolutional layers.

Reference issue #15703

Issue #15703, opened 2021-11-25
Padding="same" documentation is incorrect in all Conv layers


kkraoj:
**URL(s) with the issue:**.
https://keras.io/api/layers/convolution_layers/convolution1d/
https://keras.io/api/layers/convolution_layers/convolution2d/
https://keras.io/api/layers/convolution_layers/convolution3d/
https://keras.io/api/layers/convolution_layers/separable_convolution1d/
https://keras.io/api/layers/convolution_layers/separable_convolution2d/
https://keras.io/api/layers/convolution_layers/depthwise_convolution2d/
https://keras.io/api/layers/convolution_layers/convolution2d_transpose/
https://keras.io/api/layers/convolution_layers/convolution3d_transpose/

**Description of issue (what needs to be changed):**.
The documentation for padding in all conv layers listed above states that, " "same" results in padding with zeros evenly to the left/right or up/down of the input such that **output has the same height/width dimension as the input.**". The part in bold is incorrect (example shown below). setting padding = "same" only means that enough padding is applied to the input so that the kernel can fully cover the input.

Example:
```
from tensorflow import keras
from tensorflow.keras import layers

def get_model(img_size, num_classes):
    inputs = keras.Input(shape=img_size + (3,))
    x = layers.Conv2D(32, 3, strides=2, padding="same")(inputs)
    outputs = layers.Conv2D(64, 3, strides=1, padding="same")(x)

    # Define the model
    model = keras.Model(inputs, outputs)
    return model

img_size = (224, 224)
num_classes = 2
model = get_model(img_size, num_classes)
model.summary()
```
Results in:
```
_________________________________________________________________
 Layer (type)                Output Shape              Param #   
=================================================================
 input_4 (InputLayer)        [(None, 224, 224, 3)]     0         
                                                                 
 conv2d_4 (Conv2D)           (None, 112, 112, 32)      896       <-**Padding=same did not preserve image shape**
                                                                 
 conv2d_5 (Conv2D)           (None, 112, 112, 64)      18496     <-**Padding=same preserved image shape only because strides=1**
                                                                 
=================================================================
```
As you can see, when `strides=2`, the image shape is not preserved, it is halved in spite of setting `padding="same"`. Only when `strides=1`, the input shape is preserved. Thus the current documentation `padding` reflects its behaviour only in the special case of `strides=1` and not in the general case. I suggest two solutions: 

[...]

PR #18826, opened 2023-11-24
Change Conv layers Doc string wrt padding.

SuryanarayanaY:
"Updating the args documentation of padding='same' for all convolutional layers.

Reference issue #15703.

Reference PRs #15771 closed due to stale.

The documentation for padding in all conv layers listed above states that, "" ""same"" results in padding with zeros evenly to the left/right or up/down of the input such that output has the same height/width dimension as the input."". This seems incorrect as it is true only when `strides=1`. Hence update the same.


fchollet (approved):
LGTM, thanks!

PR #18835, opened 2023-11-27
Bug fix for TF backend using random zoom with flatten

haifeng-jin:
RandomZoom would lose the tensor shape when using with TF backend.
The multiple `None` dims would error out when used combined with the `Flatten` layer.

This PR simply adds back the shape in the underlying ops of TF backend.


fchollet (approved):
LGTM

PR #18843, opened 2023-11-28
Adding stop_evaluating and stop_predicting attributes for Callbacks

albertaillet:
### Summary

This pull request introduces the `stop_evaluating` and `stop_predicting` attributes available for callbacks, allowing users to halt the evaluation and prediction loops.

### Motivation

The existing `stop_training` attribute makes it possible to interrupt the training loop based on custom criteria in callbacks. However, similar capabilities are not available for the evaluation and prediction loops. The introduction of `stop_evaluating` and `stop_predicting` attributes extends this functionality, allowing users to use callbacks to stop these loops.

### Changes Made

1. **Added `stop_evaluating` attribute:**
   - Enables users to interrupt the evaluation loop based by marking is as `True` in a callback.
   
2. **Added `stop_predicting` attribute:**
   - Provides the ability to halt the prediction loop based by marking is as `True` in a callback.

### Example Usage

```python
import keras


class TimeOutCallback(keras.callbacks.Callback):
    """Stop evaluation loop after a certain amount of time."""

    def __init__(self, timeout: int) -> None:
        """Initialize the callback."""
        super().__init__()
        self.timeout = timeout

    def on_test_begin(self, logs: dict[str, Any] | None = None) -> None:
        """Initialize the stopping time."""
        self.stopping_time = time.time() + self.timeout

    def on_test_batch_end(self, epoch: int, logs: dict[str, Any] | None = None) -> None:
        """Stop evaluation if the timeout has been reached."""
        if time.time() > self.stopping_time:
            self.model.stop_evaluating = True
```

### DIFF
--- a/keras/backend/jax/trainer.py
+++ b/keras/backend/jax/trainer.py
@@ -579,6 +579,7 @@ class JAXTrainer(base_trainer.Trainer):
         self._record_training_state_sharding_spec()
 
         self.make_test_function()
+        self.stop_evaluating = False
         callbacks.on_test_begin()
         logs = None
         self.reset_metrics()
@@ -616,6 +617,8 @@ class JAXTrainer(base_trainer.Trainer):
                 "metrics_variables": metrics_variables,
             }
             callbacks.on_test_batch_end(step, self._pythonify_logs(logs))
+            if self.stop_evaluating:
+                break
 
         # Reattach state back to model.
         self.jax_state_sync()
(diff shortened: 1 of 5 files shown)


fchollet (commented):
Thanks for the PR! This addition makes sense for consistency. Can you add a unit test?

albertaillet:
I added the parametrized test `test_stop_loop` for testing `stop_training`, `stop_predicting` and `stop_evaluating`, based on the structure of the `test_validation_data_infinite_generator` test.

fchollet (approved):
LGTM, thank you

PR #18847, opened 2023-11-28
Add support for Tensorflow SparseTensors: preserve static shapes.

hertschuh:
Most Tensorflow sparse operation lose the static shape because the shape is only propagated using the `dense_shape` property, which is a tensor and is dynamic.

This is a workaround for this issue to allow static shape to be propagated as expected with Keras ops.


fchollet (approved):
LGTM, thank you

PR #18850, opened 2023-11-29
support numpy int dtypes in shapes

slevang:
One of two very small issues I ran into trying to run an existing code base on keras 3. This allows passing a tuple of numpy (or any other valid) int types to `Input(shape=...)`. I am preloading some data to generate the model shape on the fly, and the lib I'm using to wrangle that data happens to output shape as numpy ints.


fchollet (commented):
Thanks for the PR. Please add a unit test!

fchollet (approved):
LGTM, thanks!

PR #18851, opened 2023-11-29
support dict items in PyDataset

slevang:
The second of two very small issues I ran into trying to run an existing code base on keras 3. This allows the `PyDataset.__getitem__` method to return a dict of tensors, which is a valid way to supply inputs as `Model(inputs={...})` and very convenient for formatting complex inputs. We could add checks in the case of a dict to make sure the batch is not deeply nested, and doesn't contain other invalid objects. This was just a minimal fix.


fchollet (approved):
LGTM, thank you

PR #18867, opened 2023-12-01
[TODO] : Complete test for orthogonal regularizer

pranavvp16:
I have added the assert condition for the test and it passes successfully


fchollet (approved):
Nice, thank you! LGTM

PR #18868, opened 2023-12-01
Add a `force_download` arg to `get_file`

mattdangerw:
Useful for users not using hashes to force cache invalidation. Or for large files even with hashes, if you don't to read the entire cached file off disk before starting a new download.


fchollet (commented):
Thanks for the PR!

fchollet (approved):
LGTM, thanks!

PR #18902, opened 2023-12-07
Fix `ops.tril` and `ops.triu` with tensorflow backend

### LINKED ISSUE #18890: keras.ops.tril doesn't support dynamic shape with tensorflow backend
keras.ops.tril doesn't support tensor with unknown dimension with tensorflow backend.
it is OK with torch backend.
could you update keras.ops.tril implement with tensorflow ops to support input with unknown dimension?

```python
#!/usr/bin/python3

from os import environ
environ['KERAS_BACKEND'] = 'tensorflow'  # FAIL
#environ['KERAS_BACKEND'] = 'torch' # OK
import keras as K

def Test():
  inputs = K.Input((None, 100)) # inputs.shape = (batch x seq_len x hidden)
  attn = K.layers.Lambda(lambda x: K.ops.tril(K.ops.ones((K.ops.shape(x)[1],K.ops.shape(x)[1]))))(inputs)
  return K.Model(inputs = inputs, outputs = attn)

test = Test()
import numpy as np
print(test(np.random.normal(size = (1,10,100))).shape)
```

I find the following tensorflow implement support dimension with unknown dimension

```python
#!/usr/bin/python3

import tensorflow as tf

def Test():
  inputs = tf.keras.Input((None, 100))
  outputs = tf.keras.layers.Lambda(lambda x: tf.linalg.band_part(tf.ones((tf.shape(x)[1], tf.shape(x)[1])), -1, 0))(inputs)
  return tf.keras.Model(inputs = inputs, outputs = outputs)

test = Test()
import numpy as np
print(test(np.random.normal(size = (1,10,100))))
```

james77777778:
Fix #18890 

The corresponding test has been added.

### DIFF
--- a/keras/backend/tensorflow/numpy.py
+++ b/keras/backend/tensorflow/numpy.py
@@ -1462,20 +1462,26 @@ def tri(N, M=None, k=0, dtype=None):
 
 def tril(x, k=0):
     x = convert_to_tensor(x)
-    # TODO: tfnp.tril doesn't support bool
-    if standardize_dtype(x.dtype) == "bool":
-        x = tf.cast(x, "uint8")
-        return tf.cast(tfnp.tril(x, k=k), "bool")
-    return tfnp.tril(x, k=k)
+    if k >= 0:
+        return tf.linalg.band_part(x, -1, k)
+
+    # deal with negative k using mask
+    k = -k - 1
+    mask = tf.ones_like(x, dtype="bool")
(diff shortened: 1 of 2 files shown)


fchollet (approved):
Thank you for the fix! It seems like quite a few tfnp ops lack support for dynamic shapes, often due to relying to `x.shape`.

LGTM

PR #18908, opened 2023-12-07
Remove versioning restrictions for export from JAX backend

nkovela1:
Previously, JAX release 0.4.16 was incompatible with stable TF versions. However, this has recently been fixed, and JAX export can now be used again.


sampathweb (approved):
LGTM

fchollet (approved):
LGTM, Thank you

PR #18917, opened 2023-12-09
Rename kokoro to .kokoro

sampathweb:
Matt suggested to uses `.kokoro` as the folder name so its hidden in default view.  That's how its setup for CV and NLP. So renaming for `keras` as well.  Requires a internal cl change that's also in review.


fchollet (approved):
LGTM, thank you

PR #18927, opened 2023-12-12
Update `alpha` broadcasting and put `beta` on device (https://github.com/keras-team/keras/pull/18926#discussion_r1423237034)

KhawajaAbaid:
Learning from the improvements proposed in the PR https://github.com/keras-team/keras/pull/18926, specifically https://github.com/keras-team/keras/pull/18926#discussion_r1423237034 I've updated the existing `gamma` function in the `torch/random.py` module to:
1. Use `torch.broadcast_to` instead of `torch.ones` to broadcast `alpha` to the `shape`
2. Explicitly specify `device` parameter when using `torch.ones`  for `beta` values.


fchollet (approved):
LGTM, thank you!

PR #18928, opened 2023-12-12
Fix descrepancies in Conv Module docstrings regarding data_format

khdlr:
Stumbled upon some inconsistencies in the docs for conv layers, e.g. Conv2d, which currently states

> `"channels_last"` corresponds to inputs with shape `(batch_size, channels, height, width)`

which seems wrong, especially since a bit further down, it mentions

> Input Shape: If `data_format="channels_last"`: A 4D tensor with shape: `(batch_size, height, width, channels)`

Similar inconsisties exist for other conv layers, namely conv3d, conv3d_transpose, depthwise_conv2d, separable_conv2d.

So here's a simple Doc Fix to get the documentation for the `data_format` argument to conv layers in line with the expected input shape.


fchollet (approved):
Thanks for the fix! LGTM

PR #18932, opened 2023-12-12
Remove unused folders from docker image

sampathweb:
Docker images are limited to 14GB.  Lots of customers have raised the issue of running out of disk space in docker and GitHub suggested the following in https://github.com/actions/runner-images/issues/2840#issuecomment-790492173


sampathweb:
Issue in Keras-IO was unrelated to this.  Its due to requirements file had conflicts causing multiple downloads.  So closing this PR since no action required for Keras Repo.

PR #18934, opened 2023-12-13
added missing returns

jackd:


jackd:
It's not clear to me why `losses.get("mse")` shouldn't raise an error (i.e. I believe the tests are incorrect).

PR #18936, opened 2023-12-13
Fix Adafactor when variables have 2 dimensions or more.

hertschuh:
Adafactor would crash when using variables with 2 dimensions or more with most backends. It would work with the TensorFlow backend because `TensorShape` allows adding an int.


fchollet (approved):
LGTM, thank you for the fix!

PR #18937, opened 2023-12-13
Tweaks to TensofFlow support for SparseTensor/IndexedSlices.

hertschuh:
This is in preparation for JAX support of sparse tensors.

Added explicit support for `ops.isfinite`. This op is used by optimizers (`LossScaleOptimizer`) with sparse gradients. Somehow, `tf.isfinite` works out of the box with `tf.IndexedSlices` but this change makes the support explicit.

Removed support for `ops.mod` and `ops.floor_divide`. These ops are not used with sparse tensors in any context right now; support was incomplete with TensorFlow and there is no support for them in JAX for JAXSparse tensors.

### DIFF
--- a/keras/backend/tensorflow/numpy.py
+++ b/keras/backend/tensorflow/numpy.py
@@ -807,6 +807,7 @@ def isclose(x1, x2):
     return tfnp.isclose(x1, x2)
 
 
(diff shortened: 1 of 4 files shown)


fchollet (approved):
LGTM, thank you for the PR!

PR #18969, opened 2023-12-19
Use TorchEpochIterator for fit and evaluate too.

hertschuh:
`TorchEpochIterator` was only used for `predict`.


haifeng-jin (approved):
Thanks for the PR! LGTM

fchollet (commented):
Thanks for the PR. Does this lead to performance gains?

hertschuh:
> Thanks for the PR. Does this lead to performance gains?

@haifeng-jin what was the reason to add DataLoader support?

hertschuh:
These change were actually made as part of https://github.com/keras-team/keras/pull/18991

PR #18982, opened 2023-12-21
Save "layers" first for subclassed Functional models

mattdangerw:


mattdangerw:
@fchollet @nkovela1 after thinking a bit about our saved model issues for CV and NLP, I think this is the "right" fix.

Sequential and Functional are unique in that state attached to the models is primarily tracked internally and not via attrs set up by the user. When we are making a subclassed Sequential/Functional, this internal state should take precedence when de-duplicating layers to save.

Concrete examples--we recently added a `token_embedding` shorthand property to our NLP backbones. This was to allow a simple recipe for writing a language model... `logits = backbone.token_embedding(hidden_states, reverse=True)`. The layer is always captured in the functional graph (though in different places depending on the model).

We do not want adding something like this to impact the checkpoints structure at all! It would suck if we had to regenerate all our weights when we add attributes or properties for layers already in the functional DAG. Definitely we do not want to reason about which comes first alphabetically, attribute name or "layers".

This seems like the simplest fix. Alternately, we could add some sort of `do_not_track` annotation, but that seems like needless complexity right now.

fchollet (approved):
LGTM, thanks for the PR. The idea and mechanism make sense.

PR #18988, opened 2023-12-24
Fix `ops.split` in torch and tensorflow

james77777778:
It's a common use case to use a tensor to split another tensor.

This PR addresses the incompatibility issue of using a tensor input as `indices_or_sections` in `ops.split` for tensorflow and torch.

It is worth noting that I have implemented a more robust solution for tensorflow backend, supporting XLA compilation for tensor input of `indices_or_sections`.
The corresponding test `test_split_with_jit_in_tf` has been added.


fchollet (approved):
LGTM, thank you!

Issue #18947, opened 2023-12-15
remove reliance on this (or alternatively, turn it on by default).


pakapong777:
```
# TODO: remove reliance on this (or alternatively, turn it on by default).
```

https://github.com/keras-team/keras/blob/e70f28ff3f3b2060b83c23a1aabc85b4ccece991/keras/ops/nn_test.py#L27

divyashreepathihalli:
@james77777778 looks like the TODO was added by you. Can you please take a look?

james77777778:
> @james77777778 looks like the TODO was added by you. Can you please take a look?

Hi @divyashreepathihalli 

It was added to align the behavior in CI with `ops/numpy_test` for dtype inference.
This line should not affect the actual behavior of the codebase because `np_config.enable_numpy_behavior()` is not executed.

hertschuh:
Fixed

PR #18994, opened 2023-12-26
Remove reliance on Tensorflow enable_numpy_behavior for some ops.

hertschuh:
https://github.com/keras-team/keras/issues/18947

Some ops would only work properly if `enable_numpy_behavior` was on. This reimplements these ops using standard tf operations.


fchollet (approved):
LGTM, thank you!

PR #18998, opened 2023-12-28
Replace `tfnp` with native tf ops wherever possible

james77777778:
tf ops should be less buggy and more battle-tested. Some `tf.cond` used by `tfnp.*` could also be eliminated.

In addition, this PR adds dtype inference for `floor_divide`, which was missing before.

The following are left to be `tfnp.*` because they are much more complex to implement using native tf ops:
- max
- average
- copy
- cross
- diag
- diagonal
- diff
- linspace
- logspace
- min
- moveaxis
- roll
- swapaxes
- take
- take_along_axis
- trace
- tri
- eye

### DIFF
--- a/keras/backend/jax/numpy.py
+++ b/keras/backend/jax/numpy.py
@@ -207,6 +207,9 @@ def argmin(x, axis=None):
 
 
 def argsort(x, axis=-1):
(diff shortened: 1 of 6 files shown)


fchollet (approved):
LGTM, thank you

PR #19028, opened 2024-01-05
Text Vectorization Preprocessing Layer - Fixes

sampathweb:
Fixes - 
1. TextVectorization layer: local variable 'outputs' referenced before assignment error on some instances in the call method of TextVectorization. 
2. allow output_sequence_length when ragged is True - #18576 

Abstracted repeated use of `convert_to_tensor` on preprocessing layer outputs for non-TF backends into common utility in backend.

FYI: @SasankYadati.  I had made the change for issue #1 independently and then noticed that you had similar issue in #18980, so combined that issue also into this PR.  Hope that's OK.


fchollet (approved):
LGTM, thank you for the fix!

PR #19042, opened 2024-01-08
Remove stale `examples/keras_io`

mattdangerw:
Happy to leave these if we still have a reason, but as far as I know we have merged these back into keras-team/keras-io, the proper source of truth.

They have been popping up in code search results making it slightly harder to check things out in the repo (but not a big deal).


fchollet (approved):
Yep, I should have done that a while ago.

Issue #19043, opened 2024-01-09
Likely misbehaving keras.ops.triu()


gtikhonov:
**Describe the bug**
`keras.ops.triu(x, k)` nullifies only elements below main diagonal as if it always treats the `k` argument as if it is negative.

**To Reproduce**
```
import numpy as np
import keras
print(np.__version__, keras.__version__)
print(np.triu(np.ones([5,5]), k=1))
print(keras.ops.triu(keras.ops.ones([5,5]), k=1))
print(keras.ops.triu(keras.ops.ones([5,5]), k=-1))
```

**Expected behavior**
`keras.ops.triu(x, k)` would act according to its documentation for all integer `k` values.

**Additional context**
Numpy v1.23.5 
Keras v3.0.2

dugujiujian1999:
keras.ops.triu Actually calls keras.backend.tensorflow.numpy when backend is tensorflow.
I am check it right now.
I think this causes https://github.com/keras-team/keras/issues/19045

dugujiujian1999:
Looks like the bug was introduced in this pr:
https://github.com/keras-team/keras/commit/ef339e8adf005e24508fe2c2d1a885576be1378e

dugujiujian1999:
I am working on a fix to solve this.

Issue #19045, opened 2024-01-10
Stable diffusion is broken since Keras 3.0.2. Works fine with older versions of keras.


divyashreepathihalli:
Stable diffusion is broken since Keras 3.0.2. Works fine with older versions of keras.
Repro colab : https://colab.sandbox.google.com/drive/1CLsnhWJEa8ulwMXQ1SPN4tRU-LvkD9bf#scrollTo=Vo_YuNKb0DBj

I have narrowed it down to this part of the code 
this line is returning NaNs for keras 3.0.2 https://github.com/keras-team/keras-cv/blob/master/keras_cv/models/stable_diffusion/stable_diffusion.py#L123
calling predict_on_batch with TextEncoder class defined here https://github.com/keras-team/keras-cv/blob/master/keras_cv/models/stable_diffusion/text_encoder.py#L21

dugujiujian1999:
Fixed.
Check out my pr:
https://github.com/keras-team/keras/pull/19046

PR #19046, opened 2024-01-10
Rewrite triu and tril in tensorflow backend

dugujiujian1999:
* Use a new way to handle them.
* Add a new unit test.

Fix: https://github.com/keras-team/keras/issues/19043 and https://github.com/keras-team/keras/issues/19045


dugujiujian1999:
Maybe we don't need document here?

dugujiujian1999:
WIP, make it faster

dugujiujian1999:
Looks like something tf.linalg.band_part breaks something.

divyashreepathihalli (approved):

fchollet (approved):
LGTM, thank you

PR #19074, opened 2024-01-20
Allow bfloat16 as default dtype

mattdangerw:
Useful for llms! The tradeoff in precision can often be worth it in memory constrained environments, and unlike float16, does not have the same overflow/underflow issues during training.


fchollet (approved):
LGTM, thanks!

PR #19076, opened 2024-01-20
Fix repeat pytest in CONTRIBUTING.md

tylertking:
In `CONTRIBUTING.md`, `pytest` was repeated twice in the first code block of `Run tests` (i.e. `pytest pytest keras/losses/losses_test.py` fails but `pytest keras/losses/losses_test.py` doesn't).

Apologies if I missed something, I was just comparing it to tf-keras `CONTRIBUTING.md`.


fchollet (approved):
Thanks for the fix.

PR #19079, opened 2024-01-21
Add tests for  enabling LORA on a Dense and Embedding layer

Faisal-Alsrheed:
Add tests for enabling LORA on a Dense layer.
Add tests for enabling LORA on Embedding layer


fchollet (approved):
Thanks for adding those tests! LGTM

PR #19080, opened 2024-01-22
fix and add tests for /ops/math.py

Faisal-Alsrheed:
.


fchollet (approved):
Nice! Thanks for the PR.

PR #19092, opened 2024-01-24
Remove aliases for multi_head_attention layers

mattdangerw:
These aliases, when combined with the saving change below, breaks all saving involving the MultiHeadAttention layer and .keras style weights. https://github.com/keras-team/keras/blob/f97e3c7963c3c82e6a8bcea49e20ad6063054ed7/keras/saving/saving_lib.py#L290-L294

We either need to remove these alises, or change the linked saving logic. Basically, old saved model weights path would be to the private variable, new loading expects at a public attr without the underscore.


mattdangerw:
I think this was a false alarm. Misunderstood the code, and the error messages can be strange in the fail case.

PR #19095, opened 2024-01-24
Fix for `py_dataset_adapter_test` on GPU.

hertschuh:
Removed the parts of the test that were invalid. We were using libraries (TensorFlow, JAX, Torch) in multiple threads or multiple processes in a way that is not supported. This was causing issues with CUDA.


fchollet (approved):
LGTM, thanks for the fix!

Issue #19070, opened 2024-01-19
Operator Contribution


haohuanw:
i am working on keras2->keras3 adoptions for some of the models i am working with and i started to maintain a set of backend agnostic ops that used in our model that is not available in current keras ops. i am wondering if keras team open to have me upstream these operators.

operator lists:
1. [sequence_mask](https://www.tensorflow.org/api_docs/python/tf/sequence_mask)
2. [norm](https://www.tensorflow.org/api_docs/python/tf/norm)
3. [gather with batch_dim != 0](https://www.tensorflow.org/api_docs/python/tf/gather)
4. [adjust_{contrast,brightness}](https://www.tensorflow.org/api_docs/python/tf/image/adjust_contrast?hl=en)
5. [rgb_to_hsv (and vice versa)](https://www.tensorflow.org/api_docs/python/tf/image/rgb_to_hsv?hl=en)
6. [adjust_{hue, saturation}](https://www.tensorflow.org/api_docs/python/tf/image/adjust_hue?hl=en)
7. [flip_{up_down, lef_right}](https://www.tensorflow.org/api_docs/python/tf/image/flip_left_right?hl=en)

one additional thing i have implemented is equivalent to `tf.latest_checkpoint`, `tf.checkpoint_iterator` to ease some of our setup. i am wondering if this is something keras team is also interested to have in keras 3.

SuryanarayanaY:
Hi @haohuanw ,

If you feel any of Ops are useful to community and not available you are free to propose a PR for it. Team will review and take call on it. Please refer [contributing.md](https://github.com/keras-team/keras/blob/master/CONTRIBUTING.md) for more details. Thanks!

SuryanarayanaY:
Hi @haohuanw ,

I think we can close this issue. If you are ready with any implementation of the mentioned Ops please feel free to raise a PR as done in attached PR above.

Please feel free to close the issue. Thanks!

PR #19099, opened 2024-01-25
Add `ops.math.norm`

james77777778:
Related to #19070 

The `norm` in jax and torch is the same as numpy's `norm`.
However, significant modifications are required for tensorflow to match the output of the other backends.


fchollet (approved):

PR #19113, opened 2024-01-28
Refactor tf's `norm` and apply `result_type` to `norm`

james77777778:
This PR remove the redundant computation in #19099 by using native `tf.linalg.norm`.
It also includes the dtype tests for `norm`.

If I'm available, I will add more tests for other operators in `math`.


fchollet (approved):
Nice! LGTM, thank you

PR #19122, opened 2024-01-30
Fix TensorFlow sparse tests on GPU.

hertschuh:
Some TensorFlow sparse operations are not supported with XLA. This change turns off jit compilation for sparse tensor tests with the TensorFlow backend. This reverts it to what it was before https://github.com/keras-team/keras/pull/19109


fchollet (approved):
LGTM, thanks for the fix!

PR #19135, opened 2024-02-02
Support nested structures in while loops

mattdangerw:
Previously, we had the following support for loop vars:
- jax and tensorflow support nest structures (dicts/tuples) in a tuples
- torch and numpy support only flat tuples

With this change, we support any nested structure. If loops vars is a loop or tuple, args as passed positionally.

Alternatively, we could:
- demand the toplevel be a tuple/list, with nested structure inside
- pass any structure through unaltered (never pass args positionally). this is the jax approach, and IMO it's the cleanest, but to get there we would need to break compat.

### DIFF
--- a/keras/backend/jax/core.py
+++ b/keras/backend/jax/core.py
@@ -315,7 +315,8 @@ def while_loop(
     loop_vars,
     maximum_iterations=None,
 ):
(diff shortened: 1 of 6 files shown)


fchollet (approved):
LGTM, thanks!

PR #19136, opened 2024-02-02
Add `self._jax_state_synced` in `__init__` in jax's trainer

james77777778:
This PR fix the CI by adding`self._jax_state_synced=True` in `__init__`.

Additionally, it utilizes the new `self._get_jax_state` for `*_on_batch` and updates the logic behind `self.jax_state_sync` to avoid `if not self._jax_state_synced`


fchollet (approved):
LGTM, thanks for the fix!

PR #19162, opened 2024-02-09
Fix for `keras.metrics.Mean` when the sum of weights is negative.

hertschuh:
The new behavior is similar to Keras 2 i.e. `tf.math.divide_no_nan`.


fchollet (approved):
LGTM, thank you!

PR #19184, opened 2024-02-15
Add weights param to category_encoding

grasskin:
Weights tensor has to be a keyword argument unless we rework the tf_preprocessing layer class.


fchollet (approved):
Nice work, thanks for the fix. I have no idea why this argument was dropped in the first place.

PR #19186, opened 2024-02-16
Add `convert_to_tensor` in certain ops in JAX

james77777778:
This PR fixes the TypeError that occurs when certain ops do not retrieve the `value` from `KerasVariable` when using jax backend.

colab to reproduce the error:
[https://colab.research.google.com/drive/1LeIIJsj29K3uI2t2dXmd0yhHCCkKeHj4?usp=sharing](https://colab.research.google.com/drive/1LeIIJsj29K3uI2t2dXmd0yhHCCkKeHj4?usp=sharing) 

```bash
Argument '<KerasVariable shape=(1, 2, 3), dtype=float32, path=test_layer_4/variable_4>' of type <class 'keras.src.backend.jax.core.Variable'> is not a valid JAX type.

Arguments received by TestLayer.call():
  • inputs=jnp.ndarray(shape=(10, 2, 3), dtype=float32)
```

The solution is to add `convert_to_tensor` in these ops.

### DIFF
--- a/keras/backend/jax/numpy.py
+++ b/keras/backend/jax/numpy.py
@@ -421,6 +421,8 @@ def count_nonzero(x, axis=None):
 
 
 def cross(x1, x2, axisa=-1, axisb=-1, axisc=-1, axis=None):
+    x1 = convert_to_tensor(x1)
+    x2 = convert_to_tensor(x2)
     return jnp.cross(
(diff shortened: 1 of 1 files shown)


fchollet (approved):
LGTM -- thanks for the fix.

PR #19202, opened 2024-02-20
Check raised regex in convolutional tests

Frightera:
This PR updates tests to precisely match detailed error messages using regex, enhancing test reliability.


fchollet (approved):
LGTM, thanks!

PR #19214, opened 2024-02-22
Security Awareness for users on Security Policy

### LINKED ISSUE #19213: Security Awareness for users on Security Policy
Hi, I'm Joyce from [GOSST](https://github.com/joycebrum#about-gosst-ghost), and I'd like to suggest for you to adopt a security policy that not only allows security researchers to privately report security vulnerabilities in Keras, but also informs users of common security practices they should consider when using it.

This information will benefit:

1. the user, that will have guidelines on how to safely run a model for their application
2. the project, that can avoid receiving false positive vulnerability reports

I'll send a PR along with this issue with some suggestions such as data privacy, untrusted inputs, etc.
Thanks!

joycebrum:
Closes #19213 

Here are some topic ideas for using Keras securely and that also are not considered Keras vulnerabilities.

Let me know what you think.
Thanks!

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,5 +1,66 @@
 # Security Policy
 
+ - [**Using Keras Securely**](#using-keras-securely)
+   - [Untrusted inputs](#untrusted-inputs)
+   - [Data privacy](#data-privacy)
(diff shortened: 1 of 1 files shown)


fchollet (approved):
Thanks for the PR. The guidelines look good!

PR #19221, opened 2024-02-23
A few simplifications related to `ops.cond`.

hertschuh:
- Fixed an issue where an error resulting from using `ops.cond` was obscured by a different error caused by doing a symbolic call.
- Remove support for `KerasTensor` as return values for the functions passed to `ops.cond` as it is not a valid use case.
- Remove support for functions instead of tensors as inputs in `backend.compute_output_spec` as there is no use case for it.

### DIFF
--- a/keras/backend/jax/core.py
+++ b/keras/backend/jax/core.py
@@ -1,12 +1,9 @@
-import types
-
 import jax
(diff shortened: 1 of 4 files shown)


fchollet (approved):
Good simplification!

PR #19222, opened 2024-02-23
Update CI to TF 2.16 RC0

sampathweb:
Now that RC0 is released, pin to TF 2.16 RC0 instead of Nightly.

Spectral Norm test is failing on the Master as well for TF GPU Backend.


fchollet (approved):
Thanks for the PR! We're working on a fix for SpectralNorm.

PR #19232, opened 2024-02-27
Fix for TF XLA compilation error for `SpectralNormalization`.

hertschuh:
- On the Tensorflow backend, reimplemented all cases for `ops.linalg.norm` instead of sometimes relying on `tf.linalg.norm`. `tf.linalg.norm` sometimes fails to compile on XLA or returns tensors with no shape. Note that there is not special op used by `tf.linalg.norm`.
- Added more test cases for `ops.linalg.norm`:
  - vector norms are now also tested with 2 dimensional inputs
  - axis as a tuple of 2 values is tested
  - many more cases of invalid combinations of rank / ord / axis are verified


fchollet (approved):
LGTM, thank you!

PR #19233, opened 2024-02-27
Decorate `binomial` and `beta` functions with `keras_export`

KhawajaAbaid:
I noticed that the `@keras_export` decorator was missing from the `random.binomial` and `random.beta` functions which is causing issues when importing these functions from the `random` module.


fchollet (approved):
LGTM, thanks for the fix.

PR #19275, opened 2024-03-09
fix typo in `keras.distribution.initialize` (`proceed_id` -> `process_id`)

RobinKa:
Hey, excited about Keras 3 and trying out its distributed training, thanks for your work! I found a parameter name typo in `keras.distribution.initialize`.


fchollet (approved):
Nice catch, thanks!

Issue #19288, opened 2024-03-12
`draw_seed` method causes `Tensor on device cpu is not on the expected device meta!` error during `torch`'s symbolic execution


KhawajaAbaid:
Hi,
Using the `gamma`, `beta`  or `binomial` functions from the `keras.random` module on `torch` backend inside a model class causes the following error if we pass them an instance of `SeedGenerator` as a seed,
```
Tensor on device cpu is not on the expected device meta!
```
After a little digging, it became clear that this error is being caused by the `draw_seed`  method during `torch`'s symbolic execution.

I also noticed that the reason other random functions like `normal` do not cause this error is because they contain an explicit check to not call the `torch_seed_generator` and hence `draw_seed` method during the symbolic execution.
https://github.com/keras-team/keras/blob/0d0be6aadb7d9bfafaf8bea895789f4aa3b0abf7/keras/backend/torch/random.py#L31-L35

I've created this PR https://github.com/keras-team/keras/pull/19289 that fixes this issue by adding a similar check that avoids calling the `draw_seed` method during the symbolic execution.

PR #19289, opened 2024-03-12
Fix `draw_seed` causing device discrepancy issue during `torch`'s symbolic execution

KhawajaAbaid:
Fixes https://github.com/keras-team/keras/issues/19288


fchollet (approved):
Thanks for the fix!

Issue #19292, opened 2024-03-12
TestCase.run_layer_test for multi-output layers


shkarupa-alex:
Current implementation of TestCase.run_layer_test can't test layers with multiple outputs (returned as tuple).
Here is a test where such behavior may be required https://github.com/keras-team/keras/blob/master/keras/layers/rnn/rnn_test.py#L158

Even more current implementation have some misspellings.
For example here https://github.com/keras-team/keras/blob/master/keras/testing/test_case.py#L275 output_shape should be taken from output but not from expected_output_shape.

PR #19293, opened 2024-03-12
Fix TestCase.run_layer_test for multi-output layers

shkarupa-alex:
Fixes https://github.com/keras-team/keras/issues/19292


fchollet (approved):
Thanks, that's reasonable. LGTM

PR #19298, opened 2024-03-13
ArrayDataAdapter no longer converts to NumPy and supports sparse tens…

hertschuh:
…ors.

Instead, the passed arrays can be sliced or indexed in their native format.
- This addresses https://github.com/keras-team/keras/issues/18408 and improves performance, especially with Tensorflow and Torch. It improves TF -> TF and Torch -> Torch, but also TF -> Torch and Torch -> TF.
- This allows the support of sparse tensors (`tf.SparseTensor`, `jax.experimental.sparse.BCOO` and `scipy.sparse`). These sparse tensors are sliced as sparse and the iterators yield sparse tensors in the requested format (either TF or JAX).
- The `validation_split` argument of `Model.fit()` can now be used with anything supported by `ArrayDataAdapter`, in particular, sparse tensors are now supported.

In summary, `ArrayDataAdapter` now supports:
- native Python arrays
- NumPy arrays
- Tensorflow tensors, ragged tensors, sparse tensors (new)
- JAX arrays and BCOO sparse tensors (new)
- pandas DataFrames
- pandas Series
- scipy sparse matrices (new)

Also:
- Fixed bug where batch level shuffling would shuffle inconsistently the different arrays (in particular inputs and labels) when using a TF dataset or a NumPy iterator.
- Fixed bug where `tf.RaggedTensor`s would only work when using a TF dataset.
- Fixed bug where `tf.RaggedTensor`s would not work when doing batch level shuffling.
- Added a workaround for a bug where `tf.cast`ing a `tf.SparseTensor` would lose the static shape.
- Added test coverage for `tf.RaggedTensor`s and `pandas.Series`.
- Added verification in tests that inputs and labels are shuffled consistently.

### DIFF
--- a/keras/backend/jax/core.py
+++ b/keras/backend/jax/core.py
@@ -47,21 +47,23 @@ class Variable(KerasVariable):
 def convert_to_tensor(x, dtype=None, sparse=True):
     if dtype is not None:
         dtype = standardize_dtype(dtype)
-    if isinstance(x, (jnp.ndarray, jax.Array)) and dtype == x.dtype:
+    if isinstance(x, (jnp.ndarray, jax.Array)) and (
+        dtype is None or x.dtype == dtype
+    ):
         # Skip the conversion early if the instance is already a JAX array.
         # This is important in the multi-process context since jax.array(x) for
         # an existing distributed jax array will raise error.
         return x
 
     if isinstance(x, Variable):
-        if dtype and dtype != x.dtype:
+        if dtype is not None and x.dtype != dtype:
             return x.value.astype(dtype)
         return x.value
 
     if isinstance(x, jax_sparse.JAXSparse):
         if sparse is not None and not sparse:
             x = x.todense()
-        elif dtype and dtype != x.dtype:
+        elif dtype is not None and x.dtype != dtype:
             return x.astype(dtype)
         else:
             return x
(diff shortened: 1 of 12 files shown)


hertschuh:
> Great work! Are you seeing any performance impact?

Yes, here is the summary of the benchmark. I used a trivial model, clocked `model.fit()`, this is the time difference as a percentage, so a negative value is good (-50% means 2x faster).

Summary:
- Torch to Torch is a lot faster
- TF to TF is quite a bit faster
- the rest is about the same

|  ↓ Input ↓ \ Backend (output) → | Tensorflow | JAX | Torch |
|---------|--------|--------|--------|
| Numpy arrays          | 0 to -21%         | -3% to -4% | -1% to -2% |
| Tensorflow tensors | -23% to -44% | 14% to -1%‡ | -18% |
| JAX arrays               | 0 to -20%         | -3% to -9% | -2% |
| Torch tensors          | -14% to -32%  | 2% to -14% | -53% to -55% |

Notes:
- it's a range not a single value because performance impact is different for `shuffle=True / False / "batch"`
- Anything less than 4% is not significant
- ‡ TF tensors to JAX is a bit slower only when `shuffle=False`

fchollet (approved):
LGTM, thank you!

PR #19327, opened 2024-03-18
Set dtype policy for uint8

sampathweb:
Fixes issue for KerasCV where we use `uint8` for dtype.  Set to `FloatDtypePolicy` to get past the quantization dtype policy error. Tested locally with KerasCV to verify the fix.


fchollet (approved):
Thanks for the fix!

PR #19352, opened 2024-03-21
Allow save_model/load_model to accept a file-like object

adriangb:
I'm working on getting https://github.com/adriangb/scikeras and one of the needed features is being able to pickle models (e.g. for `GridSearchCV`). This is the first step for this, I can follow up with a PR to implement `Model.__reduce__` although that is not strictly necessary (it can be implemented via `copyreg.pickle` in SciKeras).

Since saving already worked with a file-like object internally I just had to refactor the logic into `_fileobj` methods making this a pretty straightforward change.


fchollet (approved):
Thanks for the PR. The code looks good.

PR #19359, opened 2024-03-22
Move `shape_utils.map_shape_structure` to `keras.utils.tree`.

hertschuh:
Now uses `optree` to do the traversal.


fchollet (approved):
LGTM, thank you!

Issue #19361, opened 2024-03-22
Initializing classes inside random_initializers.py with a seed value of 0 ignores the seed


untilted:
If one of the classes in `random_initializers.py` is initialized with the `seed` parameter set to 0, the code will ignore the seed and choose a random seed instead, resulting in non-deterministic results. I don't think this is intended, since it isn't documented and seems to have worked as expected in earlier versions.

The cause of this is the inadvertent use of `seed` as a boolean, e.g., in `RandomNormal`:
https://github.com/keras-team/keras/blob/ba43e3b876d1053cd0388e4fc2c9599eeb44b394/keras/initializers/random_initializers.py#L50
The same bug exists for other classes in this file as well (e.g., `TruncatedNormal`, `RandomUniform`, etc`).

sachinprasadhs:
Thanks for reporting the bug, I have created a PR to handle this edge case scenario.

PR #19364, opened 2024-03-23
fix seed value when seed=0

sachinprasadhs:
This fixes edge case scenario when `seed=0`.

According to the existing implementation for `seed=0`, `self.seed = seed or random.make_default_seed()` evaluates to 
`0 or random.make_default_seed()` which forces the seed value from  `random.make_default_seed()` and invalidate this logic for `0` input.

Fixes: https://github.com/keras-team/keras/issues/19361


fchollet (approved):
Nice catch, thanks for the fix!

PR #19369, opened 2024-03-24
only save to temp directories in tests

haifeng-jin:
When running tests locally, it would add an untracked file to git.
So changing it to use a temporary directory.


fchollet (approved):
LGTM, thanks for the fix. All and any file writing should always use `self.get_temp_dir()`.

PR #19370, opened 2024-03-25
Bug fix, model.evaluate return metrics in order.

haifeng-jin:
Related to #19267.
Before this fix, when returning a list from `model.evaluate()`, it would always sort the value by the metric names,
not respecting the order of the metrics configed by the user in `model.compile()`.

Added a test to ensure the behavior.
Before the fix, the test fails.
After the fix, the test passes.


fchollet (approved):
Thanks for the fix! LGTM

PR #19378, opened 2024-03-25
Add `dtype=float32` to PadSequencesTest

qaqland:
For function `keras.utils.pad_sequences`, it says

    sequences: List of sequences (each sequence is a list of integers).

but floats here also work. This commit add test for float type

Modifications to the function itself and document may involve API stability, so no changes have been made, further consideration may be needed


fchollet (approved):
LGTM, thanks

Issue #19395, opened 2024-03-28
RNN layers ignore recurrent_dropout parameter


drasmuss:
Setting the `recurrent_dropout` parameter on e.g. `SimpleRNN` has no effect. For example
```python
import keras
import tensorflow as tf

layer = keras.layers.SimpleRNN(10, recurrent_dropout=0.5)

inp = tf.ones((1, 10, 10))
out0 = layer(inp, training=True)
out1 = layer(inp, training=True)
assert tf.reduce_any(out0 != out1)
```
The assertion fails here because the layer produces the same output each time (with 0 dropout applied).

This is because this line https://github.com/keras-team/keras/blob/master/keras/layers/rnn/dropout_rnn_cell.py#L38 references `rate=self.dropout`. It should be `rate=self.recurrent_dropout`.

sachinprasadhs:
Thanks for pointing out the mistake, linked PR to change it to `recurrent_dropout`

PR #19397, opened 2024-03-28
Fix dropout for recurrent dropout

sachinprasadhs:
Fix typo for the the `recurrent dropout` implementation to use `recurrent_dropout` instead of `dropout`.

recurrent_dropout has no effect in the below usage as per existing implementation.

`layer = keras.layers.SimpleRNN(10, recurrent_dropout=0.5)` 

Fixes: https://github.com/keras-team/keras/issues/19395


fchollet (approved):
LGTM, thank you!

PR #19399, opened 2024-03-28
Fix variables binding in JAX export.

hertschuh:
In the current JAX export, while the variables are exported as `tf.Variable`s, these `tf.Variable`s are not at all connected to the graph. Instead, all the weights of the variables are inlined in the graph as constants.

As a consequence:
- The export is almost twice the size it should be
- One easily runs into the 2GB limit for the graph size with models with a lot of parameters
- Models that use RNGs in inference mode fail to export with a "leak" error.

This PR fixes the issues above by creating a stateless function, jax2tf converting it and binding the variables to the function in a `tf.function` wrapper. This follows the pattern documented at https://github.com/google/jax/blob/main/jax/experimental/jax2tf/README.md#saved-model-with-parameters .

Also:
- Fixed an issue where with the JAX export, the Python signature of the exported function was lost

### DIFF
--- a/keras/export/export_lib.py
+++ b/keras/export/export_lib.py
@@ -1,9 +1,12 @@
 """Library for exporting inference-only Keras models/layers."""
 
+import inspect
+
 from absl import logging
 
 from keras import backend
 from keras.api_export import keras_export
+from keras.backend.common.stateless_scope import StatelessScope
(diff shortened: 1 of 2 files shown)


fchollet (approved):
Nice -- thank you for catching this!

PR #19401, opened 2024-03-29
Refactor dtypes and add float8_* dtypes

james77777778:
This PR:
- Centralizes the dtype-related variables into `keras.backend.common.dtypes`
- Uses `tuple` instead of `list` for dtype string initialization to prevent future overwriting.
- Introduces `float8_e4m3fn` and `float8_e5m2` for the upcoming fp8 training
- Adds a simple test for `ops.cast` of `float8_*`

It is worth noting that `dtypes.result_type` will intentionally fail when encountering `float8_*` because there are no implicit conversions from `float8_*` to other dtypes. (following the same rule as in JAX)


fchollet (approved):
LGTM, thank you!

PR #19442, opened 2024-04-04
Fix for `JaxLayer` and `FlaxLayer` building in tracing scope.

hertschuh:
When building a sub-class model or layer that doesn't implement `build`, nested layers are first built in a tracing scope. This fix makes it so that the `JaxLayer` and `FlaxLayer` is built correctly the first time `build` is called in an eager scope.


fchollet (approved):
LGTM, thanks for the fix.

PR #19444, opened 2024-04-04
Update keras/losses/__init__.py to include Binary and Categorical Focal Crossentropy

Murhaf:
Binary and categorical focal crossentropy losses are not added to `ALL_OBJECTS_DICT` in `keras/losses/__init__.py` and therefore the following would raise a ValueError 

```python
losses.get("binary_focal_crossentropy")
```  

This PR imports the binary and categorical focal crossentropy classes and functions and adds them to `ALL_OBJECTS` (and, consequently, to `ALL_OBJECTS_DICT`)


fchollet (approved):
LGTM, thanks for the fix!

PR #19450, opened 2024-04-05
Fix a crash with PyTorch backend when using `weighted_metrics` with a GPU.

foxik:
This patch makes sure the `sample_weights` are on the default device. Before, it can happen that `sample_weights` are on CPU while values are on GPU (unless the `sample_weights` were moved to GPU by a dataloader, for example).

Note that the `y_true` are moved to the default device already during metric computation; for example https://github.com/keras-team/keras/blob/d4f23eac3a9469646a070129d723a532fdb07090/keras/metrics/accuracy_metrics.py#L9-L10 or https://github.com/keras-team/keras/blob/d4f23eac3a9469646a070129d723a532fdb07090/keras/metrics/f_score_metrics.py#L159-L160

I understand that a corresponding issue, reproducible example and a suitable test would be the ideal way to proceed, but unfortunately, my current time constraints only allow me to create this pull-request.

### DIFF
--- a/keras/metrics/reduction_metrics.py
+++ b/keras/metrics/reduction_metrics.py
@@ -10,7 +10,7 @@ def reduce_to_samplewise_values(values, sample_weight, reduce_fn, dtype):
     mask = getattr(values, "_keras_mask", None)
     values = ops.cast(values, dtype=dtype)
     if sample_weight is not None:
-        sample_weight = ops.cast(sample_weight, dtype=dtype)
+        sample_weight = ops.convert_to_tensor(sample_weight, dtype=dtype)
         if mask is not None:
             sample_weight = losses.loss.apply_mask(
                 sample_weight, mask, dtype=dtype, reduction="sum"


foxik:
Hm, according to tests, the `ops.convert_to_tensor` does not produce an output of a specified dtype on TensorFlow, which seems super-weird. Does anyone have an idea why it crashes? Maybe the solution would be to have `convert_to_tensor` followed by another `cast`? But it seems weird.

foxik:
Ah, now I understand. The test is given `dtype=None`, `values` of type `tf.float32` and `sample_weight` of type `tf.int32`. But while `keras.ops.cast(sample_weight, dtype=None)` produces an output of type `keras.config.floatx()`, because that is what `standardize_dtype` does, the `keras.ops.convert_to_tensor(sample_weight, dtype=None)` keeps the result as `tf.int32`.

Personally I would argue that that is a bad test. Should I fix it too?

fchollet (approved):
LGTM, thanks for the fix.

PR #19462, opened 2024-04-08
Add support for dynamic dimensions with model export.

hertschuh:
- Make the batch size dynamic for `model.export()` in the same way that it is with `ExportArchive`.
- Add support for non batch size dynamic dimensions for JAX. Previously, all dynamic dimensions where assumed to be the batch size, which only works for dimension 0.

Also added support for passing `polymorphic_shapes` in `jax2tf_kwargs`. This allows having constraints between dynamic dimensions. This allows easy control of dimensions without having the user call `jax2tf` by hand outside of `ExportArchive`.

### DIFF
--- a/keras/export/export_lib.py
+++ b/keras/export/export_lib.py
@@ -1,6 +1,8 @@
 """Library for exporting inference-only Keras models/layers."""
 
 import inspect
+import itertools
+import string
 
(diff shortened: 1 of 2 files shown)


fchollet (approved):
LGTM, thank you!

PR #19464, opened 2024-04-09
Fix type annotation for `Model`, `Sequential` and `Functional`

### LINKED ISSUE #19152: Change keras.Model inheritance for improved static analysis support
Hi all,

The `keras.Model` class may inherit from one of many different trainer classes. This is chosen at runtime by a conditional statement, which makes static analysis difficult. 

Knowing Python's Method Resolution Order (MRO) and the fact that all trainer classes have a common base class, I propose that `keras.Model` should inherit from the base trainer class. The change is minimal and would look like this:

```python
class Model(Trainer, Layer, base_trainer.Trainer):
```

This change fixes the syntax highlighting in my IDE, so I would really appreciate it.

Also, I believe it would also make the direct assignment of docstrings unnecessary, as they would be inherited from the base trainer class. The current state of the code is as follows:

```python
Model.fit.__doc__ = base_trainer.Trainer.fit.__doc__
Model.predict.__doc__ = base_trainer.Trainer.predict.__doc__
...
```

Thank you for your time and consideration.

james77777778:
Fixes #19152 (credit to @marcelo-souzaf)
Related to #19453 

This PR addresses incorrect type annotations for `Model`, `Sequential` and `Functional`
The root cause is the overridden `__new__`

I'm using VSCode to demonstrate:

|Class|Before|After|
|-|-|-|
|`Model`|  |  |
|`Sequential`|  |  |
|`Functional`|  |  |

`add` from `Sequential`:

### DIFF
--- a/keras/models/model.py
+++ b/keras/models/model.py
@@ -1,5 +1,6 @@
 import inspect
 import json
+import typing
 import warnings
 
 from keras import backend
@@ -27,7 +28,7 @@ else:
 
 
 @keras_export(["keras.Model", "keras.models.Model"])
-class Model(Trainer, Layer):
+class Model(Trainer, base_trainer.Trainer, Layer):
     """A model grouping layers into an object with training/inference features.
 
     There are three ways to instantiate a `Model`:
@@ -138,7 +139,7 @@ class Model(Trainer, Layer):
             from keras.models import functional
 
             return functional.Functional(*args, **kwargs)
-        return super().__new__(cls)
+        return typing.cast(Model, super().__new__(cls))
 
     def __init__(self, *args, **kwargs):
         Trainer.__init__(self)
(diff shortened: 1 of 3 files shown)


keras/models/model.py
```diff
@@ -599,11 +600,3 @@ def inject_functional_model_class(cls):
     cls.__new__(cls)
 
     return cls
-
-
-Model.fit.__doc__ = base_trainer.Trainer.fit.__doc__
```
fchollet, line 604: These are important -- they're how we get the right docstrings.

fchollet (commented):
Thanks for the fix!

keras/models/model.py
james77777778: These docstrings will be preserved because I have added `base_trainer.Trainer` to the inheritance of `Model`
(Actually, this solution was proposed by @marcelo-souzaf )

The screenshot:

adriangb:
Is Keras adopting some level of typing?

SamuelMarks:
@adriangb I hope so. I've got a bunch of tooling to automatically add typing to all of Keras, TensorFlow, and arbitrary other Python libraries. So when they're ready I am too.

PR #19472, opened 2024-04-10
Added GELU activation function. This commit introduces the Gaussian Error Linear Unit (GELU) activation function to the Keras activations module.

MadhumithaKolkar:
GELU is a smooth, non-monotonic function that has shown promising results in certain types of neural network architectures. The implementation follows the standard definition of GELU, utilizing TensorFlow's native functions for efficient computation. This enhancement expands the range of available activation functions in Keras, providing users with more options for building and training neural networks.


fchollet (commented):
Thanks for the PR -- but `gelu` is already an entry in `keras.activations` (on line 281).

PR #19485, opened 2024-04-10
Fix for `distribution_lib.list_device()` for TensorFlow backend.

hertschuh:
`distribution_lib.list_device(device_type=None)` would return CPU devices instead of returning GPU/TPU devices if available. This makes the behavior of `device_type=None` consistent with the JAX implementation and also consistent with the docstring.

Fixes `testing.tensorflow_uses_gpu()`.


fchollet (approved):
Thanks for the fix!

PR #19493, opened 2024-04-11
Add missing tests for `ops/math.py`

Faisal-Alsrheed:
I have added missing tests for `ops/math.py`

and 

I have suggested #19494


fchollet (approved):
LGTM -- thanks for the increased test coverage! 👍

PR #19497, opened 2024-04-11
Workaround for Tensorflow unit test crash on GPU.

hertschuh:
What was crashing was `tf.sparse.to_dense`, which was called by `convert_to_numpy`, which was called by `assertAllClose`.

Special handling is needed on GPU to densify sparse tensors that represent a scalar.

Also found a simpler way to create empty (i.e. with 0 dimensions) tensors.


fchollet (approved):
Thanks for the fix!

PR #19499, opened 2024-04-12
Add PSNR

IMvision12:
question: How to add tests for this functionality? should we load 2 images using `keras.utils` and compare the psnr values?

Here's PSNR values in different backends on sample images I used on colab:

```python
Tensorflow: 60.896583557128906
Numpy: 60.89658856391907
JAX: 60.89658737182617
Torch: 60.89658960237466
````

@fchollet

### DIFF
--- a/keras/backend/numpy/image.py
+++ b/keras/backend/numpy/image.py
@@ -17,7 +17,8 @@ def psnr(image1, image2, max_val):
     if image1.shape != image2.shape:
         raise ValueError(
             "Image shapes must match for PSNR calculation. "
-            f"Received shapes: image1.shape={image1.shape}, image2.shape={image2.shape}."
+            f"Received shapes: image1.shape={image1.shape}, "
+            "image2.shape={image2.shape}. "
         )
 
     max_val = np.float32(max_val)
--- a/keras/backend/numpy/image.py
+++ b/keras/backend/numpy/image.py
@@ -12,26 +12,21 @@ RESIZE_INTERPOLATIONS = (
     "bicubic",
 )
 
-def psnr(image1, image2, max_val, data_format="channels_last"):
-    if data_format == "channels_first":
-        if len(image1.shape) == 4:
-            image1 = np.transpose(image1, (0, 2, 3, 1))
-            image2 = np.transpose(image2, (0, 2, 3, 1))
-        elif len(image1.shape) == 3:
-            image1 = np.transpose(image1, (1, 2, 0))
-            image2 = np.transpose(image2, (1, 2, 0))
-        else:
-            raise ValueError(
-                "Invalid input rank: expected rank 3 (single image) "
-                "or rank 4 (batch of images). Received input with shape: "
-                f"image1.shape={image1.shape}"
-            )
+
+def psnr(image1, image2, max_val):
+    if image1.shape != image2.shape:
+        raise ValueError(
+            "Image shapes must match for PSNR calculation. "
+            f"Received shapes: image1.shape={image1.shape}, image2.shape={image2.shape}."
+        )
+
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: keras/backend/tensorflow/image.py [hunk 1]
```diff
@@ -14,6 +14,22 @@
...
+        elif len(image1.shape) == 3:
+            image1 = tf.transpose(image1, (1, 2, 0))
+            image2 = tf.transpose(image2, (1, 2, 0))
+        else:
+            raise ValueError(
+                "Invalid input rank: expected rank 3 (single image) "
+                "or rank 4 (batch of images). Received input with shape: "
+                f"image1.shape={image1.shape}"
+            )
+    psnr = tf.image.psnr(image1, image2, max_val, name=None)
```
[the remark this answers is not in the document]
IMvision12: lower-level op takes less time:

```python
TensorFlow op PSNR: 60.896584
TensorFlow lower op PSNR: 60.896584
Time taken for TensorFlow op: 0.009052515029907227
Time taken for TensorFlow lower op: 0.006928205490112305
```

keras/backend/numpy/image.py
```diff
@@ -12,6 +12,26 @@
     "bicubic",
 )
 
+def psnr(image1, image2, max_val, data_format="channels_last"):
+    if data_format == "channels_first":
```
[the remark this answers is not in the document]
IMvision12: Yup, I will change it!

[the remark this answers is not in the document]
IMvision12: should i remove `data_format` argument from all backends?

keras/backend/numpy/image.py
```diff
@@ -13,6 +13,22 @@
...
+            "Image shapes must match for PSNR calculation. "
+            f"Received shapes: image1.shape={image1.shape}, "
+            "image2.shape={image2.shape}. "
+        )
+
+    max_val = np.float32(max_val)
+    image1 = image1.astype(np.float32)
+    image2 = image2.astype(np.float32)
+    mse = np.mean(np.square(image1 - image2), axis=(-3, -2, -1))
+    psnr = 20 * np.log(max_val) / np.log(10.0) - (10 / np.log(10)) * np.log(mse)
```
fchollet, line 28: Instead of dividing by log(10) you should use the `log10` operator everywhere. I think it's available in all backends.

keras/backend/numpy/image.py
```diff
@@ -13,6 +13,22 @@
...
+def psnr(image1, image2, max_val):
+    if image1.shape != image2.shape:
+        raise ValueError(
+            "Image shapes must match for PSNR calculation. "
+            f"Received shapes: image1.shape={image1.shape}, "
+            "image2.shape={image2.shape}. "
+        )
+
+    max_val = np.float32(max_val)
+    image1 = image1.astype(np.float32)
```
fchollet, line 25: Is the float32 casting done for the log operation? In general it's preferred for ops not to cast inputs to a different dtype. What we should do is resolve the output dtype  based on `image1.dtype` and `image2.dtype`, and cast to that via `convert_to_tensor(image1, dtype=dtype)` (resp image2). Please check out how other ops with 2 inputs do this.

fchollet (commented):
Thanks for the update!

keras/backend/numpy/image.py
IMvision12: Ok will change that!

keras/backend/numpy/image.py
IMvision12: But i need to use float32 but image1.dtype gives float64

sampathweb:

PR #19516, opened 2024-04-15
Add LoRA to ConvND layers

james77777778:
This PR adds LoRA to `Conv1D`, `Conv2D` and `Cond3D`.

There is no official implementation from the original LoRA repo at:
https://github.com/microsoft/LoRA

I've adapted the implementation from some popular repos like:
- https://github.com/huggingface/peft/blob/main/src/peft/tuners/lora/layer.py#L754
- https://github.com/cloneofsimo/lora/blob/master/lora_diffusion/lora.py#L73

Basicly, `lora_kernel_a` would be a noraml conv and `lora_kernel_b` would be a pointwise conv as implemented from the above repos.


fchollet (approved):
Nice work -- LGTM.

This might break some folks who were doing direct setting of the `.kernel` attribute. For these users, I would recommend setting `._kernel` instead.

PR #19522, opened 2024-04-15
Models and layers now return owned metrics recursively.

hertschuh:
- added `Layer.metrics` to return all metrics owned by the layer and its sub-layers recursively.
- `Layer.metrics_variables` now returns variables from all metrics recursively, not just the layer and its direct sub-layers.
- `Model.metrics` now returns all metrics recursively, not just the model level metrics.
- `Model.metrics_variables` now returns variables from all metrics recursively, not just the model level metrics.
- added test coverage to test metrics and variables 2 levels deep.

This is consistent with the Keras 2 behavior and how `Model/Layer.variables` and `Model/Layer.weights` work.

### DIFF
--- a/keras/layers/layer.py
+++ b/keras/layers/layer.py
@@ -636,14 +636,19 @@ class Layer(BackendLayer, Operation):
         return [v for v in self.weights if not v.trainable]
 
     @property
+    def metrics(self):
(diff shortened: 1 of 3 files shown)


fchollet (approved):
LGTM, thank you!

PR #19534, opened 2024-04-17
Fix gradient accumulation when using `overwrite_with_gradient` during float8 training

james77777778:
Ref implementation:
https://github.com/google/flax/pull/3623

We should use `ops.maximum` for gradient accumulation with `overwrite_with_gradient=True` variables.

This PR also fixes a potential bug in gradient accumulation where directly using `self._accumulated_gradients` may mismatch with the filtered `grads` in the code path.

The new test `test_overwrite_with_gradient_with_gradient_accumulation` can only pass after the changes.

EDITED:

Fix all optimizers arguments hint by including `loss_scale_factor` and `gradient_accumulation_steps`:
|Before|After|
|-|-|
|before1|before|

### DIFF
--- a/keras/src/backend/jax/optimizer.py
+++ b/keras/src/backend/jax/optimizer.py
@@ -25,23 +25,24 @@ class JaxOptimizer(base_optimizer.BaseOptimizer):
             ]
             current_optimizer_vars_value = [v.value for v in self.variables]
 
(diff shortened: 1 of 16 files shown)


fchollet (approved):
LGTM, thanks for the fix!

PR #19541, opened 2024-04-18
Fix incorrect usage of `quantize`

james77777778:
This PR prevents incorrect usage of `quantize` on the layer with different quantization modes.

```python
from keras import layers
from keras import models

model = models.Sequential([layers.Input(shape=[16]), layers.Dense(32)])
model.quantize("int8")
model.quantize("float8")  # <- will succeed with the current codebase

```

This PR also replaces the original error msg with a more informative one:

```bash
ValueError: Layer 'dense' is already quantized with dtype_policy='int8_from_float32'. Received: mode=float8
```

### DIFF
--- a/keras/src/layers/core/dense.py
+++ b/keras/src/layers/core/dense.py
@@ -1,7 +1,6 @@
 import ml_dtypes
 
 from keras.src import activations
-from keras.src import backend
 from keras.src import constraints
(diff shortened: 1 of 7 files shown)


fchollet (approved):
LGTM, thanks for the PR!

PR #19566, opened 2024-04-20
Added tensordot

hazemessamm:
Added tensordot by using `mlx.core.tensordot(...)` function.

Note: I added `median()` implementation but I removed it because it will be better to add it in a different PR.


fchollet (approved):
LGTM, thank you!

PR #19568, opened 2024-04-20
Implemented `median(...)` function.

hazemessamm:
Currently, MLX does not have a bultin `median(...)` function so I implemented it by using `mx.sort(...)` and `mx.take(...)`.


fchollet (approved):
LGTM, thank you!

PR #19570, opened 2024-04-20
Fix the bug in `_int8_call` of `EinsumDense`

james77777778:
Actually, this is a bug introduced by my previous PR...

I've added the unit tests for this and the specified equations are retrieved from Gemma.


fchollet (approved):
LGTM, thanks for the fix.

PR #19575, opened 2024-04-20
list insert requires index

AlexanderLavelle:
super() calls python's builtin list class.

As per Python docs: 
```python
list.insert(i, x)
```
"Insert an item at a given position. The first argument is the index of the element before which to insert, so a.insert(0, x) inserts at the front of the list, and a.insert(len(a), x) is equivalent to a.append(x)."

https://docs.python.org/3/tutorial/datastructures.html


fchollet (approved):
LGTM, thanks for the fix.

PR #19576, opened 2024-04-21
mlx - skip float64 tests in nn_test.py

lkarthee:


fchollet (approved):
LGTM, thanks for the fix.

PR #19581, opened 2024-04-22
Fix keys of `save_own_variables` and `load_own_variables`

james77777778:
Consider following script:

```python
import keras

layer = keras.layers.Dense(4, use_bias=False)
layer.build([None, 2])
layer.quantize("int8")

store = {}
layer.save_own_variables(store)
print(list(store.keys()))

```

With the current codebase, we will get `['0', '2']`.

This PR updates the logic to strictly increments the key numbers. (`['0', '1']`)

This change will prevent a potential bug when adding quantization techniques to downstream projects such as
`keras_nlp.layers.ReversibleEmbedding`

### DIFF
--- a/keras/src/layers/convolutional/base_conv.py
+++ b/keras/src/layers/convolutional/base_conv.py
@@ -307,9 +307,11 @@ class BaseConv(Layer):
         # Do nothing if the layer isn't yet built
         if not self.built:
             return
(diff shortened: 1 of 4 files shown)


fchollet (approved):
LGTM, thanks for the fix.

PR #19592, opened 2024-04-23
Pickle support for all saveables

LukeWood:
I was doing some toying and realized that optimizers still do not pickle.  

I tried to refactor things into a MixIn to minimize code impact and improve reusability - but an issue of circular imports did arise.  As such I lazily import saving_lib in the depickle function, and this seemed to resolve the issue.

Basically you can put this in any Keras serializable class and it should just work.  I did:
```
class BaseOptimizer(KerasPickleable):
    ...
```
and it seems to just work.  Hopefully it should still be trivial to maintain as it relies on the saving logic.

---
One question: do you foresee any issues using the `._save_model_to_fileobj()` call on optimizers/layers?  Maybe I should use a different function?

---

Added tests for all optimizers, and a base mixin class that can trivially be mixed in if any other failures come up.


keras/src/optimizers/base_optimizer.py
```diff
@@ -6,11 +6,12 @@
 from keras.src import ops
 from keras.src.optimizers.schedules import learning_rate_schedule
 from keras.src.saving import serialization_lib
+from keras.src.saving.keras_pickleable import KerasPickleable
 from keras.src.utils import tracking
 from keras.src.utils.naming import auto_name
 
 
-class BaseOptimizer:
+class BaseOptimizer(KerasPickleable):
```
fchollet, line 14: The Pickleable thingy calls `_save_model_to_fileobj`. The name seems to indicate that it only works with models, but does it actually work for any serializable? If that's the case, then we should make the base layer class subclass Pickleable rather than Model.

LukeWood: It seems like it works (unit tests pass? though perhaps I should do more rigorous testing like training a model with a round trip pickled optimizer?)

perhaps there are some edge cases I'm missing?  I can try adding to Layer as well, and see if there are any issues.

keras/src/saving/keras_pickleable.py
```diff
@@ -0,0 +1,31 @@
+import io
+
+
+class KerasPickleable:
```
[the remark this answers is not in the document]
LukeWood: yeah that seems reasonable - I'll see what I can come up with.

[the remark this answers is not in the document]
LukeWood: Any ideas for things we could possibly streamline by rolling out a "Saveable" class?  I'm glancing over the code - and the only thing I can think of is maybe replacing:
```
    if isinstance(trackable, Sequential):
        obj_type = "Sequential"
    elif isinstance(trackable, Functional):
        obj_type = "Functional"
    elif isinstance(trackable, Layer):
        obj_type = "Layer"
    elif isinstance(trackable, Optimizer):
        obj_type = "Optimizer"
    elif isinstance(trackable, Metric):
        obj_type = "Metric"
    elif isinstance(trackable, Loss):
        obj_type = "Loss"
    else:
```

with some sort of method on the base class - though perhaps I'm missing something obvious?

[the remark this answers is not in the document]
fchollet: Yes, I think that's the applicable part -- registering what needs to be known by the saving logic. It's not much.

keras/src/optimizers/base_optimizer.py
LukeWood: It seems to work all over from what I can tell 👍 I think this is because under the hood they basically just call `serialize_keras_object`/`deserialize_keras_object`.

LukeWood:
Ok awesome - I think it looks pretty good overall.  I'm happy with it.  I tried to add basic coverage, let me know if you find the existing tests insufficient.  I'm pretty happy with how it turned out overall.

LukeWood:
Do I need to make the `obj_type` methods `_` prefixed to prevent them from showing up in `keras/api/_tf_keras`?

keras/src/saving/saving_lib.py
```diff
@@ -319,23 +319,16 @@ def _name_key(name):
 
 
 def _walk_trackable(trackable):
```
fchollet, line 321: Can you also update `def _is_keras_trackable` in this file?

fchollet (commented):
Thanks for the update!

fchollet (approved):
Actually I'll just update on top, I realize I have a few changes I'd like to make

LukeWood:
Sweet!  Thanks for the patches.

PR #19595, opened 2024-04-23
Reduce int8 quantization error for `Embedding` layer

james77777778:
Related to https://github.com/keras-team/keras-nlp/pull/1591

Mainly, this PR reduces quantization error by choosing `output_dim` as reduced axis because, typically, `input_dim` is larger than `output_dim`.


fchollet (approved):
Thanks for the PR -- makes sense.

PR #19621, opened 2024-04-26
Making extract_patches compatible with images of dtype float16 along with float32

VachanVY:
Error when an image of dtype `float16` is used in `nn.ops.image.extract_patches`
[Colab link](https://colab.research.google.com/drive/1y41vpFs6Cd4NHmSY6IRmU53PXgl_SAFE?usp=sharing)
Fixed it.


fchollet (approved):
LGTM, thanks for the fix

PR #19624, opened 2024-04-26
Add support for infinite `PyDataset`s.

hertschuh:
`PyDataset` now uses the `num_batches` property instead of `__len__` to support `None`, which is how one indicates the dataset is infinite. Note that infinite datasets are not shuffled.

Fixes https://github.com/keras-team/keras/issues/19528

Also added exception reporting when using multithreading / multiprocessing. Previously, the program would just hang with no error reported.


fchollet (approved):
LGTM -- thank you for the feature addition!

The JAX tests appear to have timed out. I think this may be related, since you're adding a test product over a PyDataset test case (each of which tends to take a bit of time). Let's see if it keeps happening or not.

PR #19633, opened 2024-04-28
Fix `ops.ctc_decode`

james77777778:
1. Compute `label_length` by counting non-`mask_index` elements in `losses.CTC`
2. Fix a bug in greedy decoding
3. Fix a bug where TF's beam search decoding didn't respect `mask_index`
4. Update the argument so that `mask_index` defaults to `0` which is consistent with `losses.CTC`

The unit tests have been updated.

I've also verified that this example will be compatible with Keras3 after the PR:
https://keras.io/examples/vision/handwriting_recognition/

|Backend|Training Speed|Validation Loss (at epoch=10)|
|-|-|-|
|tensorflow|~53ms/step|2.9052|
|jax|~9ms/step (super fast...)|2.9896|

- `tensorflow`, `ops.ctc_decode(strategy="beam_search")` at epoch=10:
tf_beam_search

- `jax`, `ops.ctc_decode(strategy="greedy")` at epoch=10:
inference

### DIFF
--- a/keras/src/backend/jax/nn.py
+++ b/keras/src/backend/jax/nn.py
@@ -593,7 +593,7 @@ def ctc_loss(target, output, target_length, output_length, mask_index=0):
     output = convert_to_tensor(output)
     target_length = convert_to_tensor(target_length, "int32")
     output_length = convert_to_tensor(output_length, "int32")
-    batch_size, _, num_classes = output.shape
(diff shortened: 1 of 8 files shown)


fchollet (approved):
Excellent work -- thanks for the fixes.

PR #19634, opened 2024-04-28
Fix the device issue with `argpartition` in torch

james77777778:
`torch.topk` is not guaranteed to be stable in cuda, so I've adjusted the test to prevent identical values


fchollet (approved):
LGTM, thanks

PR #19663, opened 2024-05-03
Fix bug in `model.export()` when inputs in a structure.

hertschuh:
The code computing the `call` signature was not going through the structure correctly.

Also tweaked tracking in `Layer` class:
- the first attribute set was not tracked.
- `TrackedDict` does not play nice with `jax2tf`/JAX so `_build_shapes_dict` is not tracked.

### DIFF
--- a/keras/src/layers/layer.py
+++ b/keras/src/layers/layer.py
@@ -1350,10 +1350,10 @@ class Layer(BackendLayer, Operation, KerasSaveable):
     def __setattr__(self, name, value):
         # Track Variables, Layers, Metrics, SeedGenerators.
         name, value = self._setattr_hook(name, value)
-        if hasattr(self, "_tracker"):
+        if name != "_tracker":
+            if not hasattr(self, "_tracker"):
+                self._initialize_tracker()
             value = self._tracker.track(value)
-        elif name != "_tracker":
-            self._initialize_tracker()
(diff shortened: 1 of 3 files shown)


keras/src/layers/layer.py
```diff
@@ -1350,10 +1350,11 @@ def __str__(self):
...
         # Track Variables, Layers, Metrics, SeedGenerators.
         name, value = self._setattr_hook(name, value)
-        if hasattr(self, "_tracker"):
-            value = self._tracker.track(value)
-        elif name != "_tracker":
-            self._initialize_tracker()
+        if name != "_tracker":
+            if not hasattr(self, "_tracker"):
+                self._initialize_tracker()
+            if name != "_build_shapes_dict":
```
hertschuh, line 1356: This is not very generic. I can think of some alternatives:
1. A "do not track" option on a per value basis: `self._build_shapes_dict = tracking.do_not_track(shapes)`
2. A "do not track" option on a per attribute name basis in `Tracker.init()`
3. A way to transform a structure by turning custom collections to plain `dict`s/`list`s. For the better or the worse `optree` and `dmtree` are smart enough to use the same custom classes when doing `map_structure`.

hertschuh: Reverted in favor of 3.

fchollet (approved):

Issue #19571, opened 2024-04-20
MLX Backend 


lkarthee:
### `mlx.math` 

- [ ] `fft` 
- [ ] `fft2`
- [ ] `rfft`
- [ ] `irfft`
- [ ] `stft`
- [ ] `istft` 
- [ ] `logsumexp`
- [ ] `qr`
- [ ] `segment_sum`
- [ ] `segment_max`
- [ ] `erfinv`

### `mlx.image`

- [ ] `rgb_to_grayscale`
- [ ] `resize`  - fix  crop_to_aspect_ratio param

### `mlx.nn`

- [ ]  `max_pool`
- [ ]  `avg_pool`
- [ ]  `conv`
- [ ]  `depthwise_conv`
- [ ]  `separable_conv`
- [ ]  `conv_transpose`
- [ ]  `ctc_loss`

### `mlx.rnn`

- [ ]  `rnn`
- [ ]  `lstm`
- [ ]  `gru`

### `mlx.linalg`

- [ ] `cholesky`
- [ ] `det`
- [ ] `eig`
- [ ] `eigh`
- [ ] `inv`
- [ ] `lu_factor`
- [ ] `norm`
- [ ] `qr`
- [ ] `solve`
- [ ] `solve_triangular`
- [ ] `svd`

lkarthee:
<details>
<summary>PyTest Output</summary>

```console
=========================================================================== test session starts ============================================================================
platform darwin -- Python 3.12.2, pytest-8.1.1, pluggy-1.4.0 -- /Users/kartheek/erlang-ws/github-ws/latest/keras/.venv/bin/python3.12
cachedir: .pytest_cache
rootdir: /Users/kartheek/erlang-ws/github-ws/latest/keras
configfile: pyproject.toml
plugins: cov-5.0.0
collected 6 items

keras/src/ops/operation_test.py::OperationTest::test_autoconfig PASSED                                                                                               [ 16%]
keras/src/ops/operation_test.py::OperationTest::test_eager_call PASSED                                                                                               [ 33%]
keras/src/ops/operation_test.py::OperationTest::test_input_conversion FAILED                                                                                         [ 50%]
keras/src/ops/operation_test.py::OperationTest::test_serialization PASSED                                                                                            [ 66%]
keras/src/ops/operation_test.py::OperationTest::test_symbolic_call PASSED                                                                                            [ 83%]
[...]

PR #19666, opened 2024-05-05
mlx - handle int64 and uint64 for scatter and cum*

lkarthee:


keras/src/backend/mlx/core.py
```diff
@@ -211,6 +209,10 @@ def vectorized_map(function, elements):
 def scatter(indices, values, shape):
     indices = convert_to_tensor(indices)
     values = convert_to_tensor(values)
+    if values.dtype == mx.int64:
```
lkarthee, line 212: Temporary fix for until scatter bug is resolved in mlx. See [mlx_backend](https://github.com/keras-team/keras/issues/19571#issue-2254529820) .

fchollet (approved):
Thanks for the PR! LGTM

PR #19679, opened 2024-05-07
Fix for rendering issues with Keras Layers API Documentation (Convolution and Pooling layers)

pranshupant:
This change adds an extra newline character after the `Input shape:` and `Output shape:` sections for docstrings with nested bullet points. This ensures the correct markdown translation on the keras.io website for the convolution and pooling layers which have these bullet points right after these section headings. 

**Screeenshot of issue for MaxPooling2D from the keras.io website:**
Screenshot from 2024-05-06 22-26-15

The markdown translation logic in the keras-io repo for docstrings requires an additional \n character. Some layers such as the ConvLSTM and GlobalAveragePooling already do this correctly in their docstrings. This change replicates the correct structure for the docstrings of the pooling and conv layers which have this rendering issue

**Screenshot of MaxPooling2D after the fix:**
Screenshot from 2024-05-06 22-35-38

### DIFF
--- a/keras/src/layers/convolutional/conv1d.py
+++ b/keras/src/layers/convolutional/conv1d.py
@@ -63,12 +63,14 @@ class Conv1D(BaseConv):
             bias after being updated by an `Optimizer`.
 
     Input shape:
+
     - If `data_format="channels_last"`:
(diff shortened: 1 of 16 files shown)


fchollet (approved):
LGTM, thanks for the fix!

PR #19681, opened 2024-05-07
Add `scan` op

james77777778:
Related to #19519

Originally, `tf.scan` stacks the output of the `f` as the final output and doesn't support the separated `carry` throughout the loop iterations.
I've reimplemented `scan` for tensorflow backend to align with the behavior of `jax.lax.scan`.

However, the current implementation requires that the output (`y`) of `f` must have the same shape/dtype as `carry`. This is not required in other backends.


fchollet (approved):
LGTM -- great job!

PR #19682, opened 2024-05-07
Add `training=False` behavior for float8-trained `Dense` and `EinsumDense`

james77777778:
Related to #19671

This PR introduces `training=False` behavior for float8-trained `Dense` and `EinsumDense` layers.

We could eliminate `amax_history` and preprocess the weights to bypass the transpose op in compiled graph, but this would make the layers unrecoverable for further training. (We want to continue training in `fitting`)

Perhaps the post-processing could be considered as a future plan.

### DIFF
--- a/keras/src/layers/core/dense.py
+++ b/keras/src/layers/core/dense.py
@@ -146,7 +146,7 @@ class Dense(Layer):
             )
         return self._kernel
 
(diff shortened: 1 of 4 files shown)


fchollet (approved):
LGTM -- thanks for the PR!

PR #19686, opened 2024-05-08
Fix dtype test for `trace`

james77777778:
Additionally, fix the formatting error


james77777778:
I have enforced the result dtype for `ops.trace` to be int32 if the input dtype is uint8 or uint16 regardless of the jax version we are using.
It's impossible to get a common dtype because we run the tests with 0.4.23 for GPU and 0.4.27 for CPU.

We can fully resolve this once we have jax>=0.4.27 for GPU tests.

fchollet (approved):
Thanks for the fix!

PR #19690, opened 2024-05-08
Fix unexpected dtype policy changes when quantization fails

james77777778:
This PR fixes unexpected dtype policy changes when quantization fails.

For example, the current codebase will output `False` in the following snippet:

```python
import numpy as np

from keras import layers

layer = layers.Embedding(10, 16)
layer.build()
x = np.random.randint(0, 9, size=(1, 3))
original_dtype_policy = layer.dtype_policy

try:
    layer.quantize("float8")  # Will fail
except:
    pass
print(original_dtype_policy == layer.dtype_policy)  # False

```

With this PR, the output will be `True`.

### DIFF
--- a/keras/src/dtype_policies/dtype_policy.py
+++ b/keras/src/dtype_policies/dtype_policy.py
@@ -266,6 +266,8 @@ class QuantizedDTypePolicy(DTypePolicy):
 
 @keras_export("keras.dtype_policies.QuantizedFloat8DTypePolicy")
 class QuantizedFloat8DTypePolicy(QuantizedDTypePolicy):
+    default_amax_history_length = 1024
+
     def __init__(self, name, amax_history_length=1024):
         super().__init__(name)
         if not isinstance(amax_history_length, int):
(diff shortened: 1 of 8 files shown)


keras/src/dtype_policies/dtype_policy.py
```diff
@@ -266,6 +266,8 @@ def _get_all_valid_policies(self):
 
 @keras_export("keras.dtype_policies.QuantizedFloat8DTypePolicy")
 class QuantizedFloat8DTypePolicy(QuantizedDTypePolicy):
+    default_amax_history_length = 1024
```
james77777778, line 269: I've added a static variable that can be utilized in `_float8_build`

keras/src/layers/core/dense.py
```diff
@@ -580,10 +572,19 @@ def quantize(self, mode):
...
         self._tracker.lock()
 
+        # Set new dtype policy
+        if not isinstance(
+            self.dtype_policy, dtype_policies.QuantizedDTypePolicy
+        ):
+            quantized_dtype = f"{mode}_from_{self.dtype_policy.name}"
+            # We set the internal `self._dtype_policy` instead of using the
+            # setter to avoid double `quantize` call
+            self._dtype_policy = dtype_policies.get(quantized_dtype)
```
james77777778, line 586: I've moved the dtype policy setting to the bottom of `quantize` to prevent unexpected dtype policy changes when  quantization fails.

fchollet (commented):
LGTM, thanks for the fix.

PR #19707, opened 2024-05-10
Fix `ops.shape()` for `tf.RaggedTensor`.

hertschuh:
With the TensorFlow backend, when passing a `RaggedTensor`, the dynamic shape returned by `tf.shape` is an object that raises an exception when a ragged dimension is accessed. We catch it and leave the dimension as `None`.


fchollet (approved):
LGTM, thank you!

PR #19719, opened 2024-05-16
Dense layer docstring typo fix

jm-willy:
**Dense layer does matmul(matrix multiplication) for tensors with rank <= 2, not dot product.** 

It can be easily checked in the code, the call method of the dense layer class has:
`x = ops.matmul(inputs, self.kernel)`
not 
`output = activation(dot(input, kernel) + bias)`
as in the docstring

It can also be checked by passing rank<=2 inputs of different [-1] dimension to dense layers with variable number of units, which is only possible if dense layers are doing matmul and not dot for rank<=2 tensors.

### DIFF
--- a/keras/src/layers/core/dense.py
+++ b/keras/src/layers/core/dense.py
@@ -17,7 +17,7 @@ class Dense(Layer):
     """Just your regular densely-connected NN layer.
 
     `Dense` implements the operation:
-    `output = activation(dot(input, kernel) + bias)`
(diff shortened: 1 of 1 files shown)


fchollet (commented):
`dot` represents the dot operator. See: https://en.wikipedia.org/wiki/Multiplication#Notation

jm-willy:
I see, however it's unambiguous to say matmul instead.

jm-willy:
It's unambiguous to put matmul instead, no confusion here.

PR #19720, opened 2024-05-16
Fix export for subclass models with multiple inputs.

hertschuh:
The export now supports subclasses of `Model` for which the `call` method takes more than one input argument. Note that it is required for the model class to implement a `build` method with a signature that matches the `call` method.


fchollet (approved):
LGTM

PR #19727, opened 2024-05-17
Add preliminary support of OpenVINO as Keras 3 backend

rkazants:
**Details:** Support OpenVINO as Keras 3 backend. This is inference-only backend. In order to switch on this, define environment variable as follows: `os.environ["KERAS_BACKEND"] = "openvino"` or use `set_backend`.

Here is an example how it works:
Install OpenVINO `pip install openvino -U`

```python
import os
import numpy as np
import keras

# build a simple network
x1 = keras.Input(shape=(2,), dtype='float32')
x2 = keras.Input(shape=(2,), dtype='float32')
add = keras.layers.Add()([x1, x2])
sigmoid = keras.activations.sigmoid(add)
model = keras.Model(inputs={'x1': x1, 'x2': x2}, outputs={'sigmoid': sigmoid})
model.summary()

# infer using OpenVINO
out = model.predict({'x1': np.array([[1.0, 2.0]], dtype=np.float32),
                     'x2': np.array([[3.0, 4.0]], dtype=np.float32)})
print('openvino out = ', out)
```

Added support for operations that allow to infer mobilenet_v3_small and bert_base_en_uncased from Keras Hub.

Example 1:
```python
# from here https://keras.io/api/keras_nlp/models/bert/bert_text_classifier/#from_preset-method
import os

import numpy as np
import keras
#import keras_hub

features = {
    "token_ids": np.ones(shape=(2, 12), dtype="int32"),
    "segment_ids": np.array([[0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 0, 0]] * 2),
    "padding_mask": np.array([[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0]] * 2),
}

# Pretrained classifier without preprocessing.
#classifier = keras_hub.models.BertTextClassifier.from_preset(
#    "bert_base_en_uncased",
#    num_classes=4,
#    preprocessor=None,
#)
#keras.saving.save_model(classifier, "bert_base_en_uncased.keras")
classifier = keras.saving.load_model("bert_base_en_uncased.keras")
#output = classifier.predict(x=features, batch_size=2)
output = classifier.predict(features)
```

Example 2:

```python
# from here https://keras.io/api/keras_cv/models/backbones/mobilenet_v3/
import os

import numpy as np
import keras
import keras_cv

rng = np.random.default_rng(23345243)

#input_data = np.random.rand(8, 224, 224, 3)
input_data = rng.uniform(0.0, 1.0, [8, 224, 224, 3]).astype(np.float32)
model = keras.saving.load_model("mobilenet_v3_small.keras")
#model = keras_cv.models.MobileNetV3Backbone.from_preset("mobilenet_v3_small",)

output = model.predict(input_data)
```


keras/src/backend/openvino/linalg.py
```diff
@@ -0,0 +1,88 @@
+import numpy as np
+import scipy.linalg as sl
+
+from keras.src.backend import standardize_dtype
+from keras.src.backend.common import dtypes
+from keras.src.backend.openvino.core import convert_to_tensor
+
+
+def cholesky(a):
+    return np.linalg.cholesky(a)
```
fchollet, line 10: All of the code in many of these files still refers to numpy. If something isn't implemented, it's better to raise `NotImplementedError`, so we can keep track of it.

fchollet (commented):
Thanks for the PR!

keras/src/ops/function.py
```diff
@@ -81,6 +81,32 @@ def __init__(self, inputs, outputs, name=None):
         self._nodes_by_depth = nodes_by_depth
         self._operations = operations
         self._operations_by_depth = operations_by_depth
+        if backend() == "openvino":
```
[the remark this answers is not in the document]
rkazants: Hi @fchollet, OpenVINO does not support eager mode for inference. We should build OV graph preliminary and run inference for the whole graph. So I decided to construct a graph in `init` and use it for `call`. Please propose how to avoid backend-specific code here due to this OV specific.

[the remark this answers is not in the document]
fchollet: Just disable what doesn't work, with a clear error message, at the level of the openvino backend. If openvino is only usable via `evaluate` and `predict`, that's ok.

Though tbh since this backend is inference only and doesn't have eager support, it sounds like maybe it should be a backend, but instead should be an export format in `model.export()`.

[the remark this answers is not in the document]
rkazants: @fchollet, please clarify. Your point is to exclude this backend-specific code and configure the backend via `_...` variables. This configuration should tell Keras 3 that OpenVINO only works in non-eager mode. Am I correct? I just want ops helpers to work only with symbolic tensors `ov::Output` or `ov::Node` instances (for graph construction). And I want this graph to be constructed only one for one device and input shape specification. Is it possible to do in Keras 3?

Or your point only to implement `export` method? It will allow only to infer using OpenVINO API after model loading and not Keras 3. Please correct me.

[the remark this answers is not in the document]
fchollet: So there are two entirely separate options:

1. Make an openvino backend. If we do this it should have roughly the same scope of functionality as the numpy backend. It should "fit" the backend format. I am not sure this is feasible if openvino has no support for eager execution.
2. Make `model.export()`/`ExportArchive` able to save a model in the openvino format (the model would be coming from either Keras + JAX, or Keras + torch, of Keras + TF). That was folks can train their models in another backend, then export it to openvino.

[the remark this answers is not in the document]
rkazants: Hi @fchollet, thanks a lot for explanation!
Our original plan has been to have OpenVINO as backend. It helps users do not override keras code and infer via OpenVINO. Different code snippets for Keras models (NLP, CV models) would be inferred by OpenVINO. `Export` option makes user to write OpenVINO code after export for these models that is less convenient, unfortunately.

I am thinkng about implementing functionality (for graph construction and compilation) in OpenVINO trainer class but I do not see how to access to Keras model nodes. I see inference is triggered via callback mechanism. I have two questions:
1. Is any option to access Keras model nodes from Trainer class?
2. Is any option to configure inference callback in Trainer class to set my custom one implemented in OpenVINO Trainer class?
3. Do you have any plan to extend Keras developer API for adding inference frameworks that do not support eager mode? I think for optimal inference non-eager mode can be more optimal than eager mode.

Thanks a lot for your help,
Roman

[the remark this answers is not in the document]
fchollet: How do you create a compiled graph in OpenVINO?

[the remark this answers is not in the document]
rkazants: Here is an example of how `compiled_model` is created:

```
import openvino.runtime.opset9 as ov
from openvino.runtime import Model, Core

# create a model with TopK
x = ov.parameter([1], name="x", dtype=np.int32)
y = ov.parameter([1], name="y", dtype=np.int32)
divide = ov.divide(x, y)
ov_model = Model([divide], [x, y], "model")

# compile the model for CPU device
core = ov.Core()
compiled_model = core.compile_model(ov_model, 'CPU')
```

At the first step, we build `ov::Model` ov_model object and compile then.
 @fchollet, may be I can set up a meeting with you on Discord (@rkazants on Discord) to discuss further steps and potential solution?

Really appreciate your time and help,
Roman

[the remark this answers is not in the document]
fchollet: And then you call the `compiled_model` on numpy data?

That sort of logic would belong in the `Trainer` class. You can try prototyping an OpenVINO Trainer that works like this.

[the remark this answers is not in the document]
rkazants: > And then you call the `compiled_model` on numpy data?

Exactly, it calls `compiled_model` on numpy data for prediction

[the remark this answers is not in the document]
fchollet: If we implement a backend like this, it won't be usable eagerly, but only through `Model.evaluate()` and `Model.predict()`. That's quite limiting. It's still feasible though. We won't be able to test it on CI, since most of our tests need eager mode, instead we'd need a new set of integration tests specifically for this backend.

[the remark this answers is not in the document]
rkazants: For inference such functionality is sufficient) Let me try to do this.

Best regards,
Roman

gbaned:
Hi @rkazants Can you please resolve the conflicts? Thank you!

rkazants:
> Hi @rkazants Can you please resolve the conflicts? Thank you!

I am working on this PR. I am trying to resolve comments.

Best regards,
Roman

gbaned:
Hi @rkazants Any update on this PR? Please. Thank you!

rkazants:
I am just back from vacation. I plan to continue from the next week.

gbaned:
Hi @rkazants Any update on this PR? Please. Thank you!

rkazants:
> Hi @rkazants Any update on this PR? Please. Thank you!

Hi @gbaned, please anticipate the update early this week. Sorry for the delay.

Thanks,
Roman

keras/src/backend/openvino/linalg.py
rkazants: Done.

keras/src/layers/layer.py
```diff
@@ -289,6 +291,9 @@ def __init__(
         self._convert_input_args = True
         # Whether to allow non-tensors as positional arguments in `call()`.
         self._allow_non_tensor_positional_args = False
+        if backend.backend() == "openvino":
```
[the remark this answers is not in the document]
rkazants: Hi @fchollet, please advice how to avoid this code. Keras common requires it to be of `KerasTensor` type during calling op-by-op. How to relax this check and allow to pass ov type object into ops wrappers?

rkazants:
@fchollet, please review. all other comments are addressed.

rkazants:
@fchollet, could you please advice how to re-format sources to pass CI check? What tool needs to use and how?

Thanks,
Roman

rkazants:
Hi @fchollet, look forward to receive your review:)
I am ready to address your comments asap these days.

Thanks in advance,
Roman

rkazants:
@fchollet, kindly reminder about review.

Thank you,
Roman

rkazants:
Hi @fchollet, any update? Please inform me if any concerns.

Thanks,
Roman

rkazants:
Hi @fchollet, @gbaned,

Could you please take a look? Let me know if any minimal scope of supported ops are required for this PR. I will add it here.

Thanks,
Roman

keras/src/layers/layer.py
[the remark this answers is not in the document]
fchollet: All you need is for `backend.is_tensor(x)` to return True for the types you consider to be tensors.

fchollet (commented):
Sorry, only seeing this now. It looks better! It seems the set of ops currently implemented is quite small, however. What kinds of end-to-end workflows are currently working? What level of op coverage do you think you can achieve?

keras/src/layers/layer.py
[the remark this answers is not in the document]
rkazants: @fchollet, I redefined it for my backend but `openvino.runtime.Output` does not have `dtype` attribute and exception is raised here: https://github.com/keras-team/keras/blob/master/keras/src/dtype_policies/dtype_policy.py#L200
We retrieve dtype using method `get_element_type`. I think `backend.standardize_dtype` should accept `x` instead of `x.dtype`.

rkazants:
> Sorry, only seeing this now. It looks better! It seems the set of ops currently implemented is quite small, however. What kinds of end-to-end workflows are currently working? What level of op coverage do you think you can achieve?

Hi @fchollet, currently we support `predict` method only for inference. `evaluate` is not supported because it requires to support loss function and run it on the same backend. That is not convenient to do with our backend because eager execution is absent.

In future, from my quick glance we will be able to cover solid number of operations because OpenVINO opset is quite well developed and support a lot of TF, PyTorch, ONNX models from different hubs (TF, ONNX, HF, PyTorch, etc,) for inference and we permanently develop it to cover TF, PyTorch, ONNX opsets.

Let me know what amount of operations I should cover in this PR to get it merged. So we will have the base OpenVINO support here and continue ops coverage support in future PRs. For the base OpenVINO support, I propose to support operations required by BERT model from Keras Hub, for example, and demonstrate that it works. Will it be sufficient for merge?

In the description of this PR, I demonstrated very simple model inference. Please check.

fchollet:
> Let me know what amount of operations I should cover in this PR to get it merged. So we will have the base OpenVINO support here and continue ops coverage support in future PRs. For the base OpenVINO support, I propose to support operations required by BERT model without preprocessor from Keras Hub, for example, and demonstrate that it works. Example below. Will it be sufficient for merge?

We will have to run only a very similar subset of unit tests on CI for OpenVINO since we cannot skip every test that isn't passing (too many). So we would have to create basically an OpenVINO-specific integration test that checks precisely the end-to-end workflows that are intended to be supported.

Are you going to add support for CV workflows? I would recommend also having support for conv/pooling ops.

rkazants:
> We will have to run only a very similar subset of unit tests on CI for OpenVINO since we cannot skip every test that isn't passing (too many). So we would have to create basically an OpenVINO-specific integration test that checks precisely the end-to-end workflows that are intended to be supported.
> 
> Are you going to add support for CV workflows? I would recommend also having support for conv/pooling ops.

Yes, I will add separate integration tests for OpenVINO to cover currently supported ops. For this PR, let me enable MobileNetV3 from KerasCV that will require to support conv/pooling operations.

Summarizing it, my to-do list for this PR looks as follows:

1. Enable BertClassifier model from KerasNLP
2. Enable MobileNet model from KerasCV
3. Add integration tests for OpenVINO to cover supported ops

@fchollet, will it be fine and sufficient for this PR merge?

Thanks a lot,
Roman

fchollet:
> @fchollet, will it be fine and sufficient for this PR merge?

Yes, that sounds good to me!

keras/src/layers/layer.py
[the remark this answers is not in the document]
rkazants: Removed

rkazants:
Hi @fchollet,

I added support for Keras Hub Bert and MobileNet models and removed openvino-specific code from `layer.py`. Please do preliminary code-review for it. In the meantime I am working on tests and GHA configuration for them.

Best regards,
Roman

.github/workflows/actions.yml
```diff
@@ -84,11 +84,18 @@ jobs:
         run: |
           pytest integration_tests/torch_workflow_test.py
           python integration_tests/torch_custom_fit_test.py
+      - name: Test OpenVINO-specific integrations
+        if: ${{ matrix.backend == 'openvino' }}
+        run: |
+          python integration_tests/import_test.py
```
rkazants, line 90:
```suggestion
```

rkazants:
@fchollet, please take a look at this PR.

Thanks,
Roman

keras/src/layers/attention/attention_test.py
```diff
@@ -146,6 +150,7 @@ def test_attention_calculate_score_mask_no_causal_no_vmask(self):
             "are used",
         )
 
+    @pytest.mark.openvino_backend
```
rkazants, line 153:
```suggestion
```

keras/src/layers/attention/additive_attention_test.py
```diff
@@ -70,6 +72,7 @@ def test_attention_with_mask(self):
         self.assertAllClose(output, [[[1.0, 1.0], [0.0, 0.0]]])
         self.assertAllClose(scores, [[[1.0, 0.0], [1.0, 0.0]]])
 
+    @pytest.mark.openvino_backend
```
rkazants, line 75:
```suggestion
```

keras/src/layers/attention/grouped_query_attention_test.py
```diff
@@ -202,6 +205,7 @@ def test_shape_mismatch_error(self, query_shape, value_shape, key_shape):
         with self.assertRaisesRegex(ValueError, r"must be equal"):
             layer.compute_output_shape(query_shape, value_shape, key_shape)
 
+    @pytest.mark.openvino_backend
```
rkazants, line 208:
```suggestion
```

rkazants:
@fchollet, please take a look.

Best regards,
Roman

rkazants:
Hi @fchollet, please take a look. Above 300 tests are now switched on for openvino backend.

Best regards,
Roman

rkazants:
Hi @fchollet, kindly reminder, please review. Now we have 400+ tests running for openvino backend.

Best regards,
Roman

integration_tests/openvino/basic_full_flow.py
```diff
@@ -0,0 +1,24 @@
+import numpy as np
```
fchollet, line 1: There should be much more extensive integration testing -- covering each of the supported use cases (e.g. CNNs, Transformers for LLMs, RNNs, etc.)

keras/src/ops/core_test.py
```diff
@@ -607,6 +608,7 @@ def test_stop_gradient_return(self):
         y = ops.stop_gradient(x)
         self.assertAllClose(x, y)
 
+    @pytest.mark.openvino_backend
```
fchollet, line 611: It seems impractical to mark individual tests (at the method level as whitelisted for the openvino backend).

In general you should have fewer than ~30 backend-specific annotations across the unit tests codebase.

Here's what you could do:

1. Remove all annotations, not running any unit tests on CI, but have very extensive integration testing.
2. Only do opt-out (rather than opt-in) annotations for a small number of unit tests (e.g. at the class level rather than at the method level).

fchollet (commented):
Thanks for the update!

integration_tests/openvino/basic_full_flow.py
rkazants: Now we are able to check each operation support using existing tests. So no need to implement separate set of tests for openvino.
I am managed to adopt the existing tests despite there is no eager execution for openvino. For this, I run inference for `ov::Output` resulting node in `convert_tensor_to_numpy` and it provides the requested output data computed using openvino.

.github/workflows/actions.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.9]
-        backend: [tensorflow, jax, torch, numpy]
+        backend: [tensorflow, jax, torch, numpy, openvino]
```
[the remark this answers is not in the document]
rkazants: I removed `integration_tests/openvino` and use existing net for import_test.py

keras/src/ops/core_test.py
rkazants: Hi @fchollet,

Am I correct that instead I should use this mark for all test suites that are not passing now?

```
@pytest.mark.skipif(
    backend.backend() != "openvino",
    reason="OpenVINO backend does not support `some op`",
)
```

And should I remove `@pytest.mark.openvino_backend` in all tests?

Best regards,
Roman

fchollet: > Am I correct that instead I should use this mark for all test suites that are not passing now?

You should do this *if you enable the regular pytest CI for OpenVino*, which you should only do *if a small number of tests are failing*. If lots of tests are failing, you may be better off not enabling regular CI and instead rely on integration tests only.

Also, we might want to introduce single-line shortcuts to skip something with a given backend.

fchollet:
When I say "integration tests", I was thinking that since the OpenVino backend is only supposed to support a specific set of use case, you could tests for these specific use cases in `integration_tests/openvino`.

rkazants:
> When I say "integration tests", I was thinking that since the OpenVino backend is only supposed to support a specific set of use case, you could tests for these specific use cases in `integration_tests/openvino`.

Hi @fchollet,

I see a problem to implement OpenVINO specific integration tests:
1. they should be constantly extended when we add support for new keras operations;
2. we cannot cover multiple variants of input types (keras variable, python value, numpy array, and different data types) for each layer
3. they will not be maintained by keras maintainers while Keras API design and concept are developing. So I want OV backend to be on trend and tested against fresh tests.

In order to solve these problems, I propose to re-use existing pytest tests from keras directory and step-by-step switch them on while we are adding new operations support. I managed to re-use multiple such tests as you can see the current state of PR.
What we can do:
1. for openvino backend, extend a list of ignore test files (https://github.com/keras-team/keras/blob/master/.github/workflows/actions.yml#L89). It will have a custom list of files that we will reduced step-by-step. In addition, inside the test file we can use `skipif` if there are a few tests failing. So we will use combination of ignore file list and `skipif`.
I think the most convenient and reliable way for backend development is usage of such generic tests. @fchollet, how do you think and will you allow to proceed with this way instead of OV specific integration tests (concepts of such tests is quite unclear and weak).

Best regards,
Roman

keras/src/ops/core_test.py
rkazants: responded below, please see

fchollet:
> for openvino backend, extend a list of ignore test files 

Sounds good -- so the changes would be limited to an OpenVino specific config file that lists the test files to exclude?

rkazants:
> Sounds good -- so the changes would be limited to an OpenVino specific config file that lists the test files to exclude?

Yes, I will try to create openvino specific config files with a list of excluded test files.

rkazants:
Hi @fchollet,

Just in case let me ask you: do you know I receive failures "NotImplementedError: `fit` is not supported with openvino backend" in my tests despite that I have updated `requires_trainable_backend` with openvino in `conftest.py`?

```
def pytest_collection_modifyitems(config, items):
    requires_trainable_backend = pytest.mark.skipif(
        backend() == "numpy" or backend() == "openvino",
        reason="Trainer not implemented for NumPy and OpenVINO backend.",
    )
    for item in items:
        if "requires_trainable_backend" in item.keywords:
            item.add_marker(requires_trainable_backend)
```

It looks that tests requiring trainable backend still are running for openvino.

Best regards,
Roman

rkazants:
Hi @fchollet,

PR is ready for review:
- Bert and MobileNet models from Keras Hub can be inferred using openvino backend
- Tests are run for openvino backend in GHA

Best regards,
Roman

rkazants:
Hi @fchollet, friendly reminder. All is done, please review.

Thanks,
Roman

fchollet:
I am at NeurIPS, will review when available.  I would also recommend a review from @jeffcarp and team.

rkazants:
Hi @fchollet, @jeffcarp,

These nearest days I have a high bandwidth for OpenVINO integration activity so I can apply your comments and questions quickly. Looking forward to receive your reviews. Sorry for constant reminder.
Thanks a lot for your effort and time!

Best regards,
Roman

keras/src/backend/openvino/trainer.py
```diff
@@ -0,0 +1,272 @@
...
+        class_weight=None,
+        sample_weight=None,
+        initial_epoch=0,
+        steps_per_epoch=None,
+        validation_steps=None,
+        validation_batch_size=None,
+        validation_freq=1,
+    ):
+        raise NotImplementedError(
+            "`fit` is not supported with openvino backend"
```
jeffcarp, line 169: This seems like an important thing to support, can you attach a TODO with an issue number to this line?

requirements.txt
```diff
@@ -13,5 +13,8 @@ torchvision>=0.16.0
 jax[cpu]
 flax
 
+# OpenVINO
+openvino
```
jeffcarp, line 17: Sorry I'm not super familiar, are these user-facing or development dependencies? Want to confirm this won't add another dep for users unless they opt in.

jeffcarp (commented):
Thanks Roman! Is there any way to try this without the hardware?

keras/src/backend/openvino/trainer.py
rkazants: Hi @jeffcarp, `openvino` backend is only inference backend like numpy one. So no `fit` method is planned to support.
The plan is user can train his model using PT/JAX/TF and use openvino for inference right away without code change.

requirements.txt
rkazants: @fchollet, can you please comment?
AFAIK, it used for testing preparation.

rkazants:
> Thanks Roman! Is there any way to try this without the hardware?

Sure, you can try this out on your laptop with x86, ARM CPU (Win, Linux, MacOS) or Intel GPU, NPU hardware? Just install OpenVINO by `pip install -U openvino`

integration_tests/numerical_test.py
```diff
@@ -137,6 +137,7 @@ def numerical_test():
 
 
 if __name__ == "__main__":
-    keras.utils.set_random_seed(1337)
-    tf_keras.utils.set_random_seed(1337)
-    numerical_test()
+    if keras.backend.backend() != "openvino":
```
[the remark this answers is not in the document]
rkazants: Sure, I will add comment. The reason of skipping this for `openvino` is that numerical test requires trainable backend.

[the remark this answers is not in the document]
rkazants: Added a comment

requirements.txt
fchollet: `requirements.txt` is dev-facing not user facing. User facing deps are listed in pyproject.toml.

fchollet: If there is never a need to import openvino when testing another backend, we should create an openvino specific requirements file (like what we have for jax/torch/tf with cuda) and install that for the openvino CI run only.

keras/src/backend/openvino/trainer.py
jeffcarp: Thanks for clarifying, I was not aware! Sounds good.

requirements.txt
jeffcarp: Sounds good, thanks

rkazants: @fchollet, @jeffcarp, I added custom requirements file for openvino.

rkazants:
Hi @fchollet, @jeffcarp, all comments were resolved so far. Please let me know what else is needed for merging this PR.

Thanks a lot,
Roman

keras/src/backend/openvino/trainer.py
```diff
@@ -0,0 +1,272 @@
...
+
+        if self.steps_per_execution > 1:
+            predict_step = multi_predict_steps
+        else:
+            predict_step = one_predict_step
+
+        self.predict_function = predict_step
+
+    def fit(
+        self,
```
jeffcarp, line 150: I assume all these args can be replaced with `*, **` since they're unused

jeffcarp (approved):
Thanks Roman! LGTM from my side. Please also get an LGTM from Francois before proceeding.

jeffcarp:
> sure, you can try this out on your laptop with x86

Thanks! I was able to confirm running this branch locally.

keras/src/backend/openvino/trainer.py
rkazants: It is done with the same style as for numpy backend to be aligned so may be it is better to leave unchanged.

jeffcarp: Thanks sounds reasonable

keras/src/backend/common/variables_test.py
```diff
@@ -16,6 +17,11 @@
 from keras.src.testing.test_utils import named_product
 
 
+@skip_if_backend(
+    "openvino",
+    "AttributeError: 'openvino._pyopenvino.Tensor' "
```
fchollet, line 22: ?

keras/src/initializers/constant_initializers_test.py
```diff
@@ -57,6 +58,7 @@ def test_constant_initializer_array_value(self):
 
         self.run_class_serialization_test(initializer)
 
+    @skip_if_backend("openvino", "openvino backend does not support `eye`")
```
fchollet, line 61: What's the plan/timeline for adding support for the various missing ops?

rkazants: I think the plan will be to implement these missing ops for the whole next year. I will create good-first-issues for each operation with a good description and example for some other operation. These GFIs will be performed by open-source community and I will review it these contributions. From my side, I will do all important fixes (if any needed) in the core part of openvino backend.

keras/src/backend/common/variables_test.py
rkazants: Let me try to fix it in this PR

fchollet: Ok, I'll wait for that before merging then. Thanks!

rkazants: Fixed and removed this `skip_if_backend`.

rkazants:
@fchollet, all is done. Please review it again.

Thank you very much,
Roman

fchollet (approved):
LGTM -- thank you for the contribution! 👍 

Should we expect an announcement from Intel here? How should we communicate this to the developer community?

rkazants:
> LGTM -- thank you for the contribution! 👍
> 
> Should we expect an announcement from Intel here? How should we communicate this to the developer community?

Hi François ([fchollet](https://github.com/fchollet)),

Really thanks for your big help in reviewing this PR.
From OpenVINO side, so plan to announce about this feature in the upcoming OpenVINO 2025.0 release in January. It will be declared as experimental feature. We will add this into OpenVINO release notes and add into documentation "how to deploy Keras 3 models using OpenVINO" (https://docs.openvino.ai/2024/openvino-workflow.html), add OpenVINO notebook to showcase Keras 3 + OpenVINO. Once we reach close parity in terms of operation coverage and reach quite the same test passrate,  we will remove "experimental". I will also make a post in LinkedIn and our Discord to invite open-source community to contribute to this very soon.

From Keras side, I will update Keras documentation as well and declare it as experimental feature. Please let me know how to proceed with Keras documentation. Also, can you add this into Keras release notes of the upcoming Keras release from master? Please let me know what will be Keras release number from master that will include this changes?

Best regards,
Roman

fchollet:
We will include the new backend in the release notes for the next release. Please feel free to propose edits to the docs website via a PR to `keras-team/keras-io`!

rkazants:
> Should probably update the README to mention openvino in the same places tensorflow/jax/pytorch are currently mentioned. At the moment, it isn't obvious how to use this backend (other than reading the release notes). Also, the release notes say "open_vino" but this PR says "openvino", not sure which is correct.

Let me create PR with README update today. "openvino" is correct one.

Best regards,
Roman

PR #19728, opened 2024-05-17
Tweaks for serialization of DTypePolicy in ops/layers.

hertschuh:
- Subclasses of `Operation` / `Layer` which override `__init__` and use the `dtype` parameter don't expect a `dict`. We deserialize the `DTypePolicy` in `from_config`.
- The auto `get_config` feature would break when a `DTypePolicy` was passed to the constructor of any `Operation` or `Layer` subclass not implementing `get_config`.


fchollet (approved):
LGTM, thanks for the PR!

PR #19729, opened 2024-05-18
Add more tests for `DTypePolicy`

james77777778:
These tests should prevent the future breakages.

Also, add support for `DTypePolicy` in `keras.dtype_policies.get`


fchollet (approved):
LGTM, thank you

Issue #19191, opened 2024-02-16
StringLookup Layer not working with sequences when output mode is one-hot


pinsondg:
I am reviving this error when trying to use the 'one-hot' output mode on a string lookup layer.
`When output_mode is not 'int', maximum supported output rank is 2. Received output_mode one_hot and input shape (None, 30), which would result in output rank 3.`

I am trying to input a sequence of strings using a `tf.data.Dataset`. The length of the sequence is 30. Here is what the element spec looks like for my dataset object.
```python
((TensorSpec(shape=(None, 30, 25), dtype=tf.float64, name=None),
  TensorSpec(shape=(None, 30), dtype=tf.string, name=None),
  TensorSpec(shape=(None, 30), dtype=tf.string, name=None),
  TensorSpec(shape=(None, 30), dtype=tf.int32, name=None),
  TensorSpec(shape=(None, 30), dtype=tf.string, name=None),
  TensorSpec(shape=(None, 30), dtype=tf.int32, name=None),
  TensorSpec(shape=(None, 30), dtype=tf.string, name=None),
  TensorSpec(shape=(None, 30), dtype=tf.string, name=None),
  TensorSpec(shape=(None, 30), dtype=tf.string, name=None),
  TensorSpec(shape=(None, 28), dtype=tf.float64, name=None),
  TensorSpec(shape=(None,), dtype=tf.string, name=None),
  TensorSpec(shape=(None,), dtype=tf.string, name=None),
  TensorSpec(shape=(None,), dtype=tf.string, name=None),
  TensorSpec(shape=(None,), dtype=tf.string, name=None)),
 TensorSpec(shape=(None,), dtype=tf.int32, name=None))
```

I the StringLookup layer is being passed as the first layer in my model. It seems like the batch dimension is being taken into account in the rank calculation, which is causing my model to fail. I also notice this comment in the[ code](https://github.com/keras-team/keras/blob/7ce3d62af7cc6959fc5a5841cfe17043dfcb8615/keras/utils/tf_utils.py#L75). Is this still on the docket to remove this rank restriction?

sachinprasadhs:
Hi, Could you please provide the sample reproducible code to replicate the reported behavior. Thanks!

pinsondg:
Hi @sachinprasadhs here is a basic example where this problem can be reproduced.
```
import tensorflow as tf

# Dataset of a sequence of strings
[...]

PR #19742, opened 2024-05-22
Consolidate implementations of category encoding and remove rank limit.

hertschuh:
There were three similar implementations of "one_hot", "multi_hot", "int" and "count" encodings:
- in `category_encoding.py` (a cross-backend implementation with no rank limit)
- in `tf_utils.py` (a TensorFlow specific implementation with a rank limit of 2)
- in `numerical_utils.py` (a cross-backend implementation with no rank limit but no sparse support and weights support)

They are now combined in `numerical_utils.py`. The implementation is cross-backend, supports sparse, weights and has no rank limitation.

Also:
- removed `tf_utils.ensure_tensor` in favor of `tf_backend.convert_to_tensor`.

Fixes https://github.com/keras-team/keras/issues/19191


fchollet (approved):
Thank you for the extensive fix. LGTM

PR #19750, opened 2024-05-23
Remove `convert_to_numpy` in confusion metrics and fix `ops.nonzero` …

hertschuh:
…for torch.

- Removed the use of `convert_to_numpy` confusion metrics as it prevents jit compilation.
- Fixed Torch implementation of `ops.nonzero`, it was returning a tuple of one tensor instead of just the tensor.

The `convert_to_numpy` call was presumably a workaround for the second issue.


fchollet (approved):
Thanks for the fix -- surprising bug.

PR #19751, opened 2024-05-23
Re-introduce `tf_utils.ensure_tensor` in preprocessing layers.

hertschuh:
It turns out `tf_backend.convert_to_tensor` does not support on-GPU Torch tensors. This also removes the need for some late import.


fchollet (approved):
Thanks for the fix!

PR #19752, opened 2024-05-24
Fix `dtype` serialization

james77777778:
We should use `dtype_policies.serialize` to serialize the `DTypePolicy` in `Operation`. A simple test has been added to verify this.
A note has been added in `from_config` to indicate that when overriding `from_config`, users should deserialize the `dtype` if needed.

### DIFF
--- a/keras/src/dtype_policies/__init__.py
+++ b/keras/src/dtype_policies/__init__.py
@@ -87,9 +87,7 @@ def get(identifier):
 
     if identifier is None:
         return dtype_policy.dtype_policy()
-    if isinstance(
-        identifier, (DTypePolicy, FloatDTypePolicy, QuantizedDTypePolicy)
-    ):
+    if isinstance(identifier, DTypePolicy):
         return identifier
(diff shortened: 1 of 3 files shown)


james77777778:
Apart from this PR, is it a good idea to have an api to override the `dtype` during deserialization using global dtype policy?

This should help with the following:
- Training and saving a mixed precision model but wanting to load it with `"float32"`
- Adding flexibility to the quantized dtype policy

james77777778:
> Thanks for the PR. In order to ensure backwards compatibility, we need to make sure that by default, for standard cases (e.g. float32), the `"dtype"` entry in the config is a string. Is it the case now?

I've updated the code to ensure that, in the config, `dtype` will be a plain string if it is a `FloatDTypePolicy`. Conversely, it will be a serialized config if it is a `QuantizedDTypePolicy`.
The corresponding tests have been added.

fchollet (approved):
LGTM, thank you!

PR #19763, opened 2024-05-26
mlx - linalg ops

lkarthee:
Add implementation for ops - qr, cholesky, svd, inv


fchollet (approved):
Thank you for the PR!

PR #19764, opened 2024-05-27
StringLookup layer can now take tf.SparseTensors as input.

hertschuh:


fchollet (approved):
LGTM -- thanks for the PR!

PR #19777, opened 2024-05-30
Refactor `ops.image.*` and the tests

james77777778:
This PR is quite large; sorry about that.

## Polish the APIs
- Use `images` as the argument instead of mixing `image` and `images`
- Align the signature of `map_coordinate` with `jax.scipy.ndimage.map_coordinates`
- Ensure the same error msg in function calls and symbolic inference for all backends
- Update docstrings
- EDITED: Fix the signature of `pad_images` and `crop_images`

## Backend Specific
- Simplify the logic of `rgb_to_grayscale` in jax, numpy and tensorflow (without performance regression)
- Add `torchvision` to `LazyModule`

## Unit Tests
- Test both "channels_last" and "channels_first" for all `ops.image.*`
- Improve test coverage


fchollet (approved):
LGTM, thanks for the fixes. `inputs` and `images` do work better.

Issue #19724, opened 2024-05-17
CategoryEncoding generate output with wrong shape when output_mode='one_hot' and the last dim is 1


AGFACBNNR:
As shown in this [colab](https://colab.research.google.com/drive/1dnw83wHCkcSYUJw_cso-zrU4rHzokkvY?usp=sharing),  when output_mode='one_hot' and the last dim is 1, as described  in Keras 3 Doc,

>  If the last dimension is size 1, will encode on that dimension.

The compute_output_shape() generate a correct static shape, while the output actually is generated by appending a new dimension for the encoded output, which is different from the documents.

Currently the output is:
![1715927195410](https://github.com/keras-team/keras/assets/20224019/093d1d73-ebcf-45a2-95bc-9640249594c1)

I've tried it on tensorflow and jax backend and got the same result.
The results in line 1 of output shows the symptom I described above.

SuryanarayanaY:
Hi @AGFACBNNR ,

I have replicated the issue and observed the discrepancy wrt actual output shape and shape from `compute_output_shape` method. Attached [gist](https://colab.sandbox.google.com/gist/SuryanarayanaY/0f7ab4c07269065415fc66742f7cbf32/19724_categoryencodingshape.ipynb#scrollTo=UGr03CvLuhgi) for reference.

Need to dig more and comeback. Thanks!

PR #19786, opened 2024-06-01
Fix `one_hot` generates output with an additional dimension in TensorFlow backend (#19724)

Grvzard:
Make it consistent with the behavior of the NumPy backend
refer to:
https://github.com/keras-team/keras/blob/master/keras/src/backend/numpy/nn.py#L445


fchollet (commented):
Thanks for the PR! Please add a unit test for this case.

Grvzard:
problems here, need to reconsider

PR #19790, opened 2024-06-02
Fix `CategoryEncoding` generates output with 1 additional dimension in all backends except NumPy (#19724)

Grvzard:
Make the behavior consistent across all backends.

### DIFF
--- a/keras/src/backend/numpy/nn.py
+++ b/keras/src/backend/numpy/nn.py
@@ -448,10 +448,6 @@ def one_hot(x, num_classes, axis=-1, dtype="float32", sparse=False):
     x = convert_to_tensor(x)
     input_shape = x.shape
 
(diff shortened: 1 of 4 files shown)


hertschuh:
@Grvzard ,

Thank you for catching and fixing this bug. This all looks correct. 

The fact that there was a discrepancy with one backend shows that we're missing a unit test in `nn_test.py`. Can you please add an input covering this case of a last dimension of 1 here: https://github.com/keras-team/keras/blob/master/keras/src/ops/nn_test.py#L1652

Thanks!

Grvzard:
The original one_hot in numpy will fail on this test

hertschuh (approved):
Great, thank you!

Grvzard:
Fixed #19724

PR #19807, opened 2024-06-05
Fix `LayerNormalization.get_config`

### LINKED ISSUE #19806: Can't load LayerNormalization with RMS scaling
I can't load a model with a LayerNormalization layer with RMS scaling in Keras 3.3.3. I think this is a bug.

Here is a script to reproduce the issue:

```
import keras

model = keras.Sequential([keras.layers.LayerNormalization(rms_scaling=True)])
model.build((100, 10))
model.summary()

keras.models.save_model(model, '/tmp/test.keras')
model2 = keras.models.load_model('/tmp/test.keras')
# Fails with:
# Layer 'layer_normalization_5' expected 2 variables, but received 1 variables during loading. Expected: ['gamma', 'beta']
```

Grvzard:
Fixed #19806

### DIFF
--- a/keras/src/layers/normalization/layer_normalization.py
+++ b/keras/src/layers/normalization/layer_normalization.py
@@ -243,6 +243,7 @@ class LayerNormalization(Layer):
             "epsilon": self.epsilon,
             "center": self.center,
             "scale": self.scale,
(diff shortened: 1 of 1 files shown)


fchollet (approved):
LGTM, thanks for the fix!

PR #19814, opened 2024-06-06
Faster in_top_k implementation for Jax backend

Hilly12:
Microbenchmarks are here [1]. Seems to be faster on CPU, GPU, TPU. Also XLA is sometimes able to dedup everything before k is referenced when it's called multiple times with different ks. This results in nice gains for retrieval applications with multiple top k metrics.

[1] https://gist.github.com/Hilly12/85460873d9786924159f2377f320df48


fchollet (approved):

PR #19822, opened 2024-06-10
fixing dmtree.is_nested() and parameterized tree test

haohuanw:
i encountered an issue where below code snippet doesn't work and later found that the issue is specific to dmtree:

```
import keras
input1 = keras.layers.Input(shape=(32,))
input2 = keras.layers.Input(shape=(32,))
added = keras.layers.Add()([input1, input2])
```

throws exception

```
  File "/home/haohuanw/Documents/keras/keras/src/layers/merging/base_merge.py", line 102, in build
    if not isinstance(input_shape[0], (tuple, list)):
TypeError: 'NoneType' object is not subscriptable
```

then found the issue was due to a small bug in `dmtree::is_nested()`. i then also expand to make sure both tree implementation is tested in ci. it also discovered 2 subtle differences between optree and dmtree.

additional help needed: is `requirements-common.txt` the right place for adding the test deps?

### DIFF
--- a/keras/src/tree/dmtree_impl.py
+++ b/keras/src/tree/dmtree_impl.py
@@ -6,7 +6,7 @@ def register_tree_node_class(cls):
 
 
 def is_nested(structure):
-    dmtree.is_nested(structure)
+    return dmtree.is_nested(structure)
 
 
 def traverse(func, structure, top_down=True):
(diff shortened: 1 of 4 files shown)


fchollet (approved):
Nice, thanks for the fix!

PR #19830, opened 2024-06-10
`Metric.variables` is now recursive.

hertschuh:
This allows it to surface variables from metrics nested at any depth.

Previously, metrics within metrics within metrics would not have their variables tracked in JAX, causing them to not be updated.


fchollet (approved):
LGTM, thanks for the PR!

Issue #19831, opened 2024-06-10
Type Error in keras.utils.get_file()


RParkerE:
**System**: Windows 10
**Python**: 3.9.19
**Tensorflow/Keras**: 2.10

Usage:
`dataset_path = keras.utils.get_file("auto-mpg.data", "https://archive.ics.uci.edu/ml/machine-learning-databases/auto-mpg/auto-mpg.data")`

Error: 
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[2], line 2
      1 # Datapath to import auto-mpg data
----> 2 dataset_path = keras.utils.get_file("auto-mpg.data", "https://archive.ics.uci.edu/ml/machine-learning-databases/auto-mpg/auto-mpg.data")

File ~\anaconda3\envs\AVTech\lib\site-packages\keras\utils\data_utils.py:296, in get_file(fname, origin, untar, md5_hash, file_hash, cache_subdir, hash_algorithm, extract, archive_format, cache_dir)
    294 try:
    295     try:
--> 296         urlretrieve(origin, fpath, DLProgbar())
    297     except urllib.error.HTTPError as e:
    298         raise Exception(error_msg.format(origin, e.code, e.msg))

File ~\anaconda3\envs\AVTech\lib\site-packages\keras\utils\data_utils.py:86, in urlretrieve(url, filename, reporthook, data)
     84 response = urlopen(url, data)
     85 with open(filename, "wb") as fd:
---> 86     for chunk in chunk_read(response, reporthook=reporthook):
     87         fd.write(chunk)

File ~\anaconda3\envs\AVTech\lib\site-packages\keras\utils\data_utils.py:78, in urlretrieve.<locals>.chunk_read(response, chunk_size, reporthook)
     76 count += 1
     77 if reporthook is not None:
---> 78     reporthook(count, chunk_size, total_size)
     79 if chunk:
     80     yield chunk

File ~\anaconda3\envs\AVTech\lib\site-packages\keras\utils\data_utils.py:287, in get_file.<locals>.DLProgbar.__call__(self, block_num, block_size, total_size)
    285     self.progbar = Progbar(total_size)
    286 current = block_num * block_size
--> 287 if current < total_size:
    288     self.progbar.update(current)
    289 elif not self.finished:

TypeError: '<' not supported between instances of 'int' and 'NoneType'
```

RParkerE:
[...]

PR #19833, opened 2024-06-11
Fix `get_file` when the HTTP response has no `Content-Length` header

Grvzard:
refer to:
https://github.com/python/cpython/blob/48f9d3e3faec5faaa4f7c9849fecd27eae4da213/Lib/urllib/request.py#L259-L264

Http response without a `Content-Length` header will cause `total_size = -1`

https://github.com/keras-team/keras/blob/efaaf85e19113400f23462cbafcef433cd95ad9c/keras/src/utils/file_utils.py#L279-L286

Line 280 is bypassed, so line 285 will be executed instead. `current` will be set to `self.progbar.target`, which is `None`.

### DIFF
--- a/keras/src/utils/file_utils.py
+++ b/keras/src/utils/file_utils.py
@@ -270,9 +270,9 @@ def get_file(
                 self.finished = False
 
             def __call__(self, block_num, block_size, total_size):
+                if total_size == -1:
+                    total_size = None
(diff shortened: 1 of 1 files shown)


Grvzard:
Fixes #19831

fchollet (approved):

PR #19834, opened 2024-06-11
Add `ops.switch`

james77777778:
This op should be useful and functions the same as `jax.lax.switch`


fchollet (approved):
LGTM -- thanks for the useful contribution.

PR #19842, opened 2024-06-11
Use `absl.testing.parameterized` for `tree_test.py`.

hertschuh:
For consistency, use `absl.testing.parameterized` instead of `parameterized` for `tree_test.py` since that is used for all other tests.

It's one less dependency. It also says `optree` or `dmtree` in each test name.


fchollet (approved):
Thanks!

PR #19843, opened 2024-06-12
doc: `ops.slice`

### LINKED ISSUE #19841: Keras.ops.slice documentation error
Hi, I believe there is a error in the current keras.ops documentation for the `slice` function [here](https://keras.io/api/ops/core/#slice-function). I believe the example code should look like:

```
inputs = np.zeros((5, 5))
start_indices = np.array([3, 3])
shape = np.array([2, 2])
inputs = keras.ops.slice(inputs, start_indices, shape)
```

Grvzard:
Closes #19841


fchollet (approved):
LGTM, thanks

PR #19849, opened 2024-06-13
Add `custom_gradient` (a pass-through) for the numpy backend

james77777778:
It only acts as a pass-through to support the forward pass.
This might be useful if we want to perform inference for the quantized model using the numpy backend.


fchollet (approved):
Sure, sounds good.

PR #19862, opened 2024-06-16
Fix code formatting error in docstring

taeminhong:
This pull request is a minor documentation fix, so I don't create an issue for this.

For some reason, there is an unclosed backtick the in layer.HashedClossing docstring and this breaks the code formatting of the online documentation.

Before fix:
Screen Shot 2024-06-16 at 8 42 05 PM

After fix:
Screen Shot 2024-06-16 at 8 43 10 PM


fchollet (approved):
Thanks for the fix!

PR #19864, opened 2024-06-17
Stop using import * from toplevel init

### LINKED ISSUE #19779: Keras `__init__.py` structure isn't readable by a static type checker
In keras 3.2.1, if you import say keras.models, you get no errors in VS code:

image

In keras 3.3.1, this same code shows an error:

image

I believe the reason for this was the refactoring of the `__init__.py` that happened in this commit:

https://github.com/keras-team/keras/commit/04891e89da87c2a433beb12ff7dad59403c71671

The code added in this commit dynamically adds the path to `api` to the sys path. This isn't something a static type checker can handle. 

Could this change be reverted? Or perhaps marked with a `if TYPECHECKING` special case?

mattdangerw:
It breaks typechecking. Instead we will import every symbol from /api into the toplevel, extending our api_gen.sh script to do this automatically.

I also switched the toplevel __init__.py in the repo to be the actual source of truth for the file, with just the import section being auto-generated. Keeping a python file as a string in another python file is clunky.

Fixes #19779


mattdangerw:
@rchiodo feel free to take a look as well!

rchiodo:
Thanks for asking me to take a look. It doesn't seem to be working for some reason in Pylance though. I believe I should be able to just copy this new `__init__.py` over my site-packages/keras/__init__.py? We're still not finding `keras.models` with this `__init__.py`

image

This might be our fault though. I'm debugging it now.

fchollet (approved):
LGTM, thanks for the fix!

rchiodo:
I don't think this is going to work. I can't find a way to _skip_ a module in the typechecker. 

According to this it sounds like it should work:
https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface

But that seems to only be for symbols, not modules.

@erictraut would be the person to ask. He knows a lot more about this than I do.


Eric, 

keras has a folder layout like so:

```
site-packages
  - keras
     - api
         - models 
```

And keras would like a user to be able to do this:

``` python
# Skip having to import keras.api first
from keras.models import Sequential
```

But that doesn't work even if the `__init__.py` reexports the models package. Is this just not supported? Essentially _skipping_ a package path?

It does work at runtime by changing the sys.path to find it, but we can't seem to model this behavior in Pyright/Pylance.

bschnurr:
@mattdangerw  Question. What is the reason behind the 'api' folder? is just a development thing?  Couldn't you copy the folders to the expected location on packaging?

```
so this current folder structure
site-packages
  - keras
     - api
         - models 

becomes what the user expects?
site-packages
  - keras
       - models 
 ```

mattdangerw:
@bschnurr our goals are...

- Explicit api surface, don't have everything private/public mixed together. We have an `keras_export` annotation that controls our public api.
- "Browsable codebase." E.g [jax](https://github.com/google/jax/tree/main/jax/) has a nested directory of just init files, with the good stuff hidden away in a `_src` directory. We didn't find that very readable. We wanted to avoid a ghost town of empty directories with init files.
- `pip install -e` works, `pip install git+ssh://...` works. We used to do a heavy rewrite of our source tree during packaging, but this led to a lot of confusion where people would run the non-packaged version we a slightly different API surface, as well as a more difficult time syncing with our google mono repo.

Is there a better way to achieve that? We had thought messing with the `__path__` was not that uncommon, e.g. old school namespace packages, but maybe that was a faulty assumption.

bschnurr:
Thanks Matt.

Currently we don't have support for editing '__path__'

For controlling the public API see https://microsoft.github.io/pyright/#/typed-libraries?id=library-interface

there are two options in `__init__.py`
 using the form `from X import A as A`

or adding symbols to `__all__`

PR #19887, opened 2024-06-20
Added code examples to the documentation of keras activation APIs: elu, exponential, gelu.

LakshmiKalaKadali:
tf.keras.activations APIs do not contain example code for the activations: elu, exponential, gelu. This PR will fix it.


fchollet (commented):
There's a number of issues here --

1. Code examples should only use Keras APIs, not TensorFlow APIs.
2. Code example should use markdown code formatting.
3. Code examples should convey something non-obvious about the API considered. Simply calling a function on an input that means nothing, and printing a result that means nothing, isn't helpful to readers.

PR #19892, opened 2024-06-20
Fix for broadcast in metrics reduction.

hertschuh:
This would cause dynamic shapes to not work properly with the Tensorflow backend.


fchollet (approved):
LGTM -- thanks for the fix!

PR #19893, opened 2024-06-21
Add `dtype` argument to all losses and export `CTC`

james77777778:
This PR allows users to set the `dtype` for `Loss`. Previously, it was only configurable via `keras.backend.set_floatx()`.


fchollet (approved):
LGTM, thank you!

PR #19896, opened 2024-06-21
Fix `dtype` argument passed to `Layer` when it is a `FloatDTypePolicy` in config

james77777778:
This PR is a small but important fix.

We should ensure that the `dtype` argument passed to `Layer` is a str when it is a `FloatDTypePolicy` in config for backward compatibility.
Otherwise, it will break the downstream project like KerasNLP. (Currently, master branch should break it.)

EDITED:
It would be a good idea to replace `is_quantized` with `quantization_mode` in `DTypePolicy`.
Now, `quantization_mode` returns `None` if the policy is not quantized. This behavior matches `Layer`.


fchollet (approved):
LGTM, thank you!

> It would be a good idea to replace is_quantized with quantization_mode in DTypePolicy.

That sounds good to me.

PR #19901, opened 2024-06-21
Fix for JAX exported model not updating non-trainable variables.

hertschuh:
Modification of non-trainable variables were not reflected in the `tf.Variable`s for exported JAX models. In particular, this caused random functions to use the same seed on every call.

This change binds the non-trainable variables of the model in the exported function.


fchollet (approved):
LGTM, thanks for the fix!

PR #19902, opened 2024-06-21
Fix JAX export support of `TFDataLayer`.

hertschuh:
During export, the `tf.function` wrapping the `jax2tf` function is traced. In this scenario, we're in a JAX tracing scope within a Tensorflow graph. `TFDataLayer` now detects this correctly.

Also improved `jax_utils.is_in_jax_tracing_scope()`: there are multiple tracer array classes that all extend `jax._src.core.Tracer`.


fchollet (approved):
LGTM, thank you

PR #19910, opened 2024-06-24
Fix `ops.Nonzero.compute_output_spec()`.

hertschuh:
`nonzero` returns a tuple of N 1D integer tensors, one per dimension in the input.

Also changed `nonzero` unit tests to be more interesting by actually exercising the case where there are zeros.


fchollet (approved):
Thanks for the fix!

PR #19923, opened 2024-06-26
Fix compatibility of `random.beta` in `tf.data`

james77777778:
`if tf.rank(*)` will fail if we want to use `random.beta` within a `TFDataLayer`


fchollet (approved):
LGTM, thanks for the fix!

PR #19938, opened 2024-06-30
add associative_scan

SeKim12:
Addresses #19904, adds `associative_scan` support for all backends. 

This is my first time contributing, so would greatly appreciate any feedback or something I may be missing (and would be more than happy to apply them)! Thanks!

PR #19940, opened 2024-06-30
resolve torch device placement issue for associative_scan

SeKim12:
Ref: #19938


fchollet (approved):
Thanks for the fix! LGTM

Issue #19725, opened 2024-05-17
Resizing layer meets division by zero crash with jax backend


AGFACBNNR:
As shown in this [colab](https://colab.research.google.com/drive/1P-z9KrXKhxsxLXgcnT_JJkArHQE--a9q?usp=sharing), Resizing meets a 'division by zero' crash with jax backend.
This is triggered when 'height' is large and crop_to_aspect_ratio=True.
According to the error message, it seems that after the preprocessing in Keras, jax receives an image whose dimension list contains a zero, which is directly used as a divisor in jax.
Currently the output is:
![1715930082639](https://github.com/keras-team/keras/assets/20224019/352c5582-0e49-4e8e-9a10-c4c04b92112b)
where we can observe that the API can execute successfully when  crop_to_aspect_ratio=False or 'height' is small, while meets a crash in the third time.

AGFACBNNR:
The symptom looks similar to issue #19723, but I'm not sure whether the reasons behind are same too.
Besides, I also found this problem in RandomCrop layer. 
Please let me know if you need a code snippet to reproduce the crash in RandomCrop.

mehtamansi29:
Hi @AGFACBNNR 

Thanks for reporting the issue. I have tested the code snippet and reproduces the reported behaviour. Attached [gist](https://colab.sandbox.google.com/drive/1TgLDFTJBf3I2TkDcGiZmn50fOZXfvrlp?resourcekey=0-PyF_IClRt7CNFf_ij1crcg#scrollTo=TfPSad7CAgeU) file for reference.

We will look into the issue and update you the same.

sachinprasadhs:
This issue is now fixed with the above linked PR and here is the working Gist in the `keras-nightly` version [![](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.sandbox.google.com/gist/sachinprasadhs/13dd7921f1f5cd58a6f8b4a4bec09711/jax_resizing.ipynb)

Issue #19723, opened 2024-05-17
CenterCrop meets a division-by-zero with jax backend


AGFACBNNR:
As shown in this [colab](https://colab.research.google.com/drive/1e5H7NEdw58OgxScH-PcUgmV1rQMvUxWv?usp=sharing),  I have an image of size (11,3). Using CenterCrop   I can resize it to (7,17) and (5,4), but I meet a crash when trying to resize it to (5,17).
According to the error message, it seems that Keras generates an scaled image of height 0 and passes it to backends.
I've tried to change the backend and the same problem also shows up with tensorflow and torch backend. 
The only difference is that tensorflow and torch checks the existence of 0, while jax directly uses it as a divisor.
Currently the output is:
![1715918255039](https://github.com/keras-team/keras/assets/20224019/2667b137-6eac-4bea-a49b-3b5d33b0beee)

mehtamansi29:
Hi @AGFACBNNR 

Thanks for reporting the issue. I have tested the code snippet and reproduces the reported behaviour. Attached [gist](https://colab.sandbox.google.com/gist/mehtamansi29/9d27d62f369dbb89673b20030e0641f2/19723-centercrop-meets-a-division-by-zero-with-jax-backend.ipynb) file for reference.

We will look into the issue and update you the same.

mattdangerw:
Just running you colab, I actually see a failure on all backends, jax/tf/torch. So this might just be a general bug.

Either way, open for contributions if someone want to take this on. I'll take a look later too to see if its an easy fix.

Thanks for filing!

sachinprasadhs:
Hi, The issue has been fixed and I can see the proper output in the keras-nighlty version, attaching the Gist here for reference [![](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.sandbox.google.com/gist/sachinprasadhs/f2289467afb9e85e93e2e15c9ff483e7/jax_centercrop.ipynb)

PR #19941, opened 2024-07-01
address zero division in crops

SeKim12:
Addresses #19723, #19725 
Also added test cases to verify that the erroring code mentioned in the above issue now works. Thanks!


fchollet (approved):
Thanks for the PR. This change means that cropping now has a minimum width/height of 1 (previously it was 0). Since 0-sized axes are generally invalid this seems fine. LGTM

PR #19945, opened 2024-07-01
Fix `Trainer._symbolic_build`

### LINKED ISSUE #19929: Keras fails to train on custom generator (Torch backend)
Keras model would throw an error on training when using Torch backend and a custom (Python) data generator.

Steps to reproduce:
1) Define a model
2) Define a simple data generator that yields one batch of random data
3) Start training with `epoch=1` and `steps_per_epoch=1`
4) Get an `UnboundLocalError: local variable 'logs' referenced before assignment` suggesting that the training loop itself was skipped, and `epoch_iterator.enumerate_epoch()` has issues wrapping provided generator.

The same steps would work just fine with TF.

[Colab](https://colab.research.google.com/drive/1kOO9W-I2A3wrD3NOuwgH0_MPB-YFs7vr)

Edit: torch==2.3 seems to be incompatible with Keras, but it would have the same fail on 2.2 which should be supported

Grvzard:
https://github.com/keras-team/keras/blob/24ffc7cadb0e54a56d77527a0b9e47053dfebc97/keras/src/trainers/trainer.py#L1004-L1010

In `Trainer.fit()`, there's a call to `self._symbolic_build()` before training.
It consumes one extra piece of data which may be unexpected behavior if the amount of data is calculated.
Only the JAX and Torch backends are affected.

Closes #19929

### DIFF
--- a/keras/src/trainers/trainer.py
+++ b/keras/src/trainers/trainer.py
@@ -1001,12 +1001,6 @@ class Trainer:
         optimizer_unbuilt = (
             self.optimizer is not None and not self.optimizer.built
         )
-        if model_unbuilt or compile_metrics_unbuilt or optimizer_unbuilt:
-            if data_batch is None:
-                for _, data in iterator.enumerate_epoch():
-                    data_batch = data[0]
-                    break
-
         if model_unbuilt or compile_metrics_unbuilt:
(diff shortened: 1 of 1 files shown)


fchollet (approved):
The fix looks good, thanks for the PR!

PR #19946, opened 2024-07-01
Fix for a change in JAX's promotion rules for `floor` and `ceil`.

hertschuh:
JAX has an unreleased change https://github.com/google/jax/pull/21441 that make it so that `int`s and `bool`s are no longer promoted to floats by `floor` and `ceil`.

Our current implementation follows the Numpy promotion rule, i.e. `int`s and `bool`s are promoted to floats. Therefore:
- updated unit tests, which use `jax.numpy` as the reference.
- updated our JAX implementation of `floor` and `ceil` (note that `ceil` would already cast).

### DIFF
--- a/keras/src/backend/jax/numpy.py
+++ b/keras/src/backend/jax/numpy.py
@@ -381,7 +381,8 @@ def ceil(x):
         dtype = config.floatx()
     else:
         dtype = dtypes.result_type(x.dtype, float)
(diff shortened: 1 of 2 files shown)


SamanehSaadat (approved):
Thanks, Fabien!

fchollet (approved):
LGTM, thank you for the fix!

PR #19948, opened 2024-07-02
Fix `tf.save_model.save()` for models which take an input structure a…

hertschuh:
…s a single input argument.

Fixes https://github.com/tensorflow/tensorflow/issues/69307


fchollet (approved):
LGTM, thanks for the fix!

PR #19949, opened 2024-07-03
Add `type_check` for `quantize`

james77777778:
Related to https://github.com/keras-team/keras-nlp/pull/1670
The logic could be further simplified if we could call `super().quantize` in subclasses.

A minor simplification in the tests for quantization is also included.


fchollet (approved):
LGTM, thank you for the PR.

PR #19950, opened 2024-07-03
Fix docstring: typo for setting environment variables

kandolfp:
Just came across some minor typo in the code example that results in an error as the string is not closed.


fchollet (approved):
LGTM, thanks

PR #19958, opened 2024-07-06
add on_epoch_begin for dataset

SeKim12:
Addresses issue #19749 

I wanted to double-check a couple of things, though: 
  1. `on_epoch_end` and `on_epoch_begin` are for `PyDataset` and not for other backend-specific datasets.
  2. The user is responsible for making these callbacks thread-safe and free of side effects. I made a [notebook](https://colab.research.google.com/drive/1LPQRLw0ujJkPZF_VScZ0zxuAe2YUzB2t?usp=sharing) to demonstrate this.
   e.g. if the user were to shuffle the dataset at the end of each epoch, there is a chance that the dataset is shuffled twice, as demonstrated in the notebook. 

I implemented `on_epoch_begin()` based on the above understanding. But if that's a misunderstanding please let me know and would be happy to fix! Thanks!

### DIFF
--- a/guides/training_with_built_in_methods.py
+++ b/guides/training_with_built_in_methods.py
@@ -592,6 +592,7 @@ A `PyDataset` must implement two methods:
 
 The method `__getitem__` should return a complete batch.
 If you want to modify your dataset between epochs, you may implement `on_epoch_end`.
+You may also implement `on_epoch_begin` to at the start of each epoch.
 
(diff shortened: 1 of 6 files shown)


fchollet (approved):
That's a reasonable addition, consistent with `on_epoch_end`. Thanks for the PR -- LGTM

PR #19963, opened 2024-07-08
fixing segment_max edge cases and add more exception handling

haohuanw:
`tf.unsorted_segment_*` is actually able to handle n dimension segment_ids but all other backends and `tf.segment_*` doesn't support this case. this pr adds an actual validation check to let user know about the requirements.

in addition to above, i also:
1. reduced boilerplate code in numpy and torch implementation
2. fixed numpy and torch's segment_max initialization state handling issue
3. reduced boilerplate code for segment reduction in unit tests and test for more cases.


fchollet (approved):
LGTM, thanks for the PR.

PR #19966, opened 2024-07-09
Cast random.categorical seed to int32 in TF backend

jeffcarp:
src/samplers/top_k_sampler_test.py to run into a compilation error:

```
(OpKernel was found, but attributes didn't match) Requested Attributes: T=DT_FLOAT, Tseed=DT_INT64, output_dtype=DT_INT64){{node while/stateless_categorical/StatelessMultinomial}}
```

### DIFF
--- a/keras/src/backend/tensorflow/random.py
+++ b/keras/src/backend/tensorflow/random.py
@@ -29,6 +29,8 @@ def uniform(shape, minval=0.0, maxval=1.0, dtype=None, seed=None):
 
 def categorical(logits, num_samples, dtype="int64", seed=None):
     seed = draw_seed(seed)
+    # Only int32 seeds are supported in stateless_categorical.
+    seed = tf.cast(tf.math.floormod(seed, tf.int32.max - 1), dtype="int32")
     output = tf.random.stateless_categorical(logits, num_samples, seed=seed)
     return tf.cast(output, dtype)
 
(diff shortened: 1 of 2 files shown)


hertschuh (changes requested):
Thanks for working on the fix!

keras/src/backend/tensorflow/random.py
```diff
@@ -29,6 +29,7 @@ def uniform(shape, minval=0.0, maxval=1.0, dtype=None, seed=None):
 
 def categorical(logits, num_samples, dtype="int64", seed=None):
     seed = draw_seed(seed)
+    seed = tf.cast(seed, dtype="int32") # Only int32 seeds are supported.
```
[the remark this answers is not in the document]
jeffcarp: Good point, thanks - updated.

hertschuh (approved):

keras/src/backend/tensorflow/random_test.py
```diff
@@ -0,0 +1,20 @@
+import tensorflow as tf
+
+from keras.src.backend.tensorflow import random
+from keras.src.testing import TestCase
+
+
+class TFRandomTest(TestCase):
```
hertschuh, line 7: You'll have to mark this one as tensorflow only, it's failing with the torch backend.

jeffcarp: Done, thanks.

fchollet (approved):
LGTM, thanks!

PR #19972, opened 2024-07-10
Add `bool` support to `ops.nn.one_hot` and `ops.nn.multi_hot` for the…

hertschuh:
… Tensorflow backend.

Other backends already supported it.
Also added unit tests for `bool` support for both `sparse=False` and `sparse=True`.


fchollet (approved):
Nice, thanks for the PR!

PR #19975, opened 2024-07-11
Fix TF's GPU CI

james77777778:
Just a simple fix.


fchollet (approved):
Thanks for the fix!

To be fair, testing the numerical output of a random sampling function is going to be quite brittle since the RNG behavior might change at any time. Ideally this test would not have been written like this in the first place.

james77777778:
Yeah, it was kind of weird for me when I first saw this, but I was unsure of the actual purpose.
I can try to improve it in another PR.

PR #19979, opened 2024-07-12
Fix torch average pooling

### LINKED ISSUE #19721: averagePooling2D calculates wrongly with torch backend
As shown in this [colab](https://colab.research.google.com/drive/1jZZ7eK_heTTCmSp22xk2IiZ16NsGByqS?usp=sharing),
AveragePooling2D generate an output with wrong shape with torch backend.
Specifically, the bug is triggered only when padding='same' and input_height != input_weight.
Details can be checked in the shared colab,  currently the output is:

When the bug is triggered(line 2 and line 5 in output), the output has an wrong shape.

mattdangerw:
I might be missing something, but whatever custom logic we had for torch average pooling appears to be broken, and the common function we have to compute same padding on torch seems to fix it.

Fixes #19721

### DIFF
--- a/keras/src/backend/torch/nn.py
+++ b/keras/src/backend/torch/nn.py
@@ -308,35 +308,21 @@ def average_pool(
     data_format = backend.standardize_data_format(data_format)
     if data_format == "channels_last":
         inputs = _transpose_spatial_inputs(inputs)
-    padding_value = 0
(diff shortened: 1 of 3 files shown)


fchollet (approved):
Simpler is better -- LGTM

PR #19980, opened 2024-07-12
Updated docstring for `Model.compute_metrics()`.

hertschuh:
Added details and a more complete example of how to use custom metrics.

Also corrected inaccurate statement about the tracking of custom metrics.


fchollet (approved):
LGTM, thanks for the fix.

PR #19981, opened 2024-07-12
add lamb optimizer

huberb:
This PR adds the Lamb optimizer described in the paper "Large Batch Optimization for Deep Learning: Training BERT in 76 minutes" (https://arxiv.org/abs/1904.00962)

The optimizer was previously implemented by the authors of the paper in the tensorflow_addons repository (https://github.com/tensorflow/addons) which is deprecated.

I was interested in this optimizer and opened an issue before. It is available here: https://github.com/keras-team/keras/issues/19965

I compared its output against `tensorflow_addons.optimizers.LAMB` and behavior is identical **only if `weight_decay=0.0`**. This is because the implementation in tensorflow_addons couples weight decay to the optimization procedure (as did previous implementations of Adam, see https://arxiv.org/abs/1711.05101 for details)

When looking at other optimizer implementations in keras 3 it seems to me that the preferred way to handle weight decay is to leave it to the optimizer base class and therefore I did the same. Contrary to pytorch where `Adam` and `AdamW` treat weight decay differently, keras 3 only implements a single version of weight decay and `keras.optimizers.AdamW` is just a child class of `keras.optimizers.Adam` that enforces the usage of weight decay (see here: https://github.com/keras-team/keras/blob/master/keras/src/optimizers/adamw.py#L91)

This however means that this implementation differs from the implementation in `tensorflow_addons` which might lead to confusion. Please let me know if this is the preferred behaviour. I also created a colab to highlight this:

https://colab.research.google.com/drive/13oSX0daqRV199v0wCFtFAriRA_wlAj8w

Furthermore, it seems like there are different versions of Lamb floating around the internet. The paper "ResNet strikes back: An improved training procedure in timm" (https://arxiv.org/abs/2110.00476) uses the implementation that is available here: https://github.com/huggingface/pytorch-image-models/blob/main/timm/optim/lamb.py

This implementation adds a normalization over all gradients before applying the Lamb algorithm. More details are available on this blog post from Nvidia: https://developer.nvidia.com/blog/pretraining-bert-with-layer-wise-adaptive-learning-rates/

For now, I'd like some feedback on this PR to see if I understood the handling of weight decay correctly. Please also let me know if you would prefer the version with gradient norm instead of this. Otherwise I plan to add another PR later which implements it as a different optimizer (probably going to call it `keras.optimizers.NVLamb`).

### DIFF
--- a/keras/api/_tf_keras/keras/optimizers/__init__.py
+++ b/keras/api/_tf_keras/keras/optimizers/__init__.py
@@ -16,6 +16,7 @@ from keras.src.optimizers.adam import Adam
 from keras.src.optimizers.adamax import Adamax
 from keras.src.optimizers.adamw import AdamW
 from keras.src.optimizers.ftrl import Ftrl
+from keras.src.optimizers.lamb import Lamb
 from keras.src.optimizers.lion import Lion
 from keras.src.optimizers.loss_scale_optimizer import LossScaleOptimizer
 from keras.src.optimizers.nadam import Nadam
--- a/keras/api/optimizers/__init__.py
+++ b/keras/api/optimizers/__init__.py
@@ -16,6 +16,7 @@ from keras.src.optimizers.adam import Adam
 from keras.src.optimizers.adamax import Adamax
 from keras.src.optimizers.adamw import AdamW
 from keras.src.optimizers.ftrl import Ftrl
+from keras.src.optimizers.lamb import Lamb
 from keras.src.optimizers.lion import Lion
 from keras.src.optimizers.loss_scale_optimizer import LossScaleOptimizer
 from keras.src.optimizers.nadam import Nadam
--- a/keras/src/optimizers/lamb.py
+++ b/keras/src/optimizers/lamb.py
@@ -1,0 +1,158 @@
+from keras.src import ops
+from keras.src.api_export import keras_export
+from keras.src.optimizers import optimizer
+
+
+@keras_export("keras.optimizers.Lamb")
+class Lamb(optimizer.Optimizer):
+    """Optimizer that implements the Lamb algorithm.
+
+    Lamb is a stochastic gradient descent method that
+    uses layer-wise adaptive moments to adjusts the
+    learning rate for each parameter based on the ratio of the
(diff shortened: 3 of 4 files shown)


fchollet (approved):
The code looks good -- LGTM. Thank you for the contribution.

james77777778:
We might need to add LAMB implementation in `keras/src/backend/torch/optimizers` as well

Also here:
https://github.com/keras-team/keras/blob/master/keras/src/backend/torch/optimizers/torch_optimizer.py

ErfanMowlaei:
I'm opening this issue again because I am seeing large difference in performance between what is currently on Keras3 and that of tensorflow_addons. I am getting confused by the addons code and it is hard for me to port it accurately to keras3. Is it possible for someone to check to see what is the difference?

Addons implementation is providing a much better convergence for my models across different tasks.

huberb:
Hi Erfan

There is the above mentioned Colab link that shows that the parameters of a model after some training steps are identical to the tensorflow_addons implementation

https://colab.research.google.com/drive/13oSX0daqRV199v0wCFtFAriRA_wlAj8w

Note that the parameters actually differ if you use weight_decay, but weight_decay is implemented only once for all optimizers in keras3 and there's some literature that shows quite extensively that keras3's version of weight decay should have better convergence, although your concrete results may vary depending on your usecase.

There are quite some other factors that might play a role if you switch from one of the older versions of tensorflow that still support tensorflow_addons to the most up to date keras3 version. Could you give some more details about what your training code looks like? Or even show a small example where the convergence differs?

ErfanMowlaei:
Thanks for the quick response. I have not compared the keras and addons implementations directly, but I have compared the keras implementation to that of Adam on my code and they were performing the same, and I have compared addons to adam and it was doing much better. The code for which I have compared addons and adam is here: 
https://github.com/shilab/STICI/blob/HPC/STICI_V1.1.py

It takes me a bit of time to port this code to tf 2.16. By the way, I never used weight decay in my optimizer.

ErfanMowlaei:
Thanks for the quick response. I investigated a bit more closely and you are right, they are equivalent. Thank you!

PR #20007, opened 2024-07-17
Use Numpy arrays in data adapters for JAX backend.

hertschuh:
- Slicing and splitting is more efficient with Numpy arrays on CPU than JAX arrays on CPU.
- Converting to `jax.Array` sends them to the first device by default, which is incorrect in the case of distribution and slows down distribution.

Also
- Keep JAX BCOO tensors on CPU within data adapters.
- Use `np.split` instead of `jax.numpy.split` in distribution_lib per the JAX distribution guide.

### DIFF
--- a/keras/src/trainers/data_adapters/data_adapter.py
+++ b/keras/src/trainers/data_adapters/data_adapter.py
@@ -30,7 +30,8 @@ class DataAdapter:
         raise NotImplementedError
 
     def get_jax_iterator(self):
-        """Get a Python iterable for the `DataAdapter`, that yields JAX arrays.
+        """Get a Python iterable for the `DataAdapter`, that yields arrays that
+        that can be fed to JAX. NumPy arrays are preferred for performance.
 
         Returns:
             A Python iterator.
--- a/keras/src/trainers/data_adapters/data_adapter_utils.py
+++ b/keras/src/trainers/data_adapters/data_adapter_utils.py
@@ -176,10 +176,21 @@ def get_tensor_spec(batches):
 
 
 def get_jax_iterator(iterable):
-    from keras.src.backend.jax.core import convert_to_tensor
+    import jax
+    import jax.experimental.sparse as jax_sparse
+
+    def convert_to_jax_compatible(x):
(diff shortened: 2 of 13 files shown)


keras/src/trainers/data_adapters/data_adapter_utils.py
```diff
@@ -176,10 +176,21 @@ def get_single_tensor_spec(*tensors):
 
 
 def get_jax_iterator(iterable):
-    from keras.src.backend.jax.core import convert_to_tensor
+    import jax
+    import jax.experimental.sparse as jax_sparse
+
+    def convert_to_jax(x):
+        if isinstance(x, (jax.Array, jax_sparse.JAXSparse, np.ndarray)):
```
SamanehSaadat, line 183: The return value is not always a jax array here since it can return `np.ndarray` too. right? I was thinking `convert_to_jax` name might be a little bit confusing then!

SamanehSaadat (approved):
Thanks, Fabien!

keras/src/trainers/data_adapters/data_adapter.py
```diff
@@ -30,7 +30,8 @@ def get_tf_dataset(self):
         raise NotImplementedError
 
     def get_jax_iterator(self):
-        """Get a Python iterable for the `DataAdapter`, that yields JAX arrays.
+        """Get a Python iterable for the `DataAdapter`, that yields arrays that
+        that get be fed to JAX. NumPy arrays are preferred for performance.
```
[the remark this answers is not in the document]
hertschuh: Indeed. Fixed.

keras/src/trainers/data_adapters/data_adapter_utils.py
hertschuh: Renamed `convert_to_jax_compatible`, like in `Sliceable`.

hertschuh:
> Thanks, Fabien!

Thanks for the review!

fchollet (approved):
LGTM, thanks for the fix!

PR #20014, opened 2024-07-19
Revert to the original default behavior of `dtype` in `Loss` and `Metric`

james77777778:
There is a regression after #19937 where, when users set a global dtype policy such as "mixed_float16" or "mixed_bfloat16", the `dtype` is wrongly set in `Loss` and `Metric`.
It should follow `keras.config.floatx()` as stated in the docstrings.


fchollet (approved):
LGTM, thanks for the fix.

PR #20018, opened 2024-07-20
Faster host to device transfer on Jax backend

Hilly12:
- Lazily compute the layout using the first batch and re-use it. This speeds up host to device transfers which were slowed down by creating a new mesh and sharding each time. `jax.device_put` is still pretty slow so longer term it might be worth looking into packing features with a similar layout so there's fewer device puts.
- Allow turning sharding across processes off in the distribution API. This allows doing `dataset.shard(index)` followed by `dataset.batch(per_process_batch_size)` which enables scaling up to larger batch sizes.

Before:

image

After:

image

### DIFF
--- a/keras/src/backend/jax/trainer.py
+++ b/keras/src/backend/jax/trainer.py
@@ -968,7 +968,7 @@ class JAXTrainer(base_trainer.Trainer):
         return tuple(state)
 
 
-def _distribute_data(data, layouts):
+def _distribute_data(data, layouts=None):
     if layouts is not None:
 
         def distribute_single_value(d, layout):
--- a/keras/src/backend/jax/trainer.py
+++ b/keras/src/backend/jax/trainer.py
@@ -14,7 +14,6 @@ from keras.src.distribution import distribution_lib
 from keras.src.trainers import trainer as base_trainer
 from keras.src.trainers.data_adapters import array_slicing
 from keras.src.trainers.data_adapters import data_adapter_utils
-from keras.src.trainers.data_adapters.tf_dataset_adapter import TFDatasetAdapter
 from keras.src.trainers.epoch_iterator import EpochIterator
 from keras.src.utils import traceback_utils
 
@@ -969,12 +968,16 @@ class JAXTrainer(base_trainer.Trainer):
 
 
 def _distribute_data(data, layouts=None):
-    if layouts is not None:
+    distribution = distribution_lib.distribution()
+    if distribution is not None:
 
-        def distribute_single_value(d, layout):
+        def distribute_single_value(d, layouts=None):
+            if layouts is None:
+                layout = distribution.get_data_layout(d.shape)
             return jax_distribution_lib.distribute_data_input(d, layout)
 
         return tree.map_structure(distribute_single_value, data, layouts)
+
     else:
         return tree.map_structure(jax.device_put, data)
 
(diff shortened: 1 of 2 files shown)


keras/src/backend/jax/trainer.py
```diff
@@ -967,21 +968,35 @@ def _get_jax_state(
...
+        return tree.map_structure(distribute_single_value, data, layouts)
     else:
         return tree.map_structure(jax.device_put, data)
 
 
 class JAXEpochIterator(EpochIterator):
     def _get_iterator(self):
+        layouts = None
+        distribution = distribution_lib.distribution()
+        if isinstance(self.data_adapter, TFDatasetAdapter):
```
hertschuh, line 986: I think the criteria should be `if distribution is not None:`, you're allowed to distribute from other data adapters.
And conversely, if it is a `TFDatasetAdapter` and distribution is *not* on, you want to do the prefetch iterator, not just `device_put`.

hertschuh (changes requested):
Thanks for the optimization!

keras/src/backend/jax/trainer.py
```diff
@@ -967,21 +968,35 @@ def _get_jax_state(
         return tuple(state)
 
 
-def _distribute_data(data):
-    distribution = distribution_lib.distribution()
-    if distribution is not None:
+def _distribute_data(data, layouts=None):
+    if layouts is not None:
```
SamanehSaadat, line 972: If the distribution is on and the layout is not passed, we won't distribute the data anymore, right?
There are two other instances of calling `_distribute_data`:
1. [in train_on_batch](https://github.com/keras-team/keras/blob/f9501f53fa15fe6286ca834618a6bcf10f86f7e8/keras/src/backend/jax/trainer.py#L717)
2. [in test_on_batch](https://github.com/keras-team/keras/blob/f9501f53fa15fe6286ca834618a6bcf10f86f7e8/keras/src/backend/jax/trainer.py#L766)

Since layout hasn't been computed in these cases, data distribution probably won't happen.

Should we move `if layouts is not None:` inside `distribute_single_value` and add back `if distribution is not None:` here?

Something like this:

```python
def _distribute_data(data, layouts=None):
    distribution = distribution_lib.distribution()
    if distribution is not None:

        def distribute_single_value(d, layouts=None):
            if layouts is None:
                layout = distribution.get_data_layout(d.shape)
            return jax_distribution_lib.distribute_data_input(d, layout)

        return tree.map_structure(distribute_single_value, data, layouts)

    else:
        return tree.map_structure(jax.device_put, data)
```

keras/src/backend/jax/trainer.py
Hilly12: Done.

keras/src/backend/jax/trainer.py
Hilly12: Thanks for catching, I didn't spot those calls to `_distribute_data`. Updated.

keras/src/backend/jax/trainer.py
```diff
@@ -967,21 +967,39 @@ def _get_jax_state(
...
 
 
-def _distribute_data(data):
+def _distribute_data(data, layouts=None):
     distribution = distribution_lib.distribution()
     if distribution is not None:
 
-        def distribute_single_value(d):
-            layout = distribution.get_data_layout(d.shape)
+        def distribute_single_value(d, layouts=None):
```
hertschuh, line 974: There's a mix-up between `layout` and `layouts` here. It should be `layout` within `distribute_single_value` everywhere, right?

SamanehSaadat: Right! It was my fault! I typed in the comment box here on GH so I didn't realize it's called `layouts` everywhere else!

Hilly12: Fixed.

hertschuh (approved):

SamanehSaadat (approved):
Thanks, Aahil!

PR #20025, opened 2024-07-22
Fix test_softmax_correctness_with_axis_tuple test case failed on torch gpu ci.

shashaka:
Fixing [test case failed on torch gpu ci](https://github.com/keras-team/keras/pull/20022#issuecomment-2241951918) reported by @james77777778 .


fchollet (approved):
LGTM, thanks!

PR #20028, opened 2024-07-22
Fix log_softmax logic for tuple axis

shashaka:
Similar with [#20021](https://github.com/keras-team/keras/issues/20021),
On log_softmax function, also has the same issue.
It couldn't pass the below test case that check the sum of all the exponential values should be 1.0.
So, I fixed the logic on log_softmax function also.

```
    def test_log_softmax_correctness_with_axis_tuple(self):
        input = np.array([[[1.0, 2.0], [3.0, 4.0]], [[5.0, 6.0], [7.0, 8.0]]])
        combination = combinations(range(3), 2)
        for axis in list(combination):
            result = keras.ops.nn.log_softmax(input, axis=axis)
            normalized_sum_by_axis = np.sum(
                np.exp(ops.convert_to_numpy(result)), axis=axis
            )
            self.assertAllClose(normalized_sum_by_axis, 1.0)
```

### DIFF
--- a/keras/src/ops/nn.py
+++ b/keras/src/ops/nn.py
@@ -613,23 +613,19 @@ def log_softmax(x, axis=-1):
     if any_symbolic_tensors((x,)):
         return LogSoftmax(axis).symbolic_call(x)
     if isinstance(axis, tuple):
-        original_shape = x.shape
-        new_shape = []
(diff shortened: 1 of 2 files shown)


fchollet (approved):
LGTM, thank you

PR #20033, opened 2024-07-23
Fix jax's `in_top_k` and numpy's `top_k`

### LINKED ISSUE #19995: Incorrect handling of `nan` in `in_top_k` of Jax backend (regression from keras 3.3.3)
The `top_in_k` function in the Jax backend provides an unexpected result. 

Example to reproduce the problem:
```
import jax.numpy as jnp
from numpy import nan
from keras.src.backend.jax.math import in_top_k

r = in_top_k(targets=jnp.array([1, 0]), predictions=jnp.array([[.1, nan, .5], [.3, .2, .5]]), k=2)
print(r)
```

With keras 3.3.3, I'm getting the expected outcome:
`[False  True]`

However, keras 3.4.1 gives this:
`[ True  True]`

The new behavior is unexpected because `nan` shouldn't be be considered as large probability in the prediction.

As a first step to debug it: The change in behavior has been introduced by https://github.com/keras-team/keras/pull/19814.

james77777778:
Fix #19995

The performance will be slightly slower, but it should be acceptable.

|Before|After||
|-|-|-|
|10.842s|11.063s|+2%|

Benchmark script:

```python
import functools
import timeit

import jax

from keras.src.backend.jax.math import in_top_k

logits = jax.random.uniform(jax.random.key(42), shape=(32, 128, 50))
labels = jax.random.randint(
    jax.random.key(42), shape=(32, 128), minval=0, maxval=50
)


@functools.partial(jax.jit, static_argnames=["k"])
def jit_in_top_k(logits, labels, k=5):
    return in_top_k(labels, logits, k=k)


print(jit_in_top_k(logits, labels, k=5).mean())
print(timeit.timeit(lambda: jit_in_top_k(logits, labels, k=5)))

```

Additionally, improve numpy's `top_k` which had significantly poor performance due to double sorting (or even more times of sorting...)

### DIFF
--- a/keras/src/backend/jax/math.py
+++ b/keras/src/backend/jax/math.py
@@ -43,6 +43,10 @@ def in_top_k(targets, predictions, k):
     preds_at_label = jnp.take_along_axis(
         predictions, jnp.expand_dims(targets, axis=-1), axis=-1
     )
+    # `nan` shouldn't be considered as large probability.
+    preds_at_label = jnp.where(
+        jnp.isnan(preds_at_label), -jnp.inf, preds_at_label
+    )
     rank = 1 + jnp.sum(jnp.greater(predictions, preds_at_label), axis=-1)
     return jnp.less_equal(rank, k)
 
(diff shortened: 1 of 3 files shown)


fchollet (approved):
LGTM, thanks for the fix.

PR #20034, opened 2024-07-23
Add clear deprecation error message for add_metric()

### LINKED ISSUE #19804: Clearer errors for `add_metric`
1) It's really weird to drop `add_metric` when you still have `add_loss`, but alright
2) If you are not planning to improve compatibility with 'niche functionality' of the past, then please produce clearer errors that it does not work. Right now when you try passing any parameters to `add_metric`, it fails with a type error:

```
TypeError: Exception encountered when calling MyMetric.call().

Layer.add_metric() takes 1 positional argument but 2 were given

Arguments received by MyMetric.call():
  • y_true=tf.Tensor(shape=(2, 6), dtype=int32)
  • y_pred=tf.Tensor(shape=(2, 8), dtype=int32)
  • args=<class 'inspect._empty'>
```

From this error it's not exactly clear that the method:
1) takes no arguments
2) is deprecated / not implemented

To actually understand that the method is not even supposed to work you have to either inspect the source code, or rummage through the github, none of which is super obvious. 

Suggestions: 
1) remove the method altogether (if it's not supported, why have it in the first place?)
2) make it accept anything (*args, **kwargs), and then throw a `NotImplementedError`, possibly add suggestions on how to use the updated API instead

A Colab notebook for reference: https://colab.research.google.com/drive/1FO-GstSFwU0V0cTeCCjtCuL2NUmFB2tq

grasskin:
Fix #19804

### DIFF
--- a/keras/src/layers/layer.py
+++ b/keras/src/layers/layer.py
@@ -1284,7 +1284,7 @@ class Layer(BackendLayer, Operation, KerasSaveable):
         # Permanently disabled
         raise NotImplementedError(
             "Layer `add_metric()` method is deprecated"
-            " add your metric in `compile(metrics=[...]).`"
+            " add your metric in `Model.compile(metrics=[...]).`"
         )
 
     def count_params(self):
--- a/keras/src/layers/layer.py
+++ b/keras/src/layers/layer.py
@@ -1280,9 +1280,12 @@ class Layer(BackendLayer, Operation, KerasSaveable):
             self._tracker.lock()
         self._post_untrack_variable(variable)
 
-    def add_metric(self):
+    def add_metric(self, *args, **kwargs):
         # Permanently disabled
-        raise NotImplementedError
+        raise NotImplementedError(
+            "Layer `add_metric()` method is deprecated"
+            " add your metric in `compile(metrics=[...])`"
+        )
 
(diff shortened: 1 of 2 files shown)


keras/src/layers/layer.py
```diff
@@ -1280,9 +1280,12 @@ def _untrack_variable(self, variable):
             self._tracker.lock()
         self._post_untrack_variable(variable)
 
-    def add_metric(self):
+    def add_metric(self, *args, **kwargs):
         # Permanently disabled
-        raise NotImplementedError
+        raise NotImplementedError(
+            "Layer `add_metric()` method is deprecated"
+            " add your metric in `compile(metrics=[...]).`"
```
hertschuh, line 1287: I would say `Model.compile(metrics=[...]).` since this is the `Layer` class.

hertschuh (approved):
Thanks for adding this!

keras/src/layers/layer.py
grasskin: Good call, switched!

grasskin:
We just refactored to a helper `_not_implemented` method but I think this is helpful enough as a message we should keep it.

fchollet (approved):

PR #20040, opened 2024-07-24
Return value on KerasVariable assignment

### LINKED ISSUE #19997: Assign doesn't return a value
When assigning a placeholder value to a KerasVariable I get returned a None type as there is no return in assignment, where as when assigning to a tf.Variable I get returned the proper assign operation.

This is the code in keras.Variable:
```
        self._value.assign(tf.cast(value, self._value.dtype))
```

And here in the KerasVariable:
```
    def assign(self, value):
        value = self._convert_to_tensor(value, dtype=self.dtype)
        if not shape_equal(value.shape, self.shape):
            raise ValueError(
                "The shape of the target variable and "
                "the shape of the target value in "
                "`variable.assign(value)` must match. "
                f"variable.shape={self.value.shape}, "
                f"Received: value.shape={value.shape}. "
                f"Target variable: {self}"
            )
        if in_stateless_scope():
            scope = get_stateless_scope()
            scope.add_update((self, value))
        else:
            self._direct_assign(value)
```

Link to Keras implementation: https://github.com/keras-team/keras/blob/f9501f53fa15fe6286ca834618a6bcf10f86f7e8/keras/src/backend/common/variables.py#L224

grasskin:
To keep compatibility with previous Tensorflow implementations, we can return variable value after assignment.

Specifically, this impacts porting Tensorflow Federated to Keras 3, see https://github.com/google-parfait/tensorflow-federated/blob/0ca2550e62ca7728e5e9a45edca1feedc6bc065f/tensorflow_federated/python/learning/models/functional.py#L528 which depends on returning the value after an `assign`.

Overhead should be minimal and I could see this being useful for finding the value after an `assign_add` or `assign_sub`.

Fix #19997.

### DIFF
--- a/keras/src/backend/common/variables.py
+++ b/keras/src/backend/common/variables.py
@@ -237,12 +237,13 @@ class KerasVariable:
             scope.add_update((self, value))
         else:
             self._direct_assign(value)
+        return value
 
     def assign_add(self, value):
-        self.assign(self + value)
+        return self.assign(self + value)
 
     def assign_sub(self, value):
-        self.assign(self - value)
+        return self.assign(self - value)
 
     @property
     def dtype(self):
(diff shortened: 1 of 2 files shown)


hertschuh (approved):
LGTM, thanks!

fchollet (approved):

PR #20043, opened 2024-07-25
Remove non-used param & add initial value for callback.py

shashaka:
I couldn't find self.validation_data param's usage on callback class, and found that self.params has no initial value unlike self._model.
So, I removed self.validation_data, and add initial value as None for self.params


fchollet (approved):
LGTM, thanks for the fix. `validation_data` is a remnant of a long-deprecated feature.

PR #20049, opened 2024-07-26
Migrate numpy namespace to numpy 2.0

grasskin:
Numpy 2.0 deprecates part of their namespace, see https://numpy.org/doc/stable/numpy_2_0_migration_guide.html for all details.
Keras is affected by `np.Inf` -> `np.inf` and `np.unicode_` -> `np.str_`.


hertschuh (approved):
Thanks!

fchollet (approved):
LGTM, thanks for the fix!

adamjstewart:
Probably closes #19691

Issue #20052, opened 2024-07-28
Is 2D CNN's description correct ?


ghsanti:
>   This layer creates a convolution kernel that is convolved with the layer
    input over a single spatial (or temporal) dimension to produce a tensor of
    outputs. If `use_bias` is True, a bias vector is created and added to the
    outputs. Finally, if `activation` is not `None`, it is applied to the
    outputs as well.

[keras docs](https://github.com/keras-team/keras/blob/v3.4.1/keras/src/layers/convolutional/conv2d.py#L5)

Isn't convolved over *2 spatial dimensions* ?

mehtamansi29:
Hi @ghsanti - 

Thanks for pointing this issue. We have raised an internal fix for updating the description for 2D and 3D CNN. Soon it will be reflected in the documentation as well. Thanks.

PR #20059, opened 2024-07-29
2D CNN's and 3D CNN's description update. 

mehtamansi29:
Updated Conv2D and Conv3D description. Fixes: [#20052](https://github.com/keras-team/keras/issues/20052)

### DIFF
--- a/keras/src/layers/convolutional/conv2d.py
+++ b/keras/src/layers/convolutional/conv2d.py
@@ -7,7 +7,7 @@ class Conv2D(BaseConv):
     """2D convolution layer.
 
     This layer creates a convolution kernel that is convolved with the layer
-    input over a single spatial (or temporal) dimension to produce a tensor of
+    input over a 2D spatial (or temporal) dimension(height and width) to produce a tensor of
     outputs. If `use_bias` is True, a bias vector is created and added to the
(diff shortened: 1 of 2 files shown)


keras/src/layers/convolutional/conv2d.py
```diff
@@ -7,10 +7,10 @@ class Conv2D(BaseConv):
...
 
     This layer creates a convolution kernel that is convolved with the layer
-    input over a 2D spatial (or temporal) dimension(height and width) to produce a tensor of
-    outputs. If `use_bias` is True, a bias vector is created and added to the
-    outputs. Finally, if `activation` is not `None`, it is applied to the
-    outputs as well.
+    input over a 2D spatial (or temporal) dimension(height and width) to 
+    produce a tensor ofoutputs. If `use_bias` is True, a bias vector is created 
+    and added to the outputs. Finally, if `activation` is not `None`, it is 
+    applied to the outputs as well.
```
ghsanti, line 14: Not an expert but imho this would be more clear for most people:

> This layer creates a kernel that is convolved over the 2D spatial dimensions (height and width) of the input..

fchollet (approved):
LGTM, thanks

PR #20062, opened 2024-07-29
Let Tensorflow call the initializer of variables when callable.

hertschuh:
`tf.Variable` allows passing a callable instead of an initial value. We now use a callable when the initializer passed to the Keras `Variable` is a callable. This defers the materialization of the full initial value, which is unnecessary when tracing for instance.


fchollet (approved):
LGTM, thank you! The name `_initialize_with_initializer` is a bit curious but it makes sense.

PR #20065, opened 2024-07-30
Fix dtype issue in `abs_max_quantize`

james77777778:
I encountered an issue with dtype in `abs_max_quantize` when trying to quantize KerasNLP models using `bfloat16` setting.

The root cause is that the computed `scale` might be float32 instead of bfloat16 in some backends (at least, JAX) using a non-CUDA env.

cc @mattdangerw


fchollet (approved):
LGTM, thank you!

PR #20076, opened 2024-08-01
Fix `input_shape` in `Sequential` model

### LINKED ISSUE #20074: Issue with Loading Sequential Models
https://github.com/tensorflow/tensorflow/issues/70757 

Reposting per tensorflow developer's recommendation. 

I am encountering the same issue with pip version 2.17

Thank you

james77777778:
Fix #20074 

A weird behavior was observed where `input_shape` might be a list instead of a tuple during loading.

This PR standardizes the `input_shape` in `Sequential`'s `build`. The criterion for skipping the building process has also been simplified using try-except block.

~~EDITED:~~
~~Add support for backend values to `standardize_shape`.~~


fchollet (approved):
LGTM, thanks for the fix!

PR #20088, opened 2024-08-05
Do not adapt TextVectorization preprocessor if provided with vocabulary  in FeatureSpace. 

livanosg:
Since we can provide custom preprocessors to features of FeatureSpace, trying to adapt to data  a TextVectorization preprocessing layer, with vocabulary provided, raises `ValueError`:
For example the code bellow:
```
vocabulary = ['a', 'b', 'c', 'd']
preprocessor = TextVectorization(vocabulary=vocabulary, output_mode='int', output_sequence_length=5)
feature_space = FeatureSpace(
    features={'example': FeatureSpace.feature(dtype='string', output_mode='int', preprocessor=preprocessor)},
    output_mode='dict')

ds = tf.data.Dataset.from_tensor_slices({'example': tf.constant(['a b', 'b', 'c', 'd', 'bfr cdvew xas'])})

feature_space.adapt(ds)

for sample in ds:
    print(feature_space(sample))
```
raises:
```
  File ".../lib/python3.11/site-packages/keras/src/layers/preprocessing/index_lookup.py", line 586, in update_state
    raise ValueError(
ValueError: Cannot adapt layer 'string_lookup' after setting a static vocabulary via `vocabulary` argument or `set_vocabulary()` method.
```

With this fix we can avoid this error and get the processed features as intended.
```
{'example': <tf.Tensor: shape=(5,), dtype=int64, numpy=array([2, 3, 0, 0, 0])>}
{'example': <tf.Tensor: shape=(5,), dtype=int64, numpy=array([3, 0, 0, 0, 0])>}
{'example': <tf.Tensor: shape=(5,), dtype=int64, numpy=array([4, 0, 0, 0, 0])>}
{'example': <tf.Tensor: shape=(5,), dtype=int64, numpy=array([5, 0, 0, 0, 0])>}
{'example': <tf.Tensor: shape=(5,), dtype=int64, numpy=array([1, 1, 1, 0, 0])>}
```

### DIFF
--- a/keras/src/layers/preprocessing/feature_space.py
+++ b/keras/src/layers/preprocessing/feature_space.py
@@ -490,6 +490,10 @@ class FeatureSpace(Layer):
             if isinstance(preprocessor, layers.Normalization):
                 if preprocessor.input_mean is not None:
                     continue
+            # Special case: a TextVectorization layer with provided vocabulary.
+            elif isinstance(preprocessor, layers.TextVectorization):
+                if preprocessor._has_input_vocabulary:
+                    continue
             if hasattr(preprocessor, "adapt"):
                 adaptable_preprocessors.append(name)
         return adaptable_preprocessors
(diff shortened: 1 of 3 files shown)


fchollet (approved):
Thanks for the fix!

Issue #20063, opened 2024-07-30
Dropout in ConvLSTM Cell


dryglicki:
Really quick question here.

In the ConvLSTMCell [here](https://github.com/keras-team/keras/blob/v3.3.3/keras/src/layers/rnn/conv_lstm.py#L236-L242), the dropout code has been commented out. `recurrent_dropout` also doesn't seem to do anything, except it is used in the `DropoutRNNCell` mixin.

Is this a bug or am I missing something?

sachinprasadhs:
This code snippet was commented in the fix here https://github.com/keras-team/keras/commit/2cae421d476a980e516a19ba5d429825e8765e22 to fix Conv LSTM correctness

dryglicki:
Hi @sachinprasadhs, can you be a little more clear with this explanation? What does "correctness" mean here, exactly?

I've been looking at the base RNN layer here: https://github.com/keras-team/keras/blob/master/keras/src/layers/rnn/rnn.py. Is the dropout now done under the hood?

In the Dropout Test here (https://github.com/keras-team/keras/blob/master/keras/src/layers/rnn/dropout_rnn_cell_test.py), those two calls aren't commented out.

hertschuh:
@dryglicki ,

Thank you for the report. This was an oversight. It is now fixed.

PR #20089, opened 2024-08-05
Added dropout support in `ConvLSTMCell`.

hertschuh:
Implemented dropout and recurrent dropout in `ConvLSTMCell` with the same approach as in Keras 2.

Fixes https://github.com/keras-team/keras/issues/20063


fchollet (approved):
LGTM -- thank you for the fix.

PR #20101, opened 2024-08-09
Remove threshold limits in BinaryAccuracy()

cofri:
`BinaryAccuracy` metric asserts that threshold is in [0, 1]. But this is not the legacy BinaryAccuracy behaviour in TensorFlow 2.15 (and below) and in torcheval metrics, where threshold is not limited to [0, 1]. This behaviour is useful when `y_pred` are logits and not probabilities.

This PR proposes to remove the assertion on threshold value.

Edit: This change could be extended to `FBetaScore` metric that has the same assertion on the threshold.


fchollet (approved):
Ok, this is reasonable. Thanks for the PR.

PR #20102, opened 2024-08-09
Fix `Model.summary(show_trainable=True)`

james77777778:
Currently, the following script fails easily:

```python
import keras

inputs = keras.layers.Input((2,))
x = keras.layers.Dense(4)(inputs)
outputs = keras.ops.mean(x)
model = keras.models.Model(inputs, outputs)
model.summary(show_trainable=True)
```

The root cause is that this line might be a `keras.Operation` which doesn't have a `weights` attr:
https://github.com/keras-team/keras/blob/b45be337c6156d90f220beba9f68eeb2e52e2b0d/keras/src/utils/summary_utils.py#L287

### DIFF
--- a/keras/src/utils/summary_utils.py
+++ b/keras/src/utils/summary_utils.py
@@ -284,7 +284,7 @@ def print_summary(
         if not sequential_like:
             fields.append(get_connections(layer))
         if show_trainable:
-            if layer.weights:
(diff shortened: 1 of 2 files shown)


fchollet (approved):
LGTM, thanks for the fix!

PR #20107, opened 2024-08-10
Fix default value of `logs` in Trainer

### LINKED ISSUE #20104: Tensorflow model.fit fails on test_step: 'NoneType' object has no attribute 'items'
I am using tf.data module to load my datasets. Although the training and validation data modules are almost the same. The train_step works properly and the training on the first epoch continues till the last batch, but in the test_step I get the following error:
```shell 
    353     val_logs = {
--> 354         "val_" + name: val for name, val in val_logs.items()
    355     }
    356     epoch_logs.update(val_logs)
    358 callbacks.on_epoch_end(epoch, epoch_logs)

AttributeError: 'NoneType' object has no attribute 'items'
```

Here is the code for fitting the model:
```shell
results = auto_encoder.fit(
    train_data,
    epochs=config['epochs'],
    steps_per_epoch=(num_train // config['batch_size']),
    validation_data=valid_data,
    validation_steps=(num_valid // config['batch_size'])-1,
    callbacks=callbacks
)
```
I should mention that I have used .repeat() on both train_data and valid_data, so the problem is not with not having enough samples.

Grvzard:
Fixes #20104 .

`_pythonify_logs()` treat `logs` as dict: https://github.com/keras-team/keras/blob/933579d3c4a585f236982d05d3a74921f9567415/keras/src/trainers/trainer.py#L931
`on_test_batch_end()` `on_test_end()` same: https://github.com/keras-team/keras/blob/933579d3c4a585f236982d05d3a74921f9567415/keras/src/callbacks/callback_list.py#L114
`_get_metrics_result_or_logs()` process `logs` if it's a dict, or return it as is.

and finally: https://github.com/keras-team/keras/blob/933579d3c4a585f236982d05d3a74921f9567415/keras/src/backend/tensorflow/trainer.py#L356

### DIFF
--- a/keras/src/backend/jax/trainer.py
+++ b/keras/src/backend/jax/trainer.py
@@ -554,7 +554,7 @@ class JAXTrainer(base_trainer.Trainer):
         self.make_test_function()
         self.stop_evaluating = False
         callbacks.on_test_begin()
-        logs = None
+        logs = {}
         self.reset_metrics()
 
         self._jax_state_synced = True
--- a/keras/src/backend/numpy/trainer.py
+++ b/keras/src/backend/numpy/trainer.py
@@ -276,7 +276,7 @@ class NumpyTrainer(base_trainer.Trainer):
         self.make_test_function()
         self.stop_evaluating = False
         callbacks.on_test_begin()
-        logs = None
+        logs = {}
         self.reset_metrics()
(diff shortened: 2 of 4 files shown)


fchollet (approved):
LGTM, thanks

PR #20126, opened 2024-08-15
Add bitwise ops

james77777778:
I need `left_shift` and `right_shift` in some cases.
After implementing them, I realized that it might be useful to add these bitwise ops.


fchollet (approved):
Very nice, that's definitely useful. Thank you for the contribution -- LGTM

PR #20130, opened 2024-08-16
Update operation.py...Backslash Error in name resolved!!

mathnebula186f:
BackSlash error in name resolved
Here was the error
ValueError: Argument `name` must be a string and cannot contain character `/`. Received: name=conv1/conv (of type <class 'str'>)

### DIFF
--- a/keras/src/ops/operation.py
+++ b/keras/src/ops/operation.py
@@ -17,14 +17,15 @@ class Operation:
     def __init__(self, dtype=None, name=None):
         if name is None:
             name = auto_name(self.__class__.__name__)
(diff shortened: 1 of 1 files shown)


fchollet:
The code in this PR has several syntax errors and as a result tests are failling.

At a higher level, however, it seems that it is better to raise an error that tells the user "names cannot have slashes in them", so the user can change the name they provided, than to silently convert slashes to underscores.

mathnebula186f:
Hi sir this silent conversion is already being done in optimizer.py file
but wasnt working for me and giving errors

Issue #20155, opened 2024-08-23
hasattr() raising ValueError when used with input and output attributes of a keras layer


Doch88:
The description is self-explanatory, when using hasattr() on the input attribute of a keras layer it raises an error when the layer has not been called yet. 
```
    if hasattr(first_layer, 'input'):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: The layer resizing has never been called and thus has no defined input.
```
This happens because "input" and "output" raise a ValueError, instead of a AttributeError as it should be.

PR #20156, opened 2024-08-23
_get_node_attribute_at_index() raising AttributeError instead of ValueError

Doch88:
https://github.com/keras-team/keras/issues/20155

This PR allows using hasattr() on node attributes, without raising an error.


fchollet (approved):

TwoCraZySaid4Ten (commented):
Why are you deleting things from my deceased son's files?

Issue #20166, opened 2024-08-26
This works in keras 2.15 with TensorFlow 2.15 but does not 3.5 with TensorFlow 2.17: multiple lookup layer types in inputs


jasonbrancazio:
In Keras 3, calling a model specified via the functional API with one StringLookupLayer and one IntegerLookupLayer each connected to their own Input layers raises a "Cast int64 to string" error, but in Keras 2, the tensors flow.

Here is a minimal example to reproduce the error. Try running it first in the latest version of Keras and TensorFlow, see the error, then downgrade and observe that it works:

```
import keras
import numpy as np
import tensorflow as tf

# Create a string lookup layer
data = [["a", "c", "d"], ["d", "z", "b"]]
letters_lookup = keras.layers.StringLookup(name='letters_lookup', output_mode='one_hot')
letters_lookup.adapt(data)

# Create an integer lookup layer
data = np.array([[12, 1138, 42], [42, 1000, 36]])
integers_lookup = keras.layers.IntegerLookup(name='integers_lookup', output_mode='one_hot')
integers_lookup.adapt(data)

# Create input layers
integers_input = keras.Input(shape=(), name='integers_in', dtype="int64")
letters_input = keras.Input(shape=(), name='letters_in', dtype="string")

# Make a model using the functional API
letters_out = letters_lookup(letters_input)
integers_out = integers_lookup(integers_input)
model = keras.Model([letters_input,integers_input],[letters_out,integers_out])

# Test the model
test_tensor_dict = {
    'letters_in': tf.constant(['d']),
    'integers_in': tf.constant([42], dtype=tf.int64)
}
model(test_tensor_dict)
```

```
2024-08-26 17:50:26.960422: W tensorflow/core/framework/op_kernel.cc:1817] OP_REQUIRES failed at cast_op.cc:122 : UNIMPLEMENTED: Cast int64 to string is not supported
2024-08-26 17:50:26.960536: I tensorflow/core/framework/local_rendezvous.cc:404] Local rendezvous is aborting with status: UNIMPLEMENTED: Cast int64 to string is not supported
---------------------------------------------------------------------------
UnimplementedError                        Traceback (most recent call last)
Cell In[14], line 1
----> 1 model(test_tensor_dict)

File /opt/conda/lib/python3.10/site-packages/keras/src/utils/traceback_utils.py:122, in filter_traceback.<locals>.error_handler(*args, **kwargs)
    119     filtered_tb = _process_traceback_frames(e.__traceback__)
[...]

Issue #20136, opened 2024-08-19
Keras 3 doesn't map dictionary inputs by "key" 


MeowTheCat:
The code below runs in Tensorflow 2.11 (keras 2) but not in tf-nightly (Keras  3.4.1 ).  I think Keras 3 doesn't map inputs by dict key

Epoch 1/10
Traceback (most recent call last):
  File "/home/wangx286/rnn-base-caller/base_caller/scripts/example_metric.py", line 32, in <module>
    model.fit({'before': x_train, 'after': y_train}, epochs=10, batch_size=32)
  File "/home/wangx286/miniconda3/envs/tf216/lib/python3.10/site-packages/keras/src/utils/traceback_utils.py", line 122, in error_handler
    raise e.with_traceback(filtered_tb) from None
  File "/home/wangx286/miniconda3/envs/tf216/lib/python3.10/site-packages/keras/src/models/functional.py", line 244, in _adjust_input_rank
    raise ValueError(
ValueError: Exception encountered when calling Functional.call().

**Invalid input shape for input Tensor("data_1:0", shape=(None,), dtype=float32). Expected shape (None, 20), but input has incompatible shape (None,)**

Arguments received by Functional.call():
  • inputs={'before': 'tf.Tensor(shape=(None, 20), dtype=float32)', 'after': 'tf.Tensor(shape=(None,), dtype=float32)'}
  • training=True
  • mask={'before': 'None', 'after': 'None'}

```
import numpy as np

import tensorflow as tf
from tensorflow.keras.layers import Dense
from tensorflow.keras.models import Model


# Define the model using the Functional API
x = tf.keras.Input(shape=(20,), name="before", dtype=tf.float32)
y = tf.keras.Input(shape=(), name="after", dtype=tf.float32)
tmp = Dense(64, activation='relu')(x)
outputs = Dense(1, activation='sigmoid')(tmp)


class DummyLossLayer(tf.keras.layers.Layer):
    def call(self, *x):
        self.add_loss(tf.keras.losses.BinaryCrossentropy(from_logits=True)(*x))
        return x

outputs, _ = DummyLossLayer()(outputs, y)
model = Model(inputs=[x, y], outputs=outputs)


# Compile the model with the custom metric
model.compile(optimizer='adam')

# Dummy data for demonstration
x_train = np.random.random((1000, 20))
y_train = np.random.randint(2, size=(1000,)).astype(np.float32)

# Train the model
[...]

Issue #20086, opened 2024-08-02
A possible bug with tf 2.17.0


tanwang2020:
Here are two versions of essentially the same code. The only difference is the lengths of the names of the keys. However, the first code generates error message while the second code does not. Here are the codes.
The first version: 
import numpy as np
import tensorflow as tf
from tensorflow.keras.constraints import non_neg
from tensorflow.keras.layers import Input, Dense, concatenate
from tensorflow.keras.models import Model
print('TensorFlow Version: ',tf.__version__)
# Example input data
X_train_dict = {
    'green_fin_const': np.random.rand(558, 3),
    'green_fin_inst': np.random.rand(558, 4),
    'gov_sup': np.random.rand(558, 5),
    'com_act': np.random.rand(558, 5),
    'eco_city': np.random.rand(558, 4),
    'type': np.random.rand(558, 1)
}
Y_train = np.random.rand(558, 2)

X_test_dict = {
    'green_fin_const': np.random.rand(140, 3),
    'green_fin_inst': np.random.rand(140, 4),
    'gov_sup': np.random.rand(140, 5),
    'com_act': np.random.rand(140, 5),
    'eco_city': np.random.rand(140, 4),
    'type': np.random.rand(140, 1)
}
Y_test = np.random.rand(140, 2)

# Define input layers
inputs_1 = Input(shape=(3,), name='green_fin_const')
inputs_2 = Input(shape=(4,), name='green_fin_inst')
inputs_3 = Input(shape=(5,), name='gov_sup')
inputs_4 = Input(shape=(5,), name='com_act')
inputs_5 = Input(shape=(4,), name='eco_city')
inputs_6 = Input(shape=(1,), name='type')

# Define dense layers
score_1 = Dense(1, activation='sigmoid', kernel_constraint=non_neg())(inputs_1)
score_2 = Dense(1, activation='sigmoid', kernel_constraint=non_neg())(inputs_2)
score_3 = Dense(1, activation='sigmoid', kernel_constraint=non_neg())(inputs_3)
score_4 = Dense(1, activation='sigmoid', kernel_constraint=non_neg())(inputs_4)
score_5 = Dense(1, activation='sigmoid', kernel_constraint=non_neg())(inputs_5)

# Concatenate scores and type input
concatenated_scores = concatenate([score_1, score_2, score_3, score_4, score_5, inputs_6])

# Define output layer
outputs = Dense(2, activation='softmax', kernel_constraint=non_neg())(concatenated_scores)

[...]

PR #20170, opened 2024-08-27
Add a warning for mismatched inputs structure in `Functional`

james77777778:
There are several issues related to mismatched inputs structure:
- #20166
- #20136
- #20086
and more...

As a result, it would be beneficial to add a warning for users.
Ultimately, we might want to raise an error when a mismatch occurs. Otherwise, it could lead to subtle issues if the inputs have the same shape and dtype, as the computation could be incorrect even though the code runs.

### DIFF
--- a/keras/src/models/functional.py
+++ b/keras/src/models/functional.py
@@ -168,7 +168,7 @@ class Functional(Function, Model):
         if mask is None:
             masks = [None] * len(inputs)
         else:
-            masks = self._flatten_to_reference_inputs(mask)
+            masks = tree.flatten(mask)
             for x, mask in zip(inputs, masks):
                 if mask is not None:
                     x._keras_mask = mask
@@ -205,8 +205,21 @@ class Functional(Function, Model):
     def _assert_input_compatibility(self, *args):
         return super(Model, self)._assert_input_compatibility(*args)
 
-    def _flatten_to_reference_inputs(self, inputs):
-        return tree.flatten(inputs)
+    def _maybe_warn_inputs_struct_mismatch(self, inputs):
+        try:
+            tree.assert_same_structure(
+                inputs, self._inputs_struct, check_types=False
+            )
+        except:
+            model_inputs_struct = tree.map_structure(
+                lambda x: x.name, self._inputs_struct
+            )
+            inputs_struct = tree.map_structure(lambda x: "*", inputs)
+            warnings.warn(
+                "The structure of `inputs` doesn't match the expected "
+                f"structure: {model_inputs_struct}. "
+                f"Received: the structure of inputs={inputs_struct}"
+            )
 
     def _convert_inputs_to_tensors(self, flat_inputs):
         converted = []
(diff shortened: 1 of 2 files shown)


fchollet (approved):
Thanks for the improvement -- LGTM

ThorvaldAagaard:
After upgrading Keras to latest version I am getting the warning, but can't figure out, what is wrong with the input

This is my code
![image](https://github.com/user-attachments/assets/f98f73b2-a05d-4950-952d-0e8191ab89b5)

and this is the output:
```
[<KerasTensor shape=(None, 42), dtype=float16, sparse=False, name=X_input>, <KerasTensor shape=(None, 15), dtype=float16, sparse=False, name=B_input>]                                                                                                                                            
x dtype: float16 shape: (1, 42)                                                                                                                                                                                                                                                                   
b dtype: float16 shape: (1, 15)                                                                                                                                                                                                                                                                   
UserWarning: The structure of `inputs` doesn't match the expected structure: ['X_input', 'B_input']. Received: the structure of inputs=('*', '*')                                             
```
Hard to see, what is wrong

I have tried with: result = model.predict({'X_input': x, 'B_input': b},verbose=0)
but got the same warning

LuigiCerone:
@ThorvaldAagaard Same here. Did you figure it out?

ThorvaldAagaard:
I did not find a fix, except downgrading Tensorflow to 2.17
I assume it is a bug in the implemenation, but the fix is merged, so perhaps noone else but us are reading this.

ThorvaldAagaard:
I found this, that can suppress all the warnings from Python
```
import warnings
warnings.filterwarnings("ignore")
```
It is working

LuigiCerone:
Simple script to reproduce the error:

```python
import numpy as np
from tensorflow.keras import layers, models

# Generate model
input_tensor_1 = layers.Input(shape=(10,), name='input_1')
input_tensor_2 = layers.Input(shape=(5,), name='input_2')

combined = layers.concatenate([input_tensor_1, input_tensor_2])

output = layers.Dense(1, activation='sigmoid')(combined)

model = models.Model(inputs=[input_tensor_1, input_tensor_2], outputs=output)
model.compile(optimizer='adam', loss='binary_crossentropy', metrics=['accuracy'])

# Train
num_samples = 1000

X1 = np.random.rand(num_samples, 10)
X2 = np.random.rand(num_samples, 5)
y = np.random.randint(2, size=num_samples)

model.fit([X1, X2], y, epochs=5, batch_size=32, verbose=0)

# Prediction
X1_new = np.random.rand(5, 10)
X2_new = np.random.rand(5, 5)

model.predict([X1_new, X2_new], verbose=0)
```

which gives the mentioned error:
```
UserWarning: The structure of `inputs` doesn't match the expected structure: ['input_1', 'input_2']. Received: the structure of inputs=('*', '*')
  warnings.warn(
```

By debugging, it seems a simple list/tuple mismatch in my case, given by:
```python
            tree.assert_same_structure(
                inputs, self._inputs_struct, check_types=False
            )
```

Fixed by using tuple as input for model:

```python
model = models.Model(inputs=(input_tensor_1, input_tensor_2), outputs=output)
```

james77777778:
Hey @ThorvaldAagaard , @LuigiCerone 
Thanks for the report. I have submitted a fix to suppress the warning for mismatched tuples and lists.

ThorvaldAagaard:
Looking forward to the fix, I did not find a workaround for the predict method

PR #20175, opened 2024-08-28
Mark layers with no state as built.

james77777778:
There is an issue reported by Matt where the QLoRA tutorial failed due to unbuilt layers.

The root cause is that the line `self.built = True` was removed from `__init__` in some layers, particularly in `Dropout` and `Softmax`.
(This line was originally added in this commit: https://github.com/keras-team/keras/commit/7748a6aebba4bbc77a7866c44c1293c4bb9e0808)

This PR adds an `assert_built_after_instantiation` in the test case to prevent future breakages.

cc @mattdangerw 

Note: there are still many stateless layers not included in this PR, but I want to keep this PR manageable for review. These can be addressed in the future.

### DIFF
--- a/keras/src/backend/torch/layer.py
+++ b/keras/src/backend/torch/layer.py
@@ -8,6 +8,12 @@ from keras.src.ops.operation import Operation
 
 
 class TorchLayer(torch.nn.Module):
+    @property
+    def torch_params(self):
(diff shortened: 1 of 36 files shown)


fchollet (approved):
LGTM, thanks for the fix.

PR #20176, opened 2024-08-28
Correct failed test case for model_visualization_test.py

shashaka:
When running model_visualization_test, I found that there's failed test case,and it shows error as below.

- keras/integration_tests/model_visualization_test.py"
```
plot_model(model, "nested-functional.png", expand_nested=True)

"dot" with args ['-Tpng', '/tmp/tmpv2ec_hvt'] returned code: -6

stdout, stderr:
 b''
b"failed at node 380[1]\ndot: maze.c:313: chkSgraph: Assertion `np->cells[1]' failed.\n"

AssertionError: "dot" with args ['-Tpng', '/tmp/tmpv2ec_hvt'] returned code: -6
```

An error occurred because some edges represent destinations that do not exist on the dot.
So, I added the logic to remove this kind of edge from dot.

### DIFF
--- a/keras/src/utils/model_visualization.py
+++ b/keras/src/utils/model_visualization.py
@@ -190,6 +190,14 @@ def make_node(layer, **kwargs):
     return node
 
 
+def remove_unused_edges(dot):
+    nodes = [v.get_name() for v in dot.get_nodes()]
+    for edge in dot.get_edges():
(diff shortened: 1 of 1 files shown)


fchollet (approved):
LGTM, thanks for the fix

I think this may have been due to a recent change in pydot -- the tests used to pass.

PR #20195, opened 2024-09-02
Added parameter 'verbose' when exporting TF SavedModel

Doch88:
When exporting a Keras model into a TF SavedModel, the current output is like this:

```
The following endpoints are available:

* Endpoint 'serve'
  inputs (POSITIONAL_OR_KEYWORD): ...
  training (POSITIONAL_OR_KEYWORD): ...
  mask (POSITIONAL_OR_KEYWORD): ...
Output Type:
  ...
Captures:
  * Huge list of variables, usually in this form: TensorSpec(shape=(), dtype=tf.resource, name=None) *
```
This can be quite overwhelming when you have to export models many times.
This PR adds a parameter that allows choosing whether to print all the variables when exporting.

### DIFF
--- a/keras/src/export/export_lib.py
+++ b/keras/src/export/export_lib.py
@@ -508,7 +508,7 @@ class ExportArchive:
 
         # Print out available endpoints
         endpoints = "\n\n".join(
-            _print_signature(getattr(self._tf_trackable, name), name, verbose)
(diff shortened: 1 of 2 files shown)


fchollet (approved):
Good idea -- thanks for the PR

PR #20205, opened 2024-09-04
Fix and improve the mechansim for removing metrics in `Trainer`

james77777778:
#20197 caused an `AttributeError` if the nested `Trainer` wasn't compiled beforehand.

This PR adds placeholders to fix the issue and simplify the logic in `compile` and `_clear_previous_trainer_metrics`

cc @mattdangerw


fchollet (approved):
LGTM, thanks

PR #20228, opened 2024-09-07
Fix test `test_output_mode_dict`

AryazE:
There was a problem with checking the backend (the `()` were forgotten), and hence the condition always evaluated to False.


fchollet (approved):
Thanks for the fix!

PR #20229, opened 2024-09-07
Improve performance of load and save

AryazE:
Searching for a key in a list is slower than searching it in a set.
In `saving/saving_lib.py`, the skip list is searched multiple times, and it can contain hundreds of items (in the tests it has 137 items).
Changing `get_attr_skiplist` to return a set allows faster search.


fchollet (approved):
Thanks!

PR #20231, opened 2024-09-08
Canonical type comparison

AryazE:
Since in Numpy we have `dtype('float64') == None`, type comparison should be done using `is` or `isinstance`.
```python
import numpy as np
a = np.array([1.1, 2.2])
print(a.dtype) # prints dtype('float64')
print(a.dtype == None) # prints True
```

This PR fixes that in `accuracy_metrics`.


AryazE:
In general type comparisons are recommended to be done using `is` instead of `==`. I have fixed some other places that this has happened too.

fchollet (approved):
LGTM, thanks for the fix!

PR #20232, opened 2024-09-08
Add `ops.trunc`

james77777778:
`ops.trunc` is required for the future implementation of int4 quantization


fchollet (approved):
Thanks for the PR -- LGTM

Issue #20221, opened 2024-09-06
Invalid reduction dimension (2 for input with 2 dimension(s) [Op:Sum]


MilkFiish:
This error occurs when I do dynamic execution work with `keras.layers.UnitNormalization`, but not when I do static inference.
The code below is executed in the tensorflow backend environment.
```python
import os
import re
import tensorflow as tf
import numpy as np
os.environ['KERAS_BACKEND']='tensorflow'
import keras

layer = keras.layers.UnitNormalization(
    axis=[ 1, 2 ],
    trainable=True,
    autocast=True,
)

result_static = layer.compute_output_shape([2, 3])

result_dynamic = layer(
    inputs=np.random.rand(*[2, 3]),
)
```

When I print result_static, it's [2, 3]. However, the dynamic part causes the error
```log
File ...\lib\site-packages\tensorflow\python\eager\execute.py:53, in quick_execute(op_name, num_outputs, inputs, attrs, ctx, name)
     51 try:
     52   ctx.ensure_initialized()
---> 53   tensors = pywrap_tfe.TFE_Py_Execute(ctx._handle, device_name, op_name,
     54                                       inputs, attrs, num_outputs)
     55 except core._NotOkStatusException as e:
     56   if name is not None:

InvalidArgumentError: Exception encountered when calling UnitNormalization.call().

{{function_node __wrapped__Sum_device_/job:localhost/replica:0/task:0/device:CPU:0}} Invalid reduction dimension (2 for input with 2 dimension(s) [Op:Sum]

Arguments received by UnitNormalization.call():
  • inputs=tf.Tensor(shape=(2, 3), dtype=float32)
```

By traceback information, it seems that an illegal input has been sent to the backend, and an input check may need to be added.

sanskarmodi8:
Hi @MilkFiish, 

I looked into your issue and found out that the error is due to a mismatch between the dimensionality of the input tensor and the axis specified in the UnitNormalization layer. 
The axis=[1, 2] argument is expecting a 3D tensor, but the input you are providing is a 2D tensor (shape=(2, 3)), which doesn't have the second dimension specified in the axis.

You should use np.random.rand(2, 3, 1) instead

MilkFiish:
If there is an input error, I thought it should have an error report for both static and dynamic execution, which is why I bring up the issue.
Hopefully, the team could add checks for such incorrect input, rather than reporting errors while dynamically executing to the backend and confusing users.

sanskarmodi8:
[...]

PR #20237, opened 2024-09-09
Solving issue #20221

sanskarmodi8:
Enhanced the compute_output_shape method by adding validation for input shape types and axis bounds. The method now ensures that input_shape is either a list or tuple, and that all axis values are within the valid range of input dimensions. This provides clearer error handling and prevents potential issues during shape computation.

Reason for the Change: Previously, the compute_output_shape method did not validate the input shape type or check if the specified axes were within bounds. This could lead to ambiguous or delayed errors during dynamic execution when invalid input shapes or out-of-bound axes were used, leading to confusion for the user. 
The updated validation improves robustness by catching these issues earlier in the process, providing more informative error messages to the user. 

Hence solving the issue #20221

### DIFF
--- a/keras/src/layers/normalization/unit_normalization.py
+++ b/keras/src/layers/normalization/unit_normalization.py
@@ -43,37 +43,6 @@ class UnitNormalization(Layer):
         return ops.normalize(inputs, axis=self.axis, order=2, epsilon=1e-12)
 
     def compute_output_shape(self, input_shape):
-        """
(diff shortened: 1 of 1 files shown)


ghsanti:
not sure if all norms can be bundled together, but it seems some verbose output shape computations are here https://github.com/keras-team/keras/blob/master/keras/src/ops/operation_utils.py

fchollet (approved):

PR #20266, opened 2024-09-17
Async logs - evaluate future inside the executor context

sampathweb:
Move the future evaluation to inside the with Executor context in which its created.


sampathweb:
Deleting this PR since its addressed offline.

PR #20280, opened 2024-09-23
Reimplement `eye` op on TF to support tensors as inputs.

hertschuh:
`N`, `M` and `k` can now be tensors and not just ints, which is useful when these parameters come from dynamic dimensions in a shape.


fchollet (approved):
Nice simplification as well!

PR #20286, opened 2024-09-25
Add `ops.nn.dot_product_attention`

james77777778:
I haven't check the performance yet, but I believe adding this operation will be beneficial since both torch and jax have optimized it in their codebases.

From a performance perspective, we should be able to replace `_compute_attention` in `MultiHeadAttention` with this op if the input shapes are strictly 4D.

EDITED:
It seems that CI is using `jax<=0.4.30` due to python version limitation. I implemented a pure numpy version of `dot_product_attention` for the unit tests.

For backends:
- jax: Uses `jax.nn.dot_product_attention` if available. Otherwise, adapts the impl from `jax==0.4.33`.
- numpy: Adapts the impl from `jax==0.4.31` (no customizable `vmap`)
- tensorflow: Adapts the impl from `jax==0.4.31` (no customizable `vmap`)
- torch: Uses `torch.nn.functional.scaled_dot_product_attention`

### DIFF
--- a/keras/api/_tf_keras/keras/ops/__init__.py
+++ b/keras/api/_tf_keras/keras/ops/__init__.py
@@ -67,6 +67,7 @@ from keras.src.ops.nn import conv_transpose
 from keras.src.ops.nn import ctc_decode
 from keras.src.ops.nn import ctc_loss
 from keras.src.ops.nn import depthwise_conv
+from keras.src.ops.nn import dot_product_attention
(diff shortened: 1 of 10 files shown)


fchollet (approved):
Looking good, thank you for the contribution! It seems we need to update our JAX version number on GPU CI.

james77777778:
The main barrier should be the python version.
jax>=0.4.31 only supports python>=3.10.

Issue #20324, opened 2024-10-04
Reason for the recently added shape restriction in MultiHeadAttention


aarbabi:
Hello,

Wondering why is there a restriction on the input shape of `query` and `value` to have a matching final dimension?

This blocks having cross-attention to a source that has a different shape than query, unless adding an extra projection layer. Given that all input tensors (`query`, `key`, `value`) are immediately projected by dense layers inside `MultiHeadAttention`, I don't think any restriction on final dims is necessary.

For reference, the [pytorch doc](https://keras.io/api/layers/attention_layers/multi_head_attention/) for `MultiHeadAttention` explicitly uses 3 distinct variables to describe expected dimensions for the three tensors. The tensorflow implementation does not enforce such restriction as well.

The restriction is enforced here: https://github.com/keras-team/keras/blob/5aa5f88dc200bbf2cd765d5a213c23c58da48e80/keras/src/layers/attention/multi_head_attention.py#L214-L219

And was added as part of the PR #19973 (in response to the issue #19769)

Thanks

aarbabi:
CC @mattdangerw

mehtamansi29:
Hi @aarbabi -

[MultiHeadAttention](https://keras.io/api/layers/attention_layers/multi_head_attention/) documentation shows regarding shape of query,key and value. 
Query  shapes are (batch_size, <query dimensions>, key_dim), 
key shapes are (batch_size, <key/value dimensions>, key_dim) and 
value shapes are (batch_size, <key/value dimensions>, value_dim).

[Here](https://github.com/keras-team/keras/blob/5aa5f88dc200bbf2cd765d5a213c23c58da48e80/keras/src/layers/attention/multi_head_attention.py) you can find description regarding shape of query,key and value arguments.

aarbabi:
@mehtamansi29 Thanks for getting back to this.

Looking at the documentation:

> This layer first projects query, key and value. These are (effectively) a list of tensors of length num_attention_heads, where the corresponding shapes are (batch_size, <query dimensions>, key_dim), (batch_size, <key/value dimensions>, key_dim), (batch_size, <key/value dimensions>, value_dim).

I believe these shapes belong to the tensors **after the projection**, where `key_dim` and `value_dim` are the args passed to the constructor. It makes sense that `query` and `key` to have the same dim **after the projection**, but there shouldn't be any restrictions on the input tensors to MultiHeadAttention.

[...]

PR #20340, opened 2024-10-11
Remove MultiHeadAttention assertion that last dims of Q and V need to be equal

jeffcarp:
Via #20324.

This assertion was added in #19973, however re: #20324 it seems overly restrictive. It was not covered by an existing test case.


fchollet (approved):
Thanks for the PR! The TF-GPU failure looks like it should be unrelated.

PR #20368, opened 2024-10-17
[Bounding Boxes] Add Support for Bounding Box Transformations during Image Resizing

sineeli:
This PR covers below cases:

1. Add bounding box transformations: Here we require `orig_height` and `orig_width` for bounding boxes transformation as each image can have separate bounding boxes relative to the original image size.  Later we can use the same to transform bounding boxes as per the ratios.
2. Make `convert_format` to be compatible in `tf.data` pipeline regardless of backend.

Just a temporary file without disturbing previous implementation of `convert_format`


keras/src/layers/preprocessing/image_preprocessing/bounding_boxes/converters_new.py
```diff
@@ -0,0 +1,102 @@
+from keras.src.api_export import keras_export
```
fchollet, line 1: Why converters_new? why not just converters?

fchollet (commented):
Thanks for the PR!

keras/src/layers/preprocessing/image_preprocessing/bounding_boxes/converters_new.py
sineeli: Yes, just wanted a initial review so that I can replace the existing one.

keras/src/layers/preprocessing/image_preprocessing/base_image_preprocessing_layer.py
```diff
@@ -64,7 +64,12 @@ def transform_labels(self, labels, transformation, training=True):
         raise NotImplementedError()
 
     def transform_bounding_boxes(
-        self, bounding_boxes, transformation, training=True
+        self,
+        bounding_boxes,
+        orig_height,
```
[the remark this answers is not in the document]
sineeli: Sure, I will update the respective layers.

keras/src/layers/preprocessing/image_preprocessing/base_image_preprocessing_layer.py
```diff
@@ -144,17 +158,47 @@ def call(self, data, training=True):
                         "`bounding_box_format='xyxy'`."
                     )
                 bounding_boxes = densify_bounding_boxes(
-                    data["bounding_boxes"], backend=self.backend
+                    data["bounding_boxes"],
+                    is_batched=is_batched,
+                    backend=self.backend,
                 )
+                if "orig_width" not in data:
```
fchollet, line 165: What is the implication here -- do users have to specify this key in the data?

keras/src/layers/preprocessing/image_preprocessing/bounding_boxes/bounding_box.py
```diff
@@ -0,0 +1,519 @@
...
+
+        Relative formats, abbreviated `rel`, make use of the shapes of the
+        `images` passed. In these formats, the coordinates, widths, and heights
+        are all specified as percentages of the host image.
+
+        Example:
+
+        ```python
+        boxes = {
+            "boxes": [TODO],
```
fchollet, line 58: Please fill in

keras/src/layers/preprocessing/image_preprocessing/base_image_preprocessing_layer.py
fchollet: IMO we should be able to read this information on the input images and perhaps even pass it as part of the `transformation` argument rather than as separate arguments.

sineeli: As we are processing images in batch the images can have varying sizes and the shape of images can have `(None, None, None, 3)`.  So included in broadcasting way by adding extra keys in input dict.

Other way around is basically process single image at a time and then we can have raw image height and width dynamically then we don't need these extra data specific keys. (torchvision Object Detection model follow these). Then batched after this and sent to models.

In this case we have to make sure resize layer with bounding boxes work with only unbatched images and transform boxes at a image level

fchollet: > As we are processing images in batch the images can have varying sizes and the shape of images can have (None, None, None, 3) and bounding_boxes dimensions are respective to raw image

This isn't the case, by the time we hit the layer we are expected dense batches (no ragged tensors). That means all images in the batch have the same size.

sineeli: Oh okay, I was keeping in mind the ragged case as well, so just modified the scenario where if user provides in `ragged`, It will raise the error in transformation. They can use it before the ragged batch and do the resizing without issue. 

I will test now with different scenarios as well so that I don't miss the cases and add them as test cases as well.

keras/api/_tf_keras/keras/utils/bounding_boxes/__init__.py
```diff
@@ -0,0 +1,21 @@
+"""DO NOT EDIT.
+
+This file was autogenerated. Do not edit it by hand,
+since your modifications would be overwritten.
+"""
+
+from keras.src.layers.preprocessing.image_preprocessing.bounding_boxes.converters import (
+    affine,
```
fchollet, line 8: Please rename to affine_transform

keras/api/_tf_keras/keras/utils/bounding_boxes/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+)
+from keras.src.layers.preprocessing.image_preprocessing.bounding_boxes.converters import (
+    clip_to_images,
```
fchollet, line 11: I wonder if this would be more usable via `clip_to_image_size(boxes, image_size)`, e.g. `clip_to_size(boxes, images.shape[1:3])`? You may not always have image tensors on hand, so it's nice to be able to do `clip_to_size(boxes, (12, 34))`.

keras/src/layers/preprocessing/image_preprocessing/resizing.py
```diff
@@ -112,10 +118,78 @@ def transform_segmentation_masks(
...
+        if isinstance(data, dict):
+            input_shape = self.backend.shape(data["images"])
+        else:
+            input_shape = self.backend.shape(data)
+
+        input_height, input_width = (
+            input_shape[self.height_axis],
+            input_shape[self.width_axis],
+        )
+        if input_height is None or input_width is None:
```
fchollet, line 132: This would never actually occur; in the TF dynamic shape case you'd get int tensors in the shape

fchollet (commented):
Looking good, thanks for the updates! Please add a unit test for `Resizing` bounding box support.

keras/api/_tf_keras/keras/utils/bounding_boxes/__init__.py
sineeli: Yes, name seems misleading it actually takes height and width not the image tensors. I will change the name to `clip_to_image_size` which is apt actually.

keras/src/layers/preprocessing/image_preprocessing/max_bounding_box.py
```diff
@@ -0,0 +1,89 @@
+from keras.src.api_export import keras_export
+from keras.src.layers.preprocessing.image_preprocessing.base_image_preprocessing_layer import (  # noqa: E501
+    BaseImagePreprocessingLayer,
+)
+
+
+@keras_export("keras.layers.MaxBoundingBox")
+class MaxBoundingBox(BaseImagePreprocessingLayer):
```
fchollet, line 8: Do we really need this? What's the use case? I get the feeling this should not be a layer.

fchollet (commented):
The code looks good to me! Not sure about the max bbox layer.

A side question is whether we need "bbox" in the name of the transformations like `affine_transform` etc et distinguish them from `keras.ops.image.affine_transform` and so on. No strong opinion.

keras/src/layers/preprocessing/image_preprocessing/max_bounding_box.py
sineeli: Once we resize at each image we can have any number of bounding boxes, so we can just use  this layer and use with `.map` as part of data pipeline.
```python3
image_size = (800, 800)
batch_size = 4

resize = keras.layers.Resizing(
    height=image_size[0], 
    width=image_size[1], 
    pad_to_aspect_ratio=True, 
    bounding_box_format="xyxy"
)
max_box = keras.layers.MaxBoundingBox(
    100, 
    bounding_box_format="xyxy"
)

train_ds = tfds.load("voc/2007", split="train")
train_ds = train_ds.map(decode_tfds, num_parallel_calls=tf.data.AUTOTUNE)
train_ds = train_ds.map(max_box, num_parallel_calls=tf.data.AUTOTUNE)
train_ds = train_ds.map(resize, num_parallel_calls=tf.data.AUTOTUNE)
train_ds = train_ds.batch(batch_size, drop_remainder=True)
train_ds = train_ds.prefetch(tf.data.AUTOTUNE)'
```

sineeli: image
image

This is how the results looks when pad to aspect_ratio is enabled

sineeli:
Reference Colab: https://colab.research.google.com/gist/sineeli/77cdc067d0adcef0a267b00b4921e7b8/-keras3-resizing.ipynb

keras/src/layers/preprocessing/image_preprocessing/max_bounding_box.py
fchollet: Sounds good, please rename to `MaxNumBoundingBoxes` for explicitness

keras/src/layers/preprocessing/image_preprocessing/max_bounding_box.py
```diff
@@ -0,0 +1,89 @@
...
+    """Ensure the maximum number of bounding boxes.
+
+    Args:
+        max_number: Desired output number of bounding boxs.
```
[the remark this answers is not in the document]
sineeli: Done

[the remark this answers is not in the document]
fchollet: Did you push your changes yet? Not seeing them in the PR

keras/src/layers/preprocessing/image_preprocessing/max_bounding_box.py
sineeli: Done

keras/src/layers/preprocessing/image_preprocessing/max_bounding_box.py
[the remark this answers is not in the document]
sineeli: Just pushed all the changes with test cases, please let me know if we have to add any more test cases. 

Thanks for the review!

sineeli:
> Can you take a look at the test failures? I think they are likely related to the fact that torch tests are running with `data_format="channels_first"`. https://github.com/keras-team/keras/actions/runs/11469236802/job/31916066253?pr=20368

Oh yes!! thanks I updated the test cases for `channels_first` case.

Reference:
1. [Bounding Boxes with `pad_to_aspect_ratio=True`](https://colab.research.google.com/gist/sineeli/9149200010d20e3fffb1eba19f5f0972/-keras3-resizing-bounding-boxes-with-pad-to-aspect-ratio.ipynb)
2. [Bounding Boxes with `crop_to_aspect_ratio=True`](https://colab.research.google.com/gist/sineeli/77dd9ca451edcd237755bd68283b309b/-keras3-resizing-bounding-boxes-with-crop-to-aspect-ratio.ipynb)
3. [Normal Resize](https://colab.research.google.com/gist/sineeli/576a806e32ffa587ad19cd9003ea971c/-keras3-resizing.ipynb)

sineeli:
> The code looks good to me! Not sure about the max bbox layer.
> 
> A side question is whether we need "bbox" in the name of the transformations like `affine_transform` etc et distinguish them from `keras.ops.image.affine_transform` and so on. No strong opinion.

I guess it should be fine as user will use `keras.utils.bounding_boxes.affine_transform` and it is explicit that its under `bounding_boxes`.

fchollet (approved):

PR #20384, opened 2024-10-20
Add celu activation layer for each nn.py

shashaka:
I noticed that the CELU activation function is not currently supported in Keras, even though it is available in the Torch and JAX frameworks. To address this, I implemented the CELU layer logic for both TensorFlow and NumPy, while using the existing implementations for JAX and Torch.

If this approach seems acceptable, I would be happy to update ops.nn.py to include support for the CELU activation in the Keras backend as well.

 [CeLU](https://arxiv.org/abs/1704.07483)

### DIFF
--- a/keras/api/_tf_keras/keras/activations/__init__.py
+++ b/keras/api/_tf_keras/keras/activations/__init__.py
@@ -7,6 +7,7 @@ since your modifications would be overwritten.
 from keras.src.activations import deserialize
 from keras.src.activations import get
 from keras.src.activations import serialize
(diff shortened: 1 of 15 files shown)


fchollet (commented):
Thanks for the PR! Sounds good -- please also add it in ops/nn.py and activations.py then.

shashaka:
I added the implementation in ops/nn.py and activations.py. Additionally, I included corresponding test cases for both files.

fchollet (approved):
LGTM, thank you

PR #20398, opened 2024-10-23
Fix failing CI for coverage testing

nicolaspi:
Fixes https://github.com/keras-team/keras/issues/20397
Reference https://github.com/nedbat/coveragepy/issues/512


fchollet (approved):
Nice, thank you for the fix!

PR #20405, opened 2024-10-24
Add hard_tanh activation function

shashaka:
Since Keras does not provide built-in support for the hard_tanh activation function, I implemented it using TensorFlow and NumPy backends, leveraging the existing implementations in JAX and Torch for consistency.


fchollet (approved):
Thanks for the addition. LGTM

PR #20406, opened 2024-10-25
Fix serialization / deserialization.

hertschuh:
- Serialization was not taking the registered name and package from the registry.
- Deserialization was selecting symbols by postfix as a fallback.


fchollet (approved):
LGTM -- thank you for the fix!

PR #20416, opened 2024-10-26
Add log_sigmoid activation

shashaka:
Although an implementation for log_sigmoid already exists, it was not declared in activations.py. I have added it to activations.py and included corresponding test cases.


fchollet (approved):
LGTM, that's a fine addition to our collection.

PR #20433, opened 2024-10-31
Add `from_sorted_ids` option to `SparseTopKCategoricalAccuracy`.

hertschuh:
to consume sorted IDs of top N categories instead of scores for all categories.


fchollet (approved):
LGTM -- thank you!

PR #20439, opened 2024-11-01
Use temporary folder for testing model saving in file editor

sampathweb:
Fixes the test failiure when the local folder is not writtable in test runs.


hertschuh (approved):
Thanks, Ramesh!

fchollet (approved):
LGTM, thank you

PR #20445, opened 2024-11-04
Replace Black with Ruff formatter

mwtoews:
This PR follows on from #20442 to replace Black with Ruff as the Python code formatter.

Some manual code changes were applied to minimize automated changes or hopefully better represent the logic.

With this transition, `examples` is not excluded, as there was only a minor change to files in this directory.

Some shell scripts (in `shell/`) are modified to remove the [`-e` flag](https://www.gnu.org/software/bash/manual/bash.html#Modifying-Shell-Behavior), and `shell/lint.sh` is modified to run and accumulate error codes while running the whole script (i.e. 0=good, 1=one failure, 2=two failures).

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -10,7 +10,7 @@
                     "source.organizeImports": true
                 },
                 "[python]": {
-                    "editor.defaultFormatter": "ms-python.black-formatter"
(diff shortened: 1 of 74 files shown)


fchollet (approved):
LGTM -- thanks!

PR #20448, opened 2024-11-04
Enable flash attention

divyashreepathihalli:
This PR 

- [x] refactors the MHA layer so that its `_compute_attention` method would just call `ops.dot_production_attention`
- [x] Adds a global toggle `keras.config.enable_flash_attention` and `keras.config.is_flash_attention_enabled`
- [x] Modify attention mask to match `_masked_softmax` implementation

### DIFF
--- a/keras/src/backend/config.py
+++ b/keras/src/backend/config.py
@@ -15,6 +15,8 @@ _IMAGE_DATA_FORMAT = "channels_last"
 # Default backend: TensorFlow.
 _BACKEND = "tensorflow"
 
+_ENABLE_FLASH_ATTENTION = False
+
 
 @keras_export(["keras.config.floatx", "keras.backend.floatx"])
 def floatx():
@@ -274,3 +276,23 @@ def backend():
 
     """
     return _BACKEND
+
+
+@keras_export("keras.config.enable_flash_attention")
+def enable_flash_attention(value):
+    """Enable flash attention.
+
+    Flash attention offers performance optimization for attention layers,
+    making it especially useful for large language models (LLMs) that
+    benefit from faster and more memory-efficient attention computations.
+
+    Once enabled, supported layers like `MultiHeadAttention` will
+    use flash attention for faster computations.
+    """
+    global _ENABLE_FLASH_ATTENTION
+    _ENABLE_FLASH_ATTENTION = value
+
+
+@keras_export("keras.config.is_flash_attention_enabled")
+def is_flash_attention_enabled():
+    return _ENABLE_FLASH_ATTENTION
--- a/keras/src/backend/config.py
+++ b/keras/src/backend/config.py
@@ -15,8 +15,6 @@ _IMAGE_DATA_FORMAT = "channels_last"
 # Default backend: TensorFlow.
 _BACKEND = "tensorflow"
 
-_ENABLE_FLASH_ATTENTION = False
-
 
 @keras_export(["keras.config.floatx", "keras.backend.floatx"])
 def floatx():
@@ -276,23 +274,3 @@ def backend():
 
     """
     return _BACKEND
-
-
-@keras_export("keras.config.enable_flash_attention")
-def enable_flash_attention(value):
-    """Enable flash attention.
-
-    Flash attention offers performance optimization for attention layers,
-    making it especially useful for large language models (LLMs) that
-    benefit from faster and more memory-efficient attention computations.
-
-    Once enabled, supported layers like `MultiHeadAttention` will
-    use flash attention for faster computations.
-    """
-    global _ENABLE_FLASH_ATTENTION
-    _ENABLE_FLASH_ATTENTION = value
-
-
-@keras_export("keras.config.is_flash_attention_enabled")
-def is_flash_attention_enabled():
-    return _ENABLE_FLASH_ATTENTION
--- a/keras/src/layers/attention/multi_head_attention.py
+++ b/keras/src/layers/attention/multi_head_attention.py
@@ -541,6 +541,7 @@ class MultiHeadAttention(Layer):
             attention_mask=attention_mask,
             use_causal_mask=use_causal_mask,
         )
+
         attention_output, attention_scores = self._compute_attention(
             query_dense,
             key_dense,
@@ -549,7 +550,7 @@ class MultiHeadAttention(Layer):
             attention_mask,
             training,
         )
-        attention_output = self._output_dense(attention_output)
+        attention_output = self._output_dense.call(attention_output)
 
         if return_attention_scores:
             return attention_output, attention_scores
--- a/keras/src/layers/attention/multi_head_attention.py
+++ b/keras/src/layers/attention/multi_head_attention.py
@@ -11,7 +11,6 @@ from keras.src import ops
 from keras.src import regularizers
 from keras.src.api_export import keras_export
 from keras.src.layers.activations.softmax import Softmax
-from keras.src.layers.attention.attention import enable_flash_attention
 from keras.src.layers.attention.attention import is_flash_attention_enabled
 from keras.src.layers.core.einsum_dense import EinsumDense
 from keras.src.layers.layer import Layer
@@ -430,17 +429,13 @@ class MultiHeadAttention(Layer):
           attention_output: Multi-headed outputs of attention computation.
           attention_scores: Multi-headed attention weights.
         """
-        print("### compute attention", self._flash_attention)
-        print("### compute attention", is_flash_attention_enabled())
         if self._flash_attention and return_attention_scores:
             raise ValueError(
                 "Returning attention scores is not supported when flash "
                 "attention is enabled. Please disable flash attention to access"
                 " attention scores."
             )
-        if self._flash_attention:
-            enable_flash_attention(True)
-        if is_flash_attention_enabled():
+        if not return_attention_scores:
             # Directly compute the attention output using flash attention
             attention_output = ops.dot_product_attention(
                 query=query,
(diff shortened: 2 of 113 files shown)


keras/src/backend/config.py
```diff
@@ -274,3 +276,23 @@ def backend():
...
+
+    Once enabled, supported layers like `MultiHeadAttention` will
+    use flash attention for faster computations.
+    """
+    global _ENABLE_FLASH_ATTENTION
+    _ENABLE_FLASH_ATTENTION = value
+
+
+@keras_export("keras.config.is_flash_attention_enabled")
+def is_flash_attention_enabled():
```
fchollet, line 297: Please add a docstring here as well.

keras/src/backend/config.py
```diff
@@ -274,3 +276,23 @@ def backend():
...
+
+    Flash attention offers performance optimization for attention layers,
+    making it especially useful for large language models (LLMs) that
+    benefit from faster and more memory-efficient attention computations.
+
+    Once enabled, supported layers like `MultiHeadAttention` will
+    use flash attention for faster computations.
+    """
+    global _ENABLE_FLASH_ATTENTION
+    _ENABLE_FLASH_ATTENTION = value
```
fchollet, line 293: This is needs to be threadlocal. Instead of doing it like this, use `set_global_attribute`/`get_global_attribute` from `keras.src.backend.common.global_state`. See how other global flags are implemented.

keras/src/layers/attention/multi_head_attention.py
```diff
@@ -415,9 +422,28 @@ def _compute_attention(
...
+                value=value,
+                mask=attention_mask,
+                scale=self._inverse_sqrt_key_dim,
+                is_causal=bool(self._attention_axes),
+            )
+            # Return only the attention output, as scores are not separately
+            # available
+            return attention_output, None
+
+        # Default behavior without flash attention, with explicit attention
```
fchollet, line 445: Can we always go to `dot_product_attention` unless `return_attention_scores` is enabled?

keras/src/layers/attention/multi_head_attention.py
```diff
@@ -392,7 +393,13 @@ def _masked_softmax(self, attention_scores, attention_mask=None):
         return self._softmax(attention_scores, mask=attention_mask)
 
     def _compute_attention(
-        self, query, key, value, attention_mask=None, training=None
```
fchollet, line 395: Add a MHA layer constructor argument (set to `None` as default) for `flash_attention`.

None -> use global default
True, False -> overrides global default

keras/src/layers/attention/multi_head_attention_test.py
```diff
@@ -12,10 +12,12 @@
 from keras.src import models
 from keras.src import saving
 from keras.src import testing
+from keras.src.backend import config
 
 
 class MultiHeadAttentionTest(testing.TestCase):
     def test_basics(self):
+        config.enable_flash_attention(True)
```
fchollet, line 20: Add a numerical correctness with and without FA.

fchollet (commented):
Thanks for the PR!

keras/src/backend/config.py
divyashreepathihalli: importing `global_state` inconfig created circular import error so I have moved the configs to attention.py

keras/src/layers/attention/multi_head_attention.py
```diff
@@ -131,6 +133,12 @@ def __init__(
         self._activity_regularizer = regularizers.get(activity_regularizer)
         self._kernel_constraint = constraints.get(kernel_constraint)
         self._bias_constraint = constraints.get(bias_constraint)
+        if attention.is_flash_attention_enabled is None:
+            if use_flash_attention is not None:
```
fchollet, line 137: Do `self.use_flash_attention = use_flash_attention or is_flash_attention_enabled()`

keras/src/layers/attention/multi_head_attention.py
```diff
@@ -11,6 +11,7 @@
 from keras.src import regularizers
 from keras.src.api_export import keras_export
 from keras.src.layers.activations.softmax import Softmax
+from keras.src.layers.attention import attention
```
fchollet, line 14: Import the symbol directly it will be more readable

keras/src/layers/attention/multi_head_attention.py
```diff
@@ -415,9 +429,28 @@ def _compute_attention(
           attention_output: Multi-headed outputs of attention computation.
           attention_scores: Multi-headed attention weights.
         """
-        # Note: Applying scalar multiply at the smaller end of einsum improves
-        # XLA performance, but may introduce slight numeric differences in
-        # the Transformer attention head.
+        if attention.is_flash_attention_enabled() and return_attention_scores:
```
fchollet, line 432: Use `self. use_flash_attention`

keras/src/layers/attention/multi_head_attention.py
```diff
@@ -415,9 +429,28 @@ def _compute_attention(
...
+            raise ValueError(
+                "Returning attention scores is not supported when flash "
+                "attention is enabled. Please disable flash attention to access"
+                " attention scores."
+            )
+        if attention.is_flash_attention_enabled():
```
fchollet, line 438: To always use dot_production attention, do

```
if not return_attention_scores:
    return ops.dot_product_attention(..., flash_attention=self.use_flash_attention), None

File: keras/src/layers/attention/multi_head_attention.py [hunk 10]
```diff
@@ -104,6 +105,7 @@ def __init__(
         use_bias=True,
         output_shape=None,
         attention_axes=None,
+        use_flash_attention=None,
```
fchollet, line 108: Please document it in the docstring. Also I think it should just be `flash_attention` to match the op API

keras/src/layers/attention/multi_head_attention.py
```diff
@@ -415,9 +429,29 @@ def _compute_attention(
           attention_output: Multi-headed outputs of attention computation.
           attention_scores: Multi-headed attention weights.
         """
-        # Note: Applying scalar multiply at the smaller end of einsum improves
-        # XLA performance, but may introduce slight numeric differences in
-        # the Transformer attention head.
+        if self._flash_attention and return_attention_scores:
```
fchollet, line 432: Need a similar error for dropout.

keras/src/layers/attention/multi_head_attention.py
```diff
@@ -415,9 +429,29 @@ def _compute_attention(
...
+                scale=self._inverse_sqrt_key_dim,
+                is_causal=use_causal_mask,
+                flash_attention=self._flash_attention,
+            )
+            # Return only the attention output, as scores are not separately
+            # available
+            return attention_output, None
+
+        # Default behavior without flash attention, with explicit attention
+        # scores
```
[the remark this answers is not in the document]
divyashreepathihalli: The 5d input tests fail if we do let  the code execute `ops.dot_product_attention` with `flash_attention=False` because it only accepts 4d inputs.  
Should we also just allow 4d inputs for MHA?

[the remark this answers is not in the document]
fchollet: Yes, so basically:

use_dpa_op = no dropout, no scores, ndim ==4
if use_dpa_op:
   return dot_product_attention(...)
else:
...

fchollet:
PyTorch has super weird constraints. PyTorch operations often assume tensors are stored in contiguous chunks of memory -- and this isn't automatically managed by the framework. A possible fix for the failing torch GPU test:

```
attention_output = torch.nn.functional.scaled_dot_product_attention(
    query, key.contiguous(), value, attn_mask=mask, is_causal=is_causal, scale=scale
)
```

keras/src/layers/attention/multi_head_attention_test.py
```diff
@@ -189,12 +264,23 @@ def test_initializer(self):
...
+                [[1, 2, 3, 0, 0], [3, 3, 1, 1, 2], [1, 0, 0, 0, 0]]
+            )
+            masked_query = layers.Embedding(4, 8, mask_zero=True)(query)
+            value = np.random.normal(size=(3, 3, 8))
+            output = layer(query=masked_query, value=value)
+        except RuntimeError as e:
+            if e.args[0].startswith(
+                "(*bias): last dimension must be contiguous"
+            ):
+                self.skipTest(
```
fchollet, line 280: I want to make sure that we're not breaking any existing behavior. Can we still use the MHA layer (no flash attention) with Pytorch on GPU?

fchollet (approved):

keras/src/layers/attention/multi_head_attention.py
```diff
@@ -392,7 +399,13 @@ def _masked_softmax(self, attention_scores, attention_mask=None):
         return self._softmax(attention_scores, mask=attention_mask)
 
     def _compute_attention(
-        self, query, key, value, attention_mask=None, training=None
+        self,
+        query,
+        key,
+        value,
+        return_attention_scores,
```
SamanehSaadat, line 406: @divyashreepathihalli shouldn't `return_attention_scores` have a default value here?
There might be many instances of `_compute_attention` call that don't pass `return_attention_scores` value and I think this change can break them. There is an example [here](https://github.com/keras-team/keras-hub/blob/5a7ecb644fe3ec5b04519846c2c5c54ad58fbe12/keras_hub/src/layers/modeling/cached_multi_head_attention.py#L111-L117). So I was wondering if it's possible to set a default value here so that other references of `_compute_attention` work as before.
I think this change is probably the reason that the test is failing in https://github.com/keras-team/keras-hub/pull/1977

SamanehSaadat: Isn't there any other way to check and make sure the users pass the arg value (rather than lack of default value here)?

divyashreepathihalli: I think a work around would be to add a `self._return_attention_scores` and then set it in the call method and use it in `_compute_attention`. wdyt?

PR #20453, opened 2024-11-05
Fix for https://github.com/keras-team/keras/issues/20425

hertschuh:
The issue was caused by the fact that the iterator was not fully consumed and `on_epoch_end` was not called.

Added an exception to catch this situation in the future.

Added a unit test to test `model.fit()` with all the combinations of data adapters.


fchollet (approved):
LGTM -- thank you for the fix!

PR #20456, opened 2024-11-06
Suppress warnings for mismatched tuples and lists in functional models

james77777778:
Fix #20170

It should be acceptable to suppress warnings for mismatched tuples and lists in functional models, as the TF backend will automatically convert lists to tuples for input data.


fchollet (approved):
LGTM - Thanks!

PR #20461, opened 2024-11-07
Fix steps for `TensorBoard` callback for evaluation and batch metrics.

hertschuh:
This bug caused batch level metrics and evaluation metrics to all be reported for step 0, which would not show a graph.

Epoch level metrics, were not affected by this bug.

The fix was tested manually on a training run.


fchollet (approved):
Thanks for the fix!

PR #20476, opened 2024-11-09
Update CompileLoss to report unweighted metric values (breaking change)

### LINKED ISSUE #20343: CompileLoss should use non loss weighted metric.update_state
Proposing that we should refactor `CompileLoss` to report non loss weighted metric as was done in `keras 2`

here is the suggested refactor:

```
diff --git a/keras/src/trainers/compile_utils.py b/keras/src/trainers/compile_utils.py
index 410a782db..71d37bde3 100644
--- a/keras/src/trainers/compile_utils.py
+++ b/keras/src/trainers/compile_utils.py
@@ -658,14 +658,15 @@ class CompileLoss(losses_module.Loss):
                 value = ops.cast(
                     loss_fn(y_t, y_p, sample_weight), dtype=self.dtype
                 )
-                if loss_weight is not None:
-                    value = ops.multiply(value, loss_weight)
-                loss_values.append(value)
                 # Record individual losses.
                 if metric:
                     metric.update_state(
                         value, sample_weight=tree.flatten(y_p)[0].shape[0]
                     )
+                if loss_weight is not None:
+                    value = ops.multiply(value, loss_weight)
+                loss_values.append(value)
```

jeffcarp:
Fixes #20343. Thanks to rivershah@ for pointing this out.

This changes CompileLoss metrics to report the values before weights are applied, which brings it in line with Keras 2 behavior.

### DIFF
--- a/keras/src/trainers/compile_utils.py
+++ b/keras/src/trainers/compile_utils.py
@@ -581,7 +581,7 @@ class CompileLoss(losses_module.Loss):
                     raise ValueError(
                         f"`loss_weights` must match the number of losses, "
                         f"got {len(tree.flatten(loss))} losses "
-                        f"and {len(loss_weights)} weigths."
+                        f"and {len(loss_weights)} weights."
                     )
                 loss_weights = tree.pack_sequence_as(loss, flat_loss_weights)
             loss = tree.map_structure(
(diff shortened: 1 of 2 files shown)


fchollet (approved):
LGTM, thank you

PR #20480, opened 2024-11-11
Add tanh_shrink activation

shashaka:
Since Keras lacks built-in support for the tanh_shrink activation function, I implemented it across TensorFlow, JAX, and NumPy, aiming to maintain consistency with existing implementations available in PyTorch.


fchollet (approved):
LGTM, thanks for the addition!

shashaka:
I'll work on implementing the bounding box transformation further. However, I've noticed there are additional steps involved in preprocessing, like affine transformation and smart resizing. Understanding the logic behind these steps might take me some time.

PR #20489, opened 2024-11-13
Refactor `dot_product_attention` to use flash attention when available

james77777778:
I was experimenting with the new flash attention version of `dot_product_attention` and found it to be performant using jax backend.

For example, using SD3 medium to generate a 800x800 image on a RTX4070:
|flash attention|cost time|
|-|-|
|`False`|10.61s|
|`True`|5.45s|

However, the current implementation of `dot_product_attention` relies on users to decide whether to use flash attention, which can heavily depend on the backend.

This PR refactors the implementation by setting `flash_attention=None` to allow the backend to automatically detect availability. This change should be more user-friendly and improve maintainability IMO.

cc @divyashreepathihalli @hazemessamm

### DIFF
--- a/keras/src/backend/jax/nn.py
+++ b/keras/src/backend/jax/nn.py
@@ -972,7 +972,7 @@ def psnr(x1, x2, max_val):
     return psnr
 
 
-def _can_use_flash_attention(query, key, value, bias):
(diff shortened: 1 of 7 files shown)


divyashreepathihalli (approved):
LGTM!

fchollet (approved):
This is the right call! Thank you.

PR #20492, opened 2024-11-13
Fixing example code for BinaryFocalCrossentropy in losses.py file

tilakrayal:
The example mentioned for the Keras API binarycrossentropy was failing with the error AttributeError: 'list' object has no attribute 'shape'.

```python
>>> # Example 2: (batch_size = 2, number of samples = 4)
>>> y_true = [[0, 1], [0, 0]]
>>> y_pred = [[-18.6, 0.51], [2.94, -12.8]]
>>> # Using default 'auto'/'sum_over_batch_size' reduction type.
>>> loss = keras.losses.BinaryFocalCrossentropy(
        gamma=3, from_logits=True)
>>> loss(y_true, y_pred)
```

It works if we change the input from list to arrays.

```python
>>> # Example 2: (batch_size = 2, number of samples = 4)
>>> y_true = np.array([[0, 1], [0, 0]])
>>> y_pred = np.array([[-18.6, 0.51], [2.94, -12.8]])
>>> # Using default 'auto'/'sum_over_batch_size' reduction type.
>>> loss = keras.losses.BinaryFocalCrossentropy(
        gamma=3, from_logits=True)
>>> loss(y_true, y_pred)
```

### DIFF
--- a/keras/src/losses/losses.py
+++ b/keras/src/losses/losses.py
@@ -769,8 +769,8 @@ class BinaryFocalCrossentropy(LossFunctionWrapper):
     As a standalone function:
 
     >>> # Example 1: (batch_size = 1, number of samples = 4)
-    >>> y_true = [0, 1, 0, 0]
-    >>> y_pred = [-18.6, 0.51, 2.94, -12.8]
+    >>> y_true = np.array([0, 1, 0, 0])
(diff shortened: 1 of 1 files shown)


fchollet (approved):
Thanks for the fix!

PR #20494, opened 2024-11-14
Add soft_shrink activation

shashaka:
Since Keras lacks built-in support for the soft_shrink activation function, I implemented it across TensorFlow, JAX, and NumPy, aiming to maintain consistency with existing implementations available in PyTorch.


fchollet (approved):
LGTM, thanks for the addition.

PR #20502, opened 2024-11-15
Minor changes to asserts in keras.src.layers.attention.MultiHeadAttention

lcs-crr:
Removed another assert leading to same issue that #20340 fixed.
Made _output_shape more flexible; now it can be either a scalar integer or a tuple (as previously required).
Both are discussed in #19769


lcs-crr:
Unsure what to do here, since the errors seem to stem from

PR #20503, opened 2024-11-15
More flexible output_shape computation in keras.layers.MultiHeadAttention

lcs-crr:
Made the compute_output_shape method more flexible; now _output_shape can be either an integer or a tuple (as previously required).

Fix discussed at the bottom of issue #19769


rubencorr (commented):
.

fchollet (commented):
Thanks for the PR -- please add a unit test so we avoid breaking this use case in the future.

lcs-crr:
Unsure why code format test fails... In it, it says _All checks passed!_

fchollet (approved):

PR #20510, opened 2024-11-18
Fix tensorflow `_dot_product_attention_xla` and update `enable_flash_attention`

james77777778:
This PR modifies `enable_flash_attention` to set the value to `None`. This makes it more user-friendly by allowing the backend to decide whether to use flash attention.

The MHA tests have been also updated.

Additionally, a minor but crucial bug in `dot_product_attention` in tensorflow backend has been fixed. (identified by the updated test)


fchollet (approved):
LGTM, thanks

PR #20511, opened 2024-11-18
`MultiHeadAttention._compute_attention_mask()` always returns a bool …

hertschuh:
…tensor.

Previously, if an non-bool `attention_mask` was passed and no other mask was passed, the original `attention_mask` was returned unchanged.

Now, it is always cast to bool.


divyashreepathihalli (approved):
Thanks Fabien!

PR #20514, opened 2024-11-19
Introduce flash attention to GQA and move the flash attention configs to `backend.config`

james77777778:
This PR introduces flash attention to GQA and moves the flash attention configs to `backend.config`.
The previous location of the flash attention configs was a bit weird IMO.


fchollet (approved):
LGTM, thank you

PR #20528, opened 2024-11-21
Replace `np.iinfo` and `np.finfo` with `ml_dtypes`

james77777778:
`np.finfo` fails when encountering bfloat16. With `ml_dtypes`, we can improve compatibility.


fchollet (approved):
Sounds good -- thank you for the PR.

PR #20530, opened 2024-11-21
Raise an error when `save_weights` or `load_weights` is called on an unbuilt model

james77777778:
Related to #20524

We should raise an error when `save_weights` or `load_weights` is called on an unbuilt model as the behavior is undefined


fchollet (approved):
LGTM, thanks for the fix!

Issue #20521, opened 2024-11-19
EarlyStopping instance is not reusable


maciejbiesek:
From my observations it seems that we cannot use the same instance of EarlyStopping callback when training multiple models. My assumption is that it saves the best value from one training and uses it to compare during the training of the next model.

In `on_train_begin` we have comment "# Allow instances to be re-used", however the value `self.best` is not cleared. I think it should be reinitialized with ` self.best = (float("inf") if self.monitor_op == ops.less else -float("inf"))` as in `__init__()`.

hertschuh:
@maciejbiesek ,

Thank you for the report. The fix is in and should be in the `keras-nightly` pip package tomorrow.

PR #20533, opened 2024-11-22
Allow EarlyStopping to be reused between multiple `fit`s.

hertschuh:
All values were already reset properly in `on_train_begin` except `best`.

Fixes https://github.com/keras-team/keras/issues/20521


fchollet (approved):
LGTM, thanks for the fix.

PR #20546, opened 2024-11-25
WIP: nnx support demo

cgarciae:


DavidLandup0:
@cgarciae Keras running on top of nnx sometime in the future? 👀

cgarciae:
@DavidLandup0 maybe :)

keras/src/backend/jax/core.py
```diff
@@ -33,6 +34,14 @@ def _initialize(self, value):
             self._layout = None
         self._direct_assign(value)
 
+    @property
+    def _value(self):
```
hertschuh, line 38: I don't think this can work (at least without some refactoring). `self._value` is an attribute used in the super class and other places I think.

keras/src/backend/jax/layer.py
```diff
@@ -1,2 +1,5 @@
-class JaxLayer:
-    pass
+from flax import nnx
+
+class JaxLayer(nnx.Object):
+    def __init_subclass__(cls):
+        super().__init_subclass__()
```
hertschuh, line 5: Do you need this? It doesn't do that by default?

keras/src/layers/layer.py
```diff
@@ -538,7 +541,9 @@ def add_weight(
                 initializer = "zeros"
         initializer = initializers.get(initializer)
         with backend.name_scope(self.name, caller=self):
+            value = initializer(shape, dtype)
```
hertschuh, line 544: Oh, you can't change this. With the Tensorflow backend, in some cases, we need to delay the call to initializers.

Why is this needed?

keras/src/ops/operation.py
```diff
@@ -10,10 +10,13 @@
 from keras.src.utils import python_utils
 from keras.src.utils import traceback_utils
 from keras.src.utils.naming import auto_name
+from flax import nnx
 
 
 @keras_export("keras.Operation")
 class Operation:
+    def __init_subclass__(cls):
+        super().__init_subclass__()
```
hertschuh, line 19: Why is this needed? There is no super class.

keras/src/ops/operation.py
```diff
@@ -97,6 +100,7 @@ def __new__(cls, *args, **kwargs):
         to manually implement `get_config()`.
         """
         instance = super(Operation, cls).__new__(cls)
+        vars(instance)['_object__state'] = nnx.object.ObjectState()
```
hertschuh, line 103: Oh, this cannot go here. This class is cross-backend and `Operations` are required to be stateless. This is probably what is making all tests fail.

Move to `JaxLayer`.

hertschuh (changes requested):
@cgarciae 

Thanks for the PR!

It says demo. Besides making all the tests pass, what else is needed?

hertschuh:
NNX support was submitted here https://github.com/keras-team/keras/pull/21252

PR #20547, opened 2024-11-26
Add diagflat op

IMvision12:
@fchollet


fchollet (approved):
LGTM -- thanks for the addition. It's useful for Keras to increase its coverage of the NumPy API. Question for you: how are you prioritizing which ops to add?

IMvision12:
> LGTM -- thanks for the addition. It's useful for Keras to increase its coverage of the NumPy API. Question for you: how are you prioritizing which ops to add?

I’ve compiled a list of all operations that are not yet in Keras and am gradually adding them one by one. To create this list, I systematically compared the operations available across different frameworks.

PR #20557, opened 2024-11-27
Better handling of variable creation with tensor as initializer.

hertschuh:
- when the variable dtype is not specified, the dtype of the tensor/array passed as initializer is used instead of defaulting to `backend.floatx()`.
- with the JAX backend, don't needlessly create a copy of the initializer value, reuse it if possible

### DIFF
--- a/keras/src/backend/common/variables.py
+++ b/keras/src/backend/common/variables.py
@@ -115,7 +115,6 @@ class Variable:
             self._path = current_path() + "/" + name
         else:
             self._path = name
-        self._dtype = standardize_dtype(dtype)
         self._shape = None
         self._initializer = None
         self._regularizer = None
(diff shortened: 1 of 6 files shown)


fchollet (commented):
LGTM, thanks

I think both behaviors are acceptable to be honest -- it's a bit weird to create a Variable from a np array and have it be float64 by default.

Perhaps we can compromise by first converting the value to a tensor (which might cast the dtype) and then using the resulting dtype?

hertschuh:
> I think both behaviors are acceptable to be honest -- it's a bit weird to create a Variable from a np array and have it be float64 by default.

What was also weird is that you would often get 3 copies in memory: the original array, the cast array and then the copied array.

> Perhaps we can compromise by first converting the value to a tensor (which might cast the dtype) and then using the resulting dtype?

Oh yeah, that's a good idea. This will take care of the `int` vs `float` Python array case.

hertschuh:
> > Perhaps we can compromise by first converting the value to a tensor (which might cast the dtype) and then using the resulting dtype?
> 
> Oh yeah, that's a good idea. This will take care of the `int` vs `float` Python array case.

@fchollet this is now ready.

fchollet (approved):

Issue #20574, opened 2024-12-01
MeanIoU differ from custom IOU metrics implementation


edge7:
Hi,
am running a segmentation training process and am using the following function as IoU Custom metrics:

```
@keras.saving.register_keras_serializable(package="glass_segm", name="custom_iou_metric")
def custom_iou_metric(y_true, y_pred, num_classes=3):
    y_pred = tf.argmax(y_pred, axis=-1)

    y_true = tf.cast(tf.reshape(y_true, [-1]), tf.int32)
    y_pred = tf.cast(tf.reshape(y_pred, [-1]), tf.int32)

    iou = tf.constant(0.0)

    for i in range(num_classes):
        true_mask = tf.cast(tf.equal(y_true, i), tf.float32)
        pred_mask = tf.cast(tf.equal(y_pred, i), tf.float32)

        intersection = tf.reduce_sum(true_mask * pred_mask)
        union = tf.reduce_sum(true_mask) + tf.reduce_sum(pred_mask) - intersection

        class_iou = tf.cond(
            tf.equal(union, 0), lambda: tf.constant(1.0), lambda: intersection / union
        )
        iou += class_iou

    iou /= tf.cast(num_classes, tf.float32)

    return iou
```

I was expecting that your MeanIoU would be the IoU mean across the classes and **also** the mean all the training and validation set batches, but it does not seem like that. 
example:

given this:

```
y_pred = np.array([[[0.8, 0.1, 0.1], [0.7, 0.2, 0.1], [0.1, 0.7, 0.2], [0.2, 0.6, 0.2]],
 [[0.6, 0.3, 0.1], [0.7, 0.2, 0.1], [0.2, 0.5, 0.3], [0.3, 0.4, 0.3]],
 [[0.7, 0.2, 0.1], [0.3, 0.6, 0.1], [0.4, 0.4, 0.2], [0.3, 0.3, 0.4]],
 [[0.5, 0.4, 0.1], [0.3, 0.5, 0.2], [0.3, 0.3, 0.4], [0.4, 0.5, 0.1]]]

)
y_true = np.array([[0, 0, 2, 1],
 [1, 0, 1, 2],
 [1, 2, 2, 0],
 [1, 2, 0, 0]]
)
```

If I run:
```
m = keras.metrics.MeanIoU(num_classes=3,sparse_y_pred=False)
for i in range(1):
    y_true[0][0] = i % 2
    y_true[1][0] = i % 2
    m.update_state(y_true, y_pred)
[...]

PR #20584, opened 2024-12-03
Fix confusion matrix type

edge7:
Hello,

this is an attempt to show and fix the issue I opened:
#20574

### DIFF
--- a/keras/src/metrics/iou_metrics_test.py
+++ b/keras/src/metrics/iou_metrics_test.py
@@ -26,9 +26,7 @@ class IoUTest(testing.TestCase):
         y_pred = [0, 1, 0, 1]
         y_true = [0, 0, 1, 1]
 
-        obj = metrics.IoU(
-            num_classes=2, target_class_ids=[0, 1], dtype="float32"
-        )
+        obj = metrics.IoU(num_classes=2, target_class_ids=[0, 1])
 
         result = obj(y_true, y_pred)
 
@@ -89,9 +87,7 @@ class IoUTest(testing.TestCase):
         y_pred = np.array([1], dtype=np.float32)
         y_true = np.array([1])
 
-        obj = metrics.IoU(
-            num_classes=2, target_class_ids=[0, 1], dtype="float32"
-        )
+        obj = metrics.IoU(num_classes=2, target_class_ids=[0, 1])
         result = obj(y_true, y_pred)
 
         # cm = [[0, 0],
@@ -173,9 +169,7 @@ class BinaryIoUTest(testing.TestCase):
         # sum_row = [2, 2], sum_col = [2, 2], true_positives = [1, 1]
         # iou = true_positives / (sum_row + sum_col - true_positives))
         expected_result = (1 / (2 + 2 - 1) + 1 / (2 + 2 - 1)) / 2
-        obj = metrics.BinaryIoU(
-            target_class_ids=[0, 1], threshold=0.3, dtype="float32"
-        )
+        obj = metrics.BinaryIoU(target_class_ids=[0, 1], threshold=0.3)
         result = obj(y_true, y_pred)
         self.assertAllClose(result, expected_result, atol=1e-3)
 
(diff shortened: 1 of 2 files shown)


fchollet:
Thanks for the PR. Unfortunately using `float64` will be incompatible with e.g. the JAX backend.

edge7:
ok, in that case, it might suffice to check the backend and set float64 just in case of torch/tf. However, the results might be wrong with JAX float 32. 
Alternatively, the entire thing (the confusion matrix) can be done with int64, without issues

fchollet:
What about int32? Is that enough?

edge7:
int32 is good too.
But it breaks a good amount of existing tests.

edge7:
By the way, I suggest caution with *32 as with high resolution images and big dataset the numbers can explode very quickly

fchollet:
> The weights should probably also be just int32, which makes sense IMHO

Indeed, can we just do that then?

edge7:
Ok. I'll try to modify the PR (will be much larger as some existing tests rely on float weights)

edge7:
I switched to int32 as the default, which resolved the main issue. I also made a small adjustment to the result function to address a complaint related to the epsilon (which is float). Additionally, I fixed a couple of existing tests and streamlined my own test to make it lighter and avoid OOM problems.

edge7:
Hold on with this one as I am trying to run a training with the changes and I get:

`Trying to access resource mean_io_u_1/total_confusion_matrix/600 (defined @ /anaconda/envs/keras3_jax/lib/python3.10/site-packages/tensorflow/python/ops/gen_resource_variable_ops.py:1267) located in device /job:localhost/replica:0/task:0/device:CPU:0 from device /job:localhost/replica:0/task:0/device:GPU:0`

edge7:
with int64, training starts and the metric looks ok (Tf 2.18 for training).
Also with int64, tests work fine in JAX and Torch.

edge7:
Training with JAX also works with int64, but warnings are raised:

```
/anaconda/envs/keras3_jax/lib/python3.10/site-packages/keras/src/backend/jax/core.py:77: UserWarning: Explicitly requested dtype int64 requested in asarray is not available, and will be truncated to dtype int32. To enable more dtypes, set the jax_enable_x64 configuration option or the JAX_ENABLE_X64 shell environment variable. See https://github.com/google/jax#current-gotchas for more.
  return jnp.asarray(x, dtype=dtype)
```
The metric looks fine anyway, as the truncation does not affect the result. 

To sum up:

- Float 32 is not good enough
- Int 32 is good for representing big integers, but cannot be placed in GPU with TF
- Int 64 works for TF GPU, but raise warning in JAX (but still working)
- Not tested Torch training

james77777778:
Hey @edge7 
It is well-known that we need to use int64 for tensorflow to work on GPUs. We can use a "magic" `int` to let the backend decide whether to use int32 (for jax, torch) or int64 (for tensorflow).
Please refer to https://github.com/keras-team/keras/blob/master/keras/src/backend/common/dtypes.py#L49

You should be able to implement it as follows:
```python
super().__init__(name=name, dtype=dtype or int)
```

keras/src/metrics/iou_metrics_test.py
```diff
@@ -64,7 +65,9 @@ def test_multi_dim_input(self):
         y_true = np.array([[0, 0], [1, 1]])
         sample_weight = np.array([[0.2, 0.3], [0.4, 0.1]])
 
-        obj = metrics.IoU(num_classes=2, target_class_ids=[0, 1])
+        obj = metrics.IoU(
+            num_classes=2, target_class_ids=[0, 1], dtype="float32"
```
fchollet, line 69: Why was it necessary to change the dtypes in the unit tests?

edge7: float32 is essential for tests involving float-format weights and Binary IoU calculations, as casting problems can result in all predictions being set to zero. This is somewhat inconvenient, as float would typically be the preferred choice, but it becomes unreliable for values exceeding 2^24

edge7: I'm unsure about the best way to proceed. The issue is clear, but the optimal approach isn't straightforward. Using float64 would be a quick fix, but JAX doesn't handle it well (even though it can be configured to use it, I’m not sure if there are any side effects). Using int32/int64 is a decent alternative, but given the current implementation, passing weights as floats wouldn't make sense in that case. To address this, a warning could be issued, or an exception could be raised, with clear documentation explaining the limitation. Additionally, the IoU binary calculation should be slightly adjusted to evaluate >= threshold in float format regardless of the data type, and only cast to the appropriate type afterward.

fchollet (approved):

PR #20586, opened 2024-12-04
[TextVectorization Layer] Added tests for testing the funtionality

Frightera:
Trying to address TODO:
```python
# TODO: increase coverage. Most features aren't being tested.
```

I think this PR covers most basic functionalities.


fchollet (approved):
LGTM -- thank you for the contribution! Looks like these tests adds coverage for 96 lines.

PR #20594, opened 2024-12-05
Fix masking propagation when `_keras_mask` is unset during `call`

james77777778:
#20562 causes issues in some downstream projects related to mask propagation.

The root cause is that `previous_mask` is retrieved after `__call__` but the mask might be removed during `__call__` to suppress warnings (ex: `MultiHeadAttention`). To fix it, we can cache the `previous_mask` before `__call__` and set it to the outputs.

cc @SamanehSaadat


SamanehSaadat (approved):
Thanks for the fix!

fchollet (approved):

PR #20610, opened 2024-12-09
Unscale loss value in metrics in tensorflow

james77777778:
Follow-up to #20609

We need to unscale the loss value for metrics when using distributed strategies.
The test has been updated.


fchollet (approved):
LGTM, thank you!

PR #20615, opened 2024-12-10
Allow lower precision in `einsum` in `dot_product_attention`

james77777778:
This PR updates the dtype used in `einsum` in `dot_product_attention` for tensorflow and numpy.

Initially, I followed jax's implementation which casts `key` and `query` to higher precision early to mimic the behavior of `preferred_element_type="float32"`. However, performing `einsum` in float32 is much slower than float16 or bfloat16. Additionally, the original MHA implementation doesn't cast them to float32.
Therefore, it seems reasonable to use the dtype of `key` and `query` directly for `einsum`.

Related to https://github.com/keras-team/keras-hub/pull/2014

### DIFF
--- a/keras/src/backend/numpy/nn.py
+++ b/keras/src/backend/numpy/nn.py
@@ -1096,12 +1096,14 @@ def _apply_masks(logits, mask, is_causal):
 
 
 def _dot_product_attention_xla(query, key, value, bias, mask, is_causal, scale):
+    original_dtype = key.dtype
     logits_dtype = np.promote_types(query.dtype, np.float32)
-    logits = np.einsum(
-        "BTNH,BSNH->BNTS",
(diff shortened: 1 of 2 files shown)


mattdangerw (approved):
lgtm! thanks!

mattdangerw:
@james77777778 looked like there was a legit test failure, but I see you just pushed, so guessing you got it. Will merge once this is all green!

james77777778:
> @james77777778 looked like there was a legit test failure, but I see you just pushed, so guessing you got it. Will merge once this is all green!

Yeah, it was just a dtype mismatched in numpy. Should be fixed now.

james77777778:
The failed GPU CI tests should be unrelated to this PR

fchollet (approved):

PR #20626, opened 2024-12-10
Fix Layer normalization issue with scalar mean & variance

### LINKED ISSUE #20624: Normalization layer fails to broadcast scalar mean and variance
Normalization layer errors out for scalar mean and variance arguments. The documentation states that these arguments will be broadcast to the necessary size and the examples include scalars, but instead the source code seems to only reshape them (see [here](https://github.com/keras-team/keras/blob/v3.7.0/keras/src/layers/preprocessing/normalization.py#L193)).

I'm using Python 3.12.1 and Keras 3.6.0.

```
>>> in1 = keras.Input(shape=(16, 16, 3))
>>> normLayer = keras.layers.Normalization(mean=0.1, variance=0.05)
>>> out1 = normLayer(in1)
I tensorflow/core/framework/local_rendezvous.cc:405] Local rendezvous is aborting with status: INVALID_ARGUMENT: Input to reshape is a tensor with 1 values, but the requested shape has 3
```

Surya2k1:
Fix Layer normalization issue with scalar mean & variance.

With scalar mean instead of broadcasting reshaping applied currently. Changed the same.

Fixes #20624.

### DIFF
--- a/keras/src/layers/preprocessing/normalization.py
+++ b/keras/src/layers/preprocessing/normalization.py
@@ -190,8 +190,8 @@ class Normalization(TFDataLayer):
             # with proper broadcast shape for use during call.
             mean = ops.convert_to_tensor(self.input_mean)
             variance = ops.convert_to_tensor(self.input_variance)
-            mean = ops.reshape(mean, self._broadcast_shape)
-            variance = ops.reshape(variance, self._broadcast_shape)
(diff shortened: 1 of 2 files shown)


fchollet (commented):
Thanks for the PR! Please add a unit test, so we avoid breaking this case in the future.

PR #20638, opened 2024-12-12
test: add test for class weights (py_dataset adapter)

edge7:


edge7:
Simple PR to add tests for: 

> # TODO: test class_weight
in` py_dataset_adapter.PyDataset`

Needed to add 2 calls to _standardize_batch:
- _finite_generator
- _infinite_generator

as I've noticed that `_standardize_batch` handles class_weights, but it was just called from `get_tf_dataset`. Hope there are no side effect, but tests are fine.

keras/src/trainers/data_adapters/py_dataset_adapter_test.py
```diff
@@ -217,10 +217,35 @@ def test_basic_flow(
...
+            for sub_elem in range(2):
+                self.assertTrue(
+                    np.array_equal(batch[0][sub_elem], x[index * 2 + sub_elem])
+                )
+                self.assertEqual(batch[1][sub_elem], y[index * 2 + sub_elem])
+                self.assertEqual(
+                    batch[2][sub_elem], class_w[batch[1][sub_elem]]
+                )
+
+        assert index == 1  # 2 batches
```
fchollet, line 247: Please use `self.assertEqual`

keras/src/trainers/data_adapters/py_dataset_adapter_test.py
```diff
@@ -217,10 +217,35 @@ def test_basic_flow(
...
+        class_w = {0: 2, 1: 1, 2: 3}
+        py_dataset = ExamplePyDataset(x, y, batch_size=2)
+        adapter = py_dataset_adapter.PyDatasetAdapter(
+            py_dataset, shuffle=False, class_weight=class_w
+        )
+        gen = adapter.get_numpy_iterator()
+
+        for index, batch in enumerate(gen):
+            # Batch is a tuple of (x, y, class_weight)
+            assert len(batch) == 3
```
fchollet, line 235: `self.assertLen`

fchollet (commented):
Thanks for the PR!

keras/src/trainers/data_adapters/data_adapter_utils_test.py
```diff
@@ -0,0 +1,52 @@
...
+            (
+                "all_unmapped",
+                np.array([0, 1, 0, 2]),
+                {},
+                np.array([1.0, 1.0, 1.0, 1.0]),
+            ),
+        ]
+    )
+    def test_class_weight_to_sample_weights(self, y, class_weight, expected):
+        assert np.allclose(
```
fchollet, line 50: Please use `self.assertAllClose`

edge7:
The PR turned out to be bigger than expected, as I initially thought it only required adding a test. However, I gradually noticed that class_weight was not actually being used in the case of PyDataset.
One note: astype is not supported with PyTorch tensors, so I had to adjust that case.

keras/src/trainers/data_adapters/data_adapter_utils.py
```diff
@@ -121,7 +121,15 @@ def class_weight_to_sample_weights(y, class_weight):
             y = np.argmax(y, axis=-1)
         else:
             y = np.squeeze(y, axis=-1)
-    y = np.round(y).astype("int32")
+    y = np.round(y)
+    if hasattr(y, "astype"):
```
fchollet, line 125: `y = np.round(y)` should have converted the tensor to a numpy tensor, no?

edge7: I am not at the laptop right now to paste something to reproduce it, but it was failing for me. np.round was fine (but still returning a torch tensor)

fchollet: I really don't get this -- if you apply any `np` op to an input that isn't numpy (like a torch tensor) it will call `__array__` on the object to turn it into a NumPy ndarray. You should never have a torch tensor past that point

edge7: Given this example:
```
import torch
import numpy as np
a = torch.from_numpy(np.array([1, 2.5, 3]))
b = np.round(a)
```
placing a breakpoint at the last line:

I get here:

image
which is [here](https://github.com/numpy/numpy/blob/3ca9e751af24426e5e70ec13a8ca6f4ea13f0758/numpy/_core/fromnumeric.py#L52)
it tries to apply the bound but:

image

so we enter `_wrapit`
which is [here](https://github.com/numpy/numpy/blob/3ca9e751af24426e5e70ec13a8ca6f4ea13f0758/numpy/_core/fromnumeric.py#L41)

Please note that following the debugging flow into:

`conv = _array_converter(obj)
`
bring me [here](https://github.com/pytorch/pytorch/blob/9933e59c2b16c1b0475189eef63b0ff405dfd091/torch/_tensor.py#L1212), i.e.: we are in the torch domain now.
image

then when back to `_wrapit`
the result then is indeed a numpy array:

image

the last call brings me to the [following](https://github.com/pytorch/pytorch/blob/9933e59c2b16c1b0475189eef63b0ff405dfd091/torch/_tensor.py#L1222):

image
so I get back a Torch tensor again.

edge7:
```
pip list | grep torch
torch                      2.5.1+cpu
torchvision                0.20.1+cpu

[notice] A new release of pip is available: 24.0 -> 24.3.1
[notice] To update, run: pip install --upgrade pip
❯ pip list | grep numpy
numpy                      2.0.2

```

edge7: Hi, am trying to get a better understanding and it looks like that __array_wrap__ is indeed a way to get back the original object after the numpy conversion:

https://numpy.org/doc/2.1/user/basics.interoperability.html#returning-foreign-objects

Which seems to be exactly what torch has implemented

fchollet: Ok, this is so messed up. Then we should use `ops.convert_to_numpy` at the start of the series of calls.

edge7: Do you mean at the start of this function or even before?

edge7: committed  'convert_to_numpy'

fchollet (approved):
LGTM, thank you!

edge7:
Thanks for merging @fchollet
Do you remember what you meant here https://github.com/keras-team/keras/blob/c0a1f2490c9ebd8e182d1e373be73d287e89e2e3/keras/src/trainers/data_adapters/py_dataset_adapter_test.py#L221

?

Just passing batch equal to 1 or something else?

PR #20655, opened 2024-12-17
Fix paths for pytest in contribution guide

mmicu:
It looks like that the internal structure of `keras/` changed to `keras/src` at a certain point. Hence, the paths are not valid anymore and `pytest` cannot find the proper files. For example:

```sh
$ pytest keras/losses/losses_test.py
ERROR: file or directory not found: keras/losses/losses_test.py
```

After fixing all these paths, the tests run successfully locally.


fchollet (approved):
LGTM, thanks for the fix!

Issue #20276, opened 2024-09-21
GRU + Large `recurrent_dropout` Bug


nokados:
**Keras version:** 3.5.0  
**Backend:** TensorFlow 2.17.0

I encountered a strange bug when working with the GRU layer. If you create a simple model with a GRU layer and set `recurrent_dropout=0.5`, very strange behavior occurs:

0. **With sequence length 20**: Everything works as expected.
1. **With sequence length 100**: The output of the GRU layer during training, with the default `tanh` activation, produces very large values in the range of ±1e25, even though it should be constrained to [-1, 1]. This results in an extremely large loss.
2. **With sequence length 145**: The behavior is unstable. I received the following warning:

```
   Object was never used (type <class 'tensorflow.python.ops.tensor_array_ops.TensorArray'>):
<tensorflow.python.ops.tensor_array_ops.TensorArray object at 0x7f6311231eb0>
If you want to mark it as used call its "mark_used()" method.
It was originally created here:
  File "/home/nokados/.pyenv/versions/qaclassifiers39-pdm/lib/python3.9/site-packages/keras/src/backend/tensorflow/rnn.py", line 419, in <genexpr>
    ta.write(ta_index_to_write, out)  File "/home/nokados/.pyenv/versions/qaclassifiers39-pdm/lib/python3.9/site-packages/tensorflow/python/util/tf_should_use.py", line 288, in wrapped
    return _add_should_use_warning(fn(*args, **kwargs),
```

   I was unable to reproduce this behavior in Colab; there, either the loss becomes `inf`, or it behaves similarly to the longer sequence lengths.
3. **With sequence length 200**: It throws an error:

```
Epoch 1/50

2024-09-21 22:10:35.493005: I tensorflow[/core/framework/local_rendezvous.cc:404](http://localhost:8888/core/framework/local_rendezvous.cc#line=403)] Local rendezvous is aborting with status: INVALID_ARGUMENT: indices[0] = 2648522 is not in [0, 25601)

---------------------------------------------------------------------------
InvalidArgumentError                      Traceback (most recent call last)
Cell In[15], line 1
----> 1 model.fit(
      2     dataset, onehot_target,
      3     batch_size=128,
      4     epochs=50,     
     5 )

[...]

PR #20656, opened 2024-12-18
Fix recurrent dropout for GRU.

hertschuh:
The simplified implementation, which used the same recurrent dropout masks for all the previous states didn't work and caused the training to not converge with large enough recurrent dropout values.

This new implementation is now the same as Keras 2. Note that recurrent dropout requires "implementation 1" to be turned on.

Fixes https://github.com/keras-team/keras/issues/20276


fchollet (approved):
LGTM, thanks for the fix!

PR #20657, opened 2024-12-18
Make sklearn dependency optional

mattdangerw:
Conditionally import sklearn. Only error on usage of keras sklearn symbol usage.

Kinda awkward that we need to declare empty classes for the classes we are missing from sklearn (if it isn't installed), but this seems like the simplest solution.


fchollet (approved):
LGTM -- thank you!

PR #20663, opened 2024-12-18
Change recurrent dropout implementation for LSTM.

hertschuh:
This change is to make the implementation of recurrent dropout consistent with GRU (changed as of https://github.com/keras-team/keras/pull/20656 ) and Keras 2.

Also fixed a bug where the GRU fix would break when using CUDNN with a dropout and no recurrent dropout. The solution is to create multiple masks only when needed (implementation == 1).

Added coverage for the case when dropout is set and recurrent dropout is not set.


fchollet (approved):
LGTM, thanks

hertschuh:
@nokados

PR #20665, opened 2024-12-19
Fix `PyDatasetAdapterTest::test_class_weight` test with Torch on GPU.

hertschuh:
The test was failing because arrays on device and on cpu were compared.


mattdangerw (approved):
Thanks!

PR #20666, opened 2024-12-19
Fix up torch GPU failing test for mix up

mattdangerw:
We need to make sure to use get any tensors places on cpu before using them in the tensorflow backend during preprocessing.


hertschuh (approved):

fchollet (approved):
Thanks for the fix!

PR #20678, opened 2024-12-23
Fix issue with nested model output as input

Surya2k1:
Currently with nested model output as input to functional model fails as tensor_dict not updated. The below code fails.

```
model_1 = Sequential([
    Input(shape=(6,)),
    layers.Dense(3, activation="sigmoid"),
])

model_2 = Sequential([
    Input(shape=(3,)),
], )

combined = Model(Input(shape=(6,)), model_2(model_1(Input(shape=(6,)))), name='nested_model')
combined.compile(loss='binary_crossentropy', optimizer='adam')
output = combined.train_on_batch(np.random.normal(0, 1, (8, 6)), np.random.normal(0, 1, (8, 1)))
```

Done the changes to make it work not seems elegant though. But for this case it works and no functional model tests are failing with this change.

Context from #20473

### DIFF
--- a/keras/src/models/functional_test.py
+++ b/keras/src/models/functional_test.py
@@ -714,3 +714,19 @@ class FunctionalTest(testing.TestCase):
             "The structure of `inputs` doesn't match the expected structure",
         ):
             model([x1, x2])
+
+    def test_nested_model_call_as_arg(self):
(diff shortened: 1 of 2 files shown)


Surya2k1:
Closing as this is not correct fix.

PR #20682, opened 2024-12-23
Change the import order of pydot and its older forks

### LINKED ISSUE #20681: Consider changing the import order of `pydot` and its forks
In `keras/src/utils/model_visualization.py`:

```python
try:
    # pydot-ng is a fork of pydot that is better maintained.
    import pydot_ng as pydot
except ImportError:
    # pydotplus is an improved version of pydot
    try:
        import pydotplus as pydot
    except ImportError:
        # Fall back on pydot if necessary.
        try:
            import pydot
        except ImportError:
            pydot = None
```

`pydotplus` has not been updated in 10 years, and `pydot-ng` has not been updated in 6 years.
Consider using `pydot` as the first option and the other options as fallbacks instead of the other way around.
`pydot` has been resurrected some time ago and it's the main project.

lkk7:
Fixes #20681 

Feel free to add your suggestions!

### DIFF
--- a/keras/src/utils/model_visualization.py
+++ b/keras/src/utils/model_visualization.py
@@ -8,16 +8,15 @@ from keras.src.api_export import keras_export
 from keras.src.utils import io_utils
 
 try:
-    # pydot-ng is a fork of pydot that is better maintained.
-    import pydot_ng as pydot
+    import pydot
 except ImportError:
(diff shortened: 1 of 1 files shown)


fchollet (approved):
The pydot maintenance timeline is such a rollercoaster. Thanks for the PR!

PR #20694, opened 2024-12-27
minor: Fix Functional API guide

punkeel:
Add an empty line so the list is rendered as a list, not as a single line of text

Currently:
image


fchollet (approved):
LGTM, thank you

PR #20711, opened 2025-01-02
Fix typos in additive_attention

keerthanakadiri:
I observed a few typos in additive_attention

### DIFF
--- a/keras/src/layers/attention/additive_attention.py
+++ b/keras/src/layers/attention/additive_attention.py
@@ -33,8 +33,8 @@ class AdditiveAttention(Attention):
             - `query`: Query tensor of shape `(batch_size, Tq, dim)`.
             - `value`: Value tensor of shape `(batch_size, Tv, dim)`.
             - `key`: Optional key tensor of shape `(batch_size, Tv, dim)`. If
-                not given, will use the`value` for both `key` and `value`, which is
-                the most common case.
+                not given, will use the`value` for both `key` and `value`,
+                which is the most common case.
         mask: List of the following tensors:
             - `query_mask`: A boolean mask tensor of shape `(batch_size, Tq)`.
                 If given, the output will be zero at the positions where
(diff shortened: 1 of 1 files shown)


gbaned (approved):

keras/src/layers/attention/additive_attention.py
```diff
@@ -33,7 +33,7 @@ class AdditiveAttention(Attention):
             - `query`: Query tensor of shape `(batch_size, Tq, dim)`.
             - `value`: Value tensor of shape `(batch_size, Tv, dim)`.
             - `key`: Optional key tensor of shape `(batch_size, Tv, dim)`. If
-                not given, will use `value` for both `key` and `value`, which is
+                not given, will use the`value` for both `key` and `value`, which is
```
nikolasavic3, line 36: The line you modified is now 83 characters long, which is why the code format check didn't pass. The maximum allowed length is 80.
You can rewrite it to:
```suggestion
                not given, will use the`value` for both `key` and `value`, 
                which is the most common case.
```

image

nikolasavic3 (commented):
looks good!

gbaned (approved):

NANTEGEBABRAH:
There are some three typos i would like to fix.

keras/src/layers/attention/additive_attention.py
```diff
@@ -33,8 +33,8 @@ class AdditiveAttention(Attention):
             - `query`: Query tensor of shape `(batch_size, Tq, dim)`.
             - `value`: Value tensor of shape `(batch_size, Tv, dim)`.
             - `key`: Optional key tensor of shape `(batch_size, Tv, dim)`. If
-                not given, will use `value` for both `key` and `value`, which is
-                the most common case.
+                not given, will use the`value` for both `key` and `value`,
```
hertschuh, line 36: Please add space between "the" and "value".

hertschuh (commented):
Thanks for the PR! It needs one tweak.

PR #20718, opened 2025-01-03
Add a gradient checkpoint feature

pass-lin:
[Gradient checkpoint](https://papers.cool/arxiv/1604.06174) is a widely used technique to reduce memory consumption. 
Now we are adapting it for Keras. To make minimal modifications to existing models, we add a parameter `enable_gradient_checkpoint` to the layer, which is set to `False` by default. By simply changing this parameter, we can enable gradient checkpointing. However, for specific implementations depending on different backends, the following points need to be considered:
      In the Torch backend, you should ensure that there are no dropout layers or normalization layers (such as BN, LN, GN, etc.) with inconsistent forward and backward behaviors in the layer of the function you're starting.
      In the TensorFlow backend, you can only enable this setting in eager mode.
      In the JAX backend, you should ensure that there are no strings or other non-differentiable JAX vaild types in the inputs of your function.Such as str


pass-lin:
@gbaned 
```
All checks passed!
Would reformat: keras/src/layers/layer.py
Would reformat: keras/src/ops/operation.py
2 files would be reformatted, 868 files already formatted
Error: Process completed with exit code 1
```
What does this error mean in the code format check? How should I fix it?

PR #20725, opened 2025-01-05
Patch `tf2onnx` to ensure compatibility with `numpy>=2.0.0`

james77777778:
This PR adds hacks to `tf2onnx` to ensure Keras remains compatible with `numpy>=2.0.0`.

Additionally, the export logic has been simplified, removing the need for a temporary SavedModel.


fchollet (approved):
LGTM, thank you!

PR #20729, opened 2025-01-06
Add build method to supress warning

shashaka:
To suppress the warning mentioned below, I declared the build method.

`UserWarning: `build()` was called on layer 'random_color_jitter', however the layer does not have a `build()` method implemented and it looks like it has unbuilt state. This will cause the layer to be marked as built, despite not being actually built, which may cause failures down the line. Make sure to implement a proper `build()` method.
  warnings.warn(`


fchollet (approved):
Thanks for the fix!

PR #20736, opened 2025-01-08
Allow dynamic shape in `STFTSpectrogram` layer.

hertschuh:
by simply using `ops.shape(x)` instead of `x.shape`.


fchollet (approved):
LGTM, thanks for the fix

PR #20749, opened 2025-01-10
Fix JAX GPU CI and make formatter happy

james77777778:
We need to increase the tolerance for the export test in `TestJaxLayer`

EDITED:
The new version of `ruff` introduces some changes. This PR also makes formatter happy.


fchollet (approved):
Thanks for the fix!

PR #20755, opened 2025-01-13
Fix serialization of domain packages.

hertschuh:
Not all of their symbols are exported.


mattdangerw (commented):
Seems fine for now, though I think we can update the broken KerasHub tests to either always register a serializable or not test serialization.

mattdangerw (approved):

PR #20774, opened 2025-01-17
fix(ops): Fix inconsistent padding calculation in PyTorch backend ops

harshaljanjani:
Was able to still reproduce the error, the PyTorch backend had inconsistent behavior between static shape inference and dynamic execution for pooling operations, particularly with 'same' padding and non-unit strides, figured that the root cause was by incorrect padding calculation logic that didn't properly handle asymmetric padding cases.

Key changes:
- Rewrote `_compute_padding_length()` to handle stride-based padding
- Fixed padding calculation to properly support asymmetric padding cases
- Standardize `channels_first`/`channels_last` conversion in pooling ops
- Cleaned up padding application in `_apply_same_padding()`
- Added proper handling of `data_format` throughout pooling pipeline

This fixes the issue where MaxPooling2D with 'same' padding would produce different shapes between compute_output_shape() and actual execution (e.g. (1,5,2,2) vs (1,5,2,1)) #20235.

Rebased on top of @sachinprasadhs's September 2024 PR to incorporate latest `keras:master` changes #20270.

### DIFF
--- a/keras/src/backend/torch/nn.py
+++ b/keras/src/backend/torch/nn.py
@@ -2,7 +2,6 @@ import torch
 import torch.nn.functional as tnn
 
 from keras.src import backend
-from keras.src import tree
 from keras.src.backend.common.backend_utils import (
     compute_conv_transpose_padding_args_for_torch,
 )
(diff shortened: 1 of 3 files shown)


fchollet (approved):
LGTM, thanks for the fix.

PR #20777, opened 2025-01-17
mlx fft implementation

acsweet:
This just implements fft for the mlx backend, contributing to #19571


fchollet (approved):
Yay, more MLX work! No one is contributing to the MLX backend at the moment, but I think it's a promising project.

acsweet:
@fchollet I'll see how far I can get!

PR #20779, opened 2025-01-17
Fix memory leaks in `model.evaluate`.

hertschuh:
The history is only used in `model.fit`, no need to create it for `evaluate` and `predict`. The history is attached to the model and therefore lives for as long as the model is around.

The executor used in `CallbackList` was never shut down, causing it to keep a thread around, which in turn had thread locals that were leaked.


fchollet (approved):
LGTM, thanks for the fix

PR #20784, opened 2025-01-19
fix ciou implementation bug in bounding_boxes/iou.py

### LINKED ISSUE #20780: Incorrect implementation of "v" in "compute_ciou" function
**Describe the bug**

In the current implementation of `v` in `compute_ciou` function (keras/src/layers/preprocessing/image_preprocessing/bounding_boxes/iou.py), the  `(4 / math.pi**2)` term is included inside `power` and so gets squared together with the arctan term. 

> v = ops.squeeze(
        ops.power(
            (4 / math.pi**2)
            * (ops.arctan(width_2 / height_2) - ops.arctan(width_1 / height_1)),
            2,
        ),
        axis=-1,
    )

However, according to equation (8) in the original [CIoU paper](https://arxiv.org/pdf/2005.03572), `(4 / math.pi**2)` should NOT be squared (i.e., it should be outside the `power` op) - only the arctan term should be squared. This is correctly reflected in Ultralytics' implementation of CIoU [here](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/utils/metrics.py).

Surya2k1:
Fix `ciou` implementation bug in bounding_boxes/iou.py.

As per the CIOU paper there seems a bug in implementation of `v` in `compute_ciou` as reported by the user of #20780

As per original paper and other resource referred by the user changing the implementation.

Fixes #20780.

### DIFF
--- a/keras/src/layers/preprocessing/image_preprocessing/bounding_boxes/iou.py
+++ b/keras/src/layers/preprocessing/image_preprocessing/bounding_boxes/iou.py
@@ -267,9 +267,9 @@ def compute_ciou(boxes1, boxes2, bounding_box_format, image_shape=None):
     )
 
     v = ops.squeeze(
-        ops.power(
-            (4 / math.pi**2)
-            * (ops.arctan(width_2 / height_2) - ops.arctan(width_1 / height_1)),
+        (4 / math.pi**2)
+        * ops.power(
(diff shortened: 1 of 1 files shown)


fchollet (approved):
Yes, this is correct. Thank you for the fix.

PR #20792, opened 2025-01-22
`mlx.nn.conv` implementation

acsweet:
This relates to #19571 

The tests look good! All tests are passing for `keras/src/layers/rnn` now too

### DIFF
--- a/keras/src/backend/mlx/nn.py
+++ b/keras/src/backend/mlx/nn.py
@@ -1,11 +1,13 @@
 import mlx.core as mx
 import mlx.nn as nn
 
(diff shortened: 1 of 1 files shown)


fchollet (approved):
Excellent -- thank you!

Do you anticipate any major missing op in `nn.py`? (besides `ctc_loss`)

acsweet:
At a quick glance, I think mlx has everything we need!

A few ops in `nn.py` that weren't listed in #19571 were failing, but look pretty straightforward to fix/implement.

If anything comes up Awni has been very responsive, I think we can get a quick turnaround.

Issue #20794, opened 2025-01-22
Discrepancy in RandomZoom default fill_mode: Documentation states constant, but code uses reflect


gaetanDubuc:
System Information:

- Keras Version 3.8.0

# Issue Description

There is a discrepancy in the default value of the fill_mode parameter for the RandomZoom layer between the documentation and the actual implementation:

Documentation: States that the default fill_mode is constant.
Code: The default value appears to be reflect.

This inconsistency can lead to unexpected behavior when using RandomZoom without explicitly specifying the fill_mode.

# Steps to Reproduce

Check the Keras documentation for RandomZoom.
Check the implementation of RandomZoom in the source [code](https://github.com/keras-team/keras/blob/v3.8.0/keras/src/layers/preprocessing/image_preprocessing/random_zoom.py)

# Expected Behavior

The documentation and the code should align:

If the default is constant, the code should reflect that.
If the default is reflect, the documentation should be updated.

ARforyou:
same issue!

PR #20796, opened 2025-01-22
Update fill_mode argument default value in RandomZoom class

mehtamansi29:
Update fill_mode argument default value in RandomZoom class Fixed [#20794](https://github.com/keras-team/keras/issues/20794)


mehtamansi29:
> Thanks for the PR, it is the documentation we should fix not the default value. Reflect is the correct default.

Okay. Thanks @fchollet . Let me change the default value in the documentation that.

fchollet (approved):

Issue #20802, opened 2025-01-23
Refactoring: turn the attribute `_return_attention_scores` into a method argument


apehex:
Hey there! :space_invader: 

**Issue**

The class `MultiHeadAttention` is [using a private attribute to decide whether or not to return the attention scores][mha-check-line].

This behavior is not apparent from the signature of `_compute_attention`, because there are no corresponding argument.

So, if a sub-class fails to set `_return_attention_scores` in its custom `call` method, then `_compute_attention` never returns the scores.

For example, the child class [`CachedMultiHeadAttention` from KerasHub][cmha-issue] uses `_compute_attention` and receives `None` as attention scores in all cases.

**Additional context**

The attribute `_return_attention_scores` is used in a single method, `_compute_attention`.

**Proposition**

IMHO the attribute `_return_attention_scores` should be removed.

Instead, `_compute_attention` should have an extra parameter `return_attention_scores`.
Then the `call` methods could just forward this value to `_compute_attention` explicitely.

I'll write a PR shortly if you find this relevant.

[cmha-issue]: https://github.com/keras-team/keras-hub/issues/2055

[mha-check-line]: https://github.com/keras-team/keras/blob/v3.8.0/keras/src/layers/attention/multi_head_attention.py#455
[mha-save-line]: https://github.com/keras-team/keras/blob/v3.8.0/keras/src/layers/attention/multi_head_attention.py#L528

apehex:
Well it's been fixed, thanks!

PR #20803, opened 2025-01-23
Refactoring: turn the attribute `_return_attention_scores` into an argument

apehex:
I have a small refactoring for the class `MultiHeadAttention` :space_invader: 

I propose to replace the private attribute `_return_attention_scores` with an extra argument in `_compute_attention`.

This would make subclassing more straightforward and avoid cases where the attention scores are `None`.
This [issue shows in `CachedMultiHeadAttention`][cmha-issue] on KerasHub.

The rationale is explained in more details in #20802 

[cmha-issue]: https://github.com/keras-team/keras-hub/issues/2055


fchollet (approved):
Right, this is a good change. Thanks!

PR #20805, opened 2025-01-23
extract metrics update logic into a helper method

zskendall:
this change will allow users to customize what happens in the step function while being able to use existing metrics update logic without needing to duplicate it

### DIFF
--- a/keras/src/backend/jax/trainer.py
+++ b/keras/src/backend/jax/trainer.py
@@ -85,6 +85,28 @@ class JAXTrainer(base_trainer.Trainer):
             metrics_variables,
         )
 
+    def _update_metrics_variables(
+        self, metrics_variables, unscaled_loss, x, y, y_pred, sample_weight
+    ):
+        with backend.StatelessScope(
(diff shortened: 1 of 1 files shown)


fchollet (commented):
Thanks for the PR. The change is reasonable, but the fact that the method is JAX specific and very undocumented makes it challenging to have it as part of the public API. Let's rename it `_update_metrics_variables` -- you can still reuse it, but it's clearly signaled that it's internal and thus it's your responsibility to use it correctly

zskendall:
Ah, sorry, I'm not very familiar with the PR process. To add a little more context, this came up during code development for JAX parameter server training, which is why the change is JAX-specific. I originally implemented some custom logic but it was suggested to be consistent with how Keras already handles metrics. My goal with this change is to improve maintainability so that behaviour won't diverge for my (and others') subclassed models if the core Keras metrics handling logic is updated in future. Marking it internal makes sense to me, thanks for the suggestion!

fchollet (approved):
LGTM

PR #20810, opened 2025-01-24
Fix reference to nonexistent namespace in error message

jurca:
The error message produced when using, for example, a tensorflow math operation in a layer referenced a nonexistent `keras.operations` namespace (which makes fixing the issue a lot more difficult for newcomers, given that they will encounter it while following examples from the book Deep Learning with Python, 2nd edition).

The correct name of the implied namespace is `keras.ops`.


fchollet (approved):
LGTM, thanks for the fix

Issue #20355, opened 2024-10-15
Keras argmin function returns incorrect index when handling subnormal float values.


LilyDong0127:
When using the Keras backend `argmin` function on an input array containing subnormal float values, Keras consistently returns the index of `0.0` as the minimum value, even though a smaller subnormal value (`-1.401298464324817e-45`) exists in the array. Other deep learning frameworks such as PyTorch and Chainer correctly return the index of the subnormal value, but Keras (and TensorFlow) return the index of `0`.

### Expected Behavior:
The expected behavior is for Keras's `argmin` function to return the index of the smallest value, which should be the subnormal float value (`-1.401298464324817e-45`) at index 2. Instead, Keras is returning the index of `0.0` (index 0).

### Reproduction Code:
```
import torch
import tensorflow as tf
import numpy as np
from chainer import functions as F
import jax.numpy as jnp
import tensorflow.keras.backend as K

# Input data
input_data = [
    0.0,
    1.1754943508222875e-38,
    -1.401298464324817e-45,
    0.0,
    459367.0
]

# Test PyTorch
def test_pytorch_argmin(input_data):
    tensor = torch.tensor(input_data, dtype=torch.float32)
    result = torch.argmin(tensor).item()
    print(f"PyTorch argmin result: {result}")
    return result

# Test TensorFlow
def test_tensorflow_argmin(input_data):
    tensor = tf.constant(input_data, dtype=tf.float32)
    result = tf.argmin(tensor).numpy()
    print(f"TensorFlow argmin result: {result}")
    return result

# Test Keras using backend
def test_keras_argmin(input_data):
    tensor = K.constant(input_data, dtype=tf.float32)
    result = K.argmin(tensor, axis=-1).numpy()
    print(f"Keras argmin result: {result}")
    return result

# Test Chainer
def test_chainer_argmin(input_data):
    tensor = np.array(input_data, dtype=np.float32)
    result = F.argmin(tensor).data
    print(f"Chainer argmin result: {result}")
    return result

# Test JAX
def test_jax_argmin(input_data):
    tensor = jnp.array(input_data, dtype=jnp.float32)
    result = jnp.argmin(tensor).item()
    print(f"JAX argmin result: {result}")
    return result

if __name__ == "__main__":
    pytorch_result = test_pytorch_argmin(input_data)
    tensorflow_result = test_tensorflow_argmin(input_data)
    keras_result = test_keras_argmin(input_data)
    chainer_result = test_chainer_argmin(input_data)
[...]

PR #20812, opened 2025-01-25
fix(ops): Fix ops.argmin() handling of subnormal float values in Keras backends

harshaljanjani:
- Update JAX and NumPy backends to handle subnormal float comparisons (**TODO:** Ensure TensorFlow compatibility).
- Add test case to verify subnormal float value handling.
- Addresses issue #20355.


harshaljanjani:
**TODO:** Ensure TensorFlow compatibility.

fchollet (approved):

PR #20816, opened 2025-01-27
implement remaining functions in  `mlx.numpy.py`

acsweet:
This address #19571

This implements the remaining functions in `mlx.numpy.py`

A few notes:
- the following depend on jax
    - `linalg.det`
        - I started to implement `linalg.det` with `mlx.linalg.svd`, but due to verbosity decided to use `jax.numpy.linalg.det`
    - `slogdet`
        - decided to use jax's implementation while `linalg.det` also does
    - `nonzero`
    - `quantile`
- `einsum` fails with ellipsis in subscripts (for now)
   - recently fixed here https://github.com/ml-explore/mlx/issues/1717

Questions:
- a number of data type tests are failing, it seemed like they expected some default output data types, e.g. `int32`, `uint32`, and `float32` (according to some subtypes)
    - do we need to force the mlx backend functions to output these dtypes by default?
- is it okay to add some custom handling for `backend.backend() == "mlx"` in the tests, e.g. mlx doesn't support `float64` or `complex128` (numpy's default, which we can downcast to `complex64` when testing against mlx)

(After these latest updates I think a `linalg.det` implementation from mlx, as well as boolean indexing would be nice to have)

### DIFF
--- a/keras/src/backend/mlx/linalg.py
+++ b/keras/src/backend/mlx/linalg.py
@@ -7,7 +7,7 @@ from keras.src.backend.mlx.core import convert_to_tensor
 
 
 def det(a):
-    # TODO: Swap to mlx.linalg.det when it's implemented
+    # TODO: Swap to mlx.linalg.det when supported
     a = jnp.array(a)
     output = jnp.linalg.det(a)
     return mx.array(output)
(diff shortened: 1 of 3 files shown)


fchollet (approved):
LGTM, thank you!

PR #20827, opened 2025-01-30
Fix jax version

divyashreepathihalli:
The jax version currently is causing jax.nn.dot_product_attention import error
repro - https://colab.sandbox.google.com/drive/1dRNCwDLRUYcVpOul8V94GTo9HfN0dRoE#scrollTo=jKIdMM2reib0
I am not sure why the tests in Keras are not failing but KerasHub is seeing the following error
```Flash attention is not supported in your current JAX version. Please update it by following the official guide: https://jax.readthedocs.io/en/latest/installation.html```
coming from https://github.com/keras-team/keras/blob/14d9256108a65470512fa0340ecaf9d8eb5b003a/keras/src/backend/jax/nn.py#L1017-L1037

The root of the error being `AttributeError: module 'jax.nn' has no attribute 'dot_product_attention'`

### DIFF
--- a/requirements-jax-cuda.txt
+++ b/requirements-jax-cuda.txt
@@ -11,7 +11,7 @@ torch-xla
 # Jax with cuda support.
 # TODO: Higher version breaks CI.
 --find-links https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
-jax[cuda12]==0.4.28
+jax[cuda12]==0.5.0
 flax
 
 -r requirements-common.txt
--- a/requirements-jax-cuda.txt
+++ b/requirements-jax-cuda.txt
@@ -11,7 +11,7 @@ torch-xla
 # Jax with cuda support.
 # TODO: Higher version breaks CI.
 --find-links https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
-jax[cuda12]==0.5.0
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: requirements-jax-cuda.txt [hunk 1]
```diff
@@ -11,7 +11,7 @@ torch-xla
 # Jax with cuda support.
 # TODO: Higher version breaks CI.
 --find-links https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
-jax[cuda12]==0.4.28
+jax[cuda12]==0.5.0
```
hertschuh, line 14: Should we do `jax[cuda12]>=0.5.0` for consistency with what we do elsewhere?

mattdangerw:
Shouldn't we look into why tests aren't currently failing? If this error should be popping up with the current version of jax on CI and is not, sounds like we have a coverage issue.

divyashreepathihalli:
but I do see the test coverage for this here - https://github.com/keras-team/keras/blob/14d9256108a65470512fa0340ecaf9d8eb5b003a/keras/src/ops/nn_test.py#L2452

fchollet (approved):

james77777778:
We will need to use `python>=3.10` for the testing env due to the compatibility of `jax>=0.5.0`

PR #20832, opened 2025-01-31
mlx - `nn` and `core` updates post merge

acsweet:
This addresses #19571

Patched and implemented missing methods in:
- `nn.py` (a lot has been added since the `mlx` branch started!)
- `core.py`
- `rnn.py` (very simple `scan` update)

Started to adjust tests for various layers that `mlx` doesn't support:
- `flash_attention`
- dtypes `float64`, `complex128`


fchollet (approved):
LGTM, thank you!

PR #20851, opened 2025-02-03
Don't duplicate frozen parameters during predict()

mattdangerw:
On the Jax backend we were not using donate_argnums during `predict()`. This works when a model is mostly trainable, but when a model is mostly or all frozen, this will result in 2x the memory jump (which is why we use donate_argnums for fit and evaluate).

This change adds `donate_argnums` to the predict function to avoid the memory spike. But because this means all incoming state (including the trainable variables) will be deleted by jax, this means we need to sync the trainable variables state much like in fit and evaluate. An alternative would be to change the predict_step signature (so we could only donate non-trainable variables), but this would be a breaking change and confusing.

### DIFF
--- a/keras/src/backend/jax/trainer.py
+++ b/keras/src/backend/jax/trainer.py
@@ -308,7 +308,7 @@ class JAXTrainer(base_trainer.Trainer):
             return outputs, (state[0], non_trainable_variables)
 
         if not self.run_eagerly and self.jit_compile:
-            predict_step = jax.jit(predict_step)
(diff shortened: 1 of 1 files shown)


mattdangerw:
Notably this will come up for lora + transformers.

```python
gemma = keras_hub.CausalLM.from_preset("gemma...")
gemma.backbone.enable_lora(4)
gemma.fit(...)
gemma.predict(example) # double optimal memory usage do to doubled frozen params
```

fchollet (approved):
LGTM, thanks for the fix!

PR #20854, opened 2025-02-04
Fix convnext to work with any custom input tensors

Inquisitive-ME:
Convnext fails with custom inputs that are more than 1 layer, it looks like a bug. I added a test that highlights the problem and fails for all Convnext and then fixed the issue.


fchollet (commented):
Thanks for the fix! Please fix the code format (run `sh shell/format.sh`)

fchollet (approved):
LGTM, thank you

PR #20856, opened 2025-02-04
MLX segment_sum and segment_max implementation

fbadine:
This PR is an MLX backend implementation of reduce_sum and reduce_max in backend/mlx/math.py
The relevant tests in math_test.py are now successful.


fchollet (approved):
Thanks for the PR! LGTM

PR #20862, opened 2025-02-05
mlx - passing most layers tests 

acsweet:
This addresses #19571

Most tests in `keras/src/layers` are passing. This included updated and implemented ops functions (especially in `random`, `image`, and `numpy`)
- is the change in `layers/preprocessing/image_preprocessing/random_crop.py RandomCrop.get_random_transformation()` okay?
    - mlx can't slice arrays with `mlx.arrays` (incl scalar values)
- `mlx.random.py beta()` might need adjusting how it checks if `shape` is broadcastable with `alpha` and `beta`

Couple mlx notes, these be will fixed in their next patch release:
- `mlx.core.einsum` was failing with ellipsis
- gradient updates were failing with some reverse indexing, e.g. using `::-1` (this affected transpose convolutions), it's currently being handled with a reversed index array and `mlx.core.take`

(Sorry this got a bit large! I wanted to merge back to the mlx branch before continuing)


fchollet (approved):

PR #20865, opened 2025-02-05
Clear JAX state sharding after `fit`, `evaluate` and `predict`.

hertschuh:
The state sharding is leaked at the end of `fit`, `evaluate` and `predict`. The values are not reused if `fit`, `evaluate` and `predict` is called again.


divyashreepathihalli (approved):
LGTM

PR #20867, opened 2025-02-06
fix(layers): Update Conv2D docstring to clarify numerical precision across backends

### LINKED ISSUE #20804: Results of the Conv2D layer are not identical across backends
The [docs](https://keras.io/keras_3/) state:

> numerics are identical across backends [...]
> [...] up to 1e-7 precision in float32, per function execution

However, this minimal example does not confirm this:

```python3
import os.path

import numpy as np
from keras import layers, Model
from keras.src.saving import load_model

np.random.seed(0)
data = np.random.rand(1, 256, 256, 1024)

if os.path.isfile("model.keras"):
    model = load_model("model.keras")
else:
    inputs = layers.Input(shape=(256, 256, 1024))
    outputs = layers.Conv2D(1024, kernel_size=(4, 7), padding="same", dilation_rate=(3, 2))(inputs)
    model = Model(inputs=[inputs], outputs=outputs)
    model.save("model.keras")

print(np.sum([data]))
print(os.environ["KERAS_BACKEND"])
print(np.sum(np.array(model([data]))))
```

Output with `tensorflow` backend:

```bash
KERAS_BACKEND=tensorflow python main.py
```

```
33550919.07926151
tensorflow
58094.56
```

Output with `jax` backend:

```bash
KERAS_BACKEND=jax python main.py
```

```
33550919.07926151
jax
58094.523
```

Versions used:

```bash
python -c "import keras; import tensorflow; import jax;  print(keras.__version__); print(tensorflow.__version__); print(jax.__version__)"
```

```
3.8.0
2.18.0
0.5.0
```

harshaljanjani:
- Make it clear that, contrary to what the Keras 3 [docs](https://keras.io/keras_3/) state, Conv2D operations may exceed the documented 1e-7 precision difference across backends.
- Document that large convolutions can show notable variations due to accumulated floating-point operations.

Fixes #20804.

### DIFF
--- a/keras/src/layers/convolutional/conv2d.py
+++ b/keras/src/layers/convolutional/conv2d.py
@@ -12,6 +12,15 @@ class Conv2D(BaseConv):
     and added to the outputs. Finally, if `activation` is not `None`, it is
     applied to the outputs as well.
 
+    Note on numerical precision: While Keras documentation states that results
+    are identical across backends up to 1e-7 precision in float32 per function
+    execution, Conv2D operations may show larger variations. Due to the large
+    number of element-wise multiplications and additions in convolution
+    operations, especially with large inputs or kernel sizes, accumulated
+    floating-point differences can exceed this 1e-7 threshold. These variations
+    are particularly noticeable when using different backends (e.g., TensorFlow
+    vs JAX) or hardware implementations.
+
     Args:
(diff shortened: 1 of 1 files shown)


fchollet (approved):
Thank you!

PR #20868, opened 2025-02-06
Remove `torchvision` dep and simplify `resize` and `rgb_to_grayscale` in torch backend

james77777778:
We can simplify the logic of `resize` and `rgb_to_grayscale` and remove the dep of `torchvision`


fchollet (approved):
Very nice, thank you!

PR #20877, opened 2025-02-08
Fix Torch GPU CI

james77777778:
This PR ensures the exposed APIs convert the inputs to the tensors.


shashaka:
@james77777778 
Thank you for resolving this issue.

fchollet (approved):
Thank you for the fix!

PR #20879, opened 2025-02-08
[Fix] solve method on linalg

shashaka:
The error occurred while using the map() and solve() methods with a TensorFlow dataset. To resolve this issue, I modified the source code accordingly.

here is my [gist](https://colab.research.google.com/drive/1dt5Kgx-bhALHkdMjbebVYAogyqkLPFp9?usp=sharing)


fchollet (approved):
LGTM -- indeed, `tf.rank` returns a tensor. `.ndims` is a Python int (you could also use `len(b.shape)`)

PR #20882, opened 2025-02-09
fix(ops): Fix TensorFlow backend keras.ops.rot90 shape transformation and improve test coverage

harshaljanjani:
- Fix shape handling in `keras.ops.rot90` to correctly swap height/width dimensions based on k rotations.
- Refactor test suite to use parameterized test cases and cover edge conditions more thoroughly.


fchollet (approved):
Thank you for the fix!

PR #20889, opened 2025-02-10
`HashedCrossing` layer preserves the static batch size when known.

hertschuh:
Previously, the output of `HashedCrossing` would always have `None` batch size as a result of the underlying Tensorflow `tf.sparse.cross_hashed`.

The previous reshaping logic in `HashedCrossing` would fix the last dimension (expected to be 1) but not the batch dimension.


fchollet (approved):
LGTM, thank you

PR #20891, opened 2025-02-10
Improve Model.layers setter error message for clarity in model.py

### LINKED ISSUE #20890: Improve Model.layers setter error message
Current error message for Model.layers setter is ambiguous and doesn't explain how to properly handle layers.

Current:
“`Model.layers` attribute is reserved and should not be used. Please use another name.”

Proposed:
“`Model.layers` is a read-only property. Use Model.add() to add new layers.”

nikolasavic3:
Fixes #20890

### DIFF
--- a/keras/src/models/model.py
+++ b/keras/src/models/model.py
@@ -170,8 +170,10 @@ class Model(Trainer, base_trainer.Trainer, Layer):
     @layers.setter
     def layers(self, _):
         raise AttributeError(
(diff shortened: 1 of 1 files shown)


fchollet (commented):
Please note, `Model.add()` is only valid for Sequential models. We cannot recommend it in the general case (in general people are subclassing `Model`, not `Sequential`).

PR #20900, opened 2025-02-13
Fix tril/triu ops

abheesht17:
Tril/triu ops fail (post compilation) because `tf.cond` expects `k` to be a tensor:

```
>>> inputs = keras.Input((5,))
>>> outputs = keras.ops.tril(inputs)
>>> model = keras.Model(inputs, outputs)
>>> model.predict(keras.ops.ones((2, 5)))
  File "<stdin>", line 1, in <module>
  File "/Users/abheesht/Work/repos/keras/keras/src/utils/traceback_utils.py", line 122, in error_handler
  File "/Users/abheesht/Work/repos/keras/keras/src/utils/traceback_utils.py", line 122, in error_handler
TypeError: Exception encountered when calling Tril.call().

pred must not be a Python bool

Arguments received by Tril.call():
  • x=tf.Tensor(shape=(2, 5), dtype=float32)
```

Post this PR, this gets fixed:

```
>>> model.predict(keras.ops.ones((2, 5)))
1/1 ━━━━━━━━━━━━━━━━━━━━ 0s 24ms/step
array([[1., 0., 0., 0., 0.],
       [1., 1., 0., 0., 0.]], dtype=float32)
```

### DIFF
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -2382,9 +2382,14 @@ def tril(x, k=0):
         mask = i >= j - k
         return tf.where(tf.broadcast_to(mask, shape), x, tf.zeros_like(x))
 
-    if k >= 0:
-        return tf.linalg.band_part(x, -1, k)
-    return _negative_k_branch()
+    if isinstance(k, int):
+        if k >= 0:
+            return tf.linalg.band_part(x, -1, k)
+        return _negative_k_branch()
+
+    return tf.cond(
+        k >= 0, lambda: tf.linalg.band_part(x, -1, k), _negative_k_branch
+    )
 
 
 def triu(x, k=0):
@@ -2397,9 +2402,14 @@ def triu(x, k=0):
         mask = i <= j - k
         return tf.where(tf.broadcast_to(mask, shape), x, tf.zeros_like(x))
 
-    if k <= 0:
-        return tf.linalg.band_part(x, -k, -1)
-    return _positive_k_branch()
+    if isinstance(k, int):
+        if k <= 0:
+            return tf.linalg.band_part(x, -k, -1)
+        return _positive_k_branch()
+
+    return tf.cond(
+        k <= 0, lambda: tf.linalg.band_part(x, -k, -1), _positive_k_branch
+    )
 
 
 def trunc(x):
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -2389,7 +2389,9 @@ def tril(x, k=0):
 
     # when `k` is a tensor
     return tf.cond(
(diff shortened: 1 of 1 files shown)


keras/src/backend/tensorflow/numpy.py
```diff
@@ -2382,9 +2382,9 @@ def _negative_k_branch():
         mask = i >= j - k
         return tf.where(tf.broadcast_to(mask, shape), x, tf.zeros_like(x))
 
-    return tf.cond(
-        k >= 0, lambda: tf.linalg.band_part(x, -1, k), _negative_k_branch
-    )
+    if k >= 0:
+        return tf.linalg.band_part(x, -1, k)
+    return _negative_k_branch()
```
hertschuh, line 2387: You will need a mixed approach to support both constants and tensors:

```
if isinstance(k, int):
  if k >= 0:
    return tf.linalg.band_part(x, -1, k)
  else:
    return _negative_k_branch()
else:
  return tf.cond(
     k >= 0, lambda: tf.linalg.band_part(x, -1, k), _negative_k_branch
  )
```

hertschuh (changes requested):

keras/src/backend/tensorflow/numpy.py
```diff
@@ -2382,6 +2382,12 @@ def _negative_k_branch():
...
         return tf.where(tf.broadcast_to(mask, shape), x, tf.zeros_like(x))
 
+    if isinstance(k, int):
+        if k >= 0:
+            return tf.linalg.band_part(x, -1, k)
+        return _negative_k_branch()
+
+    # when `k` is a tensor
     return tf.cond(
         k >= 0, lambda: tf.linalg.band_part(x, -1, k), _negative_k_branch
```
hertschuh, line 2392: Actually, we turn off autograph, so this should be: `tf.greater_equal(k, 0)` instead of `k >= 0`.

I think it works with the `__ge__` override, but I would prefer to be explicit.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -2397,6 +2403,12 @@ def _positive_k_branch():
...
         return tf.where(tf.broadcast_to(mask, shape), x, tf.zeros_like(x))
 
+    if isinstance(k, int):
+        if k <= 0:
+            return tf.linalg.band_part(x, -k, -1)
+        return _positive_k_branch()
+
+    # when `k` is a tensor
     return tf.cond(
         k <= 0, lambda: tf.linalg.band_part(x, -k, -1), _positive_k_branch
```
hertschuh, line 2413: `tf.less_equal(k, 0)` instead of `k <= 0`.

hertschuh (changes requested):

hertschuh (approved):

hertschuh (approved):
Thanks for the fix!

PR #20901, opened 2025-02-13
Add iterations property to LossScaleOptimizer

### LINKED ISSUE #20878: Loss scale optimizer does not increment iterations
Loss scale optimizer does not automatically increment iteration counter when `apply` method is called.

As a side effect, this breaks `evaluation_loss_vs_iterations` metric in TensorBoard when mixed precision is enabled.

See example here: [colab](https://colab.research.google.com/gist/itmo153277/295f75633b6b4f5dd9485027a93ec5a6/lossscaleiterations.ipynb)

jeffcarp:
Fixes #20878. TensorBoard isn't able to report the correct step because this optimizer doesn't forward the `iterations` property.


fchollet (approved):
Thanks for the PR -- please fix code format.

jeffcarp:
Oops - done, thanks! Merging since it's approved.

PR #20906, opened 2025-02-14
Always allow `ExportArchive.track` to track TensorFlow resources.

hertschuh:
Previously, `track` would only work with `Layer`s or `Model`s unless the backend was TensorFlow. It would raise an error on JAX for instance.

It is now possible to export saved models with a mix of Keras models and TensorFlow native preprocessing involving resources even with the JAX backend.

- Added example on how to use `ExportArchive` to export a function combining a model with some TensorFlow native preprocessing with a resource.
- Added unit test testing the combining of a model with some TensorFlow native preprocessing with a resource.
- Renamed `track` to `_track_layer` in backend specific `ExportArchive` classes because that is the use case.
- Use `super()` instead of `BackendExportArchive` for consistency.

### DIFF
--- a/keras/src/backend/jax/export.py
+++ b/keras/src/backend/jax/export.py
@@ -4,7 +4,6 @@ import itertools
 import string
 import warnings
 
-from keras.src import layers
 from keras.src import tree
 from keras.src.backend.common.stateless_scope import StatelessScope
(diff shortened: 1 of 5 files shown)


fchollet (approved):
LGTM, thanks!

PR #20907, opened 2025-02-15
`mlx` - `stft`, `istft`, and dtype handling in `numpy.py`

acsweet:
This addresses #19571
- implements `stft` and `istft`
- aligns dtype handling in `mlx/numpy.py` with other backends (matmul, tensordot, etc. still in progress)
- swaps jax dependency to numpy (for ease of testing)


fchollet (approved):
LGTM, thank you!

PR #20913, opened 2025-02-17
[OpenVINO backend] Support numpy.argsort

nikolasavic3:
Implements the numpy.argsort operation for the Keras 3 OpenVINO backend.
Fixes [#29012](https://github.com/openvinotoolkit/openvino/issues/29012)

@rkazants, could you please review?

### DIFF
--- a/keras/src/backend/openvino/core.py
+++ b/keras/src/backend/openvino/core.py
@@ -73,7 +73,10 @@ def get_ov_output(x, ov_type=None):
             ov_type = Type.i32
         x = ov_opset.constant(x, ov_type).output(0)
     elif isinstance(x, np.ndarray):
-        x = ov_opset.constant(x).output(0)
+        if x.dtype == np.dtype("bfloat16"):
+            x = ov_opset.constant(x, OPENVINO_DTYPES["bfloat16"]).output(0)
+        else:
+            x = ov_opset.constant(x).output(0)
     elif np.isscalar(x):
         x = ov_opset.constant(x).output(0)
     elif isinstance(x, KerasVariable):
--- a/keras/src/backend/openvino/excluded_concrete_tests.txt
+++ b/keras/src/backend/openvino/excluded_concrete_tests.txt
@@ -93,7 +93,6 @@ NumpyOneInputOpsCorrectnessTest::test_any
 NumpyOneInputOpsCorrectnessTest::test_argmax
 NumpyOneInputOpsCorrectnessTest::test_argmin
 NumpyOneInputOpsCorrectnessTest::test_argpartition
-NumpyOneInputOpsCorrectnessTest::test_argsort
 NumpyOneInputOpsCorrectnessTest::test_array
 NumpyOneInputOpsCorrectnessTest::test_average
 NumpyOneInputOpsCorrectnessTest::test_bincount
--- a/keras/src/backend/openvino/excluded_concrete_tests.txt
+++ b/keras/src/backend/openvino/excluded_concrete_tests.txt
@@ -10,7 +10,6 @@ NumpyDtypeTest::test_arctan2_
 NumpyDtypeTest::test_argmax
 NumpyDtypeTest::test_argmin
 NumpyDtypeTest::test_argpartition
-NumpyDtypeTest::test_argsort
 NumpyDtypeTest::test_array
 NumpyDtypeTest::test_average_
 NumpyDtypeTest::test_bincount
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -304,9 +304,35 @@ def argmin(x, axis=None, keepdims=False):
 
 
 def argsort(x, axis=-1):
-    raise NotImplementedError(
-        "`argsort` is not supported with openvino backend"
-    )
+    axis_dim = None
+    x = get_ov_output(x)
+    x_shape = x.get_partial_shape()
+    rank = x_shape.rank.get_length()
+    if rank == 0:
+        return OpenVINOKerasTensor(ov_opset.constant([0], Type.i32).output(0))
+    if axis is None:
+        flatten_shape = ov_opset.constant([-1], Type.i32).output(0)
+        x = ov_opset.reshape(x, flatten_shape, False).output(0)
+        print("shape_of: ", shape_1 := ov_opset.shape_of(x, Type.i32).output(0))
+        k = ov_opset.reduce_prod(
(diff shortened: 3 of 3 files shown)


keras/src/backend/openvino/excluded_concrete_tests.txt
```diff
@@ -10,7 +10,6 @@ NumpyDtypeTest::test_arctan2_
 NumpyDtypeTest::test_argmax
 NumpyDtypeTest::test_argmin
 NumpyDtypeTest::test_argpartition
-NumpyDtypeTest::test_argsort
```
rkazants, line 13: please exclude another test for correctness as well

keras/src/backend/openvino/core.py
```diff
@@ -73,7 +73,10 @@ def get_ov_output(x, ov_type=None):
             ov_type = Type.i32
         x = ov_opset.constant(x, ov_type).output(0)
     elif isinstance(x, np.ndarray):
-        x = ov_opset.constant(x).output(0)
+        if x.dtype == np.dtype("bfloat16"):
```
nikolasavic3, line 76: I added this because ov_opset.constant needs the explicit type of the np.ndarray. Should there be added checks for other types as well?

rkazants, line 76: I don't understand why we need this change. I even see that numpy has no built-in type for bfloat16:
{1A792B0D-8839-45D1-9776-239C46C4759B}

nikolasavic3: Hi!,
I added that check because the method `ov_opset `requires that the type is explicitly provided as the argument in the case`x` is `bfloat16`. Without specifying the type I was getting the error in the ov_opset funciton:
```
> return Constant(_value, shared_memory=_shared_memory)
E       IndexError: map::at
```
I was getting that error only for type `bfloat16`.

adding:
` if x.dtype == np.dtype("bfloat16"):`
enables me to call ov_constant with the dtype argument:
`x = ov_opset.constant(x, OPENVINO_DTYPES["bfloat16"]).output(0)
`
then inside `ov_constant`, the openvino Constant method can be executed without an error :
```
elif dtype in [Type.bf16, Type.f8e8m0, Type.f8e4m3, Type.f8e5m2]:
            display_shared_memory_warning(f"Constant initialized with OpenVINO custom {dtype}")
            return Constant(dtype, Shape(_value.shape), _value.flatten().tolist()) 
```

nikolasavic3: When I say constant method I mean: https://github.com/openvinotoolkit/openvino/blob/a4bfef7cd1eb28b7044adf1dfc58b223345c2eee/src/bindings/python/src/openvino/opset13/ops.py#L276

rkazants: Hi @nikolasavic3,

I do not understand how this code works on your side `x.dtype == np.dtype("bfloat16")`. For me, `np.dtype("bfloat16")` raises exception `TypeError`. Does it work for you?

nikolasavic3: Hi @rkazants,
Yes, this passes the 
keras/src/ops/numpy_test.py::NumpyDtypeTest::test_argsort_bfloat16 and other argsort related tests for me locally and on the github actions CI tests. 
image
`np.dtype("bfloat16")` raises `TypeError` since bfloat16 isn't nativly supported by NumPy.
After importing Keras, bfloat16 is automatically registered with NumPy.

I understand that you are concerned with this line of code, because it seems it lacks robustness. Since this code is in keras/src/backend/openvino/core.py, keras is already imported so bfloat16 is always available in this context.

keras/src/backend/openvino/numpy.py
```diff
@@ -304,9 +304,33 @@ def argmin(x, axis=None, keepdims=False):
...
+    if rank == 0:
+        return OpenVINOKerasTensor(ov_opset.constant([0], Type.i32).output(0))
+    if axis is None:
+        flatten_shape = ov_opset.constant([-1], Type.i32).output(0)
+        x = ov_opset.reshape(x, flatten_shape, False).output(0)
+        x_shape = ov_opset.shape_of(x, Type.i32).output(0)
+        k = ov_opset.reduce_prod(
+            x_shape, ov_opset.constant([0], Type.i32), keep_dims=False
+        )
+        axis_dim = x.shape[0]
```
rkazants, line 319:
```suggestion
```
looks redundant

keras/src/backend/openvino/numpy.py
```diff
@@ -304,9 +304,33 @@ def argmin(x, axis=None, keepdims=False):
...
+        axis_dim = k
+        axis = 0
+    else:
+        if axis < 0:
+            axis = rank + axis
```
rkazants, line 324:
```suggestion
    if axis < 0:
        axis = rank + axis
```

keras/src/backend/openvino/numpy.py
```diff
@@ -304,9 +304,33 @@ def argmin(x, axis=None, keepdims=False):
...
+        return OpenVINOKerasTensor(ov_opset.constant([0], Type.i32).output(0))
+    if axis is None:
+        flatten_shape = ov_opset.constant([-1], Type.i32).output(0)
+        x = ov_opset.reshape(x, flatten_shape, False).output(0)
+        x_shape = ov_opset.shape_of(x, Type.i32).output(0)
+        k = ov_opset.reduce_prod(
+            x_shape, ov_opset.constant([0], Type.i32), keep_dims=False
+        )
+        axis_dim = x.shape[0]
+        axis_dim = k
```
rkazants, line 320: please remove `axis_dim` and use just `k` to pass into `topk` construction

keras/src/backend/openvino/numpy.py
```diff
@@ -304,9 +304,33 @@ def argmin(x, axis=None, keepdims=False):
...
+        axis = 0
+    else:
+        if axis < 0:
+            axis = rank + axis
+        axis_dim = x_shape[axis].get_length()
```
rkazants, line 325: it is not always working. For example, for dynamic shapes it won't. In order to solve it, please use graph with `shapeof` and `gather` operations to retrieve the dimension size

keras/src/backend/openvino/core.py
rkazants: Got it clear. Then let us leave it here

rkazants (commented):
@nikolasavic3, please handle dynamic shape case. Other part looks good.

rkazants (commented):

PR #20921, opened 2025-02-18
[OpenVINO backend] Support arctan2. (#29010)

AsVoider:
This is implementation of numpy.arctan2 on the OpenVINO backend.

@rkazants, could you please review?


rkazants (commented):
Good job, thank you!

@fchollet, could you please merge it?

PR #20922, opened 2025-02-18
[OpenVINO backend] Support np.average

Mohamed-Ashraf273:
adding np.average support to openvino BE.
@rkazants , 
I've made this new pr due to the many unwanted commits in the old one, can you please review it.

### DIFF
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -291,7 +291,7 @@ def average(x, axis=None, weights=None):
         weighted_sum = ov_opset.multiply(x, weights)
     else:
         x = get_ov_output(x)
-        weighted_sum = x  
+        weighted_sum = x
     if axis is None:
         flatten_shape = ov_opset.constant([-1], Type.i32).output(0)
         x = ov_opset.reshape(x, flatten_shape, False).output(0)
@@ -303,7 +303,7 @@ def average(x, axis=None, weights=None):
     if axis == () or axis == []:
         return OpenVINOKerasTensor(weighted_sum)
     axis_const = ov_opset.constant(axis, dtype=Type.i32).output(0)
-    mean_ops = ov_opset.reduce_mean(weighted_sum, axis_const, False)  
+    mean_ops = ov_opset.reduce_mean(weighted_sum, axis_const, False)
     return OpenVINOKerasTensor(mean_ops.output(0))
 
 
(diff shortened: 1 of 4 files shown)


Mohamed-Ashraf273:
hi @rkazants 
I have finished this pr and all tests are passed, can you review it.

keras/src/backend/openvino/numpy.py
```diff
@@ -278,9 +278,33 @@ def array(x, dtype=None):
...
+        weighted_sum = ov_opset.multiply(x, weights)
+    else:
+        x = get_ov_output(x)
+        weighted_sum = x
+    if axis is None:
+        flatten_shape = ov_opset.constant([-1], Type.i32).output(0)
+        x = ov_opset.reshape(x, flatten_shape, False).output(0)
+        if weights is not None:
+            weights = ov_opset.reshape(weights, flatten_shape, False).output(0)
+        axis = 0
```
rkazants, line 300: looks unused piece of code so the case of None axis looks not handled

rkazants: You computed `weights` but I don't see that it is used below in the code. Please clarify.

Mohamed-Ashraf273: it handles the axis = none in case the weight is none too, because the weighted sum is referencing to the same object x does, so when flattening x weighted sum also gets flattened, the missing test case is if axis = none and weights are not.
so I will modify  it and add the test case

Mohamed-Ashraf273: It handles the axis = none in case the weights is none too because the weighted sum is referencing the same object x does (as I defined it, weighted_sum = None at the beginning of the code), so when flattening x weighted sum also gets flattened, the missing test case is if axis = none and weights are not.
soSo will modify it and add the test case

Mohamed-Ashraf273: @rkazants 
I've added the test case and modified it, can you please review it?

Mohamed-Ashraf273:
@rkazants 
I've added the test case and modified the function, can you please review it?

Mohamed-Ashraf273:
@rkazants 
when I added the new test case, a torch test failed because it didn't handle it.
may I modify it?
or just leave it?

Mohamed-Ashraf273:
I closed this PR due to too many unnecessary commits, I made another PR, #20925, and added the missing test case and torch modification cause it failed due to this test case.

PR #20923, opened 2025-02-18
Only load OpenVINO excludes file when backend is "openvino".

hertschuh:
It is not necessary to decorate excluded openvino tests with other backends.


fchollet (approved):
Thanks!

PR #20925, opened 2025-02-18
[OpenVINO backend] Support np.average

Mohamed-Ashraf273:
adding np.average support to openvino BE.
@rkazants ,
I've made this new pr due to the many unwanted commits in the old one, and here I solved this problem of 
the test case (weights != None ) and (axis = None), which fails in troch and I modified it too.
so this is the final pr and its ready to be reviewed.


Mohamed-Ashraf273:
@rkazants 
this is the final pr for which I modified the code in #20922 to handle the (axis = none and weights != none) test case and modified the orch to handle it cause it was failing too, this pr is ready to be reviewed.

Mohamed-Ashraf273:
Hi @rkazants 
Can you please review my pr?

Mohamed-Ashraf273:
Hi @rkazants 
do i need to remove the test case and the torch modification?

keras/src/backend/openvino/excluded_concrete_tests.txt
```diff
@@ -9,7 +9,6 @@ NumpyOneInputOpsCorrectnessTest::test_argmin
 NumpyOneInputOpsCorrectnessTest::test_argpartition
 NumpyOneInputOpsCorrectnessTest::test_argsort
 NumpyOneInputOpsCorrectnessTest::test_array
-NumpyOneInputOpsCorrectnessTest::test_average
```
rkazants, line 12: please also include NumPyDType test for average. We just recently merge PR for this.

https://github.com/keras-team/keras/pull/20929/files#diff-60b6b6e36719bf6926e4fcd3ca37f40d5b2ab11e55a3d4b3f40e79a092d64d23R15

keras/src/backend/openvino/numpy.py
```diff
@@ -278,9 +278,34 @@ def array(x, dtype=None):
...
+    if axis is None:
+        flatten_shape = ov_opset.constant([-1], Type.i32).output(0)
+        x = ov_opset.reshape(x, flatten_shape, False).output(0)
+        if weights is not None:
+            weights = ov_opset.reshape(weights, flatten_shape, False).output(0)
+        axis = 0
+    else:
+        x = get_ov_output(x, element_type)
+        if weights is not None:
+            weights = get_ov_output(weights, element_type)
```
rkazants, line 296: Please move this code block to the beginning and work with OV `weights`

PR #20927, opened 2025-02-19
`mlx` - numpy.searchsorted and numpy.historgram implemented

acsweet:
This addresses #19571

This implements:
- `numpy.searchsorted`
- `numpy.histogram`
- `integration_tests/mlx_custom_fit_test.py` (if this was okay to add)
    - very similar to jax

Question/comment:
- mlx added float64, but only on the cpu. I'm not sure the cleanest way to handle this, any thoughts? 
    - I was thinking about a custom wrapper that executed functions within a context manager to the cpu when a float64 is present
    - usually, it's (a) passed as the stream, (b) with a context manager, or (c) setting the default device
```python
import mlx.core as mx

x = mx.array([1, 2, 3], dtype=mx.float64)
y = mx.array([4, 5, 6], dtype=mx.float64)

z = mx.add(x, y, stream=mx.cpu) # (a)

with mx.stream(mx.cpu): # (b)
    z = mx.add(x, y)

mx.set_default_device(mx.cpu) # (c)
z = mx.add(x, y)
```

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -78,6 +78,10 @@ jobs:
         if: ${{ matrix.backend == 'jax'}}
         run: |
           python integration_tests/jax_custom_fit_test.py
+      - name: Test MLX-specific integrations
+        if: ${{ matrix.backend == 'mlx'}}
+        run: |
+          python integration_tests/mlx_custom_fit_test.py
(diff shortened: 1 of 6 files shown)


fchollet (approved):
LGTM -- thank you!

PR #20929, opened 2025-02-19
[OpenVINO Backend] Include NumpyDtype tests

rkazants:
**Details:** Include NumpyDtype tests


rkazants:
@fchollet, could you please review it? It includes tests to check output types for Numpy ops support.

fchollet (approved):

PR #20930, opened 2025-02-20
[Keras Ops] Add `keras.ops.polar` operation 

DavidLandup0:
### Why?

Specifically to make it simpler to have multi-backend `polar()` for DeepSeek in KerasHub (https://github.com/keras-team/keras-hub/issues/2077), but is otherwise applicable as a Keras operation.

Equivalent to `torch.polar()`. Note that this is _not_ equivalent to SciPy's [polar()](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.polar.html), which performs decomposition.

### Basic Usage

```
>>> import keras
>>> import numpy as np
>>> import torch
>>> 
>>> abs = np.array([1, 2])
>>> angle = [np.pi / 2, 5 * np.pi / 4]
>>> keras.src.ops.polar(abs, angle)
Array([-4.3711388e-08+1.j       , -1.4142137e+00-1.4142134j], dtype=complex64)
>>> torch.polar(torch.tensor(abs).float(), torch.tensor(angle).float())
tensor([-4.3711e-08+1.0000j, -1.4142e+00-1.4142j])
```

### DIFF
--- a/keras/api/_tf_keras/keras/ops/__init__.py
+++ b/keras/api/_tf_keras/keras/ops/__init__.py
@@ -87,6 +87,7 @@ from keras.src.ops.nn import moments
 from keras.src.ops.nn import multi_hot
 from keras.src.ops.nn import normalize
 from keras.src.ops.nn import one_hot
+from keras.src.ops.nn import polar
 from keras.src.ops.nn import psnr
 from keras.src.ops.nn import relu
 from keras.src.ops.nn import relu6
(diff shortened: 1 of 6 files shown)


fchollet (commented):
Thanks for the PR! Please take a look at the test failure:


```
FAILED keras/src/ops/nn_test.py::NNOpsCorrectnessTest::test_polar_corectness - tensorflow.python.framework.errors_impl.InvalidArgumentError: cannot compute Mul as input #1(zero-based) was expected to be a int64 tensor but is a float tensor [Op:Mul] name:
```

DavidLandup0:
> Thanks for the PR! Please take a look at the test failure:
> 
> ```
> FAILED keras/src/ops/nn_test.py::NNOpsCorrectnessTest::test_polar_corectness - tensorflow.python.framework.errors_impl.InvalidArgumentError: cannot compute Mul as input #1(zero-based) was expected to be a int64 tensor but is a float tensor [Op:Mul] name:
> ```

Whoops - forgot to commit/push the fix. Thanks for pointing this out!

fchollet (approved):
LGTM, thank you!

PR #20932, opened 2025-02-20
Remove unused dependency

abheesht17:
`google` isn't used anywhere in Keras. It was added by mistake in a previous commit.

GitHub Page: https://github.com/MarioVilas/googlesearch/
PyPi Page: https://pypi.org/project/google/


hertschuh (approved):
Good catch.

PR #20939, opened 2025-02-22
[OpenVINO backend] Support numpy.argmax and numpy.argmax

chiruu12:
@fchollet, could you please review?


chiruu12 (commented):
I was trying to make the code more readable but forgot to replace res with result at that place ! sorry for the inconvenience.

PR #20941, opened 2025-02-22
[Openvino Backend] support numpy.arange, modify dtype check

AsVoider:
Support numpy.arange operation for openvino backend and pass the tests. 
Besides, I add dtype checks and conversions to other operations to avoid situations where the dtype is a built-in Python type, such as **dtype=int**.


AsVoider:
This is a pr related to [#29009](https://github.com/openvinotoolkit/openvino/issues/29009) and some additional modifications.
@rkazants , could you please review?

rkazants:
Looks good to me.
@fchollet, recommend to merge.

PR #20954, opened 2025-02-24
Fix mean metrics to allow dict/list inputs

abheesht17:
Fixes internal tests broken by https://github.com/keras-team/keras/pull/20782


hertschuh (approved):
LGTM, thanks!

PR #20957, opened 2025-02-25
`mlx` - trainer update for compilation and data adapters

acsweet:
This addresses #19571

This update includes:
- `MLXTrainer` update for compilation (heavily mimics jax's trainer)
- all data adapters work with mlx
- faster `bincount`
- added `mlx.core.enable_float64()` similar to jax's `enable_x64()` to put in tests and let the user manage `float64` (must be on cpu)

Note:
- compilation in `MLXTrainer` works great except when `ops.cond` is involved, but can mostly be handled by evaluating both branches and doing a `where`. The most notable occurrence of this is in `loss_scale_optimizer`. I'm going to propose a `_mlx_stateless_apply()` in the next PR that avoids `where`

### DIFF
--- a/keras/src/backend/common/dtypes_test.py
+++ b/keras/src/backend/common/dtypes_test.py
@@ -47,10 +47,15 @@ class DtypesTest(test_case.TestCase):
 
         self.jax_enable_x64 = enable_x64()
         self.jax_enable_x64.__enter__()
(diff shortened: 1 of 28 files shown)


fchollet (approved):
Excellent work -- thanks for the PR!

PR #20971, opened 2025-02-26
Fix `Discretization` serialization when `num_bins` is used.

hertschuh:
Previously, serialization / deserialization would fail if:
- the layer was saved / restored before `adapt` was called
- the layer was saved / restored after `adapt` was called, but the dataset was such that the number of bins learned was fewer than `num_bins`

The fix consists in not serializing `bin_boundaries` if init was called with `num_bins`. This is consistent with how `tf_keras` works.

Tightened the error checking:
- never allow `num_bins` and `bin_boundaries` to be specified at the same time, even if they match (same as `tf_keras`)
- don't allow `num_bins` and `bin_boundaries` to be `None` at the same time
- verify that `adapt` has been called in `call`

Also removed `init_bin_boundaries` as the value was never used and its presence can be inferred.

### DIFF
--- a/keras/src/layers/preprocessing/discretization.py
+++ b/keras/src/layers/preprocessing/discretization.py
@@ -34,7 +34,7 @@ class Discretization(TFDataLayer):
             and `[2., +inf)`.
             If this option is set, `adapt()` should not be called.
         num_bins: The integer number of bins to compute.
-            If this option is set,
+            If this option is set, `bin_boundaries` should not be set and
             `adapt()` should be called to learn the bin boundaries.
         epsilon: Error tolerance, typically a small fraction
             close to zero (e.g. 0.01). Higher values of epsilon increase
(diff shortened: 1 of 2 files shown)


keras/src/layers/preprocessing/discretization.py
```diff
@@ -225,6 +225,13 @@ def load_own_variables(self, store):
         return
 
     def call(self, inputs):
+        if self.bin_boundaries is None:
+            raise ValueError(
+                "You need to either pass the `bin_boundaries` argument at "
+                "construction time or call `adapt(dataset)` before you can "
+                "start using the `Dicretization` layer."
```
mattdangerw, line 232: Dicretization  -> Discretization

mattdangerw (approved):
Thanks!

keras/src/layers/preprocessing/discretization.py
hertschuh: Fixed

PR #20972, opened 2025-02-27
Add `antialias` to `layers.Resizing` and add more tests.

james77777778:
Sometimes we need `antialias` to be consistent with the numerics.


fchollet (approved):
Thank you for the PR -- LGTM

PR #20983, opened 2025-03-04
Enable cuDNN RNNs when dropout is set and `training=True`

miticollo:
After [this PR](https://github.com/keras-team/keras/pull/19645) we can dropout input in cuDNN-based RNN. This PR fixes docstring to reflect the new implementation. Furthermore we don't want to apply dropout during inference like happens when `self.use_cudnn` is `False`: https://github.com/keras-team/keras/blob/c356cae26e36ecd726dc1d5d41282dccdadf5f65/keras/src/layers/rnn/lstm.py#L274-L276


fchollet (approved):
LGTM, thanks

PR #20991, opened 2025-03-06
20954.  adding a some changes line no.423

SahajDubey-2005:
made a some changes for testing purpose.

Fixed Syntax: Removed incorrect __init__ method and used proper decorator syntax.
Used wraps: Ensures the original function’s metadata is preserved.
Applied Warning in Decorator: The warning is issued when the decorator is used, not on every function call.
Ensured Proper Output Handling: The function now correctly returns only outputs.

### DIFF
--- a/keras/src/backend/numpy/core.py
+++ b/keras/src/backend/numpy/core.py
@@ -420,7 +420,7 @@ def random_seed_dtype():
     return "uint32"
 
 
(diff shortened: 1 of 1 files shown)


gbaned:
Hi @SahajDubey-2005 Can you please sign the CLA? Thank you.

PR #20993, opened 2025-03-06
Fixes inconsistent serialization logic for inputs

JyotinderSingh:
## Issue
Consider a model where its `input` property is represented as `[[[input_layer_1, 0, 0]]]`.
When serialized and deserialized, it turns into `[[input_layer_1, 0, 0]]`.

## Fix
This change removes the input_tensor un-nesting logic from `functional_from_config` method which adds special handling for the above case, now the code proceeds with the regular flow irrespective of triply nested input arrays and maintains consistent representations across serialization round-trips.

### DIFF
--- a/keras/src/models/functional_test.py
+++ b/keras/src/models/functional_test.py
@@ -5,7 +5,6 @@ import numpy as np
 import pytest
 from absl.testing import parameterized
 
-import keras
 from keras.src import applications
 from keras.src import backend
 from keras.src import layers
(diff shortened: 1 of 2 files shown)


keras/src/models/functional_test.py
```diff
@@ -5,6 +5,7 @@
 import pytest
 from absl.testing import parameterized
 
+import keras
```
hertschuh, line 8: Oh `keras` is often not imported in unit tests to not import everything and to avoid circular references.

In this case you can just do `from keras.src.models import model_from_json` and remove this.

hertschuh (changes requested):
Thanks for the bug fix!

keras/src/models/functional_test.py
JyotinderSingh: Got it, fixed

hertschuh (approved):

PR #20994, opened 2025-03-06
Fix flash attention TPU error

divyashreepathihalli:
- Flash attention was not being called on TPU because of a if condition that skipped it
- the condition has been fixed also the implementation has been updated to fix the dimension of q,k and v


hertschuh (approved):

divyashreepathihalli:
image

PR #20995, opened 2025-03-06
Wrap tf variables in keras variables for TFSMLayer

### LINKED ISSUE #20955: Error with variable.regularizer in keras.layers.TFSMLayer
I created a universal sentence encoder layer as an instance of `keras.layers.TFSMLayer` and used it in a binary classifier. Training the model results in the following error:
```
Epoch 1/10
AttributeError                            Traceback (most recent call last)
[<ipython-input-7-d7863a64cd24>](https://localhost:8080/#) in <cell line: 0>()
      2 y_train = keras.ops.array([[1], [0], [0]])
      3 
----> 4 nn_model_history = nn_model.fit(
      5   x = x_train,
      6   y = y_train,

1 frames
[/usr/local/lib/python3.11/dist-packages/keras/src/layers/layer.py](https://localhost:8080/#) in _get_regularization_losses(self)
   1152         weight_regularization_losses = []
   1153         for variable in self.trainable_weights:
-> 1154             if variable.regularizer is None:
   1155                 continue
   1156             if backend.in_stateless_scope() and not in_symbolic_scope():

AttributeError: 'UninitializedVariable' object has no attribute 'regularizer'
```

The [Keras layers code](https://github.com/keras-team/keras/blob/master/keras/src/layers/layer.py) fails when the trainable weights lack a `regularizer` attribute.

I've reproduced the error in [this Colab notebook](https://colab.research.google.com/drive/1F-RsSZNrr0bH4EUwkjkm0LE8XyVg6IMB?usp=sharing).

mattdangerw:
Fixes #20955

### DIFF
--- a/keras/src/backend/tensorflow/core.py
+++ b/keras/src/backend/tensorflow/core.py
@@ -36,13 +36,16 @@ class Variable(
         return self.value.handle
 
     def _initialize(self, value):
-        self._value = tf.Variable(
-            value,
-            dtype=self._dtype,
-            trainable=self.trainable,
-            name=self.name,
-            aggregation=self._map_aggregation(self.aggregation),
-        )
+        if isinstance(value, tf.Variable):
+            self._value = value
+        else:
+            self._value = tf.Variable(
+                value,
+                dtype=self._dtype,
+                trainable=self.trainable,
+                name=self.name,
+                aggregation=self._map_aggregation(self.aggregation),
+            )
 
     def _initialize_with_initializer(self, initializer):
         self._initialize(lambda: initializer(self._shape, dtype=self._dtype))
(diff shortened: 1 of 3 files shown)


keras/src/backend/tensorflow/core.py
```diff
@@ -124,6 +127,8 @@ def _map_aggregation(self, aggregation):
 
 
 def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
+    if isinstance(x, tf.Variable):
```
hertschuh, line 130: This shortcut needs to make sure the `dtype` matches or is `None`, otherwise it's incorrect.

I think you'll have to move this up:
```python
if dtype is not None:
        dtype = standardize_dtype(dtype)
```

and then do:
```python
if isinstance(x, tf.Variable) and (dtype is None or x.dtype == dtype):
  return x
```

hertschuh (changes requested):
Thanks for the fix!

keras/src/backend/tensorflow/core.py
hertschuh: I believe that's why the unit tests are failing.

mattdangerw: Thanks! Actually looks like we can just stick to the main path in convert_to_tensor, as `tf.is_tensor(variable) == True`. I think we were missing a conversion when comparing dtypes though, we were comparing tf dtype, to standardized string dtype in one spot.

Trying a fix.

hertschuh (approved):

PR #21000, opened 2025-03-07
Removed jax dependency from numpy image layer

shashaka:
I've replaced the JAX version of jax.image.resize with a NumPy-based implementation that closely mirrors the original behavior. Please review the updated code and let me know if any further corrections or improvements are needed.


fchollet (approved):
LGTM, it's a good move to reduce the numpy backend's reliance on JAX.

PR #21006, opened 2025-03-08
[OpenVINO Backend] Support numpy exp and expand_dims

rkazants:
**Details:** Support numpy exp and expand_dims


rkazants:
@fchollet, could you please take a look?

fchollet (approved):

PR #21008, opened 2025-03-10
[OpenVINO backend] Support numpy.full_like

jpy794:
Hi, @rkazants, could you review? This would close openvinotoolkit/openvino#29360


rkazants (commented):

PR #21010, opened 2025-03-10
[OpenVINO backend] Support numpy.empty

arkhamHack:
Hi, @rkazants, could you review this pr? This would close https://github.com/openvinotoolkit/openvino/issues/29357

### DIFF
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -362,7 +362,7 @@ def average(x, axis=None, weights=None):
         x_type = x.get_element_type()
         weights_type = weights.get_element_type()
         if (weights_type.is_integral() or weights_type == Type.boolean) and (
-            x_type.is_integral() or x_type == Type.boolean
+                x_type.is_integral() or x_type == Type.boolean
         ):
             x = ov_opset.convert(x, Type.f32).output(0)
             weights = ov_opset.convert(weights, Type.f32).output(0)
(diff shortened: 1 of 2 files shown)


arkhamHack:
Also, do let me know if any improvements needed , and what changes might be needed to increase code coverage. @rkazants 
Thanks.

keras/src/backend/openvino/numpy.py
```diff
@@ -565,7 +565,16 @@ def dot(x, y):
...
+    if dtype is not None:
+        ov_type = OPENVINO_DTYPES[standardize_dtype(dtype)]
+    else:
+        ov_type = Type.f32
+    if isinstance(shape, tuple):
+        shape = list(shape)
+    elif isinstance(shape, int):
+        shape = [shape]
+    empty_tensor = ov_opset.parameter(shape, ov_type).output(0)
+    return OpenVINOKerasTensor(empty_tensor)
```
rkazants, line 577: not correct implementation:)
Now you are adding `parameter` that is a new input to model.

Please check how it is implemented for other backends and try to interpret it.

arkhamHack: Hi, so I noticed in tensorflow backend they return a shape matrix of zeroes. Should i implement the same? The main issue here was returning a unintialized shape (with garbage values) or array like in numpy empty. Couldn't find any good opset for that in openvino.

arkhamHack: @rkazants do let me know if this is okay, i'll push the changes in that case. Thanks :)

rkazants (commented):

PR #21013, opened 2025-03-11
[OpenVINO backend] Added support for numpy.hstack

Imokut:
Please review @rkazants [#29361](https://github.com/openvinotoolkit/openvino/issues/29361)


Imokut:
``

PR #21022, opened 2025-03-13
Introduce weights sharding

james77777778:
Continuing the work based on #19286

This PR introduces `max_shard_size` in `Model.save_weights`.

Behind the scenes, this PR refactors `H5IOStore` by combining it with `H5Entry`. This change allows for more fine-grained control when storing weights. Specifically, it enables the creation of a new shard file once the current shard file reaches its capacity due to incoming weights.

Compatibility has been verified, but please let me know if anything was overlooked.
The test for LoRA weights saving/loading in KerasHub has been included.

Ping @mattdangerw and @divyashreepathihalli for requesting this feature.

A simple demo script:

```python
import os

import numpy as np

from keras import applications

# EfficientNetB0 is about 20.3MB.
model = applications.EfficientNetB0(weights=None, input_shape=(224, 224, 3))
ref_input = np.random.random((1, 224, 224, 3)).astype("float32")
ref_output = model.predict(ref_input)

# `max_shard_size` is in GB. 0.015 means about 15MB per shard.
model.save_weights("model.weights.json", max_shard_size=0.015)
files = [x for x in os.listdir(".")]
assert "model.weights.json" in files
assert "model_00000.weighs.h5" in files
assert "model_00001.weighs.h5" in files
print("Sharded weights saved successfully!")

# Load the sharded weights with the new instance.
model = applications.EfficientNetB0(weights=None, input_shape=(224, 224, 3))
model.load_weights("model.weights.json", sharded=True)
np.testing.assert_allclose(model.predict(ref_input), ref_output, atol=1e-6)
print("Passed!")

```

The outputs:

```bash
1/1 ━━━━━━━━━━━━━━━━━━━━ 1s 1s/step
Sharded weights saved successfully!
1/1 ━━━━━━━━━━━━━━━━━━━━ 1s 996ms/step
Passed!
```

The format of `.weights.json` (similar to HF's format):

```json
{
    "metadata": {
        "total_size": 476111392.0
    },
    "weight_map": {
        "/vars": "model_00000.weighs.h5",
        "/layers/input_layer/vars": "model_00000.weights.h5",
        "/layers/rescaling/vars": "model_00000.weights.h5",
        "/layers/conv2d/vars": "model_00000.weights.h5",
        "/layers/batch_normalization/vars": "model_00000.weights.h5",
...
```

### DIFF
--- a/keras/src/models/model.py
+++ b/keras/src/models/model.py
@@ -270,6 +270,16 @@ class Model(Trainer, base_trainer.Trainer, Layer):
     def save(self, filepath, overwrite=True, zipped=None, **kwargs):
         """Saves a model as a `.keras` file.
 
+        Note that `model.save()` is an alias for `keras.saving.save_model()`.
+
+        The saved `.keras` file contains:
+
+        - The model's configuration (architecture)
+        - The model's weights
+        - The model's optimizer's state (if any)
+
+        Thus models can be reinstantiated in the exact same state.
+
         Args:
             filepath: `str` or `pathlib.Path` object.
                 The path where to save the model. Must end in `.keras`
@@ -297,16 +307,6 @@ class Model(Trainer, base_trainer.Trainer, Layer):
         x = keras.random.uniform((10, 3))
         assert np.allclose(model.predict(x), loaded_model.predict(x))
         ```
-
-        Note that `model.save()` is an alias for `keras.saving.save_model()`.
-
-        The saved `.keras` file contains:
-
-        - The model's configuration (architecture)
-        - The model's weights
-        - The model's optimizer's state (if any)
-
-        Thus models can be reinstantiated in the exact same state.
         """
         return saving_api.save_model(
             self, filepath, overwrite=overwrite, zipped=zipped, **kwargs
@@ -321,6 +321,13 @@ class Model(Trainer, base_trainer.Trainer, Layer):
         will be saved in multiple files, each with a size at most
         `max_shard_size` (in GB). Additionally, a configuration file
         `.weights.json` will contain the metadata for the sharded files.
+
+        The saved sharded files contain:
+
+        - `*.weights.json`: The configuration file containing 'metadata' and
+            'weight_map'.
+        - `*_xxxxxx.weights.h5`: The sharded files containing only the
+            weights.
 
         Args:
             filepath: `str` or `pathlib.Path` object. Path where the weights
@@ -358,13 +365,6 @@ class Model(Trainer, base_trainer.Trainer, Layer):
         x = keras.random.uniform((1, 480, 480, 3))
         assert np.allclose(model.predict(x), loaded_model.predict(x))
         ```
-
-        The saved sharded files contain:
-
-        - `*.weights.json`: The configuration file containing 'metadata' and
-            'weight_map'.
-        - `*_xxxxxx.weights.h5`: The sharded files containing only the
-            weights.
         """
         return saving_api.save_weights(
             self, filepath, overwrite=overwrite, max_shard_size=max_shard_size
--- a/keras/src/models/model.py
+++ b/keras/src/models/model.py
@@ -313,47 +313,72 @@ class Model(Trainer, base_trainer.Trainer, Layer):
         )
 
     @traceback_utils.filter_traceback
-    def save_weights(self, filepath, overwrite=True):
-        """Saves all layer weights to a `.weights.h5` file.
+    def save_weights(self, filepath, overwrite=True, max_shard_size=None):
+        """Saves all weights to a single file or sharded files.
+
+        By default, the weights will be saved in a single `.weights.h5` file.
+        If sharding is enabled (`max_shard_size` is not `None`), the weights
+        will be saved in multiple files, each with a size at most
+        `max_shard_size` (in GB). Additionally, a configuration file
+        `.weights.json` will contain the metadata for the sharded files.
 
         Args:
-            filepath: `str` or `pathlib.Path` object.
-                Path where to save the model. Must end in `.weights.h5`.
-            overwrite: Whether we should overwrite any existing model
-                at the target location, or instead ask the user
-                via an interactive prompt.
+            filepath: `str` or `pathlib.Path` object. Path where the weights
+                will be saved.  When sharding, the filepath must end in
+                `.weights.json`. If `.weights.h5` is provided, it will be
+                overridden.
+            overwrite: Whether to overwrite any existing weights at the target
+                location or instead ask the user via an interactive prompt.
+            max_shard_size: `int` or `float`. Maximum size in GB for each
+                sharded file. If `None`, no sharding will be done. Defaults to
+                `None`.
         """
-        return saving_api.save_weights(self, filepath, overwrite=overwrite)
+        return saving_api.save_weights(
+            self, filepath, overwrite=overwrite, max_shard_size=max_shard_size
+        )
 
     @traceback_utils.filter_traceback
-    def load_weights(self, filepath, skip_mismatch=False, **kwargs):
-        """Load weights from a file saved via `save_weights()`.
+    def load_weights(
+        self, filepath, skip_mismatch=False, sharded=False, **kwargs
+    ):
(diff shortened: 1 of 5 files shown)


keras/src/models/model.py
```diff
@@ -313,47 +313,72 @@ def save(self, filepath, overwrite=True, zipped=None, **kwargs):
...
-                It can either be a `.weights.h5` file
-                or a legacy `.h5` weights file.
+            filepath: `str` or `pathlib.Path` object. Path where the weights
+                will be saved.  When sharding, the filepath must end in
+                `.weights.json`. If `.weights.h5` is provided, it will be
+                overridden.
             skip_mismatch: Boolean, whether to skip loading of layers where
                 there is a mismatch in the number of weights, or a mismatch in
                 the shape of the weights.
+            sharded: Whether the saved file(s) are sharded. Defaults to
```
mattdangerw, line 373: seems like on the loading side we shouldn't need this. can't we just check if we are passed a json file here and do the write thing?

keras/src/saving/saving_lib.py
```diff
@@ -945,112 +998,311 @@ def close(self):
...
         else:
-            self.group[key] = value
+            self.h5_entry_group[key] = value
 
-    def __getitem__(self, name):
-        value = self.group[name]
-        if "dtype" in value.attrs and value.attrs["dtype"] == "bfloat16":
-            value = np.array(value, dtype=ml_dtypes.bfloat16)
-        return value
+    def __delitem__(self, key):
```
mattdangerw, line 1156: Is this new? Why do we need del?

keras/src/models/model.py
```diff
@@ -313,47 +313,72 @@ def save(self, filepath, overwrite=True, zipped=None, **kwargs):
...
+            filepath: `str` or `pathlib.Path` object. Path where the weights
+                will be saved.  When sharding, the filepath must end in
+                `.weights.json`. If `.weights.h5` is provided, it will be
+                overridden.
+            overwrite: Whether to overwrite any existing weights at the target
+                location or instead ask the user via an interactive prompt.
+            max_shard_size: `int` or `float`. Maximum size in GB for each
+                sharded file. If `None`, no sharding will be done. Defaults to
+                `None`.
         """
```
mattdangerw, line 335: Might be worth adding some code examples, for those who don't want to read and just want to see how to shard :)

keras/src/saving/saving_lib.py
```diff
@@ -945,112 +998,311 @@ def close(self):
...
-            self.archive.writestr(self.root_path, self.io_file.getvalue())
+            self.archive.writestr(str(self.path_or_io), self.io_file.getvalue())
         if self.io_file:
             self.io_file.close()
 
+    # H5 entry level methods.
 
-class H5Entry:
-    """Leaf entry in a H5IOStore."""
+    @property
```
mattdangerw, line 1078: Do we need all this property getter/setter stuff? Why not just an attr?

mattdangerw (commented):
This looks great to me! @fchollet probably worth looking at this one!

My main question was on the loading side. Do we need to ask people to pass `sharded=True`, can we remove the arg or leave it none and infer by default?

keras/src/models/model.py
james77777778: Yes. I have changed the impl to use the file extension to determine whether it is a single file or sharded files.

keras/src/saving/saving_lib.py
james77777778: I just added it because it might be helpful. I'm considering making the store object more dict-like, but it's subject to change.

keras/src/models/model.py
james77777778: Sounds good!

I have added an example:

```python
# Instantiate a EfficientNetV2L model with about 454MB of weights.
model = keras.applications.EfficientNetV2L(weights=None)

# Save the weights in a single file.
model.save_weights("model.weights.h5")

# Save the weights in sharded files. Use `max_shard_size=0.25` means
# each sharded file will be at most ~250MB.
model.save_weights("model.weights.json", max_shard_size=0.25)

# Load the weights in a new model with the same architecture.
loaded_model = keras.applications.EfficientNetV2L(weights=None)
loaded_model.load_weights("model.weights.h5")
x = keras.random.uniform((1, 1024))
assert np.allclose(model.predict(x), loaded_model.predict(x))

# Load the sharded weights in a new model with the same architecture.
loaded_model = keras.applications.EfficientNetV2L(weights=None)
loaded_model.load_weights("model.weights.json")
x = keras.random.uniform((1, 1024))
assert np.allclose(model.predict(x), loaded_model.predict(x))
```

keras/src/saving/saving_lib.py
james77777778: Agreed. That was the old design before I merged `H5Entry` into `H5IOStore`. Removed.

keras/src/models/model.py
innat: Instead of 

```python
# each sharded file will be at most ~250MB
model.save_weights("model.weights.json", max_shard_size=0.25)
```

How about for better readability.


```python
# each sharded file will be at most ~250MB
model.save_weights("model.weights.json", max_shard_size='250MB')
```

james77777778: I agree that using a string has better readability but I think @fchollet wanted it to be an int.
Maybe we can support both int and string?
https://github.com/keras-team/keras/pull/19286#discussion_r1521939194

fchollet (approved):
Great work -- LGTM

keras/src/saving/saving_lib.py
mattdangerw: Got it. I'd probably ditch any code paths that we don't actively use or test, as they might easily break with some future changes. But no strong opinion

keras/src/saving/saving_lib.py
```diff
@@ -945,112 +990,308 @@ def close(self):
...
+    """Numerical variable store backed by HDF5.
+
+    Args:
+        path_or_io: `str`, `pathlib.Path` or `io.BytesIO` object. The path where
+            to save the model.
+        archive: Optional `zipfile.ZipFile` object. If specified, the h5 file
+            will be saved inside the archive and `path_or_io` will be used as
+            the filename.
+        mode: `str`. One of {'r', 'w'}. The mode to open the h5 file. Defaults
+            to `"r"`.
```
mattdangerw, line 1002: nit: changing your quote marker in the same arg string

keras/src/models/model.py
```diff
@@ -313,47 +313,111 @@ def save(self, filepath, overwrite=True, zipped=None, **kwargs):
...
+        assert np.allclose(model.predict(x), loaded_model.predict(x))
+
+        # Load the sharded weights in a new model with the same architecture.
+        loaded_model = keras.applications.EfficientNetV2L(weights=None)
+        loaded_model.load_weights("model.weights.json")
+        x = keras.random.uniform((1, 480, 480, 3))
+        assert np.allclose(model.predict(x), loaded_model.predict(x))
+        ```
+
+        The saved sharded files contain:
```
mattdangerw, line 362: We should probably keep the general flow 1) description, 2) args, 3) code example. So move this up

keras/src/models/model.py
mattdangerw: I think we can start with a int and extend later? Let's stick with what we got!

mattdangerw (approved):
Thanks! This looks great to me

keras/src/saving/saving_lib.py
james77777778: Fixed.

keras/src/models/model.py
james77777778: Sure! Updated.

james77777778:
@mattdangerw @fchollet 
I have resolved all the comments and added tests for the low-level operations of `H5IOStore` and `ShardedH5IOStore`.

keras/src/saving/saving_lib.py
james77777778: Tests have been added, which should now guarantee the code quality.

Issue #21027, opened 2025-03-13
Use JAX's AbstractMesh in distribution lib


hertschuh:
In the JAX [distribution lib](https://github.com/keras-team/keras/blob/master/keras/src/backend/jax/distribution_lib.py#L246), use [AbstractMesh](https://github.com/jax-ml/jax/blob/main/jax/_src/mesh.py#L434) instead of Mesh since it doesn't result in a JIT cache misses when the devices change. It may also simplify the distribution API.

vedantag17:
Is anyone working on this issue?

vedantag17:
Oh, I'm a newbie in Keras, but here's my understanding:
The current function builds a concrete device array. In contrast, AbstractMesh requires an immutable tuple of (axis_name, axis_size) pairs.
Is this correct?

hertschuh:
@vedantag17 , yes, that is correct.

vedantag17:
@hertschuh can you review the PR, there are some failing checks.

hertschuh:
@vedantag17 

The truth is that it is more complicated than this. At some point you need to create a concrete Mesh object to actually do the distribution. The `AbstractMesh` part should only come into play in contexts where you use the Mesh.

Also, I'm realizing we're not ready to work on this right now because:
- the JAX APIs related to `AbstractMesh` are still in flux and changing right now, so we at least need to wait until the next JAX release.
- we're not using `shard_map` in Keras right now, which is one of the main reason for preferring `AbstractMesh`.

bitloi:
Hi @hertschuh,
I want to confirm the approach before going further.
Here is my approach:
**In JIT/tracing**: Use AbstractMesh when building the layout used in `with_sharding_constraint`, so the JIT cache key is based on axis names/sizes only and doesn’t change when devices change.
**In eager**: Keep using the concrete Mesh for `device_put`, `make_array_from_process_local_data`, and any other path that does actual placement.
So: AbstractMesh only in tracing contexts; concrete Mesh everywhere we need real device placement. That matches your note that “at some point you need to create a concrete Mesh object to actually do the distribution” and that “the AbstractMesh part should only come into play in contexts where you use the Mesh.”
Does this direction look right to you, or would you prefer to wait until the JAX AbstractMesh API is more stable / you’re using `shard_map` before merging something like this?
Thanks.

bitloi:
@hertschuh Can you answer the question? Thanks for your time.

hertschuh:
@bitloi 

[...]

PR #21031, opened 2025-03-14
Fix uses `AbstractMesh` in distributed_lib

vedantag17:
Issue - #21027


hertschuh:
@vedantag17 

I'm going to close this PR as it won't work as-is. Let's follow up in https://github.com/keras-team/keras/issues/21027 once we understand better how this is supposed to work.

Thanks!

PR #21043, opened 2025-03-15
Add numpy.logaddexp to openvino backend

DeepC004:
Fixes #29489

### DIFF
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -848,9 +848,23 @@ def log2(x):
 
 
 def logaddexp(x1, x2):
-    raise NotImplementedError(
-        "`logaddexp` is not supported with openvino backend"
-    )
+    element_type = None
+    if isinstance(x1, OpenVINOKerasTensor):
+        element_type = x1.output.get_element_type()
+    if isinstance(x2, OpenVINOKerasTensor):
+        element_type = x2.output.get_element_type()
+    x1 = get_ov_output(x1, element_type)
+    x2 = get_ov_output(x2, element_type)
+    x1, x2 = _align_operand_types(x1, x2, "logaddexp()")
+    x_type = x1.get_element_type()
+    if x_type.is_integral():
+        ov_type = OPENVINO_DTYPES[config.floatx()]
+        x1 = ov_opset.convert(x1, ov_type)
+        x2 = ov_opset.convert(x2, ov_type)
+    exp_x1 = ov_opset.exp(x1).output(0)
+    exp_x2 = ov_opset.exp(x2).output(0)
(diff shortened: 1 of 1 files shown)


keras/src/backend/openvino/numpy.py
```diff
@@ -848,9 +848,23 @@ def log2(x):
...
+    x2 = get_ov_output(x2, element_type)
+    x1, x2 = _align_operand_types(x1, x2, "logaddexp()")
+    x_type = x1.get_element_type()
+    if x_type.is_integral():
+        ov_type = OPENVINO_DTYPES[config.floatx()]
+        x1 = ov_opset.convert(x1, ov_type)
+        x2 = ov_opset.convert(x2, ov_type)
+    exp_x1 = ov_opset.exp(x1).output(0)
+    exp_x2 = ov_opset.exp(x2).output(0)
+    sum_exp = ov_opset.add(exp_x1, exp_x2).output(0)
```
rkazants, line 866: I propose to make it more numerical stable for big `x1` and `x2` numbers:

ln(e^x1 + e^x2) = ln (e^k(e^(x1-k) + e^(x2-k))) = k ln (e^(x1-k) + e^(x2-k)), where k is maximum from x1 and x2. pay attention that by this trick x1-k and x2-k will be less that one that provides more compute stable for exponent that outputs non-negative value less than one that is better represented for floating point.

keras/src/backend/openvino/numpy.py
```diff
@@ -848,9 +848,23 @@ def log2(x):
```
rkazants, line 848: please switch on tests for this functionality.

keras/src/backend/openvino/numpy.py
DeepC004: Good catch on compute stability. I have made the changes

keras/src/backend/openvino/numpy.py
DeepC004: In issue you mentioned `./keras/src/ops/numpy_test.py` for testing. I initially thought that tests automatically invoke the numpy from correct backend. But on tracking `knp` imports in `numpy_test` file they seem to come from `./keras/src/backend/numpy` module instead of openvino. I verify this by executing tests without implementing `logaddexp()` and they `PASSED` (a number of other unimplemented functions inside `./keras/src/backend/openvino/numpy.py` also pass). What should I change exactly to test the function in current test setup?

rkazants: please rebase to latest master and you will see corresponding lines for `logaddexp`

hertschuh:
@rkazants 

Can you decide between this PR and https://github.com/keras-team/keras/pull/21522 ?

hertschuh:
This was implemented by https://github.com/keras-team/keras/pull/21522

PR #21044, opened 2025-03-16
[Documentation] Updated Categorical Focal Crossentropy Loss Docstring

samitshah1:
### What's the update?
- Updated example in docstring for Categorical Focal Crossentropy Loss
- Fixed typos in docstring for Binary Focal Crossentropy Loss

### Why was the update required?
- While reading the documentation for Categorical Focal Crossentropy Loss, the example imo seems to lack clarity on how the resultant loss translates to an increased focus on harder examples, compared to Categorical Crossentropy Loss.

### DIFF
--- a/keras/src/losses/losses.py
+++ b/keras/src/losses/losses.py
@@ -2143,10 +2143,22 @@ def categorical_focal_crossentropy(
 
     >>> y_true = [[0, 1, 0], [0, 0, 1]]
     >>> y_pred = [[0.05, 0.9, 0.05], [0.1, 0.85, 0.05]]
-    >>> loss = keras.losses.categorical_focal_crossentropy(y_true, y_pred)
-    >>> assert loss.shape == (2,)
-    >>> loss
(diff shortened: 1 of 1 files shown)


keras/src/losses/losses.py
```diff
@@ -2367,19 +2378,19 @@ def binary_focal_crossentropy(
     >>> # 'easier' example.
     >>> focal_loss = keras.losses.binary_focal_crossentropy(
     ...        y_true, y_pred, gamma=2)
-    >>> assert loss.shape == (2,)
+    >>> assert focal_loss.shape == (2,)
     >>> focal_loss
     array([0.330, 0.206], dtype=float32)
     >>> # Compare with binary_crossentropy
-    >>> bce_loss = keras.losses.binary_focal_crossentropy(
+    >>> bce_loss = keras.losses.binary_crossentropy(
```
hertschuh, line 2385: This is showing an example of the wrong function.

PR #21066, opened 2025-03-19
Fix warning for random_saturation

shashaka:
I have resolved the runtime warning that occurred when the factor was set to 1.0.

```
layer = keras.layers.RandomSaturation(factor=(1, 1), value_range=(0, 1))

...
RuntimeWarning: divide by zero encountered in divide
  factor = factor / (1 - factor)
```

### DIFF
--- a/keras/src/layers/preprocessing/image_preprocessing/random_saturation.py
+++ b/keras/src/layers/preprocessing/image_preprocessing/random_saturation.py
@@ -95,7 +95,7 @@ class RandomSaturation(BaseImagePreprocessingLayer):
             maxval=self.factor[1],
             seed=seed,
         )
-        factor = factor / (1 - factor)
+        factor = factor / (1 - factor + 1e-6)
         return {"factor": factor}
 
     def transform_images(self, images, transformation=None, training=True):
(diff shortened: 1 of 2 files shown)


fchollet (approved):

keras/src/layers/preprocessing/image_preprocessing/random_saturation.py
```diff
@@ -95,7 +95,7 @@ def get_random_transformation(self, data, training=True, seed=None):
             maxval=self.factor[1],
             seed=seed,
         )
-        factor = factor / (1 - factor)
+        factor = factor / (1 - factor + 1e-6)
```
james77777778, line 98: Shouldn't we use `keras.backend.epsilon` for this?
https://github.com/keras-team/keras/blob/4aae578b7f67bb653da3c268ba0634e799f47df5/keras/src/backend/config.py#L75

shashaka: I have applied epsilon on this.Thanks.

fchollet (approved):

PR #21077, opened 2025-03-20
Use tf.int32.min rather than relying on integer overflow

### LINKED ISSUE #21047: keras version of numpy.signbit results in OverflowError on Windows
Hit overflow error while training tf.keras.Sequential classifier on Windows 

Specifically: OverflowError: Python int too large to convert to C long due to 
```
tf.constant(0x80000000, dtype=tf.int32)
``` 
in Keras verison of [tensorflow.numpy.signbit line 2068](https://github.com/keras-team/keras/blob/master/keras/src/backend/tensorflow/numpy.py#L2068)

However, if I just import tensorflow and run that line in isolation, there is no Overflow error.
```
>>> import tensorflow as tf
>>> tf.constant(0x80000000, dtype=tf.int32)
<tf.Tensor: shape=(), dtype=int32, numpy=-2147483648>
```

More info on when it fails.

Though the failing line doesn't actually use `x` the argument `x` when it fails is the following:
```
type(x)
<class 'tensorflow.python.framework.ops.SymbolicTensor'>
print(x)
Tensor("Cast_2:0", shape=(None, 2), dtype=float32)
```

```
...
  File "C:\Users\chris\code\py\bugid\training\cli.py", line 70, in train
    report = classifier.train(features, labels)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\chris\code\py\bugid\training\model.py", line 85, in train
    self.model.fit(
  File "C:\Users\chris\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\local-packages\Python312\site-packages\keras\src\utils\traceback_utils.py", line 122, in error_handler
  File "C:\Users\chris\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\local-packages\Python312\site-packages\keras\src\backend\tensorflow\numpy.py", line 2081, in signbit
    tf.constant(0x80000000, dtype=tf.int32),
OverflowError: Python int too large to convert to C long
```

Project where I am seeing this https://github.com/cm6n/bugid

wilsbj:
Fixes #21047

keras == 3.9.0 + tensorflow < 2.18.0 is throwing `OverflowError: Python int too large to convert to C long` errors.

This is due to the new function `keras.src.backend.tensorflow.numpy.signbit`, where the signbit is identified by comparing the tensor to `tf.constant(0x80000000, dtype=tf.int32)`. The construction of this constant tensor fails on tensorflow < 2.18.0.

The underlying cause is that this calls `np.array(0x80000000, dtype=np.int32)` which throws the overflow error.

The intention of this constant tensor is to create a constant with binary representation `10000...` so that the signbit can be tested. Rather than relying on overflow of `tf.int32.max + 1`, this can instead be instantiated more reliably from `tf.int32.min`.

### DIFF
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -2078,7 +2078,7 @@ def signbit(x):
             tf.bitwise.bitwise_and(
                 tf.bitcast(x, tf.int32),
                 # tf.float32 sign bit
-                tf.constant(0x80000000, dtype=tf.int32),
+                tf.constant(tf.int32.min, dtype=tf.int32),
             ),
             0,
         )


fchollet (approved):
LGTM -- thanks for the fix!

PR #21086, opened 2025-03-25
[OpenVINO Backend] Support: Implemented numpy.diagonal for keras ov backend

geeky33:
- [x] 1. Implemented a diagonal function similar to diag, handling 2D+ tensors.
- [x] 2. Extracted diagonals based on the given axis1, axis2, and offset (k).
- [x] 3. Used OpenVINO ops like gather or gather_nd to fetch diagonal elements.

**Ticket Id:**

[#29116](https://github.com/openvinotoolkit/openvino/issues/29116)

 **Concerned Authority:**

CC: @rkazants

### DIFF
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -1,5 +1,5 @@
 import numpy as np
-import openvino.runtime.opset14 as ov_opset
+import openvino.opset14 as ov_opset
 from openvino import Type
 
 from keras.src.backend import config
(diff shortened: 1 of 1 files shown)


keras/src/backend/openvino/numpy.py
```diff
@@ -1,5 +1,5 @@
 import numpy as np
```
rkazants, line 1: switch on the corresponding tests

PR #21093, opened 2025-03-26
Version bump to 3.9.1

rtg0795:


hertschuh (approved):

miticollo:
Was this version intended to be published on [PyPI](https://pypi.org/project/keras/3.9.1/#files)? Because there is no [tag](https://github.com/keras-team/keras/releases/tag/v3.9.1) with change log and [`version.py`](https://github.com/keras-team/keras/blob/master/keras/src/version.py) on `master` branch was not updated.

PR #21094, opened 2025-03-26
Fix Remat error when called with a model

divyashreepathihalli:
When RematScope was applied to a model the following error was encountered
```
TypeError                                 Traceback (most recent call last)
[<ipython-input-2-38319c893c81>](https://localhost:8080/#) in <cell line: 0>()
     12     "padding_mask": np.array([[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0]]),
     13 }
---> 14 gemma_model_backbone(input_data)

1 frames
[/usr/local/lib/python3.11/dist-packages/keras/src/utils/traceback_utils.py](https://localhost:8080/#) in error_handler(*args, **kwargs)
    120             # To get the full stack trace, call:
    121             # `keras.config.disable_traceback_filtering()`
--> 122             raise e.with_traceback(filtered_tb) from None
    123         finally:
    124             del filtered_tb

[/usr/local/lib/python3.11/dist-packages/tensorflow/python/ops/custom_gradient.py](https://localhost:8080/#) in inner(*args, **kwargs)
    709     current_var_scope = variable_scope.get_variable_scope()
    710     with record.stop_recording():
--> 711       result = f(*args, **kwargs)
    712 
    713     def grad_wrapper(*wrapper_args, variables=None):

TypeError: Functional.call() missing 1 required positional argument: 'inputs'
```
This PR should fix the issue
The fix has been tested with training a Gemma backbone here 
 https://colab.sandbox.google.com/drive/16pUDE_OfUDGY9BuXwZ8lFm8dntaruraf#scrollTo=-Gg95NUB-nHw

### DIFF
--- a/keras/src/layers/layer.py
+++ b/keras/src/layers/layer.py
@@ -1053,11 +1053,13 @@ class Layer(BackendLayer, Operation, KerasSaveable):
                 if self._remat_mode is not None:
                     outputs = self.rematerialized_call(
                         self.quantized_call, *args, **kwargs
-                    )
+                    )(*args, **kwargs)
                 else:
                     outputs = self.quantized_call(*args, **kwargs)
             elif self._remat_mode is not None:
-                outputs = self.rematerialized_call(self.call, *args, **kwargs)
+                outputs = self.rematerialized_call(self.call, *args, **kwargs)(
+                    *args, **kwargs
+                )
             else:
                 outputs = self.call(*args, **kwargs)
             if return_losses:
@@ -1602,13 +1604,6 @@ class Layer(BackendLayer, Operation, KerasSaveable):
             Rematerialized layer's `call` method.
         """
 
-        def remated_function():
-            get_function_only = kwargs.pop("get_function_only", False)
-            if get_function_only:
-                return remat.remat(layer_call)
-            else:
-                return remat.remat(layer_call)(*args, **kwargs)
-
         def compute_size(x):
             return (
                 math.prod([d or 1 for d in x.shape])
@@ -1618,13 +1613,13 @@ class Layer(BackendLayer, Operation, KerasSaveable):
 
         # Full rematerialization
         if self._remat_mode.mode == "full":
-            return remated_function()
+            return remat.remat(layer_call)
 
         # Apply rematerialization to specific layers
         elif self._remat_mode.mode == "list_of_layers" and (
             self.name in self._remat_mode.layer_names
         ):
-            return remated_function()
+            return remat.remat(layer_call)
 
         # Apply rematerialization based on output size threshold
         elif self._remat_mode.mode == "larger_than":
@@ -1636,7 +1631,7 @@ class Layer(BackendLayer, Operation, KerasSaveable):
                 output_size
                 and output_size > self._remat_mode.output_size_threshold
             ):
-                return remated_function()
+                return remat.remat(layer_call)
         elif self._remat_mode.mode == "activations":
             has_activation = (
                 hasattr(self, "activation") and self.activation is not None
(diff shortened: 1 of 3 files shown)


keras/src/layers/layer.py
```diff
@@ -1602,6 +1602,13 @@ def rematerialized_call(self, layer_call, *args, **kwargs):
             Rematerialized layer's `call` method.
         """
 
+        def remated_function():
+            get_function_only = kwargs.pop("get_function_only", False)
```
hertschuh, line 1606: Hmm... I'm not a big fan of this secret argument.

Instead, let's make `rematerialized_call` always return the rematted function only and not the result (which I believe is the way you had it initially). You'll simply need to call it afterwards.

The confusing part about it is that you still need to pass the arguments, purely to compute the size in the "larger_than" mode.

So it's going to look like this:
```python
self.rematerialized_call(self.call, *args, **kwargs)(*args, **kwargs)
```
Which is weird, but I think it's more important to be consistent.

hertschuh (changes requested):
Because we turn off traceback filtering for unit tests, this was missed this entirely.

Let's add a couple of unit tests that turn on traceback filtering specifically to cover these cases (which is the normal way most people use it).

keras/src/layers/layer.py
divyashreepathihalli: Done!

keras/src/layers/layer.py
```diff
@@ -1638,11 +1640,11 @@ def compute_size(x):
                 not_rematted_activation = self.activation
                 try:
                     self.activation = remat.remat(not_rematted_activation)
-                    return layer_call(*args, **kwargs)
+                    return layer_call
```
hertschuh, line 1643: Oh... now this one won't work anymore. If you don't call here, the activation gets reverted and by the time it's called, it's not rematerialized anymore.

keras/src/layers/layer.py
```diff
@@ -1592,6 +1594,30 @@ def _open_name_scope(self):
...
+
+        Returns:
+            The result of the layer's `call` method.
+        """
+
+        original_activation = self.activation
+        rematted_activation = remat.remat(original_activation)
+        self.activation = rematted_activation
+        try:
+            return layer_call
```
hertschuh, line 1617: This still won't work because you're returning the original call and `finally` will be called before `layer_call` is called.

keras/src/layers/layer.py
```diff
@@ -1629,20 +1655,14 @@ def compute_size(x):
...
             if has_activation:
-                not_rematted_activation = self.activation
-                try:
-                    self.activation = remat.remat(not_rematted_activation)
-                    return layer_call(*args, **kwargs)
-                finally:
-                    self.activation = not_rematted_activation
-
-        return layer_call(*args, **kwargs)
+                return self.rematerialized_activation_call_wrapper(layer_call)
```
hertschuh, line 1664: You need to do declare a local wrapper like this (bound to layer_call):

```python
    @functools.wraps(layer_call)
    def rematerialized_activation_call_wrapper(*args, **kwargs):
      original_activation = self.activation
      rematted_activation = remat(original_activation)
      self.activation = rematted_activation
      res = layer_call(*args, **kwargs)
      self.activation = original_activation

    return rematerialized_activation_call_wrapper
```

hertschuh (changes requested):

keras/src/layers/layer.py
divyashreepathihalli: I forgot why we were restoring the activation function.

keras/src/layers/layer.py
divyashreepathihalli: this would return a executed layer_call to https://github.com/keras-team/keras/blob/master/keras/src/ops/operation.py#L42 - which would error out again

hertschuh: This is returning `rematerialized_activation_call_wrapper` not the result of it.

`rematerialized_activation_call_wrapper` is function.

hertschuh (approved):

PR #21095, opened 2025-03-27
Fix `Model.export` to Saved Model for models with dict inputs.

hertschuh:
Fixes https://github.com/keras-team/keras/issues/20835

Also changed multi-input tests to exercise `model.export()` and its signature inference logic.


fchollet (approved):
LGTM -- thank you!

PR #21098, opened 2025-03-27
Update training_with_built_in_methods.py

he7d3r:
Clarify parameter name

### DIFF
--- a/guides/training_with_built_in_methods.py
+++ b/guides/training_with_built_in_methods.py
@@ -620,8 +620,8 @@ val_py_dataset = ExamplePyDataset(x_val, y_val, batch_size=32)
 """
 To fit the model, pass the dataset instead as the `x` argument (no need for a `y`
 argument since the dataset includes the targets), and pass the validation dataset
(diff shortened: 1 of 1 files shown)


guides/training_with_built_in_methods.py
```diff
@@ -620,8 +620,8 @@ def __getitem__(self, idx):
 """
 To fit the model, pass the dataset instead as the `x` argument (no need for a `y`
 argument since the dataset includes the targets), and pass the validation dataset
-as the `validation_data` argument. And no need for the `batch_size` argument, since
-the dataset is already batched!
+as the `validation_data` argument. And no need for the `validation_batch_size`
```
he7d3r, line 623: Note how `validation_batch_size` is not specified (but `batch_size` is) on line 629 below:
https://github.com/keras-team/keras/blob/a41f298f1bfd1f4aba51703c5538d4e07759dabe/guides/training_with_built_in_methods.py#L629

fchollet (approved):

PR #21100, opened 2025-03-27
Resolve two bugs in keras.callbacks.ModelCheckpoint

timsweeneyfanelli:
The _save_model method combined the logic to determine if the checkpoint should be saved, and the logic to create the paths and save the checkpoint.

This PR separates the check the determine whether the checkpoint should be saved from the I/O logic, and in doing so resolves two bugs in the current implementation:

1. Host directory is created for every for save iteration, regardless of whether the model will be saved or not. For example, when `save_freq == 'epoch'` and `save_best_only == True`, a folder is created to save the model for every epoch, even though the model is only saved when the monitored condition is satisfied. The folder should only be created when the model or model weights will be saved.

2. If  `save_best_only==True`, and the monitored value is an numpy array or backend tensor, then it falls back to `save_best_only=False` behavior and saves the model. However, in this scenario, it save saves the whole model without regard to the value of `self.save_weights_only`. When `self.save_weights_only` is `True`, then only the weights should be saved.

### DIFF
--- a/keras/src/callbacks/model_checkpoint.py
+++ b/keras/src/callbacks/model_checkpoint.py
@@ -224,6 +224,70 @@ class ModelCheckpoint(Callback):
             return True
         return False
 
+    def _should_save_model(self, epoch, batch, logs, filepath):
+        """Determines whether the model should be saved.
+
+        The model should be saved in the following cases:
+
+        - self.save_best_only is False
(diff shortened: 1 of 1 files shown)


fchollet (approved):
Thanks for the fix!

Issue #21097, opened 2025-03-27
`sparse_categorical_crossentropy` ignores feature channel (and axis parameter) when checking tensor dimensions [torch backend])


enricoschroeder:
I have two tensors which I want to apply `sparse_categorical_crossentropy` to. 

I am using the torch backend, so the tensors are of shape (batch, features, x, y).

However, the function ignores the fact that the feature dimension with torch backend is not the last dimension but the first (non-batch) dimension and thus throws `ValueError: Arguments "target" and "output" must have the same shape up until the last dimension: target.shape=torch.Size([16, 1, 224, 224]), output.shape=torch.Size([16, 2, 224, 224])`.

This happens even when specifying `axis=1` (which by the way is not exposed in `keras.losses.SparseCategoricalCrossentropy` and thus not an option when using the loss with `model.fit`).

`sparse_categorical_crossentropy` should respect the feature channel setting or at least the axis parameter.

sonali-kumari1:
Hi @enricoschroeder - 
Could you please provide a code snippet to reproduce the reported behavior? Thanks!

edge7:
This should be an easy fix. The main issue is that axis is not propagated to the keras ops (which does offer that parameter in the signature)

sonali-kumari1:
Hi @enricoschroeder, @edge7 -
I tried to reproduce this issue with the latest version of keras(3.9.0) across Pytorch, Tensorflow and Jax backends and encountered the following errors:
- **Pytorch**
`ValueError: Arguments `target` and `output` must have the same shape up until the last dimension: target.shape=torch.Size([16, 1, 224, 224]), output.shape=torch.Size([16, 2, 224, 224])`
- **Jax**
`ValueError: Arguments `target` and `output` must have the same shape up until the last dimension: target.shape=(16, 1, 224, 224), output.shape=(16, 2, 224, 224)`
- **Tensorflow**
`ValueError: Argument output must have rank (ndim) target.ndim - 1. Received: target.shape=(16, 1, 224, 224), output.shape=(16, 2, 224, 224)`

I am attaching [gist](https://colab.research.google.com/gist/sonali-kumari1/2a0f6bcd3a804c94260c010e4af4df79/21097.ipynb) for your reference.

edge7:
I opened a PR, if merged one could pass the axis parameter down to the real function (Torch only).
The only thing is that the target (the ground truth) should already be squeezed

PR #21104, opened 2025-03-28
Enable SparseCategoricalCrossentropy to accept and propagate axis

edge7:


edge7:
This is mainly to accommodate issue #21097 
By the way, it does not completely fix it, as I have not modified the squeeze part, which means the target still needs to be something like (BATCH_SIZE, N, W) and not  (BATCH_SIZE, 1, N, W)

fchollet (approved):
LGTM, thanks

PR #21117, opened 2025-04-02
fixes activation layer serialization logic

JyotinderSingh:
This PR fixes a bug in `keras.activations.get` that prevented Keras layers from being correctly deserialized when their activation argument was configured using a Keras Layer instance (e.g., activation=layers.ReLU()) instead of a string identifier (e.g., activation="relu").

## Problem
Keras layers generally support specifying activation functions either via string identifiers ("relu") or by passing an actual activation Layer instance (layers.ReLU()).

While using a Layer instance works during model definition, serialization and subsequent deserialization fail.

The typical error encountered is:
```
ValueError: Could not interpret activation function identifier: {'module': 'keras.layers', 'class_name': 'ReLU', 'config': {...}, 'registered_name': None}
```

## Root Cause
1. When deserializing a layer, the `from_config` method typically retrieves the activation config (which is a dictionary in this case) and passes it to `keras.activations.get()`.
2. `keras.activations.get()` detects the dictionary input and calls `keras.activations.deserialize()`.
3. `keras.activations.deserialize()` is a wrapper around `keras.saving.deserialize_keras_object()`, passing the dictionary config along with `module_objects=ALL_OBJECTS_DICT` (where `ALL_OBJECTS_DICT` maps activation strings like "relu" to their callables).
4. Inside `keras.saving.deserialize_keras_object()`, there's a specific code path executed when `module_objects` is provided. This path checks if the input config (our dictionary) is a key within `module_objects`. Since the dictionary (config of the activation layer) is not a string key like "relu", this check fails.
5. Crucially, in this specific scenario (dictionary input + module_objects provided + key not found in module_objects), the function incorrectly returns the original dictionary config instead of proceeding to the logic that deserializes Keras objects from their dictionary representations using from_config.
6. Back in `keras.activations.get()`, the returned value is the original dictionary. The final if callable(obj): check fails because a dictionary is not callable, leading to the `ValueError`.

### DIFF
--- a/keras/src/activations/__init__.py
+++ b/keras/src/activations/__init__.py
@@ -116,7 +116,7 @@ def get(identifier):
     if identifier is None:
         return linear
     if isinstance(identifier, dict):
-        obj = deserialize(identifier)
+        obj = serialization_lib.deserialize_keras_object(identifier)
     elif isinstance(identifier, str):
         obj = ALL_OBJECTS_DICT.get(identifier, None)
     else:
--- a/keras/src/saving/serialization_lib_test.py
+++ b/keras/src/saving/serialization_lib_test.py
@@ -367,6 +367,44 @@ class SerializationLibTest(testing.TestCase):
         obj = serialization_lib.deserialize_keras_object(config)
         self.assertIs(obj, custom_registered_fn)
 
+    def test_layer_instance_as_activation(self):
+        """Tests serialization when activation is a Layer instance."""
+
+        # Dense layer with ReLU layer as activation
+        layer_dense_relu = keras.layers.Dense(
+            units=4, activation=keras.layers.ReLU(name="my_relu")
+        )
+        # Build the layer to ensure weights/state are initialized if needed
+        layer_dense_relu.build(input_shape=(None, 8))
+        _, restored_dense_relu, _ = self.roundtrip(layer_dense_relu)
+
+        # Verify the activation is correctly deserialized as a ReLU layer instance
+        self.assertIsInstance(restored_dense_relu.activation, keras.layers.ReLU)
+        # Verify properties are preserved
+        self.assertEqual(restored_dense_relu.activation.name, "my_relu")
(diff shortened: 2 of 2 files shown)


keras/src/activations/__init__.py
```diff
@@ -116,7 +116,7 @@ def get(identifier):
     if identifier is None:
         return linear
     if isinstance(identifier, dict):
-        obj = deserialize(identifier)
+        obj = serialization_lib.deserialize_keras_object(identifier)
```
hertschuh, line 119: I think you should fix `deserialize` instead. External users using `deserialize` should be able to use activation layers.

But then `deserialize` is just an alias for `deserialize_keras_object`, so that's not very useful.

JyotinderSingh: Maybe we can just remove `module_objects=ALL_OBJECTS_DICT` argument when making this call from within the `deserialize` method?
I'm concerned that might cause some unintended behavior change for clients depending on that API.

Maybe we can swap out the `deserialize(identifier)` call for the `serialization_lib.deserialize_keras_object(identifier)` call here, and then later investigate/refactor the `deserialize` method separately.

hertschuh: > Maybe we can just remove module_objects=ALL_OBJECTS_DICT argument when making this call from within the deserialize method?

Yes, exactly. That's what I meant by "fixing deserialize".

> I'm concerned that might cause some unintended behavior change for clients depending on that API.

All that `module_objects=ALL_OBJECTS_DICT` does is restrict the possibilities to the list of predefined activation **functions**. So it's not going to break existing use cases. But it will make some use cases that didn't work before work. I don't think anybody relies on the error happening for non-supported use case.

The only concern if we fix `deserialize` is that there no guarantee that the returned object is actually an activation function or layer. It could be anything. That's a bit weird.

hertschuh (approved):

PR #21127, opened 2025-04-03
Add OpenVINO backend support for numpy.expm1

DakshBegani:
This pull request adds OpenVINO backend support for numpy.expm1. The expm1 function is implemented using OpenVINO operations to perform element-wise exp(x) - 1.

Changes include:
- Added expm1 implementation in `numpy.py` for OpenVINO backend.
- Updated relevant files to ensure compatibility with the new operation.
- Verified correctness by running relevant tests.

### DIFF
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -1379,3 +1379,13 @@ def argpartition(x, kth, axis=-1):
     raise NotImplementedError(
         "`argpartition` is not supported with openvino backend"
     )
+
+@numpy_support
+def expm1(x):
+    """Calculates exp(x) - 1 element-wise.
+
+    >>> x = np.array([0.0, 1.0])
+    >>> print(expm1(x))
+    [0.        1.7182817]
+    """
+    return ov.exp(x) - ov.constant(1, dtype=x.dtype)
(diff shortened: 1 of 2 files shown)


keras/src/backend/openvino/numpy.py
```diff
@@ -1379,3 +1379,12 @@ def argpartition(x, kth, axis=-1):
     raise NotImplementedError(
         "`argpartition` is not supported with openvino backend"
     )
+
+
+def expm1(x):
```
rkazants, line 1384: there exists `expm1` declaration for which you need provide implementation. No need to create a new declaration.
Also, please activate test for this op

DakshBegani: implemented to its previous declaration, what exactly do you mean to activate test for this op?

keras/src/backend/openvino/numpy.py
```diff
@@ -1378,4 +1383,4 @@ def slogdet(x):
 def argpartition(x, kth, axis=-1):
     raise NotImplementedError(
         "`argpartition` is not supported with openvino backend"
-    )
+    )
```
rkazants, line 1386: revert this change

keras/src/backend/openvino/numpy.py
rkazants: check other PRs how they remove

rkazants:
Please close this PR. It was already completed: https://github.com/keras-team/keras/pull/21141

PR #21142, opened 2025-04-07
Add alpha argument description to elu docstring

shashaka:
I updated the docstring for the elu activation function to include a description of the alpha argument.


fchollet (approved):
LGTM, thanks!

Issue #21126, opened 2025-04-03
`LayerTest.test_functional_model_with_remat` fails with `NameError: name 'MockRemat' is not defined`


GaetanLepage:
Shouldn't have https://github.com/keras-team/keras/pull/21094 also updated the test case?
It now fails with:
```
=================================== FAILURES ===================================
__________________ LayerTest.test_functional_model_with_remat __________________

self = <keras.src.layers.layer_test.LayerTest testMethod=test_functional_model_with_remat>

    def test_functional_model_with_remat(self):
        if backend.backend() in ("openvino", "numpy"):
            self.skipTest(
                "remat is not supported in openvino and numpy backends."
            )
        traceback_utils.enable_traceback_filtering()
>       mock_remat = MockRemat()
E       NameError: name 'MockRemat' is not defined

keras/src/layers/layer_test.py:224: NameError
=========================== short test summary info ============================
FAILED keras/src/layers/layer_test.py::LayerTest::test_functional_model_with_remat - NameError: name 'MockRemat' is not defined
= 1 failed, 12710 passed, 280 skipped, 2 deselected, 4 xfailed, 6 xpassed in 574.57s (0:09:34) =
```

Context: [bumping `keras` in nixpkgs](https://github.com/NixOS/nixpkgs/pull/395604)

hertschuh:
@GaetanLepage you're right, the 3.9.2 branch is missing it.

hertschuh:
Fix is in.

PR #21157, opened 2025-04-11
Add verification to remat tests. (#21102)

hertschuh:
The functions that go through `remat()` should actually be called, if not, remat is not really applied.


hertschuh:
Fixes https://github.com/keras-team/keras/issues/21126

SamanehSaadat (approved):

PR #21168, opened 2025-04-15
[OpenVINO Backend] Support numpy min operation

srinjoydutta03:
@rkazants 

Closes: [https://github.com/openvinotoolkit/openvino/issues/30117](https://github.com/openvinotoolkit/openvino/issues/30117)

### DIFF
--- a/keras/src/backend/openvino/excluded_concrete_tests.txt
+++ b/keras/src/backend/openvino/excluded_concrete_tests.txt
@@ -35,7 +35,6 @@ NumpyDtypeTest::test_max
 NumpyDtypeTest::test_mean
 NumpyDtypeTest::test_median
 NumpyDtypeTest::test_meshgrid
-NumpyDtypeTest::test_min
 NumpyDtypeTest::test_moveaxis
 NumpyDtypeTest::test_multiply
 NumpyDtypeTest::test_nan
@@ -95,7 +94,6 @@ NumpyOneInputOpsCorrectnessTest::test_max
 NumpyOneInputOpsCorrectnessTest::test_mean
 NumpyOneInputOpsCorrectnessTest::test_median
 NumpyOneInputOpsCorrectnessTest::test_meshgrid
-NumpyOneInputOpsCorrectnessTest::test_min
 NumpyOneInputOpsCorrectnessTest::test_moveaxis
 NumpyOneInputOpsCorrectnessTest::test_nan_to_num
 NumpyOneInputOpsCorrectnessTest::test_ndim
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -1006,7 +1006,83 @@ def meshgrid(*x, indexing="xy"):
 
 
 def min(x, axis=None, keepdims=False, initial=None):
-    raise NotImplementedError("`min` is not supported with openvino backend")
+    x = get_ov_output(x)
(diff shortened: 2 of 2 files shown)


keras/src/backend/openvino/excluded_concrete_tests.txt
```diff
@@ -35,7 +35,6 @@ NumpyDtypeTest::test_max
 NumpyDtypeTest::test_mean
 NumpyDtypeTest::test_median
 NumpyDtypeTest::test_meshgrid
-NumpyDtypeTest::test_min
 NumpyDtypeTest::test_moveaxis
```
rkazants, line 38: Please add line with `NumpyDtypeTest::test_minimum_python_types`

srinjoydutta03: I have added it. Thank you!

keras/src/backend/openvino/numpy.py
```diff
@@ -1006,7 +1006,100 @@ def meshgrid(*x, indexing="xy"):
...
+    is_bool = x_type == Type.boolean
+    if is_bool:
+        x = ov_opset.convert(x, Type.i32).output(0)
+        x_type = Type.i32
+
+    is_empty = False
+    for dim in x_shape:
+        if dim == 0:
+            is_empty = True
+            break
```
rkazants, line 1023: let us skip `is_empty` case at all because this is invalid case for which it throws exception.
We assume that `x` is not empty always.

rkazants: please remove `is_empty` case related code

keras/src/backend/openvino/numpy.py
```diff
@@ -1006,7 +1006,100 @@ def meshgrid(*x, indexing="xy"):
...
+                ).output(0)
+            else:
+                result = init_tensor
+
+        if is_bool:
+            result = ov_opset.convert(result, Type.boolean).output(0)
+
+        return OpenVINOKerasTensor(result)
+
+    if axis is None and initial is not None:
```
rkazants, line 1060: let us handle initial input at the final step. It is the same for all cases.

keras/src/backend/openvino/numpy.py
```diff
@@ -1006,7 +1006,100 @@ def meshgrid(*x, indexing="xy"):
...
+        if keepdims:
+            result_shape = [1] * len(x_shape)
+            result = ov_opset.reshape(
+                result,
+                ov_opset.constant(result_shape, Type.i32).output(0),
+                False,
+            ).output(0)
+
+        if is_bool:
+            result = ov_opset.convert(result, Type.boolean).output(0)
```
rkazants, line 1077: let us handle it finally. It same for all cases

keras/src/backend/openvino/numpy.py
srinjoydutta03: I have removed the `is_empty` case. Thanks

keras/src/backend/openvino/numpy.py
srinjoydutta03: Handled at the end. Thanks

keras/src/backend/openvino/numpy.py
srinjoydutta03: Handled finally. Thanks

srinjoydutta03:
@rkazants Could you please review the changes?

PR #21170, opened 2025-04-15
Improve `tf.RaggedTensor` support in `DataAdapter`s.

hertschuh:
Previously, only 2D Tensorflow ragged tensors were supported. This adds support for any rank.

Also added tests for ragged tensors with `GeneratorDataAdapter`.


SamanehSaadat (approved):
Thanks, Fabien!

fchollet (approved):

PR #21171, opened 2025-04-16
Add `tf.RaggedTensor` support to `Embedding` layer.

hertschuh:
Adds support for indices indices in the form of a `tf.RaggedTensor` to the `Embedding` layer by adding support to `ops.take`. The output is also ragged.

Also:
- adds support for negative indices in the sparse tensor use case.
- adds support for ragged tensors in `TestCase.run_layer_test`.


fchollet (approved):
LGTM, thank you for the PR!

PR #21175, opened 2025-04-16
Add sparse_sigmoid activation

shashaka:
Since Keras lacks built-in support for the sparse_sigmoid activation function, I implemented it across TensorFlow, PyTorch, and NumPy, aiming to maintain consistency with existing implementations available in Jax.


fchollet (approved):
LGTM, thanks for the addition.

PR #21181, opened 2025-04-17
Add sparse support to `ops.ones_like` and `ops.zeros_like`.

hertschuh:
`ops.zeros_like` is in particular useful for creating a mask of the populated values in the sparse tensor.


fchollet (approved):
Thanks for the PR!

PR #21184, opened 2025-04-18
Fix dtype detection for JAX types.

cantonios:
The jax types like `jax.float32` have a string representation of
```
<class 'jax.numpy.float32'>
```
so with the previous code, would be "standardized" as `float32'>` (trailing quote and angle bracket), which is an invalid type.  But, the JAX dtypes _do_ have a `__name__` property, so should be properly detected if we switch the order around.

Kept the old `jax.numpy` string version in place in case that worked with older versions of JAX.


cantonios:
@hertschuh

hertschuh (approved):

PR #21185, opened 2025-04-18
WIP: fix RandAugment behavior in tf graph mode

gregstarr:
fix for #21169

`RandAugment` changes:
* samples random execution order
* samples transform params for all sub layers
* uses `fori` and `switch` to execute the sub layers according to the sampled order

this is passing the tests in all three backends. However for the `test_rand_augment_model` test, jax runs very slowly, way slower than the other backends, by a factor of 10 - 50. I have no experience with jax, can you suggest anything that might help?

thanks

### DIFF
--- a/keras/src/layers/preprocessing/image_preprocessing/rand_augment.py
+++ b/keras/src/layers/preprocessing/image_preprocessing/rand_augment.py
@@ -1,11 +1,10 @@
-import random
-
 import keras.src.layers as layers
+import keras.src.ops as ops
 from keras.src.api_export import keras_export
 from keras.src.layers.preprocessing.image_preprocessing.base_image_preprocessing_layer import (  # noqa: E501
     BaseImagePreprocessingLayer,
 )
-from keras.src.random import SeedGenerator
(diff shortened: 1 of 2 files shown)


shashaka:
In my opinion, some exception handling should be added for the NumPy backend in your test case, particularly due to the use of the fit method.

gregstarr:
> In my opinion, some exception handling should be added for the NumPy backend in your test case, particularly due to the use of the fit method.

yes I agree, the test should be skipped for numpy backend.

Any ideas about the layer implementation? it seems to have problems for jax backend

shashaka:
I’m not entirely sure about that. I think a step-by-step breakdown is needed to identify what’s causing the slowdown in the JAX backend.

gregstarr:
> I’m not entirely sure about that. I think a step-by-step breakdown is needed to identify what’s causing the slowdown in the JAX backend.

Ok I will take a look when I get a chance.

hertschuh:
This was addressed in https://github.com/keras-team/keras/pull/21499

PR #21186, opened 2025-04-18
[OpenVINO Backend] : add support for numpy.nan_to_num

11happy:
**Overview:**
- fixes https://github.com/openvinotoolkit/openvino/issues/30119

**Testing:**
- Currently passsing all tests with this approach but having issues with isinf approach . I have mentioned in detail in comments below.

**CC:**
- @rkazants


11happy:
current approach is passing all tests but when I use `is_inf` to create masks for neg_inf & pos_inf somehow i think it is not able to create those masks properly any insights on that is appreciated.

previous approach to create masks:
```
posinf_mask = ov_opset.is_inf(x,{'detect_negative': False,'detect_positive': True}).output(0)
neginf_mask = ov_opset.is_inf(x,{'detect_negative': True,'detect_positive': False}).output(0)
```
error faced:
Screenshot from 2025-04-18 23-22-34

Thank you

11happy:
as numpy documentation mentions to use dtype max I create new dictionaries is there any other approach to directly get dtype maxes & mins please let me know.
Thank you
image

keras/src/backend/openvino/excluded_concrete_tests.txt
```diff
@@ -96,7 +96,6 @@ NumpyOneInputOpsCorrectnessTest::test_median
 NumpyOneInputOpsCorrectnessTest::test_meshgrid
```
rkazants, line 96: please also remove `NumpyDtypeTest::test_nan`

keras/src/backend/openvino/numpy.py
```diff
@@ -1030,9 +1032,29 @@ def moveaxis(x, source, destination):
...
 def nan_to_num(x, nan=0.0, posinf=None, neginf=None):
-    raise NotImplementedError(
-        "`nan_to_num` is not supported with openvino backend"
-    )
+    x = get_ov_output(x)
+    dtype = x.get_element_type()
+    shape_x = ov_opset.shape_of(x).output(0)
+    nan_vector = ov_opset.broadcast(
+        ov_opset.constant(nan, dtype), shape_x
+    ).output(0)
```
rkazants, line 1040: it can be just a scalar. select operation will broadcast it automatically. No need broadcast by yourself.

rkazants: the same comment for `pos_inf` and `neg_inf`:)

rkazants:
> current approach is passing all tests but when I use `is_inf` to create masks for neg_inf & pos_inf somehow i think it is not able to create those masks properly any insights on that is appreciated.
> 
> previous approach to create masks:
> 
> ```
> posinf_mask = ov_opset.is_inf(x,{'detect_negative': False,'detect_positive': True}).output(0)
> neginf_mask = ov_opset.is_inf(x,{'detect_negative': True,'detect_positive': False}).output(0)
> ```
> 
> error faced(infs are not detected here I checked them ): Screenshot from 2025-04-18 23-22-34
> 
> Thank you

Can you please create standalone code that demonstrates this issue and create issue in openvino github? It will be greatly to fix it on our side.

Best regards,
Roman

11happy:
> > current approach is passing all tests but when I use `is_inf` to create masks for neg_inf & pos_inf somehow i think it is not able to create those masks properly any insights on that is appreciated.
> > previous approach to create masks:
> > ```
> > posinf_mask = ov_opset.is_inf(x,{'detect_negative': False,'detect_positive': True}).output(0)
> > neginf_mask = ov_opset.is_inf(x,{'detect_negative': True,'detect_positive': False}).output(0)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > error faced(infs are not detected here I checked them ): Screenshot from 2025-04-18 23-22-34
> > Thank you
> 
> Can you please create standalone code that demonstrates this issue and create issue in openvino github? It will be greatly to fix it on our side.
> 
> Best regards, Roman

Sure I will create a new issue there , also I have correctec the code based on new tests & previous reviews.
Thank you

keras/src/backend/openvino/numpy.py
11happy: done

keras/src/backend/openvino/excluded_concrete_tests.txt
11happy: done

11happy:
it is interesting behaviour i dont know, when I tried to reproduce the same error in a colab
image
now the is_inf are detecting it properly  however still the error persists when I try to do it in numpy.nan_to_num implementation .. not sure why but am debugging it , any insights on this is appreciated.
Thank you

rkazants:
> when I tried to reproduce the same error in a colab image now the is_inf are detecting it properly however still the error persists when I try to do it in numpy.nan_to_num implementation .. not sure why, but am debugging it , any insights on this is appreciated. Thank you

try to create constant node instead of parameter, i.e. no params and one result node.

11happy:
in is_inf we need to I dont know why have to send the input by  convert it to float inputss then only it works.

getting to this conclusion was very interesting , I first tried running the tests for is_inf by excluding it & saw they were also failing 
```
def isinf(x):
    x = get_ov_output(x)
    return OpenVINOKerasTensor(ov_opset.is_inf(x).output(0)) 
```
correct usage would be  & this change passes all tests same thing goes for is_finite I cheecked that as welll.
```
def isinf(x):
    x = get_ov_output(x)
    return OpenVINOKerasTensor(ov_opset.is_inf(ov_opset.convert(x, Type.f32)).output(0)) 
```
also on colab it was always correctly workin also with constant node & single result.
is this intentional to pass them as float ?? 
Thank you

rkazants (commented):

keras/src/backend/openvino/numpy.py
```diff
@@ -1030,9 +1032,28 @@ def moveaxis(x, source, destination):
...
+    dtype = x.get_element_type()
+    nan_val = ov_opset.constant(nan, dtype).output(0)
+    posinf_val = ov_opset.constant(
+        posinf if posinf is not None else DTYPES_MAX[dtype], dtype
+    ).output(0)
+    neginf_val = ov_opset.constant(
+        neginf if neginf is not None else DTYPES_MIN[dtype], dtype
+    ).output(0)
+    posinf_mask = ov_opset.is_inf(
+        ov_opset.convert(x, Type.f32),
```
rkazants, line 1045: probably it is a bug on OpenVINO side. Please create issue for non-f32 types in OpenVINO repo. For now, this WA absolutely works for us

11happy: Done , I have created the issue.
Thank you

rkazants:

keras/src/backend/openvino/numpy.py
```diff
@@ -1030,9 +1032,28 @@ def moveaxis(x, source, destination):
 
 
 def nan_to_num(x, nan=0.0, posinf=None, neginf=None):
-    raise NotImplementedError(
-        "`nan_to_num` is not supported with openvino backend"
-    )
+    x = get_ov_output(x)
+    dtype = x.get_element_type()
+    nan_val = ov_opset.constant(nan, dtype).output(0)
+    posinf_val = ov_opset.constant(
```
rkazants, line 1038: @11happy, let us do not call `is_inf` and `is_nan` for integer types. You can return input as is as I understand.
For floating point types you need to call but convert to fp32 is not needed.
Can you fix it please in this PR?

rkazants: @11happy, please clarify what cases exactly are failing. For int types, it should pass because there is no inf and nan values.

11happy: so the test case that is failing is previous one only
```
>       self.assertAllClose(
            knp.nan_to_num(x, nan=2, posinf=3, neginf=4), [1.0, 2.0, 3.0, 4.0]
        )

keras/src/ops/numpy_test.py:4806: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
keras/src/testing/test_case.py:48: in assertAllClose
    np.testing.assert_allclose(x1, x2, atol=atol, rtol=rtol, err_msg=msg)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

args = (<function assert_allclose.<locals>.compare at 0x722684c8ed40>, array([ 1.00000000e+00,  2.00000000e+00,  3.40282347e+38, -3.40282347e+38]), array([1., 2., 3., 4.]))
kwds = {'equal_nan': True, 'err_msg': 'None', 'header': 'Not equal to tolerance rtol=1e-06, atol=1e-06', 'verbose': True}

    @wraps(func)
    def inner(*args, **kwds):
        with self._recreate_cm():
>           return func(*args, **kwds)
E           AssertionError: 
E           Not equal to tolerance rtol=1e-06, atol=1e-06
E           None
E           Mismatched elements: 2 / 4 (50%)
E           Max absolute difference: 3.40282347e+38
E           Max relative difference: 1.13427449e+38
E            x: array([ 1.000000e+00,  2.000000e+00,  3.402823e+38, -3.402823e+38])
E            y: array([1., 2., 3., 4.])

/usr/lib/python3.12/contextlib.py:81: AssertionError
```
I tried to debug it , my observations:
this is failing for f64 dtype, earlier I checked is_inf on cloab with f64 its working fine but here the test case with dtype f64 is failing , also in conversion logic if I intentionally convert it to f64 type instead of f32 the above test fails. here specifically it passes with f16,f32, but fails with f64.

I am also attaching a ss of PDB :
Screenshot from 2025-04-25 15-36-00

11happy: I have now a reproducible code on f64 dtype for the same erroe

rkazants: Well, can you please do the implementation as follows:
1. return input as is for integer types
2. use `is_inf` and `is_nan` mask for floating point types. In addition, use WA with convert for fp64, for other types it is not needed.
Best regards,
Roman

11happy: done

11happy: I have also added a comment referencing the issue

fchollet (approved):
Thanks!

PR #21187, opened 2025-04-18
Redundant api imports; no path hacking in package

### LINKED ISSUE #19779: Keras `__init__.py` structure isn't readable by a static type checker

mattdangerw:
This makes two changes to our package structure.

1. Export all API with *redundant imports*. E.g. `from keras.src.layers import Dense as Dense`. This is a hint to type checkers that this is an exported symbol, not a module that is just used directly by code in the init file. https://typing.python.org/en/latest/spec/distributing.html#library-interface-public-and-private-symbols
2. Remaps the `api/` directory to the toplevel directory in the package, so no more `__path__` hacking for our installed package.

```python
keras
└── __init__.py
├── activations
│   └── __init__.py
....  # Public init files with redundant imports.
├── src
....  # Private stuff subject to changes at any release.
```

For the latter, the main change is setting `package-dirs` in the pyproject.toml. Which lets us remap the github structure to package structure without needing our own custom build process.

I pushed a package to a [keras-test](https://pypi.org/project/keras-test/) name on pypi. Try any of these (probably all good to check):
```
pip install keras-test
pip install git+https://github.com/mattdangerw/keras@fix-package
git clone https://github.com/mattdangerw/keras -b fix-package && pip install -e keras
```

Fixes #19779


pyproject.toml
```diff
@@ -44,8 +44,10 @@ Repository = "https://github.com/keras-team/keras"
 [tool.setuptools.dynamic]
 version = {attr = "keras.src.version.__version__"}
 
-[tool.setuptools.packages.find]
-include = ["keras", "keras.*"]
+[tool.setuptools.package-dir]
+"" = "."
+"keras" = "keras/api"
+"keras.src" = "keras/src"
```
mattdangerw, line 50: We might be able to remove this line, I'll check.

mattdangerw:
This should pass tests now with the new namex.

fchollet (approved):

PR #21189, opened 2025-04-20
Support numpy.moveaxis operation

Prabha14039:
### Details

- fixes openvinotoolkit/openvino#30118

cc , @rkazants plz review

### DIFF
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -1042,7 +1042,7 @@ def moveaxis(x, source, destination):
         axes.pop(axis)
 
     for dest, src in sorted(zip(destination, source)):
-        axes[dest] = src
+        axes.insert(dest, src)
 
     perm = ov_opset.constant(axes, dtype=Type.i32).output(0)
 
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -1042,7 +1042,7 @@ def moveaxis(x, source, destination):
         axes.pop(axis)
 
     for dest, src in sorted(zip(destination, source)):
-        axes.insert(dest, src)
+        axes[dest] = src
 
     perm = ov_opset.constant(axes, dtype=Type.i32).output(0)
 
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -1024,9 +1024,37 @@ def mod(x1, x2):
 
 
 def moveaxis(x, source, destination):
-    raise NotImplementedError(
-        "`moveaxis` is not supported with openvino backend"
-    )
+    if isinstance(source, int):
+        source = [source]
+    if isinstance(destination, int):
+        destination = [destination]
+
+    x = get_ov_output(x)
+    x_shape = x.get_partial_shape()
+    rank = x_shape.rank.get_length()
+
+    source = [s + rank if s < 0 else s for s in source]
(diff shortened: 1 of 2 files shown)


keras/src/backend/openvino/numpy.py
```diff
@@ -1024,9 +1024,37 @@ def mod(x1, x2):
...
+    source = [s + rank if s < 0 else s for s in source]
+    destination = [d + rank if d < 0 else d for d in destination]
+
+    axes = list(range(rank))
+
+    for axis in sorted(source, reverse=True):
+        axes.pop(axis)
+
+    for dest, src in sorted(zip(destination, source)):
+        axes.insert(dest, src)
```
rkazants, line 1045: I think it is more clear this code as:

```
for dest, src in sorted(zip(destination, source)):
    axes[dest] = src
```

Prabha14039: Done

Prabha14039: i dont think we can do this because this results in broken permuation 
as this assumes axes already has the correct length

rkazants: Can you share a concrete example that shows it will not work?

Prabha14039: image

This is happening because the following code assumes that the index is of equal length 
but after popping the elements the length is not the same hence it results in index error if `dest >= len(axes)`

```
def moveaxis(x, source, destination):

    if isinstance(source, int):
        source = [source]
    if isinstance(destination, int):
        destination = [destination]

    x = get_ov_output(x)
    x_shape = x.get_partial_shape()
    rank = x_shape.rank.get_length()

    source = [s + rank if s < 0 else s for s in source]
    destination = [d + rank if d < 0 else d for d in destination]

    axes = list(range(rank))

    for axis in sorted(source, reverse=True):
        axes.pop(axis)

    for dest, src in sorted(zip(destination, source)):
        axes[dest] = src

    perm = ov_opset.constant(axes, dtype=Type.i32).output(0)

    x_type = x.get_element_type()
    if x_type == Type.bf16:
        x = ov_opset.convert(x, Type.f32).output(0)
        result = ov_opset.transpose(x, perm).output(0)
        return OpenVINOKerasTensor(
            ov_opset.convert(result, Type.bf16).output(0)
        )

    return OpenVINOKerasTensor(ov_opset.transpose(x, perm).output(0))
```

rkazants: @Prabha14039, once again. Share with me a concrete example with concrete values source and destination for which my code willn't work.

Prabha14039: i apologise here is the test case failing 
image

rkazants: you need to remove as well:
{BB02C2C5-31A5-4E32-8BED-46BC31665DA8}

Prabha14039: image

i am facing the following issue after removing the axes.pop

rkazants: Here I don't see source and destination values. Please create standlalone code that proves my code does not work.

rkazants: @Prabha14039, any update on my latest comment? Thanks!

fchollet:

Please resolve merge conflicts.

hertschuh:
This was addressed by https://github.com/keras-team/keras/pull/21269

PR #21191, opened 2025-04-21
OpenVINO: linspace and logspace

vi-shruti:
@rkazants for review of linspace and logspace Implementations for OpenVINO

### DIFF
--- a/keras/src/backend/openvino/excluded_concrete_tests.txt
+++ b/keras/src/backend/openvino/excluded_concrete_tests.txt
@@ -34,7 +34,7 @@ NumpyDtypeTest::test_max
 NumpyDtypeTest::test_mean
 NumpyDtypeTest::test_median
 NumpyDtypeTest::test_meshgrid
-NumpyDtypeTest::test_min
+NumpyDtypeTest::test_minimum_python_types
 NumpyDtypeTest::test_moveaxis
 NumpyDtypeTest::test_multiply
 NumpyDtypeTest::test_nan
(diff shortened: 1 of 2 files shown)


keras/src/backend/openvino/excluded_concrete_tests.txt
```diff
@@ -26,10 +26,8 @@ NumpyDtypeTest::test_inner
 NumpyDtypeTest::test_isfinite
 NumpyDtypeTest::test_isinf
 NumpyDtypeTest::test_isnan
-NumpyDtypeTest::test_linspace
 NumpyDtypeTest::test_log1p
 NumpyDtypeTest::test_logaddexp
-NumpyDtypeTest::test_logspace
 NumpyDtypeTest::test_matmul_
```
rkazants, line 31: please also remove lines to enable additional tests:

```
NumpyTwoInputOpsCorrectnessTest::test_linspace
NumpyTwoInputOpsCorrectnessTest::test_logspace
```

pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
```
rkazants, line 1: please remove this file. It is only needed for developer purpose

vi-shruti:
A quick note on pytest, I was able to run the tests using:
pytest ./keras/src/ops/numpy_test.py

However, https://github.com/openvinotoolkit/openvino/issues/30114 mentions the usage of:
pytest-c ./pytest.ini ./keras/src/ops/numpy_test.py

Is that something that might need an update?

PR #21193, opened 2025-04-21
 Don't scale gradients if overwriting variable with gradient.

cantonios:
If `variable.overwrite_with_gradient`, the gradient represents the desired final value of the variable.  If we did scale it, we're changing that final value.

### DIFF
--- a/keras/src/optimizers/base_optimizer.py
+++ b/keras/src/optimizers/base_optimizer.py
@@ -424,6 +424,11 @@ class BaseOptimizer(KerasSaveable):
             self._check_variables_are_known(trainable_variables)
 
         with backend.name_scope(self.name, caller=self):
+            # Filter empty gradients.
+            grads, trainable_variables = self._filter_empty_gradients(
+                grads, trainable_variables
+            )
+
             # Overwrite targeted variables directly with their gradients if
             # their `overwrite_with_gradient` is set.
             grads, trainable_variables = (
@@ -432,10 +437,6 @@ class BaseOptimizer(KerasSaveable):
                 )
             )
 
-            # Filter empty gradients.
-            grads, trainable_variables = self._filter_empty_gradients(
-                grads, trainable_variables
-            )
             if len(list(grads)) == 0:
                 return
 
--- a/keras/src/optimizers/loss_scale_optimizer.py
+++ b/keras/src/optimizers/loss_scale_optimizer.py
@@ -102,6 +102,12 @@ class LossScaleOptimizer(optimizer.Optimizer):
             ),
         )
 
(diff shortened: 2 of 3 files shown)


cantonios:
@hertschuh found this when looking to use this property for the large-embedding case.

keras/src/optimizers/loss_scale_optimizer.py
```diff
@@ -171,8 +174,12 @@ def apply(self, grads, trainable_variables=None):
     def _stateful_handle_finite_grads(self, grads, trainable_variables):
         scale = self.dynamic_scale
         # Unscale gradients.
+        tvs = trainable_variables or self._trainable_variables
```
hertschuh, line 177: You shouldn't need this. Line 185 it uses `trainable_variables` only.

keras/src/optimizers/base_optimizer.py
```diff
@@ -432,17 +443,18 @@ def apply(self, grads, trainable_variables=None):
...
-                grads, trainable_variables
-            )
             if len(list(grads)) == 0:
                 return
 
             # Unscale gradients.
             scale = self.loss_scale_factor
             if scale is not None:
-                grads = [g if g is None else g / scale for g in grads]
+                grads = [
```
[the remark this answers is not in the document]
cantonios: Ah, good catch.  Removed.

keras/src/optimizers/loss_scale_optimizer.py
cantonios: Can't it be `None` though?  It's called on line 211 with a default of `None`, which can be called from line 172, which also uses a default of `None`.

We need the actual `trainable_variables` to check the `overwrite_with_gradient` property.

cantonios: Ah yes, the line numbers changed when I uploaded a new commit.
```
apply(self, grads, trainable_variables=None)
```
calls
```
_common_apply(self, grads, trainable_variables=None)
```
which calls this function,
```
_stateful_handle_grads(self, grads, trainable_variables)
```
So if the `apply(...)` method is called with `trainable_variables=None` (or left absent to use the default `None`, then we end up in `_stateful_handle_grads(self, grads, None)`.  This is why I need to set:
```
tv = trainable_variables or self._trainable_variables.
```
if `trainable_variables` is `None` (because no trainable variables were passed to `apply`), then the variables are `self.trainable_variables`.

hertschuh: Ok, I did not realize that it is optional in `apply`.

hertschuh (approved):

PR #21194, opened 2025-04-21
Return explicitly layout if already set on variable.

cantonios:
If explicitly overwriting a variable._layout, we want to keep this layout in any future calls.  This allows auxiliary variables (e.g. gradients, optimizer momentums) to use the same explicit layout.


cantonios:
@hertschuh

fchollet (approved):
Sounds good

PR #21195, opened 2025-04-21
Allow `Embedding` subclasses to only override `compute_output_shape`.

hertschuh:
Without the need to also override `compute_output_spec`.


SamanehSaadat (approved):
Thanks, Fabien!

PR #21202, opened 2025-04-22
`mlx` - new image ops implemented

acsweet:
This continues work on #19571

This PR includes the following:
- implemented `image` ops: `compute_homography_matrix`, `perspective_transform`, `gaussian_blur`, and `elastic_transform` (very similar to jax implementations)
- adjust `random.normal` to use `loc` and `mean` directly in `mlx` method


fchollet (approved):
Thank you for the PR!

How much work do you estimate remains before we can ship the MLX backend? Should we try to get help from Apple?

PR #21203, opened 2025-04-23
Fix nightly releases

mattdangerw:
They have been broken for a month

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -126,7 +126,5 @@ jobs:
           pip install -r requirements.txt --progress-bar off --upgrade
           pip uninstall -y keras keras-nightly
           pip install -e "." --progress-bar off --upgrade
-      - name: Install pre-commit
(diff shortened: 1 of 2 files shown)


.github/workflows/actions.yml
```diff
@@ -126,7 +126,5 @@ jobs:
           pip install -r requirements.txt --progress-bar off --upgrade
           pip uninstall -y keras keras-nightly
           pip install -e "." --progress-bar off --upgrade
-      - name: Install pre-commit
-        run: pip install pre-commit && pre-commit install
```
hertschuh, line 130: Why is this not needed?

mattdangerw: It comes in the requirements. and this is the right way we shouldn't manually list packages in our GitHub action flow. We keep forgetting to update these and we want them easily maintainable.

hertschuh (approved):

PR #21209, opened 2025-04-25
OpenVINO: linspace

vi-shruti:
@rkazants for review of linspace Implementations for OpenVINO.


vi-shruti:
Requesting review: @rkazants  / @mattdangerw

keras/src/backend/openvino/numpy.py
```diff
@@ -910,9 +910,156 @@ def less_equal(x1, x2):
...
+    """
+
+    dtype = standardize_dtype(dtype) or config.floatx()
+    out_dtype = OPENVINO_DTYPES[dtype]
+    dtype = OPENVINO_DTYPES[config.floatx()]
+
+    start = get_ov_output(start, dtype)
+    stop = get_ov_output(stop, dtype)
+    start = ov_opset.convert(start, dtype).output(0)
+    stop = ov_opset.convert(stop, dtype).output(0)
```
rkazants, line 967: not needed because `get_ov_output` already performs `convert`. So please remove

keras/src/backend/openvino/numpy.py
```diff
@@ -910,9 +910,156 @@ def less_equal(x1, x2):
...
+
+    if isinstance(num, OpenVINOKerasTensor):
+        num = get_ov_output(num, Type.i32)
+    elif isinstance(num, int):
+        num = ov_opset.constant(num, Type.i32).output(0)
+    else:
+        raise TypeError("`num` must be an int or OpenVINOKerasTensor.")
+    num = ov_opset.convert(num, Type.i32).output(0)
```
rkazants, line 975: not needed

keras/src/backend/openvino/numpy.py
```diff
@@ -910,9 +910,156 @@ def less_equal(x1, x2):
...
+    stop = get_ov_output(stop, dtype)
+    start = ov_opset.convert(start, dtype).output(0)
+    stop = ov_opset.convert(stop, dtype).output(0)
+
+    if isinstance(num, OpenVINOKerasTensor):
+        num = get_ov_output(num, Type.i32)
+    elif isinstance(num, int):
+        num = ov_opset.constant(num, Type.i32).output(0)
+    else:
+        raise TypeError("`num` must be an int or OpenVINOKerasTensor.")
```
rkazants, line 974: I think `num = get_ov_output(num, Type.i32)` should be enought for all cases

keras/src/backend/openvino/numpy.py
```diff
@@ -910,9 +910,156 @@ def less_equal(x1, x2):
...
+        post = ov_opset.range(axis_p1, rank, one_i).output(0)
+        zero_i = ov_opset.reshape(
+            zero_i, ov_opset.constant([1], Type.i32), False
+        ).output(0)
+        perm = ov_opset.concat([pre, zero_i, post], 0).output(0)
+        y = ov_opset.transpose(y, perm).output(0)
+
+    y = ov_opset.convert(y, out_dtype).output(0)
+    if retstep:
+        return OpenVINOKerasTensor(y), ov_opset.convert(step, out_dtype).output(
```
rkazants, line 1059: create `return_step` variable and assign. `return_step` should be wrapped with `OpenVINOKerasTensor`

keras/src/backend/openvino/numpy.py
vi-shruti: Thanks for your review, @rkazants.
I've incorporated the rest of the changes but removing this seems to cause type errors.

keras/src/backend/openvino/numpy.py
```diff
@@ -910,9 +910,151 @@ def less_equal(x1, x2):
...
-    )
+    """
+    Return evenly spaced numbers over a specified interval.
+
+    Returns `num` evenly spaced samples, calculated over the
+    interval [`start`, `stop`].
+
+    The endpoint of the interval can optionally be excluded.
+
+    Parameters
```
hertschuh, line 921: Remove, this doc will never be generated or seen, only the one in `keras.ops`. Also, it's not following the format that we use for docstring.

hertschuh (changes requested):

PR #21213, opened 2025-04-26
Added numpy.outer function using opset14

blue-dreaming-jay:
Modified import statement for opset14 in numpy.py (original one commented out)
Added numpy.outer function using opset14
Deleted test_outer and test_outer_ lines from excluded_concrete_tests.txt
Implementation of numpy.outer assumes that the given tensors are 1-dimensional arrays, and will likely produce bad results


gbaned:
Hi @blue-dreaming-jay Can you please resolve the conflicts? Thank you.

blue-dreaming-jay:
I will resolve them. I'd like to clarify if the type that the produced tensor should be is the type produced by applying dtypes.result_type() on the dtypes of inputted tensors x1 and x2?

hertschuh:
This was addressed by https://github.com/keras-team/keras/pull/21270

PR #21214, opened 2025-04-27
Implement bartlett function in keras.ops

shashaka:
Adds keras.ops.bartlett, which computes the Bartlett window. Supported across NumPy, TensorFlow, PyTorch, and JAX backends, but not supported on OpenVINO.


fchollet (approved):
Thanks for the PR -- very nice work! As usual, any op that is in NumPy is in scope for inclusion in the Keras API.

LGTM

PR #21216, opened 2025-04-27
implementation of logaddexp in openvino backend (issue #29489)

mathys-mt:
I implemented the numpy logaddexp function in openvino backend using functions already coded in this file. The tests seem to pass, but maybe implementing logaddexp from zero without using the openvino numpy functions would be more effective. @rkazants could you review the PR ?


rkazants:
do NOT merge!!!

hertschuh:
This was implemented by https://github.com/keras-team/keras/pull/21522

Issue #21201, opened 2025-04-22
Embedding Layer trainable bug?


Fricker95:
In the class `Embedding(Layer)` [keras/src/layers/core/embedding.py](https://github.com/keras-team/keras/blob/master/keras/src/layers/core/embedding.py)

``` python
    def build(self, input_shape=None):
        if self.built:
            return
        if self.quantization_mode is not None:
            self.quantized_build(input_shape, mode=self.quantization_mode)
        if self.quantization_mode != "int8":
            self._embeddings = self.add_weight(
                shape=(self.input_dim, self.output_dim),
                initializer=self.embeddings_initializer,
                name="embeddings",
                regularizer=self.embeddings_regularizer,
                constraint=self.embeddings_constraint,
                trainable=True,
            )
        self.built = True
        if self.lora_rank:
            self.enable_lora(self.lora_rank)
```

Here the embeddings weight is built using `trainable=True`, this doesn't reflect the `keras.layers.Layer` parent class attribute `_trainable`. Therefore if we build the `Embedding` class using `trainable=False` as a parameter the underlying `self._embeddings` weight will still be trainable.

Is this set to true for a specific reason or is it simply a bug?
Thought I would point this out.

JyotinderSingh:
The existing layer behavior works as expected.

Note that whether a layer is affected by backprop or not is controlled by the `_trainable` argument at the base `Layer` level.

1. When we instantiate the `Embedding` layer and pass a value for `trainable` (either `True` or `False`), this value is captured by the `**kwargs` in the `Embedding`'s `__init__` method.
2. Inside the `Embedding`'s `__init__` method, the line `super().__init__(**kwargs)` calls the `__init__` method of its parent class, which is `Layer`. This passes along the `trainable` we provide.
https://github.com/keras-team/keras/blob/37eacb012dff7bc6ffece2d5992faa4d279ed244/keras/src/layers/core/embedding.py#L100
3. The `Layer`'s `__init__` method then retrieves the `trainable` value from the `kwargs` (or uses the default True if it's not provided) and sets the `self._trainable` attribute accordingly.
https://github.com/keras-team/keras/blob/37eacb012dff7bc6ffece2d5992faa4d279ed244/keras/src/layers/layer.py#L303

This is then made available for reading and mutation by this property decorator. You can also see how the setter works (and dynamically controls the training value in case of callbacks)

[...]

PR #21226, opened 2025-04-30
updated embedding.py, cannot have hardcoded value True for trainable.

pctablet505:
Can't hardcode trainable to be true.
#21201

### DIFF
--- a/keras/src/layers/core/embedding.py
+++ b/keras/src/layers/core/embedding.py
@@ -128,7 +128,7 @@ class Embedding(Layer):
                 name="embeddings",
                 regularizer=self.embeddings_regularizer,
                 constraint=self.embeddings_constraint,
-                trainable=True,
+                trainable=self._trainable,
             )
         self.built = True
         if self.lora_rank:
--- a/keras/src/layers/core/embedding.py
+++ b/keras/src/layers/core/embedding.py
@@ -128,7 +128,6 @@ class Embedding(Layer):
                 name="embeddings",
                 regularizer=self.embeddings_regularizer,
                 constraint=self.embeddings_constraint,
-                trainable=True,
             )
         self.built = True
         if self.lora_rank:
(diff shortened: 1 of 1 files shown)


JyotinderSingh:
The existing layer behavior works as expected.

Note that whether a layer is affected by backprop or not is controlled by the _trainable argument at the base `Layer` level.

1. When we instantiate the `Embedding` layer and pass a value for `trainable` (either `True` or `False`), this value is captured by the `**kwargs` in the `Embedding`'s `__init__` method.
2. Inside the `Embedding`'s `__init__` method, the line `super().__init__(**kwargs)` calls the `__init__` method of its parent class, which is `Layer`. This passes along the `trainable` we provide.
https://github.com/keras-team/keras/blob/37eacb012dff7bc6ffece2d5992faa4d279ed244/keras/src/layers/core/embedding.py#L100
3. The `Layer`'s `__init__` method then retrieves the `trainable` value from the `kwargs` (or uses the default True if it's not provided) and sets the `self._trainable` attribute accordingly.
https://github.com/keras-team/keras/blob/37eacb012dff7bc6ffece2d5992faa4d279ed244/keras/src/layers/layer.py#L303

This is then made available for reading and mutation by this property decorator. You can also see how the setter works (and dynamically controls the training value in case of callbacks)

https://github.com/keras-team/keras/blob/37eacb012dff7bc6ffece2d5992faa4d279ed244/keras/src/layers/layer.py#L586-L609

The following API is used to determine whether a layer is affected by backprop:

https://github.com/keras-team/keras/blob/37eacb012dff7bc6ffece2d5992faa4d279ed244/keras/src/layers/layer.py#L638-L646

You can see in the trainer code here that we retrieve the weights we want to optimize using the above API:
https://github.com/keras-team/keras/blob/37eacb012dff7bc6ffece2d5992faa4d279ed244/keras/src/backend/tensorflow/trainer.py#L75-L80

PR #21230, opened 2025-04-30
Add numpy.flip support for OpenVINO backend. 

chuinzer:
This change resolves #29359. @rkazants  please review this PR.
My bad for spamming PRs, had some problems with CLA.


keras/src/backend/openvino/excluded_concrete_tests.txt
```diff
@@ -82,7 +82,6 @@ NumpyOneInputOpsCorrectnessTest::test_cumprod
 NumpyOneInputOpsCorrectnessTest::test_diag
```
rkazants, line 82: also, please remove `NumpyDtypeTest::test_flip` above

keras/src/backend/openvino/numpy.py
```diff
@@ -1,4 +1,5 @@
 import numpy as np
+import openvino.runtime.opset1 as ov1
```
rkazants, line 2: not needed. use `ov_opset`

chuinzer:
Looking into the error. Will take some time.

PR #21231, opened 2025-04-30
Openvino numpy ravel

SiddharthV147:
Added numpy.ravel 

@rkazants can you please review it


rkazants (commented):

fchollet (approved):
Thanks!

PR #21232, opened 2025-04-30
Don't create unused optimizer variables.

cantonios:
If `variable.overwrite_with_gradient == True`, then the only optimizer variable ever used for that variable is `base_optimizer._accumulated_gradients`. All other optimizer variables are unused.  This can be extremely wasteful if the training variables are large, for example in the case of large embedding tables that span multiple hosts/devices.

Added a convenience function in the base optimizer `add_optimizer_variables(...)` that loops through the variable list and automatically adds a variable only if appropriate.  If a variable would otherwise be unused, a `None` is inserted into the list.  This is needed to keep `optimizer._get_variable_index()` consistent. Updated all built-in optimizers to use this.

NOTE: if a custom optimizer that exists out in the wild still does create unused optimizer variables, the optimizer should still work - it will just be wasteful.  IOW this should not be a breaking change.

### DIFF
--- a/keras/src/backend/torch/optimizers/torch_parallel_optimizer.py
+++ b/keras/src/backend/torch/optimizers/torch_parallel_optimizer.py
@@ -15,7 +15,9 @@ class TorchParallelOptimizer(BaseOptimizer):
 
     @torch_utils.no_grad
     def _backend_reset_gradient_accumulators(self):
-        acc_list = [v.value for v in self._accumulated_gradients]
+        acc_list = [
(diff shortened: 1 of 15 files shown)


cantonios:
@hertschuh, I pulled this change out from the updater PR.  This is somewhat independent, in that it applies to the existing `overwrite_with_variable` attribute as it currently exists.

fchollet (approved):
That's a nice simplification as well -- LGTM

PR #21233, opened 2025-04-30
Clear static loss-scale for inner optimizer in LossScaleOptimizer.

cantonios:
The outer `LossScaleOptimizer` ignores the inner's loss-scale factor when scaling the loss.  When computing unscaled gradients, we therefore need to eliminate the inner's loss scale factor, otherwise the gradients get incorrectly scaled.


cantonios:
@hertschuh this fixes a small loss-scale issue.

The other way to handle this would be to stack the two loss scales.  That would have the advantage of not needing to modify the `inner_optimizer` directly, but has the disadvantage that the gradients would potentially need to be scaled twice for correctness. 
 Wasn't sure which was most appropriate, so chose the potentially most-efficient one.

fchollet (approved):
LGTM, thanks!

PR #21241, opened 2025-05-02
Fix message formatting

emmanuel-ferdman:
# PR Summary
This small PR updates message strings by adding missing f-strings to properly evaluate variable values. I initially noticed one missing f-string and then reviewed the codebase for similar cases. Hopefully, I’ve caught them all.


fchollet (approved):
Thanks for the fix!

Issue #21150, opened 2025-04-10
Memory Usage Monitoring Callback


DimiChatzipavlis:
Hi!
I 'd like to contribute to the keras repository along with one colleague of mine. We had an idea to build a new callback -named MemoryUsageCallback- to to enhance memory monitoring during training. The callback could leverage the "psutil" library for CPU memory usage monitoring and "tf.config.experimental.get_memory_info()" for GPU memory monitoring. We believe this feature would be a valuable addition to Keras, mainly because it will provide detailed insights into resource usage and it will assist users in optimizing training performance and debugging memory-related issues. Should we move forward with this contribution? 
We are eager to hear your feedback!

DimiChatzipavlis:
In the callbacks section, we wrote 2 files, the memory_usage_callback.py, for CPU,GPU memory usage monitoring, Tensorboard logging and the memory_usage_callback_test.py (all tests passed) -and we properly modified the __init__.py-. We add "psutil" at the requirements-common.txt. Should be confident to create a PR for the contribution now?

divyashreepathihalli:
@DimiChatzipavlis This sounds very useful. However, the solution needs to work across backend and not just with TF. If you have a working solution for all backends, please feel free to open a PR

DimiChatzipavlis:
Hello keras' members! I need your help to complete my pull request. I am giving a try to constantly reshape my code, but I fail in some tests in the Pull Request. Although, my main conern is the openvino test failure. Look my output from the test integrations in openvino:  
                                                                                                                                                                                               
Run python integration_tests/import_test.py
Traceback (most recent call last):
  File "/home/runner/work/keras/keras/integration_tests/import_test.py", line 5, in <module>
    from keras.src import backend
  File "/home/runner/work/keras/keras/keras/api/__init__.py", line 7, in <module>
    from keras import _tf_keras as _tf_keras
  File "/home/runner/work/keras/keras/keras/api/_tf_keras/__init__.py", line 1, in <module>
    from keras._tf_keras import keras
  File "/home/runner/work/keras/keras/keras/api/_tf_keras/keras/__init__.py", line 7, in <module>
    from keras import activations as activations
  File "/home/runner/work/keras/keras/keras/api/activations/__init__.py", line 7, in <module>
    from keras.src.activations import deserialize as deserialize
  File "/home/runner/work/keras/keras/keras/src/__init__.py", line 1, in <module>
    from keras.src import activations
[...]

PR #21245, opened 2025-05-03
Add memory usage monitor callback

DimiChatzipavlis:
We made the memory monitor callback(CPU/GPU monitoring) according to devs instructions (Issue No:#21150-->Tensorboard integration, support for all backends).We are looking forward to receiving your informative feedback!


fchollet:
Thanks for the PR! Can you link to a Colab showing the callback in action (maybe with different backends)?

DimiChatzipavlis:
Hi everyone!
I’ve put together a Colab demo showing the MemoryUsageCallback in action (including a TensorBoard integration):
https://colab.research.google.com/drive/1-vV1D98TtGN5A9Cx37aW_7qE-CtoFBfd?usp=sharing  

My callback works for CPU, TensorFlow, PyTorch, and JAX—and even writes scalars to TensorBoard—but OpenVINO doesn’t expose any memory-stats API (and isn’t typically used for training workloads), so it isn’t strictly required here and its tests keep failing in CI. Does anyone have suggestions for:

* Skipping or mocking out the OpenVINO memory tests?

* Alternatively, adding minimal OpenVINO support (even just a warning) so that the import_test passes without installation?

Thanks in advance and I’m eager for your feedback, both for the colab and openvino test.

divyashreepathihalli:
Thank you for the PR @DimiChatzipavlis - taking a look!

keras/src/callbacks/memory_usage_callback.py
```diff
@@ -0,0 +1,192 @@
...
+        (TensorFlow, PyTorch, JAX)
+
+    Logs are printed to stdout at the start/end of each epoch and,
+    if `log_every_batch=True`, after every batch.
+    If `tensorboard_log_dir` is provided, scalars are also written
+    via `tf.summary` (TensorBoard).
+
+    Args:
+        monitor_gpu (bool): If True, attempt to measure accelerator memory.
+        log_every_batch (bool): If True, also log after each batch.
```
divyashreepathihalli, line 33: what is the default behavior? 
Log at the end of each epoch? - You would document  default behavior in docstring

keras/src/callbacks/memory_usage_callback.py
```diff
@@ -0,0 +1,192 @@
...
+        self,
+        monitor_gpu=True,
+        log_every_batch=False,
+        tensorboard_log_dir=None,
+    ):
+        super().__init__()
+        if psutil is None:
+            raise ImportError(
+                "MemoryUsageCallback requires the 'psutil' library. "
+                "Install via `pip install psutil`."
```
divyashreepathihalli, line 51: NIT : "To install please use `pip install psutil`"

keras/src/callbacks/memory_usage_callback.py
```diff
@@ -0,0 +1,192 @@
...
+                self._writer = None
+
+    def on_train_begin(self, logs=None):
+        self._step_counter = 0
+
+    def on_epoch_begin(self, epoch, logs=None):
+        self._log_epoch("start", epoch)
+
+    def on_epoch_end(self, epoch, logs=None):
+        self._log_epoch("end", epoch, offset=1)
```
divyashreepathihalli, line 80: from the colab output I am observing that Epoch end is not logged when log_every_batch is False

keras/src/callbacks/memory_usage_callback.py
```diff
@@ -0,0 +1,192 @@
...
+            with self._writer.as_default(step=int(step)):
+                tf.summary.scalar("Memory/CPU_MB", cpu_mb)
+                if gpu_mb is not None:
+                    tf.summary.scalar("Memory/GPU_MB", gpu_mb)
+            # flush happens inside writer
+
+    def _get_cpu_memory(self):
+        return self._proc.memory_info().rss / (1024**2)
+
+    def _get_gpu_memory(self):
```
divyashreepathihalli, line 119: another function to get tpu memory would be needed as well

divyashreepathihalli (approved):
Thanks for the PR!
I tried the Colab on A100 GPU with monitor_gpu set to True
but the logs did not show GPU memory usage
image

fchollet:
Please make sure to insert line breaks in log messages so that the logs do not interfere too much with the progress bar printouts.

divyashreepathihalli:
Thank you for the updates!
looks like GPU is working but not TPU
image

DimiChatzipavlis:
Hi everyone! I’ve added two small tweaks to running_on_tpu():
For TensorFlow TPUs, we now call TPUClusterResolver + initialize_tpu_system() before checking list_logical_devices("TPU") so Colab’s TPU actually comes up.
In _get_tpu_memory(), we fall back to summing bytes_in_use for JAX-TPU.

These changes should finally let the callback detect TPUs in a Colab TPU runtime. Unfortunately I haven’t been able to verify end-to-end in Colab’s TPU runtime (TPUs always report empty-neither me nor my colleague, perhaps due to resources' shortage in the colab), so any tips on a working TPU setup would be appreciated.

Thanks in advance and I am waiting your feedback!

laxmareddyp:
Hi @DimiChatzipavlis ,

I cloned your repository and ran the notebook, but noticed the GPU memory allocation reported during each epoch is much lower than expected. For example, with a batch size of 64 and image size 224x224, the first layer should output a tensor of shape 64x224x224x32, which requires about 411MB (calculated as 4×64×224×224×32). However, the callback only reports 60MiB, which is far below the actual memory needed for training—even without rematerialization. This suggests the reported values do not accurately reflect true GPU memory consumption

Refer this gist [rematerizalization-with-callback](https://colab.research.google.com/gist/laxmareddyp/4cf805991894826f34d7381e5367f516/-rematerizalization-with-callback.ipynb)

laxmareddyp:
Hi @DimiChatzipavlis ,

In commit `bd7fc07`, you switched to using peak memory, which worked well. However, in `8f37649`, the use of peak memory appears to have been removed, and I'm seeing the same low VRAM readings as before. Could you please take a look at this?

PR #21246, opened 2025-05-04
Add NotImplementedError for angle & bartlett on openvino

shashaka:
Raise NotImplementedError in the OpenVINO implementation for:
- angle
- bartlett


fchollet (approved):
LGTM, thanks!

PR #21247, opened 2025-05-04
Interleave optimizer variable creation to restore backward-compatibility

cantonios:
Added optimizer variables were orginally interleaved during `build` prior to #21232, e.g. `{momentum0, velocity0, momentum1, velocity1, ...}`.  In #21232, the order was changed to non-interleaved for some optimizers, e.g. `{momentum0, momentum1, ..., velocity0, velocity1, ...}`.  This broke some custom checkpoint serialization compatibility that relied on the order of variables remaining consistent.

Here we modify the base function `add_optimizer_variables(...)` to support creating multiple optimizer variables per training variable, and interleaves creation to restore backward compatibility.

### DIFF
--- a/keras/src/optimizers/adadelta.py
+++ b/keras/src/optimizers/adadelta.py
@@ -75,11 +75,10 @@ class Adadelta(optimizer.Optimizer):
         if self.built:
             return
         super().build(var_list)
-        self._accumulated_grads = self.add_optimizer_variables(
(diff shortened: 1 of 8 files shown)


cantonios:
@hertschuh, this allows the RecML tests to all pass.

fchollet (approved):
Thanks for the fix.

PR #21248, opened 2025-05-04
Simplify the logic of `quantize` and float8 build.

james77777778:
I'm working on bringing QAT in Keras and want to clean up the current quantization impl.

This PR simplifies the logic of `quantize` in `Dense`, `Einsum`, and `Embedding` by utilizing `quantized_build`.
It also adds `overwrite_with_gradient` to `Layer.add_weight` to avoid manually assigning the property to variables.

### DIFF
--- a/keras/src/layers/core/dense.py
+++ b/keras/src/layers/core/dense.py
@@ -107,15 +107,15 @@ class Dense(Layer):
         self.supports_masking = True
 
     def build(self, input_shape):
-        input_dim = input_shape[-1]
+        kernel_shape = (input_shape[-1], self.units)
         if self.quantization_mode:
-            self.quantized_build(input_shape, mode=self.quantization_mode)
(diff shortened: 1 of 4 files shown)


fchollet (approved):
LGTM, thank you!

james77777778:
The CI issue seems to be a false alarm.
Any updates needed?

SamanehSaadat:
Hi @james77777778!
Since the `_int8_build` signature has changed,  we need to update its references in the KerasHub as well. For example, this change is breaking this: 
https://github.com/keras-team/keras-hub/blob/c00db4e2c64d0f4bba5c8845ed3024cef90137db/keras_hub/src/layers/modeling/reversible_embedding.py#L194-L196

james77777778:
@SamanehSaadat thanks for informing me. Will update it soon!

PR #21250, opened 2025-05-05
Update nn.py

pctablet505:
Added support for flash attention with sharding, fixed issue when using flash attention on tpu.

### DIFF
--- a/keras/api/_tf_keras/keras/ops/__init__.py
+++ b/keras/api/_tf_keras/keras/ops/__init__.py
@@ -60,6 +60,8 @@ from keras.src.ops.math import segment_max as segment_max
 from keras.src.ops.math import segment_sum as segment_sum
 from keras.src.ops.math import stft as stft
 from keras.src.ops.math import top_k as top_k
+from keras.src.ops.math import view_as_complex as view_as_complex
+from keras.src.ops.math import view_as_real as view_as_real
 from keras.src.ops.nn import average_pool as average_pool
 from keras.src.ops.nn import batch_normalization as batch_normalization
 from keras.src.ops.nn import binary_crossentropy as binary_crossentropy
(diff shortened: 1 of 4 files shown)


keras/src/backend/jax/nn.py
```diff
@@ -1177,54 +1178,111 @@ def dot_product_attention(
...
-            custom_mask=mask,
-            attn_logits_soft_cap=attn_logits_soft_cap,
+    # Check if inputs are sharded
+    inputs_sharded = any(
+        hasattr(t, "sharding") and not t.sharding.is_fully_replicated
+        for t in (query, key, value)
+    )
+    is_tpu = jax.devices()[0].platform == "tpu"
+
+    # Determine flash attention compatibility
```
divyashreepathihalli, line 1189: I am very confused about the logic here.

- why is FA disabled in inputs are sharded?

keras/src/backend/jax/nn.py
```diff
@@ -1177,54 +1178,111 @@ def dot_product_attention(
...
+    if flash_attention is None:
+        flash_attention = (
+            not inputs_sharded or is_tpu
+        ) and _can_use_flash_attention(query, key, value, bias)
+    elif flash_attention and inputs_sharded and not is_tpu:
```
divyashreepathihalli, line 1194: this condition is weird
if FA is enabled and inputs are sharded and is not running on TPU - you are disabling FA? why? can you please explain?

keras/src/backend/jax/nn.py
```diff
@@ -1177,54 +1178,111 @@ def dot_product_attention(
...
+        segment_ids = jnp.zeros([bs, q_len], dtype=jnp.int32)
+        decoder_segment_ids = splash_attention_kernel.SegmentIds(
+            q=segment_ids, kv=segment_ids
         )
-        out = jnp.transpose(out, axes=(0, 2, 1, 3))
-        return out
 
-    # `dot_product_attention` is only available in jax>=0.4.31
+        # Process mask for Splash Attention
+        custom_mask = None
```
divyashreepathihalli, line 1219: lets verify numerics remain consistent with this updated code to mask

divyashreepathihalli (commented):
Thanks for the PR!
The gating logic is a little confusing for me. I left some comments. Thanks!

PR #21252, opened 2025-05-05
Keras <> NNX integration

divyashreepathihalli:
The PR integrates NNX into JAX backend!

The following snippet shows how you would enable the nnx backend
```
import os
os.environ["KERAS_BACKEND"]="jax"
os.environ["KERAS_NNX_ENABLED"]="true"
import keras
```
Demo colab here : https://colab.sandbox.google.com/drive/1mK-4qbce2HGRIkcb4v5n4niWGDezL_6n#scrollTo=m-ZH9Mpnphfz
Added a github workflow action for nnx backend. Note this will fail - because this needs a new release of flax to work.


keras/src/random/random_test.py
```diff
@@ -380,12 +381,11 @@ def test_uniform_dtype_validation(self):
         reason="This test requires `jax` as the backend.",
     )
     def test_dropout_jax_jit_stateless(self):
-        import jax
         import jax.numpy as jnp
 
         x = ops.ones(3)
 
-        @jax.jit
+        @nnx.jit
```
fchollet, line 388: Would the integration prevent the use of `jax.jit` with Keras layers?

divyashreepathihalli: yes! it would only work with nnx.jit for now ( They might be working on adding support for jax.jit)

keras/src/ops/operation.py
```diff
@@ -119,7 +122,10 @@ def __new__(cls, *args, **kwargs):
         to manually implement `get_config()`.
         """
         instance = super(Operation, cls).__new__(cls)
+        if backend.backend() == "jax":
+            from flax import nnx
```
mattdangerw, line 124: I don't think we can assume flax is installed when jax is installed. This is an unsafe assumption. Might need to conditional import and handle gracefully.

keras/src/backend/jax/trainer.py
```diff
@@ -5,13 +5,15 @@
 
 import jax
 import numpy as np
+from flax import nnx
```
mattdangerw: Same here. Unsafe to assume flax is installed with every jax install. We will need to handle this gracefully when flax is not around.

keras/src/backend/jax/core.py
```diff
@@ -3,6 +3,7 @@
 import jax.numpy as jnp
 import ml_dtypes
 import numpy as np
+from flax import nnx
```
mattdangerw: same here

keras/src/backend/jax/trainer.py
```diff
@@ -233,7 +235,10 @@ def concatenate(outputs):
                     return output
 
                 if not self.run_eagerly and self.jit_compile:
-                    concatenate = jax.jit(concatenate)
+                    if is_nnx_backend_enabled():
```
mattdangerw: Can we put this in a util or something? So we can just call the same fn that resolves correctly.

requirements-jax-cuda.txt
```diff
@@ -9,6 +9,6 @@ torch==2.6.0
 # Jax with cuda support.
 --find-links https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
 jax[cuda12]==0.6.0
-flax
+flax>=0.10.1
```
mattdangerw: This is redundant. requirement-common is included two lines below this. Just update requirement common and you are done, no other changes needed.

requirements.txt
```diff
@@ -14,7 +14,7 @@ torch-xla==2.6.0;sys_platform != 'darwin'
 # Pinned to 0.5.0 on CPU. JAX 0.5.1 requires Tensorflow 2.19 for saved_model_test.
 # Note that we test against the latest JAX on GPU.
 jax[cpu]==0.5.0
-flax
+flax>=0.10.1
```
mattdangerw: same here, remove this

keras/src/backend/config.py
```diff
@@ -15,6 +15,9 @@
 # Default backend: TensorFlow.
 _BACKEND = "tensorflow"
 
+# Default for NNX backend features.
+_NNX_ENABLED_KEY = "nnx_enabled"
```
mattdangerw: probably just get rid of this? we aren't tracking our other json keys as constants, doing it for only one is just confusing

keras/src/backend/config.py
```diff
@@ -274,36 +298,19 @@ def keras_home():
...
-    try:
-        with open(_config_path, "w") as f:
-            f.write(json.dumps(_config, indent=4))
-    except IOError:
-        # Except permission denied.
-        pass
-
-# Set backend based on KERAS_BACKEND flag, if applicable.
+else:
+    _config = {}
```
mattdangerw: i don't think this line does anything?

keras/src/backend/config.py
```diff
@@ -403,3 +411,36 @@ def max_steps_per_epoch():
...
+        with open(_config_path, "w") as f:
+            f.write(json.dumps(_config_to_save, indent=4))
+    except IOError:
+        # Except permission denied.
+        pass
+
+if "KERAS_NNX_ENABLED" in os.environ:
+    env_val = os.environ["KERAS_NNX_ENABLED"].lower()
+    if env_val == "true":
+        _initial_nnx_value_from_config = True
```
mattdangerw: Try to rewrite the changes to this file trying to follow the local style a little more. This is overcomplicated.

- no need to move big blocks around.
- keep a constant `_NNX_ENABLED` at the top of the file, somewhere after `_BACKEND`.
- set it from config the same way `_BACKEND` is handled
- write it to config the same way `_BACKEND` is handled

Get rid of `_NNX_ENABLED_KEY`, `_initial_nnx_value_from_config`, `_current_nnx_status_for_saving`.

keras/src/backend/jax/core.py
```diff
@@ -68,7 +293,7 @@ def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
         # an existing distributed jax array will raise error.
         return x
 
-    if isinstance(x, Variable):
+    if isinstance(x, (JaxVariable, NnxVariable)):
```
mattdangerw: This will need some thought on how to handle this gracefully when flax is not installed. Maybe we should still have one class `JaxVariable` that only has a nnx stuff when flax is installed and nnx enabled. I am not sure.

mattdangerw: Maybe NNX variable should just subclass JaxVariable so you can keep `isinstance` checking and know it will work (as well as any funky user code doing the same thing). Override what you need to from the super class. Will still be a little tricky to figure out how to declare the class when nnx might not be importable in the env.

keras/src/backend/config.py
divyashreepathihalli: There are a lot of circular import errors, because of which i had a move some of these blocks lower.

keras/src/backend/jax/core.py
divyashreepathihalli: I went for a simpler approach adding a _JAX_VARIABLE_TYPES tuple that would contain NnxVariable or JaxVariable depending on if they get successfully defined

divyashreepathihalli:
Added a github workflow action for nnx backend. Note **this will FAIL** - because this needs a new release of flax to work.

keras/src/random/random_test.py
divyashreepathihalli: Added nnx as a opt in with this flag - os.environ["KERAS_NNX_ENABLED"]

requirements-common.txt
```diff
@@ -24,3 +24,4 @@ coverage!=7.6.5  # 7.6.5 breaks CI
 # for onnx_test.py
 onnxruntime
 openvino
+flax>=0.10.1
```
hertschuh: Why here?

It cannot work because `requirements-common.txt` doesn't install jax (and it shouldn't because we don't know if we need `jax[cpu]` or `jax[cuda12]`).

So undo.

requirements-jax-cuda.txt
```diff
@@ -9,6 +9,5 @@ torch==2.6.0
 # Jax with cuda support.
 --find-links https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
 jax[cuda12]==0.6.0
-flax
```
hertschuh: Just change to `flax>=0.10.1`

requirements.txt
```diff
@@ -14,7 +14,6 @@ torch-xla==2.6.0;sys_platform != 'darwin'
 # Pinned to 0.5.0 on CPU. JAX 0.5.1 requires Tensorflow 2.19 for saved_model_test.
 # Note that we test against the latest JAX on GPU.
 jax[cpu]==0.5.0
-flax
```
hertschuh, line 17: Just change to `flax>=0.10.1`

.github/workflows/actions.yml
```diff
@@ -50,6 +50,10 @@ jobs:
           pip install -r requirements.txt --progress-bar off --upgrade
           pip uninstall -y keras keras-nightly
           pip install -e "." --progress-bar off --upgrade
+      - name: Install Flax for NNX backend
+        if: matrix.backend == 'nnx'
+        run: |
+          pip install flax --progress-bar off --upgrade
```
hertschuh: Not needed, it's already part of the requirements file.

.github/workflows/actions.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ['3.10']
-        backend: [tensorflow, jax, torch, numpy, openvino]
+        backend: [tensorflow, jax, torch, numpy, openvino, nnx]
```
hertschuh: I think you need to do something else in addition (or instead of this?), right now, this does not turn on NNX.

examples/demo_jax_distributed.py
```diff
@@ -12,7 +12,6 @@
 import jax.numpy as jnp
 import tensorflow as tf  # just for tf.data
 import keras  # Keras multi-backend
-
```
hertschuh: Undo this file.

keras/src/backend/config.py
```diff
@@ -230,6 +233,25 @@ def is_flash_attention_enabled():
     return global_state.get_global_attribute("flash_attention", default=None)
 
 
+@keras_export("keras.config.is_nnx_backend_enabled")
+def is_nnx_backend_enabled():
```
hertschuh: Should be call it `is_nnx_enabled` since we're not making it a "backend"?

keras/src/backend/jax/core.py
```diff
@@ -352,7 +597,7 @@ def fori_loop(lower, upper, body_fun, init_val):
 
 
 def stop_gradient(variable):
-    if isinstance(variable, Variable):
+    if isinstance(variable, (JaxVariable, NnxVariable)):
```
hertschuh: `_JAX_VARIABLE_TYPES`

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,248 @@ def __jax_array__(self):
...
+                    stateless_value = scope.get_current_value(self)
+                    if stateless_value is not None:
+                        return self._maybe_autocast(stateless_value)
+
+                return self._maybe_autocast(current_value)
+
+            def __hash__(self):
+                return id(self)
+
+        _JAX_VARIABLE_TYPES += (NnxVariable,)
```
hertschuh: Why do we need both? Under what circumstances would we have a mix of both types?

keras/src/backend/jax/layer.py
```diff
@@ -1,2 +1,19 @@
...
+
+
+if config.is_nnx_backend_enabled():
+    try:
+        from flax import nnx
+
+        class NnxLayer(nnx.Module):
+            def __init_subclass__(cls):
+                super().__init_subclass__()
+    except ImportError:
```
hertschuh: Instead of having this try/except block replicated in many places, let's just do it once in `config.py`.
When `set_nnx_enabled()` is called with `True`, then you try to import `flax.nnx` and fail with the message once for all.

keras/src/backend/jax/trainer.py
hertschuh: +1, create a wrapper decorator like you did in the guide.

guides/distributed_training_with_jax.py
```diff
@@ -185,8 +196,24 @@ def compute_loss(trainable_variables, non_trainable_variables, x, y):
 compute_gradients = jax.value_and_grad(compute_loss, has_aux=True)
 
 
+def conditional_jit(condition, *args, **kwargs):
```
hertschuh: You don't need to pass the condition here, you can just use `is_nnx_backend_enabled()` directly.

Also, I would just call it `jit`.

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,248 @@ def __jax_array__(self):
...
+                # Initialize nnx.Variable first.
+                if shape is not None and dtype is not None:
+                    # If initializer is a Keras callable, it's not ready yet.
+                    # If initializer is already a value, KerasVariable will
+                    # handle it. We need a concrete array for the placeholder.
+                    _placeholder_value = jnp.zeros(
+                        shape, dtype=standardize_dtype(dtype)
+                    )
+                elif shape is not None:
+                    _placeholder_value = jnp.zeros(shape, dtype=jnp.float32)
```
hertschuh: What about the passed dtype?

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,248 @@ def __jax_array__(self):
...
+                    actual_nnx_mutable = mutable
+
+                # Ensure 'mutable' is in nnx_metadata, but explicit 'mutable'
+                # param takes precedence.
+                if "mutable" in nnx_metadata and mutable is not None:
+                    nnx_metadata["mutable"] = actual_nnx_mutable
+                elif "mutable" not in nnx_metadata:
+                    nnx_metadata["mutable"] = actual_nnx_mutable
+
+                # Initialize nnx.Variable first.
```
hertschuh: Do we need a placeholder value? Does it allocate HBM? The issue is that we may be using twice the memory we need by initializing the variable twice.

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,248 @@ def __jax_array__(self):
...
+                    # handle it. We need a concrete array for the placeholder.
+                    _placeholder_value = jnp.zeros(
+                        shape, dtype=standardize_dtype(dtype)
+                    )
+                elif shape is not None:
+                    _placeholder_value = jnp.zeros(shape, dtype=jnp.float32)
+                else:
+                    _placeholder_value = jnp.array(0.0, dtype=jnp.float32)
+
+                # Call nnx.Variable.__init__ directly.
```
hertschuh: What about the sharding on the placeholder_value, do we need it?

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,248 @@ def __jax_array__(self):
...
+
+                if in_stateless_scope():
+                    scope = get_stateless_scope()
+                    stateless_value = scope.get_current_value(self)
+                    if stateless_value is not None:
+                        return self._maybe_autocast(stateless_value)
+
+                return self._maybe_autocast(current_value)
+
+            def __hash__(self):
```
hertschuh: Per the discussion we had, we can't do that. It's not compatible with the semantics of `__eq__` that we have.

I think my preferred fix would be to change NNX to not use variables as keys, but instead use `id(variable)` in dictionaries.

keras/src/ops/operation.py
```diff
@@ -119,6 +123,17 @@ def __new__(cls, *args, **kwargs):
         to manually implement `get_config()`.
         """
         instance = super(Operation, cls).__new__(cls)
+        if backend.backend() == "jax":
+            if is_nnx_backend_enabled():
```
hertschuh: nitpick: combine the 2 conditions with an and.

keras/src/ops/operation.py
```diff
@@ -119,6 +123,17 @@ def __new__(cls, *args, **kwargs):
...
+                try:
+                    from flax import nnx
+
+                    vars(instance)["_object__state"] = nnx.object.ObjectState()
```
hertschuh: We've talked about this, I'm really confused why it needs to go here instead of the Layer class, all the more so as `Operation` doesn't extend `nnx.Module`, so NNX should not care about Operations that are not Layers.

Maybe it's just a question of placing it early enough in `Layer.__init__`.

.github/workflows/actions.yml
divyashreepathihalli: It does, because I am seeing the nnx related error on this test workflow

keras/src/backend/jax/core.py
divyashreepathihalli: Yes,the place holder values is needed. The flax.nnx.Variable requires a concrete array for initialization (not an initializer function) when it's created. so added this.
Regarding memory usage - yep maybe

keras/src/backend/jax/core.py
divyashreepathihalli: Sharding is not applied to the place holder value. In Keras's NnxVariable, sharding (if specified with the layout parameter) is applied later during _direct_assign, which then uses self._layout to distribute the variable.

keras/src/ops/operation.py
divyashreepathihalli: set_attr is called before the object is registered in jax/layer class which was causing the object state error.
Moving it to layer  base class also does not solve the issue and causes the following error 
```
AttributeError                            Traceback (most recent call last)
[<ipython-input-8-2692540659>](https://localhost:8080/#) in <cell line: 0>()
      7 
      8 model = keras.Sequential([
----> 9     keras.layers.Dense(units=1, input_shape=(10,))
     10 ])
     11 

10 frames
[/usr/local/lib/python3.11/dist-packages/flax/nnx/object.py](https://localhost:8080/#) in _check_valid_context(self, error_msg)
    201 
    202   def _check_valid_context(self, error_msg: tp.Callable[[], str]) -> None:
--> 203     if not self._object__state.trace_state.is_valid():
    204       raise errors.TraceContextError(error_msg())
    205 

AttributeError: 'Dense' object has no attribute '_object__state'
```

keras/src/backend/config.py
hertschuh: Oh you renamed the setter `set_nnx_enabled`, but not the getter `is_nnx_enabled`.

keras/src/backend/jax/__init__.py
```diff
@@ -10,7 +11,11 @@
 from keras.src.backend.jax.core import IS_THREAD_SAFE
 from keras.src.backend.jax.core import SUPPORTS_RAGGED_TENSORS
 from keras.src.backend.jax.core import SUPPORTS_SPARSE_TENSORS
-from keras.src.backend.jax.core import Variable
+
+if is_nnx_backend_enabled():
```
hertschuh: I would prefer reverting this block and instead handling this in `keras/src/backend/jax/core.py`.

In fact, you already do it, but you called it `_JAX_VARIABLE_TYPE` instead of `Variable`, all you have to do is rename.

Then, you can also revert `keras/src/backend/__init__.py` to keep it consistent with the other backends.

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,249 @@ def __jax_array__(self):
...
+            # Ensure 'mutable' is in nnx_metadata, but explicit 'mutable'
+            # param takes precedence.
+            if "mutable" in nnx_metadata and mutable is not None:
+                nnx_metadata["mutable"] = actual_nnx_mutable
+            elif "mutable" not in nnx_metadata:
+                nnx_metadata["mutable"] = actual_nnx_mutable
+
+            # Initialize nnx.Variable first.
+            # Determine the dtype for the placeholder.
+            _placeholder_value = None
```
hertschuh: - if `dtype` is `None`, `standardize_dtype(dtype)` will return `config.floatx()`, so you don't need those ifs.
- you don't need to handle shape is `None` completely separately

This whole block of 4 ifs  from line 98 to 116 can be replaced with:
```python
_placeholder_value = jnp.zeros(shape or (), dtype=standardize_dtype(dtype))
```

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,249 @@ def __jax_array__(self):
...
+            if mutable is None:
+                actual_nnx_mutable = (
+                    trainable  # Keras 'trainable' maps to NNX 'mutable'
+                )
+            else:
+                actual_nnx_mutable = mutable
+
+            # Ensure 'mutable' is in nnx_metadata, but explicit 'mutable'
+            # param takes precedence.
+            if "mutable" in nnx_metadata and mutable is not None:
```
hertschuh: Because `mutable` is an explicit parameter in the function signature, "mutable" can never be in `nnx_metadata`, so these tests using `"mutable" in nnx_metadata` are always false.

Then, the intermediate value of `actual_nnx_mutable` is not needed.

This whole block from line 80 to 94 can be replaced with:
```python
nnx_metadata["mutable"] = trainable if mutable is None else mutable
```

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,249 @@ def __jax_array__(self):
...
+            )
+
+            # Store JAX-specific layout using object.__setattr__ BEFORE
+            # KerasVariable init.
+            # This is because KerasVariable.__init__ will call
+            # self._initialize, which uses self._layout.
+            object.__setattr__(self, "_layout", layout)
+
+            # Initialize KerasVariable.
+            super(NnxVariable, self).__init__(
```
hertschuh: Let's be explicit since this class has multiple super classes and relying on the order is brittle:

```python
KerasVariable.__init__(
    self,
    ...
```

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,249 @@ def __jax_array__(self):
...
+            self._direct_assign(new_keras_value)
+
+        def __getstate__(self):
+            # Get the state from KerasVariable (attributes in __dict__)
+            # KerasVariable does not have a custom __getstate__, so we mimic
+            # default behavior.
+            keras_state = self.__dict__.copy()
+
+            # Get the state from nnx.Variable
+            nnx_specific_state = super(KerasVariable, self).__getstate__()
```
hertschuh: Let's be explicit since this class has multiple super classes and relying on the order is brittle:
```python
nnx_specific_state = nnx.Variable.__getstate__(self)
```

keras/src/backend/jax/trainer.py
```diff
@@ -233,7 +234,7 @@ def concatenate(outputs):
                     return output
 
                 if not self.run_eagerly and self.jit_compile:
-                    concatenate = jax.jit(concatenate)
+                    concatenate = jit()(concatenate)
```
fchollet: Why not just `jit(fn)`? Why `jit()(fn)`

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,249 @@ def __jax_array__(self):
...
+
+        @_value.setter
+        def _value(self, new_keras_value):
+            self._direct_assign(new_keras_value)
+
+        def __getstate__(self):
+            # Get the state from KerasVariable (attributes in __dict__)
+            # KerasVariable does not have a custom __getstate__, so we mimic
+            # default behavior.
+            keras_state = self.__dict__.copy()
```
hertschuh: But that's going to grab the NNX state too, right? Also, why not use `__getstate__()` from the super class like for NNX below?

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,249 @@ def __jax_array__(self):
...
+            if "_trace_state" in nnx_specific_state:
+                keras_state["_trace_state"] = nnx_specific_state["_trace_state"]
+            if "_var_metadata" in nnx_specific_state:
+                keras_state["_var_metadata"] = nnx_specific_state[
+                    "_var_metadata"
+                ]
+
+            # Remove elements that might be problematic or redundant if
+            # nnx.Variable's __getstate__
+            keras_state.pop("raw_value", None)
```
hertschuh, line 151: This goes back to my comment about `__dict__` also grabbing the NNX state.

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,249 @@ def __jax_array__(self):
...
+                pass
+
+            if nnx_var_metadata is not None:
+                object.__setattr__(self, "_var_metadata", nnx_var_metadata)
+            else:
+                pass
+
+            # Ensure Keras's self._value is also consistent with the
+            # restored raw_value
+            object.__setattr__(self, "_value", nnx_raw_value)
```
hertschuh: But `_value` is a property, so you can't do that. You could do `self._value = nnx_raw_value` to actually invoke the property setter.

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,249 @@ def __jax_array__(self):
...
+
+            if hasattr(self, "_shape") and self._shape is not None:
+                self._ndim = len(self._shape)
```
hertschuh, line 184: Were those needed? `_ndim` should have been part of the state.

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,249 @@ def __jax_array__(self):
...
+            else:
+                # Fallback if shape isn't immediately available.
+                self._ndim = len(self.raw_value.shape)
+
+        def _initialize(self, value):
```
hertschuh: For `_initialize`, `_convert_to_tensor`, and `__jax_array__`, how about inheriting from `JaxVariable` instead of `KerasVariable`.

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,249 @@ def __jax_array__(self):
...
+            if self._layout is not None:
+                processed_value = distribution_lib.distribute_variable(
+                    value, self._layout
+                )
+            else:
+                processed_value = value
+
+            # Ensure that nnx.Variable part is initialized
+            if not hasattr(self, "_var_metadata"):
+                # todo: should add a warning
```
hertschuh: Please implement

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,249 @@ def __jax_array__(self):
...
+            self._direct_assign(value)
+
+        def _direct_assign(self, value):
+            # Apply JAX-specific distribution if layout is present
+            if self._layout is not None:
+                processed_value = distribution_lib.distribute_variable(
+                    value, self._layout
+                )
+            else:
+                processed_value = value
```
hertschuh: A common pattern is to just reuse the variable, `value` in this case.

```python
value = distribution_lib.distribute_variable(value, self._layout)
```

This way you don't need the else blocks, and it's more readable. In the same way, remove `final_value`.

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,249 @@ def __jax_array__(self):
...
+            return convert_to_tensor(value, dtype=dtype, sparse=False)
+
+        # Overload native accessor.
+        def __jax_array__(self):
+            return self.value
+
+        @property
+        def value(self):
+            if not hasattr(self, "raw_value"):
+                if not hasattr(self, "_value") or self._value is None:
```
hertschuh: `_value` is a property, so remove this test `hasattr(self, "_value")`, it will always be true.

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,249 @@ def __jax_array__(self):
...
+                current_value = self.raw_value
+                if (
+                    hasattr(self, "_var_metadata")
+                    and "on_get_value" in self._var_metadata
+                ):
+                    current_value = self._var_metadata["on_get_value"](
+                        self, current_value
+                    )
+
+            if in_stateless_scope():
```
hertschuh, line 209: That's supposed to be first, before the other options.

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,249 @@ def __jax_array__(self):
...
+
+        def _convert_to_tensor(self, value, dtype=None):
+            return convert_to_tensor(value, dtype=dtype, sparse=False)
+
+        # Overload native accessor.
+        def __jax_array__(self):
+            return self.value
+
+        @property
+        def value(self):
```
hertschuh, line 208: Can we call `super().value()` somewhere?

My concern is that if we add features or fix bug in the super class, we'll forget to replicate the changes here.

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,249 @@ def __jax_array__(self):
...
+                            " no initializer."
+                        )
+                current_value = self._value
+            else:
+                current_value = self.raw_value
+                if (
+                    hasattr(self, "_var_metadata")
+                    and "on_get_value" in self._var_metadata
+                ):
+                    current_value = self._var_metadata["on_get_value"](
```
hertschuh: Shouldn't this be called even if the value came from `self._value`?

Conversely, `self._maybe_autocast(...)` should happen in this branch too.

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,249 @@ def __jax_array__(self):
...
+        def _convert_to_tensor(self, value, dtype=None):
+            return convert_to_tensor(value, dtype=dtype, sparse=False)
+
+        # Overload native accessor.
+        def __jax_array__(self):
+            return self.value
+
+        @property
+        def value(self):
+            if not hasattr(self, "raw_value"):
```
hertschuh: I thought we're using `raw_value` as the source of truth. Do we need that whole branch using `_value`?

keras/src/backend/jax/core.py
```diff
@@ -55,6 +58,249 @@ def __jax_array__(self):
...
+
+            if in_stateless_scope():
+                scope = get_stateless_scope()
+                stateless_value = scope.get_current_value(self)
+                if stateless_value is not None:
+                    return self._maybe_autocast(stateless_value)
+
+            return self._maybe_autocast(current_value)
+
+        def __hash__(self):
```
hertschuh: The comment got lost. But we can't do this because it conflicts with our semantics of `__eq__`. It will create bizarre and really hard to diagnose bugs.

I believe the fix is that NNX should key by id, not variable object.

keras/src/backend/jax/layer.py
```diff
@@ -1,2 +1,10 @@
+from flax import nnx  # noqa F401
+
+
 class JaxLayer:
     pass
+
+
+class NnxLayer(nnx.Module):
+    def __init_subclass__(cls):
+        super().__init_subclass__()
```
hertschuh: Why is this needed if it just calls `super`?

keras/src/layers/layer.py
```diff
@@ -218,9 +222,11 @@ def call(self, inputs):
     ```
     """
 
+    def __init_subclass__(cls):
+        super().__init_subclass__()
```
hertschuh: Why is this needed if it just calls `super`?

keras/src/layers/layer.py
```diff
@@ -1533,7 +1539,19 @@ def __setattr__(self, name, value):
             if not hasattr(self, "_tracker"):
                 self._initialize_tracker()
             value = self._tracker.track(value)
-        return super().__setattr__(name, value)
+
+        # NNX-specific bypass for `_called` and `built` attributes
```
hertschuh, line 1546: Oh, I thought we were fixing that in NNX, is that not the case?

keras/src/ops/operation.py
```diff
@@ -14,6 +15,9 @@
 
 @keras_export("keras.Operation")
 class Operation:
+    def __init_subclass__(cls):
+        super().__init_subclass__()
```
hertschuh: Same question about why this is needed.

keras/src/ops/operation.py
```diff
@@ -118,7 +122,10 @@ def __new__(cls, *args, **kwargs):
         to manually implement `get_config()`.
         """
         instance = super(Operation, cls).__new__(cls)
+        if backend.backend() == "jax" and is_nnx_backend_enabled():
+            from flax import nnx
 
+            vars(instance)["_object__state"] = nnx.object.ObjectState()
```
hertschuh, line 126: The comment got lost, can we move this to `Layer`?

requirements-common.txt
hertschuh: Oh you still need to remove from now that it's back in `requirements.txt`. This won't work because we don't install jax, and we shouldn't here.

keras/src/utils/jax_utils.py
```diff
@@ -9,3 +10,18 @@ def is_in_jax_tracing_scope(x=None):
             if c.__name__ == "Tracer" and c.__module__.startswith("jax"):
                 return True
     return False
+
+
+def jit(*args, **kwargs):
+    def decorator(func):
```
hertschuh: You don't need a decorator function, you can immediately delegate to either `jax.jit` or `nnx.jit` and return that.

In fact, you can do:

```python
if is_nnx_backend_enabled():
  jit = nnx.jit
else:
  jit = jax.jit
```

keras/src/layers/layer.py
divyashreepathihalli: Apparently not.

keras/src/backend/jax/core.py
divyashreepathihalli: I have let Cristian know about this. He will be adding a fix.

keras/src/backend/jax/core.py
divyashreepathihalli: calling  `super().value()` would be complicated. `KerasVariable.value` typically reads from `self._value`
 In `NnxVariable`, `self._value` is a property that reads from `self.raw_value`. A naive call to `super().value` could lead to recursion if it tries to access `self._value` or `self.value`, or it might bypass `raw_value` and the NNX hooks.

keras/src/ops/operation.py
divyashreepathihalli: Unfortunately that did not work. It is strange. Linking my comment here
https://github.com/keras-team/keras/pull/21252#discussion_r2141251161

keras/src/ops/operation.py
divyashreepathihalli: not needed removing

keras/src/backend/jax/core.py
divyashreepathihalli: Spoke to Cristian! They will be updating this -  I will remove it once that is done

abheesht17:
Noticed the following for when we use a custom NNX training loop.

- When model is just a Keras layer, grads are non-zero and the loss goes down.
- But whenever I have a `keras.Model` instance (for example, a `keras.Sequential` model), grads are all 0.

Here is a notebook demonstrating the same: https://colab.research.google.com/drive/1hvGYN00aETHlrwjUlLCjfwVJrCcQqySN?resourcekey=0-oyfeCvzJzqOf0aLroY4rhg&usp=sharing.

Let me know if I've messed up somewhere! Trying to figure out why it's happening. Will changing the `trainable_variables` filter work? Probably not, because the filter is working properly (see the cell with `nnx.state`; it returns the correct variables).

divyashreepathihalli:
> Let me know if I've messed up somewhere! Trying to figure out why it's happening. Will changing the `trainable_variables` filter work? Probably not, because the filter is working properly (see the cell with `nnx.state`; it returns the correct variables).

Just fixed this issue : The core issue was that Functional models were not executing the actual NNX layers during JAX-traced functions. Instead, they were going through the symbolic computation graph (_run_through_graph()), which didn't connect to the real NNX variables.
can you try now?

abheesht17:
Thanks! Working now for sequential and functional models: https://colab.sandbox.google.com/gist/abheesht17/95be01582bc9371dfc059bd80466b851/keras-nnx-model-training-with-custom-nnx-loop.ipynb.

keras/src/backend/config.py
```diff
@@ -230,6 +233,33 @@ def is_flash_attention_enabled():
...
+    global _NNX_ENABLED
+    from keras.src.backend.common import global_state
+
+    _NNX_ENABLED = bool(value)
+    if _NNX_ENABLED:
+        try:
+            from flax import nnx  # noqa F401
+        except ImportError:
+            raise ImportError(
+                "To use the NNX backend, you must install `flax`."
```
mattdangerw: this seems confusingly worded. this isn't a backend, so much a feature for the jax backend right?

keras/src/backend/config.py
```diff
@@ -274,36 +305,18 @@ def keras_home():
...
-        "image_data_format": image_data_format(),
-    }
-    try:
-        with open(_config_path, "w") as f:
-            f.write(json.dumps(_config, indent=4))
-    except IOError:
-        # Except permission denied.
-        pass
-
-# Set backend based on KERAS_BACKEND flag, if applicable.
```
mattdangerw: i don't think you meant to delete this comment?

keras/src/backend/config.py
```diff
@@ -403,3 +417,35 @@ def max_steps_per_epoch():
...
+        # Except permission denied and potential race conditions
+        pass
+
+if not os.path.exists(_config_path):
+    _config_to_save = {
+        "floatx": floatx(),
+        "epsilon": epsilon(),
+        "backend": _BACKEND,  # Use the final _BACKEND value
+        "image_data_format": image_data_format(),
+        "nnx_enabled": _NNX_ENABLED,
```
mattdangerw: why not use `is_nnx_enabled()` here? same for now, and more future proof if more logic ever goes in that function

keras/src/backend/config.py
```diff
@@ -403,3 +417,35 @@ def max_steps_per_epoch():
...
+        os.makedirs(_KERAS_DIR)
+    except OSError:
+        # Except permission denied and potential race conditions
+        pass
+
+if not os.path.exists(_config_path):
+    _config_to_save = {
+        "floatx": floatx(),
+        "epsilon": epsilon(),
+        "backend": _BACKEND,  # Use the final _BACKEND value
```
mattdangerw: why do we need this change? Before, we would save this default keras.json before reading the `KERAS_BACKEND` variable (default is always tensorflow). Now, we will save this default keras.json after reading `KERAS_BACKEND`, so the default might be Jax or Pytorch depending on the first Keras program you run.

This seems ripe to cause some hard to track down bugs, and has nothing to do with the other changes here right? Let's move this block back up so we can preserve the original behavior.

integration_tests/import_test.py
```diff
@@ -11,7 +11,9 @@
         "torch",
         "--extra-index-url https://download.pytorch.org/whl/cpu ",
     ),
-    "jax": ("jax[cpu]", ""),
+    # please update the jax version here if jax version is updated in
+    # requirements file
+    "jax": ("jax[cpu]==0.5.1 flax>=0.10.1", ""),
```
mattdangerw: I think we want two version here. We want to test that Keras can run on Jax without Flax installed at all (currently untested I think).

And we want to test that the basic flow can work with flax installed and nnx enabled (e.g. by setting the environment variable).

mattdangerw (commented):
I haven't dug into the details on the nnx backend implementations. But trying to focus on the basics of Keras operation.

We need to make sure we at least have coverage for a basic training flow with nnx, and coverage for running on the jax backend without nnx installed. I think we can do both of those in the import_test.

keras/src/backend/config.py
divyashreepathihalli: updated

keras/src/backend/config.py
divyashreepathihalli: restored

integration_tests/import_test.py
divyashreepathihalli: cleaned this up. Kept the jax tests independent and separate

keras/src/backend/config.py
divyashreepathihalli: Used is_nnx_enabled() in the config save logic.
Moved the config file save logic to before environment variable reading.

divyashreepathihalli:
FLAX installation is pointing to my branch currently until the CL and PR changes makes it through the Flax Github

.github/workflows/config/nnx/keras.json
```diff
@@ -0,0 +1,7 @@
+{
+    "floatx": "float32",
+    "epsilon": 1e-07,
+    "backend": "jax",
+    "image_data_format": "channels_last",
+    "nnx_enabled": true
```
mattdangerw: maybe add `"nnx_enabled": false` to `config/jax/` for clarity?

.github/workflows/nnx-tests.yml
```diff
@@ -0,0 +1,61 @@
+name: Tests
```
mattdangerw: this should be folded into actions.yml, by extending the matrix there, rather than a separate file. if you think we should try to enable all tests with nnx someday, we could add a todo.

integration_tests/basic_full_flow.py
```diff
@@ -50,5 +50,5 @@ def test_basic_fit(self):
     def test_basic_fit_no_training(self):
         model = MyModel(hidden_dim=2, output_dim=1)
         x = np.random.random((128, 4))
-        model.predict(x)
         model(x)
+        model.predict(x)
```
mattdangerw: why do we need this change? this seems to indicate a bug perhaps

keras/src/backend/config.py
mattdangerw: I don't think is has been? Restored where?

keras/src/backend/config.py
```diff
@@ -261,6 +290,29 @@ def keras_home():
...
+    except OSError:
+        # Except permission denied.
+        pass
+
+if not os.path.exists(_config_path):
+    _config_to_save = {
+        "floatx": _FLOATX,
+        "epsilon": _EPSILON,
+        "backend": _BACKEND,
+        "image_data_format": _IMAGE_DATA_FORMAT,
```
mattdangerw: didn't you want to extend this with `nnx_enabled`? that change is now lost with the move

keras/src/backend/config.py
```diff
@@ -274,36 +326,18 @@ def keras_home():
     _backend = _config.get("backend", _BACKEND)
     _image_data_format = _config.get("image_data_format", image_data_format())
     assert _image_data_format in {"channels_last", "channels_first"}
+    _nnx_enabled_config = _config.get("nnx_enabled", _NNX_ENABLED)
+    if not isinstance(_nnx_enabled_config, bool):
+        _NNX_ENABLED = str(_nnx_enabled_config).lower() == "true"
```
mattdangerw: hmm actually looking at this closer, this would allow people to write the string literal `"true"` in json instead of `true`. I get why we need this for an environment variable where all values are strings, but this seems silly for json. we should expect the json value to be True/False, and proper json

keras/src/backend/jax/core.py
divyashreepathihalli: done

keras/src/layers/layer.py
```diff
@@ -1655,8 +1670,22 @@ def get_config(self):
...
-            self._parent_path = current_path()
+            # Avoid mutating _parent_path during a JAX trace if it's part of
+            # nnx.Object state and the object was created at a different trace
+            # level. We check if we are in NNX mode and if we are in a JAX
+            # trace.
+            if not (is_nnx_enabled() and jax_utils.is_in_jax_tracing_scope()):
+                try:
+                    self._parent_path = current_path()
+                except Exception as e:
+                    warnings.warn(
```
mattdangerw: this is an obtuse warning that just refers to internal stuff. where would expect this code path to be hit, and what's a helpful message to relay to an end user -- what will and will not work for user facing code?

requirements.txt
```diff
@@ -14,7 +14,6 @@ torch-xla==2.6.0;sys_platform != 'darwin'
...
-
+flax>=0.10.1
```
mattdangerw: this seems to run counter to your change above, where you are explicitly installing nnx only for the nnx enabled testing you are running. if you add it here, the `pip install flax` line above does nothing, and we have no coverage for running jax tests without flax installed.

i think you probably want to remove this line.

shell/lint.sh
mattdangerw, line 1: what's up with this? you are adding an empty file? let's remove

divyashreepathihalli: oops sorry!! this was a mistake

.github/workflows/nnx-tests.yml
fchollet: +1

.github/workflows/config/nnx/keras.json
fchollet: +1

integration_tests/basic_full_flow.py
fchollet: Looks like a model building bug? `predict()` should work out of the box because the model should auto-build using the data as reference

keras/src/backend/config.py
fchollet: +1

keras/src/backend/config.py
```diff
@@ -274,36 +326,18 @@ def keras_home():
...
+    else:
+        _NNX_ENABLED = _nnx_enabled_config
 
+    # Apply basic configs that don't cause circular import
     set_floatx(_floatx)
     set_epsilon(_epsilon)
     set_image_data_format(_image_data_format)
     _BACKEND = _backend
 
-# Save config file, if possible.
```
fchollet: Why did you need to move this block?

divyashreepathihalli: to resolve some circular import errors

integration_tests/basic_full_flow.py
divyashreepathihalli: this is now fixed. The error was from https://github.com/keras-team/keras/blob/master/keras/src/backend/jax/trainer.py#L641 model(x) being called from stateless scope inside predict. I updated that for nnx backend.

keras/src/backend/__init__.py
```diff
@@ -39,7 +39,7 @@
     from keras.src.backend.tensorflow.core import Variable as BackendVariable
 elif backend() == "jax":
     from keras.src.backend.jax import *  # noqa: F403
-    from keras.src.backend.jax.core import Variable as BackendVariable
+    from keras.src.backend.jax import Variable as BackendVariable
```
hertschuh: Can you undo this? Let's stay consistent with the other backends. I believe this change is no longer needed.

keras/src/backend/jax/core_test.py
```diff
@@ -0,0 +1,68 @@
...
+
+@pytest.mark.skipif(
+    backend.backend() != "jax",
+    reason="JAX backend specific test for core Variable integration with NNX.",
+)
+@pytest.mark.skipif(
+    not is_nnx_enabled(),
+    reason="Test requires NNX backend to be enabled by default for setup.",
+)
+class JaxCoreVariableTest(testing.TestCase):
```
hertschuh: nit: rename `NnxVariableTest`.

keras/src/ops/function.py
```diff
@@ -102,6 +107,28 @@ def outputs(self):
         """Flat list of the symbolic outputs of the Function."""
         return self._outputs
 
+    def _setup_nnx_op_mapping(self):
+        """Setup operation mapping for NNX"""
+        # Create a mapping from operation id to operation instance
+        self._nnx_op_mapping = {}
+
+        # Store operations as direct attributes for NNX traversal
```
hertschuh: Why is this needed? How does NNX "traverse" operations?

Operations are a Keras concept so I'm really surprised NNX understands them and cares about them.

keras/src/layers/layer.py
```diff
@@ -1655,8 +1670,23 @@ def get_config(self):
...
         if self._parent_path is None:
-            self._parent_path = current_path()
+            # Avoid mutating _parent_path during a JAX trace if it's part of
+            # nnx.Object state and the object was created at a different trace
+            # level. We check if we are in NNX mode and if we are in a JAX
+            # trace.
+            if not (is_nnx_enabled() and jax_utils.is_in_jax_tracing_scope()):
+                try:
+                    self._parent_path = current_path()
+                except Exception:
```
hertschuh: You're not calling `self._parent_path = current_path()` when using NNX and in a tracing scope. Why do you _also_ need to catch the exception?

keras/src/layers/layer.py
```diff
@@ -1542,7 +1545,19 @@ def __setattr__(self, name, value):
...
+            backend.backend() == "jax"
+            and is_nnx_enabled()
+            and (name == "_called" or name == "built")
+        ):
+            object.__setattr__(self, name, value)
+            return
+
+        super().__setattr__(
+            name, value
+        )  # Default path, including for NnxLayer -> nnx.Module
```
hertschuh: nit: instead of having this comment here, which is only relevant for NNX but the code is for every backend, how about removing this comment and instead adding this comment line 1555 or adding to the comment line 1549: `bypass nnx.Module.__setattr__ which cannot be called while tracing`.

keras/src/layers/layer.py
```diff
@@ -52,7 +53,10 @@
 if backend.backend() == "tensorflow":
     from keras.src.backend.tensorflow.layer import TFLayer as BackendLayer
 elif backend.backend() == "jax":
-    from keras.src.backend.jax.layer import JaxLayer as BackendLayer
+    if is_nnx_enabled():
```
hertschuh: Let's undo this change and put the `if` in the `jax` folder instead for consistency with other backends, see my other comment.

keras/src/backend/jax/layer.py
```diff
@@ -1,2 +1,11 @@
+from keras.src.backend.config import is_nnx_enabled
+
+if is_nnx_enabled():
```
hertschuh, line 3: Per my other comment let's do something so that it's always called `JaxLayer`, like it's done for Variable:

```python
if is_nnx_enabled():
    from flax import nnx

    class JaxLayer(nnx.Module):
        pass

else:
    class JaxLayer:
        pass
```

Or

```python
if is_nnx_enabled():
    from flax import nnx

    BaseLayer = nnx.Module
else:
    BaseLayer = object

class JaxLayer(BaseLayer)
    pass
```

keras/src/backend/common/variables.py
```diff
@@ -211,6 +211,12 @@ def __init__(
 
     def _deferred_initialize(self):
         if self._value is not None:
+            # If NNX is enabled, it's possible the variable was already
+            # initialized by a concrete call. In this case,
+            # _deferred_initialize becomes a no-op for this variable.
+            if config.is_nnx_enabled():
+                self._initializer = None  # Clear initializer as it's now "used"
```
hertschuh: This NNX specific code path looks suspicious, I think it's hiding a bug.

The only way that this could happen (from what I can see) is that the value generated in a stateless scope is captured. I'm surprised this can work because it should be a TracedArray instead of a concrete array and should fail later when used in a non-tracing context.

Do you have an example of where it fails without this code?

keras/src/ops/function.py
divyashreepathihalli: NNX doesn't understand Keras operations in the sense of knowing their semantics. it needs to traverse the object structure to build its internal representation for transformations and state management.

divyashreepathihalli: changes to this file was a result of nnx traning loops not tracing the same variables as Keras was.

keras/src/backend/common/variables.py
divyashreepathihalli: NNX is slightly different in this aspect compared to JAX. NNX can manage state and variables during tracing, so it initializes the variables with concrete values during tracing. So after tracing it can be used in concrete state. 
But I removed  this code and everything is still working.

keras/src/backend/common/variables.py
```diff
@@ -211,6 +211,9 @@ def __init__(
 
     def _deferred_initialize(self):
         if self._value is not None:
+            # If NNX is enabled, it's possible the variable was already
+            # initialized by a concrete call. In this case,
+            # _deferred_initialize becomes a no-op for this variable.
```
hertschuh: But this comment is incorrect. It raises an error in this case, so it's not a no-op.

.github/workflows/actions.yml
```diff
@@ -17,11 +20,17 @@ jobs:
...
+        include:
+          - python-version: '3.10'
+            backend: jax
+            nnx_enabled: true
+    name: ${{ matrix.backend == 'jax' && format('Run tests ({0}, {1}, nnx_enabled = {2})', matrix.python-version, matrix.backend, matrix.nnx_enabled) || format('Run tests ({0}, {1})', matrix.python-version, matrix.backend) }}
     runs-on: ubuntu-latest
     env:
       PYTHON: ${{ matrix.python-version }}
       KERAS_HOME: .github/workflows/config/${{ matrix.backend }}
+      KERAS_BACKEND: ${{ matrix.backend }}
```
mattdangerw: why do we need this and the KERAS_HOME variable (which will set backend)? seems redundant

.github/workflows/config/nnx/keras.json
```diff
@@ -0,0 +1,7 @@
+{
```
mattdangerw: where is this used? from the logic above looks like we are setting NNX via env variable and will always read keras.json based on the backend string (jax), so this will not be used?

integration_tests/import_test.py
```diff
@@ -55,16 +56,25 @@ def manage_venv_installs(whl_path):
         "pip install " + backend_extra_url + backend_pkg,
         "pip install -r requirements-common.txt",
         "pip install pytest",
-        # Ensure other backends are uninstalled
```
mattdangerw: i think you could leave this block as is right? and just add the few lines about installing flax afterwards? it will only get hit on the jax backend, seems like the rest of this can stay unaltered

keras/src/backend/config.py
```diff
@@ -261,6 +290,30 @@ def keras_home():
...
+        # Except permission denied.
+        pass
+
+if not os.path.exists(_config_path):
+    _config_to_save = {
+        "floatx": _FLOATX,
+        "epsilon": _EPSILON,
+        "backend": _BACKEND,
+        "image_data_format": _IMAGE_DATA_FORMAT,
+        "nnx_enabled": is_nnx_enabled(),
```
mattdangerw: Use _NNX_ENABLED here for consistency with the rest of this block.

keras/src/backend/config.py
```diff
@@ -274,36 +327,17 @@ def keras_home():
     _backend = _config.get("backend", _BACKEND)
     _image_data_format = _config.get("image_data_format", image_data_format())
     assert _image_data_format in {"channels_last", "channels_first"}
+    _nnx_enabled_config = _config.get("nnx_enabled", _NNX_ENABLED)
+    if isinstance(_nnx_enabled_config, bool):
+        _NNX_ENABLED = _nnx_enabled_config
```
mattdangerw: remove these two lines

keras/src/backend/config.py
```diff
@@ -274,36 +327,17 @@ def keras_home():
...
+    # else: ignore non-bool values for nnx_enabled
 
+    # Apply basic configs that don't cause circular import
     set_floatx(_floatx)
```
mattdangerw, line 309: add `set_nnx_enabled(_nnx_enabled_config)` here

keras/src/backend/config.py
```diff
@@ -403,3 +438,13 @@ def max_steps_per_epoch():
             `None`, no limit is applied.
     """
     return _MAX_STEPS_PER_EPOCH
+
+
+if "KERAS_NNX_ENABLED" in os.environ:
+    env_val = os.environ["KERAS_NNX_ENABLED"].lower()
+    if env_val:
```
mattdangerw: This seems bad, don't we want to check if val.lower() = "true" or "1" here? Otherwise KERAS_NNX_ENABLED=False will enable nnx.

mattdangerw (approved):
Left some final comments on the config and configuration part. All good on my end once fixed.

keras/src/backend/common/variables.py
divyashreepathihalli: updated the comment and also if nnx is enabled, I added a return instead of retuning an error.

keras/src/backend/config.py
divyashreepathihalli: done

.github/workflows/config/nnx/keras.json
divyashreepathihalli: deleted

keras/src/backend/config.py
```diff
@@ -274,8 +303,13 @@ def keras_home():
     _backend = _config.get("backend", _BACKEND)
     _image_data_format = _config.get("image_data_format", image_data_format())
     assert _image_data_format in {"channels_last", "channels_first"}
+    _nnx_enabled_config = _config.get("nnx_enabled", _NNX_ENABLED)
 
+    # else: ignore non-bool values for nnx_enabled
+
+    # Apply basic configs that don't cause circular import
     set_floatx(_floatx)
+    _NNX_ENABLED = _nnx_enabled_config
```
mattdangerw, line 310: can we literally call `set_nnx_enabled(_nnx_enabled_config)` (just going for consistency)

keras/src/backend/config.py
```diff
@@ -274,8 +303,13 @@ def keras_home():
     _backend = _config.get("backend", _BACKEND)
     _image_data_format = _config.get("image_data_format", image_data_format())
     assert _image_data_format in {"channels_last", "channels_first"}
+    _nnx_enabled_config = _config.get("nnx_enabled", _NNX_ENABLED)
 
+    # else: ignore non-bool values for nnx_enabled
```
mattdangerw: delete stale comment here?

mattdangerw (approved):

keras/src/backend/config.py
divyashreepathihalli: this leads to circular import error 
```
    from keras.src.utils import dataset_utils
keras/src/utils/dataset_utils.py:9: in <module>
    from keras.src import tree
keras/src/tree/__init__.py:1: in <module>
    from keras.src.tree.tree_api import assert_same_paths
keras/src/tree/tree_api.py:8: in <module>
    from keras.src.tree import optree_impl as tree_impl
keras/src/tree/optree_impl.py:4: in <module>
    from keras.src.backend.config import backend
E   ImportError: cannot import name 'backend' from partially initialized module 'keras.src.backend.config' (most likely due to a circular import) (/workspaces/keras/keras/src/backend/config.py)
```

So decided to revert it.  2 lines below backend is also set in the same way ` _BACKEND = _backend` - in a way it is still consistent?

keras/src/backend/config.py
divyashreepathihalli: done

PR #21253, opened 2025-05-06
[Keras 3 OpenVINO Backend]: Support numpy.logspace operation #30114

vi-shruti:


keras/src/backend/openvino/numpy.py
```diff
@@ -910,9 +910,151 @@ def less_equal(x1, x2):
 def linspace(
     start, stop, num=50, endpoint=True, retstep=False, dtype=None, axis=0
 ):
-    raise NotImplementedError(
-        "`linspace` is not supported with openvino backend"
-    )
+    """
+    Return evenly spaced numbers over a specified interval.
```
hertschuh, line 914: You have a separate PR for `linspace`, please undo all changes related to `linspace`.

keras/src/backend/openvino/numpy.py
```diff
@@ -993,9 +1135,152 @@ def logical_or(x1, x2):
...
-    raise NotImplementedError(
-        "`logspace` is not supported with openvino backend"
+    """
+    Return numbers spaced evenly on a log scale.
+
+    In linear space, the sequence starts at ``base ** start``
+    (`base` to the power of `start`) and ends with ``base ** stop``
+    (see `endpoint` below).
+
+    Parameters
```
hertschuh, line 1145: Remove, this doc will never be generated or seen, only the one in `keras.ops`. Also, it's not following the format that we use for docstrings.

PR #21256, opened 2025-05-07
Fix `take_along_axis` with TensorFlow with non-int32 indices.

hertschuh:
Change https://github.com/keras-team/keras/pull/21239 broke one use case when the axis dimension is dynamic, the type of the indices is not int32, and the op is run in graph mode.

Note that the additional unit tests don't actually cover this.


fchollet (approved):
Thanks for the fix!

PR #21260, opened 2025-05-07
Several linalg functions support

fbadine:
As part of mlx backend support (#19571), this PR adds support for the following functions in linalg.py:

- lu_factor (supported no in mlx)
- solve_triangular (currently supported in mlx)
- lstsq (implemented the same way as TensorFlow's one was implemented)
- eig (used numpy's implementation until it is supported on mlx)


fchollet (approved):
LGTM - Nice work!

PR #21261, opened 2025-05-07
Initialize random seed for distributed models.

cantonios:
Keep track of a `global_random_seed`, and ensure it is set when initializing `keras.distribution.initialize(...)`.

In multi-host processes in JAX, all processes require consistent random number generation.  Otherwise, initializers on different hosts would produce inconsistent values, resulting in both compilation and runtime failures.


fchollet (commented):
LGTM, thank you

Issue #20953, opened 2025-02-24
Function `openvino.core.custom_gradient()` should be a class


JulianJvn:
I'm not using this, but I saw by accident while browsing the code that `keras.backend.openvino.core.custom_gradient()` is a function that defines nested functions `__init__()` and `__call__()`, but doesn't do anything (in particular, it seems to return `None`):

https://github.com/keras-team/keras/blob/c03ae353f0702387bff1d9a899115c1b9daeca37/keras/src/backend/openvino/core.py#L598-L617

I think line 598 should be

```python
class custom_gradient:
````

Introduced by https://github.com/keras-team/keras/pull/19727/commits/5c401a92677995888de2b2c5d397fb78e6dddf32 in #19727. It was a function throwing a `NotImplementedError` before, but should have been changed to a class.

sonali-kumari1:
Hi @JulianJvn - 
As mentioned by [rkazants](https://github.com/rkazants) in this PR [#21265 ](https://github.com/keras-team/keras/pull/21265), this change would break tests. Please feel free to close this issue if that resolves your concern. Thanks!

PR #21265, opened 2025-05-08
Update core.py

sonali-kumari1:
Replaced function based decorator `openvino.core.custom_gradient()`  with a class based decorator
Fixes [#20953](https://github.com/keras-team/keras/issues/20953)


rkazants:
do NOT merge!!! It breaks tests

sachinprasadhs:
@rkazants , I the existing code is intentional and if there is no changes required to that, can we close this PR then?

PR #21266, opened 2025-05-08
This PR implements the numpy.logspace function for the OpenVINO backend as requested in issue #30114.

sanleo-wq:
- Follows the same pattern as other similar functions in the file
- Passes all tests (63 test cases)

@rkazants for review
All tests are passing:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ examples/**/*.jpg
 .coverage
 *coverage.xml
 .ruff_cache
+venv/
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,3 @@
+[pytest]
+env =
+    KERAS_BACKEND=openvino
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -19,4 +19,5 @@ examples/**/*.jpg
 .python-version
 .coverage
 *coverage.xml
-.ruff_cache
\ No newline at end of file
+.ruff_cache
```
rkazants, line 22: please remove it

pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
```
rkazants, line 1: please remove it

.gitignore
rkazants: not fixed

pytest.ini
rkazants: not addressed comment

PR #21269, opened 2025-05-09
Implementation np.moveaxis function 

sanleo-wq:
issue: #30118

Follows the same pattern as other similar functions in the file
Passes all tests

@rkazants for review
All tests are passing:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ examples/**/*.jpg
 .coverage
 *coverage.xml
 .ruff_cache
+venv/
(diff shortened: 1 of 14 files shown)


.gitignore
```diff
@@ -19,4 +19,5 @@ examples/**/*.jpg
 .python-version
 .coverage
 *coverage.xml
-.ruff_cache
\ No newline at end of file
+.ruff_cache
+venv/
```
rkazants, line 23: remove these changes

pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
```
rkazants, line 1: remove it. it is needed only for developer purpose

fchollet (approved):
Thanks!

PR #21271, opened 2025-05-09
[OpenVINO backend] support split op and modify convert_to_numpy in core to solve split op issue for gsoc project

Mohamed-Ashraf273:
Hi @rkazants ,

Could you please review these operations?
This is the first batch of operations I’m submitting as part of my GSoC project. I’ll be opening additional PRs for the remaining operations in upcoming batches.

Thank you!

### DIFF
--- a/keras/src/backend/openvino/core.py
+++ b/keras/src/backend/openvino/core.py
@@ -450,8 +450,16 @@ def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
 def convert_to_numpy(x):
     if isinstance(x, np.ndarray):
         return x
-    elif isinstance(x, (int, float, list, tuple)):
+    elif isinstance(x, (int, float)):
         return np.array(x)
+    elif isinstance(x, (list, tuple)):
+        x_new = []
+        for elem in x:
+            if isinstance(elem, OpenVINOKerasTensor):
+                x_new.append(convert_to_numpy(elem))
+            else:
+                x_new.append(elem)
+        return np.array(x_new)
     elif np.isscalar(x):
         return x
     elif isinstance(x, ov.Tensor):
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -1305,7 +1305,40 @@ def sort(x, axis=-1):
 
 
 def split(x, indices_or_sections, axis=0):
-    raise NotImplementedError("`split` is not supported with openvino backend")
+    x = get_ov_output(x)
+    dim_at_axis = x.shape[axis]
+    axis_tensor = ov_opset.constant(axis, dtype=Type.i32).output(0)
+    if isinstance(indices_or_sections, int):
+        num_splits = indices_or_sections
+        if dim_at_axis % num_splits != 0:
+            raise ValueError(
+                f"""Cannot split axis of
+                size {dim_at_axis} into {num_splits} equal parts."""
+            )
+        splits = ov_opset.split(x, axis_tensor, num_splits=num_splits)
+        outputs = []
+        for i in range(num_splits):
+            output_tensor = splits.output(i)
+            outputs.append(OpenVINOKerasTensor(output_tensor))
+        return outputs
+    if isinstance(indices_or_sections, (list, tuple, np.ndarray)):
+        indices = list(indices_or_sections)
+        split_lengths = []
+        split_lengths.append(indices[0])
+        for i in range(1, len(indices)):
+            diff = indices[i] - indices[i - 1]
+            split_lengths.append(diff)
+        split_lengths.append(dim_at_axis - indices[-1])
+        length_tensor = ov_opset.constant(split_lengths, dtype=Type.i32)
+        splits = ov_opset.variadic_split(x, axis_tensor, length_tensor)
+        outputs = []
+        for i in range(len(split_lengths)):
+            output_tensor = splits.output(i)
+            outputs.append(OpenVINOKerasTensor(output_tensor))
+        return outputs
+    raise TypeError(
+        f"unsupported type of indices_or_sections: {type(indices_or_sections)}"
+    )
 
 
 def stack(x, axis=0):
--- a/keras/src/backend/openvino/excluded_concrete_tests.txt
+++ b/keras/src/backend/openvino/excluded_concrete_tests.txt
@@ -49,7 +49,6 @@ NumpyDtypeTest::test_round
 NumpyDtypeTest::test_searchsorted
 NumpyDtypeTest::test_signbit
 NumpyDtypeTest::test_sort
-NumpyDtypeTest::test_split
 NumpyDtypeTest::test_sqrt
 NumpyDtypeTest::test_stack_
 NumpyDtypeTest::test_std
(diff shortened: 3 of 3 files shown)


Mohamed-Ashraf273:
Hi @rkazants ,
Could you please review this PR?
Thanks

keras/src/backend/openvino/numpy.py
```diff
@@ -1426,7 +1478,44 @@ def vectorize(pyfunc, *, excluded=None, signature=None):
...
+            x2 = get_ov_output(x2, x1_type)
+    elif isinstance(x2, int):
+        x1 = get_ov_output(x1)
+        x2 = get_ov_output(x2, x1_type)
+    elif isinstance(x1, int):
+        x2 = get_ov_output(x2)
+        x1 = get_ov_output(x1, x2_type)
+    else:
+        x1 = get_ov_output(x1)
+        x2 = get_ov_output(x2)
```
rkazants, line 1516: do we really need this code? Why `x1, x2 = _align_operand_types(x1, x2, "select()")` does not help? The main logic of this helper to align arguments

keras/src/backend/openvino/core.py
```diff
@@ -450,8 +450,16 @@ def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
...
+    elif isinstance(x, (int, float)):
         return np.array(x)
+    elif isinstance(x, (list, tuple)):
+        x_new = []
+        for elem in x:
+            if isinstance(elem, OpenVINOKerasTensor):
+                x_new.append(convert_to_numpy(elem))
+            else:
+                x_new.append(elem)
+        return np.array(x_new)
```
rkazants, line 462: why it is needed?

keras/src/backend/openvino/numpy.py
Mohamed-Ashraf273: Yeah, we need this code,  ```_align_operand_types``` alone isn't sufficient to handle cases where xi is an integer or float. Without this fix, related tests fail.

I could remove it and make a separate commit to demonstrate that the tests fail without it, if you want.

keras/src/backend/openvino/core.py
Mohamed-Ashraf273: This change handles cases where a list contains ```OpenVINOKerasTensor``` elements. I first encountered the issue while running ```split``` tests, the tests took an unusually long time and never completed. After investigating, I found that the issue was caused by split returning a list of ```OpenVINOKerasTensor``` objects, and ```np.array(x)``` struggled to convert them. RAM usage kept increasing, and although I’m not sure whether it would have eventually finished, it was taking far too long.

Mohamed-Ashraf273: image
Look tests stop here and take too long and never finish.

keras/src/backend/openvino/numpy.py
Mohamed-Ashraf273: image
image

keras/src/backend/openvino/core.py
Mohamed-Ashraf273: @rkazants 
what do you recommend me to do?

keras/src/backend/openvino/numpy.py
```diff
@@ -1246,9 +1262,10 @@ def real(x):
 
 
 def reciprocal(x):
-    raise NotImplementedError(
-        "`reciprocal` is not supported with openvino backend"
-    )
+    x = get_ov_output(x)
+    one_constant = ov_opset.constant(1, dtype=x.get_element_type()).output(0)
+    x = ov_opset.divide(one_constant, x).output(0)
+    return OpenVINOKerasTensor(x)
```
rkazants, line 1268: now please move this op implementation to separate PR. So we will get it merged fastly.

keras/src/backend/openvino/numpy.py
```diff
@@ -1103,10 +1103,26 @@ def mod(x1, x2):
...
+    destination = [(d + rank if d < 0 else d) for d in destination]
+
+    axes = list(range(rank))
+    for s, d in zip(source, destination):
+        axes.remove(s)
+        axes.insert(d, s)
+
+    perm_const = ov_opset.constant(axes, Type.i32).output(0)
+    return OpenVINOKerasTensor(ov_opset.transpose(x, perm_const).output(0))
+
```
rkazants, line 1125: this operation is implemented in this PR https://github.com/keras-team/keras/pull/21269
I asked to merge it. So you can remove it.

keras/src/backend/openvino/numpy.py
rkazants: Need to make the alignment logic more clear. Now it looks a bit dirty. Need to think.

Mohamed-Ashraf273: @rkazants 
I tried to make ir more clear, can you take a look?

Mohamed-Ashraf273:
@rkazants
I tried to make ```where``` op to be more clear, can you take a look?

keras/src/backend/openvino/core.py
Mohamed-Ashraf273: @rkazants 
What about this ```convert_to_numpy``` should I remove the changes?
The function will still work, but the tests won't complete.

Mohamed-Ashraf273:
Hi @rkazants 
Could you please let me know the next steps for this PR?
Thank you!

keras/src/backend/openvino/numpy.py
```diff
@@ -1305,10 +1305,45 @@ def sort(x, axis=-1):
...
+        length_tensor = ov_opset.constant(split_lengths, dtype=np.int64)
+        splits = ov_opset.variadic_split(x, axis_tensor, length_tensor)
+        outputs = []
+        for i in range(len(split_lengths)):
+            output_tensor = splits.output(i)
+            outputs.append(OpenVINOKerasTensor(output_tensor))
+        return outputs
+    raise TypeError(
+        f"unsupported type of indices_or_sections: {type(indices_or_sections)}"
+    )
```
rkazants, line 1341: please create separate PR for each new operation support. Now I don't understand for what op changes in `convert_to_numpy` is needed.

keras/src/backend/openvino/core.py
rkazants: Please share a simple reproducer on Keras that requires changes in `convert_to_numpy`.

keras/src/backend/openvino/numpy.py
Mohamed-Ashraf273: Hi @rkazants 
I removed ```where``` to another PR, the ```slpit``` op is the operation needs the changes iin ```conver_to_nump```.

keras/src/backend/openvino/core.py
Mohamed-Ashraf273: issue
```python
from keras import ops
import numpy as np
from datetime import datetime

ov_tensor = ops.convert_to_tensor(np.array([1, 2, 3]))

tensor_list = [ov_tensor for _ in range(3)]

start_time = datetime.now()
print("Start:", start_time.strftime("%I:%M:%S %p"))

# This is where the problem is
# convert list of OpenVINOKerasTensor to a numpy array (this hangs)
_ = np.array(tensor_list)

# Print time after (won't reach this if it hangs)
end_time = datetime.now()
print("End:", end_time.strftime("%I:%M:%S %p"))
```
The issue is with ```np.array(x)``` where x is a list of ```OpenVINOKerasTensor```, that call hangs indefinitely or takes way too long.
So here's a minimal reproducer showing that np.array() hangs and took about 3 mins when passed a list of ```OpenVINOKerasTensor``` objects. This justifies the need to fix convert_to_numpy().

PR #21273, opened 2025-05-11
Implement hamming function in keras.ops

shashaka:
Adds keras.ops.hamming, which computes the Hamming window. Supported across NumPy, TensorFlow, PyTorch, and JAX backends, but not supported on OpenVINO.


fchollet (approved):
Neat, thanks for the op addition! LGTM

PR #21280, opened 2025-05-13
`mlx` - api patch for tests

acsweet:
This is related to #19571 

@fchollet my bad! 
I was behind one version on `namex` and it generated bad api files


fchollet (approved):
Nice work, thank you for the fix!

Do you know why compilation for MLX seems to be failing on CI? Is it specifically *our* CI or is it an issue with the CPU/Unix version of MLX in general?

acsweet:
I believe it's an issue with the CPU/Unix version of MLX, we've been trying to work through those. 

There are a few outstanding issues for the CPU build, I'll push on those a bit and ping the Apple team where we're at. See if we can get the CI tests passing. 

I'll let you know if we get stuck too!

PR #21281, opened 2025-05-13
[OpenVINO Backend] support reciprocal

Mohamed-Ashraf273:
@rkazants 
I removed the reciprocal to another PR.


rkazants:

PR #21282, opened 2025-05-13
Implement numpy.prod method for OpenVino backend

sanleo-wq:
issue: #30212

Follows the same pattern as other similar functions in the file
Passes all tests

@rkazants for review
All tests are passing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,5 @@ examples/**/*.jpg
 .python-version
 .coverage
 *coverage.xml
-.ruff_cache
+.ruff_cachekeras_venv/
+keras_venv/
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -1230,7 +1230,31 @@ def pad(x, pad_width, mode="constant", constant_values=None):
 
 
 def prod(x, axis=None, keepdims=False, dtype=None):
-    raise NotImplementedError("`prod` is not supported with openvino backend")
+    x = get_ov_output(x)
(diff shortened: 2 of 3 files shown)


keras/src/backend/openvino/numpy.py
```diff
@@ -1230,7 +1230,31 @@ def pad(x, pad_width, mode="constant", constant_values=None):
...
+        axis = 0
+
+    if isinstance(axis, tuple):
+        axis = list(axis)
+
+    axis_const = ov_opset.constant(axis, Type.i32).output(0)
+    result = ov_opset.reduce_prod(x, axis_const, keepdims).output(0)
+
+    if dtype is None and original_type != result.get_element_type():
+        result = ov_opset.convert(result, original_type).output(0)
```
rkazants, line 1255: according to the logic, if dtype is None, original_type is always equal to result's one. Please correct me.
That is because reshape and reduce_prod preserve tensor type:)

.gitignore
```diff
@@ -19,4 +19,4 @@ examples/**/*.jpg
 .python-version
 .coverage
 *coverage.xml
-.ruff_cache
\ No newline at end of file
+.ruff_cachekeras_venv/
```
rkazants, line 22: please remove changes here.
Switch on tests for prod.

hertschuh:
What's the status of this PR? There's also:
- https://github.com/keras-team/keras/pull/21403
- https://github.com/keras-team/keras/pull/21405

hertschuh:
This was implemented by https://github.com/keras-team/keras/pull/21567

PR #21293, opened 2025-05-16
update JAX GPU version

sachinprasadhs:
Update JAX GPU version and fix failing GPU tests

### DIFF
--- a/.kokoro/github/ubuntu/gpu/build.sh
+++ b/.kokoro/github/ubuntu/gpu/build.sh
@@ -13,6 +13,7 @@ source venv/bin/activate
 python --version
 python3 --version
 
+# setting the LD_LIBRARY_PATH manually is causing segmentation fault
 #export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/cuda/lib64:/usr/local/cuda/extras/CUPTI/lib64:"
 # Check cuda
 nvidia-smi
--- a/.kokoro/github/ubuntu/gpu/build.sh
+++ b/.kokoro/github/ubuntu/gpu/build.sh
@@ -13,7 +13,7 @@ source venv/bin/activate
 python --version
 python3 --version
 
-export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/cuda/lib64:/usr/local/cuda/extras/CUPTI/lib64:"
+#export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/cuda/lib64:/usr/local/cuda/extras/CUPTI/lib64:"
 # Check cuda
 nvidia-smi
 nvcc --version
--- a/requirements-jax-cuda.txt
+++ b/requirements-jax-cuda.txt
@@ -8,7 +8,7 @@ torch==2.6.0
 
 # Jax with cuda support.
 --find-links https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
-jax[cuda12]==0.4.28
+jax[cuda12]==0.6.0
 flax
 
 -r requirements-common.txt
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,7 +13,7 @@ torch-xla==2.6.0;sys_platform != 'darwin'
 # Jax.
 # Pinned to 0.5.0 on CPU. JAX 0.5.1 requires Tensorflow 2.19 for saved_model_test.
 # Note that we test against the latest JAX on GPU.
-jax[cpu]==0.5.0
+jax[cpu]==0.6.0
 flax
 
 # Common deps.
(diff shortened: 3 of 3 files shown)


sachinprasadhs:
@hertschuh , I replicated the similar changes here [#2252]( https://github.com/keras-team/keras-hub/pull/2252) and JAX GPU is working.
Additionally when I try to update tensorflow version as well, I'm getting the MemoryError, not sure if it is the Tensorflow package issue or build script issue here. 

In my local instance  when I tried to build the package by updating the tensorflow version(2.19.0) with Python 3.11 version, I'm able to generate build package .whl file.
Here is the build error for my tensorflow 2.19 commit:
https://github.com/keras-team/keras/actions/runs/15076576568/job/42385286142

@pctablet505 , do you have any suggestions here.

hertschuh:
> @hertschuh , I replicated the similar changes here [#2252](https://github.com/keras-team/keras-hub/pull/2252) and JAX GPU is working. Additionally when I try to update tensorflow version as well, I'm getting the MemoryError, not sure if it is the Tensorflow package issue or build script issue here.
> 
> In my local instance when I tried to build the package by updating the tensorflow version(2.19.0) with Python 3.11 version, I'm able to generate build package .whl file. Here is the build error for my tensorflow 2.19 commit: https://github.com/keras-team/keras/actions/runs/15076576568/job/42385286142
> 
> @pctablet505 , do you have any suggestions here.

@sachinprasadhs 

I tried this in https://github.com/keras-team/keras/pull/21026 and it doesn't work.

I spend quite a bit of time and hit a roadblock (there might be others):
JAX >= 0.5.1 requires Tensorflow 2.19 (at least for the saved_model export / reload tests)
Tensorflow 2.19 requires protobuf >= 4.21.6 (with lower versions, it crashes)
tf2onnx requires protobuf ~= 3.20 context on this here https://github.com/onnx/tensorflow-onnx/issues/2261 they want to support old TF versions. tf2onnx may work ok with more recent version but they claim some other dependencies need 3.20.
The trouble is that the `requirement.txt` format doesn't allow overrides to solve conflicts or targeted `--no-deps` options.

Possible solutions:

Force tf2onnx to run with protobuf 4.21.6. That means that we have to remove tf2onnx from requirements.txt and do some extra "manual" pip installs in the CI after the requirements.txt. We also have to document that for ONNX users.
Run export tests (ONNX but also jax2tf potentially) separately with a custom requirements-??.txt . We might even need multiple of those. Export tests have generally been a difficult because that's where JAX meets TF, Torch meets TF, etc. and the versions have to be compatible. Then there is the question of whether we also run those on cuda, requiring even more separate requirements-??-cuda.txt.

hertschuh:
@sachinprasadhs 

Oh, so the MemoryError you're facing is because you're running Tensorflow 2.19 with the wrong version of protobuf (3.20 instead of >= 4.21.6.

sachinprasadhs:
Well that is bit complicated then!
Not sure, with protobuf 3.20 in Python 3.11 environment, there was no build issue.  

I didn't see your PR, in that case this PR is passing all the tests and the failing JAX test is not related to this PR.

To solve the core dump issue in JAX GPU and continue running JAX GPU tests, may be we can have this merged and then you can probably think of any workaround for updating Tensorflow.

.kokoro/github/ubuntu/gpu/build.sh
```diff
@@ -13,7 +13,7 @@ source venv/bin/activate
 python --version
 python3 --version
 
-export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/cuda/lib64:/usr/local/cuda/extras/CUPTI/lib64:"
+#export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/cuda/lib64:/usr/local/cuda/extras/CUPTI/lib64:"
```
hertschuh, line 16: Oh, so this is what was causing the errors on JAX GPU?

hertschuh: If so, remove this line or add a comment explaining

sachinprasadhs: Yes, detailed discussion here https://github.com/keras-team/keras-hub/pull/2252

sachinprasadhs: added comment

james77777778:
Hey @sachinprasadhs @hertschuh
I think tf2onnx is not actively maintained (the last release was about 1 year ago).

Maybe we can test the export functionality only on CPU and provide guidance for users who try to export the model using JAX GPU. (e.g.  JAX_PLATFORMS=cpu)

hertschuh:
> Hey @sachinprasadhs @hertschuh
> I think tf2onnx is not actively maintained (the last release was about 1 year ago).
> 
> Maybe we can test the export functionality only on CPU and provide guidance for users who try to export the model using JAX GPU. (e.g.  JAX_PLATFORMS=cpu)

Sure, let's do that. In fact, I'd like to upgrade JAX and TF for cpu too, so maybe we can have a separate setup for onnx.

fchollet (approved):

PR #21296, opened 2025-05-17
[OpenVINO Backend] support ops.split

Mohamed-Ashraf273:
Hi @rkazants 

I've supported ```ops.split```, which I need for my GSoC project.
Could you please review it? Thanks!

While working on it, I found an issue with ```convert_to_numpy```, it doesn’t handle a list of ```OpenVINOKerasTensor```. I included a simple reproducer to demonstrate the problem.

issue
```python
from keras import ops
import numpy as np
from datetime import datetime

ov_tensor = ops.convert_to_tensor(np.array([1, 2, 3]))

tensor_list = [ov_tensor for _ in range(3)]

start_time = datetime.now()
print("Start:", start_time.strftime("%I:%M:%S %p"))

# This is where the problem is
# convert list of OpenVINOKerasTensor to a numpy array (this hangs)
_ = ops.convert_to_numpy(tensor_list)

# Print time after (won't reach this if it hangs)
end_time = datetime.now()
print("End:", end_time.strftime("%I:%M:%S %p"))
```
The issue is with ```np.array(x)``` in ```convert_to_numpy``` where x is a list of ```OpenVINOKerasTensor```, which is being returned by ```split``` op because ```split``` returns lists of ```OpenVINOKerasTensor```, that call hangs indefinitely or takes way too long.
So here's a minimal reproducer showing that ```convert_to_numpy()``` hangs and took about 3 mins when passed a list of ```OpenVINOKerasTensor``` objects. This justifies the need to fix convert_to_numpy().

I resolved the issue by first checking if any element in the list is an ```OpenVINOKerasTensor```. If so, I convert those elements to ```NumPy``` before converting the entire list.

### DIFF
--- a/keras/src/backend/openvino/core.py
+++ b/keras/src/backend/openvino/core.py
@@ -450,8 +450,13 @@ def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
 def convert_to_numpy(x):
     if isinstance(x, np.ndarray):
         return x
-    elif isinstance(x, (int, float, list, tuple)):
+    elif isinstance(x, (int, float)):
         return np.array(x)
+    elif isinstance(x, (list, tuple)):
+        x_new = []
+        for elem in x:
+            x_new.append(convert_to_numpy(elem))
+        return np.array(x_new)
     elif np.isscalar(x):
         return x
     elif isinstance(x, ov.Tensor):
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -1326,7 +1326,49 @@ def sort(x, axis=-1):
 
 
 def split(x, indices_or_sections, axis=0):
-    raise NotImplementedError("`split` is not supported with openvino backend")
+    x = get_ov_output(x)
+    axis_tensor = ov_opset.constant(axis, dtype=Type.i32).output(0)
+
+    shape_tensor = ov_opset.shape_of(x)
+    axis_i32 = ov_opset.constant([axis], dtype=Type.i32)
+    dim_at_axis_tensor = ov_opset.gather(
+        shape_tensor, axis_i32, ov_opset.constant(0, dtype=Type.i32)
+    )
+
+    if isinstance(indices_or_sections, int):
+        num_splits = indices_or_sections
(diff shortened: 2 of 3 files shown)


keras/src/backend/openvino/core.py
```diff
@@ -450,8 +450,16 @@ def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
...
-    elif isinstance(x, (int, float, list, tuple)):
+    elif isinstance(x, (int, float)):
         return np.array(x)
+    elif isinstance(x, (list, tuple)):
+        x_new = []
+        for elem in x:
+            if isinstance(elem, OpenVINOKerasTensor):
+                x_new.append(convert_to_numpy(elem))
+            else:
+                x_new.append(elem)
```
rkazants, line 461: I think we should just recursively handle elements:
```
        x_new = []
        for elem in x:
            x_new.append(convert_to_numpy(elem))
        return np.array(x_new)
```
because this way we can handle different types: KerasVariable, ov.Tensor, etc.

keras/src/backend/openvino/numpy.py
```diff
@@ -1326,7 +1326,40 @@ def sort(x, axis=-1):
 
 
 def split(x, indices_or_sections, axis=0):
-    raise NotImplementedError("`split` is not supported with openvino backend")
+    x = get_ov_output(x)
+    dim_at_axis = x.shape[axis]
```
rkazants, line 1330: shape can be with dynamic dimensions. So please use ShapeOf and `Gather` operations from OV opset to retrieve dimension size.

keras/src/backend/openvino/core.py
Mohamed-Ashraf273: yeah, that's better. I added it to the code.

keras/src/backend/openvino/numpy.py
Mohamed-Ashraf273: Thank you for the clarification. I've implemented it using ```ShapeOf``` and ```Gather``` as suggested.

rkazants (commented):

fchollet (approved):

PR #21297, opened 2025-05-17
[OpenVINO Backend] update ops.stack to support tuples

Mohamed-Ashraf273:
Hi @rkazants 

I’ve updated ```ops.stack``` to support tuples, which is needed for my GSoC project to enable pipelines.
Could you please review the changes when you have a moment?

Thank you!

### DIFF
--- a/keras/src/backend/openvino/excluded_concrete_tests.txt
+++ b/keras/src/backend/openvino/excluded_concrete_tests.txt
@@ -51,7 +51,6 @@ NumpyDtypeTest::test_signbit
 NumpyDtypeTest::test_sort
 NumpyDtypeTest::test_split
 NumpyDtypeTest::test_sqrt
-NumpyDtypeTest::test_stack_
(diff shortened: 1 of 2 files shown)


rkazants (commented):
could you please remove exclusion `NumpyDtypeTest::test_stack_` in https://github.com/keras-team/keras/blob/master/keras/src/backend/openvino/excluded_concrete_tests.txt

other part looks good

Mohamed-Ashraf273:
> could you please remove exclusion `NumpyDtypeTest::test_stack_` in https://github.com/keras-team/keras/blob/master/keras/src/backend/openvino/excluded_concrete_tests.txt
> 
> other part looks good

Some tests initially failed due to ```dtype``` mismatches after the changes, but I’ve resolved the issue and now all tests are passing successfully.

rkazants (commented):

fchollet (approved):
Thanks!

PR #21302, opened 2025-05-18
Revert "Add sparse support to `ops.ones_like` and `ops.zeros_like`."

hertschuh:
Arguably, the result of `ones_like` and `zeros_like` should not be sparse.


hertschuh:
@cantonios

fchollet (approved):

PR #21303, opened 2025-05-19
Implement kaiser function in keras.ops

shashaka:
Adds keras.ops.kaiser, which computes the Kaiser window. Supported across NumPy, TensorFlow, PyTorch, and JAX backends, but not supported on OpenVINO.


fchollet (approved):
Nice work, thank you! LGTM

PR #21307, opened 2025-05-20
Bump up minimum version of python to 3.10.

cantonios:
The latest versions of JAX (0.6.0) and TensorFlow (2.20) no longer support Python 3.9.  They do support 3.12.


cantonios:
I haven't checked that _all_ dependencies support 3.12, but at least TF/JAX/PyTorch do.  Most probably should these days.

cantonios:
@hertschuh @fchollet

PR #21321, opened 2025-05-23
`api_gen` now excludes backend code.

hertschuh:
This:
- Allows development (`api_gen` / git presubmit hooks) without all backends installed and working. For instance, OpenVino currently doesn't import on MacOS Sequoia, this allows running `api_gen` regardless.
- Makes sure we don't accidentally create and honor exports that are backend specific.


hertschuh:
In draft for now, requires https://github.com/fchollet/namex/pull/7

fchollet (approved):

Issue #21313, opened 2025-05-21
Special optimization for torch backend


pass-lin:
PyTorch provides specialized cudnn operators for many common implementations, such as F.group_norm

In the past speed test benchmarks of the keras-team, we can also find that the torch backend of keras is usually slower than the native torch backend. This may be related to the fact that many implementations in keras do not use the cudnn operators provided by torch.

Is it necessary to make some special optimizations for the torch backend for the existing layer implementation to improve the inference performance of the torch backend.

If necessary, we can submit a PR.

divyashreepathihalli:
@pass-lin Thank you!! Please feel free to submit your contribution for these optimizations! Thank you!!

PR #21325, opened 2025-05-24
improve RMSln preformance when torch backend

pass-lin:
from https://github.com/keras-team/keras/issues/21313 and https://github.com/keras-team/keras/pull/21323
@fchollet Please check if this form of optimization is reasonable.


fchollet (approved):
LGTM, thanks for the update.

PR #21327, opened 2025-05-27
add ops.layer_normalization

pass-lin:
from https://github.com/keras-team/keras/issues/21313 and https://github.com/keras-team/keras/pull/21323
To improve the performance of the torch backend, we add a new op and use the cudnn operator of the torch in this op


fchollet (approved):
LGTM, thanks for the refactor.

PR #21329, opened 2025-05-27
Revert "Fixed issue with dot_product_attention when using TPU. "

pctablet505:
Reverts keras-team/keras#21254


fchollet (approved):
What's the issue?

PR #21332, opened 2025-05-28
fix csv logger bug

divyashreepathihalli:
fixes the issue about `CSVLogger is setting val_* metrics to nan despite no validation data being provided`: https://github.com/keras-team/keras/issues/21173

Colab to verify : https://colab.sandbox.google.com/drive/1tZmeU2x_DpB6q5hBtsg2pGTR4KhqK2_7#scrollTo=RtemKVh36rI6

The values to be matched between progbar and cvs logger is fixed in https://github.com/keras-team/keras/pull/21331

### DIFF
--- a/keras/src/callbacks/csv_logger.py
+++ b/keras/src/callbacks/csv_logger.py
@@ -37,6 +37,7 @@ class CSVLogger(Callback):
         self.writer = None
         self.keys = None
         self.append_header = True
+        self.csv_file = None
 
     def on_train_begin(self, logs=None):
         if self.append:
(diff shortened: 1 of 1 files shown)


keras/src/callbacks/csv_logger.py
```diff
@@ -65,22 +72,20 @@ def handle_value(k):
...
-            val_keys_found = False
-            for key in self.keys:
-                if key.startswith("val_"):
-                    val_keys_found = True
-                    break
-            if not val_keys_found:
-                self.keys.extend(["val_" + k for k in self.keys])
+
+            # Check if the model is expected to produce validation metrics.
+            if hasattr(self.model, "metrics_names"):
```
hertschuh, line 77: nitpick: you don't need `if hasattr(self.model, "metrics_names"):`, it's a property on the `Model` class:

https://github.com/keras-team/keras/blob/master/keras/src/trainers/trainer.py#L271

PR #21340, opened 2025-06-02
Fix comment

dayo09:
Let's fix comment in AveragePooling2D.


keras/src/layers/pooling/average_pooling2d.py
```diff
@@ -17,7 +17,7 @@ class AveragePooling2D(BasePooling):
     (when `input_shape >= pool_size`)
 
     The resulting output shape when using the `"same"` padding option is:
-    `output_shape = math.floor((input_shape - 1) / strides) + 1`
```
dayo09, line 20: This comment is about `valid` padding. Let's fix this to represent `same` padding.

fchollet (approved):
LGTM

Issue #21328, opened 2025-05-27
Type of model.input is KerasTensor in 3.9 and list in 3.10


mmtrebuchet:
This seems to be a breaking change and I don't see it documented in the release notes. 

If you save a model and then load it, the type of model.input was KerasTensor in 3.9.2 (and as far back as I can remember), but it's now list in 3.10.0. If you *don't* save and then re-load the model, the input type is KerasTensor, so I think this qualifies as a bug.

Here is a minimal example:

```
#!/usr/bin/env python3

import keras as k
import keras.layers as kl
from keras import models as km
def runTest():
    il = kl.Input((1024, 4))
    cl = kl.Conv1D(filters=5, kernel_size=10, padding="valid", activation="relu")(il)
    m = km.Model(inputs=il, outputs=[cl])
    print(m.input.shape)
    m.save("test.keras")

def runLoad():
    m = km.load_model("test.keras")
    print(type(m.input))
    print(m.input.shape)

runTest()
runLoad()
```

Under Keras 3.9.2, I get the following: 

```
(None, 1024, 4)
<class 'keras.src.backend.common.keras_tensor.KerasTensor'>
(None, 1024, 4)
```

Whereas with 3.10.0, I get:

```
(None, 1024, 4)
<class 'list'>
Traceback (most recent call last):
  File "/dev/shm/./testModel.py", line 22, in <module>
    runLoad()
  File "/dev/shm/./testModel.py", line 16, in runLoad
    print(m.input.shape)
          ^^^^^^^^^^^^^
AttributeError: 'list' object has no attribute 'shape'
```

I'm using the Tensorflow backend (TF 2.19.0 in both environments I'm using).

dhantule:
Hi @mmtrebuchet, thanks for reporting this. 

I've tried to run this code and I faced the same error with keras version `3.10.0` in this colab [gist](https://colab.sandbox.google.com/gist/dhantule/0be432615d3e6fa6be00ddebdb58d60d/type-of-model-input-is-kerastensor-in-3-9-and-list-in-3-10-21328.ipynb). We'll look into this and update you.

nithin9000:
Hey can i start working on this?

divyashreepathihalli:
@mmtrebuchet Thank you for reporting this. this is Indeed a breaking change. The change is reverted here - https://github.com/keras-team/keras/pull/21336

monicadsong:
Hi @mmtrebuchet -- 

[...]

PR #21341, opened 2025-06-02
Ensures consistent serialization for Functional model

JyotinderSingh:
Related issues: #21328
Previous fix: #20993

## Original Bug
- Models created with nested list inputs (e.g., `inputs=[[input_a]]`) would have their structure flattened after a serialization round-trip.
- A previous fix for this issue removed an un-nesting step during deserialization. While it fixed the nested-input case, it introduced a regression where a standard single-input model (`inputs=input_a`) would have its `model.input` type change from `KerasTensor` to list upon loading.

The root cause was that the serialization format was ambiguous. It represented both a single tensor and a list containing one tensor identically.

## Fix
This change makes the serialization format unambiguous. The logic in `get_config` is updated to:
1. Serialize a single `KerasTensor` input/output as a direct pointer.
2. Serialize a list, tuple, or dict input/output as a corresponding structure containing pointers.
3. The deserializer can now perfectly reconstruct the original I/O structure without naive un-nesting logic, as the format is no longer lossy.

### DIFF
--- a/keras/src/models/functional.py
+++ b/keras/src/models/functional.py
@@ -452,11 +452,11 @@ class Functional(Function, Model):
             new_node_index = node_reindexing_map[node_key]
             return [operation.name, new_node_index, tensor_index]
 
-        def create_io_structure(tensors):
+        def map_tensors(tensors):
             return tree.map_structure(get_tensor_config, tensors)
 
-        config["input_layers"] = create_io_structure(self._inputs_struct)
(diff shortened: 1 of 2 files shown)


divyashreepathihalli (approved):

hertschuh (approved):

hertschuh:
Thanks for the fix!

PR #21362, opened 2025-06-06
[OpenVINO Backend] support ops.slice_update

Mohamed-Ashraf273:
Hi @rkazants ,
I've supported `ops,silce_update` for my GSoC project, but I can't enable tests for it until getting `__getitem__` merge.

It could be implemented more easily if we use `opset15` instead of `opset14`.

### DIFF
--- a/keras/src/backend/openvino/core.py
+++ b/keras/src/backend/openvino/core.py
@@ -843,6 +843,22 @@ def slice_update(inputs, start_indices, updates):
 
     updates_shape = ov_opset.shape_of(updates_tensor, Type.i32).output(0)
     rank = updates_tensor.get_partial_shape().rank.get_length()
+    if rank == 0:
+        # Handle scalar update
+        start_tensor = ov_opset.concat(processed_start_indices, axis=0).output(
+            0
+        )
+        # For scatter_nd_update,
+        # indices should be of shape [num_updates, rank_of_inputs]
+        # and updates should be of shape [num_updates]. Here num_updates is 1.
+        absolute_indices = ov_opset.unsqueeze(start_tensor, zero_scalar).output(
+            0
+        )
+        updates_flat = ov_opset.unsqueeze(updates_tensor, zero_scalar).output(0)
+        result = ov_opset.scatter_nd_update(
+            inputs, absolute_indices, updates_flat
+        ).output(0)
+        return OpenVINOKerasTensor(result)
 
     # Compute the total number of elements in the updates tensor.
     # Example:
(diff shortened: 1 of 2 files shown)


rkazants:
let us wait until `slice_update` tests are switched on

Mohamed-Ashraf273:
@rkazants 
I've switched on tests for `slice_update`.

keras/src/backend/openvino/core.py
```diff
@@ -810,9 +810,83 @@ def prepare_slice_index(val):
 
 
 def slice_update(inputs, start_indices, updates):
```
rkazants, line 817: please add comments for each block

Mohamed-Ashraf273:
@rkazants 
I can't change it to `opest15` because I'm facing a lot of `segfaults` when compiling models. I tried the version `2024.5.0` as mentioned here: https://docs.openvino.ai/2025/documentation/openvino-ir-format/operation-sets/available-opsets.html, but it didn't work either.
So I added comments.

Mohamed-Ashraf273:
@rkazants 
I've added some detailed comments :)

Mohamed-Ashraf273:
@rkazants

Mohamed-Ashraf273:
@fchollet 
@rkazants 
I tried to make it one slice instead of loop in this part:
```python

for dim in range(rank):
    dim_size = ov_opset.gather(
        updates_shape, ov_opset.constant([dim], Type.i32), zero_scalar
    ).output(0)
    dim_size_scalar = ov_opset.squeeze(dim_size, zero_tensor).output(0)
    dim_sizes.append(dim_size_scalar)

    # Strides to convert a flat index into a multi-dimensional index.
    # This allows us to map each element in the flattened updates tensor
    # to its correct N-dimensional position, so we can compute the absolute
    # index in the input tensor for the scatter update.
    # Stride for a dimension is the product of all dimensions after it.
    # For the last dimension, stride is 1.
    # Example:
    # For a 3D tensor with shape [2, 3, 4]:
    #   - stride for dim=0 (first axis) is 3*4=12
    #     (to move to the next "block" along axis 0)
    #   - stride for dim=1 is 4 (to move to the next row along axis 1)
    #   - stride for dim=2 is 1 (to move to the next element along axis 2)
    # This is equivalent to how numpy flattens multi-dimensional arrays.
    if dim < rank - 1:
        remaining_dims = ov_opset.slice(
            updates_shape,
            ov_opset.constant([dim + 1], Type.i32),
            ov_opset.constant([rank], Type.i32),
            one_tensor,
            zero_tensor,
        ).output(0)
        stride = ov_opset.reduce_prod(
            remaining_dims, zero_tensor, keep_dims=False
        ).output(0)
    else:
        stride = one_scalar
    strides.append(stride)
```
but it took more memory usage, so I think this PR is Ready too.

Mohamed-Ashraf273:
@fchollet

Mohamed-Ashraf273:
@fchollet 
Could you please take a look at this PR?
Thanks!

PR #21365, opened 2025-06-09
Add support for all dict methods to `ShardedH5IOStroe`.

james77777778:
Currently, `len`, `keys`, `items`, `values`, `in` and `del` may raise erros in sharding mode.

This PR adds `self.current_shard_filenames` to address these issues.
The original sharding test has been updated to a more rigorous one.

cc @divyashreepathihalli @laxmareddyp

### DIFF
--- a/keras/src/saving/saving_lib.py
+++ b/keras/src/saving/saving_lib.py
@@ -1040,15 +1040,20 @@ class H5IOStore:
         # will mistakenly using `__len__` to determine the value.
         return self.h5_file.__bool__()
 
-    def _get_h5_file(self, path_or_io):
+    def _get_h5_file(self, path_or_io, mode=None):
+        mode = mode or self.mode
+        if mode not in ("r", "w", "a"):
+            raise ValueError(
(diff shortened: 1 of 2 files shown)


keras/src/saving/saving_lib.py
```diff
@@ -1202,7 +1209,7 @@ def __init__(self, path_or_io, max_shard_size=5, archive=None, mode="r"):
         self.archive = archive
         self.io_file = None
 
-        self.max_shard_size = float(max_shard_size)
+        self.max_shard_size = float(max_shard_size) * 1024**3  # To bytes.
```
fchollet, line 1216: Is this a behavior change?

fchollet (commented):
Thanks for the PR!

keras/src/saving/saving_lib.py
james77777778: No. I simplified the size computation by using a unified unit: bytes.
We don't need to add conversion between bytes and gigabytes now.

fchollet (approved):
LGTM, thank you!

PR #21367, opened 2025-06-09
Move `_dtype_policy` from `Operation` to `Layer`.

hertschuh:
- The `dtype` in `Operation.__init__` was used to populate `self._dtype_policy`, but was not used by `Operation` itself, nor by any of the subclasses besides `Layer`.
- `Operation`s never apply the dtype policy.
- The serialization of `self._dtype_policy` is already implemented in `Layer`, and in a different way than in `Operation`.
- Many ops (e.g. `cast`) have a `dtype` parameter in their `__init__` which has different semantics from `_dtype_policy` and which cannot be a `DTypePolicy`. While it doesn't appear to interfere, it just makes it very confusing.

### DIFF
--- a/keras/src/layers/layer.py
+++ b/keras/src/layers/layer.py
@@ -271,7 +271,8 @@ class Layer(BackendLayer, Operation, KerasSaveable):
     ):
         BackendLayer.__init__(self)
         self._lock = False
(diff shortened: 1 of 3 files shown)


fchollet (approved):
Makes sense. LGTM

Issue #2295, opened 2016-04-13
Image Captioning example in keras Approach?


bhaveshoswal:
I have used the keras example code of Image Captioning in that I have used the VGG pretrained model for extracting image features(4096) and for text part I have done indexing to the unique words and post zero padding according the max caption length(which is equal the length of biggest sentence in data) and for next words I created a numpy array of (number of example, vocabulary size) vocabulary size is equal to number of unique words in data. The next words is a 1's 0's matrix 1 means present of word in the sentence 0 absence.

for predicting what should be given exactly in partial caption and in next words if we consider "cat sat on mat".

And in training data I am appending "START" and "END" tokens in the start and end of training captions.

here is my code:

correct me if I am doing wrong in code

```
from keras.models import Sequential
from keras.layers.core import Dense, Dropout, Activation, Flatten
from keras.layers.convolutional import Convolution2D, MaxPooling2D, ZeroPadding2D
from keras.optimizers import SGD, Adadelta, Adagrad
from keras.utils import np_utils, generic_utils
from keras.callbacks import EarlyStopping
from keras.layers.advanced_activations import PReLU, LeakyReLU
from keras.layers import Embedding,GRU,TimeDistributedDense,RepeatVector,Merge
from keras.preprocessing.text import one_hot
from keras.preprocessing import sequence
import cv2
import numpy as np

max_caption_len = 21
vocab_size = 43
def VGG_16(weights_path=None):
    model = Sequential()
    model.add(ZeroPadding2D((1,1),input_shape=(3,224,224)))
    model.add(Convolution2D(64, 3, 3, activation='relu'))
    model.add(ZeroPadding2D((1,1)))
    model.add(Convolution2D(64, 3, 3, activation='relu'))
    model.add(MaxPooling2D((2,2), strides=(2,2)))

    model.add(ZeroPadding2D((1,1)))
    model.add(Convolution2D(128, 3, 3, activation='relu'))
    model.add(ZeroPadding2D((1,1)))
    model.add(Convolution2D(128, 3, 3, activation='relu'))
    model.add(MaxPooling2D((2,2), strides=(2,2)))

    model.add(ZeroPadding2D((1,1)))
    model.add(Convolution2D(256, 3, 3, activation='relu'))
    model.add(ZeroPadding2D((1,1)))
    model.add(Convolution2D(256, 3, 3, activation='relu'))
    model.add(ZeroPadding2D((1,1)))
[...]

PR #21371, opened 2025-06-10
Update saving_lib.py to properly convert from h5py to numpy array.

pctablet505:
## What does this PR do?

Fixes an incompatibility when loading model weights in environments where TensorFlow's NumPy behavior is enabled (`tf.experimental.numpy.experimental_enable_numpy_behavior`). Previously, some variables loaded as h5py datasets were not converted to numpy arrays, causing downstream errors during model construction and weight assignment, especially for layers like `ReversibleEmbedding` and `EinsumDense`.

This update ensures that all h5py datasets are converted to numpy arrays before further processing in `saving_lib.py`.

## Why is this needed?

When using TensorFlow's NumPy behavior, type promotion and array handling can differ, leading to incompatibilities if weights remain as h5py objects. This change guarantees consistent array types regardless of backend or NumPy mode.

## Related Issues/PRs

- Fixes part of [keras-hub#2136](https://github.com/keras-team/keras-hub/issues/2136)
- See also: [keras-hub PR #2295](https://github.com/keras-team/keras-hub/pull/2295)

### DIFF
--- a/keras/src/saving/saving_lib.py
+++ b/keras/src/saving/saving_lib.py
@@ -1148,6 +1148,16 @@ class H5IOStore:
             and value.attrs["dtype"] == "bfloat16"
         ):
             value = np.array(value, dtype=ml_dtypes.bfloat16)
+
+        elif (
+            hasattr(value, "shape")
+            and hasattr(value, "dtype")
+            and not isinstance(value, np.ndarray)
+        ):
+            # Convert h5py datasets to numpy arrays
+            # to avoid tensor conversion issues.
+            value = np.array(value)
+
         return value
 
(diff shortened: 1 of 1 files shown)


james77777778:
Hey @pctablet505 
I just encountered a similar issue and fixed it in this PR:
https://github.com/keras-team/keras/pull/21365

Previously, I simply added an else block for this but I have updated it to your solution.

pctablet505:
> Hey @pctablet505 I just encountered a similar issue and fixed it in this PR: #21365
> 
> Previously, I simply added an else block for this but I have updated it to your solution.

Thanks for including it.

PR #21372, opened 2025-06-10
Implement corrcoef function in keras.ops

shashaka:
Adds keras.ops.corrcoef, which computes the Pearson correlation coefficient matrix. Supported across NumPy, TensorFlow, PyTorch, and JAX backends, but not supported on OpenVINO.

### DIFF
--- a/keras/src/ops/numpy.py
+++ b/keras/src/ops/numpy.py
@@ -6888,6 +6888,35 @@ def logical_xor(x1, x2):
     return backend.numpy.logical_xor(x1, x2)
 
 
+class Corrcoef(Operation):
+    def call(self, x):
+        return backend.numpy.corrcoef(x)
+
+    def compute_output_spec(self, x):
+        dtype = backend.standardize_dtype(getattr(x, "dtype", backend.floatx()))
+        if dtype == "int64":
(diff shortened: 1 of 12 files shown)


keras/src/ops/numpy.py
```diff
@@ -6888,6 +6888,35 @@ def logical_xor(x1, x2):
...
+            dtype = dtypes.result_type(dtype, float)
+        return KerasTensor(x.shape, dtype=dtype)
+
+
+@keras_export(["keras.ops.corrcoef", "keras.ops.numpy.corrcoef"])
+def corrcoef(x):
+    """Compute the Pearson correlation coefficient matrix.
+
+    Args:
+        x: A 2D tensor of shape (N, D), where N is the number of variables
```
fchollet: Please use backticks around tuples in docstrings.

fchollet (commented):
Nice work, thank you for the PR!

keras/src/ops/numpy.py
shashaka: I've updated docstring as your guidance. Thanks.

fchollet (approved):
LGTM, thank you!

PR #21373, opened 2025-06-10
Add consistent support for name serialization of `Operation`s.

hertschuh:
This is to solve an inconsistency in the saving / reloading of ops. Some ops behave differently.

Consider this code:

```python
input = keras.layers.Input(shape=(4,), dtype="float32")
output = keras.ops.abs(input)
model = keras.models.Model(input, output)

json = model.to_json()
reloaded_model = keras.models.model_from_json(json)
reloaded_json = reloaded_model.to_json()
```

The reloaded model JSON is the same as the original model JSON. The `abs` op is serialized as

- `{"module": "keras.src.ops.numpy", "class_name": "Absolute", "config": {"name": "absolute"}, "registered_name": "Absolute", "name": "absolute", ...}`

Consider the same code with `abs` replaced with `sum`:

```python
input = keras.layers.Input(shape=(4,), dtype="float32")
output = keras.ops.abs(input)
model = keras.models.Model(input, output)

json = model.to_json()
reloaded_model = keras.models.model_from_json(json)
reloaded_json = reloaded_model.to_json()
```

The reloaded model JSON is different from the original JSON.

- `{"module": "keras.src.ops.numpy", "class_name": "Sum", "config": {"axis": null, "keepdims": false}, "registered_name": "Sum", "name": "sum", ...}`
- `{"module": "keras.src.ops.numpy", "class_name": "Sum", "config": {"axis": null, "keepdims": false}, "registered_name": "Sum", "name": "sum_1", ...}`

The reloaded `sum` op now has a name `"sum_1"` instead of `"sum"`.

This is because:
- `Abs` does not define `__init__` and inherits the `Operation.__init__` that has a `name` parameter.
- `Sum` defines `__init__` without a `name` parameter.

We want saving / reloading to be idempotent. Even though users cannot control the name of ops (although it would be easy to add), the auto-assigned names should be saved and reloaded. For this, `name` has to be supported in `__init__`.

This PR changes the behavior of serialization from `_auto_config` to allow `name` to be returned when the `__init__` signature has `name` or `**kwargs`. The existing logic had a bug. Also switched to `inspect.signature` which is the recommended API in Python 3.

Note that adding `name` as a parameter to all ops `__init__`s will be done as a separate PR.


keras/src/ops/operation_test.py
```diff
@@ -30,20 +30,71 @@ def compute_output_spec(self, x):
...
+        self.alpha = alpha
+
+    def call(self, x):
+        return self.alpha * x
+
+    def compute_output_spec(self, x):
+        return keras_tensor.KerasTensor(x.shape, x.dtype)
+
+    def get_config(self):
+        return {**super().get_config(), **{"alpha": self.alpha}}
```
abheesht17: Nit: This can probably just be 

```
return {**super().get_config(), "alpha": self.alpha}
```

hertschuh: Oh, right. Fixed.

PR #21374, opened 2025-06-11
Fix operations which had static parameter passed as dynamic parameters.

hertschuh:
Switched many parameters in ops from being dynamic (passed to `call` method) to being static (passed to `__init__`). These parameters should be static because they cannot be tensors. The most commonly changed parameters were `shape` and `dtype`.

Standardized the way `dtype` parameters are handled, in particular when they can be `None`.

Removed unused op classes: `Empty`, `Eye`, `Identity`, `Ones`, `Tri`, `Zeros`. These ops only have static parameters and no tensor inputs, they therefore cannot be part of a functional graph. They were missing the standard `if any_symbolic_tensors(...)`, which is why the class was dead code and only instantiated explicitly by unit tests.

Fixed `arange` and `full` ops. The code was missing the standard `if any_symbolic_tensors(...)`, which means that adding them to a functional graph would fail before and the op class was dead code until now.


fchollet (approved):
Sounds good -- it's fairly inconsequential API wise since everybody is only using the functional version of these ops.

PR #21381, opened 2025-06-14
Openvino backend improvements

Mezalfa:
This pull request implements keras.numpy.prod for the OpenVINO backend.

Features:
Handles reduction along specified axes, including single integers, tuples, and None.
Implements promotion for dtype when dtype is passed as None ,converting to a suitable OpenVINO type.
If dtype is specified, the output is converted to the specified dtype.
Properly manages the keepdims argument to retain or drop the reduced dimensions in the output shape.

In addition to numpy.prod, the openvino.runtime import style has been changed to the newer openvino import style in the openvino backend folder.

@rkazants


hertschuh:
Can we close this PR? 

`numpy.prod` is a separate PR, removing `runtime` is a separate PR, adding the `venv_keras_ov` is a mistake I believe.

PR #21385, opened 2025-06-16
Fix SKLearnClassifier and SKLearnRegressor examples

### LINKED ISSUE #21384: Documentation: Docstring examples for SKLearnClassifier and SKLearnRegressor do not work
The examples in the docstrings for `SKLearnClassifier` and `SKLearnRegressor` are currently not working. This is because of several small issues:
- General issues:
  - Importing layers from `keras.src.layers` instead of `keras.layers`
  - Importing Model from `keras.src.layers` instead of `keras.models`
  - Unnecessary usage of `[]` around the output layer
- `SKLearnClassifier` docstring issue:
  - The example uses `n_classes=3`, but this leads to a scikit-learn error: `ValueError: n_classes(3) * n_clusters_per_class(2) must be smaller or equal 2**n_informative(2)=4`
- `SKLearnRegressor` docstring issue:
  - Uses `softmax` activation where `linear` would be more appropriate (regression problem that does not fall between `[0, 1]`), causing warnings.

I will open a PR to address these issues shortly.

timovdk:
The examples in the docstrings for `SKLearnClassifier` and `SKLearnRegressor` are currently not working. This PR addresses the problems mentioned in, and closes #21384:

- General Fixes:
  - Import layers from `keras.layers`
  - Import `Model` from `keras.models`
  - Remove unnecessary `[]`
- `SKLearnClassifier` Fix:
  - Use a 2 class classification problem as an example (instead of the 3 class example)
- `SKLearnRegressor` Fix:
  - Use `linear` activation because it is a regression problem that does not fall between `[0, 1]`

### DIFF
--- a/keras/src/wrappers/sklearn_wrapper.py
+++ b/keras/src/wrappers/sklearn_wrapper.py
@@ -249,7 +249,7 @@ class SKLearnClassifier(ClassifierMixin, SKLBase):
             hidden = Dense(layer_size, activation="relu")(hidden)
 
         n_outputs = y.shape[1] if len(y.shape) > 1 else 1
-        out = [Dense(n_outputs, activation="softmax")(hidden)]
+        out = Dense(n_outputs, activation="softmax")(hidden)
         model = Model(inp, out)
         model.compile(loss=loss, optimizer="rmsprop")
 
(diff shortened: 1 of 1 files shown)


fchollet (approved):
LGTM, thanks for the fixes

PR #21387, opened 2025-06-16
Fix missing and fragile scikit-learn imports in Keras sklearn wrappers

### LINKED ISSUE #21386: AttributeError in SKLearnClassifier and SKLearnRegressor wrappers due to missing sklearn.utils.multiclass
When using `SKLearnClassifier` or `SKLearnRegressor` wrappers with no other scikit-learn imports (such as `make_classification`) the following error is raised: 
``` python
AttributeError: module 'sklearn.utils' has no attribute 'multiclass'
```

This happens because `sklearn.utils.multiclass` is used internally, but not explicitly imported by Keras, and it's not guaranteed to be present unless certain sklearn components are invoked prior.

#### Minimal Reproducible Example

``` python
from keras.layers import Dense, Input
from keras.models import Model
from keras.wrappers import SKLearnClassifier, SKLearnRegressor

def dynamic_model(X, y, loss, layers):
    # Creates a basic MLP model dynamically choosing the input and
    # output shapes.
    n_features_in = X.shape[1]
    inp = Input(shape=(n_features_in,))

    hidden = inp
    for layer_size in layers:
        hidden = Dense(layer_size, activation="relu")(hidden)

    n_outputs = y.shape[1] if len(y.shape) > 1 else 1
    out = Dense(n_outputs, activation="softmax")(hidden)
    model = Model(inp, out)
    model.compile(loss=loss, optimizer="rmsprop")

    return model

X = [[0.1, 0.2], [0.2, 0.3], [0.3, 0.4], [0.4, 0.5]]
y = [[1], [0], [1], [0]]
est = SKLearnClassifier( # Or SKLearnRegressor
    model=dynamic_model,
    model_kwargs={
        "loss": "categorical_crossentropy",
        "layers": [20, 20, 20],
    },
)

est.fit(X, y, epochs=5)
```

#### Expected Behaviour
The code should run without requiring implicit imports from `sklearn.utils.multiclass`.

#### Actual Behaviour
Fails with an `AttributeError` if no other sklearn functionality has triggered an implicit import of `sklearn.utils.multiclass`.

#### Suggested Fix
Explicitly import `sklearn.utils.multiclass` in the relevant wrapper module to ensure correct functioning regardless of prior sklearn usage.

I'll open a PR shortly to address this. While investigating, I found other similar cases of implicit sklearn imports. The PR will include those as well.

timovdk:
This PR addresses a bug where the `SKLearnClassifier` and `SKLearnRegressor` wrappers raise an `AttributeError` when used without other `scikit-learn` utilities (e.g. `make_classification`).

#### Fixes
- Explicitly imports `sklearn.utils.multiclass.type_of_target` instead of relying on indirect access via `sklearn.utils`
- Cleans up related imports in related files:
  - Replaces several indirect accesses with explicit imports (`make_pipeline, OneHotEncoder, MetadataRequest, check_is_fitted`)
  - Avoids the use of private API `sklearn.utils._array_api` by using `np.squeeze` directly, which is consistent with the docstring expectations

#### Related Issue
Fixes #21386

### DIFF
--- a/keras/src/wrappers/sklearn_wrapper.py
+++ b/keras/src/wrappers/sklearn_wrapper.py
@@ -18,6 +18,10 @@ try:
     from sklearn.base import ClassifierMixin
     from sklearn.base import RegressorMixin
     from sklearn.base import TransformerMixin
+    from sklearn.pipeline import make_pipeline
+    from sklearn.preprocessing import OneHotEncoder
+    from sklearn.utils.metadata_routing import MetadataRequest
+    from sklearn.utils.validation import check_is_fitted
 except ImportError:
     sklearn = None
 
@@ -134,9 +138,7 @@ class SKLBase(BaseEstimator):
                 "sklearn.set_config(enable_metadata_routing=True)."
             )
 
-        self._metadata_request = sklearn.utils.metadata_routing.MetadataRequest(
-            owner=self.__class__.__name__
-        )
+        self._metadata_request = MetadataRequest(owner=self.__class__.__name__)
         for param, alias in kwargs.items():
             self._metadata_request.score.add_request(param=param, alias=alias)
         return self
@@ -172,7 +174,7 @@ class SKLBase(BaseEstimator):
 
     def predict(self, X):
         """Predict using the model."""
-        sklearn.base.check_is_fitted(self)
+        check_is_fitted(self)
         X = _validate_data(self, X, reset=False)
         raw_output = self.model_.predict(X)
         return self._reverse_process_target(raw_output)
@@ -284,9 +286,9 @@ class SKLearnClassifier(ClassifierMixin, SKLBase):
                 f" Target type: {target_type}"
             )
         if reset:
-            self._target_encoder = sklearn.pipeline.make_pipeline(
+            self._target_encoder = make_pipeline(
                 TargetReshaper(),
-                sklearn.preprocessing.OneHotEncoder(sparse_output=False),
+                OneHotEncoder(sparse_output=False),
             ).fit(y)
             self.classes_ = np.unique(y)
             if len(self.classes_) == 1:
@@ -472,7 +474,7 @@ class SKLearnTransformer(TransformerMixin, SKLBase):
             X_transformed: array-like, shape=(n_samples, n_features)
                 The transformed data.
         """
-        sklearn.base.check_is_fitted(self)
+        check_is_fitted(self)
         X = _validate_data(self, X, reset=False)
         return self.model_.predict(X)
 
--- a/keras/src/wrappers/sklearn_wrapper.py
+++ b/keras/src/wrappers/sklearn_wrapper.py
@@ -13,7 +13,6 @@ from keras.src.wrappers.utils import _check_model
 from keras.src.wrappers.utils import assert_sklearn_installed
 
 try:
-    import sklearn
     from sklearn.base import BaseEstimator
     from sklearn.base import ClassifierMixin
     from sklearn.base import RegressorMixin
@@ -23,7 +22,6 @@ try:
     from sklearn.utils.metadata_routing import MetadataRequest
     from sklearn.utils.validation import check_is_fitted
 except ImportError:
-    sklearn = None
 
     class BaseEstimator:
         pass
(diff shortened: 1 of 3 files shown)


keras/src/wrappers/sklearn_wrapper.py
```diff
@@ -13,13 +13,15 @@
 from keras.src.wrappers.utils import assert_sklearn_installed
 
 try:
-    import sklearn
```
fchollet: Surely you should keep the `import sklearn` statement?

keras/src/wrappers/sklearn_wrapper.py
```diff
@@ -13,13 +13,15 @@
...
     from sklearn.base import BaseEstimator
     from sklearn.base import ClassifierMixin
     from sklearn.base import RegressorMixin
     from sklearn.base import TransformerMixin
+    from sklearn.pipeline import make_pipeline
+    from sklearn.preprocessing import OneHotEncoder
+    from sklearn.utils.metadata_routing import MetadataRequest
+    from sklearn.utils.validation import check_is_fitted
 except ImportError:
```
fchollet: It seems to be there's a possibility for `ImportError` to be raised even if sklearn is installed, if there's a version mismatch. This could confuse users.

fchollet (commented):
Thanks for the PR.

> where the SKLearnClassifier and SKLearnRegressor wrappers raise an AttributeError when used without other scikit-learn utilities (e.g. make_classification).

Why does this happen exactly?

It's generally much cleaner to only try to import `sklearn`.

timovdk:
> Thanks for the PR.
> 
> > where the SKLearnClassifier and SKLearnRegressor wrappers raise an AttributeError when used without other scikit-learn utilities (e.g. make_classification).
> 
> Why does this happen exactly?
> 
> It's generally much cleaner to only try to import `sklearn`.

Thanks for the review!

The issue is that `import sklearn` alone does not load submodules like `sklearn.utils.multiclass`. So when code later accesses `sklearn.utils.multiclass.type_of_target` it can raise an `AttributeError` because `utils` or `multiclass` hasn't been imported yet. This doesn't show up in the docstring examples, since those import `make_classification`, which indirectly imports the `multiclass` (and other `.utils`) submodules.

Switching to directly importing `type_of_target` ensures that missing dependencies raise an `ImportError` at import time. However, to raise the `ImportError` consistently, I agree that `import sklearn` should be re-added to these files, thanks!

keras/src/wrappers/sklearn_wrapper.py
timovdk: Good point, but I think failing early with an `ImportError` is better than an `AttributeError` during runtime.
Other options could be to check the `sklearn` version, or to catch the `ImportError` and re-raise with an informative error message. 
Do you have any other suggestions or preferences?

keras/src/wrappers/sklearn_wrapper.py
timovdk: Thanks, let me re-add it to retain consistent `ImportError` behaviour.

fchollet:
> The issue is that import sklearn alone does not load submodules like sklearn.utils.multiclass. So when code later accesses sklearn.utils.multiclass.type_of_target it can raise an AttributeError because utils or multiclass hasn't been imported yet

This is an issue with the design of sklearn; if the package was well-designed then you could access any member of the API from `sklearn.xyz...` (as is the case for Keras).

To route around it I would suggest leaving the top-of-file import structure unchanged, and then doing e.g. `from sklearn.utils.validation import check_is_fitted` right before you need it, in-line.

timovdk:
> This is an issue with the design of sklearn; if the package was well-designed then you could access any member of the API from `sklearn.xyz...` (as is the case for Keras).

Yes I agree, and this is only the case for `multiclass`, which is not exposed in [`sklearn/utils/__init__.py`](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/__init__.py). That's why `type_of_target` cannot be accessed through `sklearn.utils.multiclass.type_of_target`.

After fixing that import, I adjusted the other sklearn imports for consistency by using the same `from sklearn.xyz import foo` style and grouping them at the top of the file for readability.

But, if the preference is to use inline imports, I propose to revert all non-essential import changes and leave only the inline import for `type_of_target`, because I think that the readability argument no longer applies with inline imports.

(Edit): I also found that `check_is_fitted` is not exposed through `sklearn/utils/__init__.py`, so it too would require an inline import. The others (`metadata_routing`, `preprocessing`, and `pipeline`) are exposed through their respective `__init__.py` files, so I can safely revert those changes.

fchollet (approved):
LGTM, thanks for the update!

PR #21392, opened 2025-06-17
Fix crash when `sys.stdout.encoding` is `None`

mohiuddin-khan-shiam:
[print_msg](cci:1://file:///d:/Github/keras/keras/src/utils/io_utils.py:91:0-112:29) now defaults to UTF-8 when `sys.stdout.encoding` is unset, preventing a `TypeError` during logging in non-interactive or redirected-output environments.


fchollet (approved):
LGTM, thanks

Issue #21391, opened 2025-06-17
RNN constructor should use raise TypeError and not ValueError


jfdev001:
This is very nit-picky, but I happened to look at the RNN constructor and see that a `ValueError` is raised when, by definition, the exception should be a `TypeError`, right? 

https://github.com/keras-team/keras/blob/764ed95651c1c7dfa71ca523287f6eeb514ccf77/keras/src/layers/rnn/rnn.py#L221-L225

From [TypeError: python org docs](https://docs.python.org/3/library/exceptions.html#TypeError)

>  exception TypeError
> Raised when an operation or function is applied to an object of inappropriate type. The associated value is a string giving details about the type mismatch.

and [ValueError: python org docs](https://docs.python.org/3/library/exceptions.html#ValueError)

> exception ValueError
> Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as [IndexError](https://docs.python.org/3/library/exceptions.html#IndexError).

The `state_size` argument in `__init__(...)` **must** be one of the specified types. The currently raised `ValueError` implies that `state_size` is the *correct type* but *incorrect value*, but that implication is directly contradicted by "state_size must be an integer, or a list/tuple of integers". 

This is an extremely minor fix, and am happy to open a PR to do so, but wanted to open an issue first! I appreciate any feedback!

dhantule:
Hi @jfdev001, thanks for the report. 

As per this [comment ](https://github.com/keras-team/keras/pull/21395#issuecomment-2988851684)  Keras will raise `ValueError` for an invalid user-provided argument, can we close this issue if you don't have anymore questions?

jfdev001:
@dhantule I see the tag still awaiting response is present, apologies for not explicitly responding! I have no further questions.

dhantule:
@jfdev001, Thank you! Feel free to open another issue.

monicadsong:
Hi @jfdev001 -- 

The Keras team would like to learn more your Keras usage. Please thumbs up this comment if you are open to speaking with us and we'll reach out to you over email. 

Thanks!
Monica, Keras Product Manager

PR #21395, opened 2025-06-17
Fix: Raise TypeError for invalid state_size in RNN constructor

00Harshh:
Previously, the code raised a ValueError when state_size was of the wrong type. This commit updates it to raise a TypeError instead, as it's a more appropriate exception for type mismatches.

Fixes keras-team/keras#21391


fchollet:
In general, Keras will always raise `ValueError` for an invalid user-provided argument, by convention.

PR #21403, opened 2025-06-18
Fix(OpenVINO): Ensure numpy.prod passes tests and handles boolean inp…

Mezalfa:
This pull request implements keras.numpy.prod for the OpenVINO backend.

Features:
Handles reduction along specified axes, including single integers, tuples, and None.
Implements promotion for dtype when dtype is passed as None ,converting to a suitable OpenVINO type.
If dtype is specified, the output is converted to the specified dtype.
Properly manages the keepdims argument to retain or drop the reduced dimensions in the output shape.

PR #21405, opened 2025-06-19
Master

Imokutmfon:
Implementation of numpy.prod on Keras Openvino Backend

Fixes openvinotoolkit/openvino#30212

please review @rkazants

### DIFF
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -1262,7 +1262,14 @@ def pad(x, pad_width, mode="constant", constant_values=None):
 
 
 def prod(x, axis=None, keepdims=False, dtype=None):
-    raise NotImplementedError("`prod` is not supported with openvino backend")
+    x = get_ov_output(x)
+    if dtype is not None:
+        ov_type = OPENVINO_DTYPES[standardize_dtype(dtype)]
+        x = ov_opset.convert(x, ov_type).output(0)
+    axis = ov_opset.constant(axis, Type.i32).output(0)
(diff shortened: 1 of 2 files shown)


hertschuh (approved):
Thanks for the PR!

hertschuh:
Which one between this one and https://github.com/keras-team/keras/pull/21403 should I merge?

hertschuh:
@rkazants please review

keras/src/backend/openvino/numpy.py
```diff
@@ -1268,7 +1268,37 @@ def pad(x, pad_width, mode="constant", constant_values=None):
...
+    if axis is None:
+        flatten_shape = ov_opset.constant([-1], Type.i32).output(0)
+        x = ov_opset.reshape(x, flatten_shape, False).output(0)
+        axis = 0
+    elif isinstance(axis, tuple):
+        axis = list(axis)
+    axis = ov_opset.constant(axis, Type.i32).output(0)
+    if x_type == Type.boolean:
+        result = ov_opset.reduce_logical_and(x, axis, keepdims).output(0)
+        result = ov_opset.convert(result, Type.i32).output(0)
```
Mohamed-Ashraf273, line 1284: You need to check if `dtype` is not `None` first. For example, if you run:
```python
import numpy as np

a = np.array([True, False, True])
print(np.prod(a, dtype=np.float32).dtype)
```
NumPy gives `float32`, but your code always returns `int32`.

hertschuh:
This was implemented by https://github.com/keras-team/keras/pull/21567

PR #21407, opened 2025-06-21
update a small feature in __getitem__ for openvino backend

Mohamed-Ashraf273:
Hi @fchollet 
I just added a check for `np.ndarray` in the function that was recently merged.

### DIFF
--- a/keras/src/backend/openvino/core.py
+++ b/keras/src/backend/openvino/core.py
@@ -341,8 +341,13 @@ class OpenVINOKerasTensor:
                 raise ValueError(
                     "OpenVINO backend does not support boolean indexing"
                 )
-            elif isinstance(index, (int, np.integer)):
-                if isinstance(index, np.integer):
+            elif isinstance(index, (int, np.integer, np.ndarray)):
+                if isinstance(index, (np.ndarray, np.integer)):
+                    if isinstance(index, np.ndarray) and len(index.shape) != 0:
+                        raise ValueError(
+                            "OpenVINO backend does not support"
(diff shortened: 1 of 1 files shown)


keras/src/backend/openvino/core.py
```diff
@@ -341,8 +341,13 @@ def count_unsqueeze_before(dim):
                 raise ValueError(
                     "OpenVINO backend does not support boolean indexing"
                 )
-            elif isinstance(index, (int, np.integer)):
-                if isinstance(index, np.integer):
+            elif isinstance(index, (int, np.integer, np.ndarray)):
+                if isinstance(index, (np.ndarray, np.integer)):
```
rkazants, line 345: not needed

keras/src/backend/openvino/core.py
```diff
@@ -341,8 +341,13 @@ def count_unsqueeze_before(dim):
...
+                    if isinstance(index, np.ndarray):
```
rkazants, line 346: same

keras/src/backend/openvino/core.py
```diff
@@ -341,8 +341,13 @@ def count_unsqueeze_before(dim):
...
+                        assert len(index.shape) == 0, (
```
fchollet: Please use ValueError instead of assert.

Mohamed-Ashraf273: @fchollet 
Done!

Mohamed-Ashraf273:
@fchollet

rkazants:

PR #21410, opened 2025-06-22
[Keras 3 OpenVINO Backend]: Support numpy.flip operation

Imokutmfon:
implementation for numpy.flip using openvino ops

please review @rkazants

Fixes openvinotoolkit/openvino#29359

### DIFF
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -800,7 +800,35 @@ def expm1(x):
 
 
 def flip(x, axis=None):
-    raise NotImplementedError("`flip` is not supported with openvino backend")
+    x = get_ov_output(x)
+    rank = x.get_partial_shape().rank.get_length()
+    x_shape = ov_opset.shape_of(x)
+    if rank == 0:
+        return OpenVINOKerasTensor(x)
+    if axis is None:
+        axis = list(range(rank))
+    elif isinstance(axis, int):
+        axis = [axis]
+    else:
+        axis = list(axis)
+    axis = [a + rank if a < 0 else a for a in axis]
+    flipped = x
+    for ax in axis:
+        dim_size = ov_opset.gather(
+            x_shape,
+            ov_opset.constant(ax, Type.i64).output(0),
+            axis=ov_opset.constant(0, Type.i64).output(0),
+        ).output(0)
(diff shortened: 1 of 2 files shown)


rkazants:
@Mohamed-Ashraf273, could you please review this PR? Thanks

keras/src/backend/openvino/numpy.py
```diff
@@ -800,7 +800,35 @@ def expm1(x):
...
+        ).output(0)
+        start = ov_opset.subtract(
+            dim_size, ov_opset.constant(1, Type.i64).output(0)
+        ).output(0)
+        stop = ov_opset.constant(-1, Type.i64).output(0)
+        step = ov_opset.constant(-1, Type.i64).output(0)
+        indices = ov_opset.range(start, stop, step, Type.i64).output(0)
+        flipped = ov_opset.gather(
+            flipped, indices, ov_opset.constant(ax, Type.i64).output(0)
+        ).output(0)
```
Mohamed-Ashraf273, line 830: I recommend considering the use of `strided_slice` instead of building reversed indices with `range` and `gather` for each axis.

keras/src/backend/openvino/numpy.py
```diff
@@ -800,7 +800,35 @@ def expm1(x):
...
 def flip(x, axis=None):
-    raise NotImplementedError("`flip` is not supported with openvino backend")
+    x = get_ov_output(x)
+    rank = x.get_partial_shape().rank.get_length()
+    x_shape = ov_opset.shape_of(x)
+    if rank == 0:
+        return OpenVINOKerasTensor(x)
+    if axis is None:
+        axis = list(range(rank))
+    elif isinstance(axis, int):
```
Mohamed-Ashraf273, line 810: I think to make it more robust and user-friendly, we can add a check for `np.integer` and `np.ndarray`. For `np.ndarray`, we just need to check if the shape is scalar or 1-element and integer-convertible

keras/src/backend/openvino/numpy.py
rkazants: @Imokutmfon, please use ReverseSequence operation: https://docs.openvino.ai/2025/documentation/openvino-ir-format/operation-sets/operation-specs/movement/reverse-sequence-1.html

PR #21418, opened 2025-06-24
[OpenVINO backend] Remove usage of deprecated `openvino.runtime`

p-wysocki:
`openvino.runtime` is deprecated in favor of flat `openvino` and is set to be removed in the upcoming OpenVINO release.


p-wysocki:
@fchollet @mattdangerw The CI is failing with pre-commit check `api-gen`, but there's no explicit error message. I ran the generation script as instructed in https://github.com/keras-team/keras?tab=readme-ov-file#minimal-installation with `.shell/api_gen.sh`, but it updated ~130 files with changes which are irrelevant, for example:

```bash
diff --git a/keras/api/optimizers/__init__.py b/keras/api/optimizers/__init__.py
index 40f6ab401..4114fab5b 100644
--- a/keras/api/optimizers/__init__.py
+++ b/keras/api/optimizers/__init__.py
@@ -4,6 +4,7 @@ This file was autogenerated. Do not edit it by hand,
 since your modifications would be overwritten.
 """

+
 from keras.optimizers import legacy as legacy
 from keras.optimizers import schedules as schedules
 from keras.src.optimizers import deserialize as deserialize
@@ -18,9 +19,7 @@ from keras.src.optimizers.adamw import AdamW as AdamW
 from keras.src.optimizers.ftrl import Ftrl as Ftrl
 from keras.src.optimizers.lamb import Lamb as Lamb
 from keras.src.optimizers.lion import Lion as Lion
-from keras.src.optimizers.loss_scale_optimizer import (
-    LossScaleOptimizer as LossScaleOptimizer,
-)
+from keras.src.optimizers.loss_scale_optimizer import LossScaleOptimizer as LossScaleOptimizer
 from keras.src.optimizers.muon import Muon as Muon
```

Is this expected? @rkazants suggested that it may be caused by wrong ordering of imports in my PR, but I changed it in numerous different ways and the check still fails.

rkazants:
@Mohamed-Ashraf273, could you please advice how to fix code-formatting issue in this PR? Thanks

keras/src/backend/openvino/core.py
```diff
@@ -4,11 +4,9 @@
...
 import numpy as np
 import openvino as ov
-import openvino.runtime.opset14 as ov_opset
-from openvino import Model
-from openvino import Tensor
-from openvino import compile_model
-from openvino.runtime import Type
+import openvino.opset14 as ov_opset
+
+from openvino import Model, Tensor, compile_model, Type
```
Mohamed-Ashraf273: @p-wysocki 
To pass the formatter test, please update this import statement to the following:
```python
from openvino import Model
from openvino import Tensor
from openvino import Type
from openvino import compile_model
```
you can also install pre-commit by running:
```bash
pip install pre-commit
```
Then, run the following command locally to automatically fix formatting issues:
```bash
pre-commit run --all-files --hook-stage manual
```

p-wysocki:
```suggestion
from openvino import Model
from openvino import Tensor
from openvino import Type
from openvino import compile_model
```

keras/src/backend/openvino/core.py
```diff
@@ -4,11 +4,12 @@
...
 import numpy as np
 import openvino as ov
-import openvino.runtime.opset14 as ov_opset
+import openvino.opset14 as ov_opset
+
 from openvino import Model
 from openvino import Tensor
+from openvino import Type
 from openvino import compile_model
-from openvino.runtime import Type
```
p-wysocki, line 12:
```suggestion
```

p-wysocki:
Thank you @Mohamed-Ashraf273!

Issue #20861, opened 2025-02-05
Progress bar showing wrong step number


itmo153277:
When `steps_per_execution` is used, the step that is showed by progress bar does not match how many steps were actually executed.

See example here: [colab](https://colab.research.google.com/gist/itmo153277/358d490853d9b9cd662dca488834aafe/wrongstepnumber.ipynb)

The step number does not match the number of optimizer iterations

itmo153277:
@sonali-kumari1 
I don't think that this is the desired behaviour. Because if ProgressBar shows 11 in actuality the model saw 20 batches which is very confusing, especially since normally you cannot see optimizer iterations directly. Also, if an epoch has 50 steps it will show 41 and then 50 with the same metric values (+validation). I think a proper implementation should show 10, 20, 30, 40, 50, 50 just like in Keras 2

PR #21422, opened 2025-06-26
Progress bar now handles `steps_per_execution`.

hertschuh:
Progress bar would always report the starting batch + 1 at the end of the batch. Now it takes into account `steps_per_execution` for the last batch reported.

Fixes https://github.com/keras-team/keras/issues/20861


mattdangerw (approved):
Code looks good! Haven't tested out the progbar or anything like that.

PR #21425, opened 2025-06-26
Run tests on TPU

kharshith-k:
Running tests on V6 TPU runner.

### DIFF
--- a/.github/workflows/tpu_tests.yml
+++ b/.github/workflows/tpu_tests.yml
@@ -1,4 +1,4 @@
-name: Keras TPU Tests
+name: Keras Tests
 
 # TODO: Consider enabling all tests (pytest, applications, etc.) with NNX in the future
 # Currently only basic flow tests run with NNX enabled
@@ -6,7 +6,6 @@ name: Keras TPU Tests
 on:
   push:
     branches: [ master ]
-  pull_request:
   pull_request_review:
     types: [submitted]
   release:


.github/workflows/actions.yml
```diff
@@ -57,7 +57,7 @@ jobs:
         run: |
           pip install -r requirements.txt --progress-bar off --upgrade
           if [ "${{ matrix.nnx_enabled }}" == "true" ]; then
-            pip install --upgrade git+https://github.com/google/flax.git
+            pip install --upgrade flax>=0.11.1
```
hertschuh, line 60: This file should be reverted

.github/workflows/tpu-tests-jax.yml
```diff
@@ -0,0 +1,104 @@
+# name: TPU Tests
```
hertschuh: Remove all commented out code

.github/workflows/tpu-tests-jax.yml
```diff
@@ -0,0 +1,104 @@
...
+permissions:
+  contents: read
+
+jobs:
+  test-in-container:
+    name: Run Keras tests on TPU runner using JAX Backend
+    runs-on: linux-x86-ct6e-44-1tpu
+
+    container:
+      # The container image is now set to python:3.10-slim
```
hertschuh: Remove this comment

.github/workflows/tpu-tests-jax.yml
```diff
@@ -0,0 +1,104 @@
...
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+
+      - name: Install System Dependencies
+        run: |
+          apt-get update && apt-get install -y --no-install-recommends \
+            git \
+            sudo \
+            && rm -rf /var/lib/apt/lists/*
```
hertschuh: I don't think this is needed. I got it to work without this:
https://github.com/keras-team/keras-rs/blob/main/.github/workflows/actions.yml#L49

.github/workflows/tpu-tests-jax.yml
```diff
@@ -0,0 +1,104 @@
...
+      
+      - name: Install Dependencies
+        run: |
+          pip install --no-cache-dir -U pip setuptools && \
```
hertschuh: Remove `setuptools`, we're not building the wheel.

.github/workflows/tpu-tests-jax.yml
```diff
@@ -0,0 +1,104 @@
...
+          pip install --no-cache-dir -U psutil && \
```
hertschuh: Remove `psutils` we don't need it.

.github/workflows/tpu-tests-jax.yml
```diff
@@ -0,0 +1,104 @@
...
+          pip install --no-cache-dir -r requirements-jax-tpu.txt && \
+          pip uninstall -y keras keras-nightly
```
hertschuh: Remove `pip uninstall -y keras keras-nightly`, this is not needed, we install keras from source.

.github/workflows/tpu-tests-jax.yml
```diff
@@ -0,0 +1,104 @@
...
+        run: |
+          echo "Successfully running inside the public python container!"
+          echo "Verifying JAX installation..."
+          python3 -c "import jax; print(f'JAX backend: {jax.default_backend()}'); print(f'JAX devices: {jax.devices()[0].device_kind}')"
+
+          pytest keras --ignore keras/src/applications \
+                      --ignore keras/src/layers/merging/merging_test.py \
+                      --ignore keras/src/trainers/data_adapters/py_dataset_adapter_test.py \
+                      --ignore keras/src/backend/jax/distribution_lib_test.py \
+                      --ignore keras/src/distribution/distribution_lib_test.py \
```
hertschuh: Remove all the `--ignore` except for `--ignore keras/src/applications`

keras/src/backend/common/dtypes_test.py
```diff
@@ -245,4 +245,4 @@ def test_invalid_float8_dtype(self):
         with self.assertRaisesRegex(
             ValueError, "There is no implicit conversions from float8 dtypes"
         ):
-            dtypes.result_type("float8_e5m2", "bfloat16")
```
hertschuh, line 289: Revert this file.

conftest.py
```diff
@@ -58,4 +58,4 @@ def pytest_collection_modifyitems(config, items):
 
 
 def skip_if_backend(given_backend, reason):
-    return pytest.mark.skipif(backend() == given_backend, reason=reason)
+    return pytest.mark.skipif(backend() == given_backend, reason=reason)
```
hertschuh: Revert this file.

Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM --platform=linux/amd64 python:3.10-slim
```
hertschuh: Is this file needed?

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+ENV KERAS_HOME=/github/workspace/.github/workflows/config/jax \
+    KERAS_BACKEND=jax
```
hertschuh: This shouldn't be here since we want to support TF too.

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+RUN pip install --no-cache-dir -U pip setuptools && \
+    pip install --no-cache-dir -U psutil && \
+    pip install --no-cache-dir -r requirements-jax-tpu.txt && \
+    pip uninstall -y keras keras-nightly
+    # python3 -c 'import jax;print(jax.__version__);print(jax.default_backend())' && \
+    # python3 -c 'import jax;assert jax.default_backend().lower() == "tpu"' && \
+    # pytest keras --ignore keras/src/applications \
+    #              --ignore keras/src/layers/merging/merging_test.py \
+    #              --cov=keras \
+    #              --cov-config=pyproject.toml
```
hertschuh: Remove, this will be done by the action.

requirements-common-jax-tpu.txt
```diff
@@ -0,0 +1,31 @@
+pre-commit
```
hertschuh: Why can't you use the `requirements-common.txt`?

requirements-common.txt
```diff
@@ -20,12 +20,8 @@ pytest-cov
 packaging
 # for tree_test.py
 dm_tree
-coverage
+coverage!=7.6.5  # 7.6.5 breaks CI
```
hertschuh: Revert this file.

requirements-tensorflow-tpu.txt
```diff
@@ -0,0 +1,14 @@
+#tensorflow==2.18.0
+#--find-links https://storage.googleapis.com/libtpu-tf-releases/index.html
+#tensorflow-tpu==2.18.0
```
hertschuh: Use `tensorflow-tpu==2.19.1`.

.github/workflows/actions.yml
kharshith-k: Sure! Reverted

.github/workflows/tpu-tests-jax.yml
kharshith-k: Removed all the commented code

.github/workflows/tpu-tests-jax.yml
kharshith-k: Removed the comment

.github/workflows/tpu-tests-jax.yml
kharshith-k: Sure! It worked for me as well. I've removed that step

.github/workflows/tpu-tests-jax.yml
kharshith-k: Sure! Removed

.github/workflows/tpu-tests-jax.yml
kharshith-k: Done!

.github/workflows/tpu-tests-jax.yml
kharshith-k: Got it! I've Removed the line

.github/workflows/tpu-tests-jax.yml
kharshith-k: Sure! Removed all the --ignore lines except `keras/src/applications`

conftest.py
kharshith-k: Reverted the file

Dockerfile
kharshith-k: I had developed this file during the initial stages of implementing TPU support workflow. Later I started using the openly available images like python-slim and ubuntu. Currently this file is not being used. So removing it

requirements-common-jax-tpu.txt
kharshith-k: I had created this file during initial stages to segregate concerns regarding each library and tried to avoid errors by not using certain modules. Now I guess this is not required. Using the `requirements-common.txt` for consistency

requirements-common.txt
kharshith-k: Done!

requirements-tensorflow-tpu.txt
kharshith-k: Sure! I've updated the file to use `tensorflow-tpu==2.19.1`. But I've not added tensorflow in the backend matrix of tpu-tests-jax.yml since we had decided to focus on jax for now

requirements-tensorflow-tpu.txt
```diff
@@ -0,0 +1,13 @@
+--find-links https://storage.googleapis.com/libtpu-tf-releases/index.html
+tensorflow-tpu==2.19.0
```
hertschuh: `tensorflow-tpu==2.19.1`

requirements-tensorflow-tpu.txt
```diff
@@ -0,0 +1,13 @@
...
+
+#tf2onnx
```
hertschuh: Uncomment `tf2onnx`.

.github/workflows/tpu-tests-jax.yml
```diff
@@ -0,0 +1,47 @@
...
+  workflow_call:
+  pull_request:
+  release:
+    types: [created]
+
+permissions:
+  contents: read
+
+jobs:
+  test-in-container:
```
hertschuh, line 19: Can move all of the content in this file in `actions.yml`? It will be easier to have everything in one place.

.github/workflows/tpu-tests-jax.yml
```diff
@@ -0,0 +1,47 @@
...
+          pip install --no-cache-dir -r requirements-${{ matrix.backend }}-tpu.txt \
+      
+      - name: Set Keras Backend
+        run: echo "KERAS_BACKEND=jax" >> $GITHUB_ENV
+
+      - name: Run Verification and Tests
+        run: |
+          echo "Successfully running inside the public python container!"
+          echo "Verifying JAX installation..."
+          python3 -c "import jax; print(f'JAX backend: {jax.default_backend()}'); print(f'JAX device Type : {jax.devices()[0].device_kind}')"
```
hertschuh: nitpick: instead of printing just the JAX device kinds, just print the whole devices. It will tell you what they are and how many.

`print(f'JAX devices: {jax.devices()}')`

.github/workflows/tpu-tests-jax.yml
kharshith-k: Sure!

kharshith-k:
An observed was made that tests were silently falling back to CPU during execution even when TPU was available while running them on a TPU VM which lead to the suspicion that the same could be happening with current passing tests. its a good idea to log the device type on which tests are getting executed (cpu/cuda/tpu).

kharshith-k:
Hi @hertschuh, I'm not sure how to do that exactly, Should I add the TPU specific workflow in `.kokoro/github/ubuntu/gpu/build.sh`? Can you please share some resources regarding the same if you have any? In the mean time I'll do some reading from my end too!

hertschuh:
> Hi @hertschuh, I'm not sure how to do that exactly, Should I add the TPU specific workflow in `.kokoro/github/ubuntu/gpu/build.sh`? Can you please share some resources regarding the same if you have any? In the mean time I'll do some reading from my end too!

It looks like this would be a way to do it: https://github.com/orgs/community/discussions/25372#discussioncomment-3247688

But it means you need a separate yml file.

kharshith-k:
Hello @hertschuh , I've added the condition to check if the PR is reviewed and approved before running the TPU tests job. Can you please let me know if the change is valid?

.github/workflows/actions.yml
```diff
@@ -7,13 +7,55 @@ on:
   push:
     branches: [ master ]
   pull_request:
+  pull_request_review:
+    types: [submitted]
```
hertschuh: Hmm... so the issue with this is that it causes all the tests to be re-run on review (not even just approval).

So you could add an `if` to the other tests, but then it will create a bunch of `skipped` tests on review.

I think the better way would be to have a separate file for the TPU tests.

.github/workflows/tpu_tests.yml
```diff
@@ -0,0 +1,57 @@
+name: Keras TPU Tests
+
+# TODO: Consider enabling all tests (pytest, applications, etc.) with NNX in the future
+# Currently only basic flow tests run with NNX enabled
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
```
hertschuh: Remove `pull_request` here. They're always excluded by the `if` below so it just creates a "skipped" entry.

.github/workflows/tpu_tests.yml
```diff
@@ -0,0 +1,57 @@
+name: Keras TPU Tests
```
hertschuh: nitpitck: rename this file and change this name.

The GPU tests are going to move here very soon.

PR #21428, opened 2025-06-27
Fix symbolic call of `logsumexp` with int axis.

hertschuh:
Using `keras.ops.math.logsumexp` with an int for `axis` in a functional model would throw an error.


mattdangerw (commented):
Thank you!

mattdangerw (approved):

PR #21429, opened 2025-06-27
Only allow deserialization of `KerasSaveable`s by module and name.

hertschuh:
Arbitrary functions and classes are not allowed.

- Made `Operation` extend `KerasSaveable`, this required moving imports to avoid circular imports
- `Layer` no longer need to extend `KerasSaveable` directly
- Made feature space `Cross` and `Feature` extend `KerasSaveable`
- Also dissallow public function `enable_unsafe_deserialization`


fchollet (approved):
Thanks for the fix!

PR #21430, opened 2025-06-27
Support non-tf.data.Dataset instances in multi-host setting.

cantonios:
It is assumed that the data is already pre-sharded across hosts.

If `distribution.auto_shard_dataset` is `True` but the input dataset isn't a tf.data.Dataset, will error out as before. Setting `auto_shard_dataset` to `False` will allow any dataset type to be used.

Unfortunately we can't add a test for this in OSS since we don't have a multi-process test fixture.  However, this was tested internally.

### DIFF
--- a/keras/src/distribution/distribution_lib.py
+++ b/keras/src/distribution/distribution_lib.py
@@ -297,11 +297,19 @@ class Distribution:
 
     Args:
         device_mesh: A `DeviceMesh` instance.
+        batch_dim_name: Optional string name for the batch dimension.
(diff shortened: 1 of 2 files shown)


cantonios:
@hertschuh This is the keras change to allow us to use a generator dataset across multiple hosts.

fchollet (approved):
LGTM, thank you for the feature!

@hertschuh do you also want to do a review?

amitsrivastava78 (approved):
LGTM

PR #21432, opened 2025-06-28
It worth not hiding exceptions

WIgor:
What do you think about logging of exceptions, instead of hiding them? + There is an example of such hidden exception when head_shards and q_seq_shards were just not initialized because get_dist()==None.

    And despite hardware supports flash attention - this code silently falls back to the standard with O(N^2) memory.

### DIFF
--- a/keras/src/backend/jax/nn.py
+++ b/keras/src/backend/jax/nn.py
@@ -1,11 +1,10 @@
 import builtins
 import math
 
-from absl import logging
-
 import jax
 import jax.experimental.sparse as jax_sparse
 import jax.numpy as jnp
+from absl import logging
 from jax import lax
 from jax import nn as jnn
 from jax.experimental.pallas.ops.tpu.splash_attention import (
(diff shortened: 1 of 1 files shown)


divyashreepathihalli (approved):
Thanks for the PR! Please run `pre-commit run --all-files`

hertschuh:
@WIgor ,

Thanks for the PR! Can you reformat the code? Then I can merge it.

WIgor:
As far as I see it does nothing

```
~/Projects/keras on  patch-1 ⌚ 6:07:26
$ pre-commit run --all-files
api_gen..................................................................Passed
ruff.....................................................................Passed
ruff-format..............................................................Passed
(ml)
~/Projects/keras on  patch-1 ⌚ 6:08:19
$ git status
On branch patch-1
nothing to commit, working tree clean
(ml)
```

WIgor:
Ahh ok - it seems working now.

hertschuh (approved):

PR #21433, opened 2025-06-28
[OpenVINO backend] support repeat

Mohamed-Ashraf273:
@fchollet 
@rkazants

### DIFF
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -1329,7 +1329,9 @@ def repeat(x, repeats, axis=None):
         and repeats.size == 1
     ):
         repeats_val = (
-            int(repeats) if isinstance(repeats, np.ndarray) else repeats
+            int(repeats)
+            if isinstance(repeats, (np.integer, np.ndarray))
+            else repeats
         )
         dim_len = ov_opset.gather(
             ov_opset.shape_of(x, Type.i32),


keras/src/backend/openvino/numpy.py
```diff
@@ -1291,7 +1291,96 @@ def reciprocal(x):
 
 
 def repeat(x, repeats, axis=None):
-    raise NotImplementedError("`repeat` is not supported with openvino backend")
+    x = get_ov_output(x)
```
rkazants, line 1314: I think you can use `Tile` operation from our opset. It should be much easier to express `np.repeats`

Mohamed-Ashraf273: `tile` repeats the whole tensor, but repeat needs to repeat each element in a tensor, so I am using `tile` for scalar repeating and using `range`, `cumsum`, and `gather` for tensor's repeating.

Mohamed-Ashraf273:
@fchollet 
@rkazants 
I've checked:
https://docs.pytorch.org/docs/stable/generated/torch.repeat_interleave.html#torch.repeat_interleave
https://github.com/openvinotoolkit/openvino/blob/master/src/frontends/pytorch/src/op/repeat_interleave.cpp
And tried to make some simplifications, this pr is ready for review.

Mohamed-Ashraf273:
@fchollet

rkazants (commented):
@fchollet, recommend to merge. Looks good to me.

PR #21434, opened 2025-06-29
Fix torch LSTM onnx export issue.

### LINKED ISSUE #21390: Failed to convert LSTM layers to ONNX on Torch backend
Not sure if this is Torch or Keras issue, but simple LSTM models cannot be exported as ONNX.

```
TypeError: Exception encountered when calling LSTM.call().

unsupported operand type(s) for -: 'torch.return_types.max' and 'int'

Arguments received by LSTM.call():
  • sequences=torch.Tensor(shape=torch.Size([1, 32, 256]), dtype=float32)
  • initial_state=None
  • mask=None
  • training=None
```

Here's the link to colab notebook reproducing the issue:
https://colab.research.google.com/drive/1QCTYR9xCConwt4D4R3rXUxvAeO0Gd2IG

P.S. The same model will convert fine with TF backend.
P.P.S. The verbosity flag is not respected.

james77777778:
Fix #21390

The root cause is not related to ONNX export itself, but rather a type issue in torch's `rnn`.
`torch.max(..., dim=...)` returns a `torch.return_type.max`, not a tensor. Ref: https://docs.pytorch.org/docs/stable/generated/torch.max.html
After extracting the tensor via `max_len = max_len[0]`, the export works correctly.

I'm not a RNN expert and unsure how to test this at the ops level, so I have added a test based on the reported issue.

`onnxscript` dep was added to support the TorchDynamo-based ONNX exporter.

cc @mattdangerw @DLumi

### DIFF
--- a/keras/src/backend/torch/rnn.py
+++ b/keras/src/backend/torch/rnn.py
@@ -228,6 +228,8 @@ def rnn(
         elif isinstance(input_length, torch.Tensor):
             if go_backwards:
                 max_len = torch.max(input_length, dim=0)
+                if isinstance(max_len, torch.return_types.max):
+                    max_len = max_len[0]
                 rev_input_length = torch.subtract(max_len - 1, input_length)
 
                 def masking_fn(time):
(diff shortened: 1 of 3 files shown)


fchollet (approved):
LGTM, thanks for the fix!

PR #21435, opened 2025-06-29
[DRAFT] int4 quantization support

JyotinderSingh:


fchollet (commented):
Thanks for the PR! The code generally looks good to me. What is the performance profile? How did you benchmark the change?

JyotinderSingh:
> Thanks for the PR! The code generally looks good to me. What is the performance profile? How did you benchmark the change?

I hadn't yet benchmarked the code. I've now created two micro-benchmarks and have linked them in the PR description, please take a look!

fchollet:
Thanks for running the benchmark! It looks like this was obtained with the TF backend; can you try the JAX backend as well?

JyotinderSingh:
> Thanks for running the benchmark! It looks like this was obtained with the TF backend; can you try the JAX backend as well?

Looks like the JAX backend is enabling worse runtime performance (both GPU and TPU) for OPT, the accuracy metrics stay the same. The overall numbers for BERT are better on JAX with TPU, however, the benefits of int4 quantization are minimal outside of storage savings.

1. OPT 125M GPU
This is one of the better runs, in some runs the int4 performance was measured to be worse.
jax_int4_gpu_opt_1

2. OPT 125M TPU
jax_int4_tpu_opt_1

3. BERT Classifier (TPU)
jax_int4_tpu_bert

I'm presently unsure why we have such a noticeable difference. I'll try and investigate the cause for this.

fchollet:
Are the numbers you're seeing in the best case along the lines of what you expected? Based on the benchmark the benefits of int4 seem really slim. Do we have a 3rd party baseline to compare to?

divyashreepathihalli:
The accuracy with int4 is not expected to be better right. I do see the model size is reducing correct, which is the main motivation.

JyotinderSingh:
> Are the numbers you're seeing in the best case along the lines of what you expected? Based on the benchmark the benefits of int4 seem really slim. Do we have a 3rd party baseline to compare to?

Basic int4 quantization suffers from high precision loss, however we plan to integrate support for GPTQ and AWQ moving ahead. That should allow us to bring the error down (already have some promising results working in some hacked together scripts)

We're also exploring adding support for Q4 dtype (used by Gemma from llama.cpp) and QAT. These should also bring improvements to inference and runtime performance.

keras/src/quantizers/quantizers.py
```diff
@@ -374,3 +374,106 @@ def quantize_and_dequantize(inputs, scale, quantized_dtype, compute_dtype):
...
     x = ops.multiply(ops.cast(x, compute_dtype), ops.cast(scale, compute_dtype))
     return x
+
+
+@keras_export("keras.quantizers.pack_int4")
+def pack_int4(arr, axis=0):
+    """Pack an int4 tensor into an int8 tensor with packed nibbles.
+
+    Accepts a Keras-compatible tensor. The input values must already be int8
+    in the signed range ``[-8, 7]`` and represent the desired int4 values.
```
fchollet: Use single backticks in docstrings.

keras/src/quantizers/quantizers.py
```diff
@@ -374,3 +374,106 @@ def quantize_and_dequantize(inputs, scale, quantized_dtype, compute_dtype):
...
+
+    * For every two consecutive rows, the **low nibble** of the output byte
+      stores the value from the first row, and the **high nibble** stores
+      the value from the second row.
+
+    Returns a tuple ``(packed, packed_shape, orig_rows)`` where ``packed``
+    is the packed ``int8`` tensor, ``packed_shape`` is its shape, and
+    ``orig_rows`` is the original (unpacked) row count prior to any padding
+    that may have been inserted when an odd number of rows is supplied.
+    """
```
fchollet, line 452: Please provide some examples in the docstring. Also add a `Args:` and `Returns` section.

keras/src/quantizers/quantizers.py
```diff
@@ -374,3 +374,106 @@ def quantize_and_dequantize(inputs, scale, quantized_dtype, compute_dtype):
...
+    packed = ops.cast(packed, "int8")
+
+    # 5-6. Restore shape.
+    packed = ops.transpose(packed, inv_perm)  # back to original order
+    orig_len = rows  # number of slices before padding
+    return packed, ops.shape(packed), orig_len
+
+
+@keras_export("keras.quantizers.unpack_int4")
+def unpack_int4(packed, orig_len, axis=0):
```
fchollet, line 503: Same here.

JyotinderSingh:
There was some issue with the original benchmarking script. I've fixed it, and now we're seeing significantly better results for GPU memory usage.

1. Text Generation Micro-Benchmark with OPT 125M using KerasHub: [colab notebook](https://colab.research.google.com/gist/JyotinderSingh/918fcdb8a1433dea12800f8ca4a240f5/int4_quantization_micro_benchmark_opt.ipynb)

opt_tf_gpu

2. Classifier Micro-Benchmark with DistilBERT (fine-tuned on SST2) using KerasHub: [colab notebook](https://colab.research.google.com/gist/JyotinderSingh/11358103789158d7a9b36dcf977a482b/int4-quantization-micro-benchmark-distilbert.ipynb)

distilbert_tf_gpu

We can expect further speedups once we support quantization in more layers.

keras/src/layers/core/dense.py
```diff
@@ -337,6 +352,38 @@ def _int8_build(self, kernel_shape):
             trainable=False,
         )
 
+    def _int4_build(self, kernel_shape):
+        """Build variables for int4 quantization.
```
fchollet, line 356: Add line break below

keras/src/quantizers/quantizers.py
```diff
@@ -374,3 +374,266 @@ def quantize_and_dequantize(inputs, scale, quantized_dtype, compute_dtype):
     # Dequantize
     x = ops.multiply(ops.cast(x, compute_dtype), ops.cast(scale, compute_dtype))
     return x
+
+
+@keras_export("keras.quantizers.pack_int4")
+def pack_int4(arr, axis=0):
+    """
+    Pack an int4 tensor into an int8 tensor with packed nibbles.
```
fchollet: Move this line to the line above

keras/src/quantizers/quantizers.py
```diff
@@ -374,3 +374,266 @@ def quantize_and_dequantize(inputs, scale, quantized_dtype, compute_dtype):
...
+    stores the value from the first row, and the **high nibble** stores
+    the value from the second row.
+
+    Args:
+        arr: An int8 tensor containing int4 values in the range `[-8, 7]`.
+        axis: The axis along which to pack the tensor. Defaults to 0.
+
+    Returns:
+        tuple: A tuple `(packed, packed_shape, orig_rows)` where `packed` is
+        the packed int8 tensor with int4 values stored in nibbles,
```
fchollet: Add indent

keras/src/quantizers/quantizers.py
```diff
@@ -374,3 +374,266 @@ def quantize_and_dequantize(inputs, scale, quantized_dtype, compute_dtype):
...
+        `packed_shape` is the shape of the packed tensor, and `orig_rows` is
+        the original (unpacked) row count prior to any padding that may have
+        been inserted when an odd number of rows is supplied.
+
+    Example:
+        >>> import numpy as np
```
fchollet: Unindent the code example and add line break below "Example:"

fchollet (commented):
Thanks for the PR! The code looks good. The new results look reasonable.

divyashreepathihalli (approved):

PR #21438, opened 2025-06-30
Fix `rms_normalization` and `layer_normalization` to ensure consistent precision in computations.

james77777778:
- Added tests.
- Fixed numeric stability issues in half-precision training by using float32 for computations (which is also the default in Flax)
- Updated the fn signature of the `Operation`.

Ref: https://github.com/google/flax/blob/main/flax/linen/normalization.py


james77777778:
@fchollet 
I think we should prioritize this PR because without it mixed-precision training will be unstable or broken.

Here I need to explicitly set `dtype="float32"` for the CLIP model to overcome the dtype issues:
https://github.com/keras-team/keras-hub/pull/2316

PR #21443, opened 2025-07-01
Enable ROCm devices to correctly use keras.layers.CuDNNGRU and keras.layers.CuDNNLSTM

JoeNan1:
Using keras.layers.CuDNNGRU and keras.layers.CuDNNLSTM on ROCm devices may introduce accuracy issues due to incorrect weight layout. This modification ensures compatibility with both CUDA and ROCm weight layouts.

### DIFF
--- a/keras/layers/rnn/cudnn_gru.py
+++ b/keras/layers/rnn/cudnn_gru.py
@@ -152,23 +152,34 @@ class CuDNNGRU(_CuDNNRNN):
         input_h = initial_state[0]
         input_h = tf.expand_dims(input_h, axis=0)
 
+        
+        weights = [
+            self.kernel[:, : self.units],                      
+            self.kernel[:, self.units : self.units * 2],       
+            self.kernel[:, self.units * 2 :],                  
+            self.recurrent_kernel[:, : self.units],            
+            self.recurrent_kernel[:, self.units : self.units * 2], 
+            self.recurrent_kernel[:, self.units * 2 :],        
+        ]
+
+        biases = [
+            self.bias[: self.units],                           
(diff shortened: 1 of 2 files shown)


.github/workflows/format.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           python -m pip install --upgrade pip setuptools
           echo "::set-output name=dir::$(pip cache dir)"
       - name: pip cache
-        uses: actions/cache@v2
+        uses: actions/cache@v4
```
fchollet: Please remove these 2 files from the PR

keras/layers/rnn/cudnn_gru.py
```diff
@@ -152,23 +152,34 @@ def _process_batch(self, inputs, initial_state):
...
+        biases = [
+            self.bias[: self.units],                           
+            self.bias[self.units : self.units * 2],            
+            self.bias[self.units * 2 : self.units * 3],        
+            self.bias[self.units * 3 : self.units * 4],        
+            self.bias[self.units * 4 : self.units * 5],        
+            self.bias[self.units * 5 :],                       
+        ]
+        
+        if tf.sysconfig.get_build_info()["is_cuda_build"]:
```
fchollet, line 174: ?

JoeNan1: The default layout is compatible with ROCm; here, the CUDA layout is adapted by checking whether the device is CUDA.

.github/workflows/format.yml
JoeNan1: Reverted to the original state before the changes.

PR #21444, opened 2025-07-01
Implement deg2rad function in keras.ops

shashaka:
Adds keras.ops.deg2rad, which converts angles from degrees to radians.

### DIFF
--- a/keras/src/backend/jax/numpy.py
+++ b/keras/src/backend/jax/numpy.py
@@ -598,6 +598,10 @@ def cumsum(x, axis=None, dtype=None):
     return jnp.cumsum(x, axis=axis, dtype=dtype)
 
 
+def deg2rad(x):
+    return jnp.deg2rad(x)
+
+
 def diag(x, k=0):
     x = convert_to_tensor(x)
     return jnp.diag(x, k=k)
--- a/keras/src/backend/numpy/numpy.py
+++ b/keras/src/backend/numpy/numpy.py
@@ -515,6 +515,10 @@ def cumsum(x, axis=None, dtype=None):
     return np.cumsum(x, axis=axis, dtype=dtype)
 
 
+def deg2rad(x):
+    return np.deg2rad(x)
+
+
 def diag(x, k=0):
     return np.diag(x, k=k)
 
--- a/keras/src/backend/numpy/numpy.py
+++ b/keras/src/backend/numpy/numpy.py
@@ -516,7 +516,16 @@ def cumsum(x, axis=None, dtype=None):
 
 
 def deg2rad(x):
-    return np.deg2rad(x)
+    if x.dtype in ["int64", "float64"]:
+        dtype = "float64"
+    elif x.dtype in ["bfloat16", "float16"]:
+        dtype = x.dtype
+    else:
+        dtype = config.floatx()
+
+    x = convert_to_tensor(x)
+
+    return np.deg2rad(x).astype(dtype)
 
 
 def diag(x, k=0):
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -1257,6 +1257,7 @@ def cumsum(x, axis=None, dtype=None):
 
 
 def deg2rad(x):
+    x = convert_to_tensor(x)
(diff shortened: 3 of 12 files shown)


keras/src/ops/numpy.py
```diff
@@ -2258,6 +2258,36 @@ def cumsum(x, axis=None, dtype=None):
...
+    `rad = deg * (π / 180)`
+
+    Args:
+        x: Input tensor of angles in degrees.
+
+    Returns:
+        A tensor containing angles converted to radians.
+
+    Examples:
+    >>> from keras.src import ops
```
fchollet: Use `from keras import ops`

keras/src/ops/numpy.py
```diff
@@ -2258,6 +2258,36 @@ def cumsum(x, axis=None, dtype=None):
...
+    >>> ops.deg2rad(180.0)
+    3.141592653589793
+    >>> ops.deg2rad([0.0, 90.0, 180.0])
+    array([0.        , 1.57079633, 3.14159265])
```
fchollet: Trim whitespace

keras/src/backend/tensorflow/numpy.py
```diff
@@ -1256,6 +1256,27 @@ def cumsum(x, axis=None, dtype=None):
...
+        "uint8",
+        "uint16",
+        "uint32",
+    ]:
+        dtype = config.floatx()
+    elif standardize_dtype(dtype) in ["int64"]:
+        dtype = "float64"
+    x = tf.cast(x, dtype)
+
+    x = convert_to_tensor(x)
```
james77777778: We should first use `x = convert_to_tensor(x)` then do the dtype check you impl above.
Otherwise a python scalar will fail.

keras/src/backend/torch/numpy.py
```diff
@@ -670,6 +670,13 @@ def cumsum(x, axis=None, dtype=None):
     return torch.cumsum(x, dim=axis, dtype=to_torch_dtype(dtype))
 
 
+def deg2rad(x):
+    if standardize_dtype(x.dtype) == "int64":
+        return cast(torch.deg2rad(x), "float64")
+    x = convert_to_tensor(x)
```
james77777778: same here

keras/src/backend/numpy/numpy.py
```diff
@@ -515,6 +515,19 @@ def cumsum(x, axis=None, dtype=None):
...
 
+def deg2rad(x):
+    if x.dtype in ["int64", "float64"]:
+        dtype = "float64"
+    elif x.dtype in ["bfloat16", "float16"]:
+        dtype = x.dtype
+    else:
+        dtype = config.floatx()
+
+    x = convert_to_tensor(x)
```
james77777778: We should first use `x = convert_to_tensor(x)` then do the dtype check you impl above.
Otherwise a python scalar will fail.

keras/src/backend/jax/numpy.py
```diff
@@ -598,6 +598,10 @@ def cumsum(x, axis=None, dtype=None):
     return jnp.cumsum(x, axis=axis, dtype=dtype)
 
 
+def deg2rad(x):
+    return jnp.deg2rad(x)
```
james77777778, line 603: We better add `x = convert_to_tensor(x)` here.

shashaka: I've updated.

keras/src/backend/numpy/numpy.py
shashaka: Thanks for review. I have updated.

keras/src/backend/torch/numpy.py
shashaka: updated also.

keras/src/backend/tensorflow/numpy.py
shashaka: Thanks for review. I've applied the change.

keras/src/ops/numpy.py
shashaka: I have updated.

keras/src/ops/numpy.py
shashaka: I have changed it.

PR #21452, opened 2025-07-04
Fix torch's `convert_to_tensor` not respecting `dtype` when input is a `Variable`

james77777778:
There is a bug in torch's `convert_to_tensor` when the input is a `Variable`.

The root cause is that we return `x.variable` directly without checking its dtype.
This PR fixes the issue.

### DIFF
--- a/keras/src/ops/nn.py
+++ b/keras/src/ops/nn.py
@@ -2816,8 +2816,7 @@ def _rms_normalization(x, scale=None, axis=-1, epsilon=None):
 
     x = backend.convert_to_tensor(x, dtype=compute_dtype)
     if scale is not None:
-        scale = backend.convert_to_tensor(scale)
-        scale = backend.cast(scale, x.dtype)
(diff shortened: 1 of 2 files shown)


keras/src/ops/nn.py
```diff
@@ -2816,7 +2816,8 @@ def _rms_normalization(x, scale=None, axis=-1, epsilon=None):
 
     x = backend.convert_to_tensor(x, dtype=compute_dtype)
     if scale is not None:
-        scale = backend.convert_to_tensor(scale, x.dtype)
+        scale = backend.convert_to_tensor(scale)
+        scale = backend.cast(scale, x.dtype)
```
fchollet: Why would 2 separate lines be necessary?

james77777778: No. I was testing the behavior and forgot to revert them.

Fixed. I also modified the tests in `nn_test.py` to use variables as inputs to verify the changes.

fchollet (approved):
LGTM, thank you!

PR #21453, opened 2025-07-04
Implement cbrt function in keras.ops

shashaka:
Adds keras.ops.cbrt, which computes the real-valued cube root of each element in the input tensor.


fchollet (approved):
LGTM, thank you for the PR!

PR #21454, opened 2025-07-04
Improve `load_weights` and the test coverage.

james77777778:
Fix #21414

This PR adds error msg in `load_weights` and increases test coverage.


fchollet (approved):
LGTM, thanks for the fix!

PR #21456, opened 2025-07-06
Fix: UpSampling2D bilinear set_image_data_format(channels_first) bug

### LINKED ISSUE #21401: keras.layers.UpSampling2D bilinear channel first bug
import tensorflow as tf
import keras
from keras.src import ops

keras.backend.set_image_data_format('channels_first')

x = tf.zeros((3, 12, 224, 224))

output_upsampling_layer = keras.layers.UpSampling2D(size=(2, 2), interpolation="bilinear")(x)
output_bilinear_resize_only = ops.image.resize(x, (448, 448), interpolation="bilinear")

results in:
output_upsampling_layer shape: (3, 448, 224, 448)
output_bilinear_resize_only shape (3, 12, 448, 448)

tristandb8:
The current approach transforms the tensor to channels_last, before passing it in ops.image.resize, which has been defined as channels_first if keras.backend.set_image_data_format has been called on channelsfirst in the user's code. This creates a bug, a passed in tensor [16, 3, 224, 224] will return as shape [16, 448, 224, 448] instead of [16, 3, 448, 448]. Setting the data_format as the expected channels_last fixes that issue.

I changed the unit tests to integrate keras.backend.set_image_data_format, also I noticed 
layer = layers.UpSampling2D(
    size=(length_row, length_col),
    data_format=data_format,
)
Inside test_upsampling_2d_bilinear, wasn't using interpolation="bilinear", so I corrected that too

Closes #21401.

### DIFF
--- a/keras/src/layers/reshaping/up_sampling2d_test.py
+++ b/keras/src/layers/reshaping/up_sampling2d_test.py
@@ -68,7 +68,9 @@ class UpSampling2dTest(testing.TestCase):
         length_col=[2, 3],
     )
     @pytest.mark.requires_trainable_backend
-    def test_upsampling_2d_bilinear(self, data_format, use_set_image_data_format, length_row, length_col):
+    def test_upsampling_2d_bilinear(
+        self, data_format, use_set_image_data_format, length_row, length_col
+    ):
         num_samples = 2
         stack_size = 2
         input_num_row = 11
--- a/keras/src/layers/reshaping/up_sampling2d_test.py
+++ b/keras/src/layers/reshaping/up_sampling2d_test.py
@@ -10,6 +10,14 @@ from keras.backend import set_image_data_format
 
 
 class UpSampling2dTest(testing.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.original_image_data_format = backend.image_data_format()
+
+    @classmethod
+    def tearDownClass(cls):
+        backend.set_image_data_format(cls.original_image_data_format)
+
     @parameterized.product(
         data_format=["channels_first", "channels_last"],
         length_row=[2],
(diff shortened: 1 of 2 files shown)


keras/src/layers/reshaping/up_sampling2d_test.py
```diff
@@ -62,15 +63,22 @@ def test_upsampling_2d(self, data_format, length_row, length_col):
...
+    def test_upsampling_2d_bilinear(
+        self, data_format, use_set_image_data_format, length_row, length_col
+    ):
         num_samples = 2
         stack_size = 2
         input_num_row = 11
         input_num_col = 12
+
+        if use_set_image_data_format:
+            set_image_data_format(data_format)
```
fchollet, line 88: You should not set this globally in a unit test, instead pass the argument when you create the UpSampling2D layer.

fchollet (commented):
Thanks for the PR!

keras/src/layers/reshaping/up_sampling2d_test.py
tristandb8: Unfortunately, that's the only way the error happens, as I've stated here
https://github.com/keras-team/keras/issues/21401#issue-3158026406

There are no issues if I simply do
`keras.layers.UpSampling2D(size=(2, 2), interpolation="bilinear", data_format="channels_first")(x)`


I'm okay with just removing that line from the test, but it won't be testing my specific issue. Let me know if that's our best solution.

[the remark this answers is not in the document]
fchollet: This is a good suggestion

tristandb8:
The last commit follows similar logic in
`keras\src\applications\applications_test.py`

fchollet (approved):

Issue #21465, opened 2025-07-09
keras.ops.searchsorted doesn't work with symbolic tensors


iezepov:
I'm trying to extend `keras.ops.searchsorted` to work on 2D tensors. Natively it supports 1D only, and the error message suggests to use vectorized_map. However this throws an error because searchsorted tries to call `len` on a symbolic tensor instead of shape.

Minimal example:
```
import keras
xs = keras.ops.stack([keras.ops.linspace(0, 1, 5) for _ in range(20)], axis=-1)
inputs = keras.random.uniform((100, 20), seed=42)

# this is okay
keras.ops.searchsorted(xs[:, 0], inputs[:, 0])  

# this thwors
keras.ops.vectorized_map(
    lambda i: keras.ops.searchsorted(xs[:, i], inputs[:, i]), 
    keras.ops.arange(20)
) 
```

Throws:
```
TypeError: in user code:

    File "/tmp/ipython-input-416-1753265487.py", line 2, in <lambda>
        keras.ops.vectorized_map(lambda i: keras.ops.searchsorted(xs[:, i], inputs[:, i]), keras.ops.arange(20)
    File ".../py/keras/src/ops/numpy.py", line 5278, in searchsorted
        return backend.numpy.searchsorted(sorted_sequence, values, side=side)
    File ".../keras/src/backend/tensorflow/numpy.py", line 2103, in searchsorted
        "int32" if len(sorted_sequence) <= np.iinfo(np.int32).max else "int64"

    TypeError: len is not well defined for a symbolic Tensor (loop_body/strided_slice:0). Please call `x.shape` rather than `len(x)` for shape information.
```

iezepov:
Note that this is true for TF backend. With jax backend it just works somehow.

dhantule:
Hi @iezepov, thanks for reporting this. 

After testing your example, I've faced the same 
`TypeError: len is not well defined for a symbolic Tensor (loop_body/strided_slice:0). Please call x.shape rather than len(x) for shape information.` in this [gist](https://colab.sandbox.google.com/gist/dhantule/8c559296147199673f716fa9097b1fe3/keras-ops-searchsorted-doesn-t-work-with-symbolic-tensors-21465-open.ipynb). 
We'll look into this issue and update you.

PR #21472, opened 2025-07-14
Fix TypeError in keras.ops.searchsorted by using shape instead of len

dhantule:
Replaces `len(sorted_sequence)` with `sorted_sequence.shape[0]` in the `searchsorted` function to prevent a TypeError when `sorted_sequence` is a symbolic Tensor, error reproduced in this [gist](https://colab.sandbox.google.com/gist/dhantule/8c559296147199673f716fa9097b1fe3/keras-ops-searchsorted-doesn-t-work-with-symbolic-tensors-21465-open.ipynb).

Fixes: #21465

### DIFF
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -2134,7 +2134,9 @@ def searchsorted(sorted_sequence, values, side="left"):
             f"sorted_sequence.shape={sorted_sequence.shape}"
         )
     out_type = (
-        "int32" if len(sorted_sequence) <= np.iinfo(np.int32).max else "int64"
+        "int32" 
+        if (sorted_sequence).shape[0] <= np.iinfo(np.int32).max 
+        else "int64"
     )
     return tf.searchsorted(
         sorted_sequence, values, side=side, out_type=out_type
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -2134,9 +2134,7 @@ def searchsorted(sorted_sequence, values, side="left"):
             f"sorted_sequence.shape={sorted_sequence.shape}"
         )
     out_type = (
-        "int32" 
-        if (sorted_sequence).shape[0] <= np.iinfo(np.int32).max 
-        else "int64"
+        "int32" if sorted_sequence.shape[0] <= np.iinfo(np.int32).max else "int64"
     )
     return tf.searchsorted(
         sorted_sequence, values, side=side, out_type=out_type


keras/src/backend/tensorflow/numpy.py
```diff
@@ -2134,7 +2134,7 @@ def searchsorted(sorted_sequence, values, side="left"):
             f"sorted_sequence.shape={sorted_sequence.shape}"
         )
     out_type = (
-        "int32" if len(sorted_sequence) <= np.iinfo(np.int32).max else "int64"
+        "int32" if sorted_sequence.shape[0] <= np.iinfo(np.int32).max else "int64"
```
fchollet, line 2137: You also need to check `sorted_sequence.shape[0] is not None` (it could be None for a symbolic tensor)

iezepov:
Sorry, if I'm not familiar with the codebase enough, but would "backend/numpy" and "backend/torch" also need the same fix? Code looks identical there: https://github.com/search?q=repo%3Akeras-team%2Fkeras%20len(sorted_sequence)&type=code

keras/src/backend/tensorflow/numpy.py
dhantule: Hi @fchollet, I have tested searchsorted with sorted_sequence as None  and I am getting value error, traceback is shown here. Should I need to raise another value error ?

```
keras_sorted= keras.ops.searchsorted(None, [13, 12])
print("keras_sorted", keras_sorted)
```


```
  File "/Users/dhantule/Desktop/keras-test/searchsorted#21465.py", line 22, in <module>
    keras_sorted= keras.ops.searchsorted(None, [13, 12])
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dhantule/Desktop/dhantule-keras/keras/keras/src/ops/numpy.py", line 5329, in searchsorted
    sorted_sequence = backend.convert_to_tensor(sorted_sequence)
  File "/Users/dhantule/Desktop/dhantule-keras/keras/keras/src/backend/tensorflow/core.py", line 153, in convert_to_tensor
    return tf.convert_to_tensor(x, dtype=dtype)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/tensorflow/python/util/traceback_utils.py", line 153, in error_handler
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/tensorflow/python/framework/constant_op.py", line 108, in convert_to_eager_tensor
    return ops.EagerTensor(value, ctx.device_name, dtype)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: Attempt to convert a value (None) with an unsupported type (<class 'NoneType'>) to a Tensor.
```

JyotinderSingh:
> Sorry, if I'm not familiar with the codebase enough, but would "backend/numpy" and "backend/torch" also need the same fix? Code looks identical there: https://github.com/search?q=repo%3Akeras-team%2Fkeras%20len(sorted_sequence)&type=code

Yes, fixes need to be ported across backends to maintain consistent behavior

JyotinderSingh:
@dhantule just checking in on the status of this fix, are you facing any blockers?

dhantule:
> @dhantule just checking in on the status of this fix, are you facing any blockers?

Hi @JyotinderSingh, I'll port the fix across backends, just wanted to know about this 
https://github.com/keras-team/keras/pull/21472#discussion_r2212562205

keras/src/backend/tensorflow/numpy.py
hertschuh: Oh the `sorted_sequence` would not be `None`, simply, it's shape would have a `None` dimension.

I think this is one way to trigger it is:
```python
keras_sorted= keras.ops.searchsorted(keras.KerasTensor((None,), dtype="float32"), [13, 12])
```

PR #21474, opened 2025-07-14
Implement heaviside function in keras.ops

shashaka:
Adds keras.ops.heaviside, which computes the element-wise Heaviside step function.

### DIFF
--- a/keras/src/ops/numpy.py
+++ b/keras/src/ops/numpy.py
@@ -1339,10 +1339,10 @@ def heaviside(x1, x2):
 
     Args:
         x1: A tensor input.
-        x2: A scalar or tensor, the value to return when x1 == 0.
+        x2: A scalar or tensor, the value to return when `x1 == 0`.
 
     Returns:
-        A tensor with a shape determined by broadcasting x1 and x2.
+        A tensor with a shape determined by broadcasting `x1` and `x2`.
 
     Example:
     >>> x1 = keras.ops.convert_to_tensor([-2.0, 0.0, 3.0])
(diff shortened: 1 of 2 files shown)


keras/src/ops/numpy.py
```diff
@@ -1317,6 +1317,44 @@ def hanning(x):
...
+@keras_export(["keras.ops.heaviside", "keras.ops.numpy.heaviside"])
+def heaviside(x1, x2):
+    """Heaviside step function.
+
+    The Heaviside step function is defined as:
+    `heaviside(x1, x2) = 0 if x1 < 0, 1 if x1 > 0, x2 if x1 == 0`
+
+    Args:
+        x1: A tensor input.
+        x2: A scalar or tensor, the value to return when x1 == 0.
```
fchollet: Please use backticks around code keywords.

keras/src/ops/numpy.py
```diff
@@ -1317,6 +1317,44 @@ def hanning(x):
...
+
+    Returns:
+        A tensor with a shape determined by broadcasting x1 and x2.
```
fchollet: Likewise.

keras/src/ops/numpy.py
shashaka: I have updated.

keras/src/ops/numpy.py
shashaka: I have updated.

fchollet (approved):
LGTM -- thank you for the contribution!

PR #21486, opened 2025-07-16
[OpenVINO backend] suppor export model using openvino format

Mohamed-Ashraf273:
Implement a logic in Keras that allows to serialize a model loaded in any backend, including openvino backend into IR on a disk
```python
import os
os.environ["KERAS_BACKEND"] = "backend"

import keras_hub

causal_lm = keras_hub.models.GPT2CausalLM.from_preset("gpt2_medium_en", dtype="float16")
model.export("path/to/location", format="openvino")
```


keras/src/export/openvino.py
```diff
@@ -0,0 +1,107 @@
...
+        import tempfile
+
+        with tempfile.TemporaryDirectory() as temp_dir:
+            model.export(temp_dir, format="tf_saved_model")
+            ov_model = ov.convert_model(temp_dir)
+    else:
+        raise NotImplementedError(
+            "`export_openvino` is only compatible with OpenVINO and "
+            "TensorFlow backends."
+        )
```
rkazants: not sure, why do we need it here?

Mohamed-Ashraf273: If the model is loaded in tf backend and the user needs to convert it to OpenVINO IR.

rkazants: `ov.convert_model` supports conversion of TF object from memory as well. Will it work?

Mohamed-Ashraf273: No, it doesn't work in our case because the model has dynamic input shapes and no `example_input` is passed. This causes `ov.convert_model` to fail with a `ValueError` related to unknown `TensorShape`. Exporting to a `SavedModel` first resolves this by providing concrete input signatures.

james77777778: If the `example_input` is the only blocker, we can use `convert_spec_to_tensor` to create a dummy input.

Ref: https://github.com/keras-team/keras/blob/master/keras/src/export/onnx.py#L91-L95

Mohamed-Ashraf273: You're right that `convert_spec_to_tensor()` can provide a dummy input which already is being used via `OpenVINO` backend, but the issue here isn’t just about having any input, it’s about the shape of that input.
In our case, the model was traced with a specific shape, and later we ran inference with a different shape, which caused a mismatch error in `OpenVINO`.
It might be solvable in more complex ways, or like how `onnx` did in `TensorFlow` part, but for `OpenVINO`, I think exporting to `tf_saved_model` first is the safest and easiest method.

Mohamed-Ashraf273:
@rkazants 
@fchollet 
I've supported all backends, including `torch` and `jax`. Also I avoided loading the model on disk in case of `tf`, and now it is ready for review.

keras/src/export/openvino.py
```diff
@@ -0,0 +1,174 @@
...
+        filepath: `str` or `pathlib.Path`. Path to the output `.xml` file.
+        The corresponding `.bin` file will be saved alongside it.
+        verbose: Optional `bool`. Whether to print a confirmation message
+        after export. If `None`, it uses the default verbosity configured
+        by the backend.
+        input_signature: Optional. Specifies the shape and dtype of the
+        model inputs. If not provided, it will be inferred.
+        **kwargs: Additional keyword arguments (currently unused).
+     Example:
+        import keras
```
fchollet: Please use a code block (and no starting indent) for the code example

keras/src/export/openvino.py
```diff
@@ -0,0 +1,174 @@
...
+    Args:
+        filepath: `str` or `pathlib.Path`. Path to the output `.xml` file.
+        The corresponding `.bin` file will be saved alongside it.
+        verbose: Optional `bool`. Whether to print a confirmation message
+        after export. If `None`, it uses the default verbosity configured
+        by the backend.
+        input_signature: Optional. Specifies the shape and dtype of the
+        model inputs. If not provided, it will be inferred.
+        **kwargs: Additional keyword arguments (currently unused).
+     Example:
```
fchollet, line 36: Add line break above

fchollet (commented):
Looks good, thank you!

keras/src/export/openvino.py
Mohamed-Ashraf273: Done!

keras/src/export/openvino.py
Mohamed-Ashraf273: Done!

fchollet (approved):

PR #21490, opened 2025-07-18
Update compute_output_spec method to support cbrt operation

shashaka:
Adds support for the cbrt operation in compute_output_spec.
Ensures correct output shape and dtype inference for cube root computations.


fchollet (approved):
LGTM, thank you!

Issue #20071, opened 2024-07-31
tf.data.Dataset Pipeline Preprocessing Custom Layer Recommendation


apage224:
I am looking to create a number of custom preprocessing layers to be used in a TensorFlow `tf.data` pipeline. I initially assumed I could subclass `keras.Layer` and in `call` simply use any `keras.ops` operations. I only use python parameters statically for condition statements but otherwise use `keras.ops` for all commands (e.g. `keras.ops.fori_loop`). I can run the pipeline alone successfully (e.g. `iter(next(train_ds))`), however, when I try to train a model using `TensorFlow` backend it complains with several issues as it's trying to create a symbolic graph of my preprocess layers. These layers are not attached to the model- they are attached to the data pipeline via `map`. I had assumed the dataset pipeline would happen on the CPU but it seems that my layers are being mapped to the GPU in a TF graph. Are there any reference examples I could follow?

When I looked at the included preprocessing layers of `keras` they all seemed to use `keras.backend.numpy` for operations (rather than `keras.ops`). I also noticed the TF pipeline safe layers subclass `TFDataLayer` which isn't exposed in the public API. Is there a way to indicate to keras that I want to run the entire preprocessing pipeline on the CPU.

Any help would be greatly appreciated.

Below are some layers that I implemented as reference (based on what I could find from both `keras` and `keras-cv`:

```python

from typing import Callable
import keras

from .defines import NestedTensorValue


def tf_keras_map(f, xs):
    # NOTE: Workaround until (https://github.com/keras-team/keras/issues/20048)
    import tensorflow as tf

    xs = keras.tree.map_structure(tf.convert_to_tensor, xs)

    def get_fn_output_signature(x):
        out = f(x)
        return keras.tree.map_structure(tf.TensorSpec.from_tensor, out)

    # Grab single element unpacking and repacking single element
    xe = tf.nest.pack_sequence_as(xs, [y[0] for y in tf.nest.flatten(xs)])
    fn_output_signature = get_fn_output_signature(xe)
    return tf.map_fn(f, xs, fn_output_signature=fn_output_signature)


class BaseAugmentation(keras.layers.Layer):
    SAMPLES = "data"
    LABELS = "labels"
    TARGETS = "targets"
    ALL_KEYS = (SAMPLES, LABELS, TARGETS)
    TRANSFORMS = "transforms"
    IS_DICT = "is_dict"
    BATCHED = "is_batched"
    USE_TARGETS = "use_targets"
[...]

PR #21494, opened 2025-07-21
Add Grain integration.

james77777778:
This PR adds a new `GrainDatasetAdapter`.

With this PR, we can now pass a `grain.MapDataset`, `grain.IterDataset`, or `grain.DataLoader` (with `worker_count >= 0`) into `Model.fit`, `Model.predict`, and `Model.evaluate`.

Thanks to @mattdangerw for the detailed doc on this proposal!

## Overview

We propose gradually switching Keras default mode of preprocessing from [tf.data](https://www.tensorflow.org/guide/data) to [PyGrain](https://github.com/google/grain). We want the following:
1. A uniform preprocessing experience across backends. Write preprocessing once, switch backends seamlessly. Have the option, but not the requirement, to use backend specific solutions (e.g. torch data loaders).
2. Minimal dependencies. We should not depend on the TensorFlow binary (which is huge) or TensorflowText (which is largely unmaintained) on other backends. Similarly we should not need Jax when running Torch or Torch when running Jax.
3. Multi-processed execution for arbitrary python code. We should be able to run regular python preprocessing code without needing to worry about compiled graph compatibility or XLA constraints. Users should be able to mix in arbitrary python code and python packages and get reasonably fast execution by simply running across all available CPU cores.

tf.data is a good solution for 1) but not 2) and 3). This has been a long running source of user pain since the release of Keras 3 (e.g. [bug](https://github.com/keras-team/keras/issues/20071), [bug](https://github.com/keras-team/keras/issues/19895), [bug](https://github.com/keras-team/keras-hub/issues/2101)).
We are not dogmatically attached to PyGrain, but believe it is the best available option – it solves all three points above and is becoming the default preprocessing solution for JAX.

## Benchmark

|Dataloader|Backend|`fit()` cost time (step time)|`evaluate` cost time|
|-|-|-|-|
|`grain.IterDataset`|JAX|2.81s (2ms)|0.11s|
|`grain.DataLoader`|JAX|4.64s (3ms)|0.11s|
|`PyDataset`|JAX|0.73s (503us)|0.03s|
|`tf.data`|JAX|0.84s (523us)|0.04s|
|`torch.utils.data.DataLoader`|JAX|1.31s (894us)|0.06s|

Note that I didn’t enable multiprocessing or prefetching in any of the dataloaders to simplify the test, so the performance here isn’t optimized.
Also, the dataset is in memory with no heavy augmentation which might differ from real-world scenarios.

Here is the script that benchmarks various dataloaders on the MNIST dataset:

`python3 benchmark.py --type [grain_dataloader|grain_dataset|pydataset|tf_data|torch_dataloader]`

<details>

<summary>benchmark.py</summary>

```python
import argparse
import math
import time

import grain
import numpy as np
from torch.utils.data import DataLoader
from torch.utils.data import Dataset

import keras
from keras.src import datasets
from keras.src import layers
from keras.src import utils
from keras.src.trainers.data_adapters.py_dataset_adapter import PyDataset


class MNISTSource(grain.sources.RandomAccessDataSource):
    def __init__(self, x, y):
        super().__init__()
        self.x = x
        self.y = y

    def __len__(self):
        return len(self.x)

    def __getitem__(self, index):
        return self.x[index], self.y[index]


class MNISTPyDataset(PyDataset):
    def __init__(self, x, y, batch_size=128):
        super().__init__()
        self.x = x
        self.y = y
        self.batch_size = batch_size

    def __len__(self):
        return math.ceil(len(self.x) / self.batch_size)

    def __getitem__(self, idx):
        low = idx * self.batch_size
        high = min(low + self.batch_size, len(self.x))
        batch_x = self.x[low:high]
        batch_y = self.y[low:high]
        return batch_x, batch_y


class MNISTTorchDataset(Dataset):
    def __init__(self, x, y):
        self.x = x
        self.y = y

    def __len__(self):
        return len(self.x)

    def __getitem__(self, index):
        return self.x[index], self.y[index]


if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument(
        "--type",
        choices=[
            "grain_dataloader",
            "grain_dataset",
            "pydataset",
            "tf_data",
            "torch_dataloader",
        ],
        default="grain_dataloader",
    )
    args = parser.parse_args()

    # Set the parameters.
    num_classes = 10
    batch_size = 128
    epochs = 3
    input_shape = (28, 28, 1)

    # Build the MNIST train & test dataloader.
    (x_train, y_train), (x_test, y_test) = datasets.mnist.load_data()
    x_train = np.expand_dims(x_train.astype("float32") / 255, -1)
    x_test = np.expand_dims(x_test.astype("float32") / 255, -1)
    y_train = utils.to_categorical(y_train, num_classes).astype("float32")
    y_test = utils.to_categorical(y_test, num_classes).astype("float32")

    if args.type == "grain_dataloader":
        train_source = MNISTSource(x_train, y_train)
        test_source = MNISTSource(x_test, y_test)
        train_dataloader = grain.DataLoader(
            data_source=train_source,
            sampler=grain.samplers.IndexSampler(
                len(train_source), shuffle=True, num_epochs=1, seed=42
            ),
            operations=[grain.transforms.Batch(batch_size=batch_size)],
        )
        test_dataloader = grain.DataLoader(
            data_source=test_source,
            sampler=grain.samplers.IndexSampler(
                len(test_source), shuffle=False, num_epochs=1
            ),
            operations=[grain.transforms.Batch(batch_size=batch_size)],
        )
    elif args.type == "grain_dataset":
        train_source = MNISTSource(x_train, y_train)
        test_source = MNISTSource(x_test, y_test)
        train_dataloader = (
            grain.MapDataset.source(train_source)
            .shuffle(seed=42)
            .to_iter_dataset()
            .batch(batch_size)
        )
        test_dataloader = (
            grain.MapDataset.source(test_source)
            .to_iter_dataset()
            .batch(batch_size)
        )
    elif args.type == "pydataset":
        train_dataloader = MNISTPyDataset(
            x_train, y_train, batch_size=batch_size
        )
        test_dataloader = MNISTPyDataset(x_test, y_test, batch_size=batch_size)
    elif args.type == "tf_data":
        import tensorflow as tf

        train_dataloader = tf.data.Dataset.from_tensor_slices(
            (x_train, y_train)
        )
        train_dataloader = (
            train_dataloader.shuffle(60000).batch(batch_size).prefetch(1)
        )
        test_dataloader = tf.data.Dataset.from_tensor_slices((x_test, y_test))
        test_dataloader = test_dataloader.batch(batch_size)
    elif args.type == "torch_dataloader":
        train_dataloader = DataLoader(
            MNISTTorchDataset(x_train, y_train),
            batch_size=batch_size,
            shuffle=True,
        )
        test_dataloader = DataLoader(
            MNISTTorchDataset(x_test, y_test),
            batch_size=batch_size,
            shuffle=False,
        )
    else:
        raise ValueError(f"Unknown type: {args.type}")

    # Build a simple ConvNet.
    model = keras.Sequential(
        [
            keras.Input(shape=input_shape),
            layers.Conv2D(32, kernel_size=(3, 3), activation="relu"),
            layers.MaxPooling2D(pool_size=(2, 2)),
            layers.Conv2D(64, kernel_size=(3, 3), activation="relu"),
            layers.MaxPooling2D(pool_size=(2, 2)),
            layers.Flatten(),
            layers.Dropout(0.5),
            layers.Dense(num_classes, activation="softmax"),
        ]
    )
    model.compile(
        loss="categorical_crossentropy", optimizer="adam", metrics=["accuracy"]
    )

    # Warmup.
    print("Warmup...")
    model.fit(train_dataloader, batch_size=batch_size, epochs=3, verbose=0)
    model.evaluate(test_dataloader, verbose=0)

    # Benchmark.
    print("Benchmark...")
    time_start_fit = time.time()
    model.fit(train_dataloader, batch_size=batch_size, epochs=epochs)
    time_end_fit = time.time()
    score = model.evaluate(test_dataloader, verbose=0)
    time_end_evaluate = time.time()
    print("Test loss:", score[0])
    print("Test accuracy:", score[1])
    print(f"Fit cost time: {time_end_fit - time_start_fit:.2f} s")
    print(f"Evaluate cost time: {time_end_evaluate - time_end_fit:.2f} s")
```

</details>

## Reference

- https://google-grain.readthedocs.io/en/latest/
- https://github.com/google/grain/tree/main
- https://github.com/rcrowe-google/Learning-JAX (Thanks to @rcrowe-google for sharing the materials!)

## Additional Notes

cc @mattdangerw @divyashreepathihalli

### DIFF
--- a/keras/src/trainers/data_adapters/__init__.py
+++ b/keras/src/trainers/data_adapters/__init__.py
@@ -8,6 +8,9 @@ from keras.src.trainers.data_adapters.array_data_adapter import ArrayDataAdapter
 from keras.src.trainers.data_adapters.generator_data_adapter import (
     GeneratorDataAdapter,
 )
+from keras.src.trainers.data_adapters.grain_dataset_adapter import (
+    GrainDatasetAdapter,
+)
 from keras.src.trainers.data_adapters.py_dataset_adapter import PyDatasetAdapter
 from keras.src.trainers.data_adapters.tf_dataset_adapter import TFDatasetAdapter
 from keras.src.trainers.data_adapters.torch_data_loader_adapter import (
@@ -111,6 +114,32 @@ def get_data_adapter(
         #     "data `x` was provided as a torch DataLoader. The DataLoader "
         #     "is expected to already be shuffled."
         # )
+    elif is_grain_dataset(x):
+        if y is not None:
+            raise_unsupported_arg(
+                "y", "the targets", "grain.Dataset and grain.DataLoader"
+            )
+        if sample_weight is not None:
+            raise_unsupported_arg(
+                "sample_weights",
+                "the sample weights",
+                "grain.Dataset and grain.DataLoader",
+            )
+        if class_weight is not None:
+            raise ValueError(
+                "Argument `class_weight` is not supported for grain.Dataset "
+                f"and grain.DataLoader inputs. You can modify your "
+                "`__getitem__ ` method to return input tensor, label and "
+                "class_weight. "
+                f"Received: class_weight={class_weight}"
+            )
+        return GrainDatasetAdapter(x)
+        # TODO: should we warn or not?
+        # warnings.warn(
+        #     "`shuffle=True` was passed, but will be ignored since the "
+        #     "data `x` was provided as a grain dataset. The grain dataset "
+        #     "is expected to already be shuffled."
+        # )
     elif isinstance(x, types.GeneratorType):
         if y is not None:
             raise_unsupported_arg("y", "the targets", "PyDataset")
@@ -160,5 +189,17 @@ def is_torch_dataloader(x):
             if parent.__name__ == "DataLoader" and "torch.utils.data" in str(
                 parent.__module__
             ):
+                return True
+    return False
+
+
+def is_grain_dataset(x):
+    if hasattr(x, "__class__"):
+        for parent in x.__class__.__mro__:
+            if parent.__name__ in (
+                "MapDataset",
+                "IterDataset",
+                "DataLoader",
+            ) and "grain" in str(parent.__module__):
                 return True
     return False
--- a/keras/src/trainers/data_adapters/data_adapter.py
+++ b/keras/src/trainers/data_adapters/data_adapter.py
@@ -47,6 +47,21 @@ class DataAdapter:
         raise NotImplementedError
 
     @property
+    def builtin_prefetch(self):
+        """Whether the DataAdapter has built-in prefetching capabilities.
+
+        Prefetching is an optimization technique where data is loaded and
+        prepared in advance while the model is processing the current batch,
+        reducing training time by overlapping data loading with computation.
+
+        Returns:
+            bool: True if the DataAdapter implements its own prefetching
+            mechanism and handles data loading asynchronously. False if the
+            caller should implement prefetching externally.
+        """
+        return False
+
+    @property
     def num_batches(self):
         """Return the size (number of batches) for the dataset created.
 
--- a/keras/src/trainers/data_adapters/grain_dataset_adapter.py
+++ b/keras/src/trainers/data_adapters/grain_dataset_adapter.py
@@ -1,80 +1,66 @@
 import itertools
-import math
-import sys
 
 import numpy as np
 
 from keras.src import tree
 from keras.src.trainers.data_adapters import data_adapter_utils
 from keras.src.trainers.data_adapters.data_adapter import DataAdapter
+from keras.src.utils.module_utils import tensorflow as tf
+
+try:
+    import grain
+except ImportError:
+    grain = None
 
 
 class GrainDatasetAdapter(DataAdapter):
-    """Adapter that Grain dataset."""
+    """Adapter that handles `grain.DataLoader`, `grain.MapDataset` and
+    `grain.IterDataset`.
+    """
 
     def __init__(self, dataset):
-        import grain
+        """Initialize the GrainDatasetAdapter.
+
+        Args:
+            dataset: A Grain dataset instance. Must be one of
+                `grain.DataLoader`, `grain.MapDataset`, or `grain.IterDataset`.
+        """
 
         if not isinstance(
             dataset, (grain.MapDataset, grain.IterDataset, grain.DataLoader)
         ):
             raise ValueError(
-                "Expected argument `dataset` to be a grain.MapDataset, "
+                "Expected `dataset` to be a grain.MapDataset, "
                 "grain.IterDataset or grain.DataLoader. "
                 f"Received: {dataset} of type {type(dataset)}"
             )
 
         self._dataset = dataset
 
-        # Retrieve the information.
-        num_batches = None
-        batch_size = None
-        if isinstance(self._dataset, (grain.MapDataset, grain.IterDataset)):
-            from grain._src.python.dataset.transformations import batch
-
-            current_dataset = self._dataset
-            while True:
-                if isinstance(current_dataset, batch.BatchMapDataset):
-                    num_batches = len(current_dataset)
-                    batch_size = current_dataset._batch_size
-                    break
-                elif isinstance(current_dataset, batch.BatchIterDataset):
-                    batch_size = current_dataset._batch_size
-                    break
-                else:
-                    try:
-                        # We may receive a non-batched dataset.
-                        num_batches = len(current_dataset)
-                        break
-                    except TypeError:
-                        pass
-                if not hasattr(current_dataset, "_parent"):
-                    break
-                try:
-                    current_dataset = current_dataset._parent
-                except AssertionError:
-                    break
-        else:
-            drop_remainder = False
-            operations = self._dataset._operations
-            # Retrieve the `batch_size`.
-            for op in reversed(operations):
-                if isinstance(op, grain.transforms.Batch):
-                    batch_size = op.batch_size
-                    drop_remainder = op.drop_remainder
-                    break
-            if batch_size is not None:
-                data_source = self._dataset._data_source
-                if drop_remainder:
-                    num_batches = len(data_source) // batch_size
-                else:
-                    num_batches = math.ceil(len(data_source) / batch_size)
-        self._num_batches = num_batches
+        batch_size, output_signature = self._get_dataset_info(dataset)
         self._batch_size = batch_size
-        self._output_signature = None
+        self._output_signature = output_signature
+        self._output_tf_signature = None
+
+    def _get_dataset_info(self, dataset):
+        """Get the `batch_size` and `output_signature` from the dataset.
+
+        We use a small list of batches to infer the `batch_size` and
+        `output_signature`.
+        """
+        batches = list(
+            itertools.islice(
+                dataset, data_adapter_utils.NUM_BATCHES_FOR_TENSOR_SPEC
+            )
+        )
+        output_signature = data_adapter_utils.get_keras_tensor_spec(batches)
+        flat_output_signature = tree.flatten(output_signature)
+        batch_size = flat_output_signature[0].shape[0]
+        if batch_size is not None:
+            batch_size = int(batch_size)
+        return batch_size, output_signature
 
     def get_numpy_iterator(self):
-        import grain
         from grain._src.python.dataset.transformations import map
         from grain._src.python.shared_memory_array import (
             SharedMemoryArrayMetadata,
@@ -117,7 +103,6 @@ class GrainDatasetAdapter(DataAdapter):
         return dataset
 
     def get_jax_iterator(self):
-        import grain
         from grain._src.python.dataset.transformations import map
 
         def convert_to_jax_compatible(x):
@@ -154,8 +139,6 @@ class GrainDatasetAdapter(DataAdapter):
         return dataset
 
     def get_tf_dataset(self):
-        import grain
-        import tensorflow as tf
         from grain._src.python.dataset.transformations import map
 
         def convert_to_tf(x):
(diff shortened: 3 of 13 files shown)


keras/src/trainers/data_adapters/grain_dataset_adapter.py
```diff
@@ -0,0 +1,250 @@
...
+import math
+import sys
+
+import numpy as np
+
+from keras.src import tree
+from keras.src.trainers.data_adapters import data_adapter_utils
+from keras.src.trainers.data_adapters.data_adapter import DataAdapter
+
+
```
divyashreepathihalli, line 15: Add export and docstring

keras/src/trainers/data_adapters/__init__.py
```diff
@@ -111,6 +114,32 @@ def get_data_adapter(
...
+        if class_weight is not None:
+            raise ValueError(
+                "Argument `class_weight` is not supported for grain.Dataset "
+                f"and grain.DataLoader inputs. You can modify your "
+                "`__getitem__ ` method to return input tensor, label and "
+                "class_weight. "
+                f"Received: class_weight={class_weight}"
+            )
+        return GrainDatasetAdapter(x)
+        # TODO: should we warn or not?
```
divyashreepathihalli, line 137: A warning is good

keras/src/trainers/trainer_test.py
```diff
@@ -615,25 +644,44 @@ def test_fit_flow(self, run_eagerly, jit_compile, use_steps_per_epoch):
...
+            },
+            {
+                "testcase_name": "grain_dataloader",
+                "dataset_type": "grain_datast",
+                "dataset_kwargs": {"use_dataloader": True},
+            },
         ]
     )
     @pytest.mark.requires_trainable_backend
     def test_fit_with_data_adapter(
```
divyashreepathihalli, line 665: Curious if you ran any simple performance benchmark test to compare the new fit() against the previous one.

divyashreepathihalli (commented):
Looks great Hongyu!! Thank you for this contribution! I will let Matt take another look.

keras/src/trainers/data_adapters/data_adapter.py
```diff
@@ -46,6 +46,21 @@ def get_torch_dataloader(self):
...
+        Prefetching is an optimization technique where data is loaded and
+        prepared in advance while the model is processing the current batch,
+        reducing training time by overlapping data loading with computation.
+
+        Returns:
+            bool: True if the DataAdapter implements its own prefetching
+            mechanism and handles data loading asynchronously. False if the
+            caller should implement prefetching externally.
+        """
+        return False
```
mattdangerw, line 62: should we be setting this for tf data adapter in any way?

keras/src/trainers/data_adapters/grain_dataset_adapter.py
```diff
@@ -0,0 +1,250 @@
...
+class GrainDatasetAdapter(DataAdapter):
+    """Adapter that Grain dataset."""
+
+    def __init__(self, dataset):
+        import grain
```
mattdangerw: probably could do this as a toplevel import with a catch import error, or use this https://github.com/keras-team/keras/blob/55263ca64c1ea746bf1396c4a3e1b91077d3e4ef/keras/src/utils/module_utils.py#L42-L60

keras/src/trainers/data_adapters/grain_dataset_adapter.py
```diff
@@ -0,0 +1,250 @@
...
+                    try:
+                        # We may receive a non-batched dataset.
+                        num_batches = len(current_dataset)
+                        break
+                    except TypeError:
+                        pass
+                if not hasattr(current_dataset, "_parent"):
+                    break
+                try:
+                    current_dataset = current_dataset._parent
```
mattdangerw: curious you have to walk an object hierarchy like this. there's not grain utilities that will do this for us?

keras/src/trainers/data_adapters/grain_dataset_adapter.py
```diff
@@ -0,0 +1,250 @@
...
+                    num_batches = len(data_source) // batch_size
+                else:
+                    num_batches = math.ceil(len(data_source) / batch_size)
+        self._num_batches = num_batches
+        self._batch_size = batch_size
+        self._output_signature = None
+
+    def get_numpy_iterator(self):
+        import grain
+        from grain._src.python.dataset.transformations import map
```
mattdangerw: if there's any way to avoid these worth doing, since these might break anytime. wonder if we can get someone from grain to take a look here.

keras/src/trainers/data_adapters/__init__.py
fchollet: We're not doing it for PyTorch dataloaders though -- we should be consistent. Do note that `shuffle=True` is the default, so the warning would show up all the time.

IMO keep the current status quo. If we want to change in the future, do it across the board.

james77777778:
I'm working on resolving the comments and some questions from @mattdangerw might need input from the Grain team.

Should I ask them on github or do you have internal contacts?

keras/src/trainers/data_adapters/grain_dataset_adapter.py
james77777778: Do we want to export `GrainDatasetAdapter`?
Other adapters are not exposed.

mattdangerw:
@james77777778 already asked someone on the Grain team to look, should give feedback on this PR soon.

keras/src/trainers/data_adapters/grain_dataset_adapter.py
iindyk: yeah, this digs into implementation details a bit.

Taking a step back, are both batch_size and num_batches required? 

the logic included here will only work for a narrow subset of cases, but will be incorrect otherwise. e.g. if there's a filter or packing transformation (packing will be used in every gen ai use case), then the length of the dataset is not known at construction time since it depends on runtime information. Additionally, in most cases, the dataset will be repeated a number of times, so the real length will be ~infinite

keras/src/trainers/data_adapters/grain_dataset_adapter.py
```diff
@@ -0,0 +1,250 @@
...
+                x = data_adapter_utils.scipy_sparse_to_jax_sparse(x)
+            elif data_adapter_utils.is_tensorflow_sparse(x):
+                x = data_adapter_utils.tf_sparse_to_jax_sparse(x)
+            return x
+
+        class ConvertToJaxCompatible(grain.transforms.Map):
+            def map(self, x):
+                return tree.map_structure(convert_to_jax_compatible, x)
+
+        if isinstance(self._dataset, grain.MapDataset):
```
iindyk: self._dataset = self._dataset.map(ConvertToJaxCompatible())

keras/src/trainers/data_adapters/grain_dataset_adapter.py
```diff
@@ -0,0 +1,250 @@
...
+        )
+
+        def convert_to_numpy(x):
+            if isinstance(x, (np.ndarray, SharedMemoryArrayMetadata)):
+                return x
+            else:
+                # Using `__array__` should handle `tf.Tensor`, `jax.np.ndarray`,
+                # `torch.Tensor`, as well as any other tensor-like object that
+                # has added numpy support.
+                if hasattr(x, "__array__"):
```
iindyk, line 75: seems like you only need this branch, else return x?

keras/src/trainers/data_adapters/grain_dataset_adapter.py
```diff
@@ -0,0 +1,250 @@
...
+                + [ConvertToTF(), ListToTuple()],
+                worker_count=self._dataset._multiprocessing_options.num_workers,
+                worker_buffer_size=self._dataset._multiprocessing_options.per_worker_buffer_size,
+                shard_options=self._dataset._shard_options,
+                read_options=self._dataset._read_options,
+                enable_profiling=self._dataset._multiprocessing_options.enable_profiling,
+            )
+
+        if self._output_signature is None:
+            batches = list(
```
iindyk: since you produce several elements to infer spec it makes sense to get the batch size at the same time?

iindyk (commented):
hi, grain dev here. Thanks for the integration!

keras/src/trainers/data_adapters/grain_dataset_adapter.py
james77777778: I didn't do that because `self._dataset.map(...)` will change the `self._dataset`, which should be a base element fed to all `get_*` methods.

Consider two scenarios:
- Call `get_jax_iterator` twice. We will add two map operations to the dataset.
- Call `get_jax_iterator` and then call `get_tf_dataset`. We will add redundant map operations for the dataset. It might not cause errors for now but it's not an optimal situation.

james77777778: Would it be possible to expose `map` as a standalone function?

Currently, in Grain, users are required to use `dataset.map(...)`. An alternative like `new_dataset = map(dataset, ...)` (where the original dataset isn't modified) might benefit some use cases including this PR's scenario.

keras/src/trainers/data_adapters/grain_dataset_adapter.py
james77777778: Good point. This path might be more robust than the hard-coded one in `__init__`. Will try it out.

keras/src/trainers/data_adapters/data_adapter.py
james77777778: Set to `True` for `TFDatasetAdapter` and infer the value for `TorchDataLoaderAdapter`.

keras/src/trainers/data_adapters/grain_dataset_adapter.py
james77777778: Added as a toplevel import.

keras/src/trainers/data_adapters/grain_dataset_adapter.py
james77777778: Changed the impl by using `get_keras_tensor_spec` to infer `batch_size` and `output_signature`.
`num_batches` will be `None`.

keras/src/trainers/trainer_test.py
james77777778: I’ve added a table benchmarking various dataloaders on the MNIST dataset in the PR description:

https://github.com/keras-team/keras/pull/21494#issue-3247982551

keras/src/trainers/data_adapters/grain_dataset_adapter.py
iindyk: `self._dataset.map` does not mutate `self._dataset`, it does exactly what you're doing: https://github.com/google/grain/blob/d2a19d9955ff78d4fbf8fdc563df3ee86bd3b72a/grain/_src/python/dataset/dataset.py#L489

james77777778: Ooh, I misunderstood something before. Now, they are corrected. Thanks!

james77777778:
Should be ready for another round of review.

PR #21499, opened 2025-07-22
Fix RandAugment behavior in tf graph mode

shashaka:
Fix for [keras-team/keras#21169](https://github.com/keras-team/keras/issues/21169)

- Modified the logic to update the sample image processing algorithm dynamically for each batch during TensorFlow graph execution.


fchollet (approved):
Nice, thanks for the fix!

Issue #21378, opened 2025-06-12
Keras Fails to load quantized model


pctablet505:
Whenever we try to save a keras-hub model after quantization, we are unable to load the quantized model. I've tried `from_preset()` method for that model, and also `keras.models.load_model` nothing works.

I've attached notebook
https://colab.research.google.com/gist/pctablet505/b5ef8ab36dceb58527e992b571aefb70/keras-quantized-model-not-loading.ipynb

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
[<ipython-input-5-3241078411>](https://localhost:8080/#) in <cell line: 0>()
      8 import keras_hub
      9 
---> 10 gemma_lm_quantized = keras_hub.models.Gemma3CausalLM.from_preset("stored_gemma_int8")
     11 gemma_lm_quantized.generate("hello, what is your name?")

4 frames
[/content/keras_hub_repo/keras_hub/src/models/task.py](https://localhost:8080/#) in from_preset(cls, preset, load_weights, **kwargs)
    196         # images, audio).
    197         load_task_weights = "num_classes" not in kwargs
--> 198         return loader.load_task(cls, load_weights, load_task_weights, **kwargs)
    199 
    200     def load_task_weights(self, filepath):

[/content/keras_hub_repo/keras_hub/src/utils/preset_utils.py](https://localhost:8080/#) in load_task(self, cls, load_weights, load_task_weights, **kwargs)
    701             else:
    702                 jax_memory_cleanup(task.backbone)
--> 703             self._load_backbone_weights(task.backbone)
    704         return task
    705 

[/content/keras_hub_repo/keras_hub/src/utils/preset_utils.py](https://localhost:8080/#) in _load_backbone_weights(self, backbone)
    754                 # Download the sharded weights.
    755                 _ = get_file(self.preset, sharded_filename)
--> 756         backbone.load_weights(filepath)
    757 
    758 

[/content/keras_repo/keras/src/utils/traceback_utils.py](https://localhost:8080/#) in error_handler(*args, **kwargs)
    120             # To get the full stack trace, call:
    121             # `keras.config.disable_traceback_filtering()`
--> 122             raise e.with_traceback(filtered_tb) from None
    123         finally:
    124             del filtered_tb

[/content/keras_repo/keras/src/saving/saving_lib.py](https://localhost:8080/#) in _raise_loading_failure(error_msgs, warn_only)
    648         warnings.warn(msg)
    649     else:
[...]

PR #21504, opened 2025-07-23
fix quantization save and load error

divyashreepathihalli:
Fixes : https://github.com/keras-team/keras/issues/21378

When layers are quantized, their state changes. This was causing the Keras saving and loading to fail.

### DIFF
--- a/keras/src/layers/layer.py
+++ b/keras/src/layers/layer.py
@@ -1382,7 +1382,7 @@ class Layer(BackendLayer, Operation):
     def _get_quantized_variables(self):
         quantized_vars = []
         for v in self._trainable_variables + self._non_trainable_variables:
-            if not backend.is_float_dtype(v.dtype):
+            if getattr(v, "_is_quantized", False):
                 quantized_vars.append(v)
         return quantized_vars
 
--- a/keras/src/layers/core/dense.py
+++ b/keras/src/layers/core/dense.py
@@ -345,12 +345,14 @@ class Dense(Layer):
             dtype="int8",
             trainable=False,
         )
+        self._kernel._is_quantized = True
         self.kernel_scale = self.add_weight(
             name="kernel_scale",
             shape=(self.units,),
             initializer="ones",
             trainable=False,
         )
+        self.kernel_scale._is_quantized = True
 
     def _int4_build(self, kernel_shape):
         """Build variables for int4 quantization.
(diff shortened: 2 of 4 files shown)


keras/src/layers/layer.py
```diff
@@ -1359,10 +1359,33 @@ def save_own_variables(self, store):
...
+        non_quantized_vars = [
+            v for v in all_vars if v not in quantized_vars_set
+        ]
+        for i, v in enumerate(non_quantized_vars):
             store[f"{i}"] = v
 
+    def _get_quantized_variables(self):
+        quantized_vars = []
+        for v in self._trainable_variables + self._non_trainable_variables:
+            if not backend.is_float_dtype(v.dtype):
```
hertschuh: This assumes that all integral variables come from quantization. But what if you have variables that intrinsically represent ints and are unrelated to quantization? We definitely have layers using int vars.

divyashreepathihalli:
The issue is that Keras's standard save-and-load function relies on a basic, index-based system to match up the variables. This process completely breaks for a quantized model because the number and names of the variables no longer match what the saved file expects.

So, when Keras tried to load the model, it was looking for the original variables but found the new, quantized ones instead.

keras/src/layers/layer.py
divyashreepathihalli: changed this to check for `_is_quantized` instead

hertschuh:
So the unit test you have passes without any of your changes. Which is consistent with this comment: https://github.com/keras-team/keras/issues/21378#issuecomment-3043643453
I don't think this is addressing the issue.

Have you tested with Gemma?

> The issue is that Keras's standard save-and-load function relies on a basic, index-based system to match up the variables. This process completely breaks for a quantized model because the number and names of the variables no longer match what the saved file expects.

But your logic relies on having the right number of variables in *some* order. All it's changing is the order. If the `kernel_scale` variable is missing in the dense layer, it's still going to fail.

> So, when Keras tried to load the model, it was looking for the original variables but found the new, quantized ones instead.

That makes me think that the issue has to do with turning on ("building") the quantization before reloading the variables.

keras/src/layers/layer.py
```diff
@@ -1359,10 +1359,33 @@ def save_own_variables(self, store):
         Args:
             store: Dict where the state of the model will be saved.
         """
+        if not getattr(self, "_is_quantized", False):
+            all_vars = self._trainable_variables + self._non_trainable_variables
+            for i, v in enumerate(all_vars):
+                store[f"{i}"] = v
+            return
+
+        # Case: quantized layer
```
fchollet, line 1368: Can you invert the `if`? `If getattr(self, "_is_quantized", False):` ... (more readable)

JyotinderSingh:
This was fixed in https://github.com/keras-team/keras/pull/21608 and https://github.com/keras-team/keras-hub/pull/2367

Issue #19226, opened 2024-02-25
TorchModuleWrapper default `get_config` creates non serializable objects


gilfree:
Hi, and thanks to all keras-team :)

The `get_config` of `TorchModuleWrapper` uses `BytesIO` and `torch.save` to create a byte array:

https://github.com/keras-team/keras/blob/1137074a9a1c237473f2fe57ab277c697892c6f1/keras/utils/torch_utils.py#L140

In the serialization lib, `bytes` object are then decoded to utf-8:
https://github.com/keras-team/keras/blob/1137074a9a1c237473f2fe57ab277c697892c6f1/keras/saving/serialization_lib.py#L154

An arbitrary byte array, as created by `torch.save` may fail decoding as it is not a valid utf-8 string, and will cause an exception on saving.

I think that the real issue here is the decoding of arbitrary bytes to a string, but I assume this is something that will be a hard to fix as it will create format compatibility issues, as it seems to date back a few years ago to keras 2.

Maybe the torch model buffer can be dumped to a json string and read from it by default, or dumped to `str` and then use `ast.literal_eval` or some other safe binary<->string conversion.

SuryanarayanaY:
Hi @gilfree ,
Thanks for reporting. It would be helpful if you could submit a minimal reproducible snippet to check the behaviour. Thanks!

gilfree:
To reproduce, add the following to the file: keras/utils/torch_utils_test.py:

```python
 def test_save_load(self):
     import keras
     class M(keras.Model):
         def __init__(self,channels=10, **kwargs):
             super().__init__()
             self.sequence = torch.nn.Sequential(
                 torch.nn.Conv2d(1, channels, kernel_size=(3, 3)),
             )
         def call(self, x):
             return self.sequence(x)
            
         def get_config(self):
             return self.sequence.get_config()
     m = M()
     x=torch.ones((10,1,28, 28))
     m(x)
     m.save('model.keras')
```

Run with:

```
CUDA_VISIBLE_DEVICES=-1 KERAS_BACKEND=torch pytest -vvv -vs keras/utils/torch_utils_test.py -k save_load 
```

This is an artificial example as one will probably not write this get config, just to demonstrate the issue in the shortest way I have found.

SuryanarayanaY:
Hi @gilfree ,

I have added the code snippet you provided to the file: `keras/utils/torch_utils_test.py` and replicated the reported error. Attached logs below.

```
[...]

PR #21505, opened 2025-07-23
fix torch module wrapper serialization error

divyashreepathihalli:
Fixes : https://github.com/keras-team/keras/issues/19226

The `get_config` method of `TorchModuleWrapper` was attempting to decode a byte array as a UTF-8 string, which was causing an error.

### DIFF
--- a/keras/src/saving/serialization_lib.py
+++ b/keras/src/saving/serialization_lib.py
@@ -164,11 +164,9 @@ def serialize_keras_object(obj):
 
     # Special cases:
     if isinstance(obj, bytes):
-        import base64
-
         return {
             "class_name": "__bytes__",
-            "config": {"value": base64.b64encode(obj).decode("utf-8")},
+            "config": {"value": obj.decode("utf-8")},
         }
     if isinstance(obj, slice):
         return {
@@ -634,9 +632,7 @@ def deserialize_keras_object(
     if class_name == "__numpy__":
         return np.array(inner_config["value"], dtype=inner_config["dtype"])
     if config["class_name"] == "__bytes__":
-        import base64
-
-        return base64.b64decode(inner_config["value"].encode("utf-8"))
+        return inner_config["value"].encode("utf-8")
     if config["class_name"] == "__ellipsis__":
         return Ellipsis
     if config["class_name"] == "__slice__":
--- a/keras/src/saving/serialization_lib.py
+++ b/keras/src/saving/serialization_lib.py
@@ -164,9 +164,11 @@ def serialize_keras_object(obj):
 
     # Special cases:
     if isinstance(obj, bytes):
+        import base64
+
         return {
             "class_name": "__bytes__",
-            "config": {"value": obj.decode("utf-8")},
+            "config": {"value": base64.b64encode(obj).decode("utf-8")},
         }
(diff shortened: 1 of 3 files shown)


keras/src/saving/serialization_lib.py
```diff
@@ -632,7 +634,9 @@ class ModifiedMeanSquaredError(keras.losses.MeanSquaredError):
     if class_name == "__numpy__":
         return np.array(inner_config["value"], dtype=inner_config["dtype"])
     if config["class_name"] == "__bytes__":
-        return inner_config["value"].encode("utf-8")
+        import base64
+
+        return base64.b64decode(inner_config["value"].encode("utf-8"))
```
hertschuh: This is not backwards compatible. Any model serialized before this change and deserialized after will fail to deserialize. The fix needs to be a lot narrower and only affect the one `buffer.getvalue()` from `TorchModuleWrapper`.

hertschuh: That being said, the fix is correct in the sense that the issue that `TorchModuleWrapper` is facing is an issue that can affect any other byte buffer that we serialize.

The baked in assumption of the current code is that any byte buffer represents a utf8 sequence, which is kind of weird. What I don't know if how much this thing is used.

divyashreepathihalli: moved the changes to the wrapper

keras/src/utils/torch_utils.py
```diff
@@ -149,21 +149,27 @@ def compute_output_shape(self, input_shape):
     def get_config(self):
         base_config = super().get_config()
         import torch
+        import base64
 
         buffer = io.BytesIO()
         torch.save(self.module, buffer)
+        # Encode the buffer using base64 to ensure safe serialization
+        buffer_b64 = base64.b64encode(buffer.getvalue()).decode("utf-8")
```
hertschuh: nit: the base 64 encoded buffer only uses code points < 128, so UTF8 encoding it is not going to do anything. It looks like what people do is `.decode('ascii')`.

keras/src/utils/torch_utils.py
```diff
@@ -149,21 +149,27 @@ def compute_output_shape(self, input_shape):
...
 
     @classmethod
     def from_config(cls, config):
         import torch
+        import base64
 
         if "module" in config:
-            buffer = io.BytesIO(config["module"])
+            # Decode the base64 string back to bytes
+            buffer_bytes = base64.b64decode(config["module"].encode("utf-8"))
```
hertschuh: Same here `.encode('ascii')`.

keras/src/utils/torch_utils.py
```diff
@@ -149,21 +149,27 @@ def compute_output_shape(self, input_shape):
     def get_config(self):
         base_config = super().get_config()
         import torch
+        import base64
```
hertschuh: I don't see an upside to importing here vs. at the top.

hertschuh:
Thank you for the fix!

MicheleCattaneo:
Thanks a lot for this fix! It's been a longstanding issue for me, and I'm glad to see it resolved.
FYI @dnerini

PR #21514, opened 2025-07-25
Fix Discretization layer graph mode bug

divyashreepathihalli:


fchollet:
Thanks for the PR -- please take a look at the test failure:

> FAILED keras/src/layers/preprocessing/discretization_test.py::DiscretizationTest::test_discretization_basics - AssertionError: expected output dtype int64, got float16

keras/src/layers/preprocessing/discretization.py
```diff
@@ -96,8 +96,7 @@ def __init__(
         name=None,
     ):
         if dtype is None:
-            dtype = "int64" if output_mode == "int" else backend.floatx()
-
+            dtype = "float32"
```
hertschuh: I think you should just remove the whole `if dtype is None` block, the base layer class will handle `None` properly.

divyashreepathihalli: I tried that! It results in the unit tests failing 
```
FAILED keras/src/layers/preprocessing/discretization_test.py::DiscretizationTest::test_discretization_basics - AssertionError: expected output dtype int64, got int32:
- int32
+ int64
```

divyashreepathihalli:
Pytorch GPU error is unrelated

keras/src/layers/preprocessing/discretization.py
hertschuh: Well, it fails the same way with or without on JAX.

The thing is, this class has this:

```python
    @property
    def input_dtype(self):
        return backend.floatx()
```

So I don't understand why the inputs were cast to ints.

Maybe you can try to override `@property ... compute_dtype`?

takumiohym:
Hi @divyashreepathihalli, any update on this PR?

fchollet:
@divyashreepathihalli Divya, are you still working on this?

hertschuh:
Closing as it was addressed in https://github.com/keras-team/keras/pull/21769

PR #21526, opened 2025-07-28
[OpenVINO backend] fix openvino model exported names to match keras names

Mohamed-Ashraf273:
@fchollet 
@rkazants


fchollet (approved):
LGTM, thank you

Issue #21530, opened 2025-07-30
Custom call context args are not supported for the common methods for model calling (e.g. fit, predict, evaluate, ...)


RyanSaxe:
Keras internals expose a method on layers to register custom arguments: [_register_call_context_args](https://github.com/keras-team/keras/blob/7bf852c211ea41dc58bbc740c3fe64d091d37511/keras/src/layers/layer.py#L1737)

The example in the documentation looks like this:

> NOTE: the code shared below from current version of keras has a typo that you should also consider fixing: `seq.register_call_context_args("foo_mode")` should be `seq._register_call_context_args("foo_mode")`. The underscore missed means the code example does not run.

```python
class Inner(layers.Layer):

    def __init__(self):
        super().__init__()
        # Register `foo_mode` as a call-context arg
        self._register_call_context_args("foo_mode")

    def call(self, x, foo_mode=False):
        # If foo_mode=True add 1, otherwise add 0
        add_val = ops.where(foo_mode, 1.0, 0.0)
        return x + add_val

class Outer(layers.Layer):
    def __init__(self):
        super().__init__()
        self.inner = Inner()

    def call(self, x):
        # We don't explicitly pass foo_mode here—Base Layer.__call__
        # should inject it into `self.inner`
        return self.inner(x)

sample_input = np.array([[1.0], [2.0]])

# Sequential model
seq = models.Sequential([Outer()])

# Tell the Sequential model to propagate foo_mode down
# the call-stack
seq.register_call_context_args("foo_mode")

# foo_mode=True -> input + 1
out_true = seq(sample_input, foo_mode=True)
```

However, it seems like it is only possible to specify `foo_mode` when calling the model directly via `__call__`. If I were to try any of the commonly exposed KPIs, I will get an error:

```python
seq.predict(sample_input, foo_mode=True)
```

```python-traceback
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[10], line 1
----> 1 seq.predict(sample_input, foo_mode=True)

File [~/.../.venv/lib/python3.12/site-packages/keras/src/utils/traceback_utils.py:122](~/.../.venv/lib/python3.12/site-packages/keras/src/utils/traceback_utils.py#line=121), in filter_traceback.<locals>.error_handler(*args, **kwargs)
    119     filtered_tb = _process_traceback_frames(e.__traceback__)
    120     # To get the full stack trace, call:
    121     # `keras.config.disable_traceback_filtering()`
[...]

PR #21543, opened 2025-08-01
Add public register_call_context_args method to Layer

slopisgood:
Add a public wrapper for _register_call_context_args on keras.layers.Layer that forwards call-context argument names. This makes the documented example using seq.register_call_context_args(...) functional and resolves issue #21530.


hertschuh:
Hi @slopisgood 

Just to understand, what is the context for this? Why is this needed? What is the use case? `_register_call_context_args` can already be called.

Thanks!

PR #21545, opened 2025-08-02
[OpenVINO backend] fix __getitem__ and convert_to_tensor issues

Mohamed-Ashraf273:
@fchollet
I found out that `OpenVINO` backend handles some Tensors incorrectly:
Example:
https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/utils/tensor_utils.py#L161
When `x=tf.Tensor`, it returns `ov.Tensor`, which is incompatible with keras ops
I took torch approach to handle that:
https://github.com/keras-team/keras/blob/master/keras/src/backend/torch/core.py#L219-L221

Additionally, there are some missing `f`'s in `__getitem__`, I added them.

### DIFF
--- a/keras/src/backend/openvino/core.py
+++ b/keras/src/backend/openvino/core.py
@@ -654,20 +654,18 @@ def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
         if dtype and dtype != x.dtype:
             x = cast(x, dtype)
         return x
+    original_type = type(x)
     try:
         if dtype is None:
-            dtype = getattr(x, "dtype", type(x))
-            if dtype == np.dtype("bfloat16"):
-                ov_type = OPENVINO_DTYPES["bfloat16"]
-            else:
-                ov_type = OPENVINO_DTYPES[standardize_dtype(dtype)]
+            dtype = getattr(x, "dtype", original_type)
+            ov_type = OPENVINO_DTYPES[standardize_dtype(dtype)]
         else:
             ov_type = OPENVINO_DTYPES[dtype]
         x = np.array(x)
         return OpenVINOKerasTensor(ov_opset.constant(x, ov_type).output(0))
     except Exception as e:
         raise TypeError(
-            f"Cannot convert object of type {type(x)} "
+            f"Cannot convert object of type {original_type} "
             f"to OpenVINOKerasTensor: {e}"
         )
 
--- a/keras/src/backend/openvino/core.py
+++ b/keras/src/backend/openvino/core.py
@@ -661,6 +661,8 @@ def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
                 ov_type = OPENVINO_DTYPES["bfloat16"]
             else:
                 ov_type = OPENVINO_DTYPES[standardize_dtype(dtype)]
+        else:
+            ov_type = OPENVINO_DTYPES[dtype]
         x = np.array(x)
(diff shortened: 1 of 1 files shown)


Mohamed-Ashraf273:
@fchollet

keras/src/backend/openvino/core.py
```diff
@@ -654,13 +654,20 @@ def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
...
-            *[getattr(item, "dtype", type(item)) for item in tree.flatten(x)]
+    original_type = type(x)
+    try:
+        if dtype is None:
+            dtype = getattr(x, "dtype", original_type)
+            ov_type = OPENVINO_DTYPES[standardize_dtype(dtype)]
+        else:
+            ov_type = OPENVINO_DTYPES[dtype]
+        x = np.array(x)
+        return OpenVINOKerasTensor(ov_opset.constant(x, ov_type).output(0))
```
fchollet, line 665: This seems very different from the prior behavior -- returning a symbolic tensor instead of a concrete tensor. Is this the right behavior? Do you have a unit test to check it?

keras/src/backend/openvino/core.py
```diff
@@ -654,13 +654,20 @@ def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
...
+    except Exception as e:
```
fchollet, line 666: Could this exception catching be less broad, more targeted? When would it be triggered?

fchollet (commented):
Thanks for the PR!

keras/src/backend/openvino/core.py
Mohamed-Ashraf273: This exception should be triggered if the object passed to `convert_to_tensor` can’t be converted to a NumPy array.

keras/src/backend/openvino/core.py
Mohamed-Ashraf273: All operations in Keras support only `OpenVINOKerasTensor`.
Additionally, at [this line in tensor_utils.py](https://github.com/keras-team/keras-hub/blob/master/keras_hub/src/utils/tensor_utils.py#L161), the function returns an `ov.Tensor` due to receiving `tf.Tensor`, which is incompatible with the downstream operations it will pass through.

Also, all tests are passing in this PR:
https://github.com/keras-team/keras-hub/pull/2350

Moreover, if an `OpenVINOKerasTensor` originates from real input data. In that case, it can still be safely converted back to a NumPy array when needed, if it can't be converted, then it will raise an error.

Additionally, all backends do the same thing:
https://github.com/keras-team/keras/blob/master/keras/src/backend/torch/core.py#L219-L221

Mohamed-Ashraf273: Here is a unit test:
```python
import os
os.environ['KERAS_BACKEND'] = 'openvino'

import tensorflow as tf
from keras import ops
const = tf.constant([[1, 2, 3], [4, 5, 6]])
ov_tensor = ops.convert_to_tensor(const)
print(ov_tensor)
print(ov_tensor.numpy())
```
```
<keras.src.backend.openvino.core.OpenVINOKerasTensor object at 0x730938ac66f0>
[[1 2 3]
 [4 5 6]]
```

Mohamed-Ashraf273: Before fix, it was returning an `ov.Tensor`, which is incompatible with keras ops.

fchollet (approved):
LGTM, thanks for the explanation

PR #21548, opened 2025-08-04
Fix support for optional inputs in model.fit

neo-alex:
Here is an example of a model with 2 inputs, the second one being optional:

```python
class OptionalInputLayer(layers.Layer):
    def __init__(self):
        super().__init__()
        self.dense = layers.Dense(2)

    def call(self, x, y=None):
        z = x if y is None else x + y
        return self.dense(z)

    def compute_output_shape(self, x_shape):
        return x_shape

i1 = Input((2,), name="input1")
i2 = Input((2,), name="input2", optional=True)
outputs = OptionalInputLayer()(i1, i2)
model = Model({"input1": i1, "input2": i2}, outputs)
```

With this definition, the model can be called in Jax/TF/Torch without issue, even when input2 is None:

```python
model({"input1": np.ones((2, 2)), "input2": None})  # WORKS
model.predict_on_batch({"input1": np.ones((2, 2)), "input2": None})  # WORKS AS WELL
```

It is even possible to train **on a batch** when input2 is None:

```python
model.compile(loss=losses.MeanSquaredError)
model.train_on_batch(x={"input1": np.ones((2, 2)), "input2": None}, y=np.ones((2, 2)))  # WORKS
```

However, doing the same with model.fit API is currently failing on all backends:

```python
# Without generator
model.fit(x={"input1": np.ones((2, 2)), "input2": None}, y=np.ones((2, 2)))  # DOESN'T WORK

# With generator
data_generator = (({"input1": np.ones((2, 2)), "input2": None}, np.ones((2, 2))) for _ in range(3))
model.fit(x=data_generator)  # DOESN'T WORK EITHER
```

The purpose of this PR is to fix this issue (on Jax/TF/Torch), so that the last code block above becomes possible (btw. I could add 1 or 2 unit tests along those lines to demonstrate the fix but I am unsure where would be the best place for it... [model_test](https://github.com/keras-team/keras/blob/master/keras/src/models/model_test.py) maybe?).

### DIFF
--- a/keras/src/backend/tensorflow/trainer.py
+++ b/keras/src/backend/tensorflow/trainer.py
@@ -51,11 +51,7 @@ class TensorFlowTrainer(base_trainer.Trainer):
 
     def train_step(self, data):
         x, y, sample_weight = data_adapter_utils.unpack_x_y_sample_weight(data)
-
-        # Convert TF Optional implementations to None
-        x = tree.map_structure(
-            lambda i: None if isinstance(i, tf.experimental.Optional) else i, x
-        )
+        x = self._convert_optional_to_none(x)
 
         # Forward pass
         with tf.GradientTape() as tape:
@@ -91,6 +87,7 @@ class TensorFlowTrainer(base_trainer.Trainer):
 
     def test_step(self, data):
         x, y, sample_weight = data_adapter_utils.unpack_x_y_sample_weight(data)
+        x = self._convert_optional_to_none(x)
         if self._call_has_training_arg:
             y_pred = self(x, training=False)
         else:
@@ -106,11 +103,18 @@ class TensorFlowTrainer(base_trainer.Trainer):
 
     def predict_step(self, data):
         x, _, _ = data_adapter_utils.unpack_x_y_sample_weight(data)
+        x = self._convert_optional_to_none(x)
         if self._call_has_training_arg:
             y_pred = self(x, training=False)
         else:
             y_pred = self(x)
         return y_pred
+
+    def _convert_optional_to_none(self, x):
+        # Convert TF Optional implementations to None
+        return tree.map_structure(
+            lambda i: None if isinstance(i, tf.experimental.Optional) else i, x
+        )
 
     def _make_function(self, step_function):
         @tf.autograph.experimental.do_not_convert
--- a/keras/src/backend/tensorflow/trainer.py
+++ b/keras/src/backend/tensorflow/trainer.py
@@ -1,4 +1,5 @@
 import contextlib
+import functools
 import warnings
 
 import numpy as np
@@ -51,7 +52,6 @@ class TensorFlowTrainer(base_trainer.Trainer):
 
     def train_step(self, data):
         x, y, sample_weight = data_adapter_utils.unpack_x_y_sample_weight(data)
-        x = self._convert_optional_to_none(x)
 
         # Forward pass
         with tf.GradientTape() as tape:
@@ -87,7 +87,6 @@ class TensorFlowTrainer(base_trainer.Trainer):
 
     def test_step(self, data):
         x, y, sample_weight = data_adapter_utils.unpack_x_y_sample_weight(data)
-        x = self._convert_optional_to_none(x)
         if self._call_has_training_arg:
             y_pred = self(x, training=False)
         else:
@@ -103,18 +102,26 @@ class TensorFlowTrainer(base_trainer.Trainer):
 
     def predict_step(self, data):
         x, _, _ = data_adapter_utils.unpack_x_y_sample_weight(data)
-        x = self._convert_optional_to_none(x)
         if self._call_has_training_arg:
             y_pred = self(x, training=False)
         else:
             y_pred = self(x)
         return y_pred
 
-    def _convert_optional_to_none(self, x):
-        # Convert TF Optional implementations to None
-        return tree.map_structure(
-            lambda i: None if isinstance(i, tf.experimental.Optional) else i, x
-        )
+    def _autoconvert_optionals(self, step_func):
+        # Wrapper converting (nested) TF Optional in input data to None
+        @functools.wraps(step_func)
+        def wrapper(data):
+            converted_data = tree.map_structure(
+                lambda i: (
+                    None if isinstance(i, tf.experimental.Optional) else i
+                ),
+                data,
+            )
+            result = step_func(converted_data)
+            return result
+
+        return wrapper
 
     def _make_function(self, step_function):
         @tf.autograph.experimental.do_not_convert
@@ -134,6 +141,7 @@ class TensorFlowTrainer(base_trainer.Trainer):
                 reduce_retracing=True,
                 jit_compile=self.jit_compile,
             )
+        one_step_on_data = self._autoconvert_optionals(one_step_on_data)
 
         @tf.autograph.experimental.do_not_convert
         def multi_step_on_iterator(iterator):
@@ -262,6 +270,7 @@ class TensorFlowTrainer(base_trainer.Trainer):
             one_step_on_data = tf.function(
                 one_step_on_data, reduce_retracing=True, jit_compile=True
             )
+        one_step_on_data = self._autoconvert_optionals(one_step_on_data)
 
         @tf.autograph.experimental.do_not_convert
         def one_step_on_data_distributed(data):
--- a/keras/src/backend/tensorflow/trainer.py
+++ b/keras/src/backend/tensorflow/trainer.py
@@ -52,6 +52,11 @@ class TensorFlowTrainer(base_trainer.Trainer):
     def train_step(self, data):
         x, y, sample_weight = data_adapter_utils.unpack_x_y_sample_weight(data)
 
+        # Convert TF Optional implementations to None
+        x = tree.map_structure(
+            lambda i: None if isinstance(i, tf.experimental.Optional) else i, x
+        )
+
         # Forward pass
         with tf.GradientTape() as tape:
             if self._call_has_training_arg:
--- a/keras/src/trainers/data_adapters/data_adapter_utils.py
+++ b/keras/src/trainers/data_adapters/data_adapter_utils.py
@@ -101,7 +101,9 @@ def list_to_tuple(maybe_list):
 
 
 def check_data_cardinality(data):
-    num_samples = set(int(i.shape[0]) for i in tree.flatten(data))
+    num_samples = set(
+        int(i.shape[0]) for i in tree.flatten(data) if i is not None
+    )
     if len(num_samples) > 1:
         msg = (
             "Data cardinality is ambiguous. "
@@ -186,7 +188,9 @@ def get_keras_tensor_spec(batches):
         else:
             return backend.KerasTensor(shape=shape, dtype=dtype)
 
-    return tree.map_structure(get_single_tensor_spec, *batches)
+    return tree.map_structure(
+        get_single_tensor_spec, *batches, none_is_leaf=False
+    )
 
 
 def convert_to_tf_tensor_spec(keras_tensor, batch_axis_to_none=True):
@@ -199,6 +203,8 @@ def convert_to_tf_tensor_spec(keras_tensor, batch_axis_to_none=True):
     """
     from keras.src.utils.module_utils import tensorflow as tf
 
+    if keras_tensor is None:
+        return tf.OptionalSpec(None)
     if not isinstance(keras_tensor, backend.KerasTensor):
         raise TypeError(
             f"Expected a KerasTensor, but got {keras_tensor} of type "
@@ -252,7 +258,9 @@ def get_jax_iterator(iterable):
             return np.asarray(x)
 
     for batch in iterable:
-        yield tree.map_structure(convert_to_jax_compatible, batch)
+        yield tree.map_structure(
+            convert_to_jax_compatible, batch, none_is_leaf=False
+        )
 
 
 def get_numpy_iterator(iterable):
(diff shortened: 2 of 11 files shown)


keras/src/backend/tensorflow/trainer.py
```diff
@@ -52,6 +52,11 @@ def distribute_reduction_method(self, value):
     def train_step(self, data):
         x, y, sample_weight = data_adapter_utils.unpack_x_y_sample_weight(data)
 
+        # Convert TF Optional implementations to None
+        x = tree.map_structure(
+            lambda i: None if isinstance(i, tf.experimental.Optional) else i, x
+        )
```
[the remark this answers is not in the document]
neo-alex: Good point! I just fixed it in [this new commit](https://github.com/keras-team/keras/pull/21548/commits/0c9f605b4821b40061514d7c6680e749e1285bb1).

keras/src/trainers/data_adapters/generator_data_adapter.py
```diff
@@ -32,6 +32,8 @@ def get_tf_dataset(self):
         from keras.src.utils.module_utils import tensorflow as tf
 
         def convert_to_tf(x, spec):
+            if isinstance(spec, tf.OptionalSpec):
+                return x
```
hertschuh: Can't you just return `tf.experimental.Optional.empty(None)` here are remove lines 55-62?

Or `tf.experimental.Optional.empty(None) is x is None else x`?

Either way, lines 55-62 should move here.

keras/src/tree/tree_api.py
```diff
@@ -179,6 +179,7 @@ def map_structure(func, *structures):
     Args:
         func: A callable that accepts as many arguments as there are structures.
         *structures: Arbitrarily nested structures of the same layout.
+        none_is_leaf: If True, None is treated as a leaf.
```
hertschuh: Can you add more details here? The name `none_is_leaf` is pretty unintuitive actually. Basically, say something like:

`none_is_leaf=True` causes `func` to be called on `None` leaves, and `none_is_leaf=False` means `None`s are not passed to `func` and are returned in the output directly.

keras/src/tree/dmtree_impl.py
```diff
@@ -194,16 +194,32 @@ def flatten_with_path(structure):
...
+    map_func = func
+    if not none_is_leaf:
+
+        def func_skipping_none(*args):
+            # Check if the reference entry (first one) is None
+            if args[0] is None:
+                if not all(s is None for s in args):
+                    raise ValueError(
+                        "Structure mismatch: some arguments are None, others "
+                        "are not."
```
hertschuh: Issues while running `map_structure` can be hard to debug. Any bit of context can help.

Can you add `args`?
```python
                    raise ValueError(
                        "Structure mismatch: some arguments are None, others "
                        f"are not: {args}."
                    )
```

keras/src/backend/tensorflow/trainer.py
```diff
@@ -101,12 +103,19 @@ def test_step(self, data):
...
         if self._call_has_training_arg:
             y_pred = self(x, training=False)
         else:
             y_pred = self(x)
         return y_pred
 
+    def _convert_optional_to_none(self, x):
+        # Convert TF Optional implementations to None
+        return tree.map_structure(
+            lambda i: None if isinstance(i, tf.experimental.Optional) else i, x
```
hertschuh: Don't you also need to do `i.get_value() if i.has_value() else None`? So that you support both the `None` and not `None` cases?

keras/src/trainers/data_adapters/data_adapter_utils.py
```diff
@@ -199,6 +203,8 @@ def convert_to_tf_tensor_spec(keras_tensor, batch_axis_to_none=True):
     """
     from keras.src.utils.module_utils import tensorflow as tf
 
+    if keras_tensor is None:
```
hertschuh, line 206: Does this actually ever happen?

My assumption was that this would need to handle non-None inputs that have `optional=True` on them (this might require some changes), and then create a `tf.OptionalSpec(<the actual tensorspec for the KerasTensor per the code below>)`.

hertschuh (commented):
Thank you for the PR!

> I am unsure where would be the best place for it... [model_test](https://github.com/keras-team/keras/blob/master/keras/src/models/model_test.py) maybe?).

Yes, this would be the place to test. Ideally, it would test `fit`, `predict` and `evaluate`.

Also, ideally, this would be tested in `*_data_adapter_test.py` to cover all cases.

Taking a step back, is the goal to handle the case when in the dataset passed to `fit` "input2" is always `None`? Or sometimes `None` sometimes not `None`. Right now it looks like it's only supporting the latter (always `None`).

keras/src/trainers/data_adapters/generator_data_adapter.py
neo-alex: Unfortunately no (this is what I tried first): indeed, an error is then raised by `tree.map_structure` on line 63 because `batch` and `self._output_signature` do not have the same structure (more specifically: None leaves in `batch` do not match None leaves in `self._output_signature`, which have `tf.OptionalSpec` instead). This is why I had to convert None leaves in `batch` first, in lines 55-62 - please let me know if you find a more elegant solution to this issue though (I am also not a fan of having 2 `map.structure` calls in a row if it is avoidable).

keras/src/tree/tree_api.py
neo-alex: Sure, I improved its docstring accordingly in [this commit](https://github.com/keras-team/keras/pull/21548/commits/687a8658adc396f8076202dbadec27ac798786c3). By the way, I agree that the name `none_is_leaf` is not the most intuitive, but it is used consistently throughout the underlying optree library (e.g. [here](https://optree.readthedocs.io/en/latest/ops.html#optree.tree_map)), so I kept the same one.

keras/src/tree/dmtree_impl.py
neo-alex: OK, done in [this commit](https://github.com/keras-team/keras/pull/21548/commits/abe2056370b240adf1c82c874bfe0c8f4da774df).

keras/src/backend/tensorflow/trainer.py
neo-alex: Yes you are probably right, I will double-check (see also my reply below to your "taking a step back" comment wrt. mixing `None` and not `None` cases).

keras/src/trainers/data_adapters/data_adapter_utils.py
neo-alex: It does actually happen, even if the reason is not intuitive: your assumption makes a lot of sense (ideally we would like optional inputs to be represented by `KerasTensor` with `optional=True` like in the model), unfortunately all the code in data_adapters is independent from the model, and the data spec is solely inferred from the first batches of received data (typically [here](https://github.com/keras-team/keras/blob/6114a21a83c52434fd45eb5167cb9fd3cd0e63f0/keras/src/trainers/data_adapters/generator_data_adapter.py#L59))... which seems indeed a bit brittle and prone to some "hidden" constraints for the first batches of the dataset (e.g. see [this error message](https://github.com/keras-team/keras/blob/6114a21a83c52434fd45eb5167cb9fd3cd0e63f0/keras/src/trainers/data_adapters/generator_data_adapter.py#L40)).

Since it is not possible to infer a proper `KerasTensor` just from a received `None` value, the trick I am using is to keep it as `None` (by using the newly introduced `none_is_leaf=False` inside `get_keras_tensor_spec`), which explains then that the line of code you mention is actually needed.

neo-alex:
Thank you very much @hertschuh for your insightful review! To answer your "taking a step back" comment, for now the goal is at least to enable optional inputs in model fit/evaluate/predict when they are always or never `None`. Of course, the best would be to enable both (sometimes `None`, sometimes a tensor) but this can be a bit tricky since tensor specs in data adapters are actually inferred from the first batches (at least for some backends like Tensorflow) - see my last comment above for more technical details.

I will continue to investigate and see if a solution for mixed values (`None` & not `None`) is still possible, given some constraints like "in the first 2 batches, every optional input should include a `None` value and a tensor one" (similar constraints are already assumed in current data adapters, as seen [here](https://github.com/keras-team/keras/blob/6114a21a83c52434fd45eb5167cb9fd3cd0e63f0/keras/src/trainers/data_adapters/generator_data_adapter.py#L40)). I will come back to you shortly about this.

keras/src/trainers/data_adapters/generator_data_adapter.py
neo-alex: **Erratum:** sorry, I got confused in my own tests (there is actually no issue using `tree.map_structure` with None leaves on one structure and tf.OptionalSpec on another, as long as `none_is_leaf=True` - which is the default). So you are right and I simplified the logic according to your comment in [this commit](https://github.com/keras-team/keras/pull/21548/commits/c5b636abe1028c9d33711d1520b475fb71522cae).

hertschuh:
> Thank you very much @hertschuh for your insightful review! To answer your "taking a step back" comment, for now the goal is at least to enable optional inputs in model fit/evaluate/predict when they are always or never `None`. Of course, the best would be to enable both (sometimes `None`, sometimes a tensor) but this can be a bit tricky since tensor specs in data adapters are actually inferred from the first batches (at least for some backends like Tensorflow) - see my last comment above for more technical details.
> 
> I will continue to investigate and see if a solution for mixed values (`None` & not `None`) is still possible, given some constraints like "in the first 2 batches, every optional input should include a `None` value and a tensor one" (similar constraints are already assumed in current data adapters, as seen [here](https://github.com/keras-team/keras/blob/6114a21a83c52434fd45eb5167cb9fd3cd0e63f0/keras/src/trainers/data_adapters/generator_data_adapter.py#L40)). I will come back to you shortly about this.

You're right, the data spec and and the model inputs are disconnected, which, as you point out, is the source of a number of shortcomings. It might not be possible to mix `None` and not `None` without this connection, which will be complex to add. Unless we do the same hack as for dynamic dimensions that you linked.

neo-alex:
Actually, the current solution in this PR already works for model fit/evaluate/predict across Jax/TF/Torch backends on single data batches as well as batches provided by a generator - even when the latter generates "mixed" optional input values (`None` & not `None`) **except in TF backend where an error is raised in this very specific case**. The core reason is that TF backend uses TF Dataset under the hood, which doesn't support None values, so we need to convert optional input values from None/tensor to tf.experimental.Optional in the data adapter, and back to None/tensor as expected by Keras model for optional inputs. But doing this last conversion (depending on whether the Optional has a value) in a way that is traceable by TF AutoGraph (when run_eagerly=False) is not trivial, since TF control flow operators like tf.cond also don't support returning `None`...

Since my latest comment, I experimented a lot and finally did find a solution that also covers this last edge case (generator with "mixed" optional input values with TF backend). But the solution is more complex than expected and I would rather add it in a dedicated follow-up PR to enable specific discussions about it.

So, if you agree @hertschuh, I would suggest merging this PR first, which already covers most cases. To that end, I just added [in this commit](https://github.com/keras-team/keras/pull/21548/commits/f77a4cc9100c2c6024e2e42ea9c5f65de04238c8) some unit tests (within model_test.py as discussed above).

keras/src/backend/tensorflow/trainer.py
neo-alex: So, after deeper investigation, it is not that simple to handle the "mixed" case in Tensorflow backend (when a generator sometimes provides `None` and sometimes a tensor for optional inputs). In particular, the problem with your suggestion is that `i.has_value()` can be a symbolic tensor (when code is traced by TF with the default `run_eagerly=False`), which cannot be evaluated as a Python boolean. A natural solution would be to replace your Python condition with something like `tf.cond(i.has_value(), lambda: i.get_value(), lambda: None)`, unfortunately TF control flow operators don't support `None` as return values... I believe I found a more complex solution that does the trick to cover this Tensorflow edge case, but I think it deserves a dedicated follow-up PR and would suggest first merging this current one - please see [this comment](https://github.com/keras-team/keras/pull/21548#issuecomment-3192909676) for more details. Thank you!

hertschuh: Ok, sounds good.

hertschuh:
@neo-alex 

It looks like it's failing on GPU. Do you understand why?

neo-alex:
> @neo-alex
> 
> It looks like it's failing on GPU. Do you understand why?

Yes, it seems to be due to the XLA JIT-compilation (automatically activated on GPU but not on CPU) - when I force `jit_compile=True` in `model.compile`, I am able to reproduce the issue even on CPU. I wasn't aware but it turns out that tf.experimental.Optional is unfortunately not JIT-compatible (for instance, the first example on [this official TF documentation page](https://www.tensorflow.org/api_docs/python/tf/OptionalSpec) fails as soon as `jit_compile=True` is added to `@tf.function`) 😞 

I think the trick would be to move the Optional->tensor/None conversion just outside of the [one_step_on_data](https://github.com/keras-team/keras/blob/7da416d3b0d80cc3cc899d8358c6357cded5a5ad/keras/src/backend/tensorflow/trainer.py#L112) function, which is the part that Keras possibly JIT compiles. I believe this should work since a couple of `tf.experimental.Optional` objects are already successfully present in the code base outside this function (like [here](https://github.com/keras-team/keras/blob/7da416d3b0d80cc3cc899d8358c6357cded5a5ad/keras/src/backend/tensorflow/trainer.py#L149) for instance).

I will try to push an update for this in the next couple of days and let you know.

hertschuh:
> I will try to push an update for this in the next couple of days and let you know.

Great, thanks!

neo-alex:
Alright, I just pushed the suggested fix for JIT compilation compatibility (I believe everything should be fine this time, even on GPU). I also made sure that the follow-up PR covering the last TF edge case is still compatible with this change - I will submit it as soon as you merge this one @hertschuh
Thanks!

PR #21552, opened 2025-08-06
Add assume_unique and invert arguments to keras.ops.isin

shashaka:
I added support for the assume_unique and invert arguments to the isin method in keras.ops.


hertschuh (approved):
Thank you for the PR!

PR #21554, opened 2025-08-07
Added cholesky inverse operation to all the backends

buildwithsuhana:
The Cholesky inverse is a fast and numerically stable algorithm for calculating the inverse of a special type of matrix: one that is symmetric and positive-definite.

Instead of calculating the inverse directly, this method first breaks the matrix down into simpler components using Cholesky decomposition.  This approach is generally more efficient and less prone to floating-point errors than a general-purpose inversion method.

Its numerical stability makes it a key component in advanced machine learning algorithms like GPTQ quantisation, where it is used to efficiently compute the inverse of Hessian matrices for layer-wise weight compression.

### DIFF
--- a/keras/src/backend/jax/linalg.py
+++ b/keras/src/backend/jax/linalg.py
@@ -29,14 +29,6 @@ def cholesky_inverse(a):
     identity = jnp.eye(a.shape[-1], dtype=a.dtype)
     a_inv = solve_triangular(a, identity, lower=True)
     out = jnp.matmul(jnp.transpose(a_inv), a_inv)
-    try:
-        if jnp.any(jnp.isnan(out)):
-            raise ValueError(
-                "Cholesky inverse failed. The input might not be a valid "
-                "positive definite matrix."
-            )
-    except jax.errors.TracerBoolConversionError:
-        pass
     return out
 
 def det(a):
--- a/keras/src/backend/jax/linalg.py
+++ b/keras/src/backend/jax/linalg.py
@@ -26,10 +26,9 @@ def cholesky(a):
     return out
 
 def cholesky_inverse(a):
-    L = cholesky(a)
-    identity = jnp.eye(a.shape[0], dtype=a.dtype)
-    L_inv = solve_triangular(L, identity, lower=True)
-    out = L_inv.T @ L_inv
+    identity = jnp.eye(a.shape[-1], dtype=a.dtype)
+    a_inv = solve_triangular(a, identity, lower=True)
+    out = jnp.matmul(jnp.transpose(a_inv), a_inv)
     try:
         if jnp.any(jnp.isnan(out)):
             raise ValueError(
--- a/keras/src/backend/jax/linalg.py
+++ b/keras/src/backend/jax/linalg.py
@@ -10,7 +10,7 @@ from keras.src.backend.jax.core import convert_to_tensor
 
 
 def cholesky(a, upper=False):
-    out = jnp.linalg.cholesky(a)
+    out = jnp.linalg.cholesky(a, lower=not upper)
     try:
         # In eager mode, raise for nan to
         # achieve behavior consistency with numpy
@@ -23,8 +23,6 @@ def cholesky(a, upper=False):
     except jax.errors.TracerBoolConversionError:
         # Cannot raise for nan in tracing mode
         pass
-    if upper:
-        return jnp.swapaxes(out, -2, -1)
     return out
 
 
--- a/keras/src/backend/jax/linalg.py
+++ b/keras/src/backend/jax/linalg.py
@@ -28,12 +28,11 @@ def cholesky(a, upper=False):
 
 def cholesky_inverse(a, upper=False):
     identity = jnp.eye(a.shape[-1], dtype=a.dtype)
+    inv_chol = solve_triangular(a, identity, lower=not upper)
     if upper:
-        u_inv = solve_triangular(a, identity, lower=False)
-        a_inv = jnp.matmul(u_inv, jnp.transpose(u_inv))
+        a_inv = jnp.matmul(inv_chol, jnp.transpose(inv_chol))
     else:
-        l_inv = solve_triangular(a, identity, lower=True)
-        a_inv = jnp.matmul(jnp.transpose(l_inv), l_inv)
+        a_inv = jnp.matmul(jnp.transpose(inv_chol), inv_chol)
     return a_inv
 
 
--- a/keras/src/backend/jax/linalg.py
+++ b/keras/src/backend/jax/linalg.py
@@ -25,6 +25,20 @@ def cholesky(a):
         pass
     return out
 
+def cholesky_inverse(a):
+    L = cholesky(a)
+    identity = jnp.eye(a.shape[0], dtype=a.dtype)
+    L_inv = solve_triangular(L, identity, lower=True)
+    out = L_inv.T @ L_inv
+    try:
+        if jnp.any(jnp.isnan(out)):
+            raise ValueError(
+                "Cholesky inverse failed. The input might not be a valid "
+                "positive definite matrix."
+            )
+    except jax.errors.TracerBoolConversionError:
+        pass
+    return out
 
 def det(a):
     return jnp.linalg.det(a)
--- a/keras/src/backend/jax/linalg.py
+++ b/keras/src/backend/jax/linalg.py
@@ -9,7 +9,7 @@ from keras.src.backend.jax.core import cast
 from keras.src.backend.jax.core import convert_to_tensor
 
 
-def cholesky(a):
+def cholesky(a, upper=False):
     out = jnp.linalg.cholesky(a)
     try:
         # In eager mode, raise for nan to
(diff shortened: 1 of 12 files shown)


keras/src/backend/jax/linalg.py
```diff
@@ -25,6 +25,20 @@ def cholesky(a):
         pass
     return out
 
+def cholesky_inverse(a):
+    L = cholesky(a)
+    identity = jnp.eye(a.shape[0], dtype=a.dtype)
+    L_inv = solve_triangular(L, identity, lower=True)
+    out = L_inv.T @ L_inv
```
abheesht17: Prefer using the backend-specific op instead of `+`, `*`, `@`, etc.

abheesht17 (commented):
Thanks for the PR, quick review! Let's run code formatting and linting. Here are the instructions: https://github.com/keras-team/keras/pull/21554#issue-3298981981.

keras/src/backend/openvino/linalg.py
```diff
@@ -3,6 +3,11 @@ def cholesky(a):
         "`cholesky` is not supported with openvino backend"
     )
 
+def cholesky_inverse(a):
+raise NotImplementedError(
+    "`Cholesky inverse` is not supported with the OpenVINO backend"
+)
```
abheesht17: This should be indented (inside the function)?

abheesht17: nit: full stop after backend

keras/src/backend/jax/linalg.py
```diff
@@ -25,6 +25,19 @@ def cholesky(a):
...
+    a_inv = solve_triangular(a, identity, lower=True)
+    out = jnp.matmul(jnp.transpose(a_inv), a_inv)
+    try:
+        if jnp.any(jnp.isnan(out)):
+            raise ValueError(
+                "Cholesky inverse failed. The input might not be a valid "
+                "positive definite matrix."
+            )
+    except jax.errors.TracerBoolConversionError:
+        pass
```
abheesht17: Why do we have this condition for JAX but not for other backends?

keras/src/ops/linalg.py
```diff
@@ -43,6 +43,48 @@ def _cholesky(x):
...
+    def call(self, x):
+        return _cholesky_inverse(x)
+
+    def compute_output_spec(self, x):
+        _assert_2d(x)
+        _assert_square(x)
+        return KerasTensor(x.shape, x.dtype)
+
+
+def cholesky_inverse(x):
```
abheesht17: `torch.cholesky_inverse` has an arg, `upper`. Do we want this too?

https://docs.pytorch.org/docs/stable/generated/torch.cholesky_inverse.html#torch-cholesky-inverse

abheesht17: We need to add API export decorators here similar to https://github.com/keras-team/keras/blob/master/keras/src/ops/linalg.py#L19.

keras/src/ops/linalg_test.py
```diff
@@ -23,6 +23,19 @@ def test_cholesky(self):
         with self.assertRaises(ValueError):
             linalg.cholesky(x)
 
+    def test_cholesky_inverse(self):
```
abheesht17, line 26: Let's add a test for numerical verification. I know we don't do this for other ops (ideally, we should!); let's do it here

abheesht17 (commented):
Thanks! Left a few comments

keras/src/backend/tensorflow/linalg.py
```diff
@@ -7,10 +7,24 @@
 from keras.src.backend.tensorflow.core import convert_to_tensor
 
 
-def cholesky(a):
+def cholesky(a, upper=True):
```
abheesht17: Let's keep the default value for `upper` the same everywhere. It is `True` here, and `False` everywhere else. Let's keep it consistent!

keras/src/backend/torch/linalg.py
```diff
@@ -7,10 +7,18 @@
 from keras.src.backend.torch.core import convert_to_tensor
 
 
-def cholesky(x):
+def cholesky(x, upper=False):
+    if upper:
```
hertschuh: No need for an `if`, simply change lines 11-13 by `return torch.linalg.cholesky(x, upper=upper)`

keras/src/backend/torch/linalg.py
```diff
@@ -7,10 +7,18 @@
...
+        return torch.linalg.cholesky(x, upper=True)
     return torch.linalg.cholesky(x)
 
 
+def cholesky_inverse(x, upper=False):
+    if upper:
```
hertschuh: No need for an if, simply change lines 17-19 by return `torch.linalg.cholesky_inverse(x, upper=upper)`

keras/src/ops/linalg.py
```diff
@@ -7,8 +7,12 @@
 
 
 class Cholesky(Operation):
+    def __init__(self, upper=False, name=None):
```
hertschuh: Add `, *` before `name`.

keras/src/ops/linalg.py
```diff
@@ -17,32 +21,76 @@ def compute_output_spec(self, x):
...
     _assert_square(x)
     try:
-        return backend.linalg.cholesky(x)
+        return backend.linalg.cholesky(x, upper=upper)
     except Exception as e:
         raise ValueError(f"Cholesky decomposition failed: {e}")
 
 
+class CholeskyInverse(Operation):
+    def __init__(self, upper=False, name=None):
```
hertschuh: Add `, *` before `name`.

keras/src/backend/numpy/linalg.py
```diff
@@ -6,10 +6,23 @@
 from keras.src.backend.numpy.core import convert_to_tensor
 
 
-def cholesky(a):
+def cholesky(a, upper=False):
+    if upper:
```
hertschuh: No need for an if, simply change lines 10-12 by `return np.linalg.cholesky(x, upper=upper)`.

keras/src/backend/jax/linalg.py
```diff
@@ -23,9 +23,22 @@ def cholesky(a):
     except jax.errors.TracerBoolConversionError:
         # Cannot raise for nan in tracing mode
         pass
+    if upper:
```
hertschuh: Remove this, instead, line 13, do `out = jnp.linalg.cholesky(a, lower=not upper)`

keras/src/backend/openvino/linalg.py
```diff
@@ -1,9 +1,15 @@
-def cholesky(a):
+def cholesky(a, upper=False):
```
hertschuh, line 1: I think you also need to add the test to the list of tests to remove in `keras/src/backend/openvino/excluded_concrete_tests.txt`.

keras/src/backend/tensorflow/linalg.py
```diff
@@ -7,10 +7,24 @@
...
     out = tf.linalg.cholesky(a)
     # tf.linalg.cholesky simply returns NaNs for non-positive definite matrices
-    return tf.debugging.check_numerics(out, "Cholesky")
+    out = tf.debugging.check_numerics(out, "Cholesky")
+    if upper:
+        return tf.linalg.matrix_transpose(out)
```
hertschuh: Is it really just a transposition?

hertschuh (changes requested):
Thanks for the PR!

abheesht17:
@buildwithsuhana - can you set up pre-commit: https://github.com/keras-team/keras/blob/master/CONTRIBUTING.md#generating-public-api-and-formatting-the-code? This will ensure that your changes are formatted, and the symbols you've added are exported

keras/src/backend/openvino/linalg.py
```diff
@@ -1,9 +1,15 @@
...
     raise NotImplementedError(
         "`cholesky` is not supported with openvino backend"
     )
 
 
+def cholesky_inverse(a, upper=False):
+    raise NotImplementedError(
+        "`cholesky_inverse` is not supported with openvino backend"
```
abheesht17: Let's add a full stop at the end.

abheesht17 (commented):
Left small comments, will do one final pass in a bit. Thanks!

keras/src/backend/jax/linalg.py
```diff
@@ -26,6 +26,17 @@ def cholesky(a):
     return out
 
 
+def cholesky_inverse(a, upper=False):
+    identity = jnp.eye(a.shape[-1], dtype=a.dtype)
+    if upper:
+        u_inv = solve_triangular(a, identity, lower=False)
```
abheesht17: Do you think we can do `solve_triangular(a, identity, lower=not upper)` and move it outside `if...else`, and just keep the matmul op inside the `if...else`?

buildwithsuhana: Done

keras/src/ops/linalg_test.py
```diff
@@ -331,12 +353,51 @@ def test_svd(self):
...
 class LinalgOpsCorrectnessTest(testing.TestCase):
     def test_cholesky(self):
-        x = np.random.rand(4, 3, 3).astype("float32")
+        x_non_psd = np.random.rand(4, 3, 3).astype("float32")
         with self.assertRaises(ValueError):
-            linalg.cholesky(x)
-        x_psd = x @ x.transpose((0, 2, 1)) + 1e-5 * np.eye(3)
-        out = linalg.cholesky(x_psd)
-        self.assertAllClose(out, np.linalg.cholesky(x_psd), atol=1e-4)
+            linalg.cholesky(x_non_psd)
```
abheesht17, line 358: Nit: newline after this, just to separate the two tests

abheesht17:
@hertschuh - could you please drop an approval here? I don't have write access.

keras/src/ops/linalg_test.py
buildwithsuhana: Done!

hertschuh (approved):

PR #21556, opened 2025-08-07
feat: Add a predict_proba method on SKLearnClassifier

### LINKED ISSUE #21547: Adding a `predict_proba` method on SKLearnClassifier
Currently the SKLearnClassifier only exposes a `predict` method. At the moment the raw probabilities are used in the predict method like:

```python
# keras/keras/src/wrappers/sklearn_wrapper.py
class SKLBase(BaseEstimator):
    ...
    def predict(self, X):
        """Predict using the model."""
        from sklearn.utils.validation import check_is_fitted

        check_is_fitted(self)
        X = _validate_data(self, X, reset=False)
        raw_output = self.model_.predict(X) # <----- here
        return self._reverse_process_target(raw_output) # <----- transformed back
```

If a user wants to get probabilities out, they need to create a custom class like:

```python
class MyKerasSKLearnClassifier(SKLearnClassifier):
    def predict_proba(self, X):
        if not hasattr(self, 'model_'):
            raise RuntimeError("You must fit the model before calling predict_proba.")
        return self.model_.predict(X) # returns the raw probas
```

Would you consider adding a predict_proba method to the SKLearnClassifier class:

```python
class SKLearnClassifier(ClassifierMixin, SKLBase):
    ...
    def predict_proba(self, X):
        """Predict class probabilities of the input samples X."""
        from sklearn.utils.validation import check_is_fitted

        check_is_fitted(self)
        X = _validate_data(self, X, reset=False)
        return self.model_.predict(X)
```

so that downstream implementations and users can rely on the default class imported from Keras, rather then creating a custom wrapping class. `predict_proba` is the common method in the scikit-learn library for getting the probabilities out of a classifier.

divakaivan:
Closes #21547 


keras/src/wrappers/sklearn_wrapper.py
```diff
@@ -278,6 +278,14 @@ def dynamic_model(X, y, loss, layers=[10]):
     ```
     """
 
+    def predict_proba(self, X):
```
fchollet: The problem here is that since the model is configurable, we have no way to know whether the model outputs probabilities or not. This method serves no additional purpose over just `predict()`.

divakaivan: @fchollet  Could you elaborate, please? I'm not sure I understand your comment. In the case when the user expects probas they will get probas. The only difference between this predict_proba and predict is that the target is not transformed back. 

If the user expects probabilities, then they will get them. Although predict_proba might not always return proper probabilities, its inclusion allows users to interoperate with sklearn workflows that expect it. Some examples are in the original issue request.

divakaivan: I had a chat with Adrin Jalali and @ glemaitre and he suggested we put `predict_proba` under an `available_if` decorator. Maybe something like

```python
from sklearn.utils._available_if import available_if

def _returns_probas(estimator):
    return estimator.model_.layers[-1].activation.__name__ in ("sigmoid", "softmax")

class SKLearnClassifier:
    @available_if(_returns_probas)
    def predict_proba(self, X):
        ...
        return self.model_.predict(X)
```

Also pinging @adrinjalali for his thoughts on the issue/PR.

adrinjalali: yes, you basically need something of the same nature as here:

https://github.com/scikit-learn/scikit-learn/blob/d5c3201291e73e6f3dd6847e3f80557370e8f24c/sklearn/model_selection/_search.py#L602-L603

keras/src/wrappers/sklearn_wrapper.py
```diff
@@ -278,6 +279,15 @@ def dynamic_model(X, y, loss, layers=[10]):
     ```
     """
 
+    @sklearn.utils._available_if.available_if(_estimator_has("predict_proba"))
```
adrinjalali:
```suggestion
    @sklearn.utils.metaestimators.available_if(_estimator_has("predict_proba"))
```

keras/src/wrappers/utils.py
```diff
@@ -32,6 +32,21 @@ def _check_model(model):
         )
 
 
+def _estimator_has(attr):
+    def check(self):
+        from sklearn.utils.validation import check_is_fitted
+
+        check_is_fitted(self)
```
adrinjalali: we shouldn't be doing this check here, the method itself does it. This `_estimator_has` should only check if the method can exist, and that's independent of whether the estimator is fitted or not.

keras/src/wrappers/utils.py
```diff
@@ -32,6 +32,21 @@ def _check_model(model):
...
+        return (
+            True
+            if self.model_.layers[-1].activation.__name__
+            in ("sigmoid", "softmax")
+            else False
+        )
```
adrinjalali: more sth like:

```suggestion
        return (
            self.model_.layers[-1].activation.__name__ in ("sigmoid", "softmax")
        )
```

keras/src/wrappers/sklearn_wrapper.py
```diff
@@ -278,6 +279,15 @@ def dynamic_model(X, y, loss, layers=[10]):
...
+    def predict_proba(self, X):
+        """Predict class probabilities of the input samples X."""
+        from sklearn.utils.validation import check_is_fitted
+
+        check_is_fitted(self)
+        X = _validate_data(self, X, reset=False)
+        return self.model_.predict(X)
```
adrinjalali, line 288: are these probabilities? Or is this more of a `decision_function` implementation? 🤔

keras/src/wrappers/utils.py
divakaivan: oof. thanks

keras/src/wrappers/utils.py
```diff
@@ -32,6 +32,18 @@ def _check_model(model):
         )
 
 
+def _estimator_has(attr):
```
divakaivan: Should this be `_check_proba` or `_estimator_has_proba`? No reason to have `attr` as it's quite specific to what we want to check ([something like here](https://github.com/scikit-learn/scikit-learn/blob/c60dae20604f8b9e585fc18a8fa0e0fb50712179/sklearn/svm/_base.py#L829))

divakaivan:
The tests fail because `AttributeError: module 'sklearn.utils' has no attribute 'metaestimators'`. But the tests run using [`scikit-learn==1.7.2`](https://github.com/keras-team/keras/actions/runs/18229231498/job/51908193409?pr=21556#step:7:49) 🤔 Edit: Need to `from sklearn.utils.metaestimators import available_if` ✅

keras/src/wrappers/sklearn_wrapper.py
divakaivan: `model_.predict` returns probas when the last layer's activation function is softmax/sigmoid. If the last layer has no activation - I believe it's logits. 

<details>
<summary>Reproducible example in google colab</summary>

```python
from tensorflow.keras.layers import Dense, Input
from tensorflow.keras.models import Model
from tensorflow.keras.losses import SparseCategoricalCrossentropy

from sklearn.model_selection import train_test_split
from sklearn.datasets import make_classification

import random
import numpy as np
import tensorflow as tf

random.seed(42)
np.random.seed(42)

X, y = make_classification(n_samples=1000, n_features=10, n_informative=4, n_classes=4, random_state=42)
X_train, X_test, y_train, y_test = train_test_split(X, y, random_state=42)

inp = Input(shape=(10,))
x = Dense(20, activation="relu")(inp)
x = Dense(20, activation="relu")(x)
x = Dense(20, activation="relu")(x)
logits_output = Dense(4, activation=None)(x)

model_logits = Model(inp, logits_output)
model_logits.compile(loss=SparseCategoricalCrossentropy(from_logits=True), optimizer="adam")

model_logits.fit(X_train, y_train, epochs=10, verbose=0)

softmax_output = tf.keras.layers.Activation('softmax')(logits_output)
model_softmax = Model(inp, softmax_output)

test_sample = X_test[:1]

print("LOGITS OUTPUT:")
pred_logits = model_logits.predict(test_sample, verbose=0)
print(pred_logits)

print("SOFTMAX MODEL OUTPUT:")
pred_softmax = model_softmax.predict(test_sample, verbose=0)
print(pred_softmax)

print("MANUAL SOFTMAX APPLIED TO LOGITS:")
pred_manual_softmax = tf.nn.softmax(pred_logits).numpy()
print(pred_manual_softmax)

print("DIFFERENCE")
print(np.abs(pred_softmax - pred_manual_softmax))
```

</details>

```
LOGITS OUTPUT:
[[ 0.60939574  4.029889   -1.224225    1.267421  ]]
SOFTMAX MODEL OUTPUT:
[[0.02969535 0.9082174  0.00474632 0.05734099]]
MANUAL SOFTMAX APPLIED TO LOGITS:
[[0.02969535 0.9082174  0.00474632 0.05734099]]
DIFFERENCE
[[0. 0. 0. 0.]]
```

adrinjalali (commented):
We need two things to be more confident here:

- Do we want to have a list of "allowed" activation functions? Or do we allow anything?
- Do the ones we allow, work well in the `predict` method here? We need to make sure we test them all here as well.

divakaivan:
> We need two things to be more confident here:
> 
> * Do we want to have a list of "allowed" activation functions? Or do we allow anything?

In case we allow everything do we raise a warning if the user uses `predict_proba` and if the last layer's activation func isn't in an *allowed list* (so I guess we'll have such a list in this case). If no warning - then an *allowed list* isn't needed I guess. Is a warning in this context an acceptable practice in keras (or even scikit-learn)?

Looking at `skorch`, seems they allow anything (correct me if I'm wrong). (Below is the code I used to test) A skorch classifier has `predict_proba` even if probas aren't returned. 

```python
# this net's last layer's activation func is ReLU which doesn't return probas
net.predict_proba(X_test[:1])
# array([[15.860021  , 11.805934  ,  0.        ,  0.97727066,  0.        ,
#          1.0556134 ,  1.4179235 ,  0.83066434,  1.2646358 ,  0.13642114]],
#       dtype=float32)
net.predict(X_test[:1])
# array([0])
```

<details>
<summary>full skorch code</summary>

```python
# in google colab !pip install -U skorch -q
import numpy as np
from torch import nn
from skorch import NeuralNetClassifier

from skore import EstimatorReport

from sklearn.model_selection import train_test_split
from sklearn.datasets import make_classification

import torch
np.random.seed(42)
torch.manual_seed(42)

X, y = make_classification(n_samples=1000, n_features=10, n_informative=4, n_classes=2)
X = X.astype(np.float32)

X_train, X_test, y_train, y_test = train_test_split(X, y, random_state=42)

class MyModule(nn.Module):
    def __init__(self, num_units=10, nonlin=nn.ReLU()):
        super().__init__()

        self.dense0 = nn.Linear(10, num_units)
        self.nonlin = nonlin
        self.dropout = nn.Dropout(0.5)
        self.dense1 = nn.Linear(num_units, num_units)

    def forward(self, X, **kwargs):
        X = self.nonlin(self.dense0(X))
        X = self.dropout(X)
        X = self.nonlin(self.dense1(X))
        return X

net = NeuralNetClassifier(
    MyModule,
    max_epochs=10,
    lr=0.1,
)

net.fit(X_train, y_train);
net.predict_proba(X_test[:1])
# array([[15.860021  , 11.805934  ,  0.        ,  0.97727066,  0.        ,
#          1.0556134 ,  1.4179235 ,  0.83066434,  1.2646358 ,  0.13642114]],
#       dtype=float32)
net.predict(X_test[:1])
# array([0])
```

</details>

> * Do the ones we allow, work well in the `predict` method here? We need to make sure we test them all here as well.

Do you mean to test that probas get properly transformed into the respective class? (i.e. `model.predict_proba` gives `array([[0.58067167, 0.4193283 ]])` which becomes `array([0])` when `model.predict` is called)

adrinjalali:
I'd be happy to allow everything, but then I'd rather expose a `decision_functions` instead of a `predict_proba`. And then we also test that the two are compatible, i.e. argmax of `decision_function` is what `predict` returns.

keras/src/wrappers/sklearn_test.py
```diff
@@ -107,6 +107,9 @@ def use_floatx(x):
         ),
         "check_supervised_y_2d": "This test assumes reproducibility in fit.",
         "check_fit_idempotent": "This test assumes reproducibility in fit.",
+        "check_classifiers_train": (
+            "decision_function can return both probabilities and logits"
+        ),
```
divakaivan, line 113: Not sure if this is a good solution for this. [3 tests are failing without it](https://github.com/keras-team/keras/actions/runs/18348447167/job/52261410479#step:15:12521) 

```
FAILED keras/src/wrappers/sklearn_test.py::test_sklearn_estimator_checks[SKLearnClassifier(fit_kwargs={'epochs':5},model=<functiondynamic_modelat0x7fae46b3a680>,model_kwargs={'layers':[20,20,20],'loss':'categorical_crossentropy'})-check_classifiers_train] - AssertionError
FAILED keras/src/wrappers/sklearn_test.py::test_sklearn_estimator_checks[SKLearnClassifier(fit_kwargs={'epochs':5},model=<functiondynamic_modelat0x7fae46b3a680>,model_kwargs={'layers':[20,20,20],'loss':'categorical_crossentropy'})-check_classifiers_train(readonly_memmap=True)] - AssertionError
FAILED keras/src/wrappers/sklearn_test.py::test_sklearn_estimator_checks[SKLearnClassifier(fit_kwargs={'epochs':5},model=<functiondynamic_modelat0x7fae46b3a680>,model_kwargs={'layers':[20,20,20],'loss':'categorical_crossentropy'})-check_classifiers_train(readonly_memmap=True,X_dtype=float32)] - AssertionError
```

From [here](https://github.com/scikit-learn/scikit-learn/blob/f621117e4950b8f78d96a984c8d69ecc1f3e4db7/sklearn/utils/estimator_checks.py#L2546)

```python
            if hasattr(classifier, "decision_function"):
                try:
                    # decision_function agrees with predict
                    decision = classifier.decision_function(X)
                    if n_classes == 2:
                        if tags.target_tags.single_output:
>                           assert decision.shape == (n_samples,)
E                           AssertionError
```

looking at ways to avoid adding this expected failed check ~

divakaivan: So if I make the `decision_function` to match the shape in case of binary classification

```python
    def decision_function(self, X):
        """Get raw model outputs."""
        from sklearn.utils.validation import check_is_fitted

        check_is_fitted(self)

        X = _validate_data(self, X, reset=False)
        scores = self.model_.predict(X)
        if len(self.classes_) == 2:
            return scores[:, 1] # caveat that since this returns a single number, now argmax will be 0 so not good 
        return scores
```

Then the 3 errors are like this:

```
FAILED keras/src/wrappers/sklearn_test.py::test_sklearn_estimator_checks[SKLearnClassifier(fit_kwargs={'epochs':5},model=<functiondynamic_modelat0x16a8b2a20>,model_kwargs={'layers':[20,20,20],'loss':'categorical_crossentropy'})-check_classifiers_train(readonly_memmap=True)] - AssertionError:
Arrays are not equal

Mismatched elements: 91 / 200 (45.5%)
Max absolute difference among violations: 1
Max relative difference among violations: inf
 ACTUAL: array([1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
       1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
       1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,...
 DESIRED: array([0, 0, 0, 1, 1, 1, 0, 1, 1, 0, 0, 1, 0, 1, 1, 1, 0, 0, 0, 1, 0, 0,
       0, 0, 0, 1, 1, 1, 1, 0, 0, 1, 1, 1, 1, 0, 0, 1, 0, 1, 1, 1, 1, 0,
       1, 1, 1, 0, 0, 1, 1, 0, 0, 0, 0, 0, 1, 0, 1, 1, 1, 0, 1, 1, 1, 0,...
```

This comes from scores[:, 1] being probabilities. And `sklearn/utils/estimator_checks.py` does this

```python
                    dec_pred = (decision.ravel() > 0).astype(int)
```

So all our outputs are 1s.

divakaivan: - Is it a good compromise to return the shape from `self.model_.predict(X)` and ignore those test cases?
- Also is it acceptable practice in the scikit-learn community if `decision_function` is used for both both logits and probas?

adrinjalali: Yeah that's not a good constraint on our side in scikit-learn I'd say. I'm happy to always return `(n_samples, n_classes)` from `decision_function` and skip those failing tests. WDYT @omarmanzoor @glemaitre

OmarManzoor: I think @adrinjalali's suggestion makes sense.

adrinjalali (commented):
LGTM now. Probably needs a changelog.

divakaivan:
> LGTM now. Probably needs a changelog.

I got this changelog from gemini, does that work?

hertschuh:
Hi, sorry, this was ignored for a long time.

I think it looks good to me, but can you rebase?

divakaivan:
> Hi, sorry, this was ignored for a long time.
> 
> I think it looks good to me, but can you rebase?

thanks. done

hertschuh (approved):
Thanks!

PR #21559, opened 2025-08-08
Refactor(JAX): Use jax.jit's out_shardings instead of _enforce_jax_state_sharding

buildwithsuhana:
This is a refactor of the JAX trainer's sharding logic to simplify the code.

The logic previously handled by _enforce_jax_state_sharding has been moved into make_train_function, make_test_function, and make_predict_function. This is the ideal location, as it's where we call jax.jit. The now-redundant _enforce_jax_state_sharding method has been removed.

### DIFF
--- a/keras/src/backend/jax/trainer.py
+++ b/keras/src/backend/jax/trainer.py
@@ -260,7 +260,9 @@ class JAXTrainer(base_trainer.Trainer):
             return
         if not self.run_eagerly and self.jit_compile:
             out_shardings = None
-            if distribution_lib.distribution() is not None:
+            if distribution_lib.distribution() is not None and hasattr(
+                self, "_metrics_result_structure"
+            ):
                 out_shardings = (
                     tree.map_structure(lambda _: None, self._metrics_result_structure),
                     (
@@ -271,9 +273,7 @@ class JAXTrainer(base_trainer.Trainer):
                     ),
                 )
             train_step = jit(
-                self.train_step, 
-                donate_argnums=0,
-                out_shardings=out_shardings
+                self.train_step, donate_argnums=0, out_shardings=out_shardings
             )
         else:
             train_step = self.train_step
@@ -287,7 +287,9 @@ class JAXTrainer(base_trainer.Trainer):
             return
         if not self.run_eagerly and self.jit_compile:
             out_shardings = None
-            if distribution_lib.distribution() is not None:
+            if distribution_lib.distribution() is not None and hasattr(
+                self, "_metrics_result_structure"
+            ):
                 out_shardings = (
                     tree.map_structure(lambda _: None, self._metrics_result_structure),
                     (
@@ -297,15 +299,12 @@ class JAXTrainer(base_trainer.Trainer):
                     ),
                 )
             test_step = jit(
-                self.test_step, 
-                donate_argnums=0,
-                out_shardings=out_shardings
+                self.test_step, donate_argnums=0, out_shardings=out_shardings
             )
         else:
             test_step = self.test_step
 
         step_function = self._make_function(test_step)
-
         self.test_function = step_function
 
     def make_predict_function(self, force=False):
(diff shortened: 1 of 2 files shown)


keras/src/backend/jax/trainer.py
```diff
@@ -277,38 +255,57 @@ def iterator_step(state, iterator):
 
         return iterator_step
 
+    def _get_out_shardings_for_step(self, state_shardings):
+        """Helper to create the out_shardings for a jitted step function."""
+        if distribution_lib.distribution() is not None and hasattr(
+            self, "_metrics_result_structure"
```
fchollet: Best I can tell this is never set, so this function always returns `None`

keras/src/backend/jax/trainer.py
```diff
@@ -297,12 +285,19 @@ def make_test_function(self, force=False):
...
         if not self.run_eagerly and self.jit_compile:
-            # Note that we mark the state to be donated to jax,
-            # so that jax will reuse the memory buffer for outputs.
-            # This will reduce the memory usage of the training function by
-            # half.
-            test_step = jit(self.test_step, donate_argnums=0)
-
+            out_shardings = None
+            if distribution_lib.distribution() is not None:
+                state_shardings = (
```
hertschuh, line 290: Instead of calling `_record_training_state_sharding_spec` and then retrieving them via `self. ...` and then calling `_clear_jax_state_sharding`, simply:
- change `_record_training_state_sharding_spec` to return the shardings instead of saving them
- rename to `_get_training_state_shardings`
- call it right here instead of above
- remove `_clear_jax_state_sharding`

Same for `predict` and `evaluate`.

keras/src/backend/jax/trainer.py
```diff
@@ -314,12 +309,25 @@ def make_predict_function(self, force=False):
         if self.predict_function is not None and not force:
             return self.predict_function
 
-        def predict_step(state, data):
+        def predict_step_wrapper(state, data):
```
hertschuh: Why did you have to rename this one and not the others?

hertschuh (commented):
Thanks for the PR!

keras/src/backend/jax/trainer.py
```diff
@@ -884,75 +886,44 @@ def jax_state_sync(self):
...
-        step, we need to ensure that they have the same sharding spec, so that
-        nnx.jit/jax.jit won't have to recompile the train/eval function.
-
-        Note that this function will also rely on the recorded sharding spec
-        for each of states.
+    def _get_training_state_shardings(self):
+        distribution = distribution_lib.distribution()
+        mesh = distribution.device_mesh.backend_mesh
+
+        def get_sharding(variable):
```
hertschuh: I don't understand why you need `get_sharding`.

Previously, it was simply doing `v.value.sharding`, does that no longer work?

Additionally, this is how `KerasVariable` does it: https://github.com/keras-team/keras/blob/master/keras/src/backend/jax/core.py#L38-L45
One important part is to *not* use `distribution` if `variable._layout` is not `None`, we have an important use case in KerasRS where the distribution of some variables is done separately.

keras/src/backend/jax/trainer.py
```diff
@@ -281,11 +259,15 @@ def make_train_function(self, force=False):
...
             return
         if not self.run_eagerly and self.jit_compile:
-            # Note that we mark the state to be donated to jax,
-            # so that jax will reuse the memory buffer for outputs.
-            # This will reduce the memory usage of the training function by
-            # half.
-            train_step = jit(self.train_step, donate_argnums=0)
+            out_shardings = None
+            if distribution_lib.distribution() is not None:
+                state_shardings = self._record_training_state_sharding_spec()
```
hertschuh: `self._record_training_state_sharding_spec()` does not return anything (and therefore returns None). So you're setting `out_shardings` to `(None, None)`.

You do want to:
- rename `_record_training_state_sharding_spec` to something like `_get_state_sharding_spec`
- have it not set `self.` attributes
- have it return the shardings instead

The worrying part is that no unit test failed. Can we have a unit test that fails when the shardings are not set correctly?

buildwithsuhana: Done.

hertschuh (approved):
Thanks!

PR #21563, opened 2025-08-08
Fix the Doc of `alpha` in `mobilenet_v3`

### LINKED ISSUE #21562: Fix the Doc of `alpha` in `mobilenet_v3`
## Doc Issue

The doc of `mobilenet_v3` shows its description as below:

https://github.com/keras-team/keras/blob/4e1b250491627f871f6c82c0dcb577cc21093def/keras/src/applications/mobilenet_v3.py#L91-L99

Which version I tried in is of `TensorFlow 2.19.0`:

For the repro below, I accept that it should raise error, but should be noted in docs that arg `alpha` only has choices in `[0.75, 1.0]` when `weights` is of `default` value `"imagenet"`.

### Repro
```python
import tensorflow as tf
import numpy as np

file_name = 'file.npy'
np.save(file_name, (np.random.rand(160, 160, 3) * 255.0))
model = tf.keras.applications.MobileNetV3Small(input_shape=(160, 160, 3), alpha=0.5, include_top=False)
model.compile(optimizer='adam', loss='categorical_crossentropy', metrics=['accuracy'])
print(model.summary())
```
### Output
```text
ValueError: If imagenet weights are being loaded, alpha can be one of `0.75`, `1.0` for non minimalistic or `1.0` for minimalistic only.
```

Raise Codes are here:

https://github.com/keras-team/keras/blob/4e1b250491627f871f6c82c0dcb577cc21093def/keras/src/applications/mobilenet_v3.py#L268-L279

Thanks a lot!

ILCSFNO:
Fixes #21562

### DIFF
--- a/keras/src/applications/mobilenet_v3.py
+++ b/keras/src/applications/mobilenet_v3.py
@@ -90,7 +90,9 @@ Args:
         E.g. `(160, 160, 3)` would be one valid value.
     alpha: controls the width of the network. This is known as the
         depth multiplier in the MobileNetV3 paper, but the name is kept for
-        consistency with MobileNetV1 in Keras.
+        consistency with MobileNetV1 in Keras. When `weights` is of default
+        value `"imagenet"`, `alpha` can be one of `0.75`, `1.0` for non
+        minimalistic or `1.0` for minimalistic only.
         - If `alpha < 1.0`, proportionally decreases the number
             of filters in each layer.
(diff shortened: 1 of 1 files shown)


ILCSFNO:
Only fix the doc, not known why the performance is changed, could anyone point it out? Thanks a lot!

fchollet (approved):
LGTM, thanks

ILCSFNO:
Thanks a lot!

PR #21566, opened 2025-08-08
Add daily Python 3.13 CPU-only tests to nightly workflow

### LINKED ISSUE #21508: Add daily CPU tests which also test with the latest Python version
Today, all the tests are run with the lowest supported version of Python (currently 3.10).

We want to also test with the latest version of Python (3.13). However, probably not for every PR.

We should therefore add daily tests, which can be CPU only, that test with 3.13. They can be part of the nightly workflow: https://github.com/keras-team/keras/blob/master/.github/workflows/nightly.yml

This is how tests are run nightly in keras-hub:
https://github.com/keras-team/keras-hub/blob/ecef6d1cd43e34729d6c759d34a26d39560a0a70/.github/workflows/nightly.yml#L12-L13

rohithpudari:
fixes #21508 

Adds `build-python-latest` job in Nightly pipeline

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -3,7 +3,7 @@ name: Nightly
 on:
   workflow_dispatch: # To Generate wheels on demand outside of schedule.
   schedule:
-    - cron: '0 3 * * *' # run at 3 AM UTC / 8 PM PDT
+    - cron: "0 3 * * *" # run at 3 AM UTC / 8 PM PDT
 
 permissions:
   contents: read
@@ -13,9 +13,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.10']
+        python-version: ["3.10"]
         backend: [tensorflow, jax, torch, numpy]
-    name: Run tests
+    name: Run tests (Python 3.10)
     runs-on: ubuntu-latest
     env:
       PYTHON: ${{ matrix.python-version }}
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly.yml
```diff
@@ -13,9 +13,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.10']
+        python-version: ["3.10"]
         backend: [tensorflow, jax, torch, numpy]
-    name: Run tests
+    name: Run tests (Python 3.10)
```
hertschuh: Does it work if you put `${{ matrix.python-version }}` here?

The goal is to minimize the number of things to change when we upgrade Python.

.github/workflows/nightly.yml
```diff
@@ -57,6 +57,54 @@ jobs:
         run: |
           pytest keras --ignore keras/src/applications --cov=keras --cov-config=pyproject.toml
 
+  build-python-latest:
+    strategy:
+      fail-fast: false
+      matrix:
+      python-version: ["3.13"]
+      backend: [tensorflow, jax, torch, numpy]
+    name: Run tests (Python 3.13)
```
hertschuh: Does it work if you put `${{ matrix.python-version }}` here?

The goal is to minimize the number of things to change when we upgrade Python.

.github/workflows/nightly.yml
```diff
@@ -57,6 +57,54 @@ jobs:
...
+    runs-on: ubuntu-latest
+    env:
+      PYTHON: ${{ matrix.python-version }}
+      KERAS_BACKEND: ${{ matrix.backend }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python 3.13
```
hertschuh: Remove "3.13" here.

The goal is to minimize the number of things to change when we upgrade Python.

rohithpudari:
@hertschuh Thanks for the review, I updated the code to use `${{ matrix.python-version }}` in workflow names

hertschuh:
Thanks for adding this, @rohithpudari !

Issue #21519, opened 2025-07-26
Reshape layer does not handle -1 shape infor dynamically


roebel:
Hello

I am encountering a problem with keras.layers.Reshape . I am currently running 3.9, but given that the related code did not change in 3.10,  the problem will be the same.

in contrast to my expectation. and to the behavior I was accustomed to from keras-2, the keras 3 version of the reshape operation statically fixes all dimensions except the batch dimension. The current version of the code is here https://github.com/keras-team/keras/blob/3bedb9a970394879360fcb1c0264f3ffdc634a77/keras/src/layers/reshaping/reshape.py#L56:

```
def build(self, input_shape):
        sample_output_shape = operation_utils.compute_reshape_output_shape(
            input_shape[1:], self.target_shape, "target_shape"
        )
        self._resolved_target_shape = tuple(
            -1 if d is None else d for d in sample_output_shape
        )

def call(self, inputs):
        return ops.reshape(
            inputs, (ops.shape(inputs)[0],) + self._resolved_target_shape
        )
```

Clearly, this will reduce the possible uses of the reshape layer compared to the ops.reshape operator, which can handle 
a dimension with -1 without problem. It appears straightforward to extend the use of the Reshape layer to preserve the flexibility of the ops.reshape operator. Is there any reason why this is kept static? If not I may provide a suggestion for a fix.

Thanks for your comments.

hertschuh:
@roebel 

Thank you for the report.

Can you provide an example of an issue that you're encountering? I would have expected the PR to have a unit test that fails before and passes after.

> Clearly, this will reduce the possible uses of the reshape layer compared to the ops.reshape operator, which can handle
a dimension with -1 without problem

The `Reshape` layer can handle of `-1` both with static and dynamic dimensions (and dynamic batch size). Instead redoing the work of calculating the dynamically resolved shape, it just lets `ops.reshape` handle it. No need to duplicated that work.

It does so by having `self._resolved_target_shape` contain a `-1` in it if it could not be statically resolved.

roebel:
It appears I misinterpreted the current implementation, notably the name `self._resolved_target_shape`, which, given the name, I understood would resolve the -1 in all cases. I now see that it only resolves the shape when the input shape is not dynamic. This indeed is exactly my use case. I share here below a small test script that shows the problem.

The script runs file when using it with keras 2.14. 

```
KERAS_INSTALL=system ./shell/manual_reshape_test.py
```

fails with all backends under keras 3.9

```
[...]

PR #21568, opened 2025-08-09
Proposed fix for issue #21519: Reshape layer does not handle -1 shape…

roebel:
I have implemented a fix that avoids precalculating a statically resolved reshape target in the build method and instead properly calculates the proper shape information for the problematic case when the shape information is not complete in the call method. The method appears to pass the reshape tests for all backends.

### DIFF
--- a/keras/src/layers/reshaping/reshape.py
+++ b/keras/src/layers/reshaping/reshape.py
@@ -11,9 +11,9 @@ class Reshape(Layer):
 
     Args:
         target_shape: Target shape. Tuple of integers, does not include the
-            samples dimension (batch size). One element of the `target_shape` can
-            be -1 in which case the missing value is inferred from the size
-            of the array and remaining dimensions.
+            samples dimension (batch size). One element of the `target_shape`
+            can be -1 in which case the missing value is inferred from the
+            size of the array and remaining dimensions.
 
     Input shape:
         Arbitrary, but required to be compatible with `target_shape`.
--- a/keras/src/layers/reshaping/reshape.py
+++ b/keras/src/layers/reshaping/reshape.py
@@ -3,7 +3,7 @@ from keras.src.api_export import keras_export
 from keras.src.backend.common.keras_tensor import KerasTensor
 from keras.src.layers.layer import Layer
 from keras.src.ops import operation_utils
-
+import math
 
 @keras_export("keras.layers.Reshape")
 class Reshape(Layer):
@@ -37,7 +37,19 @@ class Reshape(Layer):
 
     def __init__(self, target_shape, **kwargs):
         super().__init__(**kwargs)
-        self.target_shape = tuple(target_shape)
+        target_shape = tuple(target_shape)
+        # test validity of target_shape
+        if target_shape.count(-1) > 1:
+            raise ValueError(
+                "The `target_shape` argument must not contain more than one "
+                "`-1` value. Received: target_shape={}".format(target_shape)
+            )
+        self.target_shape = target_shape
+        # precalculate all values that might be required
+        self.need_explicit_shape_for_batch_size_None = (target_shape.count(-1) == 1)
+        self.new_size_no_minus_one = math.prod(
+            d for d in target_shape if d != -1
+        )
 
     def compute_output_shape(self, input_shape):
         return (
@@ -53,18 +65,22 @@ class Reshape(Layer):
             shape=output_shape, dtype=inputs.dtype, sparse=inputs.sparse
         )
 
-    def build(self, input_shape):
-        sample_output_shape = operation_utils.compute_reshape_output_shape(
-            input_shape[1:], self.target_shape, "target_shape"
-        )
-        self._resolved_target_shape = tuple(
-            -1 if d is None else d for d in sample_output_shape
+    def call(self, inputs):
+        target_shape = self.target_shape
+        if self.need_explicit_shape_for_batch_size_None and (inputs.shape[0] is None):
+            input_nonbatch_shape = tuple(inputs.shape[1:])
+            if input_nonbatch_shape.count(None) == 0:
+                # If the input shape is fully defined, we can compute the desired target_shape
+                if True:
+                    inp_nonbatch_size = math.prod(inputs.shape[1:])
+                else:
+                    inp_nonbatch_size = ops.prod(ops.shape(inputs)[1:])
+                target_shape = tuple(d if d != -1 else (inp_nonbatch_size // self.new_size_no_minus_one) for d in self.target_shape)
+
+        return ops.reshape(
+            inputs, (ops.shape(inputs)[0],) + target_shape
         )
 
-    def call(self, inputs):
-        return ops.reshape(
-            inputs, (ops.shape(inputs)[0],) + self._resolved_target_shape
-        )
 
     def get_config(self):
         config = {"target_shape": self.target_shape}
--- a/keras/src/layers/reshaping/reshape.py
+++ b/keras/src/layers/reshaping/reshape.py
@@ -1,9 +1,11 @@
+import math
+
 from keras.src import ops
 from keras.src.api_export import keras_export
 from keras.src.backend.common.keras_tensor import KerasTensor
 from keras.src.layers.layer import Layer
 from keras.src.ops import operation_utils
-import math
+
 
 @keras_export("keras.layers.Reshape")
 class Reshape(Layer):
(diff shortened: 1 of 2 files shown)


keras/src/layers/reshaping/reshape.py
```diff
@@ -53,19 +65,23 @@ def compute_output_spec(self, inputs):
...
     def call(self, inputs):
+        target_shape = self.target_shape
+        if self.need_explicit_shape_for_batch_size_None and (inputs.shape[0] is None):
+            input_nonbatch_shape = tuple(inputs.shape[1:])
+            if input_nonbatch_shape.count(None) == 0:
+                # If the input shape is fully defined, we can compute the desired target_shape
+                if True:
+                    inp_nonbatch_size = math.prod(inputs.shape[1:])
+                else:
+                    inp_nonbatch_size = ops.prod(ops.shape(inputs)[1:])
```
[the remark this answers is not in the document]
roebel: Yes indeed, I wondered in fact whether a change like this:

```
    def call(self, inputs):
        target_shape = self.target_shape
        if self.need_explicit_shape_for_batch_size_None and (inputs.shape[0] is None):
                inp_nonbatch_size = ops.prod(ops.shape(inputs)[1:])
                target_shape = tuple(d if d != -1 else (inp_nonbatch_size // self.new_size_no_minus_one) for d in self.target_shape)

        return ops.reshape(
            inputs, (ops.shape(inputs)[0],) + target_shape
        )
```

would have any advantage. Would it?

roebel:
You find the test case in the new commit. For the context: When using convolutional models with time series data (voice processing), the time dimension is hardly ever known beforehand. You can probably force a dynamic dimension here, but for convolutional layers this is not required, and for the reshape layer it is not needed either.

> Instead redoing the work of calculating the dynamically resolved shape, it just lets ops.reshape handle it. No need to duplicated that work

I may not fully understand this, but I thought that the reshape layer has an advantage compared to the reshape operator, because it knows that the batch dimension stays the same. I wondered in fact why the test case: test_reshape_with_dynamic_batch_size_and_minus_one produces shape (None, 3, 8) and not (None, None, 8). As for example this does:

```
ops.reshape(keras.KerasTensor((None, 6, 4)), (None, -1, 8))
Out[8]: <KerasTensor shape=(None, None, 8), dtype=float32, sparse=False, ragged=False, name=keras_tensor_3>
```

roebel:
I don't see any reason why the torch test fails. It runs fine here in my Keras development environment, and it also runs fine with the mini script I shared in the issue.

keras/src/layers/reshaping/reshape.py
```diff
@@ -53,18 +69,22 @@ def compute_output_spec(self, inputs):
...
-        )
-        self._resolved_target_shape = tuple(
-            -1 if d is None else d for d in sample_output_shape
-        )
-
     def call(self, inputs):
-        return ops.reshape(
-            inputs, (ops.shape(inputs)[0],) + self._resolved_target_shape
-        )
+        target_shape = self.target_shape
```
hertschuh: I think it's easier to just transfer the code from `build` here:

```python
    def call(self, inputs):
        potentially_resolved_target_shape = (
            operation_utils.compute_reshape_output_shape(
                tuple(inputs.shape)[1:], self.target_shape, "target_shape"
            )
        )
        potentially_resolved_target_shape = tuple(
            -1 if d is None else d for d in potentially_resolved_target_shape
        )
        return ops.reshape(
            inputs, (ops.shape(inputs)[0],) + potentially_resolved_target_shape
        )
```

- you don't have to reimplement the computation of the missing dimension
- you don't have to deal with errors if the number of values is not divisible by `self.new_size_no_minus_one` (right now, that check is missing)
- you don't need `self.need_explicit_shape_for_batch_size_None` and `self.new_size_no_minus_one`

keras/src/layers/reshaping/reshape.py
```diff
@@ -37,7 +39,21 @@ class Reshape(Layer):
...
+                "`-1` value. Received: target_shape={}".format(target_shape)
+            )
+        self.target_shape = target_shape
+        # precalculate all values that might be required
+        self.need_explicit_shape_for_batch_size_None = (
+            target_shape.count(-1) == 1
+        )
+        self.new_size_no_minus_one = math.prod(
+            d for d in target_shape if d != -1
+        )
```
hertschuh: Because you removed the `build` method, add `self.built = True` at the end of `__init__`.

keras/src/layers/reshaping/reshape_test.py
```diff
@@ -100,6 +102,32 @@ def test_reshape_with_dynamic_batch_size_and_minus_one(self):
         reshaped = backend.compute_output_spec(layer.__call__, input)
         self.assertEqual(reshaped.shape, (None, 3, 8))
 
+    def test_reshape_layer_with_varying_input_size_and_minus_one(self):
+        input = KerasTensor((None, 6, 4))
+        layer = layers.Reshape((-1, 8))
+        layer.build(input.shape)
```
hertschuh: You could remove all the calls to `build` in this test and others, it's no longer needed (just to make the code shorter).

keras/src/layers/reshaping/reshape_test.py
```diff
@@ -100,6 +102,32 @@ def test_reshape_with_dynamic_batch_size_and_minus_one(self):
         reshaped = backend.compute_output_spec(layer.__call__, input)
         self.assertEqual(reshaped.shape, (None, 3, 8))
 
+    def test_reshape_layer_with_varying_input_size_and_minus_one(self):
+        input = KerasTensor((None, 6, 4))
```
hertschuh: `input = KerasTensor((None, 6, 4))` is no longer needed in this test, remove.

keras/src/layers/reshaping/reshape_test.py
```diff
@@ -100,6 +102,32 @@ def test_reshape_with_dynamic_batch_size_and_minus_one(self):
...
+        layer = layers.Reshape((-1, 8))
+        layer.build(input.shape)
+        res = layer(ops.ones((1, 6, 4), dtype="float32"))
+        self.assertEqual(res.shape, (1, 3, 8))
+        res = layer(ops.ones((1, 10, 4), dtype="float32"))
+        self.assertEqual(res.shape, (1, 5, 8))
+
+    def test_custom_reshape_model_with_varying_input_size_and_minus_one(self):
```
hertschuh: Let's do a compiled example instead, which exercise a different use case than the example right before. Also, the model can be just the `Reshape` layer:

```python
    def test_reshape_model_fit_with_varying_input_size_and_minus_one(self):
        def generator():
            yield (
                ops.ones((1, 12, 2), dtype="float32"),
                ops.zeros((1, 3, 8), dtype="float32"),
            )
            yield (
                ops.ones((1, 20, 2), dtype="float32"),
                ops.zeros((1, 5, 8), dtype="float32"),
            )

        layer = layers.Reshape((-1, 8))
        model = models.Sequential([layer])
        model.compile(loss="mean_squared_error")
        model.fit(generator())
```

keras/src/layers/reshaping/reshape.py
```diff
@@ -37,7 +39,21 @@ class Reshape(Layer):
 
     def __init__(self, target_shape, **kwargs):
         super().__init__(**kwargs)
-        self.target_shape = tuple(target_shape)
+        target_shape = tuple(target_shape)
+        # test validity of target_shape
+        if target_shape.count(-1) > 1:
+            raise ValueError(
+                "The `target_shape` argument must not contain more than one "
+                "`-1` value. Received: target_shape={}".format(target_shape)
```
hertschuh: Please use an f-string for this.

roebel:
Thanks for all the comments and suggestions.

keras/src/layers/reshaping/reshape.py
```diff
@@ -11,11 +11,13 @@ class Reshape(Layer):
 
     Args:
         target_shape: Target shape. Tuple of integers, does not include the
-            samples dimension (batch size).
+            samples dimension (batch size). One element of the target_shape can
```
hertschuh: Nitpick: `` `target_shape` `` with backquotes.

keras/src/layers/reshaping/reshape.py
```diff
@@ -11,11 +11,13 @@ class Reshape(Layer):
...
+            be -1 in which case the missing value is inferred from the length
+            of the array and remaining dimensions.
 
     Input shape:
-        Arbitrary, although all dimensions in the input shape must be
-        known/fixed. Use the keyword argument `input_shape` (tuple of integers,
+        Arbitrary, but required to be compatible with target shape.
+        Use the keyword argument `input_shape` (tuple of integers,
```
hertschuh: >  Use the keyword argument `input_shape` ...

Now I'm noticing this documentation. This is actually a deprecated way of doing this. Plus, it's not needed.

Just remove that last part.

hertschuh (approved):
Thanks a lot for the PR!

PR #21572, opened 2025-08-11
Add Distillation API to Keras

divyashreepathihalli:
This PR adds *Knowledge Distillation API* to Keras,

### Core Components
- *Distiller*: Main distillation model that combines teacher and student models
- *Strategies*: Pluggable distillation strategies (LogitsDistillation, FeatureDistillation, MultiOutputDistillation)

### Basic Knowledge Distillation
```
import keras
from keras.distillation import Distiller, LogitsDistillation

# Create models
teacher = keras.Sequential([...])  # Large, pre-trained model
student = keras.Sequential([...])  # Smaller model to train

# Set up distillation
distiller = Distiller(
    teacher=teacher,
    student=student,
    strategies=[LogitsDistillation(temperature=3.0)],
    alpha=...
)

# Standard Keras workflow
distiller.compile(optimizer='adam', loss='sparse_categorical_crossentropy')
distiller.fit(x_train, y_train, epochs=10)
predictions = distiller.predict(x_test)
```

### DIFF
--- a/keras/src/distillation/distillation_loss.py
+++ b/keras/src/distillation/distillation_loss.py
@@ -10,10 +10,10 @@ def _convert_loss_to_function(loss_item):
 
     Arguments:
         loss_item: Either a string identifier, a loss function instance,
-            or None.
+            or `None`.
 
     Returns:
-        A loss function instance, or None.
+        A loss function instance, or `None`.
 
     Raises:
         ValueError: If the loss string identifier is unknown.
@@ -53,7 +53,7 @@ class DistillationLoss:
                 tensor or a list/tuple of tensors for multi-output models.
             student_outputs: Outputs from the student model. Can be a single
                 tensor or a list/tuple of tensors for multi-output models.
-            **kwargs: Additional arguments for custom strategies.
+            **kwargs: Additional arguments for custom distillation_loss.
         Returns:
             Distillation loss tensor.
         """
@@ -77,14 +77,15 @@ class DistillationLoss:
             teacher: The teacher model.
             student: The student model.
         Raises:
-            ValueError: If models are not compatible with this strategy.
+            ValueError: If models are not compatible with this
+                distillation_loss.
         """
         pass
 
 
 @keras_export("keras.distillation.FeatureDistillation")
 class FeatureDistillation(DistillationLoss):
-    """Feature distillation strategy using intermediate layer representations.
+    """Feature distillation distillation_loss.
 
     Feature distillation transfers knowledge from intermediate layers of the
     teacher model to corresponding layers of the student model. This approach
@@ -96,46 +97,46 @@ class FeatureDistillation(DistillationLoss):
             - String identifier (e.g., 'mse', 'cosine_similarity', 'mae')
             - Keras loss instance
             - Nested structure of losses matching the layer output structure
-            - None to skip distillation for that output (useful for multi-output
-              models where you only want to distill some outputs)
+            - `None` to skip distillation for that output (useful for
+              multi-output models where you only want to distill some outputs)
             At least one loss must be non-None. Defaults to 'mse'.
         teacher_layer_name: Name of the teacher layer to extract features from.
-            If None, uses the final output. Defaults to None.
+            If `None`, uses the final output. Defaults to `None`.
         student_layer_name: Name of the student layer to extract features from.
-            If None, uses the final output. Defaults to None.
+            If `None`, uses the final output. Defaults to `None`.
 
     Examlpe(s):
 
     ```python
     # Basic feature distillation from final outputs
-    strategy = FeatureDistillation(loss="mse")
+    distillation_loss = FeatureDistillation(loss="mse")
 
     # Distill from specific intermediate layers
-    strategy = FeatureDistillation(
+    distillation_loss = FeatureDistillation(
         loss="mse",
         teacher_layer_name="dense_1",
         student_layer_name="dense_1"
     )
 
     # Use cosine similarity for different feature sizes
-    strategy = FeatureDistillation(
+    distillation_loss = FeatureDistillation(
         loss="cosine_similarity",
         teacher_layer_name="conv2d_2",
         student_layer_name="conv2d_1"
     )
 
     # With custom loss instance
-    strategy = FeatureDistillation(
+    distillation_loss = FeatureDistillation(
         loss=keras.losses.MeanAbsoluteError()
     )
 
     # For multi-output models
-    strategy = FeatureDistillation(
+    distillation_loss = FeatureDistillation(
         loss=["mse", "cosine_similarity"]
     )
 
     # For multi-output models, only distill some outputs
-    strategy = FeatureDistillation(
+    distillation_loss = FeatureDistillation(
         loss=["mse", None, "cosine_similarity"]  # Skip middle output
     )
     ```
@@ -255,11 +256,11 @@ class FeatureDistillation(DistillationLoss):
 
 @keras_export("keras.distillation.LogitsDistillation")
 class LogitsDistillation(DistillationLoss):
-    """Distillation strategy that transfers knowledge from final model outputs.
+    """Distillation loss that transfers knowledge from final model outputs.
 
-    This strategy applies temperature scaling to the teacher's logits before
-    computing the loss between teacher and student predictions. It's the most
-    common approach for knowledge distillation.
+    This distillation_loss applies temperature scaling to the teacher's logits
+    before computing the loss between teacher and student predictions. It's the
+    most common approach for knowledge distillation.
 
     Arguments:
         temperature: Temperature for softmax scaling. Higher values produce
@@ -270,36 +271,36 @@ class LogitsDistillation(DistillationLoss):
               'categorical_crossentropy')
             - Keras loss instance
             - Nested structure of losses matching the model output structure
-            - None to skip distillation for that output (useful for multi-output
-              models where you only want to distill some outputs)
-            At least one loss must be non-None. Defaults to 'kl_divergence'.
+            - `None` to skip distillation for that output (useful for
+              multi-output models where you only want to distill some outputs)
+            At least one loss must be non-`None`. Defaults to 'kl_divergence'.
 
     Examlpe(s):
 
     ```python
     # Basic logits distillation with KL divergence
-    strategy = LogitsDistillation(temperature=3.0)
+    distillation_loss = LogitsDistillation(temperature=3.0)
 
     # With categorical crossentropy loss
-    strategy = LogitsDistillation(
+    distillation_loss = LogitsDistillation(
         temperature=4.0,
         loss="categorical_crossentropy"
     )
 
     # With custom loss instance
-    strategy = LogitsDistillation(
+    distillation_loss = LogitsDistillation(
         temperature=4.0,
         loss=keras.losses.CategoricalCrossentropy(from_logits=True)
     )
 
     # For multi-output models
-    strategy = LogitsDistillation(
+    distillation_loss = LogitsDistillation(
         temperature=3.0,
         loss=["kl_divergence", "categorical_crossentropy"]
     )
 
     # For multi-output models, only distill some outputs
-    strategy = LogitsDistillation(
+    distillation_loss = LogitsDistillation(
         temperature=3.0,
         loss=["kl_divergence", None]  # Skip second output
     )
@@ -317,7 +318,7 @@ class LogitsDistillation(DistillationLoss):
 
         flat_losses = tree.flatten(self.loss)
         if all(l is None for l in flat_losses):
-            raise ValueError("At least one loss must be non-None.")
+            raise ValueError("At least one loss must be non-`None`.")
 
         if not isinstance(self.temperature, (int, float)):
             raise ValueError(
--- a/keras/src/distillation/distillation_loss_test.py
+++ b/keras/src/distillation/distillation_loss_test.py
@@ -10,7 +10,7 @@ from keras.src.testing import TestCase
 
 @pytest.mark.requires_trainable_backend
 class TestLogitsDistillation(TestCase):
-    """Test cases for LogitsDistillation strategy."""
+    """Test cases for LogitsDistillation distillation_loss."""
 
     def test_logits_distillation_basic(self):
         """Test basic logits distillation structure validation."""
@@ -22,16 +22,18 @@ class TestLogitsDistillation(TestCase):
             np.array([[2.0, 1.0, 4.0], [3.0, 6.0, 2.0]]), dtype="float32"
         )
 
-        strategy = LogitsDistillation(temperature=3.0)
-        strategy.validate_outputs(teacher_logits, student_logits)
+        distillation_loss = LogitsDistillation(temperature=3.0)
+        distillation_loss.validate_outputs(teacher_logits, student_logits)
         incompatible_logits = {"output": teacher_logits}
         with self.assertRaises(ValueError):
-            strategy.validate_outputs(teacher_logits, incompatible_logits)
+            distillation_loss.validate_outputs(
+                teacher_logits, incompatible_logits
+            )
 
 
 @pytest.mark.requires_trainable_backend
 class TestFeatureDistillation(TestCase):
-    """Test cases for FeatureDistillation strategy."""
+    """Test cases for FeatureDistillation distillation_loss."""
 
     def test_feature_distillation_basic(self):
         """Test basic feature distillation structure validation."""
@@ -43,11 +45,13 @@ class TestFeatureDistillation(TestCase):
             np.array([[1.1, 2.1, 3.1], [4.1, 5.1, 6.1]]), dtype="float32"
         )
 
-        strategy = FeatureDistillation(loss="mse")
-        strategy.validate_outputs(teacher_features, student_features)
+        distillation_loss = FeatureDistillation(loss="mse")
+        distillation_loss.validate_outputs(teacher_features, student_features)
         incompatible_features = [teacher_features, teacher_features]
         with self.assertRaises(ValueError):
-            strategy.validate_outputs(teacher_features, incompatible_features)
+            distillation_loss.validate_outputs(
+                teacher_features, incompatible_features
+            )
 
 
 @pytest.mark.requires_trainable_backend
@@ -96,7 +100,7 @@ class TestEndToEndDistillation(TestCase):
         distiller = Distiller(
             teacher=self.teacher,
             student=self.student,
-            strategies=LogitsDistillation(temperature=3.0),
+            distillation_loss=LogitsDistillation(temperature=3.0),
             student_loss_weight=0.5,
         )
 
@@ -134,7 +138,7 @@ class TestEndToEndDistillation(TestCase):
         distiller = Distiller(
             teacher=self.teacher,
             student=self.student,
-            strategies=FeatureDistillation(
+            distillation_loss=FeatureDistillation(
                 loss="mse",
                 teacher_layer_name="teacher_dense_1",
                 student_layer_name="student_dense_1",
@@ -169,10 +173,10 @@ class TestEndToEndDistillation(TestCase):
             len(distiller._student_feature_extractor.outputs), 2
         )  # final + dense_1
 
-    def test_multi_strategy_distillation_end_to_end(self):
-        """Test end-to-end distillation with multiple strategies."""
-        # Create multiple strategies
-        strategies = [
+    def test_multi_distillation_loss_distillation_end_to_end(self):
+        """Test end-to-end distillation with multiple distillation_loss."""
+        # Create multiple distillation_loss
+        distillation_loss = [
             LogitsDistillation(temperature=3.0),
             FeatureDistillation(
                 loss="mse",
@@ -190,8 +194,8 @@ class TestEndToEndDistillation(TestCase):
         distiller = Distiller(
             teacher=self.teacher,
             student=self.student,
-            strategies=strategies,
-            strategy_weights=[1.0, 0.5, 0.3],
+            distillation_loss=distillation_loss,
+            distillation_loss_weights=[1.0, 0.5, 0.3],
             student_loss_weight=0.5,
         )
 
--- a/keras/src/distillation/distiller.py
+++ b/keras/src/distillation/distiller.py
@@ -19,11 +19,13 @@ class Distiller(Model):
         teacher: A trained `keras.Model` that serves as the knowledge source.
             The teacher model is frozen during distillation.
         student: A `keras.Model` to be trained through distillation.
-        strategies: List of distillation strategies to apply. Can be a single
-            strategy or a list of strategies like `LogitsDistillation`,
-            `FeatureDistillation`, or custom distillation strategies.
-        strategy_weights: List of weights for each distillation strategy. Must
-            have the same length as `strategies`. If None, equal weights used.
+        distillation_loss: List of distillation strategies to apply.
+            Can be a single strategy or a list of strategies like
+            `LogitsDistillation`, `FeatureDistillation`, or custom distillation
+            strategies.
+        distillation_loss_weights: List of weights for each distillation
+            strategy. Must have the same length as `distillation_loss`.
+            If None, equal weights are used.
         student_loss_weight: Weight for the student's supervised loss component.
             Must be between 0 and 1. Defaults to 0.5.
         name: Name for the distiller model. Defaults to `"distiller"`.
@@ -94,8 +96,8 @@ class Distiller(Model):
         self,
         teacher,
         student,
-        strategies,
-        strategy_weights=None,
+        distillation_loss,
+        distillation_loss_weights=None,
         student_loss_weight=0.5,
         name="distiller",
         **kwargs,
@@ -122,33 +124,35 @@ class Distiller(Model):
             )
         self.student_loss_weight = student_loss_weight
 
-        # Handle strategies configuration
-        if strategies is None:
+        # Handle distillation_loss configuration
+        if distillation_loss is None:
             raise ValueError(
-                "'strategies' cannot be None. Provide a distillation "
+                "'distillation_loss' cannot be `None`. Provide a distillation "
                 "strategy (e.g., LogitsDistillation or FeatureDistillation) "
                 "or a list of strategies."
             )
 
         # Convert single strategy to list for uniform handling
-        if not isinstance(strategies, (list, tuple)):
-            self.strategies = [strategies]
-            self.strategy_weights = [1.0]
+        if not isinstance(distillation_loss, (list, tuple)):
+            self.distillation_loss = [distillation_loss]
+            self.distillation_loss_weights = [1.0]
         else:
-            self.strategies = strategies
+            self.distillation_loss = distillation_loss
             # Set default weights if not provided
-            if strategy_weights is None:
-                self.strategy_weights = [1.0] * len(strategies)
+            if distillation_loss_weights is None:
+                self.distillation_loss_weights = [1.0] * len(distillation_loss)
             else:
-                if len(strategy_weights) != len(strategies):
+                if len(distillation_loss_weights) != len(distillation_loss):
                     raise ValueError(
-                        f"Number of strategy_weights ({len(strategy_weights)}) "
-                        f"must match number of strategies ({len(strategies)})"
+                        "Number of distillation_loss_weights "
+                        f"({len(distillation_loss_weights)}) must match "
+                        "number of distillation_loss "
+                        f"({len(distillation_loss)})"
                     )
-                self.strategy_weights = strategy_weights
+                self.distillation_loss_weights = distillation_loss_weights
 
         # Validate strategy-specific compatibility and create feature extractors
-        for strategy in self.strategies:
+        for strategy in self.distillation_loss:
             self._validate_strategy_compatibility(teacher, student, strategy)
 
         self._create_multi_feature_extractors()
@@ -272,7 +276,7 @@ class Distiller(Model):
         teacher_layer_names = []
         student_layer_names = []
 
-        for strategy in self.strategies:
+        for strategy in self.distillation_loss:
             if (
                 hasattr(strategy, "teacher_layer_name")
                 and strategy.teacher_layer_name
@@ -296,15 +300,16 @@ class Distiller(Model):
     def _create_feature_extractor(self, model, layer_names):
         """Create a feature extractor for a model.
 
-        Arguments:
-            model: The model to create an extractor for.
-            layer_names: List of layer names to extract features from.
+                Arguments:
+                    model: The model to create an extractor for.
+                    layer_names: List of layer names to extract features from.
 
-        Returns:
-            Feature extractor model or None if no layer names provided.
-
-        Raises:
-            ValueError: If model has no symbolic inputs/outputs.
+                Returns:
+                    Feature extractor model or `None` if no layer names
+                        sprovided.
+        `
+                Raises:
+                    ValueError: If model has no symbolic inputs/outputs.
         """
         if not layer_names:
             return None
@@ -371,7 +376,7 @@ class Distiller(Model):
             **kwargs: Additional arguments passed to parent compile.
         """
         if loss is None:
-            raise ValueError("'loss' cannot be None.")
+            raise ValueError("'loss' cannot be `None`.")
 
         self._student_loss = tree.map_structure(_convert_loss_to_function, loss)
         self._student_loss_for_serialization = loss
@@ -437,7 +442,9 @@ class Distiller(Model):
             student_features = self._extract_all_student_features(x, y_pred)
 
             # Apply strategies using pre-extracted features
-            for strategy, weight in zip(self.strategies, self.strategy_weights):
+            for strategy, weight in zip(
+                self.distillation_loss, self.distillation_loss_weights
+            ):
                 # Get appropriate outputs/features for this strategy
                 if (
                     hasattr(strategy, "teacher_layer_name")
@@ -525,11 +532,11 @@ class Distiller(Model):
                 "student": serialization_lib.serialize_keras_object(
                     self.student
                 ),
-                "strategies": [
+                "distillation_loss": [
                     serialization_lib.serialize_keras_object(strategy)
-                    for strategy in self.strategies
+                    for strategy in self.distillation_loss
                 ],
-                "strategy_weights": self.strategy_weights,
+                "distillation_loss_weights": self.distillation_loss_weights,
                 "student_loss_weight": self.student_loss_weight,
             }
         )
@@ -547,9 +554,9 @@ class Distiller(Model):
         config["student"] = serialization_lib.deserialize_keras_object(
             config["student"]
         )
-        config["strategies"] = [
+        config["distillation_loss"] = [
             serialization_lib.deserialize_keras_object(strategy)
-            for strategy in config["strategies"]
+            for strategy in config["distillation_loss"]
         ]
 
         return cls(**config)
--- a/keras/src/distillation/distiller_test.py
+++ b/keras/src/distillation/distiller_test.py
@@ -1,6 +1,5 @@
 import json
 import os
-import tempfile
 
 import numpy as np
 import pytest
@@ -58,14 +57,14 @@ class TestDistiller(TestCase):
         self.teacher(dummy_input)
         self.student(dummy_input)
 
-        # Create distillation strategy
-        self.strategy = LogitsDistillation(temperature=2.0)
+        # Create distillation distillation_loss
+        self.distillation_loss = LogitsDistillation(temperature=2.0)
 
         # Create distiller
         self.distiller = Distiller(
             teacher=self.teacher,
             student=self.student,
-            strategies=self.strategy,
+            distillation_loss=self.distillation_loss,
             student_loss_weight=0.5,
         )
 
@@ -87,13 +86,15 @@ class TestDistiller(TestCase):
         # Check student_loss_weight
         self.assertEqual(self.distiller.student_loss_weight, 0.5)
 
-        # Check strategies (should be a list with one strategy)
-        self.assertIsInstance(self.distiller.strategies, list)
-        self.assertEqual(len(self.distiller.strategies), 1)
-        self.assertIsInstance(self.distiller.strategies[0], LogitsDistillation)
+        # Check distillation_loss (should be a list with one distillation_loss)
+        self.assertIsInstance(self.distiller.distillation_loss, list)
+        self.assertEqual(len(self.distiller.distillation_loss), 1)
+        self.assertIsInstance(
+            self.distiller.distillation_loss[0], LogitsDistillation
+        )
 
-        # Check that strategy has the correct temperature
-        self.assertEqual(self.distiller.strategies[0].temperature, 2.0)
+        # Check that distillation_loss has the correct temperature
+        self.assertEqual(self.distiller.distillation_loss[0].temperature, 2.0)
 
         # Check that model is compiled
         self.assertIsNotNone(self.distiller.optimizer)
@@ -134,7 +135,7 @@ class TestDistiller(TestCase):
         Distiller(
             teacher=new_teacher,
             student=self.student,
-            strategies=self.strategy,
+            distillation_loss=self.distillation_loss,
             student_loss_weight=0.5,
         )
 
@@ -148,31 +149,31 @@ class TestDistiller(TestCase):
             Distiller(
                 teacher="not_a_model",
                 student=self.student,
-                strategies=self.strategy,
+                distillation_loss=self.distillation_loss,
             )
 
         with self.assertRaises(ValueError):
             Distiller(
                 teacher=self.teacher,
                 student="not_a_model",
-                strategies=self.strategy,
+                distillation_loss=self.distillation_loss,
             )
 
-    def test_multi_strategy_functionality(self):
-        """Test multi-strategy functionality."""
-        # Create multiple strategies
-        strategies = [
+    def test_multi_distillation_loss_functionality(self):
+        """Test multi-distillation_loss functionality."""
+        # Create multiple distillation_loss
+        distillation_loss = [
             LogitsDistillation(temperature=3.0),
             LogitsDistillation(temperature=2.0),
         ]
-        strategy_weights = [0.7, 0.3]
+        distillation_loss_weights = [0.7, 0.3]
 
-        # Create distiller with multiple strategies
+        # Create distiller with multiple distillation_loss
         distiller = Distiller(
             teacher=self.teacher,
             student=self.student,
-            strategies=strategies,
-            strategy_weights=strategy_weights,
+            distillation_loss=distillation_loss,
+            distillation_loss_weights=distillation_loss_weights,
             student_loss_weight=0.5,
         )
 
@@ -183,9 +184,9 @@ class TestDistiller(TestCase):
             metrics=["accuracy"],
         )
 
-        # Test that strategies are stored correctly
-        self.assertEqual(len(distiller.strategies), 2)
-        self.assertEqual(distiller.strategy_weights, [0.7, 0.3])
+        # Test that distillation_loss are stored correctly
+        self.assertEqual(len(distiller.distillation_loss), 2)
+        self.assertEqual(distiller.distillation_loss_weights, [0.7, 0.3])
 
         # Test training
         x = np.random.random((10, 5)).astype(np.float32)
@@ -197,9 +198,9 @@ class TestDistiller(TestCase):
         self.assertIn("student_loss", history.history)
         self.assertIn("distillation_loss", history.history)
 
-    def test_multi_strategy_validation(self):
-        """Test multi-strategy validation."""
-        strategies = [
+    def test_multi_distillation_loss_validation(self):
+        """Test multi-distillation_loss validation."""
+        distillation_loss = [
             LogitsDistillation(temperature=3.0),
             LogitsDistillation(temperature=2.0),
         ]
@@ -208,19 +209,19 @@ class TestDistiller(TestCase):
         distiller = Distiller(
             teacher=self.teacher,
             student=self.student,
-            strategies=strategies,
+            distillation_loss=distillation_loss,
             student_loss_weight=0.5,
         )
 
-        self.assertEqual(len(distiller.strategies), 2)
+        self.assertEqual(len(distiller.distillation_loss), 2)
 
-        # Test invalid strategy weights length
+        # Test invalid distillation_loss weights length
         with self.assertRaises(ValueError):
             Distiller(
                 teacher=self.teacher,
                 student=self.student,
-                strategies=strategies,
-                strategy_weights=[1.0],  # Wrong length
+                distillation_loss=distillation_loss,
+                distillation_loss_weights=[1.0],  # Wrong length
                 student_loss_weight=0.5,
             )
 
@@ -230,7 +231,7 @@ class TestDistiller(TestCase):
         distiller_0 = Distiller(
             teacher=self.teacher,
             student=self.student,
-            strategies=self.strategy,
+            distillation_loss=self.distillation_loss,
             student_loss_weight=0.0,
         )
 
@@ -238,7 +239,7 @@ class TestDistiller(TestCase):
         distiller_1 = Distiller(
             teacher=self.teacher,
             student=self.student,
-            strategies=self.strategy,
+            distillation_loss=self.distillation_loss,
             student_loss_weight=1.0,
         )
 
@@ -288,7 +289,7 @@ class TestDistiller(TestCase):
         distiller = Distiller(
             teacher=teacher,
             student=student,
-            strategies=self.strategy,
+            distillation_loss=self.distillation_loss,
             student_loss_weight=0.5,
         )
 
@@ -363,7 +364,7 @@ class TestDistiller(TestCase):
         distiller = Distiller(
             teacher=teacher,
             student=student,
-            strategies=self.strategy,
+            distillation_loss=self.distillation_loss,
             student_loss_weight=0.5,
         )
 
@@ -407,7 +408,7 @@ class TestDistiller(TestCase):
         distiller = Distiller(
             teacher=teacher,
             student=student,
-            strategies=self.strategy,
+            distillation_loss=self.distillation_loss,
             student_loss_weight=0.5,
         )
 
@@ -452,13 +453,15 @@ class TestDistiller(TestCase):
             ]
         )
 
-        # Create distiller with single strategy
-        strategy = LogitsDistillation(temperature=3.0, loss="kl_divergence")
+        # Create distiller with single distillation_loss
+        distillation_loss = LogitsDistillation(
+            temperature=3.0, loss="kl_divergence"
+        )
 
         original_distiller = Distiller(
             teacher=teacher,
             student=student,
-            strategies=strategy,
+            distillation_loss=distillation_loss,
             student_loss_weight=0.7,
         )
 
@@ -473,8 +476,8 @@ class TestDistiller(TestCase):
         required_keys = [
             "teacher",
             "student",
-            "strategies",
-            "strategy_weights",
+            "distillation_loss",
+            "distillation_loss_weights",
             "student_loss_weight",
         ]
         for key in required_keys:
@@ -490,37 +493,39 @@ class TestDistiller(TestCase):
         # Verify reconstruction
         self.assertEqual(reconstructed_distiller.student_loss_weight, 0.7)
         self.assertIsInstance(
-            reconstructed_distiller.strategies[0], LogitsDistillation
+            reconstructed_distiller.distillation_loss[0], LogitsDistillation
         )
 
-        # Verify strategy parameters
-        self.assertEqual(reconstructed_distiller.strategies[0].temperature, 3.0)
+        # Verify distillation_loss parameters
+        self.assertEqual(
+            reconstructed_distiller.distillation_loss[0].temperature, 3.0
+        )
 
         # Test that reconstructed distiller can be used for inference
         reconstructed_output = reconstructed_distiller(x_test)
         self.assertEqual(reconstructed_output.shape, (2, 10))
 
         # Test model saving and loading (full integration test)
-        with tempfile.TemporaryDirectory() as temp_dir:
-            model_path = os.path.join(temp_dir, "distiller_model.keras")
+        temp_dir = self.get_temp_dir()
+        model_path = os.path.join(temp_dir, "distiller_model.keras")
 
-            # Compile original distiller
-            original_distiller.compile(
-                loss="sparse_categorical_crossentropy",
-            )
+        # Compile original distiller
+        original_distiller.compile(
+            loss="sparse_categorical_crossentropy",
+        )
 
-            # Save the model
-            original_distiller.save(model_path)
+        # Save the model
+        original_distiller.save(model_path)
 
-            # Load the model
-            loaded_distiller = keras.models.load_model(model_path)
+        # Load the model
+        loaded_distiller = keras.models.load_model(model_path)
 
-            # Verify loaded model works
-            loaded_output = loaded_distiller(x_test)
-            self.assertEqual(loaded_output.shape, (2, 10))
+        # Verify loaded model works
+        loaded_output = loaded_distiller(x_test)
+        self.assertEqual(loaded_output.shape, (2, 10))
 
-            # Verify parameters are preserved
-            self.assertEqual(loaded_distiller.student_loss_weight, 0.7)
+        # Verify parameters are preserved
+        self.assertEqual(loaded_distiller.student_loss_weight, 0.7)
 
         # The core serialization functionality is working
         self.assertTrue(True, "Distiller serialization test passed")


keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,437 @@
...
+
+@keras_export("keras.distillation.Distiller")
+class Distiller(Model):
+    """Knowledge Distillation model.
+
+    This class implements knowledge distillation by combining a teacher model
+    and a student model with configurable distillation strategies.
+
+    The Distiller integrates seamlessly with Keras's training infrastructure
+    by overriding the _compute_loss method, allowing standard model.fit(),
```
fchollet: Use backticks for code keyword. `compute_loss` is a public method. No need for "seamlessly". Avoid LLM vibes

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,437 @@
...
+    model.evaluate(), and model.predict() workflows to work correctly.
+
+    Args:
+        teacher: The teacher model (will be frozen during training).
+        student: The student model to be trained.
+        strategies: List of distillation strategies to apply.
+        student_loss_fn: Loss function for student predictions. Defaults to
```
fchollet: Just `student_loss` (for consistency with `loss` arg)

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,437 @@
...
+    and a student model with configurable distillation strategies.
+
+    The Distiller integrates seamlessly with Keras's training infrastructure
+    by overriding the _compute_loss method, allowing standard model.fit(),
+    model.evaluate(), and model.predict() workflows to work correctly.
+
+    Args:
+        teacher: The teacher model (will be frozen during training).
+        student: The student model to be trained.
+        strategies: List of distillation strategies to apply.
```
fchollet: What happens if you pass more than one?

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,437 @@
...
+        student_loss_fn: Loss function for student predictions. Defaults to
+            sparse categorical crossentropy.
+        alpha: Weight for combining student loss and distillation loss.
```
fchollet: Why is distillation loss not configurable?

Avoid "alpha" as an argument name (too generic)

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,437 @@
...
+            alpha=1.0 means only student loss, alpha=0.0 means only
+            distillation loss.
+        temperature: Default temperature for distillation strategies that don't
```
fchollet: This should only be configurable in 1 place (e.g. at the level of the strategy).

keras/src/distillation/strategies.py
```diff
@@ -0,0 +1,566 @@
...
+                f"Teacher has {len(teacher_outputs)} outputs, "
+                f"student has {len(student_outputs)} outputs."
+            )
+
+
+@keras_export("keras.distillation.LogitsDistillation")
+class LogitsDistillation(BaseDistillationStrategy):
+    """Logits distillation strategy using Keras built-in loss functions.
+
+    This strategy distills knowledge using the logits (pre-softmax outputs)
```
fchollet: Explain the distillation process in detail.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,437 @@
...
+from keras.src.api_export import keras_export
+from keras.src.models.model import Model
+
+
+@keras_export("keras.distillation.Distiller")
+class Distiller(Model):
+    """Knowledge Distillation model.
+
+    This class implements knowledge distillation by combining a teacher model
+    and a student model with configurable distillation strategies.
```
fchollet: Provide more info about what distillation is and why/when you would want to use it.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,437 @@
...
+
+    **Basic Knowledge Distillation:**
+
+    ```python
+    import keras
+    import numpy as np
+    from keras.distillation import Distiller, LogitsDistillation
+
+    # Create teacher and student models
+    teacher = keras.Sequential([
```
fchollet: Use LLMs from KerasHub for this

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,437 @@
...
+    ])
+
+    # Create distillation strategy (will use Distiller's default temperature)
+    strategy = LogitsDistillation()
+
+    # Create distiller with default temperature
+    distiller = Distiller(
+        teacher=teacher,
+        student=student,
+        strategies=[strategy],
```
fchollet: In-line the strategy here. Note, if the typical use case is to pass 1 strategy, then the argument should not be a list. (but it could accept a list, optionally).

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,437 @@
...
+        alpha=0.7,  # 70% student loss, 30% distillation loss
+        temperature=4.0  # Default temperature for all strategies
+    )
+
+    # Compile and train
+    distiller.compile(
```
fchollet, line 60: A separate compile step might not be necessary. Why not pass these 2 args to the distiller instead?

Or, custom compile():

```
distiller = Distiller(
        teacher=teacher,
        student=student,
        strategy=strategy
)
distiller.compile(distillation_loss=..., student_loss=..., optimizers=..., temperature=...)
```

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,437 @@
...
+        optimizer='adam',
+        loss='sparse_categorical_crossentropy'
+    )
+
+    # Generate dummy data
```
fchollet: Consider using real data

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,437 @@
...
+    y_train = np.random.randint(0, 10, (1000,))
+
+    # Train the distiller
+    distiller.fit(x_train, y_train, epochs=10, batch_size=32)
+
+    # Use the trained student model
+    predictions = distiller.predict(x_train[:5])
+    ```
+
+    **Multi-Strategy Distillation:**
```
fchollet: We need to specify info about when to pick which strategy and what they do.

keras/src/distillation/distiller.py
divyashreepathihalli: users can override `compute_loss` to define custom loss

keras/src/distillation/distiller.py
divyashreepathihalli: changed this back to only allowing single strategy. If users have multiple outputs, multiple strategies can be configured with `MultiOutputDistillation` strategy

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,684 @@
...
+            class.
+
+    Example:
+
+    ```python
+    # Load pre-trained teacher model from KerasHub
+    import keras_hub as hub
+
+    teacher = hub.models.CausalLM.from_preset("gemma3_4b_en")
+    student = hub.models.CausalLM.from_preset("gemma2_2b_en")
```
hertschuh: This feels like cheating because `hub.models.CausalLM.from_preset("gemma2_2b_en")` is already trained. Also, probably the distillation will degrade performance, not improve it, if people try this for real.

Is there a way to instantiate an empty (randomly initialized) `gemma2_2b` in 1 line of code?

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,684 @@
...
+        optimizer='adam',
+        student_loss='sparse_categorical_crossentropy',
+        metrics=['accuracy']
+    )
+
+    # Train the distiller
+    distiller.fit(x_train, y_train, epochs=10, validation_split=0.2)
+
+    # Get the trained student model
+    trained_student = distiller.get_student_model()
```
hertschuh: Remove per discussion in the doc.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,684 @@
...
+                def compute_loss(self, x=None, y=None, y_pred=None,
+                               sample_weight=None, training=None):
+                    # Custom student loss computation
+                    student_loss = keras.losses.sparse_categorical_crossentropy(
+                        y, y_pred
+                    )
+
+                    # Custom distillation loss computation
+                    teacher_outputs = self.teacher(x, training=False)
+                    student_outputs = self.student(x, training=training)
```
hertschuh: You have to use `y_pred` and not compute `self.student(x, training=training)` again, otherwise the gradients will not be connected to the loss and it won't work.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,684 @@
...
+
+                    # Combine losses with custom weighting
+                    total_loss = 0.7 * student_loss + 0.3 * distillation_loss
+
+                    return total_loss
+
+                def _custom_distillation_loss(self, teacher_outputs,
+                                            student_outputs):
+                    # Implement custom distillation loss logic
+                    from keras import ops
```
hertschuh: Remove local import, instead write:
```
return keras.ops.mean(
   keras.ops.square(...
```
on the next line

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,684 @@
...
+
+        # Convert string loss to function if needed
+        if isinstance(student_loss, str):
+            self._student_loss = keras.losses.get(student_loss)
+            if self._student_loss is None:
+                raise ValueError(
+                    f"Unknown loss function: '{student_loss}'. "
+                    "Please provide a valid loss function name or instance."
+                )
+        elif isinstance(student_loss, list):
```
hertschuh: What if it's a `dict`? Wouldn't it be easier to do a `tree.map_structure`? It would handle all cases (lines 182-205)

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,684 @@
...
+            y_pred = [y_pred]
+        if y is not None and not isinstance(y, (list, tuple)):
+            y = [y]
+
+        # Compute student loss
+        student_loss = 0.0
+        if self.student_loss_weight > 0.0 and y is not None:
+            # Use the configured loss function
+            if (
+                hasattr(self, "_student_loss")
```
hertschuh: Remove this `if`, you have a check line 208 that `self._student_loss` exists and is not None.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,684 @@
...
+                and self._student_loss is not None
+            ):
+                if isinstance(self._student_loss, list):
```
hertschuh: Multi-output can be arbitrarily nested or can be a `dict`.

Use `tree` everywhere:
- Validate with `tree.assert_same_structures`
- `tree.flatten` before summing

It will also make the code shorter.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,684 @@
...
+                        if len(self._student_loss) != len(y):
+                            raise ValueError(
+                                f"Number of loss functions "
+                                f"({len(self._student_loss)}) must match "
+                                f"number of outputs ({len(y)}) for "
+                                f"multi-output loss computation."
+                            )
+
+                        # Compute loss for each output
+                        student_loss = sum(
```
hertschuh: Use `ops.add` instead of the Python `sum`.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,684 @@
...
+                        # Use first output for student loss (consistent
+                        # behavior)
+                        student_loss = self._student_loss(y[0], y_pred[0])
+                    else:
+                        # Single output with single loss function
+                        student_loss = self._student_loss(y, y_pred)
+            else:
+                # No loss function configured - this is an error
+                raise ValueError(
+                    "Student loss function is not configured. "
```
hertschuh: Already done line 208

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,684 @@
...
+        student_loss_weight=0.5,
+        optimizer="adam",
+        student_loss="sparse_categorical_crossentropy",
+        metrics=None,
+        name="distiller",
+        **kwargs,
+    ):
+        # Extract input_mapping and output_mapping before super().__init__
+        self.input_mapping = kwargs.pop("input_mapping", None)
+        self.output_mapping = kwargs.pop("output_mapping", None)
```
hertschuh: `self.input_mapping` and `self.output_mapping` don't appear to be used. Remove?

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,684 @@
...
+                ),
+                "student": serialization_lib.serialize_keras_object(
+                    self.student
+                ),
+                "strategy": serialization_lib.serialize_keras_object(
+                    self.strategy
+                ),
+                "student_loss_weight": self.student_loss_weight,
+                "input_mapping": self.input_mapping,
+                "output_mapping": self.output_mapping,
```
hertschuh: `self.input_mapping` and `self.output_mapping` don't appear to be used. Remove?

keras/src/distillation/strategies.py
```diff
@@ -0,0 +1,861 @@
+import keras
+from keras.src import ops
+from keras.src.api_export import keras_export
+from keras.src.saving import serialization_lib
+
+
+@keras_export("keras.distillation.BaseDistillationStrategy")
+class BaseDistillationStrategy:
```
hertschuh: Let's discuss the name:

- `BaseDistillationStrategy`
  - About **`Base`**: we do use that prefix in a few places (e.g. `BaseConv`), but it's not exposed publicly. Typically the abstract base class doesn't have `Base` in it: `Layer`, `Loss`, `Metric`, `Initializer`...
  - About **`Strategy`**: just curious why you called it "strategy". To me a strategy is a way to select an algorithm. But in this case, it's not really a pluggable "algorithm", it's just what the loss is applied on (outputs vs. intermediary output). In fact it is a loss. Also, the subclasses are not called "strategy", like `LogitsDistillation`.
  - About **`Distillation`**: distillation is the name of the whole process. Now that I've argued to remove `Base` and `Strategy` to the name, we're left with `Distillation`, but that's confusing, because what this class does is a small part of the distillation process.

So should it be called **`DistillationLoss`**?

keras/src/distillation/strategies.py
```diff
@@ -0,0 +1,861 @@
...
+        shapes, dimensions, and other requirements.
+
+        Args:
+            teacher_outputs: Outputs from the teacher model.
+            student_outputs: Outputs from the student model.
+        Raises:
+            ValueError: If outputs are not compatible.
+        """
+        # Default implementation - can be overridden by subclasses
+        if not isinstance(teacher_outputs, (list, tuple)):
```
hertschuh: Let's allow arbitrary nested structures and just do `keras.tree.assert_same_structure`.

keras/src/distillation/strategies.py
```diff
@@ -0,0 +1,861 @@
...
+        temperature=3.0,
+        output_index=1  # Use second output
+    )
+    ```
+    """
+
+    def __init__(
+        self,
+        temperature=3.0,
+        loss_type="kl_divergence",
```
hertschuh: About **`loss_type`**: can we just allow regular Keras losses (as either instance of `Loss` or strings)? And then, can we just call it `loss`?

keras/src/distillation/strategies.py
```diff
@@ -0,0 +1,861 @@
...
+        output_index=0,
```
hertschuh: This assumes the outputs are a list. What if you want to use a dict?

The way keras does losses is that the loss you pass is an arbitrary nested structure that matches the structure of the outputs. We could do that here, but then you don't need `MultiOutputDistillation` for this anymore.

keras/src/distillation/strategies.py
```diff
@@ -0,0 +1,861 @@
...
+    def get_config(self):
+        """Get configuration for serialization."""
+        return {
+            "temperature": self.temperature,
+            "loss_type": self.loss_type,
+            "output_index": self.output_index,
+        }
+
+    @classmethod
+    def from_config(cls, config):
```
hertschuh: Not needed AFAIK.

keras/src/distillation/strategies.py
```diff
@@ -0,0 +1,861 @@
...
+    def get_config(self):
+        """Get configuration for serialization."""
+        return {
+            "loss_type": self.loss_type,
+            "teacher_layer_name": self.teacher_layer_name,
+            "student_layer_name": self.student_layer_name,
+        }
+
+    @classmethod
+    def from_config(cls, config):
```
hertschuh: Not needed AFAIK.

keras/src/distillation/strategies.py
```diff
@@ -0,0 +1,861 @@
...
+    strategy = FeatureDistillation(
+        loss_type="mse",
+        teacher_layer_name=None,  # Final output
+        student_layer_name=None   # Final output
+    )
+    ```
+    """
+
+    def __init__(
+        self, loss_type="mse", teacher_layer_name=None, student_layer_name=None
```
hertschuh: About **`loss_type`**: can we just allow regular Keras losses (as either instance of `Loss` or strings)? And then, can we just call it `loss`?

keras/src/distillation/strategies.py
```diff
@@ -0,0 +1,861 @@
...
+    strategy = FeatureDistillation(loss_type="mse")
+
+    # Custom loss by subclassing
+    class CustomFeatureDistillation(FeatureDistillation):
+        def compute_loss(self, teacher_outputs, student_outputs, **kwargs):
+            # Use first output by default
+            teacher_features = teacher_outputs[0]
+            student_features = student_outputs[0]
+
+            # Custom L1 loss for feature distillation
```
hertschuh: Can we just allow passing L1 as the loss in the current `FeatureDistillation`?

Any other good reason to subclass `FeatureDistillation`?

keras/src/distillation/strategies.py
```diff
@@ -0,0 +1,861 @@
...
+    )
+
+    # Use cosine similarity for different feature sizes
+    strategy = FeatureDistillation(
+        loss_type="cosine",
+        teacher_layer_name="conv2d_2",
+        student_layer_name="conv2d_1"
+    )
+
+    # Distill from final outputs (equivalent to logits distillation)
```
hertschuh: Hmm... so why do we need `LogitsDistillation` then?

keras/src/distillation/strategies.py
```diff
@@ -0,0 +1,861 @@
...
+                        self._create_feature_extractor(
+                            teacher_model, self.teacher_layer_name
+                        )
+                    )
+                else:
+                    raise
+
+        return self._teacher_feature_model(inputs, training=False)
+
+    def _get_student_features(self, student_model, inputs):
```
hertschuh: Can this be refactored so that we have one function instead of both `_get_teacher_features` and `_get_student_features`?

keras/src/distillation/strategies.py
```diff
@@ -0,0 +1,861 @@
...
+        Returns:
+            A keras.Model that extracts features from the specified layer.
+        """
+        if layer_name is None:
+            # Return the original model if no layer specified
+            return model
+
+        # Find the layer by name
+        target_layer = None
+        for layer in model.layers:
```
hertschuh: This already exists in `Layer`:
```
target_layer = model.get_layer(name=layer_name)
```

keras/src/distillation/strategies.py
```diff
@@ -0,0 +1,861 @@
...
+        # compatibility when using feature distillation with specific layer
+        # names
+        pass
+
+    def validate_model_compatibility(self, teacher, student):
+        """Validate that teacher and student models are compatible for feature
+        distillation."""
+        # Check if specified layers exist in the models
+        if self.teacher_layer_name is not None:
+            if not self._layer_exists_in_model(
```
hertschuh: If would be a lot less code and wouldn't require `_layer_exists_in_model` to just reuse `model.get_layer(name=layer_name)` and add to the error message.

```python
try:
  model.get_layer(name=layer_name)
except ValueError as e:
  raise ValueError("In teacher model: {e}")
```

keras/src/distillation/strategies.py
```diff
@@ -0,0 +1,861 @@
...
+                Can be a single tensor or a list/tuple of tensors.
+            student_outputs: Intermediate features from student model.
+                Can be a single tensor or a list/tuple of tensors.
+            **kwargs: Additional arguments (ignored).
+        Returns:
+            Feature distillation loss tensor.
+        """
+
+        # Normalize outputs to lists
+        if not isinstance(teacher_outputs, (list, tuple)):
```
hertschuh: Same note about dict outputs are deeply nested outputs.

keras/src/distillation/strategies.py
```diff
@@ -0,0 +1,861 @@
...
+        # Feature extraction models (created when needed)
+        self._teacher_feature_model = None
+        self._student_feature_model = None
+
+        # Validate loss_type
+        valid_loss_types = ["mse", "cosine"]
+        if loss_type not in valid_loss_types:
+            raise ValueError(f"loss_type must be one of {valid_loss_types}")
+
+    def _get_teacher_features(self, teacher_model, inputs):
```
hertschuh: Wait, this never used, something is missing.

keras/src/distillation/strategies.py
hertschuh: Also, it's never used, something is missing.

keras/src/distillation/strategies.py
```diff
@@ -0,0 +1,861 @@
...
+    def compute_loss(self, teacher_outputs, student_outputs, **kwargs):
+        """Compute feature distillation loss using extracted features.
+
+        Note: This method expects the outputs to already be the extracted
+        features from the specified layers, not the final model outputs.
+        The Distiller class is responsible for extracting the features
+        using the methods provided by this strategy.
+
+        Args:
+            teacher_outputs: Intermediate features from teacher model.
```
hertschuh: > Intermediate features from teacher model.

There isn't code that makes this work. The distiller will pass the outputs to the model.

keras/src/distillation/strategies.py
```diff
@@ -0,0 +1,861 @@
...
+                f"This may happen with a subclassed model that cannot be "
+                f"traversed using the standard layer API. "
+                f"Available layers: "
+                f"{[layer.name for layer in model.layers]}"
+            )
+
+        # Create a new model that extracts features from the specified layer.
+        # This approach is robust for models created with the Functional API.
+        try:
+            return keras.Model(
```
hertschuh: First there is an issue whereby this model is not actually used.

But the second issue is that by having these models separately, you're going to re-run the same calculations many times. For instance, if you do 10 `FeatureDistillation`s with 10 different layers, you're going to re-run the initial layers 10 times.

What would be more efficient is this:

```python
keras.Model(
    inputs=model.inputs,
    outputs=[model.output, target_layer1.output, target_layer2.output, ...]
)
```

However, the issue right now is that `FeatureDistillation` only handles one intermediary layer's output. So to be able to do this, either `FeatureDistillation` should be augmented to support multiple layers, or `MultiOutputDistillation` should be smart enough to combine them in this way.

keras/src/distillation/strategies.py
```diff
@@ -0,0 +1,861 @@
...
+        }
+
+    @classmethod
+    def from_config(cls, config):
+        """Create instance from configuration."""
+        return cls(**config)
+
+
+@keras_export("keras.distillation.MultiOutputDistillation")
+class MultiOutputDistillation(BaseDistillationStrategy):
```
hertschuh: I'm a bit confused about what this class does.

Base on lines 826-833:

```python
        for output_idx, strategy in self.output_strategies.items():
            teacher_output = teacher_outputs[output_idx]
            student_output = student_outputs[output_idx]

            # Compute loss for this output
            output_loss = strategy.compute_loss(
                [teacher_output], [student_output], **kwargs
            )
```

It is indeed intended to map a separate `LogitsDistillation` per output of the model for a multi-output model. And it does so by index, which assumes that outputs are always lists (which, to me, is too strong of an assumption).

But then, the documentation clearly says that you can also use `FeatureDistillation`. However, this code:
- doesn't compute the intermediary features needed by `FeatureDistillation`
- passes a model output to any strategy given to it

So either it does handle `FeatureDistillation`, and the name should be changed (because intermediary features are not model outputs, so the makes the name confusing), or it doesn't, in which case you need another mechanism to enable the support of `FeatureDistillation` by the `Distiller`.

keras/src/distillation/strategies.py
```diff
@@ -0,0 +1,861 @@
...
+            `{0: 1.0, 1: 0.5}`
+
+    Example:
+
+    ```python
+    # Multi-output distillation for object detection
+    strategy = MultiOutputDistillation(
+        output_strategies={
+            0: LogitsDistillation(temperature=3.0, output_index=0),
+            1: LogitsDistillation(temperature=1.0, output_index=1)
```
hertschuh: The index in the dict is redundant with the `output_index` in the `LogitsDistillation` here.

Or if it's not redundant, then it's confusing.

hertschuh (commented):
Thank you for the PR!

First pass at the API.

keras/src/distillation/distiller.py
divyashreepathihalli: ah yes, we can set load_weight=False

keras/src/distillation/distiller.py
divyashreepathihalli: since ops.add only accepts two tensors, using ops.sum(with ops.stack) instead.

keras/src/distillation/strategies.py
divyashreepathihalli: updated the name and the file names

keras/src/distillation/strategies.py
divyashreepathihalli: Because it is a widely used way of distillation and this would reduce set up time for users. totally fine to remove this.

keras/src/distillation/strategies.py
divyashreepathihalli: changed this everywhere to use nested structures instead

keras/src/distillation/strategies.py
divyashreepathihalli: I think this is needed because the loss parameter by default would get serialized as a complex object. This makes sure the serialize/deserialize workflow is good.

keras/src/distillation/strategies.py
divyashreepathihalli: fixed this

keras/src/distillation/strategies.py
divyashreepathihalli: removed this

keras/src/distillation/strategies.py
divyashreepathihalli: removed this

keras/src/distillation/strategies.py
divyashreepathihalli: updated based on the suggestion

keras/src/distillation/strategies.py
divyashreepathihalli: I have subclassed from Feature distillation now

keras/src/distillation/strategies.py
divyashreepathihalli: subclassed from feature distillation

keras/src/distillation/strategies.py
divyashreepathihalli: done

keras/src/distillation/distiller.py
abheesht17: Instead of a randomly initialised model, we can probably pick an instruction-tuned model (as the teacher model), and a base (not instruction-tuned) model as the student model

keras/src/distillation/strategies.py
divyashreepathihalli: done

keras/src/distillation/strategies.py
divyashreepathihalli: fixed this

keras/src/distillation/strategies.py
divyashreepathihalli: I still kept the arg name as strategies though because a lot of "loss" arguments can get confusing

keras/src/distillation/distillation_loss.py
```diff
@@ -0,0 +1,469 @@
...
+        Raises:
+            ValueError: If outputs are not compatible.
+        """
+        # Default implementation - can be overridden by subclasses
+        if not isinstance(teacher_outputs, (list, tuple)):
+            teacher_outputs = [teacher_outputs]
+        if not isinstance(student_outputs, (list, tuple)):
+            student_outputs = [student_outputs]
+
+        if len(teacher_outputs) != len(student_outputs):
```
hertschuh: Oh, we agreed to support any structure for outputs, including dicts and nested outputs.

So remove lines 50-53 and replace lines 55-61 with `keras.tree.assert_same_structures(teacher_outputs, student_outputs)`. It's actually a lot simpler.

keras/src/distillation/distillation_loss.py
```diff
@@ -0,0 +1,469 @@
...
+    Feature distillation transfers knowledge from intermediate layers of the
+    teacher model to corresponding layers of the student model. This approach
+    helps the student learn better internal representations and often leads
+    to better performance compared to logits-only distillation.
+
+    Args:
+        loss: Loss function to use for feature distillation. Can be:
+            - String identifier (e.g., 'mse', 'cosine_similarity', 'mae')
+            - Keras loss instance
+            - List/tuple of losses for multi-output models
```
hertschuh: About "muti-output models". It's a bit confusing because this is not about the output of the model, it's about the intermediary feature(s) you're capturing. So rephrasing it to talk about the output of the layer, not the model.

keras/src/distillation/distillation_loss.py
```diff
@@ -0,0 +1,469 @@
...
+            - Dict of losses for named outputs
```
hertschuh: It's not really list vs. dict vs. tensor. It's really any nested structure.

keras/src/distillation/distillation_loss.py
```diff
@@ -0,0 +1,469 @@
...
+        # No specific layer, use the final model output
+        if layer_name is None:
+            return model(inputs, training=training)
+
+        # For intermediate layer extraction, create feature extractor if needed
+        if getattr(self, feature_model_attr) is None:
+            try:
+                setattr(
+                    self,
+                    feature_model_attr,
```
hertschuh: Why `setattr` instead of `self.feature_model_attr`? Is it to bypass tracking?

keras/src/distillation/distillation_loss.py
```diff
@@ -0,0 +1,469 @@
...
+    Args:
+        temperature: Temperature for softmax scaling. Higher values produce
+            softer probability distributions that are easier for the student to
+            learn. Typical values range from 3-5. Defaults to 3.0.
+        loss: Loss function to use for distillation. Can be:
+            - String identifier (e.g., 'kl_divergence',
+              'categorical_crossentropy')
+            - Keras loss instance
+            - List/tuple of losses for multi-output models
+            - Dict of losses for named outputs
```
hertschuh: Same comment about saying that any arbitrary nested structure is supported.

keras/src/distillation/distillation_loss.py
```diff
@@ -0,0 +1,469 @@
...
+    @classmethod
+    def from_config(cls, config):
+        """Create instance from configuration."""
+        config = config.copy()
+        config["loss"] = keras.losses.deserialize(config["loss"])
+
+        # Filter out parameters that LogitsDistillation doesn't accept
+        # (inherited from FeatureDistillation's get_config)
+        config.pop("teacher_layer_name", None)
+        config.pop("student_layer_name", None)
```
hertschuh: From what I could see, you don't use anything from `FeatureDistillation`. Should you make the super class `DistillationLoss` instead to be able to remove this?

keras/src/distillation/distillation_loss.py
```diff
@@ -0,0 +1,469 @@
...
+                list/tuple of tensors, or dict of tensors.
+            student_outputs: Logits from student model. Can be a single tensor,
+                list/tuple of tensors, or dict of tensors.
+            **kwargs: Additional arguments (ignored).
+        Returns:
+            Distillation loss tensor.
+        """
+        # Apply temperature scaling using tree.map_structure
+        teacher_scaled = tree.map_structure(
+            lambda x: x / self.temperature, teacher_outputs
```
hertschuh: nitpick: let's use `ops.divide` instead of `/`

keras/src/distillation/distillation_loss.py
```diff
@@ -0,0 +1,469 @@
...
+        )
+        student_scaled = tree.map_structure(
+            lambda x: x / self.temperature, student_outputs
```
hertschuh: nitpick: let's use `ops.divide` instead of `/`

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+        """Validate that teacher and student have compatible input shapes."""
+        # If symbolic tensors are not available (subclassed models), skip.
+        if not hasattr(teacher, "inputs") or not hasattr(student, "inputs"):
+            return
+        teacher_inputs = getattr(teacher, "inputs")
+        student_inputs = getattr(student, "inputs")
+        if teacher_inputs is None or student_inputs is None:
+            return
+
+        # Handle single input case
```
hertschuh: You want to support arbitrary inputs, not just lists.
Lines 226-255 can be replaced with:

```python
keras.tree.map_structures(teacher_inputs, student_inputs, lambda ti, si: self._assert_shapes_are_compatible(ti.shape, si.shape, "input"))
```

because it will also verify that the structures are the same.

You'll have to change `_shapes_are_compatible` to `_assert_shapes_are_compatible` and move the `raise` part in there. You can pass a specific error message component like `"input"` vs. `"output"`.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+        """Validate that teacher and student have compatible output shapes."""
+        # If symbolic tensors are not available (subclassed models), skip.
+        if not hasattr(teacher, "outputs") or not hasattr(student, "outputs"):
+            return
+        teacher_outputs = getattr(teacher, "outputs")
+        student_outputs = getattr(student, "outputs")
+        if teacher_outputs is None or student_outputs is None:
+            return
+
+        # Handle single output case
```
hertschuh: Same comment, line 267-295 should be a single line of code:

```python
keras.tree.map_structures(teacher_outputs, student_outputs, lambda to, so: self._assert_shapes_are_compatible(to.shape, so.shape, "output"))
```

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+        """Validate that teacher and student have compatible data types."""
+        # If symbolic tensors are not available (subclassed models), skip.
+        if not hasattr(teacher, "inputs") or not hasattr(student, "inputs"):
+            return
+        if teacher.inputs is None or student.inputs is None:
+            return
+        teacher_dtypes = [input.dtype for input in teacher.inputs]
+        student_dtypes = [input.dtype for input in student.inputs]
+
+        # Check input dtypes
```
hertschuh: Same here. Define a helper `def _assert_same_dtype(...)` and do a `keras.tree.map_structures`.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+                    f"Output {i} data types are incompatible. "
+                    f"Teacher output dtype: {teacher_dtype}, "
+                    f"Student output dtype: {student_dtype}. "
+                    f"Both models must use the same data type."
+                )
+
+    def _validate_strategy_compatibility(self, teacher, student, strategy):
+        """Validate that the strategy is compatible with the teacher and student
+        models."""
+        if hasattr(strategy, "validate_model_compatibility"):
```
hertschuh: I find it cleaner to just define 

```python
def validate_model_compatibility(self, teacher, student):
  pass
```

in `DistillationLoss` and avoid this test.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+            strategy.validate_model_compatibility(teacher, student)
+
+    def _shapes_are_compatible(self, shape1, shape2):
+        """Check if two shapes are compatible (allowing for batch dimension
+        flexibility)."""
+        # Convert to lists for easier handling
+        if hasattr(shape1, "as_list"):
```
hertschuh: I'm pretty sure shape are always iterable, so I don't think you need lines 343 to 355.

https://github.com/keras-team/keras-rs/blob/main/keras_rs/src/utils/keras_utils.py#L56

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+            shape2 = list(shape2)
+        else:
+            shape2 = [shape2]
+
+        # Check if they have the same number of dimensions
+        if len(shape1) != len(shape2):
+            return False
+
+        # Check all dimensions except the first (batch dimension)
+        for dim1, dim2 in zip(shape1[1:], shape2[1:]):
```
hertschuh: You're skipping the batch size, but I don't think you should in this case:
- you'll be feeding the same inputs to both models, so they have the same batch size and output batch size
- you'll be computing the difference between the output of the teacher model and the student model, so they'd better have the same batch size

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+        """
+        if self._teacher_feature_extractor is not None:
+            # Use efficient multi-output extractor
+            feature_outputs = self._teacher_feature_extractor(x, training=False)
+            if not isinstance(feature_outputs, (list, tuple)):
+                feature_outputs = [feature_outputs]
+
+            # Map outputs to layer names
+            features = {}
+            for name, output in zip(
```
hertschuh: You'd get this for free if you changed lines 421-427 to this:

```python
            # Collect outputs
            outputs = {"final_output": final_output}  # Always include final output

            for layer_name in layer_names:
                layer = model.get_layer(name=layer_name)
                outputs[layer_name](layer.output)
```

(which is also shorter than the current code)

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+        """
+        if self._student_feature_extractor is not None:
+            # Use efficient multi-output extractor
+            feature_outputs = self._student_feature_extractor(x, training=True)
+            if not isinstance(feature_outputs, (list, tuple)):
+                feature_outputs = [feature_outputs]
+
+            # Map outputs to layer names
+            features = {}
+            for name, output in zip(
```
hertschuh: Same comment about changing lines 421-427.

keras/src/distillation/distillation_loss.py
```diff
@@ -0,0 +1,469 @@
...
+            except ValueError as e:
+                raise ValueError(f"In teacher model: {e}")
+
+        if self.student_layer_name is not None:
+            try:
+                student.get_layer(name=self.student_layer_name)
+            except ValueError as e:
+                raise ValueError(f"In student model: {e}")
+
+    def _create_feature_extractor(self, model, layer_name):
```
hertschuh: Remove, this is never used, you have `_create_multi_feature_extractors` in the `Distiller`.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+            # Create extractor
+            extractor = keras.Model(
+                inputs=inputs,
+                outputs=outputs,
+                name=f"{model.name}_multi_feature_extractor",
+            )
+
+            # Store based on model type
+            if model_type == "teacher":
+                self._teacher_feature_extractor = extractor
```
hertschuh: It's cleaner to return `extractor` and let the caller assign it to the correct attribute.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+                self._teacher_output_names = output_names
+            else:
+                self._student_feature_extractor = extractor
```
hertschuh: It's cleaner to return `extractor` and let the caller assign it to the correct attribute.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+                self._student_output_names = output_names
+
+        except (ValueError, AttributeError):
+            # Fallback for subclassed models
+            if model_type == "teacher":
```
hertschuh: It's cleaner to return `None` and let the caller assign it to the correct attribute.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+            extractor = keras.Model(
+                inputs=inputs,
+                outputs=outputs,
+                name=f"{model.name}_multi_feature_extractor",
+            )
+
+            # Store based on model type
+            if model_type == "teacher":
+                self._teacher_feature_extractor = extractor
+                self._teacher_output_names = output_names
```
hertschuh: If you change lines 421-427 per my other comment, you don't need `self._teacher_output_names` at all.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+            else:
+                self._student_feature_extractor = extractor
+                self._student_output_names = output_names
```
hertschuh: If you change lines 421-427 per my other comment, you don't need `self. _student_output_names` at all.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+        # Compile with a dummy loss since we override compute_loss
+        super().compile(
+            optimizer=optimizer,
+            loss=None,  # We handle loss in compute_loss
+            metrics=metrics,
+            **kwargs,
+        )
+
+    @property
+    def student_model(self):
```
hertschuh: Why not let people use `self.student` and remove this (it can be documented as an attribute in the docstring)?

This is confusing because:
- the `__init__` arg is `student`, not `student_model`
- `self.student` and `self.student_model` return the same.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+        Returns:
+            Combined loss tensor.
+        """
+        # Handle case where y_pred is not provided
+        if y_pred is None:
+            y_pred = self(x, training=training)
+        # Compute student loss using tree operations for dicts, manual for lists
+        student_loss = 0.0
+        if self.student_loss_weight > 0.0 and y is not None:
+            if isinstance(self._student_loss, dict):
```
hertschuh: Replace lines 582-619 with:

```python
loss_values = keras.tree.map_structures(lambda l, o, o_pred: l(o, o_pred), self._student_loss, y, y_pred)
flat_losses = tree.flatten(loss_values)
student_loss = keras.ops.sum(keras.ops.stack(flat_losses)) if len(flat_losses) > 1 else flat_losses
```

`keras.tree.map_structures` will verify the 3 structures match.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+                    strategy_teacher_output = teacher_features["final_output"]
+                    strategy_student_output = y_pred
+
+                # Compute loss for this strategy
+                strategy_loss = strategy.compute_loss(
+                    strategy_teacher_output, strategy_student_output
+                )
+
+                # Apply weight and add to total
+                distillation_loss += weight * strategy_loss
```
hertschuh: nitpick: `ops.sum` and `ops.multiply` instead of `+=` and `*`.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+
+            # Ensure distillation_loss is a scalar
+            if (
+                hasattr(distillation_loss, "shape")
```
hertschuh: This should always be a tensor, I think it's safe to call `.shape` directly.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+                and len(distillation_loss.shape) > 0
+            ):
+                distillation_loss = keras.ops.mean(distillation_loss)
+
+        # Combine losses
+        total_loss = (
+            self.student_loss_weight * student_loss
```
hertschuh: nitpick: keras ops

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+        self.student_loss_tracker.reset_state()
+        self.distillation_loss_tracker.reset_state()
+        self.total_loss_tracker.reset_state()
+
+    @property
+    def metrics(self):
+        """Return list of metrics."""
+        # Get parent metrics (from compile)
+        parent_metrics = []
+        if hasattr(super(), "metrics"):
```
hertschuh: You're subclassing `Model`, `super().metrics` will always work.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+
+    def reset_metrics(self):
+        """Reset all metrics."""
+        super().reset_metrics()
+        self.student_loss_tracker.reset_state()
+        self.distillation_loss_tracker.reset_state()
+        self.total_loss_tracker.reset_state()
+
+    @property
+    def metrics(self):
```
hertschuh: Why do you override this? The auto-tracking should do its job and correctly return all your metrics.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+                    self.student
+                ),
+                "strategies": [
+                    serialization_lib.serialize_keras_object(strategy)
+                    for strategy in self.strategies
+                ],
+                "strategy_weights": self.strategy_weights,
+                "student_loss_weight": self.student_loss_weight,
+                # Save current state, not initial parameters
+                "optimizer": serialization_lib.serialize_keras_object(
```
hertschuh: Oh weird, the `Model` class doesn't do that?

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,788 @@
...
+                    self.optimizer
+                )
+                if hasattr(self, "optimizer") and self.optimizer
+                else None,
+                "student_loss": serialization_lib.serialize_keras_object(
+                    getattr(self, "_student_loss_for_serialization", None)
+                ),
+                # Note: metrics are not easily serializable due to
+                # CompileMetrics complexity, so we skip them in serialization
+                "metrics": None,
```
hertschuh: Hmm... that would be confusing. People will report this as a bug.

Taking a step back, what is the reason for passing them (and the loss and the optimizer) in `__init__` rather than letting people call `compile`?

hertschuh (commented):
Thanks Divya!

Also, `keras.tree` is your friend.

keras/src/distillation/distillation_loss.py
divyashreepathihalli: changed it to layer outputs

keras/src/distillation/distiller.py
divyashreepathihalli: modified validation to include batch dimension

keras/src/distillation/distiller.py
divyashreepathihalli: removed

keras/src/distillation/distiller.py
divyashreepathihalli: passing in init instead of compile

keras/src/distillation/distillation_loss.py
divyashreepathihalli: removed this

keras/src/distillation/distillation_loss.py
hertschuh: This wasn't changed.

keras/src/distillation/distillation_loss.py
```diff
@@ -0,0 +1,440 @@
...
```
hertschuh: This is never used, remove. `Distiller` has a more general (multi-layer) version of this.

keras/src/distillation/distillation_loss.py
```diff
@@ -0,0 +1,440 @@
...
+                tensor or a list/tuple of tensors for multi-output models.
+            student_outputs: Outputs from the student model. Can be a single
+                tensor or a list/tuple of tensors for multi-output models.
+            **kwargs: Additional arguments for custom strategies.
+        Returns:
+            Distillation loss tensor.
+        """
+        raise NotImplementedError("Subclasses must implement compute_loss")
+
+    def validate_outputs(self, teacher_outputs, student_outputs):
```
hertschuh, line 62: This is never called. You're missing some code to call it somewhere. Should it be called from the `Distiller`?

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,657 @@
...
+            teacher.outputs,
+            student.outputs,
+        )
+
+    def _validate_strategy_compatibility(self, teacher, student, strategy):
+        """Validate that the strategy is compatible with the teacher and student
+        models."""
+        strategy.validate_model_compatibility(teacher, student)
+
+    def _assert_shapes_are_compatible(self, shape1, shape2, context):
```
hertschuh: nitpick: move up before `_validate_input_compatibility` to follow the same ordering pattern as `_assert_same_dtype`?

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,657 @@
...
+                and strategy.student_layer_name
+            ):
+                if strategy.student_layer_name not in student_layer_names:
+                    student_layer_names.append(strategy.student_layer_name)
+
+        # Create multi-output feature extractors if needed
+        self._teacher_feature_extractor = (
+            self._create_feature_extractor(self.teacher, teacher_layer_names)
+            if teacher_layer_names
+            else None
```
hertschuh: nitpick: you could save a few lines here and line 322 by doing `if not teacher_layer_names: return None` in `_create_feature_extractor`.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,657 @@
...
+
+        Args:
+            x: Input data.
+
+        Returns:
+            Dict mapping layer names to their outputs.
+        """
+        if self._teacher_feature_extractor is not None:
+            # Use efficient multi-output extractor (returns dict directly)
+            return self._teacher_feature_extractor(x, training=False)
```
hertschuh, line 342: Hmm... so in the case when there are `LogitsDistillation` instances, that means that the student model is run twice. Once in `self.call` to get `y_pred` and once here to get the other logits.

keras/src/distillation/distillation_loss.py
divyashreepathihalli: done!

keras/src/distillation/distillation_loss.py
divyashreepathihalli: removed

keras/src/distillation/distillation_loss.py
divyashreepathihalli: validating this in compute loss now

keras/src/distillation/distiller.py
divyashreepathihalli: updated- the y_pred is now reused to avoid computation

hertschuh: What I meant is that `y_pred` is computed once in `self.call` and then another time when calling `_student_feature_extractor`. But let's not worry about that now.

keras/src/distillation/distillation_loss.py
```diff
@@ -0,0 +1,378 @@
...
+                raise ValueError(f"In student model: {e}")
+
+    def validate_outputs(self, teacher_outputs, student_outputs):
+        """Validate that outputs are compatible for feature distillation."""
+        super().validate_outputs(teacher_outputs, student_outputs)
+
+        # Validate that loss structure matches output structure
+        try:
+            tree.assert_same_structure(self.loss, teacher_outputs)
+            tree.assert_same_structure(self.loss, student_outputs)
```
hertschuh: nitpick: `super().valudate_outputs` already validates that `teacher_outputs` and `student_outputs` match in structure, so in theory you don't need both.

keras/src/distillation/distillation_loss.py
```diff
@@ -0,0 +1,378 @@
...
+        ):
+            # Validate that the specified layers exist and are compatible
+            self._validate_layer_compatibility(teacher_outputs, student_outputs)
+
+    def _validate_layer_compatibility(self, teacher_outputs, student_outputs):
+        """Validate that the specified layers are compatible for feature
+        distillation."""
+        # This method would be called by the distiller to validate layer
+        # compatibility when using feature distillation with specific layer
+        # names
```
hertschuh: But it's not called by the distiller, only by `validate_outputs`. The layer name is already validated in `validate_model_compatibility`.

It's also private (meaning users won't override), it `pass`es by default and is never overridden. Just remove?

keras/src/distillation/distillation_loss.py
```diff
@@ -0,0 +1,378 @@
...
+
+    def __init__(
+        self,
+        temperature=3.0,
+        loss="kl_divergence",
+    ):
+        self.temperature = temperature
+
+        # Convert loss structure to functions using tree.map_structure
+        def convert_loss_to_function(loss_item):
```
hertschuh: This function is duplicated. Please extract it and reuse.

keras/src/distillation/distillation_loss.py
```diff
@@ -0,0 +1,378 @@
...
+            loss = keras.ops.mean(loss_fn(teacher_features, student_features))
+
+            # Special handling for cosine similarity (convert similarity to
+            # distance)
+            if (
+                hasattr(loss_fn, "__name__")
+                and "cosine" in loss_fn.__name__.lower()
+            ):
+                # Convert similarity to distance: distance = 1 - similarity
+                loss = 1.0 - loss
```
hertschuh: nitpick: `ops.subtract(1, loss)`

keras/src/distillation/distillation_loss.py
```diff
@@ -0,0 +1,378 @@
...
+            student_outputs: Extracted features from the specified student
+                layer.
+            **kwargs: Additional arguments (ignored).
+        Returns:
+            Feature distillation loss tensor.
+        """
+
+        # Apply loss function(s) to corresponding features
+        def apply_loss(loss_fn, teacher_features, student_features):
+            loss = keras.ops.mean(loss_fn(teacher_features, student_features))
```
hertschuh, line 230: I think we should allow some of the losses to be `None`. If you have a model with multiple outputs, maybe you only care about the difference between one of the outputs. You could put a weight of 0 on the other losses, but users will probably try to use a `None` loss, which is also not rejected by `__init__`.

It's just a question of doing `if loss_fn is None: return 0` but we should probably verify in `__init__` that at least one loss is not `None`.

keras/src/distillation/distillation_loss.py
```diff
@@ -0,0 +1,378 @@
...
+            lambda x: keras.ops.divide(x, self.temperature), teacher_outputs
+        )
+        student_scaled = tree.map_structure(
+            lambda x: keras.ops.divide(x, self.temperature), student_outputs
+        )
+
+        # Apply loss function(s) to corresponding outputs
+        def apply_loss(loss_fn, teacher_logits, student_logits):
+            # Special handling for KL divergence (needs probabilities)
+            if (
```
hertschuh: I think we should allow some of the losses to be `None`. If you have a model with multiple outputs, maybe you only care about the difference between one of the outputs. You could put a weight of 0 on the other losses, but users will probably try to use a `None` loss, which is also not rejected by `__init__`.

It's just a question of doing `if loss_fn is None: return 0` but we should probably verify in `__init__` that at least one loss is not `None`.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,661 @@
...
+
+            # Create and return extractor
+            return keras.Model(
+                inputs=inputs,
+                outputs=outputs,
+                name=f"{model.name}_multi_feature_extractor",
+            )
+        except (ValueError, AttributeError):
+            # Fallback for subclassed models
+            return None
```
hertschuh: I think you should throw a meaningful error here.

If I read this correctly, when you try to use `LogitsDistillation` strategies on a subclass model (and all of KerasHub models are subclass models), this will return `None`, and then later, you'll get the error line 422 `f"Layer '{layer_name}' features not found in extracted "`, which will be confusing to users.

Instead you should say here (or somewhere else) that `LogitsDistillation` is not compatible with subclass models.

One alternative would be to check that in `validate_model_compatibility`. Right now, you validate that the layer name is found, but you could also validate that it's not a subclass model.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,661 @@
...
+                        )
+                    except ValueError as e:
+                        # Provide more helpful error message for feature
+                        # extraction failures
+                        raise RuntimeError(
+                            f"FeatureDistillation failed for strategy "
+                            f"targeting teacher layer "
+                            f"'{strategy.teacher_layer_name}' and student "
+                            f"layer '{strategy.student_layer_name}'. This can "
+                            f"happen with subclassed models that haven't "
```
hertschuh: Doesn't it happen with all subclass models always? Even when they're built? (based on line 369)

Based on my comment line 369, can't we catch this error earlier rather than try to guess that any `ValueError` that's happening here has this as the cause?

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,661 @@
...
+                    strategy_teacher_output, strategy_student_output
+                )
+
+                # Apply weight and add to total
+                distillation_loss = keras.ops.add(
+                    distillation_loss, keras.ops.multiply(weight, strategy_loss)
+                )
+
+            # Ensure distillation_loss is a scalar
+            if len(distillation_loss.shape) > 0:
```
hertschuh: Should this be an error (checking `strategy_loss` within the for loop)? Aren't strategies supposed to output scalars?

keras/src/distillation/distiller.py
divyashreepathihalli: KerasHub backbones are all written in functional model style. Logits distillation would work for subclass models but would not for feature distillation because model.inputs, model.outputs attribute would not be accessible.

Added checks in distiller and removed silent fallback

keras/src/distillation/distiller.py
divyashreepathihalli: raising an error inside the loop if strategy returns a non scalar

keras/src/distillation/distiller.py
hertschuh: > KerasHub backbones are all written in functional model style

But they are subclass models that under the hood use a Functional model. Does feature distillation work in this case?

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,578 @@
...
+                    y_pred,
+                )
+                flat_losses = tree.flatten(loss_values)
+                student_loss = (
+                    keras.ops.sum(keras.ops.stack(flat_losses))
+                    if len(flat_losses) > 1
+                    else flat_losses[0]
+                )
+            except (ValueError, TypeError):
+                # Fallback for TrackedDict compatibility issues
```
hertschuh: I think we should fix this in a different way. For one, this is a lot of code. And second, this doesn't handle nested structures.

What's the error?

You might just need to add `@tracking.no_automatic_dependency_tracking` on the `__init__` of `FeatureDistillation` and `LogitsDistillation`. This will prevent the tracker from turning `dict`s to `TrackingDict`s, which doesn't matter since losses are not tracked anyway.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,578 @@
...
+                            strategy, teacher_features, is_teacher=True
+                        )
+                        strategy_student_output = self._get_strategy_features(
+                            strategy, student_features, is_teacher=False
+                        )
+                    except ValueError as e:
+                        # Re-raise with context about which strategy failed
+                        raise RuntimeError(
+                            f"Failed to extract features for "
+                            f"FeatureDistillation targeting teacher layer "
```
hertschuh: nitpick `{type(strategy)}` instead of `FeatureDistillation` in case people subclass it.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,578 @@
...
+                    ]
+                    flat_losses = tree.flatten(loss_values)
+                    student_loss = keras.ops.sum(keras.ops.stack(flat_losses))
+                else:
+                    # Single output case
+                    student_loss = self._student_loss(y, y_pred)
+
+            # Ensure student_loss is a scalar
+            if hasattr(student_loss, "shape") and len(student_loss.shape) > 0:
+                student_loss = keras.ops.mean(student_loss)
```
hertschuh, line 436: The strategy loss raises an error line 504 if it's not scalar. Any reason to handle these differently?

keras/src/distillation/distillation_loss_test.py
```diff
@@ -0,0 +1,291 @@
...
+        # Create dummy logits
+        teacher_logits = keras.ops.convert_to_tensor(
+            np.array([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]]), dtype="float32"
+        )
+        student_logits = keras.ops.convert_to_tensor(
+            np.array([[2.0, 1.0, 4.0], [3.0, 6.0, 2.0]]), dtype="float32"
+        )
+
+        # Verify that teacher and student outputs have the same structure
+        keras.tree.assert_same_structure(teacher_logits, student_logits)
```
hertschuh: This test is not using `LogitsDistillation` at all. You should instantiate a `LogitsDistillation` and call the validation method instead.

keras/src/distillation/distillation_loss_test.py
```diff
@@ -0,0 +1,291 @@
...
+        # Create dummy features
+        teacher_features = keras.ops.convert_to_tensor(
+            np.array([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]]), dtype="float32"
+        )
+        student_features = keras.ops.convert_to_tensor(
+            np.array([[1.1, 2.1, 3.1], [4.1, 5.1, 6.1]]), dtype="float32"
+        )
+
+        # Verify that teacher and student outputs have the same structure
+        keras.tree.assert_same_structure(teacher_features, student_features)
```
hertschuh: This test is not using `FeatureDistillation` at all. You should instantiate a `FeatureDistillation` and call the validation method instead.

keras/src/distillation/distillation_loss_test.py
```diff
@@ -0,0 +1,291 @@
...
+
+
+@pytest.mark.requires_trainable_backend
+class TestEndToEndDistillation(TestCase):
+    """End-to-end distillation tests with real models."""
+
+    def test_logits_distillation_end_to_end(self):
+        """Test end-to-end logits distillation with real models."""
+        # Create teacher model (larger)
+        teacher = keras.Sequential(
```
hertschuh: The code to create the models is a few lines and is replicated a few times. Factor out to shorten tests.

keras/src/distillation/distiller.py
divyashreepathihalli: this is much simpler. I added the decorator @tracking.no_automatic_dependency_tracking and simplified the compute loss method

keras/src/distillation/distiller.py
divyashreepathihalli: The student loss fn returns per-sample losses by default (shape, (batch_size,)) - the code converts to scalars with keras.ops.mean

But with stategy.compute_loss() method, it is supposed to return scalar losses.

keras/src/distillation/distillation_loss_test.py
divyashreepathihalli: added

keras/src/distillation/distillation_loss_test.py
divyashreepathihalli: added

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,555 @@
...
+        **kwargs: Additional keyword arguments passed to the parent `Model`
+            class.
+
+    Attributes:
+        student: The student model being trained. Access this to get the trained
+            student model for independent use after distillation training.
+        teacher: The teacher model providing knowledge. This model is frozen
+            during training.
+
+    Examples:
```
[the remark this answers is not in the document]
hertschuh: Hmm... I wonder why Gemini got it wrong.

keras/src/distillation/distiller_test.py
```diff
@@ -0,0 +1,526 @@
...
+
+        # Verify strategy parameters
+        self.assertEqual(reconstructed_distiller.strategies[0].temperature, 3.0)
+
+        # Test that reconstructed distiller can be used for inference
+        reconstructed_output = reconstructed_distiller(x_test)
+        self.assertEqual(reconstructed_output.shape, (2, 10))
+
+        # Test model saving and loading (full integration test)
+        with tempfile.TemporaryDirectory() as temp_dir:
```
hertschuh: nitpick: You don't need a `with` and indented code if you just use `self.get_temp_dir()`, it cleans up automaticatlly.

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,555 @@
...
+    (teacher) to a smaller, simpler model (student). The student learns
+    from both ground truth labels and the teacher's predictions, often
+    achieving better performance than training on labels alone.
+
+    Arguments:
+        teacher: A trained `keras.Model` that serves as the knowledge source.
+            The teacher model is frozen during distillation.
+        student: A `keras.Model` to be trained through distillation.
+        strategies: List of distillation strategies to apply. Can be a single
+            strategy or a list of strategies like `LogitsDistillation`,
```
fchollet: "strategy instances like `keras.distillation.LogitsDistillation()`..."

keras/src/distillation/distiller.py
```diff
@@ -0,0 +1,555 @@
...
+            `FeatureDistillation`, or custom distillation strategies.
+        strategy_weights: List of weights for each distillation strategy. Must
+            have the same length as `strategies`. If None, equal weights used.
```
fchollet: Use backticks round None

keras/src/distillation/distiller.py
fchollet: There is an inconsistency in terminology, where these are sometimes referred to as losses and sometimes as strategies. Which is it? (Could be both: training_strategy or something)

We should make the arg name here consistent with the object class name.

fchollet: If these are DistillationLoss subclasses then the arg should be `distillation_losses`

A potential issue with calling them losses is that they're very different from `keras.losses`.

divyashreepathihalli: done! renamed everything to distillation loss

keras/src/distillation/distillation_loss.py
```diff
@@ -53,7 +53,7 @@ def compute_loss(self, teacher_outputs, student_outputs, **kwargs):
                 tensor or a list/tuple of tensors for multi-output models.
             student_outputs: Outputs from the student model. Can be a single
                 tensor or a list/tuple of tensors for multi-output models.
-            **kwargs: Additional arguments for custom strategies.
+            **kwargs: Additional arguments for custom distillation_loss.
```
hertschuh, line 56: distillation losses.

keras/src/distillation/distillation_loss.py
```diff
@@ -77,14 +77,15 @@ def validate_model_compatibility(self, teacher, student):
             teacher: The teacher model.
             student: The student model.
         Raises:
-            ValueError: If models are not compatible with this strategy.
+            ValueError: If models are not compatible with this
+                distillation_loss.
```
hertschuh, line 81: distillation loss.

keras/src/distillation/distillation_loss.py
```diff
@@ -77,14 +77,15 @@ def validate_model_compatibility(self, teacher, student):
...
         """
         pass
 
 
 @keras_export("keras.distillation.FeatureDistillation")
 class FeatureDistillation(DistillationLoss):
-    """Feature distillation strategy using intermediate layer representations.
+    """Feature distillation distillation_loss.
```
hertschuh, line 88: Feature distillation loss.

keras/src/distillation/distillation_loss.py
```diff
@@ -255,11 +256,11 @@ def from_config(cls, config):
 
 @keras_export("keras.distillation.LogitsDistillation")
 class LogitsDistillation(DistillationLoss):
-    """Distillation strategy that transfers knowledge from final model outputs.
+    """Distillation loss that transfers knowledge from final model outputs.
 
-    This strategy applies temperature scaling to the teacher's logits before
-    computing the loss between teacher and student predictions. It's the most
-    common approach for knowledge distillation.
+    This distillation_loss applies temperature scaling to the teacher's logits
```
hertschuh, line 261: This distillation loss...

keras/src/distillation/distillation_loss_test.py
```diff
@@ -10,7 +10,7 @@
 
 @pytest.mark.requires_trainable_backend
 class TestLogitsDistillation(TestCase):
-    """Test cases for LogitsDistillation strategy."""
+    """Test cases for LogitsDistillation distillation_loss."""
```
hertschuh, line 13: Remove "distillation_loss" (actually, you can remove the whole line, it's not needed).

keras/src/distillation/distillation_loss_test.py
```diff
@@ -22,16 +22,18 @@ def test_logits_distillation_basic(self):
...
-            strategy.validate_outputs(teacher_logits, incompatible_logits)
+            distillation_loss.validate_outputs(
+                teacher_logits, incompatible_logits
+            )
 
 
 @pytest.mark.requires_trainable_backend
 class TestFeatureDistillation(TestCase):
-    """Test cases for FeatureDistillation strategy."""
+    """Test cases for FeatureDistillation distillation_loss."""
```
hertschuh, line 36: Remove this line.

keras/src/distillation/distillation_loss_test.py
```diff
@@ -169,10 +173,10 @@ def test_feature_distillation_end_to_end(self):
             len(distiller._student_feature_extractor.outputs), 2
         )  # final + dense_1
 
-    def test_multi_strategy_distillation_end_to_end(self):
-        """Test end-to-end distillation with multiple strategies."""
-        # Create multiple strategies
-        strategies = [
+    def test_multi_distillation_loss_distillation_end_to_end(self):
+        """Test end-to-end distillation with multiple distillation_loss."""
+        # Create multiple distillation_loss
```
hertschuh, line 178: Fix or remove the docstring and comment.

divyashreepathihalli:
I have addressed all teh comments. Updated the name to be consistent - distillation loss everywhere instead of strategy. I will merge the PR

keras/src/distillation/distiller.py
```diff
@@ -94,8 +96,8 @@ def __init__(
         self,
         teacher,
         student,
-        strategies,
-        strategy_weights=None,
+        distillation_loss,
```
hertschuh, line 99: Do you want to call this `distallation_losses` instead? To make it clear it's a list.

keras/src/distillation/distiller.py
```diff
@@ -122,33 +124,35 @@ def __init__(
             )
         self.student_loss_weight = student_loss_weight
 
-        # Handle strategies configuration
-        if strategies is None:
+        # Handle distillation_loss configuration
```
hertschuh, line 127: Fix `distallation_loss` or remove

keras/src/distillation/distiller.py
```diff
@@ -122,33 +124,35 @@ def __init__(
...
+                        f"({len(distillation_loss_weights)}) must match "
+                        "number of distillation_loss "
+                        f"({len(distillation_loss)})"
                     )
-                self.strategy_weights = strategy_weights
+                self.distillation_loss_weights = distillation_loss_weights
 
         # Validate strategy-specific compatibility and create feature extractors
-        for strategy in self.strategies:
+        for strategy in self.distillation_loss:
```
hertschuh, line 155: `for distillation_loss in self.distillation_losses:`

keras/src/distillation/distiller.py
```diff
@@ -272,7 +276,7 @@ def _create_multi_feature_extractors(self):
         teacher_layer_names = []
         student_layer_names = []
 
-        for strategy in self.strategies:
+        for strategy in self.distillation_loss:
```
hertschuh, line 279: `for distillation_loss in self.distillation_losses:`

keras/src/distillation/distiller.py
```diff
@@ -296,15 +300,16 @@ def _create_multi_feature_extractors(self):
...
+                    layer_names: List of layer names to extract features from.
 
-        Raises:
-            ValueError: If model has no symbolic inputs/outputs.
+                Returns:
+                    Feature extractor model or `None` if no layer names
+                        sprovided.
+        `
+                Raises:
+                    ValueError: If model has no symbolic inputs/outputs.
```
hertschuh, line 312: Indentation seems off.

keras/src/distillation/distiller.py
```diff
@@ -437,7 +442,9 @@ def compute_loss(
             student_features = self._extract_all_student_features(x, y_pred)
 
             # Apply strategies using pre-extracted features
-            for strategy, weight in zip(self.strategies, self.strategy_weights):
+            for strategy, weight in zip(
+                self.distillation_loss, self.distillation_loss_weights
+            ):
```
hertschuh, line 447: `for distillation_loss, weight in zip(self.distillation_losses...`

keras/src/distillation/distiller.py
```diff
@@ -525,11 +532,11 @@ def get_config(self):
                 "student": serialization_lib.serialize_keras_object(
                     self.student
                 ),
-                "strategies": [
+                "distillation_loss": [
```
hertschuh, line 535: distillation_losses?

keras/src/distillation/distiller.py
```diff
@@ -547,9 +554,9 @@ def from_config(cls, config):
         config["student"] = serialization_lib.deserialize_keras_object(
             config["student"]
         )
-        config["strategies"] = [
+        config["distillation_loss"] = [
```
hertschuh, line 557: distillation_losses?

hertschuh (changes requested):
The bulk replace did some weird stuff, in particular in the docstrings.

But also, the main question is whether the Distiller argument should be called `distillation_losses` (plural.).

PR #21575, opened 2025-08-12
Disable `torch.load` in `TorchModuleWrapper` when in safe mode.

hertschuh:
Raise an exception and explain the user about the risks.


fchollet (approved):
LGTM, thank you!

PR #21577, opened 2025-08-13
Add `ops.image.scale_and_translate`.

james77777778:
This op is useful if we want to perform the same functionality as `torch.nn.functional.interpolate` with `align_corners=True`.

Some depth estimation models require this functionality.

Here is an example:

```python
import numpy as np
import torch

from keras.src import ops

torch_y = torch.nn.functional.interpolate(
    torch.from_numpy(x).unsqueeze(0).unsqueeze(0),
    size=(5, 5),
    mode="bilinear",
    align_corners=True,
)
torch_y = torch_y.detach().cpu().numpy()[0, 0]


def keras_interpolate(images, size, interpolation, antialias=False):
    shape = tuple([1] + list(size) + [1])
    spatial_dims = (1, 2)
    scale = ops.convert_to_tensor(
        [(shape[i] - 1.0) / (images.shape[i] - 1.0) for i in spatial_dims]
    )
    translation = -(scale / 2.0 - 0.5)
    return ops.image.scale_and_translate(
        images,
        shape,
        spatial_dims,
        scale,
        translation,
        interpolation,
        antialias,
    )


keras_y = keras_interpolate(
    ops.expand_dims(ops.convert_to_tensor(x, dtype="float32"), (0, -1)),
    size=(5, 5),
    interpolation="bilinear",
    antialias=False,
)
keras_y = ops.convert_to_numpy(keras_y)[0, ..., 0]
np.testing.assert_allclose(torch_y, keras_y, rtol=1e-5, atol=1e-5)
```

The benchmark:
- `(1, 256, 256, 64)` to `(1, 128, 128, 64)`
- Jitted / Compiled

|Backend|Time|
|-|-|
|jax|0.0114s|
|tensorflow|0.0668s|
|torch|0.0274s|

<details>

<summary> benchmark.py </summary>

```python
import time

import numpy as np

from keras.src import backend
from keras.src import ops

images = np.random.uniform(size=(2, 256, 256, 64)).astype("float32")
shape = (2, 128, 128, 64)
spatial_dims = (1, 2)
scale = np.array(
    [(shape[i] - 1.0) / (images.shape[i] - 1.0) for i in spatial_dims],
    dtype="float32",
)
translation = -(scale / 2.0 - 0.5)
method = "linear"

    import jax

    jit_scale_and_translate = jax.jit(
        ops.image.scale_and_translate, static_argnums=(1, 2, 5)
    )

elif backend.backend() == "tensorflow":
    import tensorflow as tf

    @tf.function(jit_compile=True)
    def jit_scale_and_translate(*args, **kwargs):
        return ops.image.scale_and_translate(*args, **kwargs)

    import torch

    @torch.compile
    def jit_scale_and_translate(*args, **kwargs):
        return ops.image.scale_and_translate(*args, **kwargs)
else:

    def jit_scale_and_translate(*args, **kwargs):
        return ops.image.scale_and_translate(*args, **kwargs)


# Warmup
for _ in range(3):
    jit_scale_and_translate(
        images, shape, spatial_dims, scale, translation, method
    )
print("Warmup done.")

st = time.time()
for i in range(5):
    resized_images = jit_scale_and_translate(
        images, shape, spatial_dims, scale, translation, method
    )
ed = time.time()

print(resized_images.shape)
print(f"Time taken: {ed - st:.4f} seconds")

```

</details>

### DIFF
--- a/keras/src/backend/jax/image.py
+++ b/keras/src/backend/jax/image.py
@@ -14,6 +14,34 @@ RESIZE_INTERPOLATIONS = (
     "lanczos5",
     "bicubic",
 )
+AFFINE_TRANSFORM_INTERPOLATIONS = {  # map to order
+    "nearest": 0,
+    "bilinear": 1,
+}
+AFFINE_TRANSFORM_FILL_MODES = {
+    "constant",
+    "nearest",
+    "wrap",
+    "mirror",
+    "reflect",
+}
+MAP_COORDINATES_FILL_MODES = {
+    "constant",
+    "nearest",
+    "wrap",
+    "mirror",
+    "reflect",
+}
+SCALE_AND_TRANSLATE_METHODS = {
+    "linear",
+    "bilinear",
+    "trilinear",
+    "cubic",
+    "bicubic",
+    "tricubic",
+    "lanczos3",
+    "lanczos5",
+}
 
 
 def rgb_to_grayscale(images, data_format=None):
@@ -372,19 +400,6 @@ def resize(
     )
 
 
-AFFINE_TRANSFORM_INTERPOLATIONS = {  # map to order
-    "nearest": 0,
-    "bilinear": 1,
-}
-AFFINE_TRANSFORM_FILL_MODES = {
-    "constant",
-    "nearest",
-    "wrap",
-    "mirror",
-    "reflect",
-}
-
-
 def affine_transform(
     images,
     transform,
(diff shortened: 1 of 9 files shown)


keras/src/backend/jax/image.py
```diff
@@ -14,6 +14,34 @@
     "lanczos5",
     "bicubic",
 )
+AFFINE_TRANSFORM_INTERPOLATIONS = {  # map to order
```
fchollet, line 17: Nice, this is clearer.

fchollet (approved):

PR #21579, opened 2025-08-13
Automate release process and uploading to PyPI

rtg0795:
Automate release process and uploading to PyPI through github actions.


fchollet:
Thanks for the PR. Releases right now are inevitably a manual process on my end since they require manually writing release notes (autogenerated notes can never be very informative). Meanwhile the actual upload process only takes a single command line. So overall it would not be very helpful to upload to pypi automatically when a new tag gets pushed. It could also be a potential security issue, since a number of credentials have write permissions to push tags.

PR #21580, opened 2025-08-14
[OpenVINO backend] fix ops.slice

Mohamed-Ashraf273:
@fchollet 

I discovered that the OpenVINO backend’s slice operation returns a fully dynamic shape after slicing, which causes all dimensions to become None and leads to downstream issues.
```python
from openvino import Type
from openvino import opset14 as ov_opset
from keras.src.backend.openvino.core import convert_to_tensor, slice


tensor = convert_to_tensor(ov_opset.parameter(shape=[1, 2, 3], dtype=Type.i32).output(0))
number = convert_to_tensor(ov_opset.parameter(shape=[], dtype=Type.i32).output(0))
print("Original tensor shape:", tensor.shape)
sliced_tensor = slice(tensor, start_indices=[0, 0, number], shape=[1, 2, 2])
print("Sliced tensor shape:", sliced_tensor.shape)
```
Before fix:
```
Original tensor shape: (1, 2, 3)
Sliced tensor shape: (None, None, None)
```
After fix:
```
Original tensor shape: (1, 2, 3)
Sliced tensor shape: (1, 2, 2)
```
This fix ensures that the sliced tensor retains as much static shape information as possible instead of defaulting to fully dynamic.

### DIFF
--- a/keras/src/backend/openvino/core.py
+++ b/keras/src/backend/openvino/core.py
@@ -843,9 +843,16 @@ def slice(inputs, start_indices, shape):
     start = ov_opset.concat(start, axis=0).output(0)
     stop = ov_opset.concat(stop, axis=0).output(0)
     axes = ov_opset.constant(axes, Type.i32).output(0)
-    return OpenVINOKerasTensor(
-        ov_opset.slice(inputs, start, stop, step, axes).output(0)
-    )
+    result = ov_opset.slice(inputs, start, stop, step, axes).output(0)
+
+    # Apply reshape to ensure output matches expected shape
+    # Convert None (dynamic) dimensions to -1 for OpenVINO compatibility
+    if all(dim is None or (isinstance(dim, int) and dim >= 0) for dim in shape):
+        reshape_pattern = [(-1 if dim is None else dim) for dim in shape]
+        target_shape = ov_opset.constant(reshape_pattern, Type.i32).output(0)
+        result = ov_opset.reshape(result, target_shape, False).output(0)
+
+    return OpenVINOKerasTensor(result)
 
 
 def slice_update(inputs, start_indices, updates):


Mohamed-Ashraf273:
> Thanks for the PR -- can you please add a simple test for this? Otherwise if we break it in the future we won't know.

@fchollet 
Done! 
Added

Mohamed-Ashraf273:
@fchollet

Mohamed-Ashraf273:
@fchollet

Mohamed-Ashraf273:
@hertschuh

keras/src/ops/core_test.py
```diff
@@ -1139,6 +1139,24 @@ def test_slice(self):
             core.Slice(shape)(inputs, start_indices), np.array([[5, 6], [8, 9]])
         )
 
+    @pytest.mark.skipif(
+        backend.backend() != "openvino",
+        reason="OpenVINO-specific test for slice operation shape preservation",
+    )
+    def test_slice_with_dynamic_indices_shape_preservation(self):
```
hertschuh: This should be a backend independent test as this applies to any backend.

There is already a `test_dynamic_slice` test: https://github.com/keras-team/keras/blob/master/keras/src/ops/core_test.py#L637

Can you just add a shape check there?

keras/src/ops/core_test.py
```diff
@@ -1139,6 +1139,24 @@ def test_slice(self):
...
+
+        tensor_param = ov_opset.parameter(shape=[1, 2, 3], dtype=Type.i32)
+        tensor = ops.convert_to_tensor(tensor_param.output(0))
+        dynamic_index = ov_opset.parameter(shape=[], dtype=Type.i32)
+        dynamic_start = ops.convert_to_tensor(dynamic_index.output(0))
+        start_indices = [0, 0, dynamic_start]
+        shape = (1, 2, 2)
+        sliced_tensor = core.slice(tensor, start_indices, shape)
+        self.assertEqual(sliced_tensor.shape, (1, 2, 2))
+        self.assertNotEqual(sliced_tensor.shape, (None, None, None))
```
hertschuh: `self.assertEqual(sliced_tensor.shape, (1, 2, 2))` already implies `self.assertNotEqual(sliced_tensor.shape, (None, None, None))`, you don't need the second one.

keras/src/ops/core_test.py
Mohamed-Ashraf273: Yeah, this is a better approach!
I changed it.
Thanks

Mohamed-Ashraf273:
@hertschuh 
Done!
I changed the test, and the PR is ready!

keras/src/ops/core_test.py
Mohamed-Ashraf273: Done!

PR #21582, opened 2025-08-14
Implement isposinf function in keras.ops

shashaka:
Adds keras.ops.isposinf, which checks whether each element in the input tensor is positive infinity.


fchollet (approved):
LGTM, thanks for the contribution.

Issue #21483, opened 2025-07-15
Attention layer issue broadcasting mask


knife982000:
# Summary

When using the `keras.layers.Attention` layer with `mask` support and `KERAS_BACKEND=torch`, a shape mismatch occurs between `scores` and `scores_mask`, leading to a broadcasting error during training.

## Code to Reproduce

```python
import os
os.environ["KERAS_BACKEND"] = "torch"


import numpy as np
from keras.layers import Attention, Add, GlobalAveragePooling1D, Dropout, Lambda, Embedding, Input, Dense
from keras.models import Model
from keras.utils import plot_model
from keras import ops

WORDS = 1000
MAXLEN = 50

#simulates a corpus
x_train = np.random.randint(low=1, high=WORDS+1, size=(20000,MAXLEN))
x_test = np.random.randint(low=1, high=WORDS+1, size=(20000,MAXLEN))
y_train = np.random.randint(low=0, high=2, size=(20000, 1))
y_test = np.random.randint(low=0, high=2, size=(20000, 1))

#simulates paddinf
x_train[:, np.random.randint(low=30, high=MAXLEN, size=(20000,))] = 0

i = Input((MAXLEN,))
e = Embedding(WORDS + 1, 60, mask_zero=True, name='base_emb')(i)

len_s = Lambda(lambda x: ops.expand_dims(ops.arange(start=0, stop=MAXLEN, step=1), axis=0), output_shape=(MAXLEN,))(i)
pos_e = Embedding(MAXLEN, 60, mask_zero=True, name='base_pos')(len_s)

sum = Add()([e, pos_e])
dq = Dense(60)(sum)
dk = Dense(60)(sum)

mask = Lambda(lambda x: x != 0, name='mask')(i)

att = Attention(name='att')([dq, dk], mask=[mask, mask])
attd = Dropout(0.1)(att)

d = GlobalAveragePooling1D()(attd)
d = Dropout(0.1)(d)
d = Dense(100)(d)
d = Dropout(0.1)(d)
d = Dense(1, activation='sigmoid')(d)


model = Model(i, d)
model.summary()
model.compile(loss='binary_crossentropy', optimizer='nadam', metrics=['accuracy'])

model.fit(x_train, y_train, epochs=3, validation_data=(x_test, y_test))
```

# Error Output

```text
RuntimeError: Unable to automatically build the model...
Exception encountered when calling Attention.call().

The size of tensor a (50) must match the size of tensor b (32) at non-singleton dimension 1

[...]

PR #21595, opened 2025-08-18
Fix shape mismatch in Keras Attention layer during masking

sonali-kumari1:
This PR fixes the `RuntimeError` encountered in the Attention layer when using masks. The error arises from a shape mismatch between the mask and attention scores, resulting in a broadcasting issue during training.
Fixes: [#21483 ](https://github.com/keras-team/keras/issues/21483)

### DIFF
--- a/keras/src/layers/attention/attention_test.py
+++ b/keras/src/layers/attention/attention_test.py
@@ -88,7 +88,7 @@ class AttentionTest(testing.TestCase):
 
     def test_attention_2D_mask_shape_mismatch(self):
         layer = layers.Attention()
-        batch_size, Tq, Tv, dim = 2, 3, 4, 4
+        batch_size, Tq, Tv, dim = 2, 3, 4, 5
         query = np.random.random((batch_size, Tq, dim)).astype(np.float32)
         value = np.random.random((batch_size, Tv, dim)).astype(np.float32)
         query_mask = np.array([[True, False, True], [True, False, True]])


sonali-kumari1:
Hi @fchollet @hertschuh - 
I have added a unit test  to verify shape mismatch with 2D mask. Thanks!

keras/src/layers/attention/attention_test.py
```diff
@@ -86,6 +86,21 @@ def test_attention_with_mask(self):
         self.assertAllClose(output, [[[1.0, 1.0], [0.0, 0.0]]])
         self.assertAllClose(scores, [[[1.0, 0.0], [1.0, 0.0]]])
 
+    def test_attention_2D_mask_shape_mismatch(self):
+        layer = layers.Attention()
+        batch_size, Tq, Tv, dim = 2, 3, 3, 4
```
hertschuh: Please use different values for `Tq` and `Tv` so that the tests check that there is no mixup.

keras/src/layers/attention/attention_test.py
```diff
@@ -88,11 +88,13 @@ def test_attention_with_mask(self):
 
     def test_attention_2D_mask_shape_mismatch(self):
         layer = layers.Attention()
-        batch_size, Tq, Tv, dim = 2, 3, 3, 4
+        batch_size, Tq, Tv, dim = 2, 3, 4, 4
```
hertschuh: Sorry to be bother you again about this, but can you make all values unique? Like use `5` for `dim`.

Again, we want to make sure dimensions are matched up correctly in unit tests.

sonali-kumari1: @hertschuh, I updated the unit test to use distinct values for `batch_size`, `Tq`, `Tv` and `dim` to properly detect shape mismatches. Thank you!

hertschuh (approved):
Thank you for the fix!

PR #21598, opened 2025-08-19
Refactor `TFDataLayer` to be more generic for Grain.

james77777778:
This PR:
1. Rename `TFDataLayer` to `DataLayer` to reflect that it now supports both `tf.data` and `grain`.
2. Adds an example and tests for `DataLayer`.
3. Update the docstrings in `DataLayer` subclasses.
4. Fixes a device placement issue with Variable in `backend.torch.core.convert_to_tensor`.


fchollet (approved):
LGTM, nice work.

PR #21606, opened 2025-08-21
Implement hypot function in keras.ops

shashaka:
Adds keras.ops.hypot, which computes the element-wise hypotenuse of right triangles given two tensor inputs (x1 and x2).
Equivalent to computing sqrt(x1**2 + x2**2) with broadcasting support.

Not supported on OpenVINO.

### DIFF
--- a/keras/src/ops/numpy.py
+++ b/keras/src/ops/numpy.py
@@ -1355,48 +1355,6 @@ def heaviside(x1, x2):
     return backend.numpy.heaviside(x1, x2)
 
 
-class Hypot(Operation):
-    def call(self, x1, x2):
-        return backend.numpy.hypot(x1, x2)
-
-    def compute_output_spec(self, x1, x2):
-        dtype = dtypes.result_type(x1.dtype, x2.dtype)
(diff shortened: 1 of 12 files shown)


keras/src/ops/numpy.py
```diff
@@ -3514,6 +3514,51 @@ def hstack(xs):
...
+        if dtype in ["int8", "int16", "int32", "uint8", "uint16", "uint32"]:
+            dtype = backend.floatx()
+        elif dtype == "int64":
+            dtype = "float64"
+        return KerasTensor(broadcast_shapes(x1.shape, x2.shape), dtype=dtype)
+
+
+@keras_export(["keras.ops.hypot", "keras.ops.numpy.hypot"])
+def hypot(x1, x2):
+    """Compute the element-wise hypotenuse of
```
fchollet: Please make the 1-line description fit in one line, e.g. 

```
Element-wise hypotenuse of right triangles with legs `x1` & `x2`.
```

shashaka: I've updated this. Thanks.

PR #21608, opened 2025-08-22
Fixes regex fallback logic in DTypePolicyMap to avoid dtype mismatches

JyotinderSingh:
## Issue
The regex fallback logic in `DTypePolicyMap.__getitem__` could incorrectly assign a quantized policy to a non-quantized layer.

This issue arises when deserializing models, particularly from KerasHub, where policies for quantized layers are stored in a `DTypePolicyMap`. The keys for these policies are intended for exact matches. However, the regex fallback can cause a non-quantized layer's path (e.g., `"attention/query_norm"`) to erroneously match a key associated with a quantized policy (e.g., `"attention/query"`).

Consequently, the non-quantized layer receives the wrong policy and fails during the forward pass when it attempts to call a quantized_call method it doesn't have.

## Fix
Changed the fallback logic from implicit component matching to explicit regex matching using `re.fullmatch`. This makes the matching behavior more predictable. Keys in the policy map are now treated as full regular expression patterns that must match the entire layer path.

This means:
1.  A key of `"attention/query"` will only match a layer with the exact path `"attention/query"`.
2.  It will **not match** related paths like `"attention/query/sub_layer"` or `"attention/query_norm"`.
3.  To match all sublayers within a scope, an explicit wildcard must be used and stored in the map, e.g., `"attention/.*"`.

## Testing
Modified the existing tests to cover all possible cases.

## Colab
1. **Failing Case**: This [notebook](https://colab.research.google.com/gist/JyotinderSingh/438b8284ba81df09a2611815fd469fa8/incorrect-dtype-policy-matching-issue-with-keras-layers.ipynb) shows that a quantized `gemma3_1b` KerasHub model fails to perform inference after being saved and loaded. The happens since the the `query_norm` (an `RMSNormalization` layer) matches the regex for another quantized layer called `query` (an `EinsumDense` layer for calculating query values) when resolving the dtype.
3. **Successful Case**: This [notebook](https://colab.research.google.com/gist/JyotinderSingh/37188d481c058930b18f55867b839e05/-fixed-dtype-policy-matching-issue-with-keras-layers.ipynb) shows that the above issue no longer occurs after this fix is applied.

### DIFF
--- a/keras/src/dtype_policies/dtype_policy_map.py
+++ b/keras/src/dtype_policies/dtype_policy_map.py
@@ -35,29 +35,54 @@ class DTypePolicyMap(DTypePolicy, MutableMapping):
     However, it is also possible to set a regex as the key. See the docstring of
     `get` for more details.
 
-    See below for a usage example. You can define the naming schema
-    of the `DTypePolicy`, and then retrieve the corresponding `DTypePolicy`
-    instance.
-
-    ```python
-    dtype_policy_map = DTypePolicyMap()
-    dtype_policy_map["layer/dense_0"] = DTypePolicy("bfloat16")
-    dtype_policy_map["layer/dense_1"] = QuantizedDTypePolicy("int8", "bfloat16")
-
-    policy_0 = dtype_policy_map["layer/dense_0"]
-    policy_1 = dtype_policy_map["layer/dense_1"]
-    policy_2 = dtype_policy_map["layer/dense_2"]  # No hit
-    assert policy_0 == DTypePolicy("bfloat16")
-    assert policy_1 == QuantizedDTypePolicy("int8", "bfloat16")
-    assert policy_2 == keras.config.dtype_policy()
-    ```
-
     Args:
         default_policy: An optional `DTypePolicy` instance specifying the
             default dtype policy. If not specified, the value will default to
             `keras.config.dtype_policy()`.
         policy_map: An optional dict that maps string to `DTypePolicy`
             instances. Defaults to `None`
+
+    Example:
+
+    ```python
+    >>> from keras.src import dtype_policies
+    >>> bfloat16 = dtype_policies.DTypePolicy("bfloat16")
+    >>> float16 = dtype_policies.DTypePolicy("float16")
+    >>> float32 = dtype_policies.DTypePolicy("float32")
+    >>> policy_map = DTypePolicyMap(default_policy=float32)
+
+    # Set policies using an exact path and a regex pattern.
+    # Note: "decoder" will only match the exact path, not its children.
+    >>> policy_map["encoder/layer_0/dense"] = bfloat16
+    >>> policy_map["encoder/.*"] = float16
+    >>> policy_map["decoder"] = bfloat16
+
+    # 1. An exact match is found and returned directly.
+    >>> policy_map["encoder/layer_0/dense"].name
+    'bfloat16'
+
+    # 2. A regex match is found for a child layer.
+    # It matches the "encoder/.*" pattern.
+    >>> policy_map["encoder/attention/query"].name
+    'float16'
+
+    # 3. No implicit prefix matching occurs.
+    # "decoder/attention" does not match the key "decoder".
+    # The default policy is returned.
+    >>> policy_map["decoder/attention"].name
+    'float32'
+
+    # 4. A ValueError is raised if a path matches multiple patterns.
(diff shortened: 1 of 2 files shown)


keras/src/dtype_policies/dtype_policy_map.py
```diff
@@ -98,35 +98,64 @@ def quantization_mode(self):
...
+            # This logic specifically handles how KerasHub serializes quantized
+            # models. In that workflow, every quantized layer is stored in the
+            # policy map with its exact path, ensuring it's found by the direct
+            # lookup at the start of this method.
+            #
+            #  For example, it stops a general key like "attention/query" which
+            # represents a quantization-compatible layer from incorrectly
+            # matching a different, incompatible layer like
+            # "attention/query_norm"
+            if not policy.quantization_mode:
```
mattdangerw: I'm not sure I'm following the exact bug here, but this seems like we are losing a lot of flexibility with the API here. The hope was this could also be used as a more granular dtype setting API for non quantized cases, say running at different precisions for different layers.

```
dtype_policy_map = DTypePolicyMap()
dtype_policy_map["layer/dense_0"] = DTypePolicy("float32")
dtype_policy_map["layer/dense_1"] = DTypePolicy("bfloat16")
```

Won't we lose that ability with this change? Is there another way we can fix whatever the issue is here?

JyotinderSingh: We don’t lose this functionality - layers can still have different dtypes, including distinct quantized dtypes, on a per-layer basis.

The change only affects what happens when **no exact match for a layer path exists** in the map.
* **Non-quantized layers**: The behavior remains unchanged. If a layer path isn’t found, we still run a regex search across all map keys to try and inherit the dtype from a parent layer. For example, if `decoder/attn/dense_01` has no explicit entry, it will inherit the dtype from `decoder/attn` if that entry exists.
* **Quantized layers**: Regex matches that resolve to quantized dtypes are intentionally ignored. There are two reasons:
    1. In KerasHub, every quantized layer is guaranteed to have an explicit entry in the map, so regex fallback should never be needed.
    2. A quantized parent does not imply its sublayers must also be quantized.

JyotinderSingh: As discussed offline, I've replaced the logic to instead use a more robust regex matching logic instead of incorporating special handling for quantized dtypes which makes this behavior easier to understand and predictable.

keras/src/dtype_policies/dtype_policy_map.py
```diff
@@ -98,26 +98,45 @@ def quantization_mode(self):
...
-            if re.search(k, key):
-                matching_keys.append(k)
+        # 2. If no exact match is found, fallback to a regex match.
+        # Check for a match that covers a full path component.
+        # The pattern must match from the start of the `key` and be
+        # followed by either a '/' or the end of the string.
+        matching_keys = [
+            pattern
+            for pattern in self._policy_map
+            if re.search(f"^{pattern}(/|$)", key)
```
mattdangerw: re.fullmatch might be better than this, and get the same behavior

JyotinderSingh: `re.fullmatch` would require a complete string match, it would fail if we try to match the pattern `a/b` in key `a/b/c`.
```py
>>> re.fullmatch("a/b(/|$)", "a/b/c") # No matches
>>> re.match("a/b(/|$)", "a/b/c")
<re.Match object; span=(0, 4), match='a/b/'>
```

We can use `re.match` though, which I have done now.

mattdangerw: Is that what we want though? I think of this as allowing our matching to be fully specific. E.g. if you have layers `model/composite` and `model/composite/subthing`, there should be a way to match only `model/composite`.

If we switch to full match we'd effectively be asking users to be explicit about the leading and trailing wildcard. So `a/b/.*` if you want to match all children. `a/b.*` if you want to also match b as a prefix to a layer name. `.*dense.*` if you want to match all layers with dense anywhere in the path.

If seems like if we are making this move toward fully specific matches, than switching to `fullmatch` might be the consistent thing to do here.

JyotinderSingh: Ah, that does make more sense. Thanks for explaining this in detail, I agree that this API would be properly configurable by the user, and generally more predictable. I've made the change.

mattdangerw (approved):
awesome! thank you

PR #21609, opened 2025-08-22
Add TF support for mixed optional inputs (None/tensor) from generator in model.fit/evaluate/predict

neo-alex:
This PR is a follow-up to the previous PR #21548 in order to address the last uncovered edge case: supporting optional inputs in model.fit/evaluate/predict with Tensorflow backend when mixed values (sometimes None, sometimes tensor) are produced by the same generator.

Here is a concrete example:

```python
class OptionalInputLayer(layers.Layer):
    def __init__(self):
        super().__init__()
        self.dense = layers.Dense(2)

    def call(self, x, y=None):
        z = x if y is None else x + y
        return self.dense(z)

# Create model with 2 inputs (the second one being optional)
i1 = Input((2,), name="input1")
i2 = Input((2,), name="input2", optional=True)
outputs = OptionalInputLayer()(i1, i2)
model = Model({"input1": i1, "input2": i2}, outputs)
model.compile(loss=losses.MeanSquaredError)

# Train from generator (optional inputs always None or always tensor)
data_generator1 = (({"input1": np.ones((2, 2)), "input2": None}, np.ones((2, 2))) for _ in range(4))
model.fit(x=data_generator1)  # WORKS

# Train from generator (optional inputs with mixed None/tensor values)
data_generator2 = (({"input1": np.ones((2, 2)), "input2": None if i % 2 == 0 else np.ones((2, 2))}, np.ones((2, 2))) for i in range(4))
model.fit(x=data_generator2)  # DOESN'T WORK IN TENSORFLOW (UNTIL THIS PR)
```

As already discussed with @hertschuh in the previous PR, covering this case in Tensorflow backend is not trivial for a few reasons:
- TF Dataset (used by default to provide data in TF backend) doesn't support None values directly, it requires tf.experimental.Optional instead, so there is a need to convert back & forth between these formats
- for None/tensor > tf.experimental.Optional conversion: detecting optional inputs from data requires observing first batches produced by the generator to try to infer, as [already done in present code](https://github.com/keras-team/keras/blob/b9ff57a8213a7353a65ad8e8759e9e134edf6420/keras/src/trainers/data_adapters/generator_data_adapter.py#L62) (since data adapters are independent from model specification)
- for tf.experimental.Optional > None/tensor conversion: the methods `optional.has_value()`& `optional.get_value()` must be used to unwrap to correct content; however they return a tensor which can be symbolic when code is traced (with the default `run_eagerly=False`) so they can't be reliably tested as a Python bool and can only be used inside a TF control flow operator like `tf.cond`/`tf.switch_case` - additionally, these TF operators don't support returning None in one branch and a tensor in the other, so we suggest to both convert optional inputs & call the step function (which returns fixed output shapes without risk of None) inside a big `tf.switch_case` with as many branches as needed to cover every possible optional presence combination

An existing unit test was also extended to test for the newly supported case.

### DIFF
--- a/keras/src/backend/tensorflow/trainer.py
+++ b/keras/src/backend/tensorflow/trainer.py
@@ -139,6 +139,12 @@ class TensorFlowTrainer(base_trainer.Trainer):
             terms = tf.bitwise.left_shift(flag_vec, shifts)  # shape [n]
             index = tf.reduce_sum(terms)  # scalar int32 in [0, 2^(n-1)]
             ncases = 1 << n  # = 2^n total cases (efficiently computed)
+            if n > 10:
+                warnings.warn(
+                    f"Model has {n} optional inputs. This will create 2^{n} "
+                    "branches in the computational graph, which may be slow to "
+                    "compile and consume a lot of memory."
+                )
 
             # Create a branch function for each possible bitmask combination
             def make_branch(mask: int):
--- a/keras/src/backend/tensorflow/trainer.py
+++ b/keras/src/backend/tensorflow/trainer.py
@@ -68,7 +68,9 @@ class TensorFlowTrainer(base_trainer.Trainer):
             )
             self._loss_tracker.update_state(
                 loss_module.unscale_loss_for_distribution(loss),
-                sample_weight=tf.shape(tree.flatten(x)[0])[0],
+                sample_weight=tf.shape(
+                    next(i for i in tree.flatten(x) if i is not None)
+                )[0],
             )
             if self.optimizer is not None:
                 loss = self.optimizer.scale_loss(loss)
@@ -96,7 +98,9 @@ class TensorFlowTrainer(base_trainer.Trainer):
         )
         self._loss_tracker.update_state(
             loss_module.unscale_loss_for_distribution(loss),
-            sample_weight=tf.shape(tree.flatten(x)[0])[0],
+            sample_weight=tf.shape(
+                next(i for i in tree.flatten(x) if i is not None)
+            )[0],
         )
         return self.compute_metrics(x, y, y_pred, sample_weight=sample_weight)
 
--- a/keras/src/backend/tensorflow/trainer.py
+++ b/keras/src/backend/tensorflow/trainer.py
@@ -109,17 +109,57 @@ class TensorFlowTrainer(base_trainer.Trainer):
         return y_pred
 
     def _autoconvert_optionals(self, step_func):
-        # Wrapper converting (nested) TF Optional in input data to None
+        # Wrapper converting (nested) TF Optional in input data to tensor/None
         @functools.wraps(step_func)
         def wrapper(data):
-            converted_data = tree.map_structure(
-                lambda i: (
-                    None if isinstance(i, tf.experimental.Optional) else i
-                ),
-                data,
-            )
-            result = step_func(converted_data)
-            return result
+            # Flatten inputs
+            flat = tree.flatten(data)
+
+            # List positions of optional inputs
+            opt_pos = [
+                i
+                for i, x in enumerate(flat)
+                if isinstance(x, tf.experimental.Optional)
+            ]
+            if not opt_pos:  # if nothing optional, just call on data (shortcut)
+                return step_func(data)
+
+            # Build bitmask for optionals (1=present, 0=empty)
+            opts = [flat[i] for i in opt_pos]
+            flags = [o.has_value() for o in opts]  # 1 Tensor[bool] per optional
+            flag_vec = tf.cast(tf.stack(flags), tf.int32)  # shape [n]
+
+            # Compute bitmask index via TF ops (traceable with symbolic tensors)
+            n = len(flags)  # number of optional inputs
+            shifts = tf.range(n, dtype=tf.int32)  # [0, 1, 2, ..., n-1]
+            terms = tf.bitwise.left_shift(flag_vec, shifts)  # shape [n]
+            index = tf.reduce_sum(terms)  # scalar int32 in [0, 2^(n-1)]
+            ncases = 1 << n  # = 2^n total cases (efficiently computed)
+
+            # Create a branch function for each possible bitmask combination
+            def make_branch(mask: int):
+                def branch():
+                    # Unwrap optional inputs to tensor/None in flat inputs
+                    inputs = list(flat)
+                    for j, i in enumerate(opt_pos):
+                        if inputs[i].element_spec is None:
+                            inputs[i] = None  # special case: always None
+                        else:
+                            present = ((mask >> j) & 1) == 1
+                            inputs[i] = opts[j].get_value() if present else None
+
+                    # Pack rebuilt inputs like original data
+                    struct_inputs = tree.pack_sequence_as(data, inputs)
+
+                    # Call step_func (same output shapes for all branches)
+                    return step_func(struct_inputs)
+
+                return branch
+
+            branches = [make_branch(m) for m in range(ncases)]
+
+            # Compute result with switch case
+            return tf.switch_case(index, branch_fns=branches)
 
         return wrapper
 
--- a/keras/src/models/model_test.py
+++ b/keras/src/models/model_test.py
@@ -1261,14 +1261,14 @@ class ModelTest(testing.TestCase):
     def test_functional_optional_inputs_generator(self, is_optional_none):
         model = _get_model_optional_inputs()
         x1 = np.ones((2, 2))
-        if is_optional_none != "sometimes":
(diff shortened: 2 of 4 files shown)


neo-alex:
One more side comment wrt. auto-inferring optional inputs from observing the first generated data batches: currently, only the [2 first batches](https://github.com/keras-team/keras/blob/b9ff57a8213a7353a65ad8e8759e9e134edf6420/keras/src/trainers/data_adapters/data_adapter_utils.py#L8) are observed for that, but we could increase it to 3 by default to cover detection of optional inputs with variable dimensions (beyond batch one), e.g. starting with None/tensor of shape (3, 4)/tensor of shape (3, 5). I can do the change if you think it is a good idea.

keras/src/backend/tensorflow/trainer.py
```diff
@@ -109,17 +113,57 @@ def predict_step(self, data):
...
+            opts = [flat[i] for i in opt_pos]
+            flags = [o.has_value() for o in opts]  # 1 Tensor[bool] per optional
+            flag_vec = tf.cast(tf.stack(flags), tf.int32)  # shape [n]
+
+            # Compute bitmask index via TF ops (traceable with symbolic tensors)
+            n = len(flags)  # number of optional inputs
+            shifts = tf.range(n, dtype=tf.int32)  # [0, 1, 2, ..., n-1]
+            terms = tf.bitwise.left_shift(flag_vec, shifts)  # shape [n]
+            index = tf.reduce_sum(terms)  # scalar int32 in [0, 2^(n-1)]
+            ncases = 1 << n  # = 2^n total cases (efficiently computed)
```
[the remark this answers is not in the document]
neo-alex: This is absolutely correct, but after many attempts to solve this issue, this solution is the only valid one I could find despite this potential performance impact. However, note that this could only meaningfully impact performance in this very special case: fit/evaluate/predict with `jit_compile=True` of a TF model with many optional inputs from a generator with mixed values (whereas today this case is not supported at all). I agree that a warning could be a good idea though, I will add it in a new commit in this PR.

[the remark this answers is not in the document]
neo-alex: Done in [this commit](https://github.com/keras-team/keras/pull/21609/commits/e23dba45e3a1bd0319864a8f131a0b6aeee94f96).

keras/src/backend/tensorflow/trainer.py
```diff
@@ -68,7 +68,9 @@ def train_step(self, data):
             )
             self._loss_tracker.update_state(
                 loss_module.unscale_loss_for_distribution(loss),
-                sample_weight=tf.shape(tree.flatten(x)[0])[0],
+                sample_weight=tf.shape(
+                    next(i for i in tree.flatten(x) if i is not None)
```
hertschuh, line 72: This is probably not a valid use case, but in theory `i for i in tree.flatten(x) if i is not None` could run out, in which case `tf.shape` can't be called.

keras/src/backend/tensorflow/trainer.py
```diff
@@ -96,7 +98,9 @@ def test_step(self, data):
         )
         self._loss_tracker.update_state(
             loss_module.unscale_loss_for_distribution(loss),
-            sample_weight=tf.shape(tree.flatten(x)[0])[0],
+            sample_weight=tf.shape(
+                next(i for i in tree.flatten(x) if i is not None)
```
hertschuh, line 102: Same here

keras/src/backend/tensorflow/trainer.py
```diff
@@ -109,17 +113,63 @@ def predict_step(self, data):
...
+
+                    # Call step_func (same output shapes for all branches)
+                    return step_func(struct_inputs)
+
+                return branch
+
+            branches = [make_branch(m) for m in range(ncases)]
+
+            # Compute result with switch case
+            return tf.switch_case(index, branch_fns=branches)
```
hertschuh, line 172: Wow. All this just to unpack `tf.Optional`s.

First I'm impressed that you got this to work. But I'm hoping we can find something simpler. I played with it a bit but couldn't find anything so far. What I'm confused about is why we can't just do `if x.has_value()`.

The related question is whether it will interfere with the existing usage of optionals here: https://github.com/keras-team/keras/blob/master/keras/src/backend/tensorflow/trainer.py#L238

keras/src/backend/tensorflow/trainer.py
neo-alex: Indeed, this assumes that not all inputs are both optional and actually None at the same time (otherwise, I don't see how to infer the batch size anyway...). In any case, this edit makes this line of code (which was present long before my 2 PRs) more robust, now that some optional inputs can be None when using model.fit/evaluate/predict. Without this change, if the first input received (e.g. first alphabetic key in a dict input) is optional and None, it would crash.

keras/src/backend/tensorflow/trainer.py
neo-alex: Same reply here :)

keras/src/backend/tensorflow/trainer.py
neo-alex: "All this just to unpack tf.Optionals" --> I know, believe me I tried many other things before to make this edge case work, without success - but if you find a better alternative, it would be great!

"What I'm confused about is why we can't just do if `x.has_value()`" --> I understand your confusion, especially since a similar `x.has_value()` is used without problem in a Python `if` in [the line of code you mention](https://github.com/keras-team/keras/blob/master/keras/src/backend/tensorflow/trainer.py#L238). Actually, this works here because this part of the code is never traced by TF Autograph (since it isn't wrapped in `tf.function`). Unfortunately, when using a data generator, our `x` of interest is fetched one item at a time from the iterator **inside** [multi_step_on_iterator](https://github.com/keras-team/keras/blob/a11ef39695d95165d193c3d80ab0fb13dfd80daf/keras/src/backend/tensorflow/trainer.py#L147)/[one_step_on_data](https://github.com/keras-team/keras/blob/a11ef39695d95165d193c3d80ab0fb13dfd80daf/keras/src/backend/tensorflow/trainer.py#L128) which are both potentially traced by TF Autograph [here](https://github.com/keras-team/keras/blob/a11ef39695d95165d193c3d80ab0fb13dfd80daf/keras/src/backend/tensorflow/trainer.py#L147)/[there](https://github.com/keras-team/keras/blob/a11ef39695d95165d193c3d80ab0fb13dfd80daf/keras/src/backend/tensorflow/trainer.py#L139) (when `run_eagerly=False`, which is the default setting). During tracing, `x.has_value()` is then a **symbolic** boolean tensor (which cannot be simply evaluated as a boolean in Python since it doesn't yet hold any actual value), so I believe we have no choice but use a TF conditional operator to handle whether x has value or not. Additionally, since TF conditional operators (like `tf.cond`/`tf.switch_case`) don't support having some branch(es) outputting `None` when other branch(es) output a tensor, we **cannot** use a simple `tf.cond(x.has_value(), lambda: x.get_value(), lambda: None)` to unwrap the required value in the format Keras expect. The only workaround I found is then to call the model inside the conditional branches, since it has consistent tensor outputs (no more possible `None` in outputs, as opposed to inputs) which is compatible with TF conditional operators... at the cost of creating a big `tf.switch_case` covering all possible optional input combinations.

"The related question is whether it will interfere with the existing usage of optionals here" --> I don't think that it creates any interference... The solution I propose is just heavier than I wished for (I would also like something simpler), but hopefully my verbose explanation above explains why I did it this way. Of course, I am more than open to a better way if you find one that still enables the new unit test I added in this PR.

neo-alex:
By the way, if you find the solution in this PR too complex/heavy (and if you can't find a better way to cover this edge case), feel free to reject it - my actual need was already covered by the previous PR #21548 and I just added this one for completeness as promised. In that scenario, I would only advocate to accept [this commit](https://github.com/keras-team/keras/pull/21609/commits/0448a389b96feff08e7b9bc58eab5ce8af1d4fb3) which is a generally useful fix (I could re-push it in a tiny dedicated PR if needed).

hertschuh:
> By the way, if you find the solution in this PR too complex/heavy (and if you can't find a better way to cover this edge case), feel free to reject it - my actual need was already covered by the previous PR #21548 and I just added this one for completeness as promised. In that scenario, I would only advocate to accept [this commit](https://github.com/keras-team/keras/pull/21609/commits/0448a389b96feff08e7b9bc58eab5ce8af1d4fb3) which is a generally useful fix (I could re-push it in a tiny dedicated PR if needed).

Let's do that (separate PR). I'll keep playing with this PR to see if there is a simpler way to do this.

Thanks!

hertschuh:
Per comments, this is quite complex for something we haven't needed yet.

PR #21613, opened 2025-08-24
Add linspace and logspace implementations in OpenVINO NumPy backend

arunthakur009:
- Added linspace for generating evenly spaced linear sequences.

- Added logspace built on linspace for log-scaled sequences.

- Supports num, endpoint, base, dtype, and axis parameters.

- Integrated OpenVINO tensor conversion and dtype handling.

Tickets:
 https://github.com/openvinotoolkit/openvino/issues/30114  
https://github.com/openvinotoolkit/openvino/issues/29485


keras/src/backend/openvino/numpy.py
```diff
@@ -997,9 +997,157 @@ def less_equal(x1, x2):
 def linspace(
     start, stop, num=50, endpoint=True, retstep=False, dtype=None, axis=0
 ):
-    raise NotImplementedError(
-        "`linspace` is not supported with openvino backend"
+    """
+    ...
```
fchollet: Please fix the docstring

arunthakur009:
Thanks for the feedback, @fchollet  I’ve fixed the docstring—please take another look when you get a chance.

keras/src/backend/openvino/numpy.py
```diff
@@ -997,9 +997,160 @@ def less_equal(x1, x2):
 def linspace(
     start, stop, num=50, endpoint=True, retstep=False, dtype=None, axis=0
 ):
-    raise NotImplementedError(
-        "`linspace` is not supported with openvino backend"
+    """
+    Return evenly spaced numbers over a specified interval.
+
+    Parameters
```
fchollet: Please fix docstring (see style of other docstrings in the codebase)

arunthakur009:
Thanks for the feedback, @fchollet. I’ve updated the docstrings accordingly. Please let me know if further adjustments are needed.

keras/src/backend/openvino/numpy.py
```diff
@@ -997,9 +997,160 @@ def less_equal(x1, x2):
...
+        step_reshaped = ov_opset.reshape(
+            step, start_target_shape, False
+        ).output(0)
+    else:
+        result_shape = ov_opset.concat([start_shape, indices_shape], 0).output(
+            0
+        )
+
+        ones_for_start = ov_opset.broadcast(
+            ov_opset.constant(1, Type.i32).output(0), start_rank
```
Mohamed-Ashraf273: Please define this constant once and reuse it throughout the code for better readability and maintainability.
Do same for `ov_opset.constant(0, Type.i32)` and `ov_opset.constant([1], Type.i32)`

keras/src/backend/openvino/numpy.py
```diff
@@ -997,9 +997,160 @@ def less_equal(x1, x2):
...
+        step_reshaped = ov_opset.reshape(
+            step, start_target_shape, False
+        ).output(0)
+
+    indices_broadcasted = ov_opset.broadcast(
+        indices_reshaped, result_shape
+    ).output(0)
+    start_broadcasted = ov_opset.broadcast(start_reshaped, result_shape).output(
+        0
     )
```
Mohamed-Ashraf273, line 1155: Too many explicit broadcasts are used here. With proper reshapes, elementwise ops handle broadcasting automatically.  
For example, `start_broadcasted = ov_opset.broadcast(start_reshaped, result_shape)` is unnecessary since `add`/`multiply` already broadcast. Use fewer `broadcast` ops as they add memory overhead.

arunthakur009:
@Mohamed-Ashraf273 I’ve applied the suggested changes. Please take a look when you have time.

keras/src/backend/openvino/numpy.py
```diff
@@ -997,9 +997,143 @@ def less_equal(x1, x2):
...
+        start_target_shape = ov_opset.concat(
+            [start_shape, one_i32_array], 0
+        ).output(0)
+        start_reshaped = ov_opset.reshape(
+            start, start_target_shape, False
+        ).output(0)
+        step_reshaped = ov_opset.reshape(
+            step, start_target_shape, False
+        ).output(0)
+
```
Mohamed-Ashraf273, line 1157: This part could be simplified by avoiding duplication. Instead of repeating both `concat` and `reshape` logic inside each branch, only determine the `concat` order in the `if/else`, and then apply the `reshape` once afterwards. For example:
```python
if axis == 0:
    indices_target_shape = ov_opset.concat([indices_shape, ones_for_start], 0).output(0)
    start_target_shape = ov_opset.concat([one_i32_array, start_shape], 0).output(0)
else:
    indices_target_shape = ov_opset.concat([ones_for_start, indices_shape], 0).output(0)
    start_target_shape = ov_opset.concat([start_shape, one_i32_array], 0).output(0)

indices_reshaped = ov_opset.reshape(indices, indices_target_shape, False).output(0)
start_reshaped = ov_opset.reshape(start, start_target_shape, False).output(0)
step_reshaped = ov_opset.reshape(step, start_target_shape, False).output(0)
```

arunthakur009:
@Mohamed-Ashraf273 Thanks for the suggestion. I’ve applied the suggested changes. Please take a look when you have time.

Mohamed-Ashraf273:
@fchollet 
Looks good to me.
Could you please review it?

PR #21623, opened 2025-08-28
Implement gcd function in keras.ops

shashaka:
Adds keras.ops.gcd, which computes the element-wise greatest common divisor of two tensor inputs (x1 and x2).
Equivalent to applying the integer gcd operation with broadcasting support.

Not supported on OpenVINO.


fchollet (approved):
LGTM -- thanks for continuing to increase ops coverage in the API!

PR #21628, opened 2025-08-30
GPTQ refactor, GPU optimization, speedup, and test enhancements

JyotinderSingh:
## Issues with existing implementation
- GPTQ logic was spread across multiple modules.
- Activation capture and Hessian building were done via ad-hoc input collection, accumulating unnecessary tensor copies and GPU residency, leading to high peak GPU memory.
- Quantization math lived in a bespoke class with duplicated functionality, limiting reuse and testability.
- `GPTQQuantizer` did not support serialization like other quantizers in the codebase.
- Dataloader utilities were not configurable and non-deterministic.
- Existing test suite lacked in-depth numeric verification for quantization math, hessian updates, and layer corrections.

## Refactor Details
- Single entry point: `model.quantize("gptq", config)` now calls `keras.src.quantizers.gptq_core.gptq_quantize(model, config)`.  
  - `GPTQConfig.quantize()` is removed to keep configuration class lean.
- Streamed Hessian capture: New `stream_hessians(...)` context manager updates the Hessian online from real activations without large buffers. Hooks are restored even on exceptions.  
  - Reduces GPU residency and improves control flow.
- Consolidated quantization math: Moves per-tensor/per-channel/grouped parameter computation into `quantizers.compute_quantization_parameters(...)` with a `GPTQQuantizer` wrapper.  
  - Deduplicates logic, improves reusability and testing.
- Blockwise GPTQ core: Extracts correction loop into `gptq_quantize_matrix(...)` with stable activation-order permutation and explicit block/group handling.

## Additional Improvements
- Deterministic calibration windows: `get_dataloader(...)` supports `{"strided","linspace","random"}` strategies, `seed`, `stride`, and optional `eos_id` insertion.
- Config surface improvements: `GPTQConfig` now includes `per_channel` (default `True`) plus `group_size`, `symmetric`, `activation_order`, `hessian_damping`.
- Stable activation order: `_stable_permutation(...)` ensures deterministic tie-breaking for reproducible results.
- Numerical improvements: Implements Hessian symmetrization (for large activations) and zero-scale guardrails

## Test Coverage Improvements
- Dataloader correctness & determinism: Shapes, dtypes, seed reproducibility, strided coverage, EOS placement.
- Hessian math: Streaming vs. big-batch equivalence, closed-form checks (`2/N * XᵀX`), reshaping, stability with large inputs.
- Layer compatibility: Dense and EinsumDense coverage, activation-order round-trip.
- Quantization math Idempotence tests, clip-boundary behavior, grouped/per-channel invariants, dtype/finite checks.
- End-to-End numerics tests: Ensures end-to-end correctness of results from quantized models by comparing against full-precision results.

## Metrics (GPT-2 & OPT-125M)

| **Metric**           | **GPT-2 (Original)** | **GPT-2 (Refactored)** | **OPT-125M (Original)** | **OPT-125M (Refactored)** | **Observation**                                |
| -------------------- | -------------------- | ---------------------- | ----------------------- | ------------------------- | ---------------------------------------------- |
| **Pre-Perplexity**   | 110.03               | 110.03                 | 77.06                   | 77.06                     | Baseline same                                  |
| **Post-Perplexity**  | 114.79               | 114.39                 | 80.50                   | 83.42                     | Similar (within margin of error)          |
| **Δ Perplexity**     | +4.76                | +4.36                  | +3.44                   | +6.36                     | Similar (within margin of error)             |
| **Runtime (sec)**    | 245.76               | 224.17                 | 230.75                  | 209.56                    | Refactor ~9% faster                            |
| **CPU Peak (bytes)** | 2.47 GB              | 2.60 GB                | 2.55 GB                 | 2.70 GB                   | Refactor uses ~130–150 MB more CPU RAM         |
| **GPU Peak (bytes)** | 1.51 GB              | 0.81 GB                | 1.51 GB                 | 0.86 GB                   | Refactor halves GPU usage (~45% lower)         |

**Summary:**  
- **Speed:** ~9% faster.  
- **GPU memory:** ~45% lower.  
- **CPU memory:** Slightly higher due to dataloader offload.
- **Accuracy:** Unchanged, differences within margin of error.

> Repro: [colab notebook](https://colab.research.google.com/drive/14wqS3s00F2F80xfTKi24-_yN0Vk0JcIu?usp=sharing)

## Next Steps
This refactor doesn't include major functional differences. The implementation still implements simulated quantization, which stores dequantized full-precision weights in the layer kernel like before. Next steps include true quantization, which would store low-precision kernels and provide measurable GEMM speedups.


keras/src/quantizers/gptq_core.py
```diff
@@ -9,15 +10,109 @@
...
+         exit, all layers are unpatched and safe to use normally.
+
+    Example:
+    ```python
+    >>> with stream_hessians(layers_map, gptq_objects):
+    ...     for sample in calibration_inputs:
+    ...         if len(sample.shape) == 2:
+    ...             sample = ops.expand_dims(sample, 0)
+    ...         _ = block(sample)   # hooks update Hessians on-the-fly
+    # <- original layer.call methods restored here
```
hertschuh: Don't you need `>>>` here ?

keras/src/quantizers/gptq.py
```diff
@@ -340,11 +436,7 @@ def quantize_and_correct_block(
             )
 
         # Set the new quantized weights in the original layer
-        new_weights = [ops.convert_to_numpy(quantized_weights)]
-        if self.original_layer.bias is not None:
-            new_weights.append(ops.convert_to_numpy(self.original_layer.bias))
-
-        self.original_layer.set_weights(new_weights)
+        self.original_layer._kernel.assign(quantized_weights)
```
hertschuh, line 439: This is not injecting the `zeros` and `scale` variables yet. If I understand the PR description, you'll do that in a separate PR, right?

However, isn't the quantized kernel just garbage if not used with the proper scale and zeros?

JyotinderSingh: What we're injecting into the layer is the _dequantized_ kernel. The GPTQ implementation at present (and in the original PR) quantizes and dequantizes the kernel to simulate the effects of actual quantization.

This is the high-precision kernel we get back after the simulated quant-dequant process. It includes some quantization error as well. It works as a proof of concept that the error correction mechanism keeps the quantization error low.

The work in upcoming PRs will involve storing the actual low-precision kernel, along with scales and zeros.

hertschuh (approved):

keras/src/quantizers/gptq_core.py
JyotinderSingh: Yes, fixed it.

PR #21636, opened 2025-09-07
Implement lcm function in keras.ops

shashaka:
Adds keras.ops.lcm, which computes the element-wise least common multiple of two tensor inputs (x1 and x2).
Equivalent to applying the integer lcm operation with broadcasting support.

Not supported on OpenVINO.

### DIFF
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -1761,12 +1761,15 @@ def lcm(x1, x2):
     x1 = convert_to_tensor(x1)
     x2 = convert_to_tensor(x2)
 
+    if not (x1.dtype.is_integer and x2.dtype.is_integer):
+        raise TypeError(
+            f"Arguments to lcm must be integers. "
+            f"Received: x1.dtype={x1.dtype.name}, x2.dtype={x2.dtype.name}"
+        )
+
     dtype = dtypes.result_type(x1.dtype, x2.dtype)
     x1 = tf.cast(x1, dtype)
     x2 = tf.cast(x2, dtype)
-
-    if not x1.dtype.is_integer:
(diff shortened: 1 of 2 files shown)


keras/src/backend/tensorflow/numpy.py
```diff
@@ -1757,6 +1757,32 @@ def expand_front(x, num):
...
+def lcm(x1, x2):
+    x1 = convert_to_tensor(x1)
+    x2 = convert_to_tensor(x2)
+
+    dtype = dtypes.result_type(x1.dtype, x2.dtype)
+    x1 = tf.cast(x1, dtype)
+    x2 = tf.cast(x2, dtype)
+
+    if not x1.dtype.is_integer:
+        raise TypeError("Arguments to lcm must be integers.")
```
fchollet: Please print what `x1` was passed (e.g. dtype)

keras/src/ops/numpy.py
```diff
@@ -3888,6 +3888,43 @@ def kron(x1, x2):
...
+    """Least common multiple of `x1` and `x2`, element-wise.
+
+    Args:
+        x1: First input tensor (integer type).
+        x2: Second input tensor (integer type).
+
+    Returns:
+        Output tensor, element-wise least common multiple of `x1` and `x2`.
+
+    Examples:
```
fchollet: Singular (there's 1 example here)

fchollet (commented):
Thanks for the PR!

keras/src/backend/tensorflow/numpy.py
shashaka: I've changed.

keras/src/ops/numpy.py
shashaka: I've updated docstring.

PR #21641, opened 2025-09-08
Adds Native GPTQ Layer, Runtime Execution, and Serialization Support

JyotinderSingh:
## Description of the change

This PR introduces a dedicated `GPTQDTypePolicy` and integrates GPTQ as a first-class quantization mode in Keras. It adds build/call paths for `"gptq"` in `Dense` and `EinsumDense`, persists GPTQ parameters (quantized weights, scales, zero points, group indices), and wires the model-level `quantize()` flow to allocate GPTQ variables, execute the GPTQ pass, and rebuild execution graphs as needed.

The core GPTQ quantizer is modified to return quantized weights and all required group parameters in a single pass. Serialization is supported by recording `weight_bits` and `group_size` in the dtype policy (and in `DTypePolicyMap` entries when used), so GPTQ-quantized KerasHub models reload correctly when using the Keras model format.

### `keras.dtype_policies.GPTQDTypePolicy`
* Encodes `mode="gptq"`, `weight_bits`, and `group_size`.

### Layer-level GPTQ support (inference path)
* `Dense` and `EinsumDense` now include `_gptq_build(...)` and `_gptq_call(...)`.
* Runtime forward pass uses `dequantize_with_sz_map()` to reconstruct weights on the fly from:
  * `quantized_kernel: uint8`
  * `kernel_scale: float32`
  * `kernel_zero: uint8`
  * `g_idx: int32`
* Group-wise Quantization/Dequantization happens purely with vectorized operations to minimize overhead.

### Model integration
* `Model.quantize("gptq", config=GPTQConfig(...))` now:
  1. Invokes layer-local `quantize(..., config=...)` to allocate GPTQ variables and set dtype policy (with `weight_bits`/`group_size` baked in).
  2. Runs the GPTQ pass (`gptq_quantize(self, config)`).
  3. Rebuilds training/eval/predict functions when the graph is modified.

### Serialization
* Modified `load_own_weights(...)` and `save_own_weights(...)` methods in `Dense`/`EinsumDense` to support serialization of GPTQ-quantized weights, scales, zeros, and group indices.

### Quantization utilities
* Public helpers `quantize_with_sz_map()` and `dequantize_with_sz_map()` for group-mapped (de)quantization.

## Benchmarks
* Native GPTQ benchmark: [colab notebook](https://colab.research.google.com/drive/1x2gjXji8jQHG-lVdRvDR-M6S8rf_59zg?usp=sharing)
Results: [benchmarks_native_gptq.csv](https://github.com/user-attachments/files/22259943/benchmarks_native_gptq.csv)

* Existing Simulated GPTQ benchmark: [colab notebook](https://colab.research.google.com/drive/1RDlTSYLh_c1SeE58AqwytzxqXTZLZ6P1?usp=sharing)
Results: [benchmarks_simulated_gptq.csv](https://github.com/user-attachments/files/22260985/benchmarks_simulated.csv)

[Complete comparative analysis](https://github.com/user-attachments/files/22261169/gptq_comparative_report.csv)

### Benchmark Summary
| Metric (native − simulated)                      |    Mean change | What it means                                                          |
| ------------------------------------------------ | -------------: | ---------------------------------------------------------------------- |
| Quantization time                                |     **−4.69%** | Native calibrates a bit faster overall.                                |
| Quantization CPU peak                            |     **−0.43%** | Roughly the same CPU peak during calibration.                          |
| Quantization GPU peak                            |    **−31.40%** | Native uses far less GPU memory while quantizing.                      |
| Disk reduction vs FP32        | **-49.67%** | Native actually shrinks checkpoints; simulated does not.               |
| Inference VRAM                                   |    **−48.49%** | About half the VRAM at inference with native.                          |
| First-token latency                              |    **+12.18%** | Native is slightly slower to first token (runtime dequant cost).       |
| Throughput                                       |     **−1.40%** | Throughput is roughly a wash overall (model-dependent).                |

### Absolute Means for Context
| Metric                                | Simulated (mean) | Native (mean) |
| ------------------------------------- | ---------------: | ------------: |
| Inference VRAM (GB)                   |            2.611 |     **1.254** |
| Quant GPU peak (GB)                   |            3.989 |     **3.093** |
| First-token latency (ms)              |        **40.13** |         42.78 |
| Throughput (tokens/s)                 |       **419.53** |        406.05 |
| Quant time (sec)                      |           662.44 |    **593.84** |
| Avg. accuracy degradation vs FP (pct) |             8.04 |      **5.31** |
| Avg. disk reduction vs FP32           |         **0.0%** |    **49.67%** |

### Per-model Comparison
| Model (preset)                                    | d_PPL_sim % | d_PPL_nat % | Disk GB (sim → nat) | VRAM GB (sim → nat) | First-token (ms) sim → nat | Throughput (t/s) sim → nat |
| ------------------------------------------------- | ---------: | ---------: | ------------------: | ------------------: | -------------------------: | -------------------------: |
| gpt2_causal_lm (gpt2_base_en_cnn_dailymail)       |       1.11 |    **0.98** |   0.465 → **0.232** |   0.582 → **0.343** |              38.64 → 38.92 |        402.63 → **483.40** |
| opt_causal_lm (opt_125m_en)                       |      11.83 |    **9.97** |   0.468 → **0.235** |   0.656 → **0.348** |          **44.51 → 47.49** |        **813.74 → 686.00** |
| bloom_causal_lm (bloom_1.1b_multi)                |    **5.23** |       7.03 |   3.985 → **2.113** |   4.995 → **2.296** |          **61.66 → 62.79** |        **156.00 → 131.55** |
| gemma3_causal_lm (gemma3_1b)                      |      14.00 |    **3.24** |   3.730 → **1.808** |   4.210 → **2.029** |          **15.72 → 21.93** |        305.75 → **323.26** |

Notes:
1. Lower d_PPL is better.
2. Disk "sim" shows no meaningful reduction vs FP32 overall; the big cuts are purely from the native layout.
3. Throughput impact varies by model; Gemma3/GPT-2 improved, OPT/BLOOM dipped a bit. First-token is consistently a touch slower with native due to on-the-fly dequant.

## Testing
* Added tests to verify that GPTQ-quantized layers can be serialized/deserialized and rebuilt with correct shapes and params.
* Added tests to exercise full quantization path: layer quantize → Hessian update → GPTQ pass → state cleanup.
* Added tests to validate per-column equivalence via `dequantize_with_sz_map`.
* Added tests to ensure activation permutation is undone (ordered vs unordered produce identical final weights).
* Expanded combination coverage: group-wise, activation-order, symmetric, per-channel, and 8-bit variants.

## Follow-up Work

* Support for 2-bit and 4-bit integer packing will be added in subsequent PRs.
* We can explore activation quantization to enable low-precision GEMM speedups.
* To reduce numeric variance and reduce surface for bugs in scope for this PR, we haven't migrated `linalg.inverse` to `choleskly`/`cholesky_inverse` (which was added in #21554). This should be a trivial to implement in future changes.


keras/src/dtype_policies/dtype_policy.py
```diff
@@ -288,6 +288,54 @@ def get_config(self):
...
+            Typically a smaller group size leads to better accuracy but
+            slower speed.
+    """
+
+    def __init__(
+        self,
+        mode,
+        source_name=None,
+        weight_bits=None,
+        group_size=None,
```
hertschuh: I don't think it will work if `weight_bits` is `None` or `group_size` is `None`.

keras/src/dtype_policies/dtype_policy.py
```diff
@@ -352,6 +400,8 @@ def _get_quantized_dtype_policy_by_str(policy):
     mode, source_name = split_name
     if policy.startswith("int8") or policy.startswith("int4"):
         return QuantizedDTypePolicy(mode, source_name)
+    elif policy.startswith("gptq"):
+        return GPTQDTypePolicy(mode, source_name)
```
hertschuh, line 444: What about the other parameters: `weight_bits` and `group_size`? Where are they suppose to come from?

Isn't the dtype string supposed to fully define how the quantization will work (so that you can serialize / deserialize the dtype policy as a string)?

For instance, should the string be like be `"gptq4/-1_from_float32"`?

@james77777778

keras/src/layers/core/dense.py
```diff
@@ -108,8 +113,11 @@ def __init__(
     def build(self, input_shape):
         kernel_shape = (input_shape[-1], self.units)
         if self.quantization_mode:
-            self.quantized_build(kernel_shape, mode=self.quantization_mode)
-        if self.quantization_mode not in ("int8", "int4"):
+            self.quantized_build(
+                kernel_shape,
+                mode=self.quantization_mode,
```
hertschuh: nitpick: remove comma and make it one line again.

keras/src/layers/core/dense.py
```diff
@@ -246,22 +282,40 @@ def load_own_variables(self, store):
...
+
+        targets = []
+        if mode != "gptq":
+            targets.append(self._kernel)
+        if self.bias is not None:
+            targets.append(self.bias)
+        targets.extend(getattr(self, name) for name in MODE_SPEC[mode])
+
+        for i, variable in enumerate(targets):
             variable.assign(store[str(i)])
```
hertschuh, line 324: It's unfortunate that we didn't keep the names and instead assigned numbers.

What was the reason @james77777778 ?

keras/src/layers/core/dense.py
```diff
@@ -323,15 +377,15 @@ def _check_load_own_variables(self, store):
                 f"Expected: {[v.name for v in all_vars]}"
             )
 
-    # Quantization-related (int8 and float8) methods
-
-    def quantized_build(self, kernel_shape, mode):
+    def quantized_build(self, input_shape, mode, config=None):
```
hertschuh: Undo the rename, i.e. go back to `kernel_shape` instead of `input_shape`. It is the kernel shape and not the input shape.

keras/src/layers/core/dense.py
```diff
@@ -352,6 +406,66 @@ def _int8_build(self, kernel_shape):
             trainable=False,
         )
 
+    def _gptq_build(self, kernel_shape, config):
+        self.gptq = False
+        # self.inputs_quantizer = quantizers.AbsMaxQuantizer(axis=-1)
```
hertschuh: Remove?

keras/src/layers/core/dense.py
```diff
@@ -662,6 +778,9 @@ def quantize(self, mode, type_check=True):
             from keras.src import dtype_policies  # local import to avoid cycle
 
             policy = dtype_policies.get(f"{mode}_from_{self.dtype_policy.name}")
+            if mode == "gptq":
+                policy.weight_bits = config.weight_bits
+                policy.group_size = config.group_size
```
hertschuh: This goes back to the comment on whether the policy name should contain the whole config or not.

Which one should it be?

keras/src/layers/core/dense.py
```diff
@@ -693,7 +812,7 @@ def _get_kernel_with_merged_lora(self):
                 `kernel_scale`: The quantization scale for the merged kernel.
                     This is `None` if the layer is not quantized.
         """
-        if self.dtype_policy.quantization_mode is None:
+        if self.dtype_policy.quantization_mode in (None, "gptq"):
```
hertschuh, line 823: So you can't do LoRa with GPTQ? Should we raise an error if you try?

keras/src/layers/core/dense.py
```diff
@@ -746,3 +865,38 @@ def _get_kernel_with_merged_lora(self):
...
+        Returns:
+            int. The determined group size for GPTQ quantization.
+        Raises:
+            ValueError: If the group size is not specified in either the
+                `config` or the `dtype_policy`.
+        """
+        if isinstance(self.dtype_policy, GPTQDTypePolicy):
+            policy_group_size = self.dtype_policy.group_size
+        elif isinstance(self.dtype_policy, DTypePolicyMap):
+            policy_group_size = self.dtype_policy[self.path].group_size
```
hertschuh: Don't you need to check if `self.dtype_policy[self.path]` is a `GPTQDTypePolicy` first?

Also, is there a precedent for having both a local config and a policy that can disagree? It can easily become confusing.

keras/src/layers/core/dense.py
```diff
@@ -746,3 +865,38 @@ def _get_kernel_with_merged_lora(self):
...
+        else:
+            policy_group_size = None
+        group_size = config.group_size if config else policy_group_size
```
hertschuh: How about doing this line 888:
```python
if config and config.group_size is not None:
  return config.group_size
```
This way it's obvious that `config` is the default and the policy is the fallback. Otherwise you have to read and parse a few lines of code to understand what's going on line 894.

keras/src/layers/core/dense_test.py
```diff
@@ -959,3 +960,18 @@ def test_quantize_int4_when_lora_enabled(self):
...
+        layer.build((None, 8))
+        layer.quantize(
+            "gptq",
+            config=GPTQConfig(
+                dataset=None, tokenizer=None, weight_bits=4, group_size=8
+            ),
+        )
+        config = layer.get_config()
+        new_layer = layers.Dense.from_config(config)
+        new_layer.build((None, 8))
```
hertschuh, line 977: Is `new_layer` quantized? Can you verify?

keras/src/layers/core/einsum_dense.py
```diff
@@ -387,15 +441,15 @@ def _check_load_own_variables(self, store):
                 f"Expected: {[v.name for v in all_vars]}"
             )
 
-    # Quantization-related (int8 and float8) methods
-
-    def quantized_build(self, kernel_shape, mode):
+    def quantized_build(self, input_shape, mode, config=None):
```
hertschuh: Same comment: `kernel_shape` was correct.

keras/src/layers/core/dense.py
JyotinderSingh: Done

keras/src/layers/core/dense.py
JyotinderSingh: Done.

keras/src/layers/core/dense.py
JyotinderSingh: Done

keras/src/layers/core/dense.py
JyotinderSingh: Done. Enabling LoRA now raises an error when used with GPTQ quantized models.

keras/src/layers/core/dense.py
JyotinderSingh: It's not really required to explicitly check the type of the policy. The fact that this layer is being detected as a GPTQ quantized layer is itself grounded in the dtype_policy found inside the DTypePolicyMap for this layer. There should be no disagreement. However, for safety, I've added a check.

This is how we check the quantization mode of a layer, you can see it comes from the policy/policy_map itself: https://github.com/keras-team/keras/blob/2d4dce388280d89a0fe3b1bb0a21690714550b44/keras/src/layers/layer.py#L785-L792

keras/src/layers/core/dense.py
JyotinderSingh: Makes sense, done.

keras/src/layers/core/einsum_dense.py
JyotinderSingh: Reverted

keras/src/dtype_policies/dtype_policy.py
JyotinderSingh: I was setting `weight_bits` and `group_size` directly on the policy object at the call sites inside `Dense` and `EinsumDense`.

But this is a great suggestion. I've modified the implementation to encode the config within the policy's string name. Something like `"gptq/4/-1"`. It didn't initially occur to me that we could do that :D

keras/src/dtype_policies/dtype_policy.py
JyotinderSingh: No longer relevant after modifications to encode config into the name based on the other comment

keras/src/layers/core/dense.py
JyotinderSingh: The policy name now encodes the complete config.

JyotinderSingh:
> Thanks for the PR!
> 
> This is a first pass. I have a high level question and now I'm having second thoughts about the `config` parameter in `quantize`.
> 
> * What is supposed to come from `mode`? All the information needed to quantize or not?
> * How are `GPTQConfig` and `GPTQDTypePolicy` supposed to work together? Do we need both?

I am okay with removing the GPTQConfig in support of independent kwargs that can be propagated down to the `quantize(...)` methods of the layers. But maybe we should take that up in a subsequent PR (since this one is already quite large).

1. This bit of information is redundant, largely kept to stay consistent with existing convention of specifying a mode parameter in the `quantize()` call.
2. You don't need both at the same time. When quantizing a full-precision model you provide the `GPTQConfig`, which provides all the necessary information required to (i) Build the layer variables (ii) Run the GPTQ quantization process. However, when loading a serialized model, the config object is not present. In that case we use the information serialized as a part of the `GPTQDTypePolicy` to restore the layer. The benefit of this approach is that KerasHub models already have a sophisticated dtype propagation mechanism in place, which ensures each layer gets restored with the correct (quantized) dtype.

keras/src/layers/core/dense_test.py
JyotinderSingh: Added an assertion for this.

keras/src/layers/core/dense.py
james77777778: The variables are built or modified in two ways when considering quantization, which is the reason:
1. Directly built from the `__init__` using `dtype` (an instance of `DTypePolicyMap`).
2. Modified during a call to `quantize` in `keras.Model`

The order of variables from these two methods is different. That's why we need to sort the variables within `load|save_own_variables`.

hertschuh: But if we kept the string keys in the dictionary, like `"kernel_scale"`, the order wouldn't matter.

keras/src/dtype_policies/dtype_policy.py
```diff
@@ -288,6 +288,55 @@ def get_config(self):
...
 
+@keras_export("keras.dtype_policies.GPTQDTypePolicy")
+class GPTQDTypePolicy(QuantizedDTypePolicy):
+    """Quantized dtype policy for GPTQ quantization.
+
+    This policy helps propagate quantization settings for GPTQ
+    when loading a GPTQ quantized model in Keras format.
+
+    Args:
+        mode: The quantization mode, "gptq".
```
hertschuh: You need to document the "qptq/b/g" format.

keras/src/dtype_policies/dtype_policy.py
```diff
@@ -288,6 +288,55 @@ def get_config(self):
...
+        source_name: The source dtype policy name, e.g. "float32".
+        weight_bits: Number of bits to quantize weights to. Supported values
```
hertschuh: Remove weight_bits as a standalone arg.

keras/src/dtype_policies/dtype_policy.py
```diff
@@ -288,6 +288,55 @@ def get_config(self):
...
+            are 2, 3, 4, and 8.
+        group_size: The group size for quantization. Supported values are
```
hertschuh: Remove group_size as a standalone arg.

keras/src/dtype_policies/dtype_policy.py
```diff
@@ -288,6 +288,55 @@ def get_config(self):
...
+            Typically a smaller group size leads to better accuracy but
+            slower speed.
+    """
+
+    def __init__(
+        self,
+        mode,
+        source_name=None,
+    ):
+        mode, weight_bits, group_size = mode.split("/")
```
hertschuh: Verify the format follows the expected format and raise a more specific ValueError if it doesn't. Right now, it will throw an error that's pretty obtuse.

keras/src/dtype_policies/dtype_policy.py
```diff
@@ -288,6 +288,55 @@ def get_config(self):
...
+        super().__init__(
+            mode=mode,
+            source_name=source_name,
+        )
+
+        self._name = f"{mode}/{weight_bits}/{group_size}_from_{source_name}"
+        self.mode = mode
+        self.weight_bits = int(weight_bits)
+        self.group_size = int(group_size)
```
hertschuh: The int casting should be part of the validation from my previous comment, you can probably have 1 error message for all the possible errors.

keras/src/layers/core/dense.py
```diff
@@ -352,6 +408,65 @@ def _int8_build(self, kernel_shape):
             trainable=False,
         )
 
+    def _gptq_build(self, kernel_shape, config):
+        self.gptq = False
```
hertschuh: It's very unintuitive that the result of `_gptq_build()` would be to "turn off" gptq.

Either we need a better name than `self.gptq` here, or at least a comment explaining.

keras/src/layers/core/dense.py
```diff
@@ -661,7 +778,12 @@ def quantize(self, mode, type_check=True):
         if self.dtype_policy.quantization_mode is None:
             from keras.src import dtype_policies  # local import to avoid cycle
 
-            policy = dtype_policies.get(f"{mode}_from_{self.dtype_policy.name}")
+            policy_name = mode
+            if mode == "gptq":
+                policy_name = config.dtype_policy_string()
+            policy = dtype_policies.get(
+                f"{policy_name}_from_{self.dtype_policy.name}"
```
hertschuh, line 790: That will give you `gptq_from_...`. Don't you need the other parameters? `qptq/.../..._from_...`?

keras/src/dtype_policies/dtype_policy.py
JyotinderSingh: Done.

keras/src/dtype_policies/dtype_policy.py
JyotinderSingh: Done.

keras/src/dtype_policies/dtype_policy.py
JyotinderSingh: Done.

keras/src/dtype_policies/dtype_policy.py
JyotinderSingh: Done.

keras/src/dtype_policies/dtype_policy.py
JyotinderSingh: Done.

keras/src/layers/core/dense.py
JyotinderSingh: The line right above this generates the policy name from the config object in the correct format.
```py
if mode == "gptq":
    policy_name = config.dtype_policy_string()
```

keras/src/layers/core/dense.py
JyotinderSingh: Renamed it to `is_gptq_calibrated`. Added a comment for explanation too.

JyotinderSingh (commented):
Thanks for the reviews @hertschuh! I've addressed your comments.

keras/src/layers/core/dense.py
james77777778: The reason for using integers is to maintain consistency with other layers. If we change the keys, older models will break.

> But if we kept the string keys in the dictionary, like `"kernel_scale"`, the order wouldn't matter.

This is true, and I agree that this is a better solution for the saving/loading process of quantization.
Maybe we can find a way to do this while remaining compatible with old models?

hertschuh: Sounds good. @JyotinderSingh can you add a line your in quantization document about this?

hertschuh (approved):

PR #21646, opened 2025-09-09
Fix Torch ONNX export not respecting InputSpec.name

### LINKED ISSUE #21637: Keras ONNX export not respecting input node name
When using `torch` backend and exporting to `onnx`, it doesn't seem to respect the `name` of the `InputSpec`.

```python
import os
os.environ["CUDA_VISIBLE_DEVICES"] = ""

import numpy as np
import keras

inputs = keras.Input(shape=(10, 16))
x = keras.layers.MultiHeadAttention(num_heads=2, key_dim=16)(inputs, inputs)
outputs = keras.layers.Dense(8)(x)
model = keras.Model(inputs=inputs, outputs=outputs)

model.compile(optimizer="adam", loss="mse")

X_dummy = np.random.rand(5, 10, 16).astype(np.float32)
y_dummy = np.random.rand(5, 10, 8).astype(np.float32)

model.fit(X_dummy, y_dummy, epochs=1, verbose=1)

spec = [keras.InputSpec(name="input", shape=(1, 10, 16), dtype="float32")]
model.export("test.onnx", format="onnx", verbose=False, input_signature=spec)
```

Produces the following names:

Image

Flakes342:
Fixes #21637 

### Summary
Currently, when exporting a Keras model with the Torch backend to ONNX, the provided InputSpec.name is ignored. Instead, inputs are saved as generic names such as args_0, which makes exported models harder to use and integrate with downstream pipelines.

This PR updates the Torch backend export logic to properly propagate InputSpec.name into torch.onnx.export via the input_names argument. If a name is not provided, it gracefully falls back to the existing default (input_i). 

### Example
Example

```
spec = [keras.InputSpec(name="input", shape=(1, 10, 16), dtype="float32")]
model.export("test.onnx", format="onnx", input_signature=spec)
```

Before:
ONNX model input name → args_0

After:
ONNX model input name → input

Please let me know what else is needed on this..

### DIFF
--- a/keras/src/export/onnx.py
+++ b/keras/src/export/onnx.py
@@ -86,8 +86,6 @@ def export_onnx(
         if name is None:
             name = f"input_{i}"
         input_names.append(name)
-else:
-    input_names = [f"input_{i}" for i in range(len(model.inputs))]
 
     if backend.backend() in ("tensorflow", "jax"):
         from keras.src.utils.module_utils import tf2onnx
--- a/keras/src/export/onnx.py
+++ b/keras/src/export/onnx.py
@@ -166,7 +166,7 @@ def export_onnx(
                     filepath,
                     verbose=actual_verbose,
                     opset_version=opset_version,
-                    input_names=input_names
+                    input_names=input_names,
                 )
     else:
         raise NotImplementedError(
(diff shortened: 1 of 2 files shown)


fchollet (commented):
Thanks for the PR!

Please fix the code format and add a unit test for the code fix.

Flakes342:
Hi @fchollet, I think this PR is ready to be reviewed and merged now. It fixes the InputSpec functionality. I have formatted the code properly and also added a unit test for the same.

Issue #21648, opened 2025-09-09
Issue in reduce_shape in keras/src/ops/operation_utils.py


miguelteixeiragomes:
`keras.ops.any` and `keras.ops.all` seem to be the only operators in latest docs that specifically say _axis may be negative_. However, looking through the code, both these ops and other reduction ops (like `keras.ops.amax`, for example) use the same utility function to calculate the output shape: `reduce_shape` in `keras/src/ops/operation_utils.py`.

I think this function does not handle lists/tuples of negative indexes correctly. The following example shows that, running in keras 3.11.3
```python
import keras
x = keras.Input((3, None, None, 1280))

max_x = keras.ops.amax(x, axis = (2, 3))
assert max_x.shape == (None, 3, 1280)  # passes

max_x = keras.ops.amax(x, axis = (-2, -3))
assert max_x.shape == (None, 3, 1280)  # does not pass, despite being equivalent
```

I believe this can be solved in the `reduce_shape` function by first doing something like `axis = [ax % len(shape) for ax in axis]`.
I can submit a pull request for this later on (I never contributed to keras, so it could be a nice initiation)

Let me know if there's important info that's missing, I did not find a template for issues that I should follow.

pctablet505:
looks like issue is fixed

PR #21650, opened 2025-09-09
fix: use canonicalize_axis function in reduce_shape function

miguelteixeiragomes:
Pull request to fix issue #21648


mgomes0:
@fchollet , sure, added a test with multiple negative axes, but without keepdims. Let me know if you think more tests are needed.
~~There is a test failing related to quantization, but I don't understand if there's any relation to the test I added.~~
Never mind this part, after a re-run if passed.
Thank you!

fchollet (approved):

PR #21656, opened 2025-09-10
Fixes int64 casting bug for ops.take with `keras.Variable` on TF backend

JyotinderSingh:
## Description of the change
For some reason when using a Keras Variable of type `int64`, TF's casting behavior goes bonkers. This adds an explicit cast to avoid a runtime exception of mismatched dtypes.

The existing code would lead to the following state at the `tf.where(...)` call site (the modified code):
```
>>> i.dtype
int64
>>> tf.shape(x)[axis].dtype
int32
>>> tf.cast(tf.shape(x)[axis], i.dtype).dtype
int64
>>> (i + tf.cast(tf.shape(x)[axis], i.dtype).dtype
int32 
```

As shown above, the addition operation weirdly ends up with `int32` dtype. This change explicitly casts the result to a 64 bit integer to avoid a dtype mismatch between the two arguments of `tf.where`

### DIFF
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -2403,7 +2403,9 @@ def take(x, indices, axis=None):
 
     def fix_negative_indices(i):
         # Correct the indices using "fill" mode which is the same as in jax
-        return tf.where(i < 0, i + tf.cast(tf.shape(x)[axis], i.dtype), i)
+        return tf.where(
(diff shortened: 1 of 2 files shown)


JyotinderSingh:
The underlying issue here was discovered to be related to https://github.com/keras-team/keras/issues/21677.

JyotinderSingh:
Fixed in https://github.com/keras-team/keras/pull/21679

PR #21672, opened 2025-09-17
Use JAX DimExpr for dynamic shapes in `compute_output_spec`.

hertschuh:
Previously, with the JAX backend, when tracing in `backend.compute_output_spec`, we would replace the `None` dimensions with 83 then 89 and infer dynamic dimensions in the output by finding the dimensions which changed.

JAX has a mechanism for tracing with dynamic dimensions, which has been exposed publicly for about a year. While the `_DimExpr` class itself is not public, one can create instances of it by using `jax.export.symbolic_shape()`.

Also replaced `jax.make_jaxpr` with `jax.eval_shape`.

## Is it fully backwards compatible?

Not strictly. Some dimensions in some contexts will now be `_DimExpr` instead of ints. But:
- You can do math on `_DimExpr`, which makes it seamless, you don't even see you're not handling ints.
- The export to TF workflow already used `_DimExpr`, so it's not completely new. In fact there was already code in `backend.standardize_shape` to handle `_DimExpr`.

## Example

Here is a theoretical example where the input size `(batch_size, width, height)` is fully dynamic and the output size is `(batch_size, width + 2, height + 2)`.

```python
class PaddedEye(keras.layers.Layer):
    def call(self, x):
        print("in shape  >", x.shape)

        shape = keras.ops.shape(x)
        y = keras.ops.eye(shape[1] + 2, shape[2] + 2)
        y = keras.ops.expand_dims(y, axis=0)
        y = keras.ops.broadcast_to(y, (shape[0], shape[1] + 2, shape[2] + 2))

        print("out shape <", y.shape)
        return y


inputs = keras.layers.Input((None, None), dtype="float32")
outputs = PaddedEye()(inputs)
model = keras.models.Model(inputs=inputs, outputs=outputs)
```

### Before
```
in shape  > (83, 83, 83)
out shape < (83, 85, 85)
in shape  > (89, 89, 89)
out shape < (89, 91, 91)
```

### After
```
in shape  > (dynamic_dimension, dynamic_dimension, dynamic_dimension)
out shape < (dynamic_dimension, dynamic_dimension + 2, dynamic_dimension + 2)
```

### What didn't change

```python
model(keras.ops.ones((2, 3, 3), dtype="float32"))
```

```
in shape  > (2, 3, 3)
out shape < (2, 5, 5)
```

```python
model.export("test_model.tf")
```

```
in shape  > (batch, a, b)
out shape < (batch, a + 2, b + 2)
...
```

`batch`, `a` and `b` are `_DimExpr`.

## Implementations notes

- We create a `_DimExpr` instance for one dimension by creating a symbolic shape with one dimension and extracting it.
- We create a single dynamic dimension and reuse it instead of creating N dynamic dimensions. This is for backwards compatibility. Previously we would fill all dynamic dimensions with the same concrete value. This can handle the case where there is an implicit assumption that two dimensions are the same (e.g. square images).
- We add the constraint "dynamic_dimension>=2" to prevent JAX from assuming that the dimension can be broadcastable or squeezable. It removes this ambiguity.

## Future

I'm planning on writing a keras.io guide explaining shapes in details:
- `x.shape` vs. `keras.ops.shape`
- dynamic shapes with TF vs. JAX (recompilation etc.)
- writing multi-backend code that works with dynamic shapes

### DIFF
--- a/keras/src/backend/common/variables.py
+++ b/keras/src/backend/common/variables.py
@@ -599,6 +599,12 @@ def standardize_shape(shape):
                 shape = shape.as_list()
         shape = tuple(shape)
 
+    if config.backend() == "jax":
+        # Replace `_DimExpr` (dimension expression) with None
+        shape = tuple(
+            [None if "_DimExpr" in str(type(d)) else d for d in shape]
+        )
+
     if config.backend() == "torch":
         # `shape` might be `torch.Size`. We need to convert the items in it to
         # either int or `None`
@@ -606,9 +612,6 @@ def standardize_shape(shape):
 
     for e in shape:
         if e is None:
-            continue
-        if config.backend() == "jax" and "_DimExpr" in str(type(e)):
-            # JAX2TF tracing uses JAX-native dimension expressions
             continue
         if not is_int_dtype(type(e)):
             raise ValueError(
--- a/keras/src/backend/jax/core.py
+++ b/keras/src/backend/jax/core.py
@@ -3,6 +3,7 @@ import jax.experimental.sparse as jax_sparse
 import jax.numpy as jnp
 import ml_dtypes
 import numpy as np
+from jax import export as jax_export
 
 from keras.src import tree
 from keras.src.backend import config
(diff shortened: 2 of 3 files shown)


mattdangerw (approved):
Thanks!

PR #21673, opened 2025-09-17
Non cumulative model.predict

### LINKED ISSUE #21642: Request for non-cumulative prediction options
The story:
I am facing an issue with large data that using model.predict + a result off-loading callback (to write the results to disk) is giving a much better performance compared to looping over batches and using model.predict_on_batch() or model(input, training=False) with offloading operation. However, model.predict is accumulating the results and causes out-of-memory error. 

The request
I was wondering if you could add an argument to model.predict that would prevent it from accumulating the model predictions, allowing one to use a callback to handle the results.

pctablet505:
there was a feature request for non cumulative model.predict, where we don't store the outputs and return it directly.
#21642 In case of very large datasets and limited memory, it is more feasible to retrieve the outputs using a callback and store to storage, instead of storing everything in memory.

This PR fixes #21642.

### DIFF
--- a/keras/src/backend/jax/trainer.py
+++ b/keras/src/backend/jax/trainer.py
@@ -694,7 +694,7 @@ class JAXTrainer(base_trainer.Trainer):
             return outputs
 
         self._jax_state_synced = True
-        outputs = None if accumulate else []
+        outputs = None
         non_trainable_variables = None
         with epoch_iterator.catch_stop_iteration():
             for begin_step, end_step, iterator in epoch_iterator:
@@ -733,9 +733,10 @@ class JAXTrainer(base_trainer.Trainer):
         callbacks.on_predict_end()
         self._jax_state = None
         if accumulate:
+            if outputs is None:
+                return None
             return tree.map_structure_up_to(batch_outputs, np.concatenate, outputs)
-        else:
-            return None
+        return outputs
 
     def train_on_batch(
         self,
--- a/keras/src/backend/jax/trainer.py
+++ b/keras/src/backend/jax/trainer.py
@@ -638,7 +638,7 @@ class JAXTrainer(base_trainer.Trainer):
 
     @traceback_utils.filter_traceback
     def predict(
-        self, x, batch_size=None, verbose="auto", steps=None, callbacks=None
+        self, x, batch_size=None, verbose="auto", steps=None, callbacks=None, accumulate=True
     ):
         # Create an iterator that yields batches of input data.
         epoch_iterator = JAXEpochIterator(
@@ -694,7 +694,7 @@ class JAXTrainer(base_trainer.Trainer):
             return outputs
 
         self._jax_state_synced = True
-        outputs = None
+        outputs = None if accumulate else []
         non_trainable_variables = None
         with epoch_iterator.catch_stop_iteration():
             for begin_step, end_step, iterator in epoch_iterator:
@@ -718,7 +718,8 @@ class JAXTrainer(base_trainer.Trainer):
                     # during predict(), but it's allowed.
                     "non_trainable_variables": non_trainable_variables,
                 }
-                outputs = append_to_outputs(batch_outputs, outputs)
+                if accumulate:
+                    outputs = append_to_outputs(batch_outputs, outputs)
 
                 # Dispatch callbacks. This takes care of async dispatch.
                 callbacks.on_predict_batch_end(
@@ -731,7 +732,10 @@ class JAXTrainer(base_trainer.Trainer):
         self.jax_state_sync()
         callbacks.on_predict_end()
         self._jax_state = None
-        return tree.map_structure_up_to(batch_outputs, np.concatenate, outputs)
+        if accumulate:
+            return tree.map_structure_up_to(batch_outputs, np.concatenate, outputs)
+        else:
+            return None
 
     def train_on_batch(
         self,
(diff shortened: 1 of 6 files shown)


fchollet (commented):
Thanks for the PR!

Can you explain the advantages of calling `predict()` (with a callback) if it returns nothing? Why not simply call `predict(batch)` or `predict_on_batch` in a loop?

pctablet505:
@fchollet there was a feature request from @ErfanMowlaei in the issue #21642.
According to him, using for loop over batches with `predict_on_batch` is much slower compared to `predict` with a callback to get the output.
I verified the performance difference in the [colab](https://colab.research.google.com/gist/pctablet505/987ff6b69c7adce8887fd9e5919c59d5/non-cumulative-model-predict.ipynb).
It might be due to passing control to python, instead of using highly optimized dataset iterator from backends.
other reason might be, less usage of memory transfer from RAM to Storage (virtual memory), as we are using lesser memory.

@ErfanMowlaei 
could you explain the performance difference?

ErfanMowlaei:
@pctablet505 I am not sure about the cause, but calling predict_on_batch was certainly slower without any other processing, and around 2 times slower.

Just as a note, pytorch lightning provides a similar approach to mine using BasePredictionWriter to handle out-of-core prediction handling in a callback to keep the code clean compared to writing a for loop using predict_on_batch.

@fchollet I started my deep learning journey with your book (first edition). A big fan!

innat:
@pctablet505 Does the timing effect same between `model.test_on_batch` and `model.predict_on_batch`?

ErfanMowlaei:
> @pctablet505 Does the timing effect same between `model.test_on_batch` and `model.predict_on_batch`?

There was a problem with that as @pctablet505 also tested here: 
https://colab.research.google.com/gist/pctablet505/cbf7c3f97d3e404b5786da6b57fa4e42/non-cumulative-model-predict.ipynb

innat:
In the Colab, only `.predict_on_batch` is being tested, right? I was curious about `test_on_batch` as well. Also, was there any mention of why `.predict_on_batch` is significantly slower? It might be better to identify and fix the underlying issue if there’s a bug.

ErfanMowlaei:
I would assume the speed difference is because of the graph execution.

hertschuh:
> I would assume the speed difference is because of the graph execution.

@ErfanMowlaei they are both compiled. If you want to be 100% sure, you can add `model.compile(jit_compile=True)` first.

hertschuh:
> In the Colab, only `.predict_on_batch` is being tested, right? I was curious about `test_on_batch` as well. Also, was there any mention of why `.predict_on_batch` is significantly slower? It might be better to identify and fix the underlying issue if there’s a bug.

@innat @ErfanMowlaei I'm not seeing a significant difference between `predict_on_batch` and `predict` using this example:  https://github.com/keras-team/keras/issues/21642#issuecomment-3268890552

Can you confirm? What hardware and backend were used?

pctablet505:
I compared the test_on_batch also, it is slower than predict_on_batch

pctablet505:
predict_on_batch is faster


```
import gc, time, tracemalloc
import numpy as np
import keras
from keras import layers, models

# ── Small model ----------------------------------------------------------
inp = layers.Input((256,))
h   = layers.Dense(1024, activation="relu")(inp)
out = layers.Dense(4096)(h)
model = models.Model(inp, out)
model.compile()

N, BS = 10_000, 256
x = np.random.rand(N, 256).astype("float32")

# warm up
model.predict(x[:BS*2], batch_size=BS, verbose=0)
for i in range(4): model.predict_on_batch(x[i*BS:(i+1)*BS])

# ── Speed ----------------------------------------------------------------
t0 = time.perf_counter()
model.predict(x, batch_size=BS, verbose=0, accumulate=False)
t_pred = time.perf_counter() - t0

t0 = time.perf_counter()
chunks = [model.predict_on_batch(x[i:i+BS]) for i in range(0, N, BS)]
_ = np.concatenate(chunks)
t_pob = time.perf_counter() - t0

# ── Memory ---------------------------------------------------------------
tracemalloc.start()
model.predict(x, batch_size=BS, verbose=0, accumulate=False)
_, peak_pred = tracemalloc.get_traced_memory(); tracemalloc.stop()

tracemalloc.start()
for i in range(0, N, BS): model.predict_on_batch(x[i:i+BS])  # no concat
_, peak_pob = tracemalloc.get_traced_memory(); tracemalloc.stop()

tracemalloc.start()
model.predict(x, batch_size=BS, verbose=0, accumulate=True)
_, peak_acc = tracemalloc.get_traced_memory(); tracemalloc.stop()

# ── Results --------------------------------------------------------------
print(f"Speed (N={N:,}, output=4096 floats/sample)")
print(f"  predict(accumulate=False) : {t_pred:.3f}s")
print(f"  predict_on_batch loop     : {t_pob:.3f}s  — ratio: {t_pred/t_pob:.2f}x")

print(f"\nPeak memory (tracemalloc)")
print(f"  predict(accumulate=True)  : {peak_acc/1e6:.1f} MB  ← PR default (old behaviour)")
print(f"  predict(accumulate=False) : {peak_pred/1e6:.1f} MB")
print(f"  predict_on_batch loop     : {peak_pob/1e6:.1f} MB  ← already zero-accumulation")
```

```
Speed (N=10,000, output=4096 floats/sample)
  predict(accumulate=False) : 2.645s
  predict_on_batch loop     : 2.534s  — ratio: 1.04x

Peak memory (tracemalloc)
  predict(accumulate=True)  : 164.0 MB  ← PR default (old behaviour)
  predict(accumulate=False) : 14.5 MB
  predict_on_batch loop     : 4.2 MB  ← already zero-accumulation
```

PR #21674, opened 2025-09-17
Model Export to liteRT

pctablet505:
This pull request adds support for exporting Keras models to the LiteRT (TensorFlow Lite) format and refactors the export utilities to improve extensibility and robustness. The changes also enhance documentation and input signature handling for various model types during export.

**LiteRT export support and extensibility:**

* Added `LiteRTExporter` to the export API and integrated it into the export format registry, enabling users to export models using `format="lite_rt"` via the unified `export_model` function. [[1]](diffhunk://#diff-969c214a6faa5ceefc8c79f648c92db366cd102d5105b39d4ec8212ae5cd822bR6) [[2]](diffhunk://#diff-4d200ca4bf4c92ea72238857274a86436d4cd729f5a6f5cb28fc57ecc13cd59aR150-R184)
* Updated the `Model.export` method to support `"lite_rt"` as a valid export format, including detailed docstrings and usage examples for LiteRT export. [[1]](diffhunk://#diff-fb2a676499790af948471633c80c52cbe5fab5dd3b3aecf704776f636eb18263L572-R571) [[2]](diffhunk://#diff-fb2a676499790af948471633c80c52cbe5fab5dd3b3aecf704776f636eb18263R595-R600) [[3]](diffhunk://#diff-fb2a676499790af948471633c80c52cbe5fab5dd3b3aecf704776f636eb18263R635-R654) [[4]](diffhunk://#diff-fb2a676499790af948471633c80c52cbe5fab5dd3b3aecf704776f636eb18263R685-R694)

**Export utility improvements:**

* Refactored `export_utils.py` to introduce a registry and lazy-loading mechanism for export formats, making it easier to add new export formats in the future.
* Improved `get_input_signature` to better handle subclassed and functional models, providing clearer error messages and fallback logic for input signature inference. [[1]](diffhunk://#diff-4d200ca4bf4c92ea72238857274a86436d4cd729f5a6f5cb28fc57ecc13cd59aR10-R17) [[2]](diffhunk://#diff-4d200ca4bf4c92ea72238857274a86436d4cd729f5a6f5cb28fc57ecc13cd59aR28-R44)

**Input signature and spec handling:**

* Enhanced `_make_input_spec` to properly handle batch dimensions and preserve input structure for both single and multiple inputs, improving compatibility with different model types.


keras/src/export/export_utils.py
```diff
@@ -105,3 +147,38 @@ def convert_spec_to_tensor(spec, replace_none_number=None):
             s if s is not None else replace_none_number for s in shape
         )
     return ops.ones(shape, spec.dtype)
+
+
+# Registry for export formats
+EXPORT_FORMATS = {
+    "tf_saved_model": "keras.src.export.saved_model:export_saved_model",
+    "lite_rt": "keras.src.export.lite_rt_exporter:LiteRTExporter",
```
pctablet505: Shall it be named 'lite_rt' or 'litert'?

keras/src/models/model.py
```diff
@@ -627,12 +632,26 @@ def export(
...
+        interpreter.invoke()
+        output_data = interpreter.get_tensor(interpreter.get_output_details()[0]['index'])
+        ```
         """
         from keras.src.export import export_onnx
         from keras.src.export import export_openvino
         from keras.src.export import export_saved_model
 
-        available_formats = ("tf_saved_model", "onnx", "openvino")
+        available_formats = ("tf_saved_model", "onnx", "openvino", "lite_rt")
```
pctablet505: `tflite` and `lite_rt` both may be supported as both generate same format `tflite` but `lite_rt` is supposed to be further optimized.

keras/src/export/export_utils.py
```diff
@@ -105,3 +147,38 @@ def convert_spec_to_tensor(spec, replace_none_number=None):
             s if s is not None else replace_none_number for s in shape
         )
     return ops.ones(shape, spec.dtype)
+
+
+# Registry for export formats
```
mattdangerw: Let's clean up the code below here in this file. I don't think we need any of it. Instead add we can just directly handle calling a lite rt export function in `model.export()` in mode.py.

keras/src/export/export_utils.py
```diff
@@ -17,19 +25,24 @@ def get_input_signature(model):
...
+        # For subclassed models, try multiple approaches
         input_signature = _infer_input_signature_from_model(model)
-        if not input_signature or not model._called:
-            raise ValueError(
-                "The model provided has never called. "
-                "It must be called at least once before export."
-            )
+        if not input_signature:
+            # Fallback: Try to get from model.inputs if available
+            if hasattr(model, 'inputs') and model.inputs:
```
mattdangerw: can you explain what case you are trying to cover here? when will a model have `inputs` but _infer_input_signature_from_model won't work?

keras/src/export/export_utils.py
```diff
@@ -45,22 +58,51 @@ def _make_input_spec(structure):
             return {k: _make_input_spec(v) for k, v in structure.items()}
         elif isinstance(structure, tuple):
             if all(isinstance(d, (int, type(None))) for d in structure):
+                # Keep batch dimension unbounded, keep other dimensions as they are
+                bounded_shape = []
```
mattdangerw: ditto here, what's the case you are trying to address?

keras/src/export/litert_exporter.py
```diff
@@ -0,0 +1,452 @@
+import tensorflow as tf
+from keras.src.export.export_utils import get_input_signature, make_tf_tensor_spec
+import tempfile
+import os
+import numpy as np
+
+# Try to import LiteRT AOT compilation if available
+try:
```
mattdangerw: I think the way to do this would be to add litert to [module utils](https://github.com/keras-team/keras/blob/master/keras/src/utils/module_utils.py), import that regularly above, and then just path from the toplevel object below (e.g. `litert.python.aot.aot_compile`). Would that work?

keras/src/export/lite_rt_exporter.py
```diff
@@ -0,0 +1,452 @@
...
+try:
+    from litert.python.aot import aot_compile
+    from litert.python.aot.core import types as litert_types
+    from litert.python.aot.vendors import import_vendor
+    LITERT_AVAILABLE = True
+except ImportError:
+    LITERT_AVAILABLE = False
+
+
+class LiteRTExporter:
```
mattdangerw: fine to have a class here if it's helpful, but make sure to expose a toplevel function like `export_openvino` that creates and uses the class.

keras/src/export/lite_rt_exporter.py
```diff
@@ -0,0 +1,452 @@
...
+    """
+    Exporter for the LiteRT (TFLite) format that creates a single,
+    callable signature for `model.call`.
+    """
+
+    def __init__(self, model, input_signature=None, verbose=0,
```
mattdangerw: please make sure to format your code

keras/src/export/lite_rt_exporter.py
```diff
@@ -0,0 +1,452 @@
...
+    """
+
+    def __init__(self, model, input_signature=None, verbose=0, 
+                 aot_compile_targets=None, **kwargs):
+        """Initialize the LiteRT exporter.
+        
+        Args:
+            model: The Keras model to export
+            input_signature: Input signature specification
+            verbose: Verbosity level (0=quiet, 1=info)
```
mattdangerw: handle verbose the exact same as other exporters. see onnx https://github.com/keras-team/keras/blob/cdc7e2993e173325cd99132e07b9dd9e0ea84aa3/keras/src/export/onnx.py#L34-L36

keras/src/export/saved_model.py
```diff
@@ -358,6 +356,7 @@ def serving_fn(x):
             self._endpoint_names.append(name)
             return decorated_fn
 
+        from keras.src.export.export_utils import make_tf_tensor_spec
```
mattdangerw: what's motivating this change?

keras/src/export/litert.py
```diff
@@ -0,0 +1,452 @@
+import tensorflow as tf
```
mattdangerw: Make sure the filenames agree with others in this dir. E.g. `onnx.py` and `openvino.py` so `litert.py`. Also add tests (use the other exporters as a model).

keras/src/export/litert_exporter.py
```diff
@@ -0,0 +1,452 @@
...
+        
+        # 3. Convert the model to TFLite.
+        tflite_model = self._convert_to_tflite(self.input_signature)
+        
+        if self.verbose:
+            final_size_mb = len(tflite_model) / (1024*1024)
+            print(f"TFLite model converted successfully. Size: {final_size_mb:.2f} MB")
+        
+        # 4. Save the initial TFLite model to the specified file path.
+        if not filepath.endswith('.tflite'):
```
mattdangerw: Again we want consistency with other exporters. We error for the wrong file extension elsewhere, so we should do the same here. https://github.com/keras-team/keras/blob/cdc7e2993e173325cd99132e07b9dd9e0ea84aa3/keras/src/export/openvino.py#L52-L55

keras/src/export/litert_exporter.py
```diff
@@ -0,0 +1,452 @@
...
+            raise ValueError(f"Failed to trace model with error: {e}")
+
+        # Final, critical check
+        if not self.model.built:
+            raise ValueError(
+                "Model could not be built even after a direct call. "
+                "Please check the model's definition and input specification."
+            )
+
+    def _infer_sequential_input_shape(self):
```
mattdangerw: This looks incredibly brittle, i think we should remove this whole function. Take some time to read sequential.py and understand how input shape is handled. It looks to me like after `build()` we can be guaranteed that the first layer is an input layer. https://github.com/keras-team/keras/blob/cdc7e2993e173325cd99132e07b9dd9e0ea84aa3/keras/src/models/sequential.py#L176-L188

mattdangerw (commented):
Left some initial comments.

General notes:
* Pay close attention to how other exporters work and try to follow the same UX exactly.
* If we are doing anything that involves enumerating layer types in Keras, let's think hard about how to avoid that and make the code less brittle.

pctablet505:
@gemini-code-assist

keras/src/export/export_utils.py
pctablet505: named it as "litert"

pctablet505:
@gemini-code-assist

keras/src/models/model.py
```diff
@@ -593,6 +592,18 @@ def export(
                     provided, they will be automatically computed.
                 - `opset_version`: Optional `int`. Specific to `format="onnx"`.
                     An integer value that specifies the ONNX opset version.
+                - `allow_custom_ops`: Optional `bool`. Specific to
```
fchollet: Maybe we should consider putting all the litert args in a single dict to simplify accounting? `litert_kwargs`.

keras/src/export/__init__.py
```diff
@@ -3,3 +3,12 @@
 from keras.src.export.saved_model import ExportArchive
 from keras.src.export.saved_model import export_saved_model
 from keras.src.export.tfsm_layer import TFSMLayer
+
+# LiteRT export requires TensorFlow, so we import conditionally
+try:
+    from keras.src.export.litert import LitertExporter
```
mattdangerw: avoid the conditional import, and use the module_utils version of tensorflow below where you need it

keras/src/export/export_utils.py
```diff
@@ -41,25 +56,36 @@ def _infer_input_signature_from_model(model):
     def _make_input_spec(structure):
         # We need to turn wrapper structures like TrackingDict or _DictWrapper
         # into plain Python structures because they don't work with jax2tf/JAX.
+        if structure is None:
```
mattdangerw: when is structure none?

keras/src/export/litert.py
```diff
@@ -0,0 +1,474 @@
+import os
+
+import tensorflow as tf
```
mattdangerw: mentioned above, but you can use `from keras.src.utils.module_utils import tensorflow as tf` to avoid the need for conditional imports above

keras/src/export/litert.py
```diff
@@ -0,0 +1,474 @@
...
+
+        Args:
+            filepath: Output path for the exported model
+
+        Returns:
+            Path to exported model or compiled models if AOT compilation is
+            performed
+        """
+        if self.verbose:
+            print("Starting Litert export...")
```
mattdangerw: i think you might want print_msg for all the prints in this file https://github.com/keras-team/keras/blob/3137cb0473c9463bc3008f634dbeb7bc5e93d000/keras/src/callbacks/progbar_logger.py#L55

keras/src/export/litert.py
```diff
@@ -0,0 +1,474 @@
...
+        # available
+        compiled_models = None
+        if self.aot_compile_targets and litert.available:
+            if self.verbose:
+                print("Performing AOT compilation for Litert targets...")
+            compiled_models = self._aot_compile(filepath)
+        elif self.aot_compile_targets and not litert.available:
+            if self.verbose:
+                print(
+                    "Warning: AOT compilation requested but LiteRT is not "
```
mattdangerw: here you might want `logging.warning` https://github.com/keras-team/keras/blob/3137cb0473c9463bc3008f634dbeb7bc5e93d000/keras/src/legacy/saving/saving_utils.py#L249-L253

keras/src/models/model.py
```diff
@@ -593,6 +592,17 @@ def export(
                     provided, they will be automatically computed.
                 - `opset_version`: Optional `int`. Specific to `format="onnx"`.
                     An integer value that specifies the ONNX opset version.
+                - `litert_kwargs`: Optional `dict`. Specific to
```
mattdangerw: This seems a bit duplicated. We already have a general passing of kwargs passed to this function to the specific export. Maybe we should just the per format kwargs that are currently supported?

keras/src/export/litert_test.py
```diff
@@ -0,0 +1,467 @@
...
+
+def _get_interpreter_outputs(interpreter):
+    output_details = interpreter.get_output_details()
+    outputs = [
+        interpreter.get_tensor(detail["index"]) for detail in output_details
+    ]
+    return outputs[0] if len(outputs) == 1 else outputs
+
+
+@pytest.mark.skipif(
```
mattdangerw: do we need both `tensorflow.available` and `backend.backend() != "tensorflow"` checks? seems like you could get away with just the latter.

keras/src/export/litert_test.py
```diff
@@ -0,0 +1,467 @@
...
+    not tensorflow.available,
+    reason="TensorFlow is required for LiteRT export tests.",
+)
+@pytest.mark.skipif(
+    backend.backend() != "tensorflow",
+    reason="`export_litert` currently supports the TensorFlow backend only.",
+)
+@pytest.mark.skipif(
+    testing.tensorflow_uses_gpu(),
+    reason="LiteRT export tests are only run on CPU to avoid CI issues.",
```
mattdangerw: What are the issues running on GPU? Working through those is probably important

keras/src/export/litert.py
```diff
@@ -0,0 +1,474 @@
...
+            # Fall back to model.inputs for Functional/Sequential models
+            elif hasattr(self.model, "inputs") and self.model.inputs:
+                input_shapes = [inp.shape for inp in self.model.inputs]
+                if len(input_shapes) == 1:
+                    self.model.build(input_shapes[0])
+                else:
+                    self.model.build(input_shapes)
+            else:
+                raise ValueError(
+                    "Cannot build model: no input_signature provided and "
```
mattdangerw: Try to write these so they are a little more understandable for a user with no context. Here something like

```
Cannot export model to the litert format as the input_signature could not be inferred.
Either pass an `input_signature` to `model.export()` or ensure that the model is
already built (called once on real inputs).
```

keras/src/export/litert.py
```diff
@@ -0,0 +1,474 @@
...
+        if self.verbose:
+            print("Building model before conversion...")
+
+        try:
+            # Try to build using input_signature if available
+            if self.input_signature:
+                input_shapes = tree.map_structure(
+                    lambda spec: spec.shape, self.input_signature
+                )
+                self.model.build(input_shapes)
```
mattdangerw: I'm not sure we want this? It looks to me like tf saved model export expects the model to be built

https://github.com/keras-team/keras/blob/3137cb0473c9463bc3008f634dbeb7bc5e93d000/keras/src/export/saved_model.py#L151-L154

and onnx export

https://github.com/keras-team/keras/blob/3137cb0473c9463bc3008f634dbeb7bc5e93d000/keras/src/export/onnx.py#L79-L82

We are just going to make things more confusing if one export format attempt to automatically build but no others do. Let's shoot for consistency.

pctablet505: We are using tf_saved_model as intermediate step, to convert to litert. 
We can't expect the model to be fully built/called/traced while calling export.
And making it leave on user, will make the export process complicated from users perspective.

for uniformity, we can change the behaviour in other formats too like onnx.

keras/src/export/litert_test.py
pctablet505: Tests are working even with GPU.
Changed it.

keras/src/export/export_utils.py
pctablet505: This condition was not necessary, removed it.

keras/src/models/model.py
mattdangerw: I think given that every export format already does this through kwargs we should probably be consistent. Either `litert_kwargs`, `onnx_kwargs`, `tf_save_model_kwargs`, etc. Or one final `**kwargs` that is interpreted per format. No strong preference.

keras/src/export/export_utils.py
```diff
@@ -45,21 +60,30 @@ def _make_input_spec(structure):
             return {k: _make_input_spec(v) for k, v in structure.items()}
         elif isinstance(structure, tuple):
             if all(isinstance(d, (int, type(None))) for d in structure):
-                return layers.InputSpec(
-                    shape=(None,) + structure[1:], dtype=model.input_dtype
+                # For export, force batch dimension to None for flexible
+                # batching
+                shape = (
+                    (None,) + structure[1:] if len(structure) > 0 else structure
```
SamanehSaadat: What's changed here is the addition of `if len(structure) > 0 else structure`. Curious to know why we need this here?

keras/src/export/litert.py
```diff
@@ -0,0 +1,497 @@
...
+
+def export_litert(
+    model,
+    filepath,
+    verbose=None,
+    input_signature=None,
+    aot_compile_targets=None,
+    **kwargs,
+):
+    """Export the model as a Litert artifact for inference.
```
SamanehSaadat: I think `LiteRT` is easier to read than `Litert`.

keras/src/export/litert.py
```diff
@@ -0,0 +1,497 @@
...
+
+    Args:
+        model: The Keras model to export.
+        filepath: The path to save the exported artifact.
+        verbose: `bool`. Whether to print a message during export. Defaults to
+            `None`, which uses the default value set by different backends and
+            formats.
+        input_signature: Optional input signature specification. If
+            ``None``, it will be inferred.
```
SamanehSaadat: There seems to be some extra backticks around `None`

keras/src/export/litert.py
```diff
@@ -0,0 +1,497 @@
...
+from keras.src import tree
+from keras.src.utils import io_utils
+from keras.src.utils.module_utils import litert
+from keras.src.utils.module_utils import tensorflow as tf
+
+
+def export_litert(
+    model,
+    filepath,
+    verbose=None,
```
SamanehSaadat: Why not set the default here to `True` so you don't have to set the default `if verbose is None` here: https://github.com/keras-team/keras/pull/21674/files#diff-9cf8ac69bf49a96a776d05beaa53d93d30af5cc8fa8af5ea1511f22268095403R33-R34

keras/src/export/litert.py
```diff
@@ -0,0 +1,497 @@
...
+            formats.
+        input_signature: Optional input signature specification. If
+            ``None``, it will be inferred.
+        aot_compile_targets: Optional list of Litert targets for AOT
+            compilation.
+        **kwargs: Additional keyword arguments passed to the exporter.
+    """
+
+    if verbose is None:
+        verbose = True  # Defaults to `True` for all backends.
```
SamanehSaadat: We can remove this and set the default value of the `verbose` to `True` in the function's signature.

keras/src/export/litert.py
```diff
@@ -0,0 +1,497 @@
...
+    )
+    exporter.export(filepath)
+    if verbose:
+        io_utils.print_msg(f"Saved artifact at '{filepath}'.")
+
+
+class LitertExporter:
+    """
+    Exporter for the Litert (TFLite) format that creates a single,
+    callable signature for `model.call`.
```
SamanehSaadat: Could you make the docstring aligned with Keras' guideline? https://github.com/keras-team/keras/blob/master/CONTRIBUTING.md#docstrings, e.g. it's missing a one-line description of the class.

keras/src/export/litert.py
```diff
@@ -0,0 +1,497 @@
...
+            **kwargs: Additional export parameters
+        """
+        self.model = model
+        self.input_signature = input_signature
+        self.verbose = verbose
+        self.aot_compile_targets = aot_compile_targets
+        self.kwargs = kwargs
+
+    def export(self, filepath):
+        """Exports the Keras model to a TFLite file and optionally performs AOT
```
SamanehSaadat: TFLite -> LiteRT?

keras/src/export/litert.py
```diff
@@ -0,0 +1,497 @@
...
+        # 2. Resolve / infer input signature
+        if self.input_signature is None:
+            if self.verbose:
+                io_utils.print_msg("Inferring input signature from model.")
+            from keras.src.export.export_utils import get_input_signature
+
+            self.input_signature = get_input_signature(self.model)
+
+        # 3. Convert the model to TFLite.
+        tflite_model = self._convert_to_tflite(self.input_signature)
```
SamanehSaadat: `TFLite` is the old name, right? It'd be better to consistently use the new name, `LiteRT`.

keras/src/export/litert.py
```diff
@@ -0,0 +1,497 @@
...
+
+        if self.verbose:
+            final_size_mb = len(tflite_model) / (1024 * 1024)
```
SamanehSaadat: hmm ... what is expected `len(tflite_model)` will return here?

keras/src/export/litert.py
```diff
@@ -0,0 +1,497 @@
...
+                )
+
+            return self._convert_with_wrapper(input_signature)
+
+    def _convert_with_wrapper(self, input_signature):
+        """Converts the model to TFLite using the tf.Module wrapper."""
+
+        # Define the wrapper class dynamically to avoid module-level
+        # tf.Module inheritance
+        class KerasModelWrapper(tf.Module):
```
SamanehSaadat: Could you point me to a design doc or something that explains why we need this wrapper?

keras/src/export/litert.py
```diff
@@ -0,0 +1,497 @@
...
+            are tracked by the SavedModel format, which is key to including
+            them in the final TFLite model.
+            """
+
+            def __init__(self, model):
+                super().__init__()
+                # Store the model reference in a way that TensorFlow won't
+                # try to track it
+                # This prevents the _DictWrapper error during SavedModel
+                # serialization
```
SamanehSaadat: Could you separate these two sentences with a `.` instead of new line?

keras/src/export/litert.py
```diff
@@ -0,0 +1,497 @@
...
+                object.__setattr__(self, "_model", model)
+
+                # Track all variables from the Keras model using proper
+                # tf.Module methods
+                # This ensures proper variable handling for stateful layers
+                # like BatchNorm
```
SamanehSaadat: Same here

keras/src/export/litert.py
```diff
@@ -0,0 +1,497 @@
...
+                )
+                if result.failed_backends:
+                    for backend, error in result.failed_backends:
+                        io_utils.print_msg(
+                            f"  Failed: {backend.id()} - {error}"
+                        )
+
+                # Print compilation report if available
+                try:
+                    report = result.compilation_report()
```
SamanehSaadat: Does this throw an exception if report is not available?

keras/src/export/litert_test.py
```diff
@@ -0,0 +1,458 @@
...
+from keras.src.saving import saving_lib
+from keras.src.testing.test_utils import named_product
+from keras.src.utils.module_utils import litert
+from keras.src.utils.module_utils import tensorflow
+
+# Set up LiteRT interpreter with fallback logic:
+# 1. Try AI Edge LiteRT interpreter (preferred)
+# 2. Fall back to TensorFlow Lite interpreter if AI Edge LiteRT unavailable
+AI_EDGE_LITERT_AVAILABLE = False
+LiteRtInterpreter = None
```
SamanehSaadat: Could you make the `LiteRT` name format consistent everywhere? I think everywhere else it's called `Litert`

keras/src/export/litert_test.py
```diff
@@ -0,0 +1,458 @@
...
+        temp_filepath = os.path.join(
+            self.get_temp_dir(), "exported_model.tflite"
+        )
+        ref_output = _convert_to_numpy(
+            model(tree.map_structure(ops.convert_to_tensor, ref_input))
+        )
+
+        # Test with model.export()
+        model.export(temp_filepath, format="litert")
+        export_path = temp_filepath
```
SamanehSaadat, line 253: Is there any reason that you need a new variable here? can we call it `export_path` from the beginning?

keras/src/export/litert.py
pctablet505: LiteRT is just a runtime built on top of TFLite, it generates same  old `.tflite` file.

keras/src/export/export_utils.py
pctablet505: Removed - it was unnecessary. Empty structures fail earlier at `model.build()`.

keras/src/export/litert.py
pctablet505: `_convert_to_tflite` returns the serialized bytes.
`len(tflite_model)` counts the bytes.

keras/src/export/litert.py
mattdangerw: > We can't expect the model to be fully built/called/traced while calling export.

I think we can right? If tf saved model and onnx expect the model to be built, let's just make the same assumption here. It's a good way to start simple. Basically, we'd expect users to hit this error in some cases. Modify their code to call the model on some inputs, and re-export.

Then we could always add this auto build feature as a follow up right? But do it more consistently across export formats.

keras/src/export/export_utils.py
pctablet505: final name LiteRT

keras/src/models/model.py
pctablet505: format = "litert"

keras/src/models/model.py
pctablet505: for now I decided to pass as it directly in `**kwargs` like below:

```
model.export(
            "gemma_model_quantized.tflite",
            format="litert",
            optimizations=[tf.lite.Optimize.DEFAULT],
            allow_custom_ops=True,
        )
```

in a separate PR we would make it consistent and packed for every format.
```
model.export(
    format="tf_saved_model",
    filepath=filepath,
    saved_model_kwargs={
        "saved_model_kwargs1": None,
        "saved_model_kwargs2": None,
        "saved_model_kwargs3": None,
    },
)
model.export(
    format="litert",
    filepath=filepath,
    litert_kwargs={
        "litert_kwargs1": None,
        "litert_kwargs2": None,
        "litert_kwargs3": None,
    },
)
model.export(
    format="onnx",
    filepath=filepath,
    onnx_kwargs={
        "onnx_kwargs1": None,
        "onnx_kwargs2": None,
        "onnx_kwargs3": None,
    },
)
```

keras/src/export/litert.py
pctablet505: We do need the build for many models. So we would be keeping it.
For consistency, we would make it common for model.export for every format by refactoring before calling specific export function for formats.

keras/src/export/litert.py
pctablet505: It does throw exception, but we are ignoring that as this is not much necessary for now.

keras/src/export/litert.py
pctablet505: [go/keras-litert](http://goto.google.com/keras-litert)

pctablet505 (commented):
I've addressed the comments.
Please review.

keras/src/models/model.py
```diff
@@ -627,18 +634,44 @@ def export(
...
-        available_formats = ("tf_saved_model", "onnx", "openvino")
+        available_formats = ("tf_saved_model", "onnx", "openvino", "litert")
         if format not in available_formats:
             raise ValueError(
                 f"Unrecognized format={format}. Supported formats are: "
                 f"{list(available_formats)}."
             )
 
+        # Check if LiteRT export is available (requires TensorFlow)
+        if format == "litert" and export_litert is None:
```
mattdangerw: this branch can never be hit, as export_litert will never be none. It's imported 10 lines up and will always resolve to a module.

remove this branch, and check that tensorflow is installed inside `export_litert` in the other file.

keras/src/models/model.py
```diff
@@ -593,6 +593,13 @@ def export(
                     provided, they will be automatically computed.
                 - `opset_version`: Optional `int`. Specific to `format="onnx"`.
                     An integer value that specifies the ONNX opset version.
+                - LiteRT-specific options: Optional keyword arguments specific
+                    to `format="litert"`. These are passed directly to the
+                    TensorFlow Lite converter and include options like
+                    `optimizations`, `representative_dataset`,
+                    `experimental_new_quantizer`, `allow_custom_ops`,
+                    `enable_select_tf_ops`, etc. See TensorFlow Lite
+                    documentation for all available options.
```
mattdangerw: where should they go? can we include a link?

pctablet505:
@gemini-code-assist review

keras/src/export/export_utils.py
```diff
@@ -60,6 +75,7 @@ def _make_input_spec(structure):
                 f"Unsupported type {type(structure)} for {structure}"
             )
 
+    # Always return a flat list preserving the order of shapes_dict values
     return [_make_input_spec(value) for value in shapes_dict.values()]
```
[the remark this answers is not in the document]
pctablet505: The suggested change is incorrect - it would break existing functionality by flattening nested structures that should remain nested to match the model's call signature. The current implementation properly handles input signature structures as needed.

pctablet505:
@gemini-code-assist review

keras/src/export/litert.py
```diff
@@ -0,0 +1,548 @@
...
+    def _get_available_litert_targets(self):
+        """Get available LiteRT targets for AOT compilation."""
+        if not litert.available:
+            return []
+
+        try:
+            # Get all registered targets
+            targets = (
+                litert.python.aot.vendors.import_vendor.AllRegisteredTarget()
+            )
```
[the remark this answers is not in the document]
pctablet505: Currently we don't have any public API for this,  so we can't do anything for now.

keras/src/export/__init__.py
```diff
@@ -1,3 +1,5 @@
+from keras.src.export.litert import LiteRTExporter
+from keras.src.export.litert import export_litert
```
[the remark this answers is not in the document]
pctablet505: if backend is set as tensorflow, it is user's responsibility to ensure tensorflow is installed. In other cases, it doesn't cause any issue. 

```
bash-3.2$ KERAS_BACKEND=numpy python3 -c "import keras.src.export; print('Import successful with numpy backend')"
Import successful with numpy backend
bash-3.2$ KERAS_BACKEND=jax python3 -c "import keras.src.export; print('Import successful with jax backend')"
Import successful with jax backend
bash-3.2$ KERAS_BACKEND=torch python3 -c "import keras.src.export; print('Import successful with torch backend')"
Import successful with torch backend```

File: keras/src/export/litert.py (hunk 24 above)
Comment (hertschuh) [reply keras/src/export/litert.py]: Let's not auto-build. Exporting a freshly built model (un-trained, not loaded) with random weights makes no sense. Please remove lines 167-198 and instead throw an error.

File: keras/src/models/model.py [hunk 45]
```diff
@@ -627,18 +634,45 @@ def export(
...
                 f"Unrecognized format={format}. Supported formats are: "
                 f"{list(available_formats)}."
             )
 
+        # Check if LiteRT export is available (requires TensorFlow backend)
+        if format == "litert" and backend.backend() != "tensorflow":
+            raise ImportError(
+                "LiteRT export requires TensorFlow backend. "
+                "Please set the backend to TensorFlow: "
+                "`keras.backend.set_backend('tensorflow')`"
```
hertschuh: Oh, doing `keras.backend.set_backend('tensorflow')` won't solve their problem, it's not going to convert in-memory, it's just going to create errors. Please remove the end of the message, just say this for now.

`"LiteRT export requires TensorFlow backend."`

keras/src/export/litert_test.py
```diff
@@ -0,0 +1,1250 @@
...
+        # Verify outputs match by position
+        sig_output_values = list(sig_output.values())
+        for i, ref_out in enumerate(ref_outputs):
+            self.assertAllClose(
+                ref_out, sig_output_values[i], atol=1e-4, rtol=1e-4
+            )
+
+
+if __name__ == "__main__":
+    pytest.main([__file__])
```
hertschuh: Remove, this is not needed

keras/src/export/litert.py
```diff
@@ -0,0 +1,548 @@
...
+        input_signature=None,
+        verbose=False,
+        aot_compile_targets=None,
+        **kwargs,
+    ):
+        """Initialize the LiteRT exporter.
+
+        Args:
+            model: The Keras model to export
+            input_signature: Input signature specification
```
hertschuh: `input_signature: Input signature specification` in what format? TF TensorSpec?

keras/src/export/litert.py
```diff
@@ -0,0 +1,548 @@
...
+                "properly defined or provide an input_signature."
+            )
+
+    def _convert_to_tflite(self, input_signature):
+        """Converts the Keras model to TFLite format.
+
+        Returns:
+            A bytes object containing the serialized TFLite model.
+        """
+        is_sequential = isinstance(self.model, tf.keras.Sequential)
```
hertschuh: Remove `is_sequential`. It's confusing because subclass models are going to be logged as "Functional", which is incorrect.

keras/src/export/litert.py
```diff
@@ -0,0 +1,548 @@
...
+
+        # Try direct conversion first for all models
+        try:
+            if self.verbose:
+                model_type = "Sequential" if is_sequential else "Functional"
+                io_utils.print_msg(
+                    f"{model_type} model detected. Trying direct conversion..."
+                )
```
hertschuh: Remove. Keras in general has almost no console output in the success case. I don't think this message brings any value to the user.

keras/src/export/litert.py
```diff
@@ -0,0 +1,548 @@
...
+    embedded devices, it creates a single callable signature based on the
+    model's `call()` method and supports optional Ahead-of-Time (AOT)
+    compilation for specific hardware targets.
+    """
+
+    def __init__(
+        self,
+        model,
+        input_signature=None,
+        verbose=False,
```
hertschuh: Remove `verbose`. This was more of a debugging tool. 

Keras in general has almost no console output in the success case. I don't think the messages brings any value to the user. Either it succeeds, and the message printed are meaningless (what is direct conversion vs. wrapper conversion??), or it fails and then the error message should be very detailed.

So remove all the `if self.verbose: print_msg` except in the case when dicts were turned to lists.

keras/src/export/litert.py
```diff
@@ -0,0 +1,548 @@
...
+            input_signature: Input signature specification
+            verbose: Whether to print progress messages during export.
+            aot_compile_targets: List of LiteRT targets for AOT compilation
+            **kwargs: Additional export parameters
+        """
+        self.model = model
+        self.input_signature = input_signature
+        self.verbose = verbose
+        self.aot_compile_targets = aot_compile_targets
+        self.kwargs = kwargs
```
hertschuh, line 59: Just for clarity, let's call it `self.tf_lite_converter_kwargs`.

Also, it's very unfortunate that it's not validated right here. If it fails later, you're no longer in `__init__` and the connection between `__init__` and the invalid kwarg is unclear.

keras/src/export/litert.py
```diff
@@ -0,0 +1,548 @@
...
+            io_utils.print_msg(
+                f"TFLite model converted successfully. Size: "
+                f"{final_size_mb:.2f} MB"
+            )
+
+        # 4. Save the initial TFLite model to the specified file path.
+        assert filepath.endswith(".tflite"), (
+            "The LiteRT export requires the filepath to end with '.tflite'. "
+            f"Got: {filepath}"
+        )
```
hertschuh: We don't use `assert`, use `if ... : raise ValueError(...)`.

keras/src/export/litert.py
```diff
@@ -0,0 +1,548 @@
...
+            if self.verbose:
+                io_utils.print_msg(
+                    "Performing AOT compilation for LiteRT targets..."
+                )
+            compiled_models = self._aot_compile(filepath)
+        elif self.aot_compile_targets and not litert.available:
+            logging.warning(
+                "AOT compilation requested but LiteRT is not available. "
+                "Skipping AOT compilation."
+            )
```
hertschuh: Raise an actual error in this case.

keras/src/export/litert.py
```diff
@@ -0,0 +1,548 @@
...
+            io_utils.print_msg(
+                f"LiteRT export completed. Base model: {filepath}"
+            )
+            if compiled_models:
+                io_utils.print_msg(
+                    f"AOT compiled models: {len(compiled_models.models)} "
+                    "variants"
+                )
+
+        return compiled_models if compiled_models else filepath
```
hertschuh: It's a bit of weird API. Should we save the compiled models as files instead? In this case, the file names should be passed to `__init__`.

keras/src/export/litert.py
```diff
@@ -0,0 +1,548 @@
...
+                io_utils.print_msg(
+                    f"{model_type} model detected. Trying direct conversion..."
+                )
+
+            converter = tf.lite.TFLiteConverter.from_keras_model(self.model)
+            converter.target_spec.supported_ops = [
+                tf.lite.OpsSet.TFLITE_BUILTINS,
+                tf.lite.OpsSet.SELECT_TF_OPS,
+            ]
+            converter.experimental_enable_resource_variables = False
```
hertschuh: `converter.experimental_enable_resource_variables = False`

Why this? We do rely heavily on resource variables in Keras 3. So this is a bit concerning.

keras/src/export/litert.py
```diff
@@ -0,0 +1,548 @@
...
+    def _convert_with_wrapper(self, input_signature):
+        """Converts the model to TFLite using the tf.Module wrapper.
+
+        Returns:
+            A bytes object containing the serialized TFLite model.
+        """
+
+        # Define the wrapper class dynamically to avoid module-level
+        # tf.Module inheritance
+        class KerasModelWrapper(tf.Module):
```
hertschuh: It's really not clear:
- what problem(s) this wrapper solve
- how each problem is solved
Can you provide some details? Is it just the multi-input / dict inputs case that is a problem?

I also suspect we don't really need a `tf.Module` and that a `tf.function` might do the job with less code.

keras/src/export/litert.py
```diff
@@ -0,0 +1,548 @@
...
+            )
+
+        # Try multiple conversion strategies for better inference compatibility
+        conversion_strategies = [
+            {
+                "experimental_enable_resource_variables": False,
+                "name": "without resource variables",
+            },
+            {
+                "experimental_enable_resource_variables": True,
```
hertschuh: Per my comment above, you should do `experimental_enable_resource_variables = True`. Keras 3 only supports resource variables.

keras/src/export/litert.py
```diff
@@ -0,0 +1,548 @@
...
+                model_type = "Sequential" if is_sequential else "Functional"
+                io_utils.print_msg(
+                    f"Direct conversion failed for {model_type} model: "
+                    f"{direct_error}"
+                )
+                io_utils.print_msg(
+                    "Falling back to wrapper-based conversion..."
+                )
+
+            return self._convert_with_wrapper(input_signature)
```
hertschuh: Instead of trying and failing and trying something else, we should have code that explicitly detects when the wrapper is needed. I think it's just a question if inspecting the input signature and finding out if there are multiple inputs or dicts.

keras/src/export/litert.py
```diff
@@ -0,0 +1,548 @@
...
+                    return self._model(*args, **kwargs)
+
+        # 1. Wrap the Keras model in our clean tf.Module.
+        wrapper = KerasModelWrapper(self.model)
+
+        # 2. Get a concrete function from the wrapper.
+        if not isinstance(input_signature, (list, tuple)):
+            input_signature = [input_signature]
+
+        from keras.src.export.export_utils import make_tf_tensor_spec
```
hertschuh: Move import at the top

keras/src/export/litert.py
```diff
@@ -0,0 +1,548 @@
...
+            io_utils.print_msg("Starting LiteRT export...")
+
+        # 1. Ensure the model is built by calling it if necessary
+        self._ensure_model_built()
+
+        # 2. Resolve / infer input signature
+        if self.input_signature is None:
+            if self.verbose:
+                io_utils.print_msg("Inferring input signature from model.")
+            from keras.src.export.export_utils import get_input_signature
```
hertschuh: Move import at the top

keras/src/export/litert.py
```diff
@@ -0,0 +1,548 @@
...
+                    if hasattr(converter.target_spec, spec_key):
+                        setattr(converter.target_spec, spec_key, spec_value)
+                        if self.verbose:
+                            io_utils.print_msg(
+                                f"Applied target_spec setting: {spec_key}"
+                            )
+            elif self.verbose:
+                io_utils.print_msg(
+                    f"Warning: Unknown converter setting '{key}' - ignoring"
+                )
```
hertschuh: else raise a `ValueError`, we should not ignore anything that is incorrect.

keras/src/export/litert.py
```diff
@@ -0,0 +1,548 @@
...
+                        io_utils.print_msg(
+                            f"  Failed: {backend.id()} - {error}"
+                        )
+
+                # Print compilation report if available
+                try:
+                    report = result.compilation_report()
+                    if report:
+                        io_utils.print_msg("Compilation Report:")
+                        io_utils.print_msg(report)
```
hertschuh: I'm ok keeping these `print_msg`, but basically these are the only ones to keep.

keras/src/export/litert.py
```diff
@@ -0,0 +1,548 @@
...
+                except Exception:
```
hertschuh: Why is this needed?

keras/src/export/litert.py
```diff
@@ -0,0 +1,548 @@
...
+                    pass
+
+            return result
+
+        except Exception as e:
+            if self.verbose:
+                io_utils.print_msg(f"AOT compilation failed: {e}")
+                io_utils.print_msg(traceback.format_exc())
+            raise RuntimeError(f"AOT compilation failed: {e}")
```
hertschuh: Just remove the try / except block and let the original exception go through.

keras/src/export/litert.py
```diff
@@ -0,0 +1,548 @@
...
+        return exporter.export(filepath)
+
+    @classmethod
+    def get_available_targets(cls):
+        """Get list of available LiteRT AOT compilation targets."""
+        if not litert.available:
+            return []
+
+        dummy_exporter = cls(model=None)
+        return dummy_exporter._get_available_litert_targets()
```
hertschuh: What is this? Is this needed? How would users use this?

It's not exported and it's not documented.

keras/src/export/litert.py
pctablet505: https://github.com/keras-team/keras/pull/21674#discussion_r2423270547

We would be dealing with `kwargs` in a separate PR, for consistency across all formats we have left it as kwargs.
later it will be renamed as `litert_kwargs`.

keras/src/export/litert.py
pctablet505: AOT compilation was earlier an opensource feature from google, now it is not opensource, removed everything related to AOT compile

keras/src/export/litert.py
pctablet505: verbose has been kept to keep consistency with other formats. 
Here I stopped using it.

keras/src/export/litert.py
pctablet505: AOT compile is not supported anymore.

pctablet505:
@gemini-code-assist review

keras/src/export/litert.py
pctablet505: Thanks for the suggestion, we can remove lot of code from here.

keras/src/export/litert.py
```diff
@@ -0,0 +1,458 @@
...
+            converter.experimental_enable_resource_variables = False
+
+            # Apply any additional converter settings from kwargs
+            self._apply_converter_kwargs(converter)
+
+            tflite_model = converter.convert()
+
+            return tflite_model
+
+        except Exception:
```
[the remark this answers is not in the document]
pctablet505: we don't want logging.

keras/src/export/export_utils.py
```diff
@@ -7,6 +7,14 @@
...
 
 def get_input_signature(model):
+    """Get input signature for model export.
+
+    Args:
+        model: A Keras Model instance.
+
+    Returns:
+        Input signature suitable for model export (always a tuple or list).
+    """
```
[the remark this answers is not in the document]
pctablet505: not required.

pctablet505:
@gemini-code-assist review

keras/src/models/model.py
```diff
@@ -627,18 +634,41 @@ def export(
...
+        interpreter = tf.lite.Interpreter(model_path="path/to/location")
+        interpreter.allocate_tensors()
+        interpreter.set_tensor(
+            interpreter.get_input_details()[0]['index'], input_data
+        )
+        interpreter.invoke()
+        output_data = interpreter.get_tensor(
+            interpreter.get_output_details()[0]['index']
+        )
+        ```
```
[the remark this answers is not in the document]
pctablet505: this is not a user guide, we already have enough info in docstring, and we have a separate user guide.

keras/src/backend/tensorflow/layer.py
```diff
@@ -62,17 +62,73 @@ def _trackable_children(self, save_type="checkpoint", **kwargs):
...
-                    children[tracked_attr] = dict(tracked_item)
-                if isinstance(tracked_item, tracking.TrackedSet):
-                    children[tracked_attr] = list(tracked_item)
+            # Convert Keras tracked collections to plain Python structures
+            # without creating TensorFlow trackable dependencies
+            self._convert_tracked_collections(children)
 
         return children
 
+    @tf.__internal__.tracking.no_automatic_dependency_tracking
```
hertschuh: Couldn't you just put `@tf.__internal__.tracking.no_automatic_dependency_tracking` on `def _trackable_children`?

keras/src/backend/tensorflow/layer.py
```diff
@@ -62,17 +62,73 @@ def _trackable_children(self, save_type="checkpoint", **kwargs):
...
+                    )
+                except (ImportError, ModuleNotFoundError):
+                    return None
+
+            # Handle dict/list/single tensor inputs
+            if isinstance(inputs, dict):
+                return {k: _make_spec(v) for k, v in inputs.items()}
+            if isinstance(inputs, (list, tuple)):
+                return [_make_spec(t) for t in inputs]
+            return _make_spec(inputs)
```
hertschuh: You don't need recursion, just use `keras.tree.map_structure`.

keras/src/backend/tensorflow/layer.py
```diff
@@ -62,17 +62,73 @@ def _trackable_children(self, save_type="checkpoint", **kwargs):
...
+
+        Returns:
+            A TensorSpec, list or dict mirroring the model inputs, or
+            `None` when specs cannot be inferred.
+        """
+        # Prefer the base implementation if available
+        try:
+            return super()._get_save_spec(dynamic_batch)
+        except AttributeError:
+            # Fall back to building specs from `self.inputs`
```
hertschuh, line 104: I think you can replace this whole implementation (lines 106 - 130) with one line:

```python
return tree.map_structure(make_tf_tensor_spec, input_signature)
```

You'll need to import from `keras.src.export.export_utils import make_tf_tensor_spec`

Look at `make_tf_tensor_spec` and how it's used. You may need to add some code to `make_tf_tensor_spec`.

requirements-tensorflow-cuda.txt
```diff
@@ -1,6 +1,7 @@
 # Tensorflow with cuda support.
 tensorflow[and-cuda]~=2.20.0
 tf2onnx
+ai-edge-litert
```
hertschuh, line 4: Why are we not testing this on CPU too?

keras/src/export/litert.py
```diff
@@ -0,0 +1,255 @@
...
+        Args:
+            filepath: Output path for the exported model
+
+        Returns:
+            Path to exported model
+        """
+        # 1. Resolve / infer input signature
+        if self.input_signature is None:
+            # Try dict-specific inference first (for models with dict inputs)
+            dict_signature = self._infer_dict_input_signature()
```
hertschuh: You are checking dict vs. not, but it's a bit more complicated than that. Inputs can be nested to any level, it can be a list of dicts of inputs, or a dict of lists, etc. But the good news is that it's actually easy to generalize with `keras.tree`, in fact, it will be less code once you handle any arbitrary structure.

There are 3 case:
1. single input
2. flat list of inputs
3. everything else (nested lists, dict...)

The way you check is:
```python
if not keras.tree.is_nested(inputs):
    # case 1
elif isintance(inputs, list) and len(inputs) == len(keras.tree.flatten(inputs)):
    # case 2
else:
    # case 3
``` 

You'll do this instead of the current `has_dict_inputs`.

keras/src/export/litert.py
```diff
@@ -0,0 +1,255 @@
...
+            if dict_signature is not None:
+                self.input_signature = dict_signature
+            else:
+                # Fall back to standard inference
+                self.input_signature = get_input_signature(self.model)
```
hertschuh: Why did you have to do that? `get_input_signature` will return the dict, no?

keras/src/export/litert.py
```diff
@@ -0,0 +1,255 @@
...
+        # Check if we have dict inputs that need adaptation
+        has_dict_inputs = isinstance(self.input_signature, dict)
+
+        if has_dict_inputs:
+            # Create adapter model that converts list to dict
+            adapted_model = self._create_dict_adapter(self.input_signature)
+
+            # Convert dict signature to list for TFLite conversion
+            # The adapter will handle the dict->list conversion
+            input_signature_list = list(self.input_signature.values())
```
hertschuh: `input_signature_list = keras.tree.flatten(self.input_signature)`

This works for any nested structure, not just dicts.

keras/src/export/litert.py
```diff
@@ -0,0 +1,255 @@
...
+                "The LiteRT export requires the filepath to end with "
+                "'.tflite'. Got: {filepath}"
+            )
+
+        with open(filepath, "wb") as f:
+            f.write(tflite_model)
+
+        return filepath
+
+    def _create_dict_adapter(self, input_signature_dict):
```
hertschuh: Let's rename it since it's for more than just dicts. `_create_nested_inputs_adapter(self, input_signature_struct)`.

keras/src/export/litert.py
```diff
@@ -0,0 +1,255 @@
...
+
+        Returns:
+            A Functional model that accepts list inputs and converts to dict
+        """
+        io_utils.print_msg(
+            f"Creating adapter for dictionary inputs: "
+            f"{list(input_signature_dict.keys())}"
+        )
+
+        input_keys = list(input_signature_dict.keys())
```
hertschuh: Not needed

keras/src/export/litert.py
```diff
@@ -0,0 +1,255 @@
...
+        Args:
+            input_signature_dict: Dictionary mapping input names to InputSpec
+
+        Returns:
+            A Functional model that accepts list inputs and converts to dict
+        """
+        io_utils.print_msg(
+            f"Creating adapter for dictionary inputs: "
+            f"{list(input_signature_dict.keys())}"
+        )
```
hertschuh: Instead, print the mapping from the nested structure to the flat list so that people know in which order to put their inputs in the list.

```python
paths_and_specs = keras.tree.flatten_with_path(input_signature_struct)
paths = [".".join(str(e) for e in p) for v, p in paths_and_specs]
io_utils.print_msg(f"Creating adapter for inputs: {paths}")
```

keras/src/export/litert.py
```diff
@@ -0,0 +1,255 @@
...
+                shape=spec.shape[1:],  # Remove batch dimension
+                dtype=spec.dtype,
+                name=name,
+            )
+            input_layers.append(input_layer)
+
+        # Create dict from list inputs
+        inputs_dict = {
+            name: layer for name, layer in zip(input_keys, input_layers)
+        }
```
hertschuh: Replace with

```python
    input_layers = []
    for path, spec in paths_and_specs:
             input_layer = layers.Input(
                shape=spec.shape[1:],  # Remove batch dimension
                dtype=spec.dtype,
                name=name,
            )
            input_layers.append(input_layer)

    inputs_structure = keras.tree.pack_sequence_as(input_signature_struct, input_layers)
    outputs = self.model(inputs_structure)
    adapted_model = models.Model(inputs=input_layers, outputs=outputs)
```

keras/src/export/litert.py
```diff
@@ -0,0 +1,255 @@
...
+        # Fall back to model.inputs if it's a dict
+        if hasattr(self.model, "inputs") and isinstance(
+            self.model.inputs, dict
+        ):
+            return {
+                name: make_input_spec(inp)
+                for name, inp in self.model.inputs.items()
+            }
+
+        return None
```
hertschuh: If needed, this could be:

```python
keras.tree.map_structure(make_input_spec, self.model._inputs_struct)
```

hertschuh (commented):
Thanks! This is getting really close.

requirements-tensorflow-cuda.txt
pctablet505: Added it to requirements.txt, now we can test on cpu also.

keras/src/backend/tensorflow/layer.py
```diff
@@ -13,7 +13,6 @@ def __init__(self, *args, **kwargs):
         self._saved_model_arg_spec = None
         self._tracked = []
 
-    @tf.__internal__.tracking.no_automatic_dependency_tracking
```
hertschuh, line 16: Why is this no longer needed?

keras/src/export/export_utils.py
```diff
@@ -86,13 +102,33 @@ def make_input_spec(x):
...
+            shape[0] = None
+            tensor_spec = tf.TensorSpec(
+                tuple(shape), dtype=tensor_spec.dtype, name=tensor_spec.name
+            )
     else:
         input_spec = make_input_spec(x)
+        shape = input_spec.shape
+        # Adjust batch dimension if needed and shape is not None
+        if dynamic_batch and shape is not None and len(shape) > 0:
+            shape = tuple(None if i == 0 else s for i, s in enumerate(shape))
```
hertschuh, line 130: `shape = tuple(None if i == 0 else s for i, s in enumerate(shape))`

nitpick: use this line to replace lines 119-120 for consistency

hertschuh:
Can you rebase and address the merge conflict?

keras/src/backend/tensorflow/layer.py
pctablet505: earlier we kept it here, then as per your suggestion, I moved it to `_trackable_children`

PR #21679, opened 2025-09-18
Fix TF int64 promotion issue.

### LINKED ISSUE #21677: Keras Policy forces 64-bit integer arithmetic results to int32
**Describe the bug**
The arithmetic operation A+B, where both A and B are `int64` Keras Variables (tested on TensorFlow), unexpectedly results in a `tf.int32` output. This is a consequence of an explicit dtype promotion policy in:
https://github.com/keras-team/keras/blob/5dbdf60fab00c9b4592b16ec6dbc31c9c3ebf394/keras/src/backend/common/dtypes.py#L280-L283

This policy downcasts the result type from int64 to int32 to align Keras's integer precision with the default float32 precision and causes weird dtype behaviors.

This behavior was introduced in: https://github.com/keras-team/keras/pull/21604

**To Reproduce**

Image

```py
import numpy as np
import tensorflow as tf
import keras

# Set up a random number generator for reproducibility
# Note: The original image uses 'np.random.default_rng()' 
# which is the modern NumPy way, so we'll stick to that.
rng = np.random.default_rng()

# 1. Create a NumPy array 'x'
# rng.standard_normal((2, 3, 4, 5)) creates a 4D array of random standard normal values
x = rng.standard_normal((2, 3, 4, 5))

# 2. Create a Keras Variable for 'indices'
# This creates a variable of shape (3,) with all ones, and data type 'int64'
indices = keras.Variable(
    initializer="ones",
    shape=(3,),
    dtype="int64",
    trainable=False
)

# 3. Print the data types of 'x' and 'indices'
print(f"Data type of x: {x.dtype}")
print(f"Data type of indices: {indices.dtype}")

# 4. TensorFlow Type Casting and Operations
# Get the value of the 2nd dimension of x's shape (which is 3)
x_shape_dim2 = tf.shape(x)[2]
print(f"tf.shape(x)[2]: {x_shape_dim2.numpy()} (dtype: {x_shape_dim2.dtype.name})")

# Cast the shape dimension value to a TensorFlow 32-bit integer
x_shape_dim2_int32 = tf.cast(x_shape_dim2, tf.int32)
print(f"tf.cast(tf.shape(x)[2], tf.int32): {x_shape_dim2_int32.numpy()} (dtype: {x_shape_dim2_int32.dtype.name})")

# Cast the 'indices' variable to a TensorFlow 32-bit integer
indices_int32 = tf.cast(indices, tf.int32)
print(f"tf.cast(indices, tf.int32) (a Keras Variable cast): dtype: {indices_int32.dtype.name}")

# Perform the final element-wise addition and print the resulting dtype
# The result will be an 'tf.int32' Tensor because both operands are cast to 'tf.int32'
final_result = indices_int32 + x_shape_dim2_int32
print(f"\nFinal operation (indices + tf.cast(tf.shape(x)[2], indices.dtype)).dtype: {final_result.dtype.name}")
```
**Expected behavior**
Adding two 64-bit integers should result in a 64-bit integer.

**Additional context**

**Would you like to help us fix it?**

james77777778:
Fix #21677

The root cause of the issue is that #21604 forced 64bit dtypes to be remapped to 32bit dtypes to be consistent with JAX's dtype promotion rules. However, TF requires int64 variables for GPU ops.

This PR fixes the issue by exclusively loosening the constraint for the TF backend and skipping the tests that were causing the errors.

cc @JyotinderSingh

### DIFF
--- a/keras/src/backend/common/dtypes_test.py
+++ b/keras/src/backend/common/dtypes_test.py
@@ -29,8 +29,11 @@ class DtypesTest(test_case.TestCase):
     if backend.backend() == "torch":
         ALL_DTYPES = [x for x in ALL_DTYPES if x not in ("uint16", "uint32")]
     elif backend.backend() == "tensorflow":
-        # TODO(hongyu): Re-enable uint32 tests when we figure out how to handle
-        # (uint32, int*) -> int64 promotion. JAX doesn't natively support int64.
+        # TODO(hongyu): Re-enable uint32 tests once we determine how to handle
+        # dtypes.result_type(uint32, int*) -> int64 promotion.
+        # Since TF variables require int64 to be placed on the GPU, we
+        # exclusively enable the int64 dtype for TF. However, JAX does not
+        # natively support int64, which prevents us from comparing the dtypes.
         ALL_DTYPES = [x for x in ALL_DTYPES if x not in ("uint32",)]
     elif backend.backend() == "openvino":
         ALL_DTYPES = [x for x in ALL_DTYPES if x not in ("complex64",)]


keras/src/backend/common/variables_test.py
```diff
@@ -807,6 +807,11 @@ class VariableOpsDTypeTest(test_case.TestCase):
             x for x in ALL_DTYPES if x not in ("uint16", "uint32", "complex64")
         ]
         INT_DTYPES = [x for x in INT_DTYPES if x not in ("uint16", "uint32")]
+    elif backend.backend() == "tensorflow":
+        # TODO(hongyu): Re-enable uint32 tests when we figure out how to handle
+        # (uint32, int*) -> int64 promotion. JAX doesn't natively support int64.
```
JyotinderSingh: It may be helpful to explain the TODO in a little more detail, so that we can provide enough context to other contributors who may want to work on this.

keras/src/ops/numpy_test.py
```diff
@@ -5774,6 +5774,11 @@ class NumpyDtypeTest(testing.TestCase):
     if backend.backend() == "torch":
         ALL_DTYPES = [x for x in ALL_DTYPES if x not in ("uint16", "uint32")]
         INT_DTYPES = [x for x in INT_DTYPES if x not in ("uint16", "uint32")]
+    elif backend.backend() == "tensorflow":
+        # TODO(hongyu): Re-enable uint32 tests when we figure out how to handle
+        # (uint32, int*) -> int64 promotion. JAX doesn't natively support int64.
```
JyotinderSingh: Same for this

JyotinderSingh (approved):
Thanks for the PR! LGTM

keras/src/backend/common/variables_test.py
james77777778: Updated.

```python
        # TODO(hongyu): Re-enable uint32 tests once we determine how to handle
        # dtypes.result_type(uint32, int*) -> int64 promotion.
        # Since TF variables require int64 to be placed on the GPU, we
        # exclusively enable the int64 dtype for TF. However, JAX does not
        # natively support int64, which prevents us from comparing the dtypes.
```

fchollet (approved):
LGTM, thank you for the fix.

PR #21343, opened 2025-06-02
[*.py] Use f-strings and `os.path.join` throughout

SamuelMarks:

PR #21680, opened 2025-09-18
Do not use `os.path.join` for variable paths.

hertschuh:
Variable paths do not represent a filesystem path and should not be OS dependent, they should always use "/" as a separator.

This reverts https://github.com/keras-team/keras/pull/21343 just for variables.


SamanehSaadat (approved):
Thanks, Fabien!

PR #21681, opened 2025-09-19
Refactor `save_own_variables` and `load_own_variables` to use the names as the keys.

james77777778:
`load_own_variables` for legacy models will now detect the key values of the incoming `store` and use `_legacy_load_own_variables`.

This PR adds a comprehensive `test_legacy_load_own_variables` and simplifies the quantization tests by consolidating the `int8` and `int4` tests.

@JyotinderSingh @hertschuh  
Let me know if this looks good. If so, I will refactor other layers as well.

### DIFF
--- a/keras/src/layers/core/dense_test.py
+++ b/keras/src/layers/core/dense_test.py
@@ -867,14 +867,14 @@ class DenseTest(testing.TestCase):
         layer = layers.Dense(units=16)
         layer.build((None, 8))
         layer.load_own_variables(float32_store)
-        self.assertAllClose(layer.kernel, float32_store["0"])
+        self.assertAllClose(layer._kernel, float32_store["0"])
         self.assertAllClose(layer.bias, float32_store["1"])
 
         # Test int8-quantized layer.
         layer = layers.Dense(units=16, dtype="int8_from_float32")
         layer.build((None, 8))
         layer.load_own_variables(int8_store)
-        self.assertAllClose(layer.kernel, int8_store["0"])
+        self.assertAllClose(layer._kernel, int8_store["0"])
         self.assertAllClose(layer.bias, int8_store["1"])
         self.assertAllClose(layer.kernel_scale, int8_store["2"])
 
@@ -882,7 +882,7 @@ class DenseTest(testing.TestCase):
         layer = layers.Dense(units=16, dtype="int4_from_float32")
         layer.build((None, 8))
         layer.load_own_variables(int4_store)
-        self.assertAllClose(layer.kernel, int4_store["0"])
+        self.assertAllClose(layer._kernel, int4_store["0"])
         self.assertAllClose(layer.bias, int4_store["1"])
         self.assertAllClose(layer.kernel_scale, int4_store["2"])
 
@@ -890,7 +890,7 @@ class DenseTest(testing.TestCase):
         layer = layers.Dense(units=16, dtype="float8_from_float32")
         layer.build((None, 8))
         layer.load_own_variables(float8_store)
-        self.assertAllClose(layer.kernel, float8_store["0"])
+        self.assertAllClose(layer._kernel, float8_store["0"])
         self.assertAllClose(layer.bias, float8_store["1"])
         self.assertAllClose(layer.inputs_scale, float8_store["2"])
         self.assertAllClose(layer.inputs_amax_history, float8_store["3"])
--- a/keras/src/layers/core/einsum_dense_test.py
+++ b/keras/src/layers/core/einsum_dense_test.py
@@ -1084,14 +1084,14 @@ class EinsumDenseTest(testing.TestCase):
         layer = layers.EinsumDense(**config)
         layer.build((None, 3))
         layer.load_own_variables(float32_store)
-        self.assertAllClose(layer.kernel, float32_store["0"])
+        self.assertAllClose(layer._kernel, float32_store["0"])
         self.assertAllClose(layer.bias, float32_store["1"])
 
         # Test int8-quantized layer.
         layer = layers.EinsumDense(**config, dtype="int8_from_float32")
         layer.build((None, 3))
         layer.load_own_variables(int8_store)
-        self.assertAllClose(layer.kernel, int8_store["0"])
+        self.assertAllClose(layer._kernel, int8_store["0"])
         self.assertAllClose(layer.bias, int8_store["1"])
         self.assertAllClose(layer.kernel_scale, int8_store["2"])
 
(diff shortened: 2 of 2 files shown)


james77777778:
> Yeah, this is great!
> 
> Can you add the same for EinsumDense?

I've updated the logic for `EinsumDense` and `Embedding` and added int4 quant support for `Embedding`.

keras/src/layers/core/dense.py
```diff
@@ -382,6 +350,27 @@ def _check_load_own_variables(self, store):
                 f"Expected: {[v.name for v in all_vars]}"
             )
 
+    # Per-mode variable spec.
+    MODE_SPEC = {
```
JyotinderSingh: I suggest renaming `MODE_SPEC` to `QUANTIZATION_VARIABLE_SPEC`. The current name is now somewhat ambiguous because the spec is stored as a class attribute. Additionally, given that this value is a constant, it might be better to define it as a private constant outside the class rather than keeping it at the class level. WDYT?

keras/src/layers/core/dense_test.py
```diff
@@ -976,3 +816,74 @@ def test_gptq_serialization(self):
         new_layer = layers.Dense.from_config(config)
         new_layer.build((None, 8))
         self.assertEqual(new_layer.quantization_mode, "gptq")
+
+    def test_legacy_load_own_variables(self):
+        # In previous versions, `load_own_variables` accepted a store with
+        # numeric keys.
+        # TODO: add gptq_store test.
```
JyotinderSingh: Are we planning to include the gptq store test case as a part of this PR? If not, feel free to assign it to me.

keras/src/layers/core/einsum_dense.py
```diff
@@ -449,6 +416,27 @@ def _check_load_own_variables(self, store):
...
+            "outputs_grad_scale",
+            "outputs_grad_amax_history",
+        ],
+        "gptq": [
+            "quantized_kernel",
+            "kernel_scale",
+            "kernel_zero",
+            "g_idx",
+        ],
+    }
```
JyotinderSingh: Same comment here.

keras/src/layers/core/embedding.py
```diff
@@ -363,49 +418,151 @@ def _int8_call(self, inputs, training=None):
...
+            ops.cast(outputs, dtype=self.compute_dtype),
+            ops.expand_dims(embeddings_scale, axis=-1),
+        )
+        if self.lora_enabled:
+            lora_outputs = ops.take(self.lora_embeddings_a, inputs, axis=0)
+            lora_outputs = ops.matmul(lora_outputs, self.lora_embeddings_b)
+            outputs = ops.add(
+                outputs, (self.lora_alpha / self.lora_rank) * lora_outputs
+            )
+        return outputs
```
JyotinderSingh, line 453: Thanks for adding `"int4"` support as part of this work!

JyotinderSingh (commented):
Thanks for this PR! The changes look good to me, added a few minor comments. Also good to see `"int4"` quantization support added to the `Embedding` layer.

JyotinderSingh:
Since we are adding `"int4"` support to the Embedding layer as a part of this PR - could you also update its `kernel` property definition to unpack the 4-bit kernel before returning it to the user?

I already have a PR doing the same for `Dense` and `EinsumeDense` which you can refer to: https://github.com/keras-team/keras/pull/21684

keras/src/layers/core/dense.py
james77777778: Thanks for pointing this out.

I think keeping it in the layer (or class) could be helpful if one wants to subclass it. I've modified it to be a property and am open to any suggestions.

keras/src/layers/core/dense_test.py
james77777778: Assigned. Thanks!

james77777778:
> Since we are adding `"int4"` support to the Embedding layer as a part of this PR - could you also update its `kernel` property definition to unpack the 4-bit kernel before returning it to the user?
> 
> I already have a PR doing the same for `Dense` and `EinsumeDense` which you can refer to: #21684

Sure. Updated.

hertschuh:
@james77777778 

Thanks for the PR!

Do you mind rebasing this?

james77777778:
> Do you mind rebasing this?

Rebased.

james77777778:
@hertschuh @JyotinderSingh 

I think this modification will cause a breakage in one scenario:
- Saving the model in a newer Keras (e.g. `3.12` in next weeks/months)
- Loading the model in an older Keras (`<=3.11`)

Also, it will be difficult to warn users to upgrade their Keras version since we don't patch released versions.
Should we consider adding an option to save the model using the old keys? This would allow users to load the model with an older version of Keras.

Something like this:
```python
Model.save(..., use_legacy_key=True)
```

cc @mattdangerw @fchollet

hertschuh:
> I think this modification will cause a breakage in one scenario:
> 
> * Saving the model in a newer Keras (e.g. `3.12` in next weeks/months)
> * Loading the model in an older Keras (`<=3.11`)
> 
> Also, it will be difficult to warn users to upgrade their Keras version since we don't patch released versions. Should we consider adding an option to save the model using the old keys? This would allow users to load the model with an older version of Keras.
> 
> Something like this:
> 
> ```python
> Model.save(..., use_legacy_key=True)
> ```

While technically it is a breaking change, I don't think it's worth the trouble. I doubt people work with multiple different versions of Keras at the same time.

james77777778:
> While technically it is a breaking change, I don't think it's worth the trouble. I doubt people work with multiple different versions of Keras at the same time.

The primary risk is that we might release KerasHub presets with `keras>3.11` which would break for users running `keras<=3.11`

hertschuh:
> > While technically it is a breaking change, I don't think it's worth the trouble. I doubt people work with multiple different versions of Keras at the same time.
> 
> The primary risk is that we might release KerasHub presets with `keras>3.11` which would break for users running `keras<=3.11`

Oh I see, that's an issue.

james77777778:
> > > While technically it is a breaking change, I don't think it's worth the trouble. I doubt people work with multiple different versions of Keras at the same time.
> > 
> > 
> > The primary risk is that we might release KerasHub presets with `keras>3.11` which would break for users running `keras<=3.11`
> 
> Oh I see, that's an issue.

Let me know if we want to add support for using legacy key names.
If so, should we enable this by default?

fchollet:
What do the weights look like before and after?

james77777778:
> What do the weights look like before and after?

For a regular `Dense` layer, the keys are as follows:
- before: `<KeysViewHDF5 ['0', '1']>`
- after: `<KeysViewHDF5 ['bias', 'kernel']>`

PR #21682, opened 2025-09-19
Do not use `os.path.sep` to detect `/` in names.

hertschuh:
Operation and layer names do not represent a filesystem path and should not be OS dependent. Namescope paths always use "/" as a separator, which is what we should be looking for.

This reverts https://github.com/keras-team/keras/pull/21343 just for `Operation`.

Note that [the unit test](https://github.com/keras-team/keras/blob/master/keras/src/ops/operation_test.py#L331) was correct. Simply, we never run it on Windows, which is why this was not caught.

### DIFF
--- a/keras/src/ops/operation.py
+++ b/keras/src/ops/operation.py
@@ -1,5 +1,4 @@
 import inspect
-import os.path
 import textwrap
 
 from keras.src import backend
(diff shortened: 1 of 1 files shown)


SamanehSaadat (approved):
Thanks!

Issue #21678, opened 2025-09-18
Whether 'jpg' is supported or not in func `keras.utils.save_img()`


ILCSFNO:
## Bug Issue

I found an issue on `keras.utils.save_img()` with arg `file_format='jpg'`, the doc of `keras.utils.save_img()` shows its description as below:

https://github.com/keras-team/keras/blob/fa9c4ba633ca184499650f210d654485cd65e7b1/keras/src/utils/image_utils.py#L171-L173

See the repro below, with TensorFlow 2.19.0 and Keras nightly:
### Repro
```python
import numpy as np
import keras

file_format = 'jpg' # jpeg, png is OK, jpg will raise error

image_array = np.random.randint(0, 256, size=(100, 100, 3), dtype=np.uint8)
image_array = (((image_array - np.min(image_array)) / (np.max(image_array) - np.min(image_array))) * 255)
keras.utils.save_img(('random_image.' + file_format), image_array, data_format='channels_last', file_format=file_format)
```
### Output
```text
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[12], line 8
      6 image_array = np.random.randint(0, 256, size=(100, 100, 3), dtype=np.uint8)
      7 image_array = (((image_array - np.min(image_array)) / (np.max(image_array) - np.min(image_array))) * 255)
----> 8 keras.utils.save_img(('random_image.' + file_format), image_array, data_format='channels_last', file_format=file_format)

File ~/anaconda3/envs/testoracle-tf/lib/python3.10/site-packages/keras/src/utils/image_utils.py:184, in save_img(path, x, data_format, file_format, scale, **kwargs)
    180     warnings.warn(
    181         "The JPG format does not support RGBA images, converting to RGB."
    182     )
    183     img = img.convert("RGB")
--> 184 img.save(path, format=file_format, **kwargs)

File ~/anaconda3/envs/testoracle-tf/lib/python3.10/site-packages/PIL/Image.py:2581, in Image.save(self, fp, format, **params)
   2579     save_handler = SAVE_ALL[format.upper()]
   2580 else:
-> 2581     save_handler = SAVE[format.upper()]
   2583 created = False
   2584 if open_fp:

KeyError: 'JPG'
```

Firstly, I think that maybe this func is not supported for `'jpg'`, but see the codes below,

[...]

PR #21683, opened 2025-09-20
Fix: Support 'jpg' format in keras.utils.save_img()

utsab345:
### Summary
This PR fixes issue #21678. 
The `save_img` function previously failed with `.jpg` extension because Pillow only supports 'JPEG'. 
Now, `.jpg` is automatically mapped to 'jpeg'.

### Changes
- Updated keras/utils/image_utils.py to map 'jpg' → 'jpeg'.
- Added test `test_save_img_with_jpg` in integration_tests/test_save_img.py.

### Notes
- Verified with pytest, all tests pass.


fchollet:
Thanks for the PR! Please fix the code format and address the relevant comments from Gemini.

fchollet:
You can fix the code format via `sh shell/format.sh`.

keras/src/utils/image_utils.py
```diff
@@ -175,13 +175,35 @@ def save_img(path, x, data_format=None, file_format=None, scale=True, **kwargs):
...
+        file_format = "jpeg"
+    if _format is not None and _format.lower() == "jpg":
+        _format = "jpeg"
+
+    # Convert array to PIL Image
     img = array_to_img(x, data_format=data_format, scale=scale)
-    if img.mode == "RGBA" and (file_format == "jpg" or file_format == "jpeg"):
+
+    # Handle RGBA → RGB if saving to JPEG
+    if img.mode == "RGBA" and (_format in ("jpeg", "jpg")):
```
hertschuh: You're checking `_format`, but you're no longer verifying `file_format`.

keras/src/utils/image_utils.py
```diff
@@ -175,13 +175,35 @@ def save_img(path, x, data_format=None, file_format=None, scale=True, **kwargs):
...
         warnings.warn(
-            "The JPG format does not support RGBA images, converting to RGB."
+            "The JPEG format does not support RGBA images, converting to RGB."
         )
         img = img.convert("RGB")
-    img.save(path, format=file_format, **kwargs)
+
+    # Finalize save format (explicit file_format wins)
+    save_format = file_format if file_format is not None else _format
```
hertschuh: It would be easier to do this line 186 instead of having to check both line 197 and lines 188-190.

in fact, you don't need `_format` at all, use `file_format` lines 181-185.

keras/src/utils/image_utils.py
```diff
@@ -175,10 +175,18 @@ def save_img(path, x, data_format=None, file_format=None, scale=True, **kwargs):
         **kwargs: Additional keyword arguments passed to `PIL.Image.save()`.
     """
     data_format = backend.standardize_data_format(data_format)
+
+    # Normalize jpg → jpeg
+    if file_format is not None and file_format.lower() == "jpg":
```
fchollet: This would be breaking

utsab345:
Any update @fchollet ??

integration_tests/test_save_img.py
```diff
@@ -0,0 +1,27 @@
+import os
```
hertschuh: Can you move this code to a `keras/src/utils/image_utils_test.py` file?

This should be fast enough, it doesn't need to be an integration test.

Then use the standard patterns:
- create a class that extends `testing.TestCase`
- use `from absl.testing import parameterized` instead of `pytest.mark.parametrize`
- use `self.get_temp_dir()` to get a temp dir

hertschuh (commented):
Thanks for the update.

integration_tests/test_save_img.py
hertschuh: You can now remove this file.

keras/src/utils/image_utils_test.py
```diff
@@ -0,0 +1,30 @@
...
+    @parameterized.parameters(
+        ((50, 50, 3), "rgb.jpg"),
+        ((50, 50, 4), "rgba.jpg"),
+    )
+    def test_save_jpg(self, shape, name):
+        tmp_dir = self.get_temp_dir()
+        path = os.path.join(tmp_dir, name)
+
+        img = np.random.randint(0, 256, size=shape, dtype=np.uint8)
+        save_img(path, img, file_format="jpg")
```
hertschuh: This is not covering the case that you added support for. That is, the case when the `file_format` was not provided and it was inferred from the path.

keras/src/utils/image_utils.py
```diff
@@ -175,13 +175,28 @@ def save_img(path, x, data_format=None, file_format=None, scale=True, **kwargs):
...
+
+    # Infer format from file_format or path (normalize only internally)
+    save_format = file_format
+    if save_format is None and isinstance(path, (str, pathlib.Path)):
+        ext = pathlib.Path(path).suffix[1:].lower()
+        save_format = ext
+
+    internal_format = (
+        "jpeg" if save_format and save_format.lower() == "jpg" else save_format
+    )
```
hertschuh: This can be simplified. There is no need for 3 variables for the file format. You're no longer using `file_format` after assigning to `save_format`, and you don't need `save_format` as it can be replaced by `internal_format`. I think the version you had earlier was correct.

hertschuh (approved):
Thank you for the fix!

PR #21685, opened 2025-09-22
Add unfold op

pass-lin:
By https://github.com/keras-team/keras/issues/20046

This is an unfold implementation; the API and documentation were referenced from [PyTorch’s version](https://docs.pytorch.org/docs/stable/generated/torch.nn.Unfold.html#unfold).  
At present we only supply backends for JAX, TensorFlow and PyTorch.  
A NumPy implementation has been finished, but its performance would be poor
Finally, I am not able to use  OpenVINO.

### DIFF
--- a/keras/src/backend/jax/nn.py
+++ b/keras/src/backend/jax/nn.py
@@ -1411,3 +1411,27 @@ def dot_product_attention(
     )
     encoded = vmapped_fn(query, key, value, bias, mask, is_causal, scale)
     return jnp.reshape(encoded, output_shape)
+
+
+def unfold(input, kernel_size, dilation=1, padding=0, stride=1):
+    def _pair(x):
+        return (x, x) if isinstance(x, int) else x
+
+    filter_shape = _pair(kernel_size)
+    window_strides = _pair(stride)
+    lhs_dilation = _pair(dilation)
+
+    if isinstance(padding, int):
+        padding = ((padding, padding), (padding, padding))
+    else:
+        padding = ((padding[0], padding[0]), (padding[1], padding[1]))
+
+    patches = lax.conv_general_dilated_patches(
+        lhs=input,
+        filter_shape=filter_shape,
+        window_strides=window_strides,
+        padding=padding,
+        lhs_dilation=lhs_dilation,
+    )
+
+    return patches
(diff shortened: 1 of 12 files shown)


keras/src/backend/tensorflow/nn.py
```diff
@@ -1077,3 +1077,41 @@ def dot_product_attention(
     return _dot_product_attention_xla(
         query, key, value, bias, mask, is_causal, scale
     )
+
+
+def unfold(input, kernel_size, dilation=1, padding=0, stride=1):
+    """
```
fchollet: Please fix docstring format

keras/src/backend/numpy/nn.py
```diff
@@ -1175,3 +1175,48 @@ def dot_product_attention(
     return _dot_product_attention_xla(
         query, key, value, bias, mask, is_causal, scale
     )
+
+
+def unfold(input, kernel_size, dilation=1, padding=0, stride=1):
+    """
```
fchollet: Please fix docstring format

keras/src/ops/nn.py
```diff
@@ -3047,3 +3047,95 @@ def _polar(abs_, angle):
...
+
+    Example:
+
+    >>> x = keras.ops.ones((1, 2, 4, 4))
+    >>> patches = keras.ops.unfold(x, kernel_size=2, stride=2)
+    >>> patches.shape
+    (1, 8, 4)
+
+    The output can be reshaped back to spatial format with
+    `keras.ops.fold` (if implemented).
```
fchollet: fold is implemented, right? So (if implemented) is not necessary

keras/src/backend/jax/nn.py
```diff
@@ -1411,3 +1411,46 @@ def _reshape_to_grouped(t):
     )
     encoded = vmapped_fn(query, key, value, bias, mask, is_causal, scale)
     return jnp.reshape(encoded, output_shape)
+
+
+def unfold(input, kernel_size, dilation=1, padding=0, stride=1):
+    """
+    Extract sliding local blocks from a **NCHW** batched image tensor.
```
fchollet, line 1418: Remove the blank line above -- docstring description should start after `"""`

fchollet (commented):
We can add this op to the API. Thanks for the PR!

keras/src/ops/nn.py
pass-lin: > fold is implemented, right? So (if implemented) is not necessary

This is an incorrect comment. fold is not implemented

pass-lin:
@fchollet  document has been fixed

pass-lin:
@fchollet Could you please check this PR? Thanks.

PR #21686, opened 2025-09-22
Adds 4-bit packing support to GPTQ Quantization

JyotinderSingh:
## Summary

Integrates native 4-bit kernel packing for GPTQ in `Dense` and `EinsumDense`. Kernels are stored packed (two nibbles/byte) and unpacked only at call time. This change also extends `pack_int4`/`unpack_int4` to support both `"int8"` and `"uint8"` paths.

This change reduces the memory requirements for 4-bit integer quantized kernels by half.

## New Logical Flow

1. Build (`_gptq_build`): Allocate `quantized_kernel` with packed columns when `weight_bits==4`
    (`units = ceil(cols/2)` for `Dense`; analogous for `EinsumDense`, storing `gptq_unpacked_column_size`).
2. Quantize (`gptq.py`): If 4-bit: `pack_int4(..., axis=0, dtype="uint8")` before assigning.
3. Call (`_gptq_call`): If 4-bit: `unpack_int4(..., axis=0, dtype="uint8")` → `dequantize_with_sz_map(...)` → matmul, Otherwise: dequantize directly.

## Tests

1. Packing reduction: Ensures param count halves after 4-bit quantize for both `Dense` and `EinsumDense`.
2. Parameterized shapes/axes (+ negative axes, higher ranks) × dtypes (`int8`, `uint8`): `pack_int4` → `unpack_int4` preserves dtype and values.
3. `EinsumDense.get_config` includes `gptq_unpacked_column_size`; GPTQ serialization still passes.

### DIFF
--- a/keras/src/layers/core/dense.py
+++ b/keras/src/layers/core/dense.py
@@ -418,9 +418,17 @@ class Dense(Layer):
         # until calibration has been performed.
         self.is_gptq_calibrated = False
         self.kernel_shape = kernel_shape
+
+        weight_bits = self._get_gptq_weight_bits(config)
+        if weight_bits == 4:
+            # For 4-bit weights, we pack two values per byte.
+            units = math.ceil(kernel_shape[1] / 2)
+        else:
+            units = kernel_shape[1]
+
         self.quantized_kernel = self.add_weight(
             name="kernel",
-            shape=(kernel_shape[1], kernel_shape[0]),
+            shape=(units, kernel_shape[0]),
             initializer="zeros",
             dtype="uint8",
             trainable=False,
@@ -456,10 +464,20 @@ class Dense(Layer):
         if not self.is_gptq_calibrated:
             W = self._kernel
         else:
+            # if 4-bit weights, unpack them
+            if self.dtype_policy.weight_bits == 4:
+                W = quantizers.unpack_int4(
+                    self.quantized_kernel,
+                    orig_len=self.units,
+                    axis=0,
+                    dtype="uint8",
+                )
+            else:
+                W = self.quantized_kernel
             W = (
                 ops.transpose(
                     dequantize_with_sz_map(
-                        self.quantized_kernel,
+                        W,
                         self.kernel_scale,
                         self.kernel_zero,
                         self.g_idx,
@@ -912,6 +930,47 @@ class Dense(Layer):
         else:
             raise ValueError(
                 "For GPTQ quantization, the group_size must be specified"
+                "either through a `dtype_policy` of type "
+                "`GPTQDTypePolicy` or the `config` argument."
+            )
+
+    def _get_gptq_weight_bits(self, config):
+        """Determine the number of weight bits for GPTQ quantization.
+
+        The number of weight bits can be specified either through the `config`
+        argument or through the `dtype_policy` if it is of type
+        `GPTQDTypePolicy`.
+
+        The config argument is usually available when quantizing the layer
+        via the `quantize` method. If the layer was deserialized from a
+        saved model, the weight bits should be specified in the `dtype_policy`.
+
+        Args:
+            config: An optional configuration object that may contain the
+                `weight_bits` attribute.
+        Returns:
+            int. The determined number of weight bits for GPTQ quantization.
+        Raises:
+            ValueError: If the weight bits is not specified in either the
+                `config` or the `dtype_policy`.
+        """
+        if config and isinstance(config, GPTQConfig):
+            return config.weight_bits
+        elif isinstance(self.dtype_policy, GPTQDTypePolicy):
+            return self.dtype_policy.weight_bits
+        elif isinstance(self.dtype_policy, DTypePolicyMap):
+            policy = self.dtype_policy[self.path]
+            if not isinstance(policy, GPTQDTypePolicy):
+                # This should never happen based on how we set the
+                # quantization mode, but we check just in case.
+                raise ValueError(
+                    "Expected a `dtype_policy` of type `GPTQDTypePolicy`."
+                    f"Got: {type(policy)}"
+                )
+            return policy.weight_bits
+        else:
+            raise ValueError(
+                "For GPTQ quantization, the weight_bits must be specified"
                 "either through a `dtype_policy` of type "
                 "`GPTQDTypePolicy` or the `config` argument."
             )
--- a/keras/src/layers/core/dense.py
+++ b/keras/src/layers/core/dense.py
@@ -147,18 +147,39 @@ class Dense(Layer):
             raise AttributeError(
                 "You must build the layer before accessing `kernel`."
             )
-        if (
-            getattr(self, "is_gptq_calibrated", False)
-            and self.quantization_mode == "gptq"
-        ):
-            return self.quantized_kernel
-        kernel = self._kernel
-        if self.quantization_mode == "int4":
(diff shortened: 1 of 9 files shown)


hertschuh:
@JyotinderSingh can you rebase? Thanks!

keras/src/layers/core/dense.py
```diff
@@ -916,3 +953,44 @@ def _get_gptq_group_size(self, config):
                 "either through a `dtype_policy` of type "
                 "`GPTQDTypePolicy` or the `config` argument."
             )
+
+    def _get_gptq_weight_bits(self, config):
```
james77777778: Should we move this function to `quantizers.gptq` or another location? It might be needed by any layers that support GPTQ.

keras/src/layers/core/einsum_dense.py
```diff
@@ -1207,6 +1251,47 @@ def _get_gptq_group_size(self, config):
                 "`GPTQDTypePolicy` or the `config` argument."
             )
 
+    def _get_gptq_weight_bits(self, config):
```
james77777778: Same here.

keras/src/layers/core/dense.py
```diff
@@ -147,18 +147,39 @@ def kernel(self):
...
-            and self.quantization_mode == "gptq"
-        ):
-            return self.quantized_kernel
-        kernel = self._kernel
-        if self.quantization_mode == "int4":
-            kernel = quantizers.unpack_int4(kernel, self._orig_input_dim)
-        if self.lora_enabled:
-            return kernel + (self.lora_alpha / self.lora_rank) * ops.matmul(
+
+        mode = getattr(self, "quantization_mode", None)
```
james77777778: `Layer.quantization_mode`  is defined in `keras>=3.8.0` so it should be safe to use this attribute directly.

https://github.com/keras-team/keras/blob/v3.8.0/keras/src/layers/layer.py#L754

keras/src/layers/core/dense.py
```diff
@@ -147,18 +147,39 @@ def kernel(self):
...
+        is_gptq = mode == "gptq"
+        is_int4 = mode == "int4"
+        calibrated = bool(getattr(self, "is_gptq_calibrated", False))
+        gptq_bits = self._get_gptq_weight_bits(None) if is_gptq else None
+
+        # Decide the source tensor first (packed vs already-quantized vs plain
+        # kernel)
+        if is_gptq and calibrated and gptq_bits != 4:
+            # calibrated GPTQ, not 4-bit, no unpacking needed
+            kernel = self.quantized_kernel
```
james77777778, line 162: I'm not clear on why we have a separate `quantized_kernel` for GPTQ.
For int8 and int4 quant, we replace the original `_kernel` with the quantized one which is also called `_kernel`. Is it feasible to do the same for GPTQ?

Using a separate `quantized_kernel` seems to add unnecessary complexity.

keras/src/layers/core/dense.py
```diff
@@ -147,18 +147,39 @@ def kernel(self):
...
+            elif is_gptq and calibrated and gptq_bits == 4:
+                kernel = quantizers.unpack_int4(
+                    self.quantized_kernel,
+                    orig_len=self.units,
+                    axis=0,
+                    dtype="uint8",
+                )
+
+        # Apply LoRA once at the end.
+        if getattr(self, "lora_enabled", False):
```
james77777778: `self.lora_enabled` was introduced a long time ago so it should be safe to use it directly.

keras/src/layers/core/einsum_dense.py
```diff
@@ -209,20 +211,43 @@ def kernel(self):
...
-            return self.quantized_kernel
-        kernel = self._kernel
-        if self.quantization_mode == "int4":
-            kernel = quantizers.unpack_int4(
-                kernel, self._orig_length_along_pack_axis, self._int4_pack_axis
-            )
-        if self.lora_enabled:
-            return kernel + (self.lora_alpha / self.lora_rank) * ops.matmul(
+
+        mode = getattr(self, "quantization_mode", None)
```
james77777778: same here.

keras/src/layers/core/einsum_dense.py
```diff
@@ -209,20 +211,43 @@ def kernel(self):
...
+            elif is_gptq and calibrated and gptq_bits == 4:
+                kernel = quantizers.unpack_int4(
+                    self.quantized_kernel,
+                    orig_len=self.gptq_unpacked_column_size,
+                    axis=0,
+                    dtype="uint8",
+                )
+
+        # Apply LoRA if enabled
+        if getattr(self, "lora_enabled", False):
```
james77777778: and here.

james77777778 (commented):
Just some personal suggestions.

keras/src/layers/core/dense.py
JyotinderSingh: Fixed.

JyotinderSingh: Fixed.

keras/src/layers/core/dense.py
JyotinderSingh: Fixed.

keras/src/layers/core/einsum_dense.py
JyotinderSingh: Fixed.

keras/src/layers/core/einsum_dense.py
JyotinderSingh: Fixed.

keras/src/layers/core/dense.py
JyotinderSingh: A separate kernel variable is required since GPTQ requires access to the original high-precision `_kernel` during the calibration process, which happens separately from the `quantized_build` process.

We perform the calibration step after we have added the quantization-specific variables: https://github.com/keras-team/keras/blob/49797f28c3f5830b500184257038b0feb9369a34/keras/src/models/model.py#L476-L477

This process requires access to the high-precision kernel when instantiating a `GPTQ` object: https://github.com/keras-team/keras/blob/49797f28c3f5830b500184257038b0feb9369a34/keras/src/quantizers/gptq.py#L320-L322

And the `GPTQ` object accesses this kernel during the `quantize_and_correct_layer` call here: https://github.com/keras-team/keras/blob/49797f28c3f5830b500184257038b0feb9369a34/keras/src/quantizers/gptq.py#L438

We delete the high-precision kernel after we obtain the quantized kernel here: https://github.com/keras-team/keras/blob/49797f28c3f5830b500184257038b0feb9369a34/keras/src/quantizers/gptq.py#L479-L484

keras/src/layers/core/dense.py
JyotinderSingh: Makes sense, done.

keras/src/layers/core/einsum_dense.py
JyotinderSingh: Fixed.

JyotinderSingh:
Thanks for the review @james77777778, I've incorporated your suggestions into the PR.

I've rebased the PR on the latest master @hertschuh.

keras/src/layers/core/dense.py
james77777778: Thanks for the clarification. No issues on my end now.

hertschuh (approved):
Thanks for the PR!

PR #21687, opened 2025-09-23
[Keras 3 OpenVINO Backend]: Support numpy.sort

samthakur587:
hii @hertschuh @mattdangerw while implementing the numpy median i found that sort is not been implemented yet so i have created this PR could you please review  ?


hertschuh:
@rkazants

PR #21691, opened 2025-09-24
Implement logaddexp2 function in keras.ops

shashaka:
Adds keras.ops.logaddexp2, which computes the element-wise base-2 logarithm of the sum of exponentiations of two tensor inputs (x1 and x2).
Not supported on OpenVINO.


fchollet (approved):
Nice work, thanks for the contribution!

PR #21697, opened 2025-09-26
Core Data Structures & Communication Primitives for Tensor Parallel for Keras

buildwithsuhana:
This pull request marks the first foundational step in introducing a powerful, backend-agnostic auto-sharding capability for Tensor Parallelism in Keras 3.0. The ultimate goal is to democratize distributed training by allowing users to train models larger than a single device's memory with minimal code changes ideally just two lines to define and use the distribution strategy.

This initial PR does not introduce the full end-to-end logic but lays the critical groundwork by establishing the core data structures and communication abstractions that the rest of the system will be built upon.

**Key Components Introduced**
This PR introduces three fundamental building blocks for the auto-sharding framework:

1. state_actions_keras.py: Abstracting Sharding Operations
This module defines how tensor transformations for distribution should be represented.

StateActionKeras: An abstract base class that represents an action to be performed on a tensor, such as splitting it for a specific worker.

SplitKeras: A concrete implementation that defines the logic for splitting a tensor along a given dimension.

2. communications_keras.py: High-Level Communication Primitives
This module provides high-level wrappers for the collective communication operations essential for tensor parallelism.

AllReduceKeras, AllGatherKeras, BroadcastKeras, ScatterKeras: These classes encapsulate the logic for cross-device communication.

Purpose: They serve as a backend-agnostic interface for collective ops. 

3. config_keras.py: The Sharding Plan Data Structure
This module introduces the data class that will hold the complete, model-wide sharding configuration.

ConfigKeras: A dataclass designed to store the rules and actions for sharding both the model's weights (state_rules) and its outputs (output_rules).

Design Document: [Autosharding for Keras](https://docs.google.com/document/d/1lwV5HiHUryiHZW_MFvYwYkUd7A_h2WTlUdOWbesib-0/edit?resourcekey=0-fnUodKbP8oh0BBTUgoHyYQ&tab=t.0#heading=h.2pgbezpiqr4j)

### DIFF
--- a/keras/src/backend/distributed/__init__.py
+++ b/keras/src/backend/distributed/__init__.py
@@ -1,0 +1,6 @@
+# keras/src/backend/distributed/__init__.py
+
+from .base import BaseDistributedBackend
+from .factory import get_distributed_backend
+
+__all__ = ["get_distributed_backend", "BaseDistributedBackend"]
--- a/keras/src/backend/distributed/__init__.py
+++ b/keras/src/backend/distributed/__init__.py
@@ -1,5 +1,3 @@
-# keras/src/backend/distributed/__init__.py
-
 from .base import BaseDistributedBackend
 from .factory import get_distributed_backend
 
--- a/keras/src/backend/distributed/base.py
+++ b/keras/src/backend/distributed/base.py
@@ -1,0 +1,59 @@
+# keras/src/backend/distributed/base.py
+
+from abc import ABC
+from abc import abstractmethod
+from typing import Any
+from typing import List
+
+
+class BaseDistributedBackend(ABC):
+    """
+    Abstract Base Class for a distributed backend.
+    """
+
+    @abstractmethod
+    def get_tensor_lib(self):
+        """Get the appropriate tensor library for the backend."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def convert_to_backend_tensor(self, tensor: Any) -> Any:
+        """Convert a tensor to the appropriate backend format."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def compute_gradients(
+        self, loss: Any, trainable_vars: List[Any]
+    ) -> List[Any]:
+        """Compute gradients using the backend's automatic differentiation."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def apply_gradients(
+        self,
+        gradients: List[Any],
+        trainable_vars: List[Any],
+        learning_rate: float = 0.001,
+    ) -> None:
+        """Apply gradients to trainable variables."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def create_optimizer(self, optimizer_class: str, **kwargs):
+        """Create an optimizer for the backend."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_device_info(self) -> dict:
+        """Get information about available devices."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def is_multi_device_capable(self) -> bool:
+        """Check if the backend supports multi-device operations."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_communication_ops(self) -> dict:
+        """Get collective communication operations for the backend."""
+        raise NotImplementedError
--- a/keras/src/backend/distributed/base.py
+++ b/keras/src/backend/distributed/base.py
@@ -1,5 +1,3 @@
-# keras/src/backend/distributed/base.py
-
 from abc import ABC
 from abc import abstractmethod
 from typing import Any
--- a/keras/src/backend/jax/distributed_backend.py
+++ b/keras/src/backend/jax/distributed_backend.py
@@ -1,0 +1,141 @@
+import logging
+from typing import Any
+from typing import List
+
+import jax
+import jax.lax as lax
+import jax.numpy as jnp
+import optax
+
+from keras.src.backend.distributed.base import BaseDistributedBackend
+
+logger = logging.getLogger(__name__)
+
+
+class JaxDistributedBackend(BaseDistributedBackend):
+    """JAX-specific implementation of distributed operations."""
+
+    def get_tensor_lib(self):
+        return jnp
+
+    def convert_to_backend_tensor(self, tensor: Any) -> Any:
+        if hasattr(tensor, "numpy"):
+            return jnp.array(tensor.numpy())
+        else:
+            return jnp.array(tensor)
+
+    def compute_gradients(
+        self, loss: Any, trainable_vars: List[Any]
+    ) -> List[Any]:
+        def safe_convert_to_jax(tensor):
+            try:
+                if hasattr(tensor, "numpy"):
+                    if hasattr(tensor, "shape") and tensor.shape is None:
+                        logger.warning("Symbolic tensor detected")
+                        return jnp.array(0.0)
+                    else:
+                        return jnp.array(tensor.numpy())
+                else:
+                    return jnp.array(tensor)
+            except Exception as e:
+                logger.warning(
+                    f"Failed to convert tensor to JAX: {e}, using dummy value"
+                )
+                return jnp.array(0.0)
+
+        loss_jax = safe_convert_to_jax(loss)
+        params_jax = [safe_convert_to_jax(param) for param in trainable_vars]
+
+        def loss_fn(params):
+            return loss_jax
+
+        try:
+            gradients = jax.grad(loss_fn)(params_jax)
+            logger.info("   - JAX gradient computation successful")
+            return gradients
+        except Exception as e:
+            logger.warning(
+                f"JAX gradient computation failed: {e}, using fallback"
+            )
+            return [jnp.zeros_like(param) for param in params_jax]
+
+    def apply_gradients(
+        self,
+        gradients: List[Any],
+        trainable_vars: List[Any],
+        learning_rate: float = 0.001,
+    ) -> None:
+        for grad, var in zip(gradients, trainable_vars):
+            if grad is not None:
+                new_value = var - (learning_rate * grad)
+                if hasattr(var, "assign"):
+                    var.assign(new_value)
+
+    def create_optimizer(self, optimizer_class: str, **kwargs):
+        if optimizer_class.lower() == "adam":
+            return optax.adam(**kwargs)
+        elif optimizer_class.lower() == "sgd":
+            return optax.sgd(**kwargs)
+        else:
+            return optax.adam(learning_rate=0.001)
+
+    def get_device_info(self) -> dict:
+        info = {"backend": "jax", "devices": [], "device_count": 0}
+        try:
+            info["devices"] = [str(d) for d in jax.devices()]
+            info["device_count"] = jax.local_device_count()
+        except Exception as e:
+            logger.warning(f"Could not get device info for JAX: {e}")
+            info["devices"] = ["cpu"]
+            info["device_count"] = 1
+        return info
+
+    def is_multi_device_capable(self) -> bool:
+        return self.get_device_info()["device_count"] > 1
+
+    def get_communication_ops(self) -> dict:
+        def all_reduce_jax(x, op="sum", axis_name="data"):
+            return lax.pmean(x, axis_name=axis_name)
+
+        def all_gather_jax(x, axis=0, axis_name="model"):
+            return lax.all_gather(x, axis_name=axis_name, axis=axis)
+
+        def broadcast_jax(x, axis_name="data"):
+            return lax.all_gather(x, axis_name=axis_name, axis=0)
+
+        def scatter_jax(x, num_devices, axis_name="data"):
+            return lax.psplit(x, axis_name=axis_name, num_splits=num_devices)
+
+        def all_reduce_simulated(x, op="sum", axis_name="data"):
+            return jnp.sum(x, axis=0)
+
+        def all_gather_simulated(x, axis=0, axis_name="model"):
+            return jnp.concatenate([x, x], axis=axis)
+
+        def broadcast_simulated(x):
+            return x
+
+        def scatter_simulated(x, num_devices):
+            return jnp.split(x, num_devices, axis=0)
+
+        try:
+            if jax.device_count() > 1:
+                logger.info("Using real JAX collective communication ops.")
+                return {
+                    "all_reduce": all_reduce_jax,
+                    "all_gather": all_gather_jax,
+                    "broadcast": broadcast_jax,
+                    "scatter": scatter_jax,
+                }
+            else:
+                raise RuntimeError("Not running on multiple JAX devices.")
+        except (ImportError, RuntimeError) as e:
+            logger.warning(
+                f"JAX collective ops not available: {e}. Using SIMULATED ops."
+            )
+            return {
+                "all_reduce": all_reduce_simulated,
+                "all_gather": all_gather_simulated,
+                "broadcast": broadcast_simulated,
+                "scatter": scatter_simulated,
+            }
--- a/keras/src/backend/numpy/distributed_backend.py
+++ b/keras/src/backend/numpy/distributed_backend.py
@@ -1,0 +1,105 @@
+import logging
+from typing import Any
+from typing import List
+
+import numpy as np
+
+import keras
+from keras.src.backend.distributed.base import BaseDistributedBackend
+
+logger = logging.getLogger(__name__)
+
+
+class NumpyDistributedBackend(BaseDistributedBackend):
+    """NumPy-based fallback implementation of distributed operations."""
+
+    def get_tensor_lib(self):
+        return np
+
+    def convert_to_backend_tensor(self, tensor: Any) -> Any:
+        return keras.ops.convert_to_numpy(tensor)
+
+    def compute_gradients(
+        self, loss: Any, trainable_vars: List[Any]
+    ) -> List[Any]:
+        epsilon = 1e-7
+        gradients = []
+        for var in trainable_vars:
+            if hasattr(var, "shape"):
+                grad = np.zeros_like(var)
+                it = np.nditer(
+                    var, flags=["multi_index"], op_flags=["readwrite"]
+                )
+                while not it.finished:
+                    idx = it.multi_index
+                    original_value = var[idx]
+                    var[idx] = original_value + epsilon
+                    # This part is flawed as loss is a scalar.
+                    # Numerical differentiation needs a function to re-evaluate.
+                    # This is a placeholder for a no-op.
+                    loss_plus = loss
+                    var[idx] = original_value - epsilon
+                    loss_minus = loss
+                    grad[idx] = (loss_plus - loss_minus) / (
+                        2 * epsilon
+                    )  # Will be 0
+                    var[idx] = original_value  # Restore
+                    it.iternext()
+                gradients.append(grad)
+            else:
+                gradients.append(0.0)
+        return gradients
+
+    def apply_gradients(
+        self,
+        gradients: List[Any],
+        trainable_vars: List[Any],
+        learning_rate: float = 0.001,
+    ) -> None:
+        for grad, var in zip(gradients, trainable_vars):
+            if grad is not None:
+                new_value = var - (learning_rate * grad)
+                if hasattr(var, "assign"):
+                    var.assign(new_value)
+                else:
+                    var[:] = new_value
+
+    def create_optimizer(self, optimizer_class: str, **kwargs):
+        class NumpyOptimizer:
+            def __init__(self, learning_rate=0.001):
+                self.learning_rate = learning_rate
+
+            def apply_gradients(self, grads_and_vars):
+                for grad, var in grads_and_vars:
+                    if grad is not None:
+                        var -= self.learning_rate * grad
+
+        return NumpyOptimizer(**kwargs)
+
+    def get_device_info(self) -> dict:
+        return {"backend": "numpy", "devices": ["cpu"], "device_count": 1}
+
+    def is_multi_device_capable(self) -> bool:
+        return False
+
+    def get_communication_ops(self) -> dict:
+        logger.info("Using SIMULATED NumPy communication ops.")
+
+        def all_reduce_np(x, op="sum"):
+            return keras.ops.sum(x, axis=0)
+
+        def all_gather_np(x, axis=0):
+            return keras.ops.concatenate([x, x], axis=axis)
+
+        def broadcast_np(x):
+            return x
+
+        def scatter_np(x, num_devices):
+            return keras.ops.split(x, num_devices, axis=0)
+
+        return {
+            "all_reduce": all_reduce_np,
+            "all_gather": all_gather_np,
+            "broadcast": broadcast_np,
+            "scatter": scatter_np,
+        }
--- a/keras/src/distribution/tensor_parallel/config.py
+++ b/keras/src/distribution/tensor_parallel/config.py
@@ -1,0 +1,65 @@
+import dataclasses
+from typing import Any
+from typing import Dict
+from typing import Sequence
+
+from keras.src.backend.distributed import get_distributed_backend
+from keras.src.distribution.tensor_parallel.communications import AllGatherKeras
+from keras.src.distribution.tensor_parallel.communications import AllReduceKeras
+from keras.src.distribution.tensor_parallel.communications import BroadcastKeras
+
+
+@dataclasses.dataclass
+class ConfigKeras:
+    state_rules: Dict[str, Any]
+    output_rules: Dict[str, Any]
+
+    def create_collective_ops(
+        self, devices: Sequence[str], distributed: bool = True
+    ):
+        world_size = len(devices)
+        backend = get_distributed_backend()
+
+        # Pass the backend instance to the constructors
+        make_allreduce = lambda ws: AllReduceKeras(
+            ws, backend=backend, op="mean"
+        )
+        make_allgather = lambda ws, dim: AllGatherKeras(
+            ws, backend=backend, dim=dim
+        )
+        make_broadcast = lambda ws: BroadcastKeras(ws, backend=backend)
+
+        def create_collective_ops(rules: Dict[str, Any]) -> Dict[str, Any]:
+            result = {}
+            for pattern, actions in rules.items():
+                if isinstance(actions, dict):
+                    result[pattern] = {}
+                    for key, action in actions.items():
+                        if isinstance(action, str):
+                            if action == "sum":
+                                result[pattern][key] = make_allreduce(
+                                    world_size
+                                )
+                            elif action.startswith("gather"):
+                                dim = -1
+                                if " " in action:
+                                    dim = int(action.split(" ")[1])
+                                result[pattern][key] = make_allgather(
+                                    world_size, dim
+                                )
+                            elif action == "broadcast":
+                                result[pattern][key] = make_broadcast(
+                                    world_size
+                                )
+                            else:
+                                result[pattern][key] = action
+                        else:
+                            result[pattern][key] = action
+                else:
+                    result[pattern] = actions
+            return result
+
+        return dataclasses.replace(
+            self,
+            output_rules=create_collective_ops(self.output_rules),
+        )
--- a/keras/src/backend/distributed/factory.py
+++ b/keras/src/backend/distributed/factory.py
@@ -1,10 +1,7 @@
-# keras/src/backend/distributed/factory.py
-
 import logging
 
 from keras.src.backend.distributed.base import BaseDistributedBackend
 
-# Import all the concrete implementation classes
 from keras.src.backend.jax.distributed_backend import JaxDistributedBackend
 from keras.src.backend.numpy.distributed_backend import NumpyDistributedBackend
 from keras.src.backend.tensorflow.distributed_backend import (
--- a/keras/src/backend/distributed/factory.py
+++ b/keras/src/backend/distributed/factory.py
@@ -1,0 +1,53 @@
+# keras/src/backend/distributed/factory.py
+
+import logging
+
+from keras.src.backend.distributed.base import BaseDistributedBackend
+
+# Import all the concrete implementation classes
+from keras.src.backend.jax.distributed_backend import JaxDistributedBackend
+from keras.src.backend.numpy.distributed_backend import NumpyDistributedBackend
+from keras.src.backend.tensorflow.distributed_backend import (
+    TensorflowDistributedBackend,
+)
+from keras.src.backend.torch.distributed_backend import (
+    PytorchDistributedBackend,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def get_distributed_backend(
+    backend_name: str = "auto",
+) -> BaseDistributedBackend:
+    """
+    Factory to get the best available or a specific distributed backend.
+    """
+    if backend_name == "auto":
+        try:
+            logger.info("Auto-detected JAX for distributed backend.")
+            return JaxDistributedBackend()
+        except ImportError:
+            try:
+                logger.info("Auto-detected TensorFlow for distributed backend.")
+                return TensorflowDistributedBackend()
+            except ImportError:
+                try:
+                    logger.info(
+                        "Auto-detected PyTorch for distributed backend."
+                    )
+                    return PytorchDistributedBackend()
+                except ImportError:
+                    logger.warning("Using NumPy distributed backend.")
+                    return NumpyDistributedBackend()
+
+    elif backend_name == "jax":
+        return JaxDistributedBackend()
+    elif backend_name == "tensorflow":
+        return TensorflowDistributedBackend()
+    elif backend_name == "pytorch":
+        return PytorchDistributedBackend()
+    elif backend_name == "numpy":
+        return NumpyDistributedBackend()
+    else:
+        raise ValueError(f"Unknown distributed backend: {backend_name}")
--- a/keras/src/backend/tensorflow/distributed_backend.py
+++ b/keras/src/backend/tensorflow/distributed_backend.py
@@ -1,0 +1,139 @@
+import logging
+from typing import Any
+from typing import List
+
+import tensorflow as tf
+
+import keras
+from keras.src.backend.distributed.base import BaseDistributedBackend
+
+logger = logging.getLogger(__name__)
+
+
+class TensorflowDistributedBackend(BaseDistributedBackend):
+    """TensorFlow-specific implementation of distributed operations."""
+
+    def get_tensor_lib(self):
+        return tf
+
+    def convert_to_backend_tensor(self, tensor: Any) -> Any:
+        if hasattr(tensor, "numpy"):
+            return tf.convert_to_tensor(tensor.numpy())
+        else:
+            return tf.convert_to_tensor(tensor)
+
+    def compute_gradients(
+        self, loss: Any, trainable_vars: List[Any]
+    ) -> List[Any]:
+        with tf.GradientTape() as tape:
+            # TensorFlow's tape automatically watches trainable variables,
+            # but explicit watching is safer.
+            for var in trainable_vars:
+                tape.watch(var)
+
+        try:
+            # Assuming loss is already a tensor computed from watched variables
+            gradients = tape.gradient(loss, trainable_vars)
+            logger.info("   - TensorFlow gradient computation successful")
+            return gradients
+        except Exception as e:
+            logger.warning(
+                f"TensorFlow gradient computation failed: {e}, using fallback"
+            )
+            return [tf.zeros_like(var) for var in trainable_vars]
+
+    def apply_gradients(
+        self,
+        gradients: List[Any],
+        trainable_vars: List[Any],
+        learning_rate: float = 0.001,
+    ) -> None:
+        for grad, var in zip(gradients, trainable_vars):
+            if grad is not None:
+                new_value = var - (learning_rate * grad)
+                var.assign(new_value)
+
+    def create_optimizer(self, optimizer_class: str, **kwargs):
+        if optimizer_class.lower() == "adam":
+            return tf.keras.optimizers.Adam(**kwargs)
+        elif optimizer_class.lower() == "sgd":
+            return tf.keras.optimizers.SGD(**kwargs)
+        else:
+            return tf.keras.optimizers.Adam(learning_rate=0.001)
+
+    def get_device_info(self) -> dict:
+        info = {"backend": "tensorflow", "devices": [], "device_count": 0}
+        try:
+            info["devices"] = [
+                d.name for d in tf.config.list_physical_devices()
+            ]
+            info["device_count"] = len(tf.config.list_physical_devices())
+        except Exception as e:
+            logger.warning(f"Could not get device info for TensorFlow: {e}")
+            info["devices"] = ["cpu"]
+            info["device_count"] = 1
+        return info
+
+    def is_multi_device_capable(self) -> bool:
+        return self.get_device_info()["device_count"] > 1
+
+    def get_communication_ops(self) -> dict:
+        def all_reduce_tf(x, op="sum"):
+            strategy = tf.distribute.get_strategy()
+            return strategy.reduce(tf.distribute.ReduceOp.SUM, x, axis=0)
+
+        def all_gather_tf(x, axis=0):
+            strategy = tf.distribute.get_strategy()
+            return tf.raw_ops.AllGather(
+                input=x,
+                group_assignment=[
+                    [i for i in range(strategy.num_replicas_in_sync)]
+                ],
+                group_size=strategy.num_replicas_in_sync,
+            )
+
+        def broadcast_tf(x, root=0):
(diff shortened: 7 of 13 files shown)


keras/src/backend/distributed/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .base import BaseDistributedBackend
+from .factory import get_distributed_backend
```
JyotinderSingh: Using relative imports goes against keras code conventions. Please use absolute imports. Additionally, you shouldn't have to add these imports manually; pre-commit should handle these for you if you're using the `@keras.export` decorator for the exported objects.
```suggestion
from keras.src.backend.distributed.base import BaseDistributedBackend
from keras.src.backend.distributed.factory import get_distributed_backend
```

keras/src/backend/distributed/base.py
```diff
@@ -0,0 +1,57 @@
+from abc import ABC
+from abc import abstractmethod
+from typing import Any
+from typing import List
+
+
+class BaseDistributedBackend(ABC):
```
JyotinderSingh: nit: `Base` prefix may not be required
```suggestion
class DistributedBackend(ABC):
```

keras/src/backend/distributed/base.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @abstractmethod
+    def get_tensor_lib(self):
+        """Get the appropriate tensor library for the backend."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def convert_to_backend_tensor(self, tensor: Any) -> Any:
+        """Convert a tensor to the appropriate backend format."""
+        raise NotImplementedError
```
JyotinderSingh: keras already has `ops.convert_to_tensor` for this.

keras/src/backend/distributed/base.py
```diff
@@ -0,0 +1,57 @@
...
+
+class BaseDistributedBackend(ABC):
+    """
+    Abstract Base Class for a distributed backend.
+    """
+
+    @abstractmethod
+    def get_tensor_lib(self):
+        """Get the appropriate tensor library for the backend."""
+        raise NotImplementedError
```
JyotinderSingh: Why do we need this? Can't simply use `keras.backend.numpy`?

keras/src/backend/jax/distributed_backend.py
```diff
@@ -0,0 +1,172 @@
...
+    def get_device_info(self) -> dict:
+        info = {"backend": "jax", "devices": [], "device_count": 0}
+        try:
+            info["devices"] = [str(d) for d in jax.devices()]
+            info["device_count"] = jax.local_device_count()
+        except Exception as e:
+            logger.warning(f"Could not get device info for JAX: {e}")
+            info["devices"] = ["cpu"]
+            info["device_count"] = 1
+        return info
```
JyotinderSingh: We can easily restructure this in the form of simple conditional checks instead of depending on try-catch. Exceptions should never be used in place of conditional flow since it reduces readability and makes the behavior harder to predict.

keras/src/backend/jax/distributed_backend_test.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def assign(self, new_value):
+        self.value = jnp.array(new_value)
+
+    def __sub__(self, other):
+        return self.value - other
+
+    @property
+    def __array_interface__(self):
+        return self.value.__array_interface__
```
JyotinderSingh: I don't think we need to create a Mock variable class. Why not just use a regular variable?

keras/src/backend/jax/distributed_backend_test.py
```diff
@@ -0,0 +1,169 @@
...
+        logging.disable(logging.NOTSET)
+
+    def test_get_tensor_lib(self):
+        """Test if the correct tensor library (jnp) is returned."""
+        self.assertIs(self.backend.get_tensor_lib(), jnp)
+
+    def test_convert_to_backend_tensor(self):
+        """Test tensor conversion from various types to JAX arrays."""
+        py_list = [1.0, 2.0, 3.0]
+        jax_tensor = self.backend.convert_to_backend_tensor(py_list)
```
JyotinderSingh: We can just use `ops.array` to create a tensor.

keras/src/backend/jax/distributed_backend_test.py
```diff
@@ -0,0 +1,169 @@
...
+class TestJaxDistributedBackend(unittest.TestCase):
+    """Unit tests for the JaxDistributedBackend class."""
+
+    def setUp(self):
+        """Set up the test case by instantiating the backend."""
+        self.backend = JaxDistributedBackend()
+
+    def tearDown(self):
+        """Re-enable logging after tests are done."""
+        logging.disable(logging.NOTSET)
```
JyotinderSingh: I'm not sure this is needed. We almost never create logs in our tests.

keras/src/backend/jax/distributed_backend_test.py
```diff
@@ -0,0 +1,169 @@
...
+        self.assertIsInstance(jax_tensor, jnp.ndarray)
+        np.testing.assert_array_equal(jax_tensor, jnp.array([1.0, 2.0, 3.0]))
+
+        np_array = np.array([4.0, 5.0, 6.0])
+        jax_tensor = self.backend.convert_to_backend_tensor(np_array)
+        self.assertIsInstance(jax_tensor, jnp.ndarray)
+        np.testing.assert_array_equal(jax_tensor, jnp.array([4.0, 5.0, 6.0]))
+
+    def test_compute_gradients_returns_zeros(self):
+        loss = jnp.array(10.0)
```
JyotinderSingh: Use `ops.array`. Same for other places.

keras/src/backend/jax/distributed_backend_test.py
```diff
@@ -0,0 +1,169 @@
...
+
+        gradients = self.backend.compute_gradients(loss, trainable_vars)
+
+        self.assertEqual(len(gradients), 2)
+        np.testing.assert_array_equal(
+            gradients[0], jnp.zeros_like(trainable_vars[0])
+        )
+        np.testing.assert_array_equal(
+            gradients[1], jnp.zeros_like(trainable_vars[1])
+        )
```
JyotinderSingh: use the assertion helpers from `keras.src.testing` to be consistent with the repository's testing conventions.

keras/src/backend/jax/distributed_backend_test.py
```diff
@@ -0,0 +1,169 @@
...
+        """Test tensor conversion from various types to JAX arrays."""
+        py_list = [1.0, 2.0, 3.0]
+        jax_tensor = self.backend.convert_to_backend_tensor(py_list)
+        self.assertIsInstance(jax_tensor, jnp.ndarray)
+        np.testing.assert_array_equal(jax_tensor, jnp.array([1.0, 2.0, 3.0]))
+
+        np_array = np.array([4.0, 5.0, 6.0])
+        jax_tensor = self.backend.convert_to_backend_tensor(np_array)
+        self.assertIsInstance(jax_tensor, jnp.ndarray)
+        np.testing.assert_array_equal(jax_tensor, jnp.array([4.0, 5.0, 6.0]))
```
JyotinderSingh: use the assertion helpers from `keras.src.testing` to be consistent with the repository's testing conventions.

JyotinderSingh: replace `jnp.array` with `ops.array`

keras/src/backend/jax/distributed_backend_test.py
```diff
@@ -0,0 +1,169 @@
...
+    def test_get_tensor_lib(self):
+        """Test if the correct tensor library (jnp) is returned."""
+        self.assertIs(self.backend.get_tensor_lib(), jnp)
+
+    def test_convert_to_backend_tensor(self):
+        """Test tensor conversion from various types to JAX arrays."""
+        py_list = [1.0, 2.0, 3.0]
+        jax_tensor = self.backend.convert_to_backend_tensor(py_list)
+        self.assertIsInstance(jax_tensor, jnp.ndarray)
+        np.testing.assert_array_equal(jax_tensor, jnp.array([1.0, 2.0, 3.0]))
```
JyotinderSingh: use the assertion helpers from `keras.src.testing` to be consistent with the repository's testing conventions.

keras/src/backend/jax/distributed_backend_test.py
```diff
@@ -0,0 +1,169 @@
...
+        grad2 = jnp.array(0.5)
+        gradients = [grad1, grad2]
+        learning_rate = 0.1
+        self.backend.apply_gradients(gradients, trainable_vars, learning_rate)
+
+        expected_var1 = np.array([1.0 - 0.1 * 0.1, 2.0 - 0.1 * 0.2])
+        expected_var2 = 5.0 - 0.1 * 0.5
+
+        np.testing.assert_allclose(var1.value, expected_var1, atol=1e-6)
+        np.testing.assert_allclose(var2.value, expected_var2, atol=1e-6)
```
JyotinderSingh: use the assertion helpers from `keras.src.testing` to be consistent with the repository's testing conventions. Same for all others.

keras/src/backend/jax/distributed_backend_test.py
```diff
@@ -0,0 +1,169 @@
...
+            "get_device_info",
+            return_value={
+                "backend": "jax",
+                "devices": ["cpu:0", "cpu:1"],
+                "device_count": 2,
+            },
+        ):
+            with patch.object(
+                self.backend, "is_multi_device_capable", return_value=False
+            ):
```
JyotinderSingh: Can't we use real implementations instead of using mocks? We can restrict this test to run only on TPUs if hardware is a concern.
Based on my current understanding this anyway doesn't test the real distributed implementation, and instead targets the simulated algorithm

keras/src/backend/numpy/distributed_backend.py
JyotinderSingh, line 1: Do we really need a numpy implementation of this algorithm? Numpy exists largely as a toy-backend in keras and doesn't even support training.

keras/src/backend/jax/distributed_backend_test.py
```diff
@@ -0,0 +1,169 @@
...
+                x_scatter = jnp.array([[1, 2], [3, 4], [5, 6], [7, 8]])
+                scattered = ops["scatter"](x_scatter)
+                expected_scatter = keras.ops.split(
+                    x_scatter, simulated_world_size, axis=0
+                )[0]
+                np.testing.assert_allclose(scattered, expected_scatter)
+
+
+if __name__ == "__main__":
+    unittest.main(argv=["first-arg-is-ignored"], exit=False)
```
JyotinderSingh: We don't use `unittest` for our tests. This block is not required.

keras/src/distribution/tensor_parallel/config.py
```diff
@@ -0,0 +1,69 @@
...
+                                result[pattern][key] = make_broadcast(
+                                    world_size
+                                )
+                            else:
+                                result[pattern][key] = action
+                        else:
+                            result[pattern][key] = action
+                else:
+                    result[pattern] = actions
+            return result
```
JyotinderSingh: Such extremely nested blocks are very hard to understand and debug. Can you try reducing the nesting? It's okay to introduce smaller helpers and repeated, un-nested for-loops

keras/src/distribution/tensor_parallel/config.py
```diff
@@ -0,0 +1,69 @@
...
+        )
+        make_allreduce_mean = lambda ws: AllReduceKeras(
+            ws, backend=backend, op="mean"
+        )
+        make_allgather = lambda ws, dim: AllGatherKeras(
+            ws, backend=backend, dim=dim
+        )
+        make_broadcast = lambda ws: BroadcastKeras(ws, backend=backend)
+
+        def create_collective_ops(rules: Dict[str, Any]) -> Dict[str, Any]:
```
JyotinderSingh: Please add a docstring here.

JyotinderSingh (changes requested):

keras/src/backend/jax/distributed_backend.py
```diff
@@ -0,0 +1,248 @@
...
+
+import jax
+import jax.lax as lax
+import jax.numpy as jnp
+
+import keras
+
+
+def compute_gradients(
+    _loss: jnp.ndarray, trainable_vars: List[jnp.ndarray]
```
hertschuh: This signature doesn't work for JAX. You cannot take the gradient of a tensor. You can only transform a function so that you can take its gradient.

keras/src/backend/jax/distributed_backend.py
```diff
@@ -0,0 +1,248 @@
...
+) -> List[jnp.ndarray]:
+    """Computes gradients of the loss with respect to trainable variables.
+
+    Note: This is a placeholder implementation that returns zeros. A real
+    implementation would use `jax.grad`.
```
hertschuh: So why are we doing this if it's not a real implementation?

keras/src/backend/jax/distributed_backend.py
```diff
@@ -0,0 +1,248 @@
...
+) -> List[jnp.ndarray]:
+    """Applies gradients and returns the updated variables."""
+    updated_vars = []
+    for grad, var in zip(gradients, trainable_vars):
+        if grad is not None:
+            new_var = var - (learning_rate * grad)
+            updated_vars.append(new_var)
+        else:
+            updated_vars.append(var)
+    return updated_vars
```
hertschuh: This is an inline implementation of SGD. Why is this needed?

keras/src/backend/jax/distributed_backend.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+def create_optimizer(optimizer_class: str, **kwargs) -> Dict[str, Any]:
```
hertschuh: Why do we need this? Does this mean that TensorParallel won't work with Keras optimizers?

hertschuh (commented):
Thank you for the PR!

Some high level comments:

- Out of context, it's really hard for me to understand why these abstractions are needed for Tensor Parallel.
     - Why do we need all these primitives? 
     - Why do we need 3 layers of abstraction for the same concepts: the `communications` layer, the `state_actions` layer and the `keras.distributed.get_communication_ops` layer? Can we just have one?
- These abstraction look Torch-like and not JAX-like. On JAX you never have to manually split and do an all-gather, you simply shard. You never have to explicitly have to do a "collective sum". You just do a sum, and if the tensors are sharded, it will magically do all the needed collectives for you. So it's unclear to me why any of these are needed for JAX.
- I wouldn't export these symbols that you added to `keras.distributed`, I don't think they are needed. What we'll expose is the "Tensor Parallel" API.
- For the better or worse, we don't do type annotations in Keras. And unfortunately, mixing code with type annotations with code without type annotation doesn't work well. It's better to not have any type annotations at all.

keras/src/backend/jax/distributed_backend.py
```diff
@@ -0,0 +1,95 @@
+import jax
+import jax.lax as lax
+
+
+def get_device_info():
```
hertschuh: Arguably, this use case is already covered by `keras.distribution.list_devices()`.

Also, I don't see this being used by subsequent PRs. Did I miss it?

keras/src/backend/jax/distributed_backend.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    available_devices = jax.devices()
+    return {
+        "backend": "jax",
+        "devices": [str(d) for d in available_devices],
+        "device_count": len(available_devices),
+    }
+
+
+def is_multi_device_capable():
```
hertschuh: Arguably, this use case is already covered by `keras.distribution.list_devices()`.

Also, I wouldn't call it `is_multi_device_capable` because JAX is always "multi device capable". The real question is whether there are multiple devices.

How about having `get_device_count` instead, this could be more useful because you get the actual number. Then you can export it as `keras.distribution.get_device_count`.

keras/src/backend/jax/distributed_backend.py
```diff
@@ -0,0 +1,95 @@
...
+    """Checks if more than one JAX device is available for computation.
+
+    Returns:
+        bool: True if the local JAX environment has more than one device,
+        False otherwise.
+    """
+    return jax.local_device_count() > 1
+
+
+def get_communication_ops():
```
hertschuh: I think it would be simpler to have them as regular ops, probably in `core.py`. But also, they're only used once each. I'll check in the subsequent PRs whether this abstraction is actually needed.

keras/src/backend/jax/distributed_backend.py
```diff
@@ -0,0 +1,95 @@
+import jax
```
hertschuh: Why not put these in `distribution_lib.py`? All `backend/jax/distribution_lib` is already a backend specific implementation, so we don't need call it `distribution_backend`.

keras/src/distribution/tensor_parallel/tensor_layout.py
```diff
@@ -0,0 +1,169 @@
...
+            NotImplementedError: This is an abstract method and must be
+                implemented by subclasses.
+
+        Returns:
+            A shard or transformation of the input tensor specific to the given
+            rank.
+        """
+        raise NotImplementedError
+
+    def undo(self, tensors):
```
hertschuh: I don't see `undo` being used in the subsequent PRs. What is it for?

keras/src/distribution/tensor_parallel/tensor_layout.py
```diff
@@ -0,0 +1,169 @@
...
+            The single tensor reconstructed by concatenating the shards.
+        """
+        if self.dim == -1:
+            dim = keras.ops.ndim(tensors[0]) - 1
+        else:
+            dim = self.dim
+        return keras.ops.concatenate(tensors, axis=dim)
+
+
+class Split(_ConcatenateMixin, LayoutAction):
```
hertschuh: `Split` is the only implementation of `LayoutAction` and the only use of `_ConcatenateMixin`, and it turns out `undo` is not used. So simplify by
- removing `LayoutAction`
- removing `_ConcatenateMixin`

keras/src/distribution/tensor_parallel/tensor_layout.py
```diff
@@ -0,0 +1,169 @@
...
+        """Initializes the LayoutMap.
+
+        Args:
+            state_rules: A dictionary of distribution rules for model states.
+            output_rules: A dictionary of distribution rules for model outputs.
+        """
+        self.state_rules = state_rules
+        self.output_rules = output_rules
+
+    def create_collective_ops(self, devices):
```
hertschuh: This always returns `self`, is never overridden, is never chained, and is only used once. Simply remove.

keras/src/distribution/tensor_parallel/tensor_layout.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def __init__(self, state_rules, output_rules):
+        """Initializes the LayoutMap.
+
+        Args:
+            state_rules: A dictionary of distribution rules for model states.
+            output_rules: A dictionary of distribution rules for model outputs.
+        """
+        self.state_rules = state_rules
+        self.output_rules = output_rules
```
hertschuh: All this class does is group two things. You can simply replace by:

```python
LayoutMap = collections.namedtuple("LayoutMap", ["state_rules", "output_rules"])
```

keras/src/backend/jax/distribution_lib.py
```diff
@@ -27,6 +27,37 @@ def list_devices(device_type=None):
...
+    """Returns the number of local JAX devices.
+
+    Returns:
+        int: The total number of devices configured in the current distribution
+             strategy.
+    """
+    return jax.local_device_count()
+
+
+def get_best_devices(count=1):
```
hertschuh: I don't think this is needed. It's only used here:

```python
    def _auto_detect_parallelism(self):
        """Auto-detects the number of available devices and sets world size."""
        from keras.src.distribution import get_best_devices

        available_devices = list_devices()
        world_size = len(available_devices)

        device_ids = get_best_devices(world_size)

        return world_size, device_ids
```

Which you should replace with:
```python
    def _auto_detect_parallelism(self):
        available_devices = list_devices()
        return len(available_devices), available_devices
```

buildwithsuhana: Done

keras/src/backend/jax/core.py
```diff
@@ -529,6 +530,63 @@ def remat(f):
...
+    Returns:
+        The result of the all-reduce operation, with the same shape as the
+        input `x`.
+    """
+    if op == "sum":
+        return lax.psum(x, axis_name=axis_name)
+    elif op == "mean":
+        sum_val = lax.psum(x, axis_name=axis_name)
+        axis_size = lax.psum(1, axis_name=axis_name)
+        return sum_val / axis_size
```
hertschuh: Why not use `jax.lax.pmean`?

keras/src/backend/jax/core.py
```diff
@@ -529,6 +530,63 @@ def remat(f):
...
+        axis: The dimension along which to concatenate the gathered tensors.
+        axis_name: The name of the distribution axis (e.g., "model",
+                                   "data") over which to perform the gather.
+                                   Defaults to "model".
+
+    Returns:
+        The gathered tensor, which will have a larger size along `axis`
+        dimension.
+    """
+    return lax.all_gather(x, axis_name=axis_name, axis=axis, tiled=True)
```
hertschuh: I have a general question. Is autosharding supposed to work for multi-host distribution or only single host?

keras/src/backend/jax/distribution_lib.py
```diff
@@ -27,6 +27,16 @@ def list_devices(device_type=None):
     return [f"{device.platform}:{device.id}" for device in jax_devices]
 
 
+def get_device_count():
+    """Returns the number of local JAX devices.
+
+    Returns:
+        int: The total number of devices configured in the current distribution
+             strategy.
```
hertschuh: The documentation line 31 is correct. The documentation here is misleading. Maybe there is no current distribution strategy configured yet. This API works regardless of whether there is a distribution strategy.

```
   Returns: int representing the number of JAX devices.
```

keras/src/distribution/distribution_lib.py
```diff
@@ -39,6 +39,19 @@ def list_devices(device_type=None):
     return distribution_lib.list_devices(device_type)
 
 
+@keras_export("keras.distribution.get_device_count")
+def get_device_count():
+    """
+    Returns the total number of devices (e.g., GPUs, TPUs) available for the
+    current distribution strategy.
```
hertschuh: In the same way as `list_devices`, this is unrelated to any distribution strategy, it is purely wrapping the backend implementation.

Also:
- please add `device_type` like `list_devices` for consistency (I should have caught this earlier).
- format wise we always do this so that the documentation is formatted properly:
```Python
"""Single line documenting the function / class.

More details after a blank line.
"""

File: keras/src/distribution/distribution_lib.py [hunk 35]
```diff
@@ -39,6 +39,19 @@ def list_devices(device_type=None):
...
+
+    Returns:
+        int: The total number of devices configured in the current distribution
```
hertschuh: Same comment, don't mention "distribution strategy".

keras/src/distribution/tensor_parallel/tensor_layout.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def __init__(self, device_count, dim, sharding_type="auto"):
+        """Initializes the Split action.
+
+        Args:
+            device_count: The total number of workers/shards.
+            dim: The dimension along which to split the tensor. If -1, the
+                last dimension is used.
+            sharding_type: If `dim` is -1, this can be 'row' (dim=0) or
+                'column' (dim=1) to infer the split axis for 2D tensors.
```
hertschuh: Oh we use a different format for documentation. All of this should move inside the documentation of the class line 14.

keras/src/distribution/tensor_parallel/tensor_layout.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class Split:
+    """Splits a tensor into shards along a specified dimension.
+
+    This is an internal utility used by a higher-level distribution API.
+    It implements sharding by slicing a tensor along one of its axes.
+    It handles cases where the dimension size is not perfectly divisible by the
+    number of workers by distributing the remainder elements one by one to the
+    first few workers.
```
hertschuh: nitpick: this class is purely a "smart split" that can handle non-divisible dimensions gracefully, but it doesn't actually shard. So mentioning "sharding" and "workers" here is a bit misleading.

Additionally, we shouldn't use the term "workers" because it's usually for multiple processes or multiple hosts, but what we're doing here is multiple devices connected to a single host and there is no "workers", at least not on JAX.

keras/src/distribution/tensor_parallel/tensor_layout.py
```diff
@@ -0,0 +1,67 @@
...
+        self.dim = dim
+        self.sharding_type = sharding_type
+
+        if dim == -1 and sharding_type != "auto":
+            if sharding_type == "row":
+                self.dim = 0
+            elif sharding_type == "column":
+                self.dim = 1
+
+    def __call__(self, tensor, rank):
```
hertschuh: Actually, why is this a class and not just a function?

keras/src/distribution/tensor_parallel/tensor_layout.py
```diff
@@ -0,0 +1,67 @@
...
+        """Splits the tensor and returns the shard corresponding to the rank.
+
+        This method calculates the correct slice of the tensor for a given
+        worker rank, handling uneven distributions gracefully.
+
+        Args:
+            tensor: The full tensor to be sharded.
+            rank: The rank of the worker for which to get the shard.
+
```
hertschuh: `rank` is a bit confusing here, normally we only use the term for the rank of a tensor. How about `index`?

Same comment about the term worker. Just say `index: The index of the slice to return`.

keras/src/distribution/tensor_parallel/tensor_layout.py
```diff
@@ -0,0 +1,67 @@
...
+                self.dim = 1
+
+    def __call__(self, tensor, rank):
+        """Splits the tensor and returns the shard corresponding to the rank.
+
+        This method calculates the correct slice of the tensor for a given
+        worker rank, handling uneven distributions gracefully.
+
+        Args:
+            tensor: The full tensor to be sharded.
```
hertschuh: Don't use the term "sharded", the tensor is not getting sharded, only split.

`tensor: the full tensor to get a slice of.`

keras/src/distribution/tensor_parallel/tensor_layout.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+
+    def __init__(self, device_count, dim, sharding_type="auto"):
+        """Initializes the Split action.
+
+        Args:
+            device_count: The total number of workers/shards.
+            dim: The dimension along which to split the tensor. If -1, the
+                last dimension is used.
+            sharding_type: If `dim` is -1, this can be 'row' (dim=0) or
```
hertschuh: Why have both `sharding_type` and `dim`? It makes things more complicated. Simply, dim should be one of `0`, `1` or `-1`.

keras/src/distribution/tensor_parallel/tensor_layout.py
```diff
@@ -0,0 +1,67 @@
+import collections
+
+from keras.src import ops
+
+
+class Split:
```
hertschuh: Actually, looking at it again, I thought of two things:
1. This exists as a NumPy API: [array_split](https://docs.jax.dev/en/latest/_autosummary/jax.numpy.array_split.html), so we should just make this a normal keras op that can be used for other purposes.
2. Support for uneven shards is limited. Intermediary values can be sharded unevenly, but the outputs of a jitted function must be evenly sharded. Can you make sure this is not a blocker for this project?
    [1] https://github.com/jax-ml/jax/issues/26946#issuecomment-2701742580
    [2] https://docs.jax.dev/en/latest/_autosummary/jax.lax.with_sharding_constraint.html

keras/src/backend/jax/core.py
buildwithsuhana: Yes, this is designed to work for both single-host and multi-host distribution.

keras/src/backend/jax/numpy.py
```diff
@@ -1167,6 +1167,10 @@ def split(x, indices_or_sections, axis=0):
     return jnp.split(x, indices_or_sections, axis=axis)
 
 
+def array_split(x, indices_or_sections, axis=0):
+    return jnp.array_split(x, indices_or_sections, axis=axis)
```
hertschuh, line 1172: Add `x = convert_to_tensor(x)`

keras/src/ops/numpy.py
```diff
@@ -7762,3 +7762,120 @@ def histogram(x, bins=10, range=None):
...
+        if not isinstance(indices_or_sections, int):
+            raise TypeError(
+                "Argument `indices_or_sections` must be of type `int`. "
+                f"Received: {indices_or_sections}"
+            )
+        if indices_or_sections <= 0:
+            raise ValueError(
+                "Argument `indices_or_sections` must be a positive integer. "
+                f"Received: {indices_or_sections}"
+            )
```
hertschuh: This is already done lines 7861-7870. You don't need it again here because the only way to instantiate this class is via line 7873.

So remove this block.

keras/src/ops/numpy.py
```diff
@@ -7762,3 +7762,120 @@ def histogram(x, bins=10, range=None):
...
+        if not isinstance(axis, int):
+            raise TypeError(
+                f"Argument `axis` must be of type `int`. Received: {axis}"
+            )
```
hertschuh: Move to line 7871.

keras/src/ops/numpy.py
```diff
@@ -7762,3 +7762,120 @@ def histogram(x, bins=10, range=None):
...
+
+        self.indices_or_sections = indices_or_sections
+        self.axis = axis
+
+    def call(self, x):
+        # Call the backend's array_split implementation directly.
+        # It handles the logic for uneven splits.
```
hertschuh: This type of statement is true for any feature or any operation and any backend, so remove this comment.

keras/src/ops/numpy.py
```diff
@@ -7762,3 +7762,120 @@ def histogram(x, bins=10, range=None):
...
+    have size `quotient`.
+
+    Args:
+        x: Input tensor.
+        indices_or_sections: An integer indicating the number of
+            sub-arrays to create.
+        axis: The axis along which to split. Defaults to 0.
+
+    Returns:
+        A tuple of sub-tensors.
```
hertschuh: It should actually be a list of tensors per NumPy. So change the documentation and implementation as needed (although it looks like the implementations already return lists).

keras/src/ops/numpy.py
```diff
@@ -7762,3 +7762,120 @@ def histogram(x, bins=10, range=None):
...
+
+        output_specs = []
+        base_shape = list(x.shape)
+        for i in range(num_splits):
+            size = split_size + (1 if i < remainder else 0)
+            shape = base_shape.copy()
+            shape[axis] = size
+            output_specs.append(KerasTensor(shape=tuple(shape), dtype=x.dtype))
+
+        return tuple(output_specs)
```
hertschuh: Because the op itself returns a list, return a list here:
`return list(output_specs)`

keras/src/ops/numpy.py
```diff
@@ -7762,3 +7762,120 @@ def histogram(x, bins=10, range=None):
...
+            f"Received: indices_or_sections={indices_or_sections}"
+        )
+
+    if any_symbolic_tensors((x,)):
+        return ArraySplit(
+            indices_or_sections=indices_or_sections, axis=axis
+        ).symbolic_call(x)
+
+    # The eager path should also call the backend's array_split.
+    # The original implementation was incorrect.
```
hertschuh: Remove this comment.

keras/src/ops/numpy_test.py
```diff
@@ -9401,3 +9401,75 @@ def call(self, x):
...
+        ("split_by_size_zero_length_non_empty_split", 1, 2, 0, (1, 0)),
+        ("3D_axis_neg1", (2, 3, 7), 3, -1, ((2, 3, 3), (2, 3, 2), (2, 3, 2))),
+    )
+    def test_array_split_static_shape_and_correctness(
+        self, shape, indices_or_sections, axis, expected_shapes
+    ):
+        if isinstance(shape, int):
+            input_array = np.arange(shape)
+        else:
+            input_array = np.arange(np.prod(shape)).reshape(shape)
```
hertschuh: It's actually quite confusion to have ints in the parameters when it's a shape. So remove the `if` here and instead turn the ints to tuples of one value in the parameters:

```
("even_split_1D", (10,), 2, 0, (5, 5)),
```

This way it makes it obvious it's a shape.

keras/src/ops/numpy_test.py
```diff
@@ -9401,3 +9401,75 @@ def call(self, x):
         model.compile(jit_compile=jit_compile)
 
         model.predict(np.random.randn(1, 8))
+
+
+class NumPyArraySplitTest(testing.TestCase):
+    @parameterized.named_parameters(
```
hertschuh: Please follow the same pattern as the other tests. Look for `test_split` in this file and do something similar, i.e.:

`NumpyOneInputOpsDynamicShapeTest::test_array_split` which tests with Keras Tensors and a None dimension where you split.
`NumpyOneInputOpsDynamicShapeTest::test_array_split`which tests with keras Tensors.
`NumpyOneInputOpsCorrectnessTest::test_array_split` which tests with values.
`NumpyDTypeTest::test_array_split` which tests with different dtypes.

Except for the `NumpyDTypeTest`, we usually don't parameterize tests because each test is one line anyway, so it's fewer lines of code:

```python
    def test_split(self):
        x = np.array([[1, 2, 3], [3, 2, 1]])
        self.assertIsInstance(knp.split(x, 2), list)
        self.assertAllClose(knp.split(x, 2), np.split(x, 2))
        self.assertAllClose(knp.Split(2)(x), np.split(x, 2))
        ...
```

keras/src/backend/jax/distribution_lib.py
```diff
@@ -27,6 +27,21 @@ def list_devices(device_type=None):
...
+    Args:
+        device_type: Optional device type to count (e.g., "cpu", "gpu", "tpu").
+            If `None`, it counts all available devices.
+
+    Returns:
+        int: The total number of JAX devices for the specified type.
+    """
+    device_type = device_type.lower() if device_type else None
+    jax_devices = jax.devices(backend=device_type)
+    return len(jax_devices)
```
hertschuh: Why not use `jax.device_count(device_type)`?

keras/src/backend/jax/distribution_lib.py
```diff
@@ -27,6 +27,21 @@ def list_devices(device_type=None):
     return [f"{device.platform}:{device.id}" for device in jax_devices]
 
 
+def get_device_count(device_type=None):
+    """Returns the number of available JAX devices.
+
+    Args:
+        device_type: Optional device type to count (e.g., "cpu", "gpu", "tpu").
+            If `None`, it counts all available devices.
```
hertschuh: That's actually not the behavior. If you pass nothing, you either get the number of GPUs or TPUs, or failing that, the number of CPUs. You never get the number of GPUs + CPUs for instance.

In any case, make sure this documentation and the documentation of `list_devices` is consistent.

hertschuh (changes requested):
Sorry for creating more work, but can you split this PR in 3 PRs.
- one for `split_array`.
- one for `get_device_count`.
- one for everything else for now.

Thanks! These are independent features that make sense on their own.

Additionally, I had a question in an earlier review:

On JAX, support for uneven shards is limited. Intermediary values can be sharded unevenly, but the outputs of a jitted function must be evenly sharded. Can you make sure this is not a blocker for this project?
    [1] https://github.com/jax-ml/jax/issues/26946#issuecomment-2701742580
    [2] https://docs.jax.dev/en/latest/_autosummary/jax.lax.with_sharding_constraint.html

keras/src/ops/numpy_test.py
```diff
@@ -2481,6 +2488,14 @@ def test_view(self):
         self.assertEqual(knp.view(x, dtype="int32").shape, (2, 2))
         self.assertEqual(knp.view(x, dtype="int32").dtype, "int32")
 
+    def test_array_split(self):
+        x = KerasTensor((6, 4))
+        splits = knp.array_split(x, 3, axis=0)
```
hertschuh: Oh, can you change this to an uneven split case, like 8 split in 3.

keras/src/backend/torch/numpy.py
```diff
@@ -1539,6 +1539,13 @@ def split(x, indices_or_sections, axis=0):
     return list(out)
 
 
+def array_split(x, indices_or_sections, axis=0):
+    x = convert_to_tensor(x)
+    axis_int = int(axis)
```
hertschuh: You already assert that axis is an `int` in the cross-backend `array_split`, so this is not needed.

keras/src/backend/openvino/numpy.py
```diff
@@ -2014,6 +2014,73 @@ def split(x, indices_or_sections, axis=0):
...
+    if not isinstance(indices_or_sections, int):
+        raise TypeError(
+            "Argument `indices_or_sections` must be of type `int`. "
+            f"Received: {indices_or_sections}"
+        )
+    if indices_or_sections <= 0:
+        raise ValueError(
+            "Argument `indices_or_sections` must be a positive integer. "
+            f"Received: {indices_or_sections}"
+        )
```
hertschuh: You already verify these in the cross-backend `array_split`, so this is not needed.

hertschuh: You already verify these in the cross-backend `array_split`, so this is not needed.

keras/src/backend/openvino/numpy.py
```diff
@@ -2014,6 +2014,73 @@ def split(x, indices_or_sections, axis=0):
...
+    ).output(0)
+
+    remainder_bcast = ov_opset.broadcast(remainder, splits_shape).output(0)
+
+    add_one_mask = ov_opset.less(range_splits, remainder_bcast).output(0)
+
+    add_one_values = ov_opset.convert(add_one_mask, Type.i64).output(0)
+
+    split_lengths = ov_opset.add(all_splits_base, add_one_values).output(0)
+    splits = ov_opset.variadic_split(x, axis_tensor, split_lengths)
```
hertschuh: This looks correct. But since everything is `int`s (`axis`, `indices_or_sections`, the shape dimensions), you could simply compute the `split_lengths` in pure Python, then convert to OV, then call `ov_opset.variadic_split`. It would be a lot less code I believe.

BTW, thank you for doing the OV implementation too, most people leave it out.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -2494,6 +2494,24 @@ def split(x, indices_or_sections, axis=0):
...
+    avg_size = tf.math.floordiv(total_size, num_splits)
+    remainder = tf.math.floormod(total_size, num_splits)
+
+    sizes = tf.concat(
+        [
+            tf.fill([remainder], avg_size + 1),
+            tf.fill([num_splits - remainder], avg_size),
+        ],
+        axis=0,
+    )
```
hertschuh: This works, but makes the sizes of the outputs dynamic (i.e. determined when the graph is executed, not when the graph is traced). Ideally we want to trace with static sizes when the input shapes are static, but keep it dynamic when the input shapes are dynamic. This helps with compilation too.

Luckily, in this case, you don't need to have 2 implementations of the code.

Instead of `tf.shape`, you can use `shape_op` (same as `keras.ops.shape`) which returns a tuple with ints for static dimensions and scalar tensors for dynamic dimensions. Then, when you use Python operators, it simply dispatches to the tf ops under the hood.

So, you can just do:
```python
total_size = shape_op(x)[axis]
avg_size = total_size // num_splits
remainder = total_size % num_splits

sizes = [[avg_size + 1] * remainder + [avg_size] * (num_splits - remainder)]
```

`sizes` may be made of tensors are not, either way it's fine.

keras/src/ops/numpy_test.py
```diff
@@ -3601,6 +3616,13 @@ def test_mean(self):
         x = np.array([65504, 65504, 65504], dtype="float16")
         self.assertAllClose(knp.mean(x), np.mean(x))
 
+    def test_array_split(self):
+        x = np.array([[1, 2, 3], [4, 5, 6]])
+        self.assertAllClose(knp.array_split(x, 2), np.array_split(x, 2))
+        self.assertAllClose(
+            knp.array_split(x, 3, axis=1), np.array_split(x, 3, axis=1)
+        )
```
hertschuh: Oh I missed that, but both tests are even splits. Can you add an uneven split or change one of them to be uneven.

buildwithsuhana: Done

PR #21706, opened 2025-10-01
Bug fixes with variable handling in `LossScaleOptimizer`.

hertschuh:
## overwrite_with_gradient

`overwrite_with_gradient` would be ineffective on JAX in real-world conditions, i.e. within `model.fit`.
    
This is because in the training loop, `stateless_apply` is passed `trainable_variables` as arrays containing the values of the trainable variables, not the variables themselves. Instead, we have to inspect the variables.

## apply with gradients only

`apply(grads)` without the `trainable_variables` argument passed in would not apply anything.
    
This is because the code uses `self._trainable_variables`. But this was an empty array for `LossScaleOptimizer`. This was fixed by adding `super().build(...)`.

## unsupported optimizer arguments
    
Also fail when other arguments from the base optimizer are passed to `LossScaleOptimizer.__init__` since they are not actually supported. They are also no longer returned by `get_config`.

### DIFF
--- a/keras/src/optimizers/loss_scale_optimizer.py
+++ b/keras/src/optimizers/loss_scale_optimizer.py
@@ -48,6 +48,7 @@ class LossScaleOptimizer(optimizer.Optimizer):
         inner_optimizer,
         initial_scale=2.0**15,
         dynamic_growth_steps=2000,
+        name=None,
         **kwargs,
     ):
         if not kwargs.pop("dynamic", True):
@@ -56,7 +57,42 @@ class LossScaleOptimizer(optimizer.Optimizer):
                 "Instead, simply set `loss_scale_factor` directly on the "
                 "`inner_optimizer`."
             )
-        super().__init__(learning_rate=0.0, **kwargs)
+
+        # Backwards compatibility code for deserialization.
+        # LossScaleOptimizer used to return all these parameters in `get_config`
+        # from `super.get_config` even though they are all non-functional. We
+        # no longer let user set them, but we have to allow the default values
+        # to be passed during deserialization to support older models.
+        base_optimizer_defaults = {
+            "weight_decay": None,
+            "clipnorm": None,
+            "global_clipnorm": None,
+            "clipvalue": None,
+            "use_ema": False,
+            "ema_momentum": 0.99,
+            "ema_overwrite_frequency": None,
+            "loss_scale_factor": None,
+            "gradient_accumulation_steps": None,
+        }
+        for arg_name, default_value in base_optimizer_defaults.items():
+            if arg_name not in kwargs:
+                continue
+            arg_value = kwargs.pop(arg_name)
+            if (
+                default_value is None and arg_value is not None
+            ) or arg_value != default_value:
+                raise ValueError(
+                    f"LossScaleOptimizer does not support `{arg_name}`. "
+                    f"Instead, set `{arg_name}` on the `inner_optimizer`."
+                )
+
+        if kwargs:
+            raise ValueError(
+                "LossScaleOptimizer does not support arguments: "
+                f"`{'`, `'.join(kwargs.keys())}`."
+            )
+
+        super().__init__(learning_rate=0.0, name=name)
         self.inner_optimizer = inner_optimizer
         self.initial_scale = initial_scale
         self.dynamic_growth_steps = dynamic_growth_steps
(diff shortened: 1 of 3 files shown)


hertschuh:
@cantonios can you review?

keras/src/optimizers/loss_scale_optimizer.py
```diff
@@ -136,7 +136,9 @@ def increment():
                 g
                 if g is None or self._overwrite_variable_with_gradient(v)
                 else ops.divide(g, scale)
-                for g, v in zip(grads, trainable_variables)
+                for g, v in zip(
+                    grads, self.inner_optimizer._trainable_variables
```
cantonios: Summarizing from our brief in-person discussion:
I don't think we should pull directly from the inner optimizer here.  Instead, we should ensure the loss-scale optimizer has the correct `_trainable_variables` (maybe from calling it's `super().build(...)`).

keras/src/optimizers/loss_scale_optimizer.py
```diff
@@ -177,7 +179,7 @@ def apply(self, grads, trainable_variables=None):
     def _stateful_handle_finite_grads(self, grads, trainable_variables):
         scale = self.dynamic_scale
         # Unscale gradients.
-        tvs = trainable_variables or self._trainable_variables
+        tvs = trainable_variables or self.inner_optimizer._trainable_variables
```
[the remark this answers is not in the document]
cantonios: ... which would fix this as well.

keras/src/optimizers/loss_scale_optimizer_test.py
```diff
@@ -40,7 +53,9 @@ def test_finite_step(self, stateless):
         if stateless:
             optimizer.build(vars)
             vars, _ = optimizer.stateless_apply(
-                optimizer.variables, grads, vars
+                [v.value for v in optimizer.variables],
```
cantonios, line 56: I guess this is because of that JAX deprecation issue?

cantonios: How prevalent is this pattern across keras?  If we can no longer rely on the `__jax_array__` built-in to auto convert, maybe there should be a helper function like `optimizer.variable_values` that unwraps the variables?

keras/src/optimizers/loss_scale_optimizer.py
hertschuh: Will call `super().build(...)` and use `self._trainable_variables`.

keras/src/optimizers/loss_scale_optimizer.py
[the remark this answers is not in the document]
hertschuh: Will revert since `super().build(...)` will take care of this.

keras/src/optimizers/loss_scale_optimizer_test.py
hertschuh: I uncovered this as part of the `__jax_array__` deprecation. See https://github.com/keras-team/keras/pull/21702

However, I discovered in the process that we're testing `stateless_apply` the wrong way. `stateless_apply` takes tensors, not variables (unlike the names suggest). So this change of the unit tests is correct, and is independent of the `__jax_array__` deprecation.

This pattern of unwrapping variables was only added in this test file (see https://github.com/keras-team/keras/pull/21702 ). This pattern does exists to some extend, but in the JAX trainer, not the optimizers.

keras/src/optimizers/loss_scale_optimizer.py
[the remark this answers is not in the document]
hertschuh: Fixed

keras/src/optimizers/loss_scale_optimizer.py
```diff
@@ -136,7 +136,9 @@ def increment():
                 g
                 if g is None or self._overwrite_variable_with_gradient(v)
                 else ops.divide(g, scale)
-                for g, v in zip(grads, trainable_variables)
+                for g, v in zip(
+                    grads, self.inner_optimizer._trainable_variables
+                )
```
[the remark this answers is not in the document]
hertschuh: Fixed

keras/src/optimizers/loss_scale_optimizer.py
hertschuh: Fixed

keras/src/optimizers/loss_scale_optimizer_test.py
cantonios: Thanks, this all looks better.

PR #21707, opened 2025-10-02
Add Autoconfig, Coordinated_Optimizer and Sharding keras implementations for Tensor Parallel Autosharding

buildwithsuhana:
This PR introduces support for tensor parallelism autosharding in Keras, enabling users to shard large model layers across multiple devices. This is a crucial feature for training models that are too large to fit into the memory of a single accelerator.

The implementation is centered around two new components:

autoconfig.py: This module contains the logic to analyze a Keras model, identify sharding candidates (e.g., Dense, EinsumDense layers), and generate a sharding plan.

coordinated_optimizer.py: This is an optimizer wrapper that consumes the sharding plan. During training, it intercepts gradients for sharded variables and performs a collective AllReduce to ensure weight updates are correctly synchronized across all devices.


keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,282 @@
...
+        layer: The Keras `layers.Dense` instance to analyze.
+        module: The parent module containing the layer (currently unused).
+        prefix (str): The name prefix for the layer in the model hierarchy
+            (currently unused).
+
+    Returns:
+        str: A string classifying the layer as 'up_projection',
+        'down_projection', or 'generic_dense'.
+    """
+    from keras.src import layers
```
hertschuh: Move to the top

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,282 @@
...
+        state_rules (Dict[str, Any]): A dictionary to be populated with rules
+            for sharding layer weights (state). Keys are regex patterns
+            matching weight names, values are `SplitKeras` actions.
+        output_rules (Dict[str, Any]): A dictionary to be populated with rules
+            for handling layer outputs. Keys are regex patterns matching layer
+            names, values describe the communication op (e.g., 'allreduce').
+        processed_layers (Set[int]): A set of `id()`s of layers that have
+            already been processed to prevent cycles and redundant work.
+    """
+    from keras.src import layers
```
hertschuh: Remove once it's at the top

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,282 @@
...
+        (
+            layers.LayerNormalization,
+            layers.BatchNormalization,
+            layers.GroupNormalization,
+        ),
+    ):
+        return
+
+    if hasattr(current_layer, "layers") and current_layer.layers:
+        for sub_layer in current_layer.layers:
```
hertschuh: `Model.layers` already returns all layers recursively, there is no recursion needed. All the code from line 188 to line 238 can be removed.

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,282 @@
...
+            for handling layer outputs. Keys are regex patterns matching layer
+            names, values describe the communication op (e.g., 'allreduce').
+        processed_layers (Set[int]): A set of `id()`s of layers that have
+            already been processed to prevent cycles and redundant work.
+    """
+    from keras.src import layers
+
+    if id(current_layer) in processed_layers:
+        return
+    processed_layers.add(id(current_layer))
```
hertschuh: Per my comment below about not needing a recursion, this is not needed

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,282 @@
...
+
+    name = current_layer.name
+    full_name = f"{prefix}.{name}" if prefix else name
```
hertschuh: Because you will never really recurse, the prefix won't work.

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,282 @@
...
+
+    if isinstance(current_layer, layers.Dense):
+        mlp_type = analyze_dense_layer_directly(
+            current_layer, module, full_name
+        )
+
+        if mlp_type == "up_projection":
+            state_rules[f"^{full_name}.kernel$"] = Split(
```
hertschuh: Why use regexes everywhere when you always force a full exact match? Why not just key by name?

```python
state_rules[f"{full_name}.kernel"] = ...
```

Also, if you really wanted regexes, you should escape at least the dots.

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,282 @@
...
+
+        else:
+            state_rules[f"^{full_name}.kernel$"] = Split(
+                world_size, 1, "column"
+            )
+            if current_layer.use_bias:
+                state_rules[f"^{full_name}.bias$"] = Split(
+                    world_size, 0, "column"
+                )
+            output_rules[f"^{full_name}$"] = {0: "gather -1"}
```
hertschuh: What does `"gather -1"` mean?

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,282 @@
...
+    This function inspects the input and output dimensions of a Dense layer
+    to determine if it functions as an expansion layer ("up-projection"), a
+    contraction layer ("down-projection"), or neither ("generic_dense"). This
+    classification is a heuristic commonly used to apply tensor parallelism
+    in Transformer-based models, such as in an MLP block where an up-projection
+    is followed by a down-projection.
+
+    Args:
+        layer: The Keras `layers.Dense` instance to analyze.
+        module: The parent module containing the layer (currently unused).
```
hertschuh: If it's unused, remove

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,282 @@
+from keras.src.distribution.tensor_parallel.tensor_layout import LayoutMap
+from keras.src.distribution.tensor_parallel.tensor_layout import Split
+
+
+def analyze_dense_layer_directly(layer, module, prefix):
```
hertschuh: Why have `directly` in the name? What does it mean?

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,282 @@
...
+            (currently unused).
+
+    Returns:
+        str: A string classifying the layer as 'up_projection',
+        'down_projection', or 'generic_dense'.
+    """
+    from keras.src import layers
+
+    if not isinstance(layer, layers.Dense):
+        return "generic_dense"
```
hertschuh: Why call it "generic_dense" and not just "dense"? What does the "generic" part signify?

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,282 @@
...
+            pass
+        else:
+            weight_name = None
+            if hasattr(current_layer, "embeddings"):
+                weight_name = "embeddings"
+            elif hasattr(current_layer, "position_embeddings"):
+                weight_name = "position_embeddings"
+
+            if weight_name:
+                state_rules[f"^{full_name}\\..*{weight_name}$"] = Split(
```
hertschuh: Oh, why do you need `.*` here?

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,282 @@
...
+                    world_size, 1, "column"
+                )
+                output_rules[f"^{full_name}$"] = {0: "no_comm"}
+            return
+
+    elif isinstance(
+        current_layer,
```
hertschuh: What about other layer types?

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,282 @@
...
+        module: The Keras `Model` or `Layer` to generate a config for.
+        device_ids (Sequence[str]): A sequence of device IDs (e.g.,
+            ["gpu:0", "gpu:1"]) across which the model will be sharded.
+
+    Returns:
+        ConfigKeras: A configuration object containing the generated sharding
+        rules for model weights (`state_rules`) and layer outputs
+        (`output_rules`).
+    """
+    world_size = len(device_ids)
```
hertschuh: I guess `world_size` is common terminology in Torch, but I find it confusing because it doesn't tell you what you're counting. `device_count` seems more intuitive.

keras/src/distribution/tensor_parallel/autoconfig_test.py
```diff
@@ -0,0 +1,138 @@
+from autoconfig import analyze_dense_layer_directly
+from autoconfig import get_default_config_keras
+
+import keras
```
hertschuh, line 6: In general, we don't `import keras` within the keras codebase because it easily creates circular import, we import specific parts of `keras.src` like you did below.

keras/src/distribution/tensor_parallel/coordinated_optimizer.py
```diff
@@ -0,0 +1,653 @@
...
+        """
+        if not self.shard_optimizer_states or not self.base_optimizer.built:
+            return
+
+        self.sharded_states = {}
+        self._state_variable_to_parameter = {}
+        self._variable_to_slot_name = {}
+        opt_name = self.base_optimizer.name
+
+        normalized_params = sorted(
```
hertschuh, line 65: I have a hard time following what the code in this method is doing. I think it's try to re-pair the optimizer variables with the corresponding model variables.

I think it would be easier to capture that information in `BaseOptimizer.add_variable_from_reference`. Today we have `_get_variable_index`, but we need something more specific.

Also what do you call a slot?

keras/src/distribution/tensor_parallel/coordinated_optimizer.py
```diff
@@ -0,0 +1,653 @@
...
+        if not gradients:
+            return []
+
+        if distributed_backend.is_multi_device_capable():
+            all_reduce_fn = distributed_backend.get_communication_ops()[
+                "all_reduce"
+            ]
+            numpy_grad = ops.convert_to_numpy(gradients[0])
+            synced_numpy = all_reduce_fn(numpy_grad, op="mean")
+            synced_tensor = ops.convert_to_tensor(synced_numpy)
```
hertschuh:
```python
numpy_grad = ops.convert_to_numpy(gradients[0])
```
This is going to copy the gradient tensor from GPU/TPU to CPU memory. At that point, there is no sharding anymore because it's going to gather all shards from all GPUs into one single recombined tensor on CPU.

```python
synced_numpy = all_reduce_fn(numpy_grad, op="mean")
```
There is no `all_reduce` needed here, this is a CPU NumPy array, you might as well do `np.mean`.

However, what JAX will do is copy the full tensor (unsharded) from CPU to device 0 and perform the mean on that one device.

```python
synced_tensor = ops.convert_to_tensor(synced_numpy)
```
`synced_numpy` is already a JAX array per my comment above, so this is a no-op.

So overall, I'm not sure what the intent is, but it look like this is not doing what you think it's doing. In particular, it's moving gradient back and forth to CPU, which will reduce the throughput.

keras/src/distribution/tensor_parallel/coordinated_optimizer.py
```diff
@@ -0,0 +1,653 @@
...
+            return [synced_tensor for _ in range(self.world_size)]
+
+        stacked_grads = keras.ops.stack(
+            [ops.convert_to_tensor(g) for g in gradients], axis=0
+        )
+        mean_grad = ops.mean(stacked_grads, axis=0)
```
hertschuh: Why would there be more than 1 gradient in this case?

keras/src/distribution/tensor_parallel/coordinated_optimizer.py
```diff
@@ -0,0 +1,653 @@
...
+        return [mean_grad for _ in range(len(gradients))]
+
+    def get_weights(self):
```
hertschuh, line 385: Oh weird, I wonder why `get_weights` is missing in `BaseOptimizer`.

keras/src/distribution/tensor_parallel/coordinated_optimizer.py
```diff
@@ -0,0 +1,653 @@
...
+
+        Args:
+            variables: A list of model variables to be optimized.
+        """
+        self.shard_optimizer_states = True
+        self._variables = variables
+        self._initialize_sharded_states()
+
+
+class TensorParallelOptimizer(optimizers.Optimizer):
```
hertschuh, line 406: It appears that `TensorParallelOptimizer` is mostly a wrapper around `CoordinatedOptimizer`.

Any reason to have both separate?

keras/src/distribution/tensor_parallel/autoconfig.py
buildwithsuhana: I've tested removing the recursion logic as you suggested, and I found an issue, If I remove the manual recursion (lines 188-238) and the processed_layers check (lines 110-112), the number of sharded parameters drops completely to zero.

I further found out that Keras's built-in layer introspection does not guarantee a full, recursive enumeration of all nested sub-layers in all architectures, especially in complex models.

For the large-scale Transformer architectures we are targeting, many Dense or EinsumDense layers are often hidden from standard traversal. They are often stored as:

- Layers inside non-recursive containers.
- Layers stored as custom attributes (e.g., self.q_proj) within a custom layer.
- Layers stored within lists or tuples (e.g., a list of attention heads or blocks).

Since the primary goal of autoconfig.py is to find and apply sharding rules to every relevant layer, the current implementation of _recursive_layer_traversal must be maintained to ensure all nested sharding targets are discovered.

keras/src/distribution/tensor_parallel/autoconfig.py
buildwithsuhana: Concatenation of the partial output activations along the last dimension to restore the full tensor.

keras/src/distribution/tensor_parallel/autoconfig.py
buildwithsuhana: The function is set up to only worry about the biggest layers in the model (Dense, Embedding, etc.). These are the only ones big enough to cause memory problems and need splitting (sharding).

We skip the smaller layers for a few reasons:

Normalization Layers (like LayerNormalization): Their weights are small. We leave them alone so we don't slow things down with extra communication.

Layers with No Weights (like Dropout, Activation): They don't have anything to split. They just use the sharded data that comes from the layer before them.

keras/src/distribution/tensor_parallel/coordinated_optimizer.py
buildwithsuhana: A "slot" is simply the name for an extra state variable an optimizer creates for each model weight like the momentum (m) or velocity (v) variable in Adam.

keras/src/distribution/tensor_parallel/coordinated_optimizer.py
buildwithsuhana: The BaseOptimizer class doesn’t have get_weights() because optimizers don’t own model weights, they just update them.

keras/src/distribution/tensor_parallel/coordinated_optimizer.py
buildwithsuhana: The fundamental reason for designing both a TensorParallelOptimizer and a CoordinatedOptimizer is to maintain a clear architectural boundary between the simple public interface and the complex internal mechanics. The TensorParallelOptimizer is the class that faces the rest of the Keras framework. 

It acts as the stable public API layer, ensuring it inherits from the necessary Keras base class and perfectly adheres to expected methods like build and apply_gradients. Meanwhile, the CoordinatedOptimizer functions as the hidden private logic engine, encapsulating all the intricate, implementation-specific details required for truly distributed training. This private engine manages complex tasks such as partitioning the sharded NumPy state dictionaries (state partitioning) and executing the crucial synchronization logic needed for parallel gradient updates (synchronize gradients).

keras/src/distribution/tensor_parallel/autoconfig.py
buildwithsuhana: Done, removed recursion.

PR #21712, opened 2025-10-03
Fix ModelParallel OOM issue during weight loading

amitsrivastava78:
- Modified load_own_variables() to use _direct_assign() for sharded variables
- Prevents loading full weight tensors on single device before distribution
- Resolves RESOURCE_EXHAUSTED errors when loading large models with ModelParallel
- Maintains backward compatibility for non-sharded variables
- Enables loading of models like Gemma2 2B/7B without OOM errors


keras/src/utils/variable_loading.py
```diff
@@ -0,0 +1,48 @@
...
+        # Use normal assign for non-sharded variables
+        logging.info(
+            f"load_own_variables: Loading non-sharded variable "
+            f"({variable.name}) with assign"
+        )
+        variable.assign(weight_data)
+        logging.info(
+            f"load_own_variables: Variable ({variable.name}) "
+            "loaded successfully"
+        )
```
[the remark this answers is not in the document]
amitsrivastava78: Fixed

keras/src/layers/preprocessing/index_lookup.py
```diff
@@ -806,7 +806,19 @@ def save_own_variables(self, store):
...
+            # and use chunked loading
+            if (
+                hasattr(self.idf_weights, "_layout")
+                and self.idf_weights._layout is not None
+            ):
+                # Use _direct_assign for sharded variables to avoid OOM
+                self.idf_weights._direct_assign(weight_data)
+            else:
+                # Use normal assign for non-sharded variables
+                self.idf_weights.assign(weight_data)
```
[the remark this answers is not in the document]
amitsrivastava78: Fixed

keras/src/backend/jax/core.py
```diff
@@ -187,24 +388,111 @@ def __setstate__(self, state):
...
+            # Block and keep reference to ALL shards
+            value = jax.block_until_ready(value)
+            self._maybe_create_strong_reference(value)
+            # Set value for NNX
+            object.__setattr__(self, "raw_value", value)
+
+            logging.info(
+                f"_initialize (NNX): Variable '{self.path}' initialized"
+            )
+
```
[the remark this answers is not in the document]
amitsrivastava78: Fixed

keras/src/backend/jax/core.py
```diff
@@ -23,33 +29,228 @@
...
 
+    def _maybe_create_strong_reference(self, value):
+        """Create a strong ref to a JAX array to prevent GC."""
+        if isinstance(value, jax.Array):
+            if hasattr(value, "addressable_shards"):
+                # For sharded arrays, hold references to the shards' data.
+                shard_data = [shard.data for shard in value.addressable_shards]
+                if not hasattr(self, "_shard_references"):
+                    self._shard_references = []
+                self._shard_references.append(shard_data)
```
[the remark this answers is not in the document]
amitsrivastava78: Fixed

keras/src/backend/jax/core.py
```diff
@@ -3,7 +3,7 @@
 import jax.numpy as jnp
 import ml_dtypes
 import numpy as np
-from jax import export as jax_export
+from absl import logging
```
hertschuh: This PR is undoing a lot of changes that were made in this file. It wasn't rebased correctly.

keras/src/backend/jax/core.py
```diff
@@ -23,33 +23,272 @@
 IS_THREAD_SAFE = True
 
 
+def _is_jax_tracer(x):
```
hertschuh: Can't you use (and potentially change) the one from `jax_utils`?

keras/src/utils/variable_loading.py
```diff
@@ -0,0 +1,40 @@
...
+        variable: The variable to load data into
+        weight_data: The weight data to load (typically a numpy array)
+
+    Returns:
+        None
+    """
+    # Check if variable has a layout (is sharded)
+    if hasattr(variable, "_layout") and variable._layout is not None:
+        # Use _direct_assign for sharded variables to avoid OOM
+        logging.info(
```
hertschuh: This logging will be very noisy, let's remove it.

keras/src/utils/variable_loading.py
```diff
@@ -0,0 +1,40 @@
...
+Utility functions for loading variables with sharded support.
+
+This module provides common utilities for loading variables that may be sharded
+across multiple devices, which is useful for distributed training scenarios.
+"""
+
+from absl import logging
+
+
+def load_variable_with_sharded_support(variable, weight_data):
```
hertschuh: I don't think we need this, you can just do `variable._direct_assign(weight_data)` everywhere you used `load_variable_with_sharded_support`.

That check for `if hasattr(variable, "_layout") and variable._layout is not None:` is already done within `_direct_assign`.

keras/src/backend/jax/core.py
```diff
@@ -23,33 +23,272 @@
...
+    logging.debug(f"{log_prefix}: Has layout: {variable._layout is not None}")
+
+    # If we have a layout, distribute the tensor to avoid OOM
+    if variable._layout is not None:
+        logging.info(
+            f"{log_prefix}: Sharded initialization (layout: {variable._layout})"
+        )
+
+        # Ensure value is on host (numpy array)
+        if not isinstance(value, np.ndarray):
```
hertschuh: But that's too late. The initializer has already been called on `device:0`. For this to work the way you intend it to, you need to run the initializer on CPU using a `with device(...)` scope.

keras/src/backend/jax/core_test.py
```diff
@@ -1,21 +1,159 @@
+"""Test for core.py."""
+
 import os
 
+os.environ["KERAS_BACKEND"] = "jax"
```
hertschuh: Remove, this should not be needed and will fail anyway if, for instance, we run this with the Torch CUDA setup

keras/src/backend/jax/core_test.py
```diff
@@ -1,21 +1,159 @@
+"""Test for core.py."""
```
hertschuh: We don't add such docstring, please remove.

keras/src/backend/jax/core_test.py
```diff
@@ -1,21 +1,159 @@
...
+        for _ in range(5):
+            value = var.value
+            self.assertIsNotNone(value)
+
+    def test_variable_loading_with_sharding(self):
+        """Test variable loading with sharding support."""
+        self._require_min_devices(2)  # Requires multiple devices for sharding
+
+        # Create a temporary file for the variable
+        with tempfile.NamedTemporaryFile(suffix=".npy", delete=False) as f:
```
hertschuh: We have a utility in the TestCase class already. You can just do:
```python
path = os.path.join(self.get_temp_dir(), "my_keras_model")
np.save(path, jax.numpy.ones((10, 10)))
```
It will cleanup at the end of the test.

keras/src/backend/jax/core_test.py
```diff
@@ -1,21 +1,159 @@
...
+            temp_file = f.name
+            np.save(temp_file, jax.numpy.ones((10, 10)))
+
+        try:
+            # Create variable with sharding
+            var = JaxVariable(jax.numpy.zeros((10, 10)))
+            # Load data into it
+            load_variable_with_sharded_support(var, np.load(temp_file))
```
hertschuh: Why do you need a file for this? How is it different from a np array?

keras/src/distribution/distribution_lib_test.py
```diff
@@ -361,6 +364,371 @@ def test_distribute_dataset(self):
...
+            original_weights = []
+            for var in model.weights:
+                if hasattr(var, "_layout") and var._layout is not None:
+                    # For sharded variables, get the full distributed value
+                    original_weights.append(var.value.copy())
+                else:
+                    original_weights.append(var.numpy().copy())
+
+            # Save model weights to temporary file
+            with tempfile.TemporaryDirectory() as temp_dir:
```
hertschuh: We have a utility in the TestCase class already. You can just do:
```python
weights_path = os.path.join(self.get_temp_dir(), "my_keras_model")
```
It will cleanup at the end of the test.

And then you don't need to intent everything.

keras/src/distribution/distribution_lib_test.py
```diff
@@ -361,6 +364,371 @@ def test_distribute_dataset(self):
...
+                    has_shard_refs_orig = (
+                        hasattr(orig_var, "_shard_references")
+                        and orig_var._shard_references
+                    )
+                    has_shard_refs_loaded = (
+                        hasattr(loaded_var, "_shard_references")
+                        and loaded_var._shard_references
+                    )
+
+                    logging.debug(
```
hertschuh: Remove all logging.

keras/src/layers/core/dense.py
```diff
@@ -306,18 +307,34 @@ def load_own_variables(self, store):
...
+                target_variables.append(self.outputs_grad_amax_history)
+            elif self.quantization_mode == "gptq":
+                target_variables.append(self.kernel_scale)
+                target_variables.append(self.kernel_zero)
+                target_variables.append(self.g_idx)
+            else:
+                raise self._quantization_mode_error(self.quantization_mode)
+        for i, variable in enumerate(target_variables):
+            weight_data = store[str(i)]
+            load_variable_with_sharded_support(variable, weight_data)
```
hertschuh: This code is being changed I believe.

But why couldn't you do a 1-line change: 
```diff
< variable.assign(store[str(i)])
---
> load_variable_with_sharded_support(variable, store[str(i)])
```

keras/src/layers/core/einsum_dense.py
```diff
@@ -374,18 +375,33 @@ def load_own_variables(self, store):
...
+                target_variables.append(self.outputs_grad_amax_history)
+            elif self.quantization_mode == "gptq":
+                target_variables.append(self.kernel_scale)
+                target_variables.append(self.kernel_zero)
+                target_variables.append(self.g_idx)
+            else:
+                raise self._quantization_mode_error(self.quantization_mode)
+        for i, variable in enumerate(target_variables):
+            weight_data = store[str(i)]
+            load_variable_with_sharded_support(variable, weight_data)
```
hertschuh: Same comment about this code being changed and a 1-line change.

keras/src/layers/core/embedding.py
```diff
@@ -266,14 +267,16 @@ def load_own_variables(self, store):
...
-            variable.assign(store[str(i)])
+        target_variables = [self._embeddings]
+        if self.quantization_mode is not None:
+            if self.quantization_mode in ("int8", "int4"):
+                target_variables.append(self.embeddings_scale)
+            else:
+                raise self._quantization_mode_error(self.quantization_mode)
+        for i, variable in enumerate(target_variables):
+            weight_data = store[str(i)]
+            load_variable_with_sharded_support(variable, weight_data)
```
hertschuh: Same comment about this code being changed and a 1-line change.

keras/src/distribution/distribution_lib_test.py
```diff
@@ -361,6 +364,371 @@ def test_distribute_dataset(self):
...
+                        f"  Original has shard references: "
+                        f"{has_shard_refs_orig}"
+                    )
+                    logging.debug(
+                        f"  Loaded has shard references: "
+                        f"{has_shard_refs_loaded}"
+                    )
+
+                    self.assertTrue(
```
hertschuh: Using `assertTrue` forces you to put a specific error message to provide context.

I would replace all of this:
```python
                    has_shard_refs_orig = (
                        hasattr(orig_var, "_shard_references")
                        and orig_var._shard_references
                    )
                    logging.debug(
                        f"  Original has shard references: "
                        f"{has_shard_refs_orig}"
                    )
                    self.assertTrue(
                        has_shard_refs_orig,
                        f"Original {var_name} should have shard references",
                    )
                    self.assertGreater(
                        len(orig_var._shard_references),
                        0,
                        f"Original {var_name} has empty shard references",
                    )
```

With line:
```python
self.assertLen(orig_var._shard_references, 1)
```

Not only it's a lot less code, it will actually gives you more information in case of error:
- if `orig_var` doesn't have `_shard_references` as an attribute, it will raise an error telling you exactly that
- if `orig_var._shard_references` is None, `len` will fail telling you it's None, so you'll know (`has_shard_refs_orig` won't tell you directly if the attribute is missing or None, you'll have to look at the debug loggin)
- `assertLen` will tell you what you're taking the len of whereas `assertGreater` will tell you `0 < 0` which is not super helpful.

hertschuh (commented):
Thanks for the PR!

It's unfortunate that this is combining 3 different things (which you describe in your doc):
1. initialization of sharded variables
2. reloading of sharded variables
3. reference counting of variables

In particular, I don't think we should do 3 because both Python and JAX already track usage of arrays. Therefore:
- I believe it's hiding some other bug
- Adding our own tracking system on top is error prone and is probably going to add memory leaks because it's very easy to forget to clear references.

keras/src/backend/jax/core.py
amitsrivastava78: ok

keras/src/utils/variable_loading.py
amitsrivastava78: ok

keras/src/backend/jax/core.py
amitsrivastava78: ok

keras/src/backend/jax/core_test.py
amitsrivastava78: ok

keras/src/backend/jax/core_test.py
amitsrivastava78: ok

keras/src/utils/variable_loading.py
amitsrivastava78: ok

keras/src/backend/jax/core_test.py
amitsrivastava78: ok

keras/src/backend/jax/core_test.py
amitsrivastava78: ok

keras/src/distribution/distribution_lib_test.py
amitsrivastava78: ok

keras/src/distribution/distribution_lib_test.py
amitsrivastava78: ok

keras/src/distribution/distribution_lib_test.py
amitsrivastava78: ok

keras/src/layers/core/einsum_dense.py
amitsrivastava78: ok

keras/src/layers/core/dense.py
amitsrivastava78: ok

keras/src/backend/jax/core.py
amitsrivastava78: ok rebased again and ensured no fixes are missing also restored jax_export and other fixes overwritten

PR #21713, opened 2025-10-03
Refactor variable serialization.

james77777778:
This is an alternative solution compared to #21681

Since the default behavior is to store the variables using their order as the keys, we can maintain this behavior by using a `variable_serialization_spec` to explicitly define the order.
`variable_serialization_spec` works like a human-readable proxy between the variables and `store` for better compatibility.

IMO, `save_own_variables` and `load_own_variables` become more readable and simpler in this PR.

cc @hertschuh @fchollet @JyotinderSingh

### DIFF
--- a/keras/src/layers/core/dense.py
+++ b/keras/src/layers/core/dense.py
@@ -258,25 +258,25 @@ class Dense(Layer):
         if not self.built:
             return
         mode = self.quantization_mode
-        if mode not in self.quantization_variable_spec:
+        if mode not in self.variable_serialization_spec:
             raise self._quantization_mode_error(mode)
 
         # Kernel plus optional merged LoRA-aware scale (returns (kernel, None)
         # for None/gptq)
         kernel_value, merged_kernel_scale = self._get_kernel_with_merged_lora()
-
-        # Save the variables using the name as the key.
-        if mode != "gptq":
-            store["kernel"] = kernel_value
-        if self.bias is not None:
-            store["bias"] = self.bias
-        for name in self.quantization_variable_spec[mode]:
-            if name == "kernel_scale" and mode in ("int4", "int8"):
+        idx = 0
+        for name in self.variable_serialization_spec[mode]:
+            if name == "kernel":
+                store[str(idx)] = kernel_value
+            elif name == "bias" and self.bias is None:
+                continue
+            elif name == "kernel_scale" and mode in ("int4", "int8"):
                 # For int4/int8, the merged LoRA scale (if any) comes from
                 # `_get_kernel_with_merged_lora()`
-                store[name] = merged_kernel_scale
+                store[str(idx)] = merged_kernel_scale
             else:
-                store[name] = getattr(self, name)
+                store[str(idx)] = getattr(self, name)
+            idx += 1
 
     def load_own_variables(self, store):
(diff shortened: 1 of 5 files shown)


fchollet:
Thanks for the PR!

To be clear, the reason we were using indices instead of string names was that string names are susceptible to changes, while indices are more stable. It makes the format more future proof.

One thing I've considered doing (and which could be done across the board, for all weights in all layers) is to have names like `"0:kernel"`, `"1:bias"`, which would be parsed as `int(name.split(":")[0])`. This would be human readable and fully backwards compatible with the old format. What do you think?

hertschuh:
> Thanks for the PR!
> 
> To be clear, the reason we were using indices instead of string names was that string names are susceptible to changes, while indices are more stable. It makes the format more future proof.

Thanks for providing the context.

> One thing I've considered doing (and which could be done across the board, for all weights in all layers) is to have names like `"0:kernel"`, `"1:bias"`, which would be parsed as `int(name.split(":")[0])`. This would be human readable and fully backwards compatible with the old format. What do you think?

Unfortunately, this is not backwards compatible in the way we were discussing.

Let's say we save the latest Gemma checkpoints with Keras 3.12 with this new format, people using Keras 3.11 won't be able to load them.

fchollet:
> Let's say we save the latest Gemma checkpoints with Keras 3.12 with this new format, people using Keras 3.11 won't be able to load them.

Yes, but historically it has been the case that file loading has been backwards compatible (newer Keras versions can load older files) but not forward compatible (older Keras versions may not necessarily load newer files).

The only alternative here is to keep integer indices.

hertschuh:
> Yes, but historically it has been the case that file loading has been backwards compatible (newer Keras versions can load older files) but not forward compatible (older Keras versions may not necessarily load newer files).
> 
> The only alternative here is to keep integer indices.

Ok, let's submit this as-is for now.

But I would not be opposed to adopting the "0:kernel", "1:bias" format at some point.

james77777778:
(I just got back from my vacation)

Thanks for the inputs @fchollet @hertschuh
I agree that `index:name` is a good solution for human readability.

I think it's a good idea to slowly migrate to this format. Here is my plan:
1. Modify `Layer.save_own_variables` and `Layer.load_own_variables` to support this format. Add a new kwarg `use_readable_format` to toggle the feature. Add a warning about this.
2. Default `use_readable_format` to `True` at a specified version.

WDYT? If this sounds good, I can start working on a PR for step 1.

hertschuh:
> (I just got back from my vacation)
> 
> Thanks for the inputs @fchollet @hertschuh I agree that `index:name` is a good solution for human readability.
> 
> I think it's a good idea to slowly migrate to this format. Here is my plan:
> 
> 1. Modify `Layer.save_own_variables` and `Layer.load_own_variables` to support this format. Add a new kwarg `use_readable_format` to toggle the feature. Add a warning about this.
> 2. Default `use_readable_format` to `True` at a specified version.
> 
> WDYT? If this sounds good, I can start working on a PR for step 1.

Sounds good. My only comment is that plumbing the `use_readable_format` option through all the layers might be a lot of work for a parameter that most people will never use. How about making it a global flag / setting?

Issue #21698, opened 2025-09-27
Does Keras have an op similar to jax.jvp?


pass-lin:
I’d like to ask whether Keras offers a backend-agnostic equivalent of jax.jvp; if not, I can contribute an implementation.

divyashreepathihalli:
jvp op is not yet supported. Please feel free to open a PR with the contribution.

PR #21720, opened 2025-10-04
Add jvp op

pass-lin:
from https://github.com/keras-team/keras/issues/21698
We implemented the jvp operator. Since numpy and OpenVINO as inference frameworks do not have automatic differentiation capabilities, no design is made.

### DIFF
--- a/keras/src/backend/jax/linalg.py
+++ b/keras/src/backend/jax/linalg.py
@@ -97,3 +97,7 @@ def lstsq(a, b, rcond=None):
     a = convert_to_tensor(a)
     b = convert_to_tensor(b)
     return jnp.linalg.lstsq(a, b, rcond=rcond)[0]
+
+
+def jvp(fun, primals, tangents, has_aux=False):
+    return jax.jvp(fun, primals, tangents, has_aux=has_aux)
--- a/keras/src/ops/linalg.py
+++ b/keras/src/ops/linalg.py
@@ -774,4 +774,4 @@ def jvp(fun, primals, tangents, has_aux=False):
     >>> print(tangents)
     0.19900084
     """
-    return backend.linalg.jvp(fun, primals, tangents)
+    return backend.linalg.jvp(fun, primals, tangents, has_aux=has_aux)
--- a/keras/src/ops/linalg.py
+++ b/keras/src/ops/linalg.py
@@ -732,3 +732,46 @@ def _assert_a_b_compat(a, b):
                 "Expected `a.shape[-1] == b.shape[-1]`. "
                 f"Received: a.shape={a.shape}, b.shape={b.shape}"
             )
+
+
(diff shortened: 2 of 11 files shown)


keras/src/ops/linalg.py
```diff
@@ -732,3 +733,97 @@ def _assert_a_b_compat(a, b):
...
+            and its length should be equal to the number of positional
+            parameters of `fun`.
+        tangents: The tangent vector for which the Jacobian-vector product
+            should be evaluated. Should be either a tuple or a list of
+            tangents, with the same tree structure and array shapes as
+            `primals`.
+        has_aux: Optional, bool. Indicates whether `fun` returns a pair where
+            the first element is considered the output of the mathematical
+            function to be differentiated and the second element is auxiliary
+            data. Default is False.
```
fchollet: Use backticks around code keywords

keras/src/ops/linalg.py
```diff
@@ -732,3 +733,97 @@ def _assert_a_b_compat(a, b):
...
+        Jacobian-vector product of `function` evaluated at `primals` with
+        `tangents`. The `tangents_out` value has the same Python tree
+        structure and shapes as `primals_out`.
+
+        If `has_aux` is True, returns a (primals_out, tangents_out, aux)
+        tuple where `aux` is the auxiliary data returned by `fun`.
+
+    Example:
+
+        >>> from keras import ops
```
fchollet: Unindent the code example (should have the same indentation as the docstring)

keras/src/ops/linalg.py
```diff
@@ -732,3 +733,97 @@ def _assert_a_b_compat(a, b):
...
+        >>> a1, a2 = ops.convert_to_tensor(0.1), ops.convert_to_tensor(0.2)
+        >>> primals, tangents = ops.jvp(ops.sin, (a1,), (a2,))
+        >>> print(primals)
```
fchollet: This is terminal syntax so you don't need `print()`

keras/src/backend/jax/linalg.py
```diff
@@ -97,3 +97,7 @@ def lstsq(a, b, rcond=None):
     a = convert_to_tensor(a)
     b = convert_to_tensor(b)
     return jnp.linalg.lstsq(a, b, rcond=rcond)[0]
+
+
+def jvp(fun, primals, tangents, has_aux=False):
+    return jax.jvp(fun, primals, tangents, has_aux=has_aux)
```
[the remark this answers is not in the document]
fchollet: I believe the gemini comment is incorrect, please confirm. Return format is `(primals_out, tangents_out, aux)` everywhere.

fchollet (commented):
Thanks for the PR!

pass-lin:
@fchollet I was designed with jax.jvp as the standard, and in fact, torch.jvp also takes jax.jvp as the standard.

fchollet (approved):

PR #21721, opened 2025-10-05
Fix negative index handling in MultiHeadAttention attention_axes

### LINKED ISSUE #21714: Incorrect results when using `MultiHeadAttention` `attention_axes` with negative indexing
Using negative indexing (eg. -2 for second last element) for `attention_axes` in `MultiHeadAttention` returns wrong results.

I think the issue is with `_dot_product_equation`, computed in `_build_attention_equation`

```python
import numpy as np
import keras

# (batch0, batch1, seq, features)
x = np.random.normal(size=(10, 5, 128, 16))

# 0) flatten the batch dimensions
x_flat = keras.ops.reshape(x, (x.shape[0] * x.shape[1],) + x.shape[2:])

mha = keras.layers.MultiHeadAttention(num_heads=4, key_dim=16)
z_flat, a_flat = mha(x_flat, x_flat, return_attention_scores=True)
z = keras.ops.reshape(z_flat, x.shape[:2] + z_flat.shape[1:])
a = keras.ops.reshape(a_flat, x.shape[:2] + a_flat.shape[1:])

print(z.shape, a.shape) # (10, 5, 128, 16) (10, 5, 4, 128, 128)

# 1) pass attention_axes with regular indexing
mha = keras.layers.MultiHeadAttention(num_heads=4, key_dim=16, attention_axes=2)
z, a = mha(x, x, return_attention_scores=True)

print(z.shape, a.shape) # (10, 5, 128, 16) (10, 5, 4, 128, 128)
print(mha._dot_product_equation) # abfde,abcde->abdcf

# 2) pass attention_axes with negative indexing <- NOT WORKING AS EXPECTED
mha = keras.layers.MultiHeadAttention(num_heads=4, key_dim=16, attention_axes=-2)
z, a = mha(x, x, return_attention_scores=True)

print(z.shape, a.shape) # (10, 5, 128, 16) (10, 5, 128, 4, 4) 
print(mha._dot_product_equation) # abcfe,abcde->abcdf
```

See this gist: 
https://gist.github.com/AmedeoBiolatti/45fdd330f06d94fbe0dad51f4a3d0f2e

utsab345:
## Description
Fixes #21714 - Incorrect results when using `MultiHeadAttention` `attention_axes` with negative indexing

## Problem
When `attention_axes` was specified with negative indices (e.g., `-2`), the normalization was happening relative to the projected tensor rank (which includes the `num_heads` dimension) rather than the input tensor rank. This caused incorrect axis selection and wrong einsum equations.

For example, with input shape `(10, 5, 128, 16)` (rank 4):
- `attention_axes=2` correctly produced equation `abfde,abcde->abdcf`
- `attention_axes=-2` incorrectly produced equation `abcfe,abcde->abcdf`

## Solution
Modified `_build_attention` method in `MultiHeadAttention` to normalize negative indices relative to the input rank (`rank - 1`) before the num_heads dimension is added during projection. This ensures:
- `attention_axes=-2` normalizes to `input_rank + (-2) = 4 + (-2) = 2`
- Both `attention_axes=2` and `attention_axes=-2` now produce identical results

## Changes
- Updated `keras/src/layers/attention/multi_head_attention.py`:
  - Added negative index normalization in `_build_attention` method
- Added `integration_tests/test_multi_head_attention_negative_axis.py`:
  - Test verifies that negative and positive indexing produce identical outputs

## Testing
```python
x = np.random.normal(size=(2, 3, 8, 4))
mha_pos = keras.layers.MultiHeadAttention(num_heads=2, key_dim=4, attention_axes=2)
mha_neg = keras.layers.MultiHeadAttention(num_heads=2, key_dim=4, attention_axes=-2)
# Both now produce identical shapes and values (with same weights)

### DIFF
--- a/keras/src/layers/attention/multi_head_attention_test.py
+++ b/keras/src/layers/attention/multi_head_attention_test.py
@@ -204,8 +204,6 @@ class MultiHeadAttentionTest(testing.TestCase):
         )
 
     def test_attention_axes_negative_indexing(self):
-        """Test that negative attention_axes indexing matches
-        positive indexing."""
         x = np.random.normal(size=(2, 3, 8, 4))
 
         # Create two layers with equivalent positive and negative indices

### CODE REVIEW

File: keras/src/layers/attention/multi_head_attention.py [hunk 1]
```diff
@@ -760,6 +770,12 @@ def _build_attention_equation(rank, attn_axes):
     Returns:
         Einsum equations.
     """
+    # Normalize negative indices to positive indices
+    if isinstance(attn_axes, (list, tuple)):
+        attn_axes = tuple(ax % rank if ax < 0 else ax for ax in attn_axes)
+    else:
+        attn_axes = (attn_axes % rank if attn_axes < 0 else attn_axes,)
```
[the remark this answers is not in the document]
hertschuh: Agree, please revert.

keras/src/utils/image_utils.py
```diff
@@ -175,10 +175,13 @@ def save_img(path, x, data_format=None, file_format=None, scale=True, **kwargs):
         **kwargs: Additional keyword arguments passed to `PIL.Image.save()`.
     """
     data_format = backend.standardize_data_format(data_format)
+    # Normalize jpg → jpeg
+    if file_format is not None and file_format.lower() == "jpg":
```
hertschuh: Unrelated changes, please revert.

integration_tests/test_save_img.py
```diff
@@ -0,0 +1,27 @@
+import os
```
hertschuh: Unrelated changes, please revert.

integration_tests/test_multi_head_attention_negative_axis.py
```diff
@@ -0,0 +1,28 @@
+import numpy as np
+
+import keras
+
+
+def test_attention_axes_negative_indexing_matches_positive():
+    x = np.random.normal(size=(2, 3, 8, 4))
```
hertschuh: Move to `multi_head_attention_test.py` and use the unit test style, i.e. `self.assertEqual`, `self.assertAllClose`, ...

keras/src/layers/attention/multi_head_attention.py
```diff
@@ -378,7 +378,17 @@ def _build_attention(self, rank):
         if self._attention_axes is None:
             self._attention_axes = tuple(range(1, rank - 2))
         else:
-            self._attention_axes = tuple(self._attention_axes)
+            # Normalize negative indices relative to INPUT rank (rank - 1)
+            input_rank = rank - 1
```
hertschuh: Why `rank - 1`?

I think this would be enough instead of lines 381-391:
```python
self._attention_axes = tuple(axis + rank if axis < 0 else axis for axis in self._attention_axes)
```

keras/src/layers/attention/multi_head_attention_test.py
```diff
@@ -203,6 +203,38 @@ def test_high_dim_attention(
             run_training_check=False,
         )
 
+    def test_attention_axes_negative_indexing(self):
+        """Test that negative attention_axes indexing matches
+        positive indexing."""
```
hertschuh: This is the only change that's left (and actually this should be undone).

It appears that you've reverted too much.

hertschuh:
@utsab345 

There is no change anymore. Basically everything was reverted.

hertschuh (approved):
Thank you for the fix!

Issue #21711, opened 2025-10-03
Note the available backends in docstring of `keras.preprocessing.image.smart_resize()`


ILCSFNO:
## Doc Issue

The doc of `keras.preprocessing.image.smart_resize()` shows its description as below:

https://github.com/keras-team/keras/blob/3fac66fe5239647d345e465d1fa456a33c588e99/keras/src/utils/image_utils.py#L357-L358

For the repro below, I know that the default backend is `keras.src.backend`, but what backend can be served as other `Backend module to use`?

I tried `keras.backend` but failed. If there are some other backends available to use, I suggest to note some examples of them in the docstring above.

Thanks a lot!

### Repro
```python
import numpy as np
import keras
size = (256, 256)
backend_module = keras.backend
image = np.random.randint(0, 256, size=(size[0], size[1], 3), dtype=np.uint8)
resized_image = keras.preprocessing.image.smart_resize(image, size, interpolation='bicubic', backend_module=backend_module)
print('Original Image Shape:', image.shape)
print('Resized Image Shape:', resized_image.shape)
```
### Output
```text
AttributeError: module 'keras.api.backend' has no attribute 'convert_to_tensor'
```

Flakes342:
Good catch @ILCSFNO. The parameter is misleading as currently documented and should be clarified to prevent any confusion. The `backend_module` parameter only accepts `keras.src.backend` (which is already the default). There are no other valid options, I also tried tf, jax, etc.

Why `keras.backend` doesn't work: The function internally calls backend_module.image.resize(), which requires the internal backend structure. https://github.com/keras-team/keras/blob/3fac66fe5239647d345e465d1fa456a33c588e99/keras/src/utils/image_utils.py#L451

The public `keras.backend` API doesn't have this image submodule. So, I guess that the docstring should be updated to state this clearly.

PR #21722, opened 2025-10-05
Fix misleading documentation for smart_resize backend_module parameter

Flakes342:
## Summary
Clarifies the `backend_module` parameter documentation in smart_resize() to prevent user confusion.

## Changes
Updated the docstring for the backend_module parameter to clarify:

- It's primarily for internal use
- Should be left as None for typical usage
- Does not allow switching between TensorFlow/JAX/PyTorch backends

## Motivation
Fixes issue #21711 where users attempt to pass keras.backend or other modules, resulting in AttributeError. The current documentation suggests this parameter offers more flexibility than it actually provides.

## Current documentation & proposed change:
```
backend_module: Backend module to use (if different from the default
    backend).
```

Proposed change:
```
backend_module: Backend module to use (if different from the default
    backend). This parameter is primarily for internal use and should
    typically be left as None (default: `keras.src.backend`).
    Note: This does not allow switching between TensorFlow/JAX/PyTorch
    backends. Use `keras.config.set_backend()` to change the global backend.
```
## Testing
-  Documentation builds successfully
-  Existing tests pass
-  No functional changes, only documentation update

### DIFF
--- a/keras/src/utils/image_utils.py
+++ b/keras/src/utils/image_utils.py
@@ -354,7 +354,7 @@ def smart_resize(
             `"lanczos3"`, `"lanczos5"`.
             Defaults to `"bilinear"`.
         data_format: `"channels_last"` or `"channels_first"`.
-        Backend module to use (if different from the default
+        backend_module: Backend module to use (if different from the default
             backend). This parameter is for internal use and should typically
             be left as None. When None, defaults to `keras.src.backend`.
             Note: This does NOT allow switching between TensorFlow/JAX/PyTorch


fchollet:
I suggest just removing the arg from the documentation (and signature) entirely. It can just be hidden in `kwargs`.

ILCSFNO:
cc @Flakes342 https://github.com/keras-team/keras/pull/21722#issuecomment-3393824132

ILCSFNO:
cc @Flakes342

PR #21723, opened 2025-10-07
Fix ModelParallel OOM issue during weight loading

amitsrivastava78:
- Modified load_own_variables() to use _direct_assign() for sharded variables
- Prevents loading full weight tensors on single device before distribution
- Resolves RESOURCE_EXHAUSTED errors when loading large models with ModelParallel
- Maintains backward compatibility for non-sharded variables
- Enables loading of models like Gemma2 2B/7B without OOM errors
- Added EinsumDense layer testing to ModelParallel sharded variable loading

### DIFF
--- a/keras/src/callbacks/orbax_checkpoint.py
+++ b/keras/src/callbacks/orbax_checkpoint.py
@@ -13,29 +13,27 @@ try:
 except ImportError:
     ocp = None
 
-# Expose orbax classes for testing purposes
+# Expose advanced Orbax functionality for users who need direct access
+# These are provided as bridge for advanced usecases like custom type handlers
 if ocp is not None:
+    # Core checkpointing classes
     CheckpointManager = ocp.CheckpointManager
-    CheckpointManagerOptions = ocp.CheckpointManagerOptions
-    CheckpointHandler = ocp.CheckpointHandler
-    CheckpointHandlerRegistry = ocp.CheckpointHandlerRegistry
     SaveArgs = ocp.SaveArgs
     StandardRestore = ocp.args.StandardRestore
-    JsonSave = ocp.args.JsonSave
-    # Expose type handler functionality for advanced users and testing
+
+    # Type handler functionality for custom serialization
     TypeHandler = ocp.type_handlers.TypeHandler
     register_type_handler = ocp.type_handlers.register_type_handler
+
+    # Direct checkpointing for custom objects
     PyTreeCheckpointer = ocp.PyTreeCheckpointer
-    # Expose metadata for testing
+
+    # Metadata functionality
     metadata = ocp.metadata
 else:
     CheckpointManager = None
-    CheckpointManagerOptions = None
-    CheckpointHandler = None
-    CheckpointHandlerRegistry = None
     SaveArgs = None
     StandardRestore = None
-    JsonSave = None
     TypeHandler = None
     register_type_handler = None
     PyTreeCheckpointer = None
--- a/keras/src/callbacks/orbax_checkpoint_test.py
+++ b/keras/src/callbacks/orbax_checkpoint_test.py
@@ -11,29 +11,24 @@ from keras.src import models
 from keras.src import testing
 
 try:
-    from keras.src.callbacks.orbax_checkpoint import CheckpointHandler
-    from keras.src.callbacks.orbax_checkpoint import CheckpointHandlerRegistry
+    # Import advanced Orbax functionality through the Keras bridge
     from keras.src.callbacks.orbax_checkpoint import CheckpointManager
-    from keras.src.callbacks.orbax_checkpoint import CheckpointManagerOptions
-    from keras.src.callbacks.orbax_checkpoint import JsonSave
     from keras.src.callbacks.orbax_checkpoint import OrbaxCheckpoint
     from keras.src.callbacks.orbax_checkpoint import PyTreeCheckpointer
     from keras.src.callbacks.orbax_checkpoint import SaveArgs
     from keras.src.callbacks.orbax_checkpoint import StandardRestore
     from keras.src.callbacks.orbax_checkpoint import TypeHandler
+    from keras.src.callbacks.orbax_checkpoint import metadata
     from keras.src.callbacks.orbax_checkpoint import register_type_handler
 except ImportError:
     OrbaxCheckpoint = None
-    CheckpointHandler = None
-    CheckpointHandlerRegistry = None
     CheckpointManager = None
-    CheckpointManagerOptions = None
-    JsonSave = None
     SaveArgs = None
     StandardRestore = None
     TypeHandler = None
     register_type_handler = None
     PyTreeCheckpointer = None
+    metadata = None
 
 
 class OrbaxCheckpointTest(testing.TestCase):
@@ -1180,8 +1175,8 @@ class OrbaxCheckpointTest(testing.TestCase):
 
         import asyncio
 
-        from keras.src.callbacks.orbax_checkpoint import TypeHandler
-        from keras.src.callbacks.orbax_checkpoint import metadata
+        # Use the classes imported through the Keras bridge
+        # TypeHandler and metadata are already imported above
 
         class MetadataHandler(TypeHandler):
             """A custom Orbax type handler to save/load the TrainingMetadata
@@ -1244,14 +1239,12 @@ class OrbaxCheckpointTest(testing.TestCase):
         )
 
         # 2. Register the type handler globally
-        from keras.src.callbacks.orbax_checkpoint import register_type_handler
-
         register_type_handler(
             ty=TrainingMetadata, handler=MetadataHandler(), override=True
         )
 
         # 3. Create a PyTreeCheckpointer to save the custom object directly
-        from keras.src.callbacks.orbax_checkpoint import PyTreeCheckpointer
+        # PyTreeCheckpointer is already imported above
 
         checkpointer = PyTreeCheckpointer()
 


keras/src/backend/jax/core.py
```diff
@@ -17,39 +17,293 @@
...
+            )
+        else:
+            variable._layout = tensor_layout
+            logging.debug(
+                f"{log_prefix}: Using layout directly: {variable._layout}"
+            )
+
+    # Log initialization details
+    total_elements = np.prod(variable._shape)
+    element_size = 4  # float32 = 4 bytes
```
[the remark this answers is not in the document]
amitsrivastava78: fixed

keras/src/backend/jax/core.py
```diff
@@ -17,39 +17,293 @@
...
+                    shard_data = [
+                        shard.data for shard in value.addressable_shards
+                    ]
+                    self._shard_references = [shard_data]
+                else:
+                    # For non-sharded arrays, hold a ref to the array itself.
+                    self._strong_reference = value
+            except Exception:
+                # If we can't set attributes (e.g., during tracing), skip
+                pass
```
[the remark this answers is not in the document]
amitsrivastava78: Fixed

keras/src/layers/core/dense.py
```diff
@@ -306,18 +306,34 @@ def load_own_variables(self, store):
...
+                target_variables.append(self.outputs_grad_amax_history)
+            elif self.quantization_mode == "gptq":
+                target_variables.append(self.kernel_scale)
+                target_variables.append(self.kernel_zero)
+                target_variables.append(self.g_idx)
+            else:
+                raise self._quantization_mode_error(self.quantization_mode)
+        for i, variable in enumerate(target_variables):
+            weight_data = store[str(i)]
+            variable._direct_assign(weight_data)
```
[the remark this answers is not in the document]
amitsrivastava78: Fixed

keras/src/layers/core/einsum_dense.py
```diff
@@ -374,18 +374,33 @@ def load_own_variables(self, store):
...
+                target_variables.append(self.outputs_grad_amax_history)
+            elif self.quantization_mode == "gptq":
+                target_variables.append(self.kernel_scale)
+                target_variables.append(self.kernel_zero)
+                target_variables.append(self.g_idx)
+            else:
+                raise self._quantization_mode_error(self.quantization_mode)
+        for i, variable in enumerate(target_variables):
+            weight_data = store[str(i)]
+            variable._direct_assign(weight_data)
```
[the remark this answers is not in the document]
amitsrivastava78: Fixed

keras/src/utils/variable_loading.py
```diff
@@ -0,0 +1,6 @@
+"""
+Utility functions for loading variables with sharded support.
+
+This module provides common utilities for loading variables that may be sharded
+across multiple devices, which is useful for distributed training scenarios.
+"""
```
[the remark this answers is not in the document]
amitsrivastava78: Fixed

keras/src/utils/variable_loading.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def get_quantized_variable_load_order(layer):
```
hertschuh: There is a conflicting PR that addresses this issue in a different way: https://github.com/keras-team/keras/pull/21713

Since these changes don't address the OOM but quantization and saving, let's remove these from this PR.

keras/src/backend/jax/core_test.py
```diff
@@ -7,15 +7,141 @@
...
 from keras.src import backend
+from keras.src import layers
+from keras.src import models
 from keras.src import testing
 from keras.src.backend.config import is_nnx_enabled
+from keras.src.backend.jax.core import JaxVariable
+from keras.src.backend.jax.core import _ProtectedShardedArray
+
+if is_nnx_enabled():
+    from keras.src.backend.jax.core import NnxVariable
```
hertschuh: This duplicates code from below. Bad merge?

keras/src/backend/jax/core.py
```diff
@@ -17,39 +18,296 @@
...
+    # Log initialization details
+    total_elements = np.prod(variable._shape)
+    element_size = np.dtype(variable.dtype).itemsize
+    total_size_mb = (total_elements * element_size) / (1024 * 1024)
+
+    logging.info(f"{log_prefix}: Creating variable '{variable.path}'")
+    logging.debug(
+        f"{log_prefix}: Shape: {variable._shape}, Size: {total_size_mb:.2f} MB"
+    )
+    logging.debug(f"{log_prefix}: Has layout: {variable._layout is not None}")
```
hertschuh: Remove all logging everywhere.

keras/src/backend/jax/core.py
```diff
@@ -17,39 +18,296 @@
...
+                )
+        elif not isinstance(value, np.ndarray):
+            value = np.array(value)
+            logging.debug(
+                f"{log_prefix}: Converted to numpy array (host memory)"
+            )
+        else:
+            logging.debug(
+                f"{log_prefix}: Value already numpy array (host memory)"
+            )
```
hertschuh: This is not a valid branch, value is always a JAX array

keras/src/backend/jax/core.py
```diff
@@ -17,39 +18,296 @@
...
+    if variable._layout is not None:
+        logging.info(
+            f"{log_prefix}: Sharded initialization (layout: {variable._layout})"
+        )
+
+        if isinstance(value, (jnp.ndarray, jax.Array)):
+            if hasattr(value, "device") and value.device.platform == "cpu":
+                logging.debug(
+                    f"{log_prefix}: JAX array already on CPU (host memory)"
+                )
```
hertschuh: The initializer created the variable value on "device:0". Do you're doing "device:0" -> "cpu" -> "multiple devices (potentially)".

I don't think it helps or is better than "device:0" -> "multiple devices (potentially)".

Let's say the sharding is such that "device:0" is actually the correct placement, than this code will move the variable twice for no reason.

keras/src/backend/jax/core.py
```diff
@@ -17,39 +18,296 @@
...
+            mem_reduction = (
+                (total_size_mb - shard_0_size_mb) / total_size_mb * 100
+            )
+            logging.debug(
+                f"{log_prefix}: Memory reduction: {mem_reduction:.1f}%"
+            )
+    else:
+        logging.debug(f"{log_prefix}: NORMAL (non-sharded) initialization")
+        # Convert to tensor using normal path
+        value = variable._convert_to_tensor(value)
```
hertschuh: Value is a tensor, not needed

keras/src/backend/jax/core.py
```diff
@@ -17,39 +18,296 @@
...
+        """
+
         if self._layout is not None:
+            logging.debug(
+                f"_direct_assign: Distributing variable '{self.path}' "
+                f"with layout"
+            )
+            logging.debug(
+                f"_direct_assign: Original value shape: {value.shape}"
+            )
```
hertschuh: Remove logging

keras/src/backend/jax/core.py
```diff
@@ -17,39 +18,296 @@
...
+
+            # Log sharding details
+            if hasattr(value, "sharding") and _safe_has_addressable_shards(
+                value
+            ):
+                shards = value.addressable_shards
+                num_devices = len(shards)
+                logging.debug(
+                    f"_direct_assign: Sharded across {num_devices} devices"
+                )
```
hertschuh: Logging only, remove

keras/src/backend/jax/core.py
```diff
@@ -17,39 +18,296 @@
...
+            )
+            # Calculate memory reduction percentage
+            mem_reduction = (
+                (total_size_mb - shard_0_size_mb) / total_size_mb * 100
+            )
+            logging.debug(
+                f"{log_prefix}: Memory reduction: {mem_reduction:.1f}%"
+            )
+    else:
+        logging.debug(f"{log_prefix}: NORMAL (non-sharded) initialization")
```
hertschuh: This is logging only, remove.

keras/src/backend/jax/core.py
```diff
@@ -17,39 +18,296 @@
...
+        if hasattr(self, "_trace_state") and SymbolicScope():
+            return
+
+        if isinstance(value, jax.Array):
+            try:
+                # Check if this is a JAX tracer (during compilation/tracing)
+                if jax_utils.is_in_jax_tracing_scope(value):
+                    # During tracing, we can't access addressable_shards
+                    # Just hold a reference to the tracer itself
+                    self._strong_reference = value
```
hertschuh: You can keep references to tracers. They are short live and only valid during one call. If you access them outside of that, you get errors.

keras/src/backend/jax/core.py
```diff
@@ -187,24 +445,59 @@ def __setstate__(self, state):
...
+                if isinstance(value, np.ndarray):
+                    logging.debug("_direct_assign (NNX): Value is numpy array")
+                elif isinstance(value, (jnp.ndarray, jax.Array)):
+                    logging.debug("_direct_assign (NNX): Value is JAX array")
+
+                # Distribute
                 value = distribution_lib.distribute_variable(
                     value, self._layout
                 )
+                logging.debug("_direct_assign (NNX): Distributed successfully")
```
hertschuh: The changes are logging only, revert.

keras/src/backend/jax/core.py
```diff
@@ -283,6 +578,8 @@ def is_tensor(x):
 
 
 def shape(x):
+    # This will work as long as we disallow
+    # dynamic shapes in JAX.
```
hertschuh: Spurious changes from a failed merge, revert.

keras/src/backend/jax/core.py
```diff
@@ -314,9 +611,6 @@ def compute_output_spec(fn, *args, **kwargs):
             else:
                 maybe_symbolic_kwargs[k] = v
 
-        # Create a _DimExpr instance for one dimension by creating a symbolic
-        # shape with one dimension and extracting it.
-        #
```
hertschuh, line 319: Spurious changes from a failed merge, revert.

keras/src/backend/jax/core.py
```diff
@@ -17,39 +18,296 @@
...
+        self._is_sharded = _safe_has_addressable_shards(array)
+
+    def __getattr__(self, name):
+        # Delegate all attribute access to the wrapped array
+        return getattr(self._array, name)
+
+    def delete(self):
+        """Intercept delete() calls and prevent deletion of sharded arrays."""
+        if self._is_sharded:
+            # Don't actually delete sharded arrays
```
hertschuh: Meaning it's permanently leaked?

keras/src/backend/jax/core.py
```diff
@@ -17,39 +18,296 @@
...
+        isinstance(x, jax.Array)
+        and not jax_utils.is_in_jax_tracing_scope(x)
+        and hasattr(x, "addressable_shards")
+    )
+
+
+class _ProtectedShardedArray:
+    """Wrapper that prevents deletion of sharded JAX arrays.
+
+    This wrapper intercepts delete() calls from jax_memory_cleanup
```
hertschuh: Ok, so the problem stems from `jax_memory_cleanup`.

And all this very complex logic (many hundreds of lines of code) does is undo what `jax_memory_cleanup` is doing. So the obvious choice is to remove `jax_memory_cleanup` (4 lines of code) instead. The way `jax_memory_cleanup` works is intrinsically dangerous and we should find a better way to do it.

keras/src/backend/jax/core.py
```diff
@@ -17,39 +18,296 @@
...
+def _safe_has_addressable_shards(x):
+    """Safely check if x has addressable_shards without tracer errors."""
+    return (
+        isinstance(x, jax.Array)
+        and not jax_utils.is_in_jax_tracing_scope(x)
+        and hasattr(x, "addressable_shards")
+    )
+
+
+class _ProtectedShardedArray:
```
hertschuh: Passing array wrappers to JAX is not supported. This will cause use trouble: https://github.com/keras-team/keras/pull/21719

hertschuh (commented):
@amitsrivastava78 

One overall note: We basically never log in the successful case, the logs are just way too noisy.

Can you create two separate small PRs:
- One for the fix for the initializer. My reading of it is that this is the only change needed:
```python
    def _initialize_with_initializer(self, initializer):
        """Initialize variable with initializer, running on CPU if sharding
        is needed."""
        if self._layout is not None:
            # For sharded variables, run initializer on CPU to avoid device
            # placement issues
            with jax.default_device(jax.devices("cpu")[0]):
                value = self._convert_to_tensor(
                    initializer(self._shape, dtype=self._dtype)
                )
        else:
            # For non-sharded variables, use the default behavior
            value = self._convert_to_tensor(
                initializer(self._shape, dtype=self._dtype)
            )
        self._initialize(value)
```
But we should check that it does what we think it does.

- One for the fix for the weight loading. It will have this change, and only this change, for the relevant layers:
```python
        for i, variable in enumerate(target_variables):
            variable._direct_assign(store[str(i)])
```

And `jax_memory_cleanup` should be addressed differently, which we can talk about.

keras/src/backend/jax/core.py
amitsrivastava78: This file changes are removed and no longer needed

keras/src/backend/jax/core.py
amitsrivastava78: Actually you are right, let me explain you what is happening
1. During model build time, the Jax Array is created 
2. Since Now during model loading time, since Jax arrays are immutable they will be duplicated.
3. No idea was that eventually gc will  run and delete the old reference
Now this is the reason to introduce the jax_memory deletion as a way to aggressively delete the memory between build time of the model and weight loading time, so that duplication of memory should not happen.
Now in ModelParallel approach that we have done, we do not need this aggressive approach as this rolling, variable-by-variable approach does not put so much strain  on memory.
So yes i have reverted all these change and in keras hub i will call jax_memory_cleanup conditionally i.e this will not be called if ModelParallel is active, let me know what you think

keras/src/backend/jax/core.py
amitsrivastava78: Changed no longer present

keras/src/backend/jax/core.py
amitsrivastava78: changes no longer present

keras/src/backend/jax/core.py
amitsrivastava78: changes no longer present

keras/src/backend/jax/core.py
amitsrivastava78: changes no longer present

keras/src/backend/jax/core.py
amitsrivastava78: changes no longer present

keras/src/backend/jax/core.py
amitsrivastava78: changes no longer present

keras/src/backend/jax/core.py
amitsrivastava78: changes no longer present

keras/src/backend/jax/core.py
amitsrivastava78: changes no longer present

keras/src/backend/jax/core.py
amitsrivastava78: changes no longer present

keras/src/backend/jax/core.py
amitsrivastava78: changes no longer present

keras/src/backend/jax/core.py
amitsrivastava78: changes no longer present

keras/src/backend/jax/core.py
amitsrivastava78: Changes no longer present

keras/src/backend/jax/core_test.py
amitsrivastava78: ok removed

keras/src/utils/variable_loading.py
amitsrivastava78: Sure

amitsrivastava78:
> This looks good to me. But what about `base_conv`, `embedding`? Don't they need this change too?
> 
> https://github.com/search?q=repo%3Akeras-team%2Fkeras%20variable.assign(store%5Bstr(i)%5D)&type=code
> 
> Oh and what about `base_optimizer`?

Yes base_conv, embedding do have weights which needs to be sharded, same for base_optimizer

amitsrivastava78:
> Taking a closer look at this, I really don't understand why calling `variable._direct_assign` instead of `variable.assign` makes any difference.
> 
> The implementation of `variable.assign` calls `variable._direct_assign` after doing the shape check (which you've now replicated 8 times).
> 
> `_direct_assign` is just the backend dependent implementation of the assignment.
> 
> Can you explain why it changes anything?

Ah.. yes you are right, the only issue then is the jax_memory_cleanup , should we remove this completely, or conditionally skip memory delete if sharding is present ?

keras/src/backend/jax/distribution_lib_test.py
```diff
@@ -33,6 +33,14 @@
     reason="Backend specific test",
 )
 class JaxDistributionLibTest(testing.TestCase):
+    def _require_min_devices(self, min_devices):
```
hertschuh, line 36: How come this was not needed before?
Line 27 should make it work.
Under what circumstances did you need this?

PR #21725, opened 2025-10-07
Add debug mode

pass-lin:
In many cases we need to pinpoint the exact line that throws an error.  
Keras, however, currently hides this information because the block

```python
try:
    return fn(*args, **kwargs)
except Exception as e:
    if hasattr(e, "_keras_call_info_injected"):
        # Only inject info for the innermost failing call
        raise e
```

swallows the original traceback.  
All you see is the exception message, not the precise location where it occurred.  

The new debug mode removes this wrapper and lets the raw backend exception propagate, so you can see the exact file and line number that triggered the failure.


fchollet:
Please note: you can already use `disable_traceback_filtering` which displays full stack traces.

PR #21726, opened 2025-10-08
Enable Automatic Tensor Parallelism

buildwithsuhana:
This pull request introduces a comprehensive framework for automatic tensor parallelism in Keras, enabling users to train models that are too large to fit on a single accelerator. The core of this feature is a new distribution strategy, AutoTPDistribution, which provides a simple, high-level API to shard an existing Keras model across multiple devices.

Description
This framework is designed to automate the complex process of model sharding and inter-device communication, making large-scale model training more accessible. The implementation is broken down into several key components:

1. High-Level User API: AutoTPDistribution
A new distribution strategy, AutoTPDistribution, is introduced in keras/src/distribution/distribution_lib.py. This class serves as the primary user entry point. The workflow is straightforward:

- A user defines their hardware topology using a DeviceMesh.
- They instantiate the AutoTPDistribution strategy.
- They pass their standard Keras model to the distribution.shard() method.
- This returns a new, sharded model instance ready for distributed training.

2. Core Sharding Engine: TensorParallelKeras
A new TensorParallelKeras model class (keras.src.distribution.tensor_parallel.tensor_parallel.py) acts as the core engine. When distribution.shard() is called, this class wraps the original model and performs the following actions:

Auto-Configures Hardware: Discovers and assigns available devices (TPU, GPU, or CPU).

Shards Parameters: It analyzes the model's layers and applies column-parallel or row-parallel sharding strategies to the weights and biases of relevant layers (e.g., Dense).

Builds a Unified Graph: It creates a single, assembled Keras Functional model that internally manages the parallel computation. This clever design encapsulates the communication logic (e.g., AllGather, ReduceScatter) within the model's call graph, simplifying the execution and enabling JIT compilation. Partial outputs from each device shard are correctly combined (e.g., concatenation for column-parallel, summation for row-parallel).

Coordinates Gradients: It overrides the compile method to wrap the user's optimizer in a TensorParallelOptimizer, which handles the synchronized computation and application of gradients across all shards.

Example usage: https://colab.research.google.com/drive/1UAINIcstDuO0aeA9lxCF5LaIj5ne5X5z?resourcekey=0-pPF4COO19KRoqS5cpWNILA&usp=sharing

This is the 4th (out of 4) PR for AutoSharding Keras.

### DIFF
--- a/keras/src/distribution/distribution_lib.py
+++ b/keras/src/distribution/distribution_lib.py
@@ -985,6 +985,7 @@ class AutoTPDistribution(Distribution):
         self._num_process = distribution_lib.num_processes()
         self._process_id = distribution_lib.process_id()
         self._is_multi_process = self._num_process > 1
+
         from keras.src.distribution.tensor_parallel.tensor_parallel import (
             TensorParallelKeras,
         )
@@ -996,22 +997,20 @@ class AutoTPDistribution(Distribution):
         )
 
     def get_data_layout(self, data_shape):
+        """Shard data on the 'data' axis of the mesh."""
         data_shard_spec = [None] * len(data_shape)
         data_shard_spec[0] = self.batch_dim_name
         return TensorLayout(data_shard_spec, self.device_mesh)
 
     def get_variable_layout(self, variable):
-        warnings.warn(
-            "Variable layout is determined automatically within "
-            "AutoTPDistribution. This method will return a replicated layout."
-        )
-        return TensorLayout([None] * len(variable.shape), self.device_mesh)
+        return None
 
     def get_tensor_layout(self, path):
+        """Intermediate tensors do not require automated layout propagation."""
         return None
 
     def distribute_dataset(self, dataset):
-        """Distributes the dataset across processes based on the device mesh."""
+        """Distributes the dataset across processes based on the 'data' axis."""
         if not self._is_multi_process or not self.auto_shard_dataset:
             return dataset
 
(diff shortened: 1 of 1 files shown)


keras/src/distribution/tensor_parallel/tensor_parallel.py
```diff
@@ -0,0 +1,349 @@
...
+from keras.src.distribution.tensor_parallel.parameter_sharding import (
+    make_parameter_sharded_model,
+)
+from keras.src.layers import Add
+from keras.src.layers import Input
+from keras.src.layers import Lambda
+from keras.src.models import Model
+
+
+class TensorParallelKeras(Model):
```
hertschuh, line 25: The whole point of `distribution_lib` is that you don't have to do that. People should not need to change the base class of their existing models. Otherwise none of the KerasHub models will support this.

The `Distribution` object has the right injection point to do this.

keras/src/distribution/distribution_lib.py
```diff
@@ -896,3 +896,183 @@ def set_distribution(value):
...
+        data_shard_spec = [None] * len(data_shape)
+        data_shard_spec[0] = self.batch_dim_name
+        return TensorLayout(data_shard_spec, self.device_mesh)
+
+    def get_variable_layout(self, variable):
+        warnings.warn(
+            "Variable layout is determined automatically within "
+            "AutoTPDistribution. This method will return a replicated layout."
+        )
+        return TensorLayout([None] * len(variable.shape), self.device_mesh)
```
hertschuh: This is your injection point. This is where you should return the layout for each variable from your LayoutMap.

keras/src/distribution/distribution_lib.py
```diff
@@ -896,3 +896,183 @@ def set_distribution(value):
...
+
+    def get_tensor_layout(self, path):
+        return None
```
hertschuh, line 1010: This is your injection point. This is where you should return the output layout from your `LayoutMap.state_rules`.

hertschuh: This is your injection point. This is where you should return the output layout from your `LayoutMap.output_rules`.

hertschuh:
@buildwithsuhana 

I'm closing this old PR as I don't think this will be done this way. Feel free to reopen when the new implementation is ready.

PR #21729, opened 2025-10-09
Fix histogram op for symbolic inputs

### LINKED ISSUE #21708: TF backend: keras.ops.histogram calls .numpy() in graph mode and crashes with AttributeError on symbolic tensors
# Summary
On the TensorFlow backend, `keras.ops.histogram` crashes in graph mode because the TF implementation converts a tensor to a Python list via `.numpy().tolist()`. This is not graph-safe for symbolic tensors in a Keras Functional/Model context. The same code works on the JAX backend.

Minimal reproducible example

```python
# Repro: Keras 3 + TensorFlow backend, graph mode, jit or not
import os

import tensorflow as tf
import keras

tf.config.run_functions_eagerly(False)  # ensure graph

class HistogramProbe(keras.layers.Layer):
    def __init__(self, bins=8, lo=0.0, hi=1.0, **kwargs):
        super().__init__(**kwargs)
        self.bins = int(bins)
        self.lo = float(lo)
        self.hi = float(hi)

    def call(self, x):
        flat = keras.ops.reshape(x, (-1,))
        # This triggers TF backend histogram implementation which does .numpy().tolist()
        counts, edges = keras.ops.histogram(flat, bins=self.bins, range=(self.lo, self.hi))
        return x  # identity on data path

inp = keras.Input(shape=(16,), name="inp")
h = keras.layers.Dense(8, activation="relu")(inp)
h = HistogramProbe(name="probe")(h)
out = keras.layers.Dense(1)(h)
model = keras.Model(inp, out)

model.compile(optimizer="adam", loss="mse", jit_compile=True)  # jit_compile can be True or False
x = tf.random.uniform((32, 16))
y = tf.zeros((32, 1))

# Fails with AttributeError in graph mode
model.fit(x, y, epochs=1, batch_size=8)
```

See https://colab.research.google.com/drive/19R8qt7UjmX6Qz4YNkmz7KWWe3wO27MEw?usp=sharing

# Observed error (trimmed)

```
AttributeError: Exception encountered when calling HistogramProbe.call().

Could not automatically infer the output shape / dtype of 'probe' ...
Error encountered:
'SymbolicTensor' object has no attribute 'numpy'

Arguments received by HistogramProbe.call():
  • args=('<KerasTensor shape=(None, 8), dtype=float32, ...>',)
```

Root cause (source references)
In the TF backend implementation of `keras.ops.histogram`, the code does:

```python
bin_edges = tf.linspace(min_val, max_val, bins + 1)
bin_edges_list = bin_edges.numpy().tolist()  # <-- requires eager tensor
bin_indices = tf.raw_ops.Bucketize(input=x, boundaries=bin_edges_list[1:-1])
```

https://github.com/keras-team/keras/blob/b491c860fc2750e2b6006b55358d3251dbb4a9f0/keras/src/backend/tensorflow/numpy.py#L2966C1-L2966C48

Calling `.numpy()` on a symbolic tensor in a graph causes the crash.

# Expected behavior
`keras.ops.histogram` should be graph-safe on the TensorFlow backend (or clearly documented as eager-only). Ideally it should run inside a Keras model without errors.

# Environment
Please replace the versions with your actual outputs:
```
keras: 3.11.3
tensorflow: 2.19.0
jax: 0.5.3
python: 3.12.11 (main, Jun  4 2025, 08:56:18) [GCC 11.4.0]
platform: Linux-6.6.97+-x86_64-with-glibc2.35
backend: tensorflow
```

# Workarounds

Something like:

```python
def tf_histogram_fixed_width_1d(x, bins, range):
    lo = range[0]
    hi = range[1]
    x = tf.reshape(x, [-1])
    x = tf.clip_by_value(x, lo, hi)
    counts = tf.histogram_fixed_width(values=x, value_range=[lo, hi], nbins=bins)
    return counts, tf.linspace(lo, hi, bins + 1)
```

However, I think there might be slight edge issues around lo and hi.

# Notes

* The same repro runs fine on the JAX backend.
* The failure happens whether `jit_compile=True` or `False`, as long as the call happens in graph mode / symbolic context.

Thanks!

abheesht17:
Resolves #21708

On the TensorFlow backend, `ops.histogram` converts a tensor to a Python list [[here](https://github.com/keras-team/keras/blob/4e1edcf1869fceae3c9ead4477211f7d70989c7e/keras/src/backend/tensorflow/numpy.py#L3164)], which won't work in graph mode with symbolic tensors. The current piece of code uses `tf.raw_ops.Bucketize` which necessarily requires `boundaries` to be a Python list [see [documentation](https://www.tensorflow.org/api_docs/python/tf/raw_ops/Bucketize)].

This PR:

- Replaces `tf.raw_ops.Bucketize` with `tf.searchsorted`;
- Adds a unit test to make sure the op runs with `predict`;
- Also adds a missing test (unrelated to the specific issue linked above), which ensures we are assigning values which are "on the boundaries" to the correct bucket. This is an important unit test to have.

We should probably add a unit test which tests `model.predict` for every op at some point.

Note: Skipping the NumPy test for `predict`. Seems to be an unrelated issue. We can solve it separately.

### DIFF
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -3165,7 +3165,7 @@ def histogram(x, bins=10, range=None):
     bin_indices = tf.searchsorted(bin_edges[1:-1], x, side="right")
 
     # tf.math.bincount does not work with XLA in this case. So, we use
-    # `tensor_scatter_nd_add`.
+    # `scatter_nd`.
     bin_counts = tf.scatter_nd(
         indices=tf.expand_dims(bin_indices, axis=-1),
         updates=tf.ones_like(bin_indices, dtype=x.dtype),


abheesht17:
Works with `jit_compile = False`, but does not work with `jit_compile = True`. Trying to figure out another way

```
FAILED keras/src/ops/numpy_test.py::HistogramTest::test_histogram_predict - tensorflow.python.framework.errors_impl.InvalidArgumentError: Graph execution error:

Detected at node functional_1/histogram_layer_1/bincount/DenseBincount defined at (most recent call last):
<stack traces unavailable>
Input 1 to node `functional_1/histogram_layer_1/bincount/DenseBincount` with op DenseBincount must be a compile-time constant.

XLA compilation requires that operator arguments that represent shapes or dimensions be evaluated to concrete values at compile time. This error means that a shape or dimension argument could not be evaluated at compile time, usually because the value of the argument depends on a parameter to the computation, on a variable, or on a stateful operation such as a random number generator.
```

abheesht17:
> Works with `jit_compile = False`, but does not work with `jit_compile = True`. Trying to figure out another way
> 
> ```
> FAILED keras/src/ops/numpy_test.py::HistogramTest::test_histogram_predict - tensorflow.python.framework.errors_impl.InvalidArgumentError: Graph execution error:
> 
> Detected at node functional_1/histogram_layer_1/bincount/DenseBincount defined at (most recent call last):
> <stack traces unavailable>
> Input 1 to node `functional_1/histogram_layer_1/bincount/DenseBincount` with op DenseBincount must be a compile-time constant.
> 
> XLA compilation requires that operator arguments that represent shapes or dimensions be evaluated to concrete values at compile time. This error means that a shape or dimension argument could not be evaluated at compile time, usually because the value of the argument depends on a parameter to the computation, on a variable, or on a stateful operation such as a random number generator.
> ```

Alright, played some ops golf here, and replaced `bincount` with `scatter_nd_add`. Seems to work now

keras/src/backend/tensorflow/numpy.py
```diff
@@ -3161,10 +3161,14 @@ def histogram(x, bins=10, range=None):
...
-    bin_indices = tf.raw_ops.Bucketize(input=x, boundaries=bin_edges_list[1:-1])
-
-    bin_counts = tf.math.bincount(
-        bin_indices, minlength=bins, maxlength=bins, dtype=x.dtype
+    bin_edges = tf.cast(bin_edges, x.dtype)
+    bin_indices = tf.searchsorted(bin_edges[1:-1], x, side="right")
+
+    # tf.math.bincount does not work with XLA in this case. So, we use
+    # `tensor_scatter_nd_add`.
+    bin_counts = tf.tensor_scatter_nd_add(
```
hertschuh: It looks like you can just use `tf.scatter_nd`:
- it creates a new tensor with zeros
- it also sums in the case of duplicate indices

keras/src/backend/tensorflow/numpy.py
```diff
@@ -3161,10 +3161,14 @@ def histogram(x, bins=10, range=None):
 
     x = tf.boolean_mask(x, (x >= min_val) & (x <= max_val))
     bin_edges = tf.linspace(min_val, max_val, bins + 1)
-    bin_edges_list = bin_edges.numpy().tolist()
-    bin_indices = tf.raw_ops.Bucketize(input=x, boundaries=bin_edges_list[1:-1])
```
hertschuh, line 3165: So `Bucketize` is not XLA compilable?

This table implies it is https://www.tensorflow.org/api_docs/python/tf/raw_ops

abheesht17: `Bucketize` necessarily requires `boundaries` to be a Python list

keras/src/backend/tensorflow/numpy.py
abheesht17: Thanks, done!

PR #21732, opened 2025-10-10
Update Torch and Tensorflow versions in cuda requirements files.

hertschuh:
Replacement for https://github.com/keras-team/keras/pull/21704

Also:
- Disabled an ONNX export test for Torch that was already disabled on GPU with both JAX and Tensorflow.
- Moved install for `tf_keras` from `requirements.txt` to `action.yml` using the `--no-deps` option because `tf_keras` depends on `tensorflow`, which installs the non-CPU version of TensorFlow and causes issues with CPU tests.

### DIFF
--- a/keras/src/export/onnx_test.py
+++ b/keras/src/export/onnx_test.py
@@ -77,9 +77,11 @@ def get_model(type="sequential", input_shape=(10,), layer_list=None):
         "backends."
     ),
 )
-@pytest.mark.skipif(testing.jax_uses_gpu(), reason="Leads to core dumps on CI")
 @pytest.mark.skipif(
-    testing.tensorflow_uses_gpu(), reason="Leads to core dumps on CI"
+    testing.jax_uses_gpu()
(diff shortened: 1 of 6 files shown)


keras/src/export/onnx_test.py
```diff
@@ -77,9 +77,11 @@ def get_model(type="sequential", input_shape=(10,), layer_list=None):
         "backends."
     ),
 )
-@pytest.mark.skipif(testing.jax_uses_gpu(), reason="Leads to core dumps on CI")
 @pytest.mark.skipif(
-    testing.tensorflow_uses_gpu(), reason="Leads to core dumps on CI"
+    testing.jax_uses_gpu()
+    or testing.tensorflow_uses_gpu
```
abheesht17: `testing.tensorflow_uses_gpu` --> `testing.tensorflow_uses_gpu()`?

hertschuh: Oh, good catch. Fixed.

PR #21737, opened 2025-10-12
Remove the unused jax `enable_x64`.

james77777778:
Following #21604 and #21734, we no longer need `enable_x64` in our tests since all 64bit tests were removed by default.

This PR further cleans up the test suit.

### DIFF
--- a/keras/src/ops/numpy_test.py
+++ b/keras/src/ops/numpy_test.py
@@ -8841,11 +8841,8 @@ class NumpyDtypeTest(testing.TestCase):
         # for bool inputs
         if dtype == "bool":
             expected_dtype = "int32"
-        # TODO: Remove the condition of uint8 and uint16 once we have
-        # jax>=0.4.27 for both CPU & GPU environments.
-        # uint8 and uint16 will be casted to uint32 when jax>=0.4.27 but to
-        # int32 otherwise.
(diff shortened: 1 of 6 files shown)


keras/src/ops/numpy_test.py
```diff
@@ -9031,32 +8821,22 @@ def test_tile(self, dtype):
...
-            self.assertDType(knp.Trace().symbolic_call(x), expected_dtype)
+        x = knp.ones((1, 1, 1), dtype=dtype)
+        x_jax = jnp.ones((1, 1, 1), dtype=dtype)
+        expected_dtype = standardize_dtype(jnp.trace(x_jax).dtype)
+        # jnp.trace is buggy with bool. We set the expected_dtype to int32
+        # for bool inputs
+        if dtype == "bool":
+            expected_dtype = "int32"
+        # TODO: Remove the condition of uint8 and uint16 once we have
+        # jax>=0.4.27 for both CPU & GPU environments.
```
hertschuh: This is already true. We have 0.5.0 on CPU and 0.6.2 on GPU.

james77777778: I have updated the test and the impl of `trace` in all backends.

hertschuh (approved):

Issue #20910, opened 2025-02-15
🚀OpenVINO Backend: Good-first-issues for Further Development🚀


rkazants:
🚀**A great opportunity to contribute to two popular AI projects with just one PR:: [Keras 3](https://github.com/keras-team/keras) and [OpenVINO](https://github.com/openvinotoolkit/openvino).**🚀 

Keras 3 enables seamless switching between supported backends—PyTorch, TensorFlow, and JAX—for both training and inference of traditional models and LLMs/GenAI pipelines. Since Keras 3.8.0, we've introduced a preview version of the OpenVINO backend (for inference only), allowing developers to leverage OpenVINO for model predictions directly within Keras 3 workflows. Activating the OpenVINO backend requires just one line of code to run inference on Keras 3-trained models. Here’s an example for a BERT model from Keras Hub:

```
import os
os.environ["KERAS_BACKEND"] = "openvino"
import numpy as np
import keras
import keras_hub

features = {
    "token_ids": np.ones(shape=(2, 12), dtype="int32"),
    "segment_ids": np.array([[0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 0, 0]] * 2),
    "padding_mask": np.array([[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0]] * 2),
}

classifier = keras_hub.models.BertTextClassifier.from_preset("bert_base_en_uncased", num_classes=4, preprocessor=None)
output = classifier.predict(features)
```

Currently, the OpenVINO backend lacks support for some operations. Our goal is to resolve this gap and to optimize it for inference on Intel devices—including CPUs, integrated GPUs, discrete GPUs, and NPUs—by supporting as many models as possible while delivering optimal performance. We aim to make the OpenVINO backend **the No. 1 choice for model inference within the Keras 3 workflow**.

We warmly welcome you to participate in further development of the OpenVINO backend. Here is a list of good-first-issues (it will be periodically updated with new ones):
- https://github.com/openvinotoolkit/openvino/issues/29008
- https://github.com/openvinotoolkit/openvino/issues/29009
- https://github.com/openvinotoolkit/openvino/issues/29010
- https://github.com/openvinotoolkit/openvino/issues/29011
- https://github.com/openvinotoolkit/openvino/issues/29012
- https://github.com/openvinotoolkit/openvino/issues/29013
- https://github.com/openvinotoolkit/openvino/issues/29014

chiruu12:
.take

[...]

PR #21739, opened 2025-10-13
Add `eye` support for OpenVINO backend

danielenricocahall:
Per https://github.com/keras-team/keras/issues/20910, add `eye` support for OpenVINO backend.


keras/src/backend/openvino/excluded_concrete_tests.txt
```diff
@@ -1,5 +1,4 @@
 NumPyTestRot90
-NumpyArrayCreateOpsCorrectnessTest::test_eye
```
hertschuh, line 2: Please also remove `NumpyDtypeTest::test_eye` line 30

PR #21742, opened 2025-10-14
Raise exception on batch_size mismatch for stateful RNNs

danielenricocahall:
Just resurrecting [this PR](https://github.com/keras-team/keras/pull/21249), along with making changes based on the feedback it originally received, as I think it's a valuable fix.

### DIFF
--- a/keras/src/layers/rnn/rnn.py
+++ b/keras/src/layers/rnn/rnn.py
@@ -390,6 +390,7 @@ class RNN(Layer):
             actual_batch_size = ops.shape(sequences)[0]
             if (
                 self._expected_batch_size is not None
+                and actual_batch_size is not None
                 and actual_batch_size != self._expected_batch_size
             ):
                 raise ValueError(


keras/src/layers/rnn/rnn.py
```diff
@@ -382,6 +382,25 @@ def call(
...
+        if self.stateful:
+            actual_batch_size = ops.shape(sequences)[0]
+            expected_batch_size = (
+                ops.shape(tree.flatten(self.states)[0])[0]
+                if self.states
+                else None
+            )
+            if (
+                expected_batch_size is not None
+                and actual_batch_size != expected_batch_size
```
fchollet: `actual_batch_size` being `None` would also be valid, in fact (and it can happen with TensorFlow)

fchollet (commented):
Thanks for the PR! I agree it's a useful feature.

keras/src/layers/rnn/rnn.py
```diff
@@ -382,6 +382,25 @@ def call(
                 initial_state = self.get_initial_state(
                     batch_size=ops.shape(sequences)[0]
                 )
+        if self.stateful:
+            actual_batch_size = ops.shape(sequences)[0]
+            expected_batch_size = (
```
fchollet: You should compute this when we create the states, and cache it as a property (it's expensive to recompute at every call in eager mode)

keras/src/layers/rnn/rnn.py
danielenricocahall: Ah okay I didn't realize that! Thank you, I added a condition to ensure `actual_batch_size` being `None` wouldn't raise an exception then

keras/src/layers/rnn/rnn.py
danielenricocahall: Thank you! Just updated by adding a property `_expected_batch_size` which computes this in the `build` method after `states` is set.

keras/src/layers/rnn/rnn.py
```diff
@@ -382,6 +386,21 @@ def call(
                 initial_state = self.get_initial_state(
                     batch_size=ops.shape(sequences)[0]
                 )
+        if self.stateful:
+            actual_batch_size = ops.shape(sequences)[0]
+            if (
+                self._expected_batch_size is not None
+                and actual_batch_size is not None
```
hertschuh: In the case of a non-constant batch size, this is dynamic and can be something that's not `None` and not an `int`. It can be a `tf.Tensor` for instance.

So instead of `actual_batch_size is not None` please do `not instance(actual_batch_size, int)`.

PR #21743, opened 2025-10-15
Fix several operations for OpenVINO backend + enable testing

danielenricocahall:
- Add `floor_divide` to list of functions to skip testing (they were being skipped due to `test_floor` being in `test_floor_divide`)
- correct implementation of the `floor` function and enable tests
- correct implementation of `isfinite` and enable tests
- correct implementation of `isinf ` and enable tests
- correct implementation of `isnan ` and enable tests
- update `isin_` skip to ensure `isinf` is not skipped due to the current skipping logic

These can be decomposed into more granular PRs if needed!  Also, not sure if there's a place to include the common logic.

### DIFF
--- a/keras/src/layers/core/dense.py
+++ b/keras/src/layers/core/dense.py
@@ -258,25 +258,25 @@ class Dense(Layer):
         if not self.built:
             return
         mode = self.quantization_mode
-        if mode not in self.quantization_variable_spec:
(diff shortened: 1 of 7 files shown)


danielenricocahall:
ah, I didn't run some of the tests locally 🙃 inf and finite checking are gnarly so I'm going to close this out, probably reopen one scoped down to floor only

PR #21751, opened 2025-10-16
`StringLookup` & `IntegerLookup` now save vocabulary loaded from file

hertschuh:
in the `.keras` archive when they are initialized with a path to a vocabulary file. This makes the `.keras` archive fully self contained.

This was already the behavior when using either `set_vocabulary(path)` or `adapt`. Simply, this behavior was extended to the case when `__init__` is called with a vocabulary file.

Note that this is technically a breaking change. Previously, upon doing `keras.saving.load_model`, it would be looking up the vocabulary file at the exact same path as when originally constructed.

Also disallow loading an arbitrary vocabulary file during model loading with `safe_mode=True` since the vocabulary file should now come from the archive.

### DIFF
--- a/keras/src/layers/preprocessing/index_lookup.py
+++ b/keras/src/layers/preprocessing/index_lookup.py
@@ -4,6 +4,7 @@ import numpy as np
 
 from keras.src import backend
 from keras.src.layers.layer import Layer
+from keras.src.saving import serialization_lib
 from keras.src.utils import argument_validation
 from keras.src.utils import numerical_utils
 from keras.src.utils import tf_utils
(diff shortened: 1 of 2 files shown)


keras/src/layers/preprocessing/index_lookup.py
```diff
@@ -382,6 +388,12 @@ def set_vocabulary(self, vocabulary, idf_weights=None):
             )
 
         if isinstance(vocabulary, str):
+            if serialization_lib.in_safe_mode():
+                raise ValueError(
+                    "Loading vocabulary files outside of the model archive "
+                    "being reloaded is not allowed"
```
fchollet: Please add details to the error message, in particular inform users of the workaround (turning off safe mode)

hertschuh: Augmented message with similar verbiage from other safe mode errors.

PR #21754, opened 2025-10-17
Fix: handle 3D input for channels_first in preprocess_input()  #21749

### LINKED ISSUE #21749: Bug in func `keras.applications.vgg16.preprocess_input()` when `x` in 3D and `data_format=='channels_first'`
## Bug Issue

The doc of `keras.applications.vgg16.preprocess_input()` shows its description as below:

https://github.com/keras-team/keras/blob/a5e6d0a94f19ca40c20056b1d3d3fbeda72a989d/keras/src/applications/imagenet_utils.py#L41-L44

https://github.com/keras-team/keras/blob/a5e6d0a94f19ca40c20056b1d3d3fbeda72a989d/keras/src/applications/imagenet_utils.py#L54-L66

For the repro below, which is expected to work well, I found a bug that `keras.applications.vgg16.preprocess_input` can't deal correctly with `x` in 3D and `data_format=='channels_first'`, using `tf 2.20.0` and `keras` latest:

### Repro
```python
import tensorflow as tf
import numpy as np
import keras
data_format = 'channels_first'
image = np.random.randint(0, 256, size=(224, 224, 3), dtype=np.uint8)
image = tf.convert_to_tensor(image, dtype=tf.float32)
image = tf.transpose(image, perm=[2, 0, 1])  # Transpose to channels_first
preprocessed_image = keras.applications.vgg16.preprocess_input(image, data_format=data_format)
print(preprocessed_image.shape)
```
### Output
```text
InvalidArgumentError: {{function_node __wrapped__AddV2_device_/job:localhost/replica:0/task:0/device:CPU:0}} Incompatible shapes: [3,224,224] vs. [1,3,1] [Op:AddV2] name: 
```

The related codes is here:

https://github.com/keras-team/keras/blob/a5e6d0a94f19ca40c20056b1d3d3fbeda72a989d/keras/src/applications/imagenet_utils.py#L279-L284

It should deal with `data_format=='channels_first'` conditionally(3D v.s. 4D) just like here:

https://github.com/keras-team/keras/blob/a5e6d0a94f19ca40c20056b1d3d3fbeda72a989d/keras/src/applications/imagenet_utils.py#L265-L270

Thanks for noting!

utsab345:
### Summary
Fixed a shape mismatch issue when `preprocess_input()` receives 3D input (single image)
with `data_format='channels_first'`. The bug occurred because the mean and std tensors
were reshaped for 4D input only.

### Changes
- Added conditional reshape for both 3D and 4D cases.
- Included shape validation for unsupported dimensions.

### Test
```python
import numpy as np, tensorflow as tf, keras
image = tf.transpose(tf.random.uniform((224, 224, 3)), perm=[2, 0, 1])
keras.applications.vgg16.preprocess_input(image, data_format='channels_first')
```
### Closes  #21749


keras/src/utils/image_utils.py
```diff
@@ -175,10 +175,13 @@ def save_img(path, x, data_format=None, file_format=None, scale=True, **kwargs):
         **kwargs: Additional keyword arguments passed to `PIL.Image.save()`.
     """
     data_format = backend.standardize_data_format(data_format)
+    # Normalize jpg → jpeg
+    if file_format is not None and file_format.lower() == "jpg":
```
fchollet, line 179: Please revert changes in this file

integration_tests/test_save_img.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+@pytest.mark.parametrize(
+    "shape, name",
+    [
+        ((50, 50, 3), "rgb.jpg"),
+        ((50, 50, 4), "rgba.jpg"),
+    ],
+)
+def test_save_jpg(tmp_path, shape, name):
```
fchollet, line 18: Please use a TestCase subclass and use `self.assert...` methods instead of naked assert staments

Issue #21634, opened 2025-09-04
Model sharding fails due to early weight loading - OOM on large models


divyashreepathihalli:
Issue: keras.distribution.ModelParallel with from_preset() causes OOM errors when loading models larger than single-device memory.
Root Cause: Model weights are downloaded and fully instantiated in memory before distribution/sharding is applied, rather than being distributed during the loading process.

Evidence
Working Example: [Gemma2 2B notebook from last week] - Memory properly distributed across 8 TPU devices
Failing Example: [Gemma2 9B notebook] - OOM during loading despite having sufficient total TPU memory
Cross-Platform Confirmation: Similar issue occurs on Kaggle runtime (different Python version and library versions), suggesting it's not environment-specific.

Error Trace Pattern
Consistent failure path across platforms:
keras_hub/src/models/task.py → keras_hub/src/utils/preset_utils.py → keras/src/saving/serialization_lib.py → keras/src/layers/init.py → keras_hub/src/models/gemma/gemma_backbone.py → keras_hub/src/layers/modeling/reversible_embedding.py → jax/_src/numpy/array_methods.py → jax/_src/interpreters/pxla.py:1362 → results = self.xla_executable.execute_sharded(input_bufs)

Additional Issues
Kaggle TPU: Consistent availability but TensorBoard profiling plugin incompatible with runtime
Colab TPU: Libtpu library crashes runtime (logs attached)
Impact: Prevents loading models >2B parameters despite having sufficient distributed memory capacity.

Through debugging, I believe I've traced this to a JAX canonicalization change that's breaking distributed model loading.

Possible Evidence of JAX Root Cause:
I traced the issue to JAX commit 19a5055 by Peter Hawkins: "Remove most canonicalization from MLIR lowering"
The commit message explicitly states:
"There is still one place where we need the older canonicalization behavior that forms a NumPy array for a Python scalar, namely in pxla.shard_args."
This matches exactly where our error occurs - in the sharding code (pxla.py) during distributed loading.

Request:
This looks like a JAX regression that's hitting Keras distributed loading. Could you help validate this isn't actually a Keras issue and maybe file a JAX bug report together?

MalyalaKarthik66:
I tried loading a big model with ModelParallel and ran into the same OOM error. Looks like the weights are loaded fully on one device before distribution. Can I work on modifying it so the model is sharded while loading to avoid this?

amitsrivastava78:
[...]

PR #21755, opened 2025-10-17
Extended fix OOM Issue #21634 on Keras side

praveenhosdrug123:
## Summary
Applies distributed initialization fix to model backbone to resolve OOM errors during initialization of 7B+ parameter models on 8GB TPU devices. This PR adds a helper function to distribute the initializers at time of instantiation.

## Issue
Token embedding initialization creates large arrays at time of creation, placing all weights on a single device.
Combined with forward passes during backbone initialization, this causes a 2X to 3X memory spike and triggers OOM on TPUs with limited HBM. 

## Solution
Implements _distribute_initializer helper that wraps embedding initializers with explicit TensorLayout, properly sharding weights across TPU devices during instantiation. Validated on 8-device TPU: models that previously OOM'd during backbone initialization now load successfully.

## Reference
For memory profiling analysis, cache locality theory, validation logs and alternative solutions considered, refer to: [Doc](https://docs.google.com/document/d/1USQj3Tn1uL7iUijbPGLAsE4NiB_BDPlRRGihAgjWj_k)

Related PR: keras-team/keras-hub#2441

Issue: keras-team/keras#21634

### DIFF
--- a/keras/src/backend/jax/core.py
+++ b/keras/src/backend/jax/core.py
@@ -53,7 +53,7 @@ class JaxVariable(KerasVariable):
         self._initialize_layout()
         layout = self._layout
         shape = self._shape
-        if check_memory_spec(layout, shape):
+        if should_shard_at_init(layout, shape):
             jitted_initializer = jax.jit(
                 initializer.__call__,
                 out_shardings=layout,
@@ -312,19 +312,16 @@ if config.is_nnx_enabled():
     NnxVariable.__setattr__ = __setattr__
 
 
-def check_memory_spec(init_layout, shape):
-    is_named_sharding = isinstance(init_layout, jax.sharding.NamedSharding)
-    # Check if PartitionSpec has any non-None values
-    spec = getattr(init_layout, "spec", None)
-    partition_spec = spec if spec is not None else ()
-    is_partitioned = any(dim is not None for dim in partition_spec)
-    is_partitioned = is_partitioned and is_named_sharding
-    # Size check: Allow sharding at init only for large arrays
-    # Sharding smaller arrays at init increases loading time
-    size_threshold = 250 * 1024 * 1024  # 250MB
-    array_size = np.prod(shape) * 4  # Assume 4 bytes per element
-    size_check = True if array_size >= size_threshold else False
-    return is_partitioned and size_check
+def should_shard_at_init(init_layout, shape):
+    if not isinstance(init_layout, jax.sharding.NamedSharding):
+        return False
+
+    if all(dim is None for dim in init_layout.spec):
+        return False
+
+    size_threshold = 250 * 1024 * 1024
+    array_size = np.prod(shape) * 4
+    return array_size >= size_threshold
 
 
 def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):


keras/src/backend/jax/distribution_lib.py
```diff
@@ -246,3 +246,80 @@ def _to_backend_layout(tensor_layout):
     partition_spec = jax.sharding.PartitionSpec(*tensor_layout.axes)
     jax_mesh = tensor_layout.device_mesh.backend_mesh
     return jax.sharding.NamedSharding(jax_mesh, partition_spec)
+
+
+def _distribute_initializer(
+    init_func=None, mean=0.0, stddev=1.0, seed=None, layout=None
+):
```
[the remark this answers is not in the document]
praveenhosdrug123: Feedback for gemini-code-assist -- This has been mentioned in the doc string of the function. From the current PR--init_func: A functools.partial-wrapped object that takes the seed
            as argument and returns a jax.Array. Must have shape and dtype
            already bound via partial.

keras/src/backend/jax/distribution_lib.py
```diff
@@ -246,3 +246,80 @@ def _to_backend_layout(tensor_layout):
...
+        ValueError: If init_func or seed is None.
+                    If init_func.func is not a supported random function.
+        TypeError: If init_func is not a functools.partial object.
+    """
+    import warnings
+    from functools import partial
+
+    # Validate all required arguments
+    if seed is None:
+        raise ValueError("seed cannot be None. Use keras.random.SeedGenerator.")
```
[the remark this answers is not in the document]
praveenhosdrug123: Feedback for gemini-code-assist -- The ValueError message already mentions 
using `keras.random.SeedGenerator`. See line 282 in the code.

For human reviewers -- This PR maintains the existing seed validation logic 
without changes. The distributed initialization path uses the same seed handling 
as the standard path. The scope of this PR is fixing the OOM issue during build, 
not refactoring seed management.

keras/src/backend/jax/distribution_lib.py
```diff
@@ -246,3 +246,80 @@ def _to_backend_layout(tensor_layout):
...
+    if not isinstance(init_func, partial):
+        raise TypeError(
+            f"init_func must be functools.partial object, got {type(init_func)}"
+        )
+
+    # Shard based on tensor layout
+    if layout is None:
+        warnings.warn(
+            f"The layout is {layout}, sharding will default to single device"
+        )
```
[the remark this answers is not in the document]
praveenhosdrug123: For example, it could state that this might lead to OOM errors if the tensor is too large to fit on a single device. --   It is not a general error, it depends on the size of the token embedding tensor relative to the available HBM or host memory. Warning at each instance would lead to redundant  and faulty warnings. It is better to let the user let know that the initialization would first default to single device at this stage.

keras/src/backend/jax/distribution_lib.py
```diff
@@ -246,3 +246,80 @@ def _to_backend_layout(tensor_layout):
...
+
+    # Apply mean/stddev only for distributions where it makes sense
+    if init_func.func in (jax.random.normal, jax.random.truncated_normal):
+        return sample * stddev + mean
+    elif init_func.func == jax.random.uniform:
+        # Uniform doesn't use mean/stddev - warn
+        if mean != 0.0 or stddev != 1.0:
+            warnings.warn(
+                "mean and stddev are ignored for uniform distribution"
+            )
```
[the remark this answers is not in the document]
praveenhosdrug123: A ValueError check on line 323 which checks for the supported initializers and  provides a list of supported initializers to the user addresses this comment.

keras/src/backend/jax/distribution_lib.py
```diff
@@ -246,3 +246,80 @@ def _to_backend_layout(tensor_layout):
...
+        return sample
+    else:
+        raise ValueError(
+            f"Unsupported initializer: {init_func.func.__name__}. "
+            f"Supported: normal, truncated_normal, uniform"
+        )
```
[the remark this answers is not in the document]
praveenhosdrug123: A ValueError check on line 323 which checks for the supported initializers and  provides a list of supported initializers to the user addresses this comment. Currently the fix aims to only fix the OOM issue pertaining to model presets's token embedding initializers. The distributions supported form the basis of all the initializers covered in the dist_initializer file under keras hub utilities. It would be better to understand the user needs to have distributions which are not covered by the expansive list of distributions covered.

keras/src/backend/jax/distribution_lib.py
```diff
@@ -246,3 +246,80 @@ def _to_backend_layout(tensor_layout):
...
+            f"The layout is {layout}, sharding will default to single device"
+        )
+        sharding = None
+    else:
+        sharding = _to_backend_layout(layout)
+
+    # The init_func has static arguments baked in as per initializer.
+    compiled_init = jax.jit(
+        lambda seed: init_func(seed), out_shardings=sharding
+    )
```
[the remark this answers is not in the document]
praveenhosdrug123: TODO: praveenhosdrug123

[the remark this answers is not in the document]
praveenhosdrug123: I can add the inherent function expected in the init_func as a comment above this line. We are expecting specific jax based random functions here. This would help in user readability.
TODO: praveenhosdrug123 to validate the seed process if it translates to the distributed framework or if the existing process were incorrect.

keras/src/backend/jax/distribution_lib.py
```diff
@@ -246,3 +246,80 @@ def _to_backend_layout(tensor_layout):
...
+        seed: Random seed for initialization.
+        layout: TensorLayout for the distributed tensor.
+
+    Returns:
+        A distributed jax array.
+
+    Raises:
+        ValueError: If init_func or seed is None.
+                    If init_func.func is not a supported random function.
+        TypeError: If init_func is not a functools.partial object.
```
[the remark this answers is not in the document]
praveenhosdrug123: Feedback for gemini-code-assist: This is part of the docstrings and not the actual ValueError mentioned in the feedback.
This has been raised in another comment.

keras/src/backend/jax/distribution_lib.py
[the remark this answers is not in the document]
praveenhosdrug123: The seed is correctly handled within the JIT context because:

1. We convert the seed to JAX PRNG key format
jax_compatible_seed = jax.numpy.array(
    [jax_seed[1], jax_seed[0]], dtype=jax.numpy.uint32
)

2. The init_func is a JAX random function (e.g., jax.random.truncated_normal)
   which properly consumes the PRNG key
init_func = partial(
    truncated_normal,  # This is jax.random.truncated_normal
    shape=shape,
    dtype=dtype,
    lower=-2.0,
    upper=2.0,
)

3. Inside the JIT-compiled function, the key is passed directly to 
    the JAX random function, which handles it correctly:
compiled_init = jax.jit(
    lambda key: init_func(key),  # key is jax_comaptible_seed key passed to jax.random.*
    out_shardings=sharding
)

 4. JAX's random functions are stateless and functional - they take a key
   and deterministically produce random values, which works correctly
    within JIT compilation
sample = compiled_init(jax_compatible_seed)

 5. Per-shard randomness is automatically handled by JAX when the output
    is sharded - each shard gets different but deterministic values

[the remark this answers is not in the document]
praveenhosdrug123: Done!

hertschuh:
@praveenhosdrug123

Thank you for the investigation. This is indeed an issue.

Somebody on the team is working on a fix that's generally applicable to all variables so that you don't have to explicitly use the fix that you provided here.

praveenhosdrug123:
@hertschuh - Thanks for the feedback and for taking the time to review the document.

I want to clarify the technical issue:
The OOM problem is about large contiguous memory allocation, not total parameter count. Token embeddings are the largest single array and exceed device memory during initialization, even when the full model would fit after sharding.

Thank you for the context on the general solution. A few follow-up questions to help me understand the timeline:

1. What's the expected completion date for the general fix?
2. Will it handle the edge cases mentioned in the document (interleaving, quantization, LoRA)?
3. Will it detect which variables actually need distribution?

The reason I ask: users are blocked on this today for 7B+ models on 8GB TPU devices. 
If the general fix is months out, would it make sense to:
- Merge this targeted fix as a stopgap
- Mark it deprecated once the general solution ships
- Remove it in a future release

Let me know if that's feasible.

hertschuh:
@praveenhosdrug123 

> Merge this targeted fix as a stopgap

I'm confused about how this particular fix addresses the issue though. The `_distribute_initializer` function is not used anywhere.

praveenhosdrug123:
@hertschuh - Thank you :) 
The _distribute_initializer function is used in keras_hub/src/utils/dist_initializer.py as part of the following pull request https://github.com/keras-team/keras-hub/pull/2441 .

praveenhosdrug123:
@hertschuh  and @amitsrivastava78 -- 
The PR is ready for review. Since the implementation revises the head and tail of the flow, there are some test failures in the distribution library and random library tests in CI. These failures occur because the implementation changes the expected flow for these tests. The same tests pass locally with the updated implementation.

The other backends are unaffected by the change.

I would thoroughly appreciate your review and guidance on the PR now.

keras/src/backend/common/variables.py
```diff
@@ -414,10 +414,13 @@ def _initialize(self, value):
...
-            initializer(self._shape, dtype=self._dtype)
-        )
-        self._initialize(value)
+        if backend.backend() == "jax":
+            raise NotImplementedError
+        else:
+            value = self._convert_to_tensor(
+                initializer(self._shape, dtype=self._dtype)
+            )
+            self._initialize(value)
```
hertschuh: Please undo this file. It's generally understood that methods can have overrides, and that if some overrides are missing bad things happens. But we don't throw when a known override is missing.

Moreover, we want to support `JaxVariable._initialize_with_initializer` calling `super()._initialize_with_initializer`.

keras/src/backend/jax/core.py
```diff
@@ -30,9 +30,7 @@ def __init__(self, *args, layout=None, **kwargs):
         self._layout = layout
         super().__init__(*args, **kwargs)
 
-    def _initialize(self, value):
-        # Note that variable.shape is needed by distribution_lib
-        self._shape = self._validate_shape(value.shape)
+    def set_tensor_layout(self):
```
hertschuh: Rename `set_tensor_layout` to `_initialize_layout`.

- it's private
- it's not really a setter, there is no `value` argument

keras/src/backend/jax/core.py
```diff
@@ -44,6 +42,29 @@ def _initialize(self, value):
...
+    def _initialize_with_initializer(self, initializer):
+        self.set_tensor_layout()
+        layout = self._layout
+        if check_layout_spec_nnx(layout):
+            jitted_initializer = jax.jit(
+                initializer.__call__,
+                out_shardings=layout,
+                static_argnames=["shape", "dtype"],
+            )
+            value = jax.device_put(
```
hertschuh: Do you need the `device_put`? Doesn't `out_shardings` put it on the device automatically?

keras/src/backend/jax/core.py
```diff
@@ -112,6 +133,12 @@ def __init__(
             # The real value is now set in self._value, sync it to raw_value
             object.__setattr__(self, "raw_value", self._value)
 
+        def _initialize_with_initializer(self, initializer):
+            value = self._convert_to_tensor(
+                initializer(self._shape, dtype=self._dtype)
+            )
+            self._initialize(value)
```
hertschuh, line 137: What's the issue with NNX? Why can't we do the sharding?

keras/src/backend/jax/core.py
```diff
@@ -288,6 +315,16 @@ def __setattr__(self, name, value):
...
 
 
+def check_layout_spec_nnx(init_layout):
+    nnx_enabled = config.is_nnx_enabled()
+    is_named_sharding = isinstance(init_layout, jax.sharding.NamedSharding)
+    # Check if PartitionSpec has any non-None values
+    spec = getattr(init_layout, "spec", None)
+    partition_spec = spec if spec is not None else ()
+    is_partitioned = any(dim is not None for dim in partition_spec)
+    return is_partitioned and is_named_sharding and not nnx_enabled
```
hertschuh: Can you explain this? What is the issue with NNX?

Also, `check_layout_spec_nnx` is only called in  `JaxVariable._initialize_with_initializer`. However, `NnxVariable` overrides `_initialize_with_initializer` and doesn't call super.

So my reading of this is that it's not actually used.

hertschuh (commented):
Thanks a lot for the update!

Were you able to test this on a large model? Also, did you notice a slowdown in the initialization?

keras/src/backend/common/variables.py
praveenhosdrug123: Done!

keras/src/backend/jax/core.py
praveenhosdrug123: Done!

keras/src/backend/jax/core.py
praveenhosdrug123: We dont need device_put , out_shardings does put it on the device automatically .

 when i was going through the jax array for
```python
value = jitted_initializer(self._shape, dtype=self._dtype)
```
I found the _committed flag to be False which indicated a problem. 
But i am trying again to recreate the problem and do not find the same problem.
The committed flag along with other flags from the jax array seem to match for both 
```python
value = jitted_initializer(self._shape, dtype=self._dtype)
```
and
```python
value = jax.device_put(jitted_initializer(self._shape, dtype=self._dtype),layout)
```
The changes in the _committed flag could be transient but made it feel the out_shardings parameter is relatively newer in jit compilation. I felt device_put is a bit more stable function for expected jax array creation with sharding.
We could remove it, but the device_put is a no-op and does ensure the expected jax array has been created.
So i went with the device put and outshardings together. Outshardings fixes the OOM and device put ensures the JAX array creation behavior.

Happy to hear your thoughts

keras/src/backend/jax/core.py
praveenhosdrug123: NNX has its own nnx.jit method from the flax module. Since both jax and nnx have their own jit methods,
I was not entirely sure if the jit from jax could impact nnx in any way. In order to isolate the jit from jax and nnx, i overrode the NnxVariable's  with _initialize_with_initializer function . 

I understand the nnx object is not being jitted and it would be the call of the initializer object but I am not very sure of the interaction(Jit : Jax vs flax).

Happy to hear your thoughts and whether we can still use the _initialize_with_initializer from the JAX variable.

keras/src/backend/jax/core.py
praveenhosdrug123: Yes, you're right the NnxVariable overrides the _initialize_with_initializer and doesn't call super.
I have removed the nnx check as it was dead code and causes confusion.

praveenhosdrug123:
@hertschuh 
Thank you for your feedback and review!

These were the results of the changes. There was a bit of a slowdown noticed while loading.


## Model Load Time Comparison

| Model | No Change (Current Approach) | All Initializers | Only Token Embedding |
|-------|------------------------------|--------------------------|------------------------------|
| gemma_7b_en |  Fails | 182.49 seconds | 29.64 seconds |
| gemma2_2b_en | 23.56 seconds | 59.85 seconds | 15.54 seconds |
| gemma3_instruct_270m | 14.92 seconds | 27.05 seconds | 8.86 seconds |

* Measurements were taken after deleting the kagglehub cache after each run.

It is a bit surprising that the scenario where only the token embedding path being sharded at creation had the fastest loading time. When we shard all variables at onset, the loading time is double the no change loading time.

If the latency is still under the threshold for model loading time then we can shard all the variables at initialization. Happy to hear your thoughts on this.

For now i have sharded the token embedding initializer at the onset.

keras/src/backend/jax/core.py
hertschuh: Let's remove the `device_put`. I really don't think it's needed and it wasn't there before either.

`jax.jit` `out_shardings` was introduced in JAX 0.4.13, so we're totally fine there.

keras/src/backend/jax/core.py
```diff
@@ -288,6 +314,18 @@ def __setattr__(self, name, value):
...
 
+def check_token_spec(init_layout, path):
+    is_named_sharding = isinstance(init_layout, jax.sharding.NamedSharding)
+    # Check if PartitionSpec has any non-None values
+    spec = getattr(init_layout, "spec", None)
+    partition_spec = spec if spec is not None else ()
+    is_partitioned = any(dim is not None for dim in partition_spec)
+    is_partitioned = is_partitioned and is_named_sharding
+    # Check if path is for token embedding
+    token_embed_path = True if path == "token_embedding/embeddings" else False
```
hertschuh: This doesn't feel right in terms of the layering of frameworks. This is using an implementation detail of the KerasHub models. But Keras should be generic.

hertschuh:
> @hertschuh Thank you for your feedback and review!
> 
> These were the results of the changes. There was a bit of a slowdown noticed while loading.
> 
> ## Model Load Time Comparison
> Model	No Change (Current Approach)	All Initializers	Only Token Embedding
> gemma_7b_en	Fails	182.49 seconds	29.64 seconds
> gemma2_2b_en	23.56 seconds	59.85 seconds	15.54 seconds
> gemma3_instruct_270m	14.92 seconds	27.05 seconds	8.86 seconds
> * Measurements were taken after deleting the kagglehub cache after each run.
> 
> It is a bit surprising that the scenario where only the token embedding path being sharded at creation had the fastest loading time. When we shard all variables at onset, the loading time is double the no change loading time.
> 
> If the latency is still under the threshold for model loading time then we can shard all the variables at initialization. Happy to hear your thoughts on this.
> 
> For now i have sharded the token embedding initializer at the onset.

I put a comment in the code regarding the reliance on the name `"token_embedding/embeddings"`.

It's unfortunate that it's such a slowdown because it's actually more correct. My assumption is that the reason is mostly the compile time of the jitted functions.

Now, the next step (for a different PR) is to add `out_shardings` or rather `layout` to all initializers and random functions. So this solution is not final, but it will be the fallback for instance for custom initializers that don't support the `layout` parameter.

Here is one idea though. What if the criteria, instead of being the `"token_embedding/embeddings"` was a minimum size? I suspect there are a lot of very small variables that make up most of the slowdown.

Thank you for all your work on this!

keras/src/backend/jax/core.py
praveenhosdrug123: That's Fair. Im exploring your idea of using size. There were many smaller variables with memory less than 10MB . So i was thinking we can gather a theoretical memory using the math.prod(shape) * 4 bytes , then compare it against a threshold. If its higher than the threshold we initialize it at onset.

keras/src/backend/jax/core.py
praveenhosdrug123: Done!

praveenhosdrug123:
My assumption is that the reason is mostly the compile time of the jitted functions. -- Makes a lot of sense! Thanks!

On the positives, sharding only large arrays at initialization helps reduce the memory pressure and invariably decreases loading time for models now. 
Please note :- There definitely exists a better and sophisticated "minimum size" based on the proportion of variable weights and Jit compilation times across different model configurations. But focussing on just large arrays will work well for this specific use case.

At this point i wanted to check in with you regarding this particular pull request -- https://github.com/keras-team/keras/pull/21853
Im interested in this particular issue as it relates to something i have been working on. 
Could that be the next step after this PR?

Thank you for all your help on this!

hertschuh:
> At this point i wanted to check in with you regarding this particular pull request -- #21853 Im interested in this particular issue as it relates to something i have been working on. Could that be the next step after this PR?

This was addressed here: https://github.com/keras-team/keras/pull/21891

Does that work for you?

keras/src/backend/jax/core.py
```diff
@@ -288,6 +312,21 @@ def __setattr__(self, name, value):
     NnxVariable.__setattr__ = __setattr__
 
 
+def check_memory_spec(init_layout, shape):
```
hertschuh: Rename `_should_shard_initializer` (it's private, the name says what it's used for).

keras/src/backend/jax/core.py
```diff
@@ -288,6 +312,21 @@ def __setattr__(self, name, value):
...
+    spec = getattr(init_layout, "spec", None)
+    partition_spec = spec if spec is not None else ()
+    is_partitioned = any(dim is not None for dim in partition_spec)
+    is_partitioned = is_partitioned and is_named_sharding
+    # Size check: Allow sharding at init only for large arrays
+    # Sharding smaller arrays at init increases loading time
+    size_threshold = 250 * 1024 * 1024  # 250MB
+    array_size = np.prod(shape) * 4  # Assume 4 bytes per element
+    size_check = True if array_size >= size_threshold else False
+    return is_partitioned and size_check
```
hertschuh: This would be a bit easier and more readable with early exits:
```python
if not isinstance(init_layout, jax.sharding.NamedSharding)
    return False

if all(dim is None for dim in init_layout.spec):
    return False

size_threshold = 250 * 1024 * 1024  # 250MB
array_size = np.prod(shape) * 4
return array_size >= size_threshold
```

hertschuh (commented):
Last nitpicks.

Thanks a lot!

keras/src/backend/jax/core.py
praveenhosdrug123: Done !

keras/src/backend/jax/core.py
praveenhosdrug123: Nit -- I was thinking could we rename it as a public function named should_shard_at_init(init_layout, shape), The name _should_shard_initializer could be misleading since variables might be sharded later, just not at initialization. As a helper function, it follows the pattern of other public helpers in the codebase.

Just a thought

praveenhosdrug123:
> > At this point i wanted to check in with you regarding this particular pull request -- #21853 Im interested in this particular issue as it relates to something i have been working on. Could that be the next step after this PR?
> 
> This was addressed here: #21891
> 
> Does that work for you?

Yes, that works - thank you!

I have a related optimization idea I'd like to experiment with. I'll follow up on PR #21891 once I have results to share. Would take time probably in the new year.

Thanks again!

hertschuh (approved):

praveenhosdrug123:
@hertschuh 

The Tensorflow GPU test under the keras/src/trainers/data_adapters/generator_data_adapter_test.py fails for the test_basic_flow function when the use_sample_weight = False, generator_type = 'tf' for the GPU device.

Note:-- It looks like multi-GPU  tests which automatically shards the array from the error 2025-12-16 01:32:46.979162: W tensorflow/core/framework/op_kernel.cc:1855] OP_REQUIRES failed at strided_slice_op.cc:117 : INVALID_ARGUMENT: slice index 32 of dimension 0 out of bounds.

This is unrelated to the current core JAX change, could you please have a quick look.

Thank you!

fchollet:
The test appears to be merely flaky.

praveenhosdrug123:
Thank you both!

Issue #21624, opened 2025-08-29
Add feature to extract_patches for volume data


innat:
In tensorflow, we can do

```python

a = tf.ones((1, 96, 96, 96, 4))   # add batch dim → (B, D, H, W, C)
p = 4
patches = tf.extract_volume_patches(
    input=a, 
    ksizes=[1, p, p, p, 1],
    strides=[1, p, p, p, 1],
    padding='VALID'
)

print(patches.shape)
(1, 24, 24, 24, 256)
```

This feature is missing in `keras.ops.image.extract_patches`. 

---

Workaround if anyone is looking for

```python
def extract_volume_patches_simple(x, patch_size):
    batch_size, depth, height, width, channels = ops.shape(x)
    
    # Calculate patch counts
    d_patches = depth // patch_size[0]
    h_patches = height // patch_size[1]
    w_patches = width // patch_size[2]
    
    # Reshape directly to the target shape
    patches = ops.reshape(
        x,
        [
            batch_size,
            d_patches, patch_size[0],
            h_patches, patch_size[1],
            w_patches, patch_size[2],
            channels
        ]
    )
    
    # Now reshape to final form
    patches = ops.reshape(
        patches,
        [
            batch_size,
            d_patches,
            h_patches,
            w_patches,
            patch_size[0] * patch_size[1] * patch_size[2] * channels
        ]
    )
    
    return patches

# Test
a = ops.ones((1, 96, 96, 96, 4))
p = 4
patches = extract_volume_patches_simple(a, (p, p, p))
print(patches.shape) 
(1, 24, 24, 24, 256)
```

SamanehSaadat:
Thanks for providing the workaround for the case of non-overlapping patches.
I labeled this issue as "contributions welcome"  for anyone who is interested in extending `keras.ops.image.extract_patches` to work for 3D images in a flexible way.

danielenricocahall:
@innat This is feature is now available in [3.12](https://github.com/keras-team/keras/compare/v3.11.0...v3.12.0) as `extract_patches_3d`

innat:
Nice work! Just a small suggestion, since this is under ops, would it be possible to unify the API for both 2D and 3D cases? For example, similar to how keras.ops.conv supports both 2D and 3D operations.

danielenricocahall:
[...]

PR #21759, opened 2025-10-20
Support for extracting volume patches

danielenricocahall:
Addresses https://github.com/keras-team/keras/issues/21624. Inspired by https://github.com/keras-team/keras/pull/21727, except made to be compatible with all backends

**Notes**
- Should we try to coerce the `ExtractPatches` and `ExtractVolumePatches` `Operation`s into a common base/mixin or is would that obfuscate?
- Should `ExtractPatches` have the same validation added to it's `__init__` as was suggested (and added) by Gemini in `ExtractVolumePatches`?

### DIFF
--- a/keras/src/ops/image.py
+++ b/keras/src/ops/image.py
@@ -712,7 +712,7 @@ def _extract_patches(
     return patches
 
 
-class ExtractVolumePatches(Operation):
+class ExtractPatches3D(Operation):
     def __init__(
         self,
         size,
@@ -880,7 +880,7 @@ def extract_patches_3d(
     (3, 3, 3, 81)
     """
     if any_symbolic_tensors((volumes,)):
-        return ExtractVolumePatches(
+        return ExtractPatches3D(
             size=size,
             strides=strides,
             dilation_rate=dilation_rate,
--- a/keras/src/ops/image_test.py
+++ b/keras/src/ops/image_test.py
@@ -2560,7 +2560,7 @@ class ImageOpsBehaviorTests(testing.TestCase):
             kimage.elastic_transform(invalid_image)
 
 
-class ExtractVolumePatchesTest(testing.TestCase):
+class ExtractPatches3DTest(testing.TestCase):
     FLOAT_DTYPES = [x for x in dtypes.FLOAT_TYPES if x not in ("float64",)]
 
     def setUp(self):


fchollet:
Thanks for the PR!

> Should we try to coerce the ExtractPatches and ExtractVolumePatches Operations into a common base/mixin or is would that obfuscate?

Keeping them separate sounds fine to me.

keras/src/ops/image.py
```diff
@@ -712,6 +712,187 @@ def _extract_patches(
...
+        data_format=data_format,
+        dilation_rate=dilation_rate,
+    )
+    if _unbatched:
+        patches = backend.numpy.squeeze(patches, axis=0)
+    return patches
+
+
+@keras_export("keras.ops.image.extract_volume_patches")
+def extract_volume_patches(
```
fchollet: For API consistency with our conv ops, we might want to name it `extract_patches_3d` or `extract_3d_patches`. "volume" could be any dimension beyond 2.

keras/src/ops/image.py
```diff
@@ -712,6 +712,187 @@ def _extract_patches(
...
+            `dilation_rate > 1` is not supported in conjunction with
+            `strides > 1` on the TensorFlow backend.
+        padding: The type of padding algorithm to use: `"same"` or `"valid"`.
+        data_format: A string specifying the data format of the input tensor.
+            It can be either `"channels_last"` or `"channels_first"`.
+            `"channels_last"` corresponds to inputs with shape
+            `(batch, depth, height, width, channels)`, while `"channels_first"`
+            corresponds to inputs with shape
+            `(batch, channels, depth, height, width)`. If not specified,
+             the value will default to `keras.config.image_data_format`.
```
fchollet: It's a function, `keras.config.image_data_format()`

keras/src/ops/image.py
danielenricocahall: Okay fixed! Went with `extract_patches_3d` to be closer to existing API

keras/src/ops/image.py
danielenricocahall: Ah yes sorry, fixed

fchollet (approved):
Nice, thanks for the updates! LGTM

Issue #21700, opened 2025-09-29
Nested outputs structures and corresponding nested losses are supported, but not nested metrics


hertschuh:
This is a follow up to https://github.com/keras-team/keras/issues/21675 and https://github.com/keras-team/keras/pull/21694

- nested output structures are supported
- nested losses matching the nested output structures are supported
- however, the same is not true for metrics, the code within the `try`/`except` block fails

```python
import numpy as np
import keras

print('keras version:', keras.__version__)

X = np.random.rand(100, 32)
Y1 = np.random.rand(100, 10)
Y2 = np.random.rand(100, 10)
Y3 = np.random.rand(100, 10)

def create_model():
    x = keras.Input(shape=(32,))
    y1 = keras.layers.Dense(10)(x)
    y2 = keras.layers.Dense(10)(x)
    y3 = keras.layers.Dense(10)(x)
    return keras.Model(inputs = x, outputs = {'a': y1, 'b': {'c': y2, 'd': y3}})

model = create_model()
model.compile(optimizer = 'adam', loss = {'a': 'bce', 'b': {'c': 'mse', 'd': 'mse'}})
model.train_on_batch(X, {'a': Y1, 'b': {'c': Y2, 'd': Y3}})
print('Successful training not using metrics\n')

try:
    model = create_model()
    model.compile(
        optimizer = 'adam',
        loss = {'a': 'bce', 'b': {'c': 'mse', 'd': 'mse'}},
        metrics = {'a': ['mae', 'acc'], 'b':{'c': 'mse', 'd': 'mse'}}
    )
    model.train_on_batch(X, {'a': Y1, 'b': {'c': Y2, 'd': Y3}})
    print('Successful training using metrics')
except ValueError as ex:
    print('Unsuccessful training with metrics')
    print(ex)

model = create_model()
model.compile(
    optimizer = 'adam',
    loss = {'a': 'bce', 'b': {'c': 'mse', 'd': 'mse'}},
    loss_weights = {'a': 1., 'b': {'c': 0.5, 'd': 0.4}}
)
model.train_on_batch(X, {'a': Y1, 'b': {'c': Y2, 'd': Y3}})
print('Successful training using loss_weights\n')
```

Ideally, metrics and losses should be handled the same. Part of the complication is that there is a mechanism to use the output names when not using the output structure.

[...]

PR #21761, opened 2025-10-20
Fix Nested Metrics Handling in CompileMetrics

Dimios45:
This PR addresses the issue reported in [Keras Issue #21700](https://github.com/keras-team/keras/issues/21700), where nested output structures and corresponding losses are supported, but nested metrics are not. This caused failures when attempting to compile models with nested metrics configurations.

### Changes
- [x] Implemented recursive handling of nested metric configurations within `CompileMetrics`.
- [x] Ensured compatibility with both nested and flat metric structures.
- [x] Updated internal methods to process nested metrics appropriately.

### Testing
- [x] Added tests to verify correct handling of nested metrics during model compilation and training.
- [x] Tests cover nested dictionaries and lists to ensure the recursive metric construction behaves as expected and raises appropriate errors for invalid configurations.

### Example Usage:

```python
model.compile(
    optimizer='adam',
    loss={'a': 'mse', 'b': {'c': 'mse', 'd': 'mse'}},
    metrics={'a': ['mae'], 'b': {'c': 'mse', 'd': 'mae'}}

### DIFF
--- a/keras/src/trainers/compile_utils.py
+++ b/keras/src/trainers/compile_utils.py
@@ -175,6 +175,40 @@ class CompileMetrics(metrics_module.Metric):
         return vars
 
     def build(self, y_true, y_pred):
+        # Handle nested structures using tree utilities similar to CompileLoss
+        if tree.is_nested(y_true) and tree.is_nested(y_pred):
+            try:
+                # Align the metrics configuration with the structure of y_pred
+                if self._has_nested_structure(self._user_metrics) or self._has_nested_structure(self._user_weighted_metrics):
+                    self._flat_metrics = self._build_nested_metrics(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: keras/src/trainers/compile_utils.py [hunk 1]
```diff
@@ -175,6 +175,40 @@ def variables(self):
...
+            return False
+        if isinstance(obj, dict):
+            for value in obj.values():
+                if isinstance(value, (dict, list)):
+                    return True
+        elif isinstance(obj, list):
+            for item in obj:
+                if isinstance(item, (dict, list)):
+                    return True
+        return False
```
hertschuh, line 209: Why not use `tree.is_nested`?

PR #21762, opened 2025-10-22
Added OrbaxCheckpoint for keras 3.0 for Data centric saving and restore

amitsrivastava78:
Following features are supported
1. Asynchronous Saving
2. Full State Preservation
3. Save Frequency Control
4. Checkpoint Management
5. Monitoring & Best Model Tracking
6. Cross-Backend Compatibility

### DIFF
--- a/keras/src/callbacks/orbax_checkpoint.py
+++ b/keras/src/callbacks/orbax_checkpoint.py
@@ -114,10 +114,10 @@ class OrbaxCheckpoint(MonitorCallback):
         save_best_only=False,
         save_weights_only=False,
         mode="auto",
-        max_to_keep=1,
-        save_on_background=True,
         save_freq="epoch",
         initial_value_threshold=None,
+        max_to_keep=1,
+        save_on_background=True,
     ):
         # Ensure orbax is available
         ocp.initialize()


keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,525 @@
...
+    def _should_save_on_batch(self, batch):
+        """Check if we should save on this batch."""
+        if self.save_freq == "epoch":
+            return False
+
+        self._batches_seen_since_last_saving += 1
+        if self._batches_seen_since_last_saving >= self.save_freq:
+            self._batches_seen_since_last_saving = 0
+            return True
+        return False
```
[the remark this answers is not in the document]
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,525 @@
...
+                    # Update best value when there's improvement
+                    self.best = current
+
+            if should_save:
+                # Use step number (e.g., optimizer iterations) for Orbax save
+                # step
+                step = self._get_current_step()
+                self._save_checkpoint(step=step, logs=logs)
+                # Ensure all processes sync after save operation
+                self.manager.wait_until_finished()
```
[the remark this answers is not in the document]
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,525 @@
...
+                should_save = False
+            else:
+                # Update best value when there's improvement
+                self.best = current
+
+        if should_save:
+            # Use epoch number as the step for Orbax save
+            self._save_checkpoint(step=epoch, logs=logs)
+            # Ensure all processes sync after save operation
+            self.manager.wait_until_finished()
```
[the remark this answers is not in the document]
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,525 @@
...
+        if hasattr(self.model, "optimizer") and hasattr(
+            self.model.optimizer, "iterations"
+        ):
+            # Convert potential backend tensor to int
+            return int(
+                backend.convert_to_numpy(self.model.optimizer.iterations)
+            )
+        else:
+            # Fallback: use batch count
+            return self._last_batch_seen
```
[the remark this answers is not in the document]
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,525 @@
...
+        save_best_only: if `save_best_only=True`, it only saves when the model
+            is considered the "best" based on the monitored quantity.
+        mode: one of {'auto', 'min', 'max'}. Used with `save_best_only`.
+        save_freq: `'epoch'` or integer. Frequency to save checkpoints.
+        max_to_keep: Integer, maximum number of recent checkpoints to keep.
+            If None, keeps all. Defaults to 5.
+        keep_period: Integer, keep one checkpoint every `keep_period` saves.
+            Useful for keeping checkpoints less frequently over long runs.
+        initial_value_threshold: Floating point initial "best" value for the
+             monitor, used with `save_best_only`.
```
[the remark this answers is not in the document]
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,525 @@
...
+            checkpointing operations complete. Defaults to None.
+        save_transforms: Dict of orbax.checkpoint.Transform objects to apply
+            during saving. Keys should match composite_state keys (e.g.,
+            'model_weights', 'optimizer_state'). Defaults to None.
+        save_decision_policy: orbax.checkpoint.SaveDecisionPolicy object to
+            control when checkpoints are saved. If provided, overrides the
+            default save frequency logic. Defaults to None.
+        save_interval: Integer, save checkpoints every N steps. If provided,
+            overrides save_freq. Defaults to None.
+    """
```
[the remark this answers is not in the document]
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,525 @@
...
+            raise ValueError("Unrecognized save_freq")
+
+        # Create should_save_fn from save_decision_policy or save_interval
+        # if provided
+        should_save_fn = None
+        if save_decision_policy is not None:
+            # For now, create a simple should_save_fn that saves every 2 steps
+            # This is a placeholder - proper integration would require
+            # PolicyCheckpointInfo
+            should_save_fn = lambda step, prev_step=None: step % 2 == 0
```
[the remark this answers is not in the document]
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py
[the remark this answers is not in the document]
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -0,0 +1,1660 @@
...
+                None,  # learning rate (no change)
+                None,  # momentum vars (no change)
+                None,  # momentum vars (no change)
+                None,  # momentum vars (no change)
+                None,  # momentum vars (no change)
+                None,  # momentum vars (no change)
+                None,  # momentum vars (no change)
+                None,  # momentum vars (no change)
+                None,  # momentum vars (no change)
+            ],
```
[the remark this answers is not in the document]
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -0,0 +1,1660 @@
...
+
+            model_weights_np = [
+                keras.ops.convert_to_numpy(w) for w in model.weights
+            ]
+            optimizer_vars_np = [
+                keras.ops.convert_to_numpy(v) for v in model.optimizer.variables
+            ]
+            return model_weights_np, optimizer_vars_np
+        except Exception:
+            return None, None
```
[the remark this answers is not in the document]
amitsrivastava78: ok

keras/src/backend/__init__.py
```diff
@@ -75,3 +75,38 @@ class name_scope(backend_name_scope):
...
+            return 0
+    elif backend_name == "torch":
+        try:
+            import torch.distributed as dist
+
+            if dist.is_available() and dist.is_initialized():
+                return dist.get_rank()
+            return 0
+        except (ImportError, AttributeError):
+            return 0
```
hertschuh: We avoid doing if/else for backends like that.

This function would be a good addition in `distribution_lib` with a separate per backend implementation. We don't need to export it for now.

keras/src/callbacks/__init__.py
```diff
@@ -8,6 +8,12 @@
 from keras.src.callbacks.learning_rate_scheduler import LearningRateScheduler
 from keras.src.callbacks.model_checkpoint import ModelCheckpoint
 from keras.src.callbacks.monitor_callback import MonitorCallback
+
+try:
+    from keras.src.callbacks.orbax_checkpoint import OrbaxCheckpoint
+except ImportError:
```
hertschuh: Let's do this differently. The issue with this approach is that the error message you'll get when you try to use `OrbaxCheckpoint` won't make sense and it will not be clear that you need to install Orbax.

I think we should use a `LazyModule` for this. See `keras/src/utils/module_utils.py`.

But at the end of the day, what you should have here is:
```python
keras.src.callbacks.orbax_checkpoint import OrbaxCheckpoint
```
And it should only fail once you actually instantiate `OrbaxCheckpoint`.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,579 @@
+import os
+import warnings
+
+import keras  # Import Keras itself
```
hertschuh: We don't import keras within `keras/src` because it eventually leads to circular imports.

Simply import the parts you need from `keras.src`.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,579 @@
...
+from keras.src import backend
+from keras.src.api_export import keras_export
+from keras.src.callbacks.monitor_callback import (
+    MonitorCallback,  # For metric monitoring logic
+)
+
+try:
+    import orbax.checkpoint as ocp
+except ImportError:
+    ocp = None
```
hertschuh: Let's do the `LazyModule` here.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,579 @@
...
+
+# Expose advanced Orbax functionality for users who need direct access
+# These are provided as bridge for advanced usecases like custom type handlers
```
hertschuh: Who are the users? We don't want external users to use this, they should use anything within `keras.src`.

I saw you used this in unit tests, but I have a separate comment about that.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,579 @@
...
+    # Metadata functionality
+    metadata = ocp.metadata
+else:
+    CheckpointManager = None
+    SaveArgs = None
+    StandardRestore = None
+    TypeHandler = None
+    register_type_handler = None
+    PyTreeCheckpointer = None
+    metadata = None
```
hertschuh: I think if you do a `LazyModule` for `ocp`, you won't need any of this.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,579 @@
...
+        optimizer_vars_np = [
+            keras.ops.convert_to_numpy(v) for v in model.optimizer.variables
+        ]
+        return model_weights_np, optimizer_vars_np
+    except Exception as e:
+        warnings.warn(f"Could not convert state to NumPy: {e}")
+        return None, None
+
+
+# Conditional export decorator
```
hertschuh: We definitely don't want to do that. The export should always be there. Simply, instantiating `keras.callbacks.OrbaxCheckpoint` will fail if Orbax is not installed.

With a conditional export, we run the risk of removing the export if the process building the pip wheel doesn't have Orbax installed. Then `keras.callbacks.OrbaxCheckpoint` will never work even if Orbax is installed.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,579 @@
...
+        post_finalization_callback=None,
+        save_transforms=None,
+        save_decision_policy=None,
+        save_interval=None,
+    ):
+        if ocp is None:
+            raise ImportError(
+                "OrbaxCheckpoint requires the 'orbax-checkpoint' package. "
+                "Install it with: pip install orbax-checkpoint"
+            )
```
hertschuh: You should get this error message for free with `LazyModule`.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,579 @@
...
+        # for now.
+        # A robust implementation would need a backend-aware way to check
+        # process_index.
+        is_primary_host = backend.get_process_index() == 0
+
+        if is_primary_host:
+            if self.verbose > 0:
+                print(
+                    f"OrbaxCheckpoint: Triggering async save for step {step}..."
+                )
```
hertschuh: Please use `keras.src.utils.io_utils.print_msg`.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,579 @@
...
+        if should_save:
+            # Use epoch number as the step for Orbax save
+            self._save_checkpoint(step=epoch, logs=logs)
+
+    def on_train_end(self, logs=None):
+        if self.verbose > 0:
+            print("OrbaxCheckpoint: Waiting for final saves to complete...")
+        self.manager.wait_until_finished()
+        if self.verbose > 0:
+            print("OrbaxCheckpoint: All saves finalized.")
```
hertschuh: Please use `keras.src.utils.io_utils.print_msg`.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,579 @@
...
+        # In distributed training, only load on primary process
+        if backend.get_process_index() != 0:
+            return True  # Return True to indicate no error, but no loading
+            # performed
+
+        try:
+            if self.verbose > 0:
+                print(
+                    f"OrbaxCheckpoint: Loading checkpoint from step {step}..."
+                )
```
hertschuh: Please use `keras.src.utils.io_utils.print_msg`.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,579 @@
...
+            tuple: (success, iterator_state) where success is True if loading
+            was successful, False otherwise, and iterator_state is the saved
+            data iterator state dict if available, None otherwise.
+        """
+        try:
+            # Get the latest step
+            latest_step = self.manager.latest_step()
+            if latest_step is None:
+                if self.verbose > 0:
+                    print("OrbaxCheckpoint: No checkpoints found")
```
hertschuh: Please use `keras.src.utils.io_utils.print_msg`.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,579 @@
...
+        model_weights_np = [
+            keras.ops.convert_to_numpy(w) for w in model.weights
+        ]
+        optimizer_vars_np = [
+            keras.ops.convert_to_numpy(v) for v in model.optimizer.variables
+        ]
+        return model_weights_np, optimizer_vars_np
+    except Exception as e:
+        warnings.warn(f"Could not convert state to NumPy: {e}")
+        return None, None
```
hertschuh: I think we should let the error go through. Otherwise people will think their model is saved when if fact the checkpoint will be empty and they'll only get a warning.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,579 @@
...
+            if iterator_state:
+                composite_state["data_iterator"] = iterator_state
+
+        # --- Save Logic ---
+        # Assuming single host or JAX backend with jax.distributed initialized
+        # for now.
+        # A robust implementation would need a backend-aware way to check
+        # process_index.
+        is_primary_host = backend.get_process_index() == 0
+
```
hertschuh: Wait, I thought `backend.get_process_index()` is the robust way to do it.

That should work in multi-host setups, right? What is missing?

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,579 @@
...
+                # Update best value when there's improvement
+                self.best = current
+
+        if should_save:
+            # Use epoch number as the step for Orbax save
+            self._save_checkpoint(step=epoch, logs=logs)
+
+    def on_train_end(self, logs=None):
+        if self.verbose > 0:
+            print("OrbaxCheckpoint: Waiting for final saves to complete...")
```
hertschuh: Please use `keras.src.utils.io_utils.print_msg`.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,579 @@
...
+
+        Returns:
+            tuple: (success, iterator_state) where success is True if loading
+            was successful, False otherwise, and iterator_state is the saved
+            data iterator state dict if available, None otherwise.
+        """
+        # In distributed training, only load on primary process
+        if backend.get_process_index() != 0:
+            return True  # Return True to indicate no error, but no loading
+            # performed
```
hertschuh: Performed? Remove

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,579 @@
...
+
+            return success, iterator_state
+
+        except Exception as e:
+            if self.verbose > 0:
+                print(
+                    f"OrbaxCheckpoint: Failed to load checkpoint from step "
+                    f"{step}: {e}"
+                )
+            return False, None
```
hertschuh: I think we should remove the try / except and let the error go through.

If the checkpoint fails to load, this accidentally trains the model from scratch instead of continuing training. Users won't realize.

Or am I missing something?

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,579 @@
...
+                if self.verbose > 0:
+                    print("OrbaxCheckpoint: No checkpoints found")
+                return False, None
+
+            return self.load_checkpoint(latest_step, model)
+
+        except Exception as e:
+            if self.verbose > 0:
+                print(f"OrbaxCheckpoint: Failed to load latest checkpoint: {e}")
+            return False, None
```
hertschuh: I think we should remove the try / except and let the error go through.

If the checkpoint fails to load, this accidentally trains the model from scratch instead of continuing training. Users won't realize.

Or am I missing something?

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,579 @@
...
+                        metric_idx += 1
+
+            if self.verbose > 0:
+                print("OrbaxCheckpoint: Successfully restored model state")
+            return True
+
+        except Exception as e:
+            if self.verbose > 0:
+                print(f"OrbaxCheckpoint: Failed to restore model state: {e}")
+            return False
```
hertschuh: I think we should remove the try / except and let the error go through.

If the checkpoint fails to load, this accidentally trains the model from scratch instead of continuing training. Users won't realize.

Or am I missing something?

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,579 @@
...
+def _get_state_as_numpy(model):
+    # Explicitly convert Keras weights/variables to NumPy arrays
+    try:
+        model_weights_np = [
+            keras.ops.convert_to_numpy(w) for w in model.weights
+        ]
+        optimizer_vars_np = [
+            keras.ops.convert_to_numpy(v) for v in model.optimizer.variables
+        ]
+        return model_weights_np, optimizer_vars_np
```
hertschuh: Oh, let's use the `get_state_tree` like in the example: https://keras.io/guides/orbax_checkpoint/#orbax-callback

It gives you nested dictionaries. This is a lot more in the spirit of how it's done in JAX and will allow model surgery / warm start in a much nicer way.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,579 @@
...
+
+        # --- Prepare Composite State (Backend-Agnostic) ---
+        model_weights_np, optimizer_vars_np = _get_state_as_numpy(self.model)
+
+        if model_weights_np is None:
+            if self.verbose > 0:
+                print("OrbaxCheckpoint: Skipping save due to conversion error")
+            return
+
+        composite_state = {"model_weights": model_weights_np}
```
hertschuh: If you use `get_state_tree` you won't need to do this anymore.

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -0,0 +1,1536 @@
...
+    from keras.src.callbacks.orbax_checkpoint import register_type_handler
+except ImportError:
+    OrbaxCheckpoint = None
+    CheckpointManager = None
+    SaveArgs = None
+    StandardRestore = None
+    TypeHandler = None
+    register_type_handler = None
+    PyTreeCheckpointer = None
+    metadata = None
```
hertschuh: For unit tests, it's ok to fail on import. This unit test will require orbax to be installed.

hertschuh (commented):
A couple more comments I forgot.

keras/src/backend/__init__.py
amitsrivastava78: ok

keras/src/callbacks/__init__.py
amitsrivastava78: Agree, removed the try except block and imported OrbaxCheckpoint directly so when the user instanciate and orbax is not installed it would throw an error
def __init__(self, ...):
    if ocp is None:  # ocp is None if orbax import failed
        raise ImportError(
            "OrbaxCheckpoint requires the 'orbax-checkpoint' package. "
            "Install it with: pip install orbax-checkpoint"
        )

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: Sure

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint_test.py
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: Yes, [backend.get_process_index() == 0] is the robust way to check for the primary host in multi-host setups. The current implementation is actually correct and backend-aware.

keras/src/backend/__init__.py
```diff
@@ -75,3 +75,38 @@ class name_scope(backend_name_scope):
 @keras_export("keras.device")
 def device(device_name):
     return device_scope(device_name)  # noqa: F405
+
+
+def get_process_index():
```
hertschuh: This should not be in `backend/__init__.py`, it should be in `keras/src/distribution/distribution_lib.py` and simply be:

```
def process_id():
    return distribution_lib.process_id()
```

With some documentation.

Also, do you want to call it `process_id` or `process_index`?

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,702 @@
...
+    pip_name="orbax-checkpoint",
+    import_error_msg=(
+        "OrbaxCheckpoint requires the 'orbax-checkpoint' package. "
+        "Install it with: pip install orbax-checkpoint"
+    ),
+)
+
+# Note: Advanced Orbax functionality is available through the ocp LazyModule
+# Users can access it via: from keras.src.utils.module_utils import LazyModule
+# ocp = LazyModule("orbax.checkpoint"); ocp.CheckpointManager
```
hertschuh: Let's move this declaration to `keras/src/utils/module_utils.py` then.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,702 @@
...
+    # Convert numpy scalar types to Python types for Orbax compatibility
+    def convert_scalars(obj):
+        if isinstance(obj, np.ndarray) and obj.ndim == 0:
+            # Convert 0-dimensional numpy arrays (scalars) to Python types
+            return obj.item()
+        elif isinstance(obj, np.generic):
+            # Convert numpy scalar types (like np.float32) to Python types
+            return obj.item()
+        elif isinstance(obj, dict):
+            return {k: convert_scalars(v) for k, v in obj.items()}
```
hertschuh: In general, use `keras.tree.map_structure` to do things like that, it recurses for you.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,702 @@
...
+# Users can access it via: from keras.src.utils.module_utils import LazyModule
+# ocp = LazyModule("orbax.checkpoint"); ocp.CheckpointManager
+
+
+def _get_state_tree(model):
+    """Get the complete model state as a nested tree structure."""
+    state_tree = model.get_state_tree(value_format="numpy_array")
+
+    # Convert numpy scalar types to Python types for Orbax compatibility
+    def convert_scalars(obj):
```
hertschuh: Is this for the NumPy backend? Or for the JAX backend?

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,702 @@
...
+            return obj.item()
+        elif isinstance(obj, dict):
+            return {k: convert_scalars(v) for k, v in obj.items()}
+        else:
+            return obj
+
+    return convert_scalars(state_tree)
+
+
+def _flatten_state_tree_values(state_tree):
```
hertschuh: In general, use `keras.tree.flatten` to do things like that, it recurses for you.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,702 @@
...
+def _reconstruct_state_tree_with_values(structure, values):
+    """Reconstruct state tree structure with provided values."""
+    value_iter = iter(values)
+
+    def _reconstruct(obj):
+        if isinstance(obj, dict):
+            new_dict = {}
+            for key in sorted(obj.keys()):
+                new_dict[key] = _reconstruct(obj[key])
+            return new_dict
```
hertschuh: Use `keras.tree.map_structure`.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,702 @@
...
+
+        # --- Prepare Composite State (Backend-Agnostic) ---
+        state_tree = _get_state_tree(self.model)
+
+        if state_tree is None:
+            if self.verbose > 0:
+                print_msg(
+                    "OrbaxCheckpoint: Skipping save due to state tree error"
+                )
+            return
```
hertschuh: Remove or raise an error. If the model has no state that can be saved, there is no point in continuing the training.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,702 @@
...
+
+        # Flatten the trainable variables values for cross-model compatibility
+        trainable_values = _flatten_state_tree_values(
+            state_tree["trainable_variables"]
+        )
```
hertschuh: Oh, why are you flattening each structure? I think it's much nicer to have a nested structure with the layer names, etc. Same for the optimizer, metrics etc.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,702 @@
...
+                return value
+
+    return _reconstruct(structure)
+
+
+def _restore_legacy_format(
+    checkpoint_data, target_model, save_optimizer_state, save_metrics_state
+):
+    """Restore from the old flat format for backward compatibility."""
+    # Restore model weights
```
hertschuh: I'm confused about this. What is this legacy format? And why do we need to support it?

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,702 @@
...
+                    else:
+                        return value.astype(obj.dtype).reshape(obj.shape)
+                else:
+                    return np.array(value, dtype=obj.dtype)
+            elif isinstance(obj, np.ndarray):
+                # obj is a numpy array
+                if isinstance(value, np.ndarray):
+                    return value.astype(obj.dtype).reshape(obj.shape)
+                else:
+                    return np.array(value, dtype=obj.dtype).reshape(obj.shape)
```
hertschuh: I'm confused why you need these. And doesn't Orbax give you JAX arrays? What about the sharding?

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,702 @@
...
+                elif isinstance(value, np.ndarray):
+                    # value is a numpy array, convert to scalar if needed
+                    if value.ndim == 0:
+                        return np.array(value.item(), dtype=obj.dtype)
+                    elif value.ndim == 1 and value.size == 1:
+                        return np.array(value.item(), dtype=obj.dtype)
+                    else:
+                        return value.astype(obj.dtype).reshape(obj.shape)
+                else:
+                    return np.array(value, dtype=obj.dtype)
```
hertschuh: Why are these needed? Doesn't Orbax give you back what you saved?

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,702 @@
...
+            in the checkpoint. Defaults to True.
+        save_on_background: Boolean, whether to save asynchronously in the
+            background. Defaults to True.
+        save_metadata: Dict or callable, additional metadata to save with each
+            checkpoint. If callable, it will be called with (epoch, logs) and
+            should return a dict. Defaults to None.
+        save_data_iterator: Dict or callable, data iterator state to save with
+            each checkpoint. If callable, it will be called with (epoch, logs)
+            and should return a dict with serializable iterator state.
+            Defaults to None.
```
hertschuh: Curious what this feature is for? It seems very similar to the one above.

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -0,0 +1,1525 @@
...
+    from keras.src.callbacks.orbax_checkpoint import CheckpointManager
+    from keras.src.callbacks.orbax_checkpoint import PyTreeCheckpointer
+    from keras.src.callbacks.orbax_checkpoint import SaveArgs
+    from keras.src.callbacks.orbax_checkpoint import StandardRestore
+    from keras.src.callbacks.orbax_checkpoint import TypeHandler
+    from keras.src.callbacks.orbax_checkpoint import metadata
+    from keras.src.callbacks.orbax_checkpoint import register_type_handler
+except ImportError:
+    # If orbax is not available, these won't be exported
+    pass
```
hertschuh: Just use the LazyLoader.

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -0,0 +1,1525 @@
...
+
+from keras.src.callbacks.orbax_checkpoint import OrbaxCheckpoint
+
+# Skip the entire test module if orbax-checkpoint is not available
+pytest.importorskip("orbax.checkpoint")
```
hertschuh: We don't do it this way because if we screw up the install, the test is skipped silently and everything passes, which we may never realize.

Instead we always try the tests. If you need to skip, you should have an explicit condition based on when you know it won't work. For instance, if it's JAX only, skip if backend is not JAX at the class level.

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -0,0 +1,1525 @@
...
+
+
+class OrbaxCheckpointTest(testing.TestCase):
+    def setUp(self):
+        super().setUp()
+        self.temp_dir = tempfile.mkdtemp()
```
hertschuh: Just use `self.get_temp_dir()` directly in the test. It will cleanup automatically. No need for `setUp` nor `tearDown`.

keras/src/backend/__init__.py
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: The convert_scalars function is not specific to any particular backend (NumPy, JAX, TensorFlow, or PyTorch). Here's why it's needed:
The function is called in _get_state_tree() after this line:
state_tree = model.get_state_tree(value_format="numpy_array")
This Keras API method converts the model's state (weights, optimizer state, etc.) from whatever backend-specific tensor format (JAX arrays, TF tensors, PyTorch tensors) into NumPy arrays, regardless of which backend you're using.
The conversion problem - After this conversion, some values in the state tree might be:

0-dimensional NumPy arrays ([array(0.5)], NumPy scalar types (np.float32(1.0), np.int64(42)
Orbax serialization requirements - Orbax checkpointing has specific requirements for data types that can be serialized. NumPy scalar types and 0-dimensional arrays may not serialize correctly or may cause compatibility issues.
Input: The function receives NumPy arrays that came from any backend (JAX/TF/PyTorch/NumPy)
Processing: It only looks at NumPy data types
utput: Standard Python types  that Orbax can reliably serialize

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: I would not recommend removing the as it helps in 
Backward Compatibility The function handles the original checkpoint format from when OrbaxCheckpoint was first implemented in Keras 3. Users may still have checkpoints saved in this format that they need to load.
The checkpoint formats have evolved:

Legacy format → Original implementation
Flattened values → After tree utilities integration
Nested structures → Current format (preserves layer names)

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: Orbax saves checkpoints as numpy arrays regardless of the backend (JAX/TensorFlow/PyTorch). When loading, you get numpy arrays back, not JAX arrays.
The obj is the raw checkpoint data: If obj but checkpoint has float64[50], this reshapes and converts it.
Orbax does NOT give you JAX arrays directly. It gives you numpy arrays for cross-backend compatibility:

Save: JAX_array → numpy_array via backend.convert_to_numpy()
Load: numpy_array ← convert back via ops.convert_to_tensor()
The JAX arrays come later when the model assigns these values to its variables.
sharding is handled in a separate PR as suggested by you https://github.com/keras-team/keras/pull/21786

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: The conversions are needed because Orbax gives you back exactly what you saved, but the saved data might not match the current model's expectations. This reconstruction code handles mismatches that can occur due to model surgery of model configuration getting changed.

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: The save_data_iterator parameter is indeed similar to save_metadata, but serves a very specific and important purpose in training resumption scenarios.

keras/src/callbacks/orbax_checkpoint_test.py

keras/src/callbacks/orbax_checkpoint_test.py
amitsrivastava78: ok done

keras/src/callbacks/orbax_checkpoint_test.py
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py
hertschuh: Can you give an example? We didn't have that in the regular checkpoint format, so I'm curious.

keras/src/callbacks/orbax_checkpoint.py
hertschuh: > state_tree = model.get_state_tree(value_format="numpy_array")

Oh, but we can't do "numpy_array", at least not for JAX. Because at that point the sharding is irreversibly lost and this is no longer distributed checkpointing.

> NumPy scalar types and 0-dimensional arrays may not serialize correctly or may cause compatibility issues.

Is this documented? Or is this a bug? Do you understand why?

keras/src/callbacks/orbax_checkpoint.py
hertschuh: > OrbaxCheckpoint was first implemented in Keras 3.

We never implemented OrbaxCheckpoint in Keras 3.

We have a guide, but we never support stuff in guides. We really don't need to add code to support this.

keras/src/callbacks/orbax_checkpoint.py
hertschuh: > Orbax saves checkpoints as numpy arrays regardless of the backend (JAX/TensorFlow/PyTorch). When loading, you get numpy arrays back, not JAX arrays.

Where is the sharding information then?

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,769 @@
...
+        elif isinstance(obj, np.generic):
+            # Convert numpy scalar types (like np.float32) to Python types
+            return obj.item()
+        else:
+            return obj
+
+    return tree.map_structure(convert_scalars, state_tree)
+
+
+def _flatten_state_tree_values(state_tree):
```
hertschuh: Remove.

keras/src/callbacks/orbax_checkpoint.py
hertschuh: You can and should pass Jax arrays with the Jax backend, that's how the sharding is supported: https://orbax.readthedocs.io/en/latest/guides/checkpoint/checkpointing_pytrees.html#basic-checkpointing

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: I have added a test for this test_save_data_iterator_state, test_load_checkpoint_with_iterator_state , test_tensorflow_iterator_restoration &     def test_jax_iterator_restoration(self):

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: This part of the code is removed after migration to orbax V1

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,521 @@
...
+        checkpointer_kwargs = {
+            "directory": directory,
+            "preservation_policy": preservation_policy,
+            "step_name_format": step_name_format,
+            "custom_metadata": custom_metadata,
+        }
+        if save_decision_policy is not None:
+            checkpointer_kwargs["save_decision_policy"] = save_decision_policy
+
+        self.checkpointer = ocp.training.Checkpointer(**checkpointer_kwargs)
```
cpgaffney1: This construction is a bit weird, you lose the type checking by doing this. Also, you can just pass save_decision_policy if it is None - Orbax will give you a default behavior. Though I suggest Keras define its own default policy.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,521 @@
...
+            should return a dict. Defaults to None.
+        save_data_iterator: Dict or callable, data iterator state to save with
+            each checkpoint. If callable, it will be called with (epoch, logs)
+            and should return a dict with serializable iterator state.
+            Defaults to None.
+        save_metrics_state: Boolean, whether to include stateful metrics
+            variables in the checkpoint. Defaults to False.
+        async_timeout_secs: Integer, timeout in seconds for async checkpointing
+            operations. Defaults to 600 (10 minutes).
+        enable_background_delete: Boolean, whether to delete old checkpoints in
```
cpgaffney1: What is the use case? I recommend leaving them to the default behaviors unless there is a known need.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,521 @@
...
+            else:
+                iterator_state = self.save_data_iterator
+            if iterator_state:
+                composite_state["data_iterator"] = iterator_state
+
+        # --- Save Logic (V1 API) ---
+        # Only save on the primary process (rank 0) in distributed setups
+        is_primary_host = process_id() == 0
+
+        if is_primary_host:
```
cpgaffney1: This will almost certainly lead to bugs with distributed arrays. Always call public Orbax APIs on all processes.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,521 @@
...
+                async_options["post_finalization_callback"] = (
+                    self.post_finalization_callback
+                )
+
+            # Add the other async options
+            if self.async_timeout_secs is not None:
+                async_options["timeout_secs"] = self.async_timeout_secs
+
+            # Map enable_background_delete to its V1 equivalent
+            async_options["create_directories_asynchronously"] = (
```
cpgaffney1: Semantically, these are not the same options. I don't see any need for you to override, unless I am missing something.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,521 @@
...
+
+            # Use a single with statement. If context_options is empty,
+            # Context() uses defaults.
+            with ocp.Context(**context_options):
+                if self.save_on_background:
+                    self.checkpointer.save_pytree_async(step, composite_state)
+                else:
+                    self.checkpointer.save_pytree(step, composite_state)
+
+    def on_train_batch_end(self, batch, logs=None):
```
cpgaffney1, line 224: This and the following method are a bit confusing because you are performing save decisions outside the Checkpointer. If possible, you should either let the Checkpointer fully decide for itself (which would encapsulate the logic here), or configure Checkpointer to save unconditionally, and make all saving decisions in keras.

Otherwise the division of responsibilities gets confusing, and the two layers might have conflicting decisions for a particular step.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,521 @@
...
+
+    def wait_until_finished(self):
+        """Wait for any in-progress checkpoint operations to complete.
+
+        This method blocks until all asynchronous checkpoint save operations
+        have completed. It should be called before attempting to load
+        checkpoints if there might be pending save operations.
+        """
+        # Wait for any async operations to complete
+        while self.checkpointer.is_saving_in_progress():
```
cpgaffney1: Probably better to use checkpointer.wait() here, unless you want to log things periodicially.

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: ok removed enable_background_delete & async_timeout_secs

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: ok

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: Removed

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: Fix confusing division of save decision responsibilities in OrbaxCheckpoint
Configure Checkpointer with AlwaysSavePolicy so it saves unconditionally
when save_pytree/save_pytree_async is called, eliminating conflicting
decision logic between Keras and Orbax layers. Keras now handles all
save decisions (save_freq, save_best_only, etc.) while Orbax just
executes the saves.

keras/src/callbacks/orbax_checkpoint.py

keras/src/distribution/distribution_lib.py
```diff
@@ -151,9 +137,11 @@ class DeviceMesh:
     represents the computation devices in the global context.
 
     See more details in [jax.sharding.Mesh](
-        https://jax.readthedocs.io/en/latest/jax.sharding.html#jax.sharding.Mesh)
+        https://jax.readthedocs.io/en/latest/jax.sharding.html
+        #jax.sharding.Mesh)
```
hertschuh: Undo this file, this will break the links.

keras/src/utils/module_utils.py
```diff
@@ -59,3 +59,11 @@ def __repr__(self):
 dmtree = LazyModule("tree")
 tf2onnx = LazyModule("tf2onnx")
 grain = LazyModule("grain")
+ocp = LazyModule(
+    "orbax.checkpoint.experimental.v1",
+    pip_name="orbax-checkpoint",
+    import_error_msg=(
+        "OrbaxCheckpoint requires the 'orbax-checkpoint' package. "
+        "Install it with: pip install orbax-checkpoint"
```
hertschuh: nitpick: just for consistency with the other ones: `"You can install it via `pip install orbax-checkpoint`".

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,497 @@
...
+def _get_state_tree(model):
+    """Get the complete model state as a nested tree structure."""
+    # For JAX backend, preserve native arrays if JAX monitoring available
+    # to avoid unnecessary conversions. Otherwise convert to numpy.
+    if backend.backend() == "jax":
+        try:
+            import jax
+
+            # Check if jax.monitoring.record_scalar exists (JAX 0.7.0+)
+            jax.monitoring.record_scalar
```
hertschuh: I'm confused by this. `jax.monitoring.record_scalar` is unrelated to this.

1. What is the issue with JAX < 0.7.0?
2. Why don't you check the jax version directly? https://github.com/keras-team/keras-rs/blob/5d7f18ae1103cec9a36fa621aa3f1f5d1128e2a7/keras_rs/src/layers/embedding/jax/distributed_embedding_test.py#L50

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,497 @@
...
+    # Convert numpy scalar types to Python types for Orbax compatibility
+    def convert_scalars(obj):
+        if isinstance(obj, np.ndarray) and obj.ndim == 0:
+            # Convert 0-dimensional numpy arrays (scalars) to Python types
+            return obj.item()
+        elif isinstance(obj, np.generic):
+            # Convert numpy scalar types (like np.float32) to Python types
+            return obj.item()
+        else:
+            return obj
```
hertschuh: Per the comments in the design doc, this conversion is not needed.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,497 @@
...
+            and should return a dict with serializable iterator state.
+            Defaults to None.
+        save_metrics_state: Boolean, whether to include stateful metrics
+            variables in the checkpoint. Defaults to False.
+        post_finalization_callback: Callable, function to call after async
+            checkpointing operations complete. Defaults to None.
+        save_decision_policy: orbax.checkpoint.SaveDecisionPolicy object to
+            control when checkpoints are saved. If None, defaults to saving
+            every epoch for save_freq="epoch" or every save_freq batches.
+            Defaults to None.
```
hertschuh: Oh, we had agreed in the design doc to trim down the list of features. I think we should start with parity with `ModelCheckpoint`. Some of the extra features have never been asked. And they would be easy to add when needed.

So let's start with:
```
        filepath,
        monitor="val_loss",
        verbose=0,
        save_best_only=False,
        save_weights_only=False,
        mode="auto",
        save_freq="epoch",
        initial_value_threshold=None,
```

In particular:
- replace `save_optimizer_state` and `save_metrics_state` with `save_weights_only` (the opposite)
- remove `save_metadata`, `save_data_iterator`, `post_finalization_callback`, `save_decision_policy` for now

I'm ok keeping `save_on_background` and `max_to_keep` though.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,497 @@
...
+            ):
+                def should_save(
+                    self, current_step_info, previous_steps=None, context=None
+                ):
+                    return True
+
+            save_decision_policy = _AlwaysSavePolicy()
+
+        # --- Orbax Checkpointer Setup (V1 API) ---
+        # Map V0 options to V1 parameters
```
hertschuh: `# Map V0 options to V1 parameters`

I'm not following this comment. We're not using anything V0. I guess just remove the comment.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,497 @@
...
+        if hasattr(self.model, "optimizer") and hasattr(
+            self.model.optimizer, "iterations"
+        ):
+            # Convert potential backend tensor to int
+            return int(
+                backend.convert_to_numpy(self.model.optimizer.iterations)
+            )
+        else:
+            # Fallback: use global batch count
+            return self._total_batches_seen
```
hertschuh: I'm not following why you need 2 mechanisms here.

Just use `self._total_batches_seen` everywhere and remove `_get_current_step`.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,497 @@
...
+
+    def _save_checkpoint(self, step, logs=None):
+        """Save a checkpoint at the given step."""
+        if self.model is None:
+            return
```
hertschuh: Not a valid use case, remove this check. If it fails it's a programmer error and we want an exception raised.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,497 @@
...
+
+        # --- Prepare Composite State (Backend-Agnostic) ---
+        state_tree = _get_state_tree(self.model)
+
+        if state_tree is None:
+            raise RuntimeError(
+                "OrbaxCheckpoint: Failed to get model state tree. "
+                "The model may not be properly built or may have no "
+                "savable state."
+            )
```
hertschuh: Looking at the code, this case is impossible. Remove.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,497 @@
...
+
+        if should_save:
+            # Use epoch number as the step for Orbax save
+            # Keras has already made the save decision - Checkpointer will
+            # save unconditionally
+            self._save_checkpoint(step=epoch, logs=logs)
+
+    def on_train_end(self, logs=None):
+        if self.verbose > 0:
+            print_msg("OrbaxCheckpoint: Training completed.")
```
hertschuh: Remove this. There are other indications that the training is complete (and it's not an Orbax feature)

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,497 @@
...
+
+# Export additional Orbax functionality for advanced users (only if available)
+if ocp.available:
+    Checkpointer = ocp.training.Checkpointer
+    save_pytree = ocp.save_pytree
+    load_pytree = ocp.load_pytree
+    save_pytree_async = ocp.save_pytree_async
+    load_pytree_async = ocp.load_pytree_async
+    preservation_policies = ocp.training.preservation_policies
+    save_decision_policies = ocp.training.save_decision_policies
```
hertschuh: Remove this. It's not useable, not even by advanced users. We don't want users to import from `keras.src` ever.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,497 @@
...
+        if self.verbose > 0:
+            print_msg("OrbaxCheckpoint: Training completed.")
+
+        # Close the Checkpointer to ensure all pending saves complete
+        try:
+            self.checkpointer.close()
+        except Exception:
+            pass  # Ignore errors during cleanup
+
+    def load_checkpoint(self, step, model=None):
```
hertschuh: Hmm... I'll have to think about where `load_checkpoint`, `load_latest`, `all_steps` should go. This is not how the Keras loading APIs have worked so far.

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -0,0 +1,1506 @@
...
+    _orbax_available = _orbax_available and True
+except ImportError:
+    OrbaxCheckpoint = None
+    _orbax_available = False
+
+
+@pytest.mark.skipif(
+    not _orbax_available,
+    reason="OrbaxCheckpoint requires the 'orbax-checkpoint' package",
+)
```
hertschuh: Remove all the `if`s and `try / except` and `@pytest.mark.skipif`. The unit test has to fail if Orbax is not installed.

keras/src/utils/module_utils.py
amitsrivastava78: Done

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: Orbax's JAX array serialization involves a function called get_replica_slices
This function is responsible for preparing JAX arrays for distributed storage/sharding.
Inside get_replica_slices, Orbax calls record_scalar, which is there on the new version, if we convert even jax backend to numpy this error does not come. i have changed the versioning part a bit

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: Yes updated it, For jax no conversion is required

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint_test.py
amitsrivastava78: This would cause the CI failure on GPU tests where orbax is not insalled

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: ok one option can be the OrbaxCheckpoint API to follow Keras' traditional separation of concerns, moving loading functionality from the callback class to standalone functions in the saving API.
@keras_export("keras.saving.load_orbax_checkpoint")
def load_orbax_checkpoint(directory, step, model=None)

@keras_export("keras.saving.load_latest_orbax_checkpoint") 
def load_latest_orbax_checkpoint(directory, model=None)

@keras_export("keras.saving.list_orbax_checkpoint_steps")
def list_orbax_checkpoint_steps(directory)

load_checkpoint(step, model=None) → move to keras.saving.load_orbax_checkpoint()
load_latest(model=None) → move to keras.saving.load_orbax_checkpoint()
all_steps() → move to keras.saving.list_orbax_checkpoint_steps()

Let me know what you think

keras/src/distribution/distribution_lib.py
amitsrivastava78: Done

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: Removed Features:
* save_metadata parameter and functionality
* save_data_iterator parameter and functionality
* post_finalization_callback parameter and functionality
* save_decision_policy parameter and functionality
* keep_period parameter and functionality
* load_checkpoint(), load_latest(), and all_steps() methods
Parameter Changes:
* Replaced save_optimizer_state and save_metrics_state with a single save_weights_only parameter (matching ModelCheckpoint exactly)
* Kept max_to_keep and save_on_background as they provide Orbax-specific value without conflicting with ModelCheckpoint
Test Cleanup:
* Removed all tests for deleted features

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: Also when we removed the features as suggested by you these functions are no longer needed, So in the latest commit these would not be there, let me know if you are ok with this ?

keras/src/callbacks/orbax_checkpoint_test.py
hertschuh: Oh, let's add it to the requirements, it's important to test this on GPU

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,294 @@
...
+        import jax
+        from packaging import version
+
+        # Check JAX version directly (JAX 0.7.0+ supports better array handling)
+        if version.parse(jax.__version__) >= version.parse("0.7.0"):
+            state_tree = model.get_state_tree()
+        else:
+            # Fallback to numpy conversion for older JAX versions
+            state_tree = model.get_state_tree(value_format="numpy_array")
+            did_numpy_conversion = True
```
hertschuh: What is the issue with JAX < 0.7.0 ?

Orbax checkpointing was available way before 0.7.0 so I'm really surprised.

Also, we have issues with our CI that prevents us from upgrading JAX, so we won't be testing the "real" use case, which is not great.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,294 @@
...
+    ```
+
+    Args:
+        directory: string, path to the directory where to save the checkpoints.
+        monitor: The metric name to monitor (e.g., 'val_loss').
+        verbose: Verbosity mode, 0 or 1.
+        save_best_only: if `save_best_only=True`, it only saves when the model
+            is considered the "best" based on the monitored quantity.
+        save_weights_only: if `save_weights_only=True`, only the model's weights
+            will be saved. Otherwise, both weights and optimizer state will be
```
hertschuh: It's not just the weights and optimizer state, it's also the metrics.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,294 @@
...
+        self.save_freq = save_freq
+        self.max_to_keep = max_to_keep
+        self.save_on_background = save_on_background
+        self._batches_seen_since_last_saving = 0
+        self._last_batch_seen = 0
+        self._current_epoch = 0  # Keep track of epoch
+        self._total_batches_seen = 0  # Global batch counter for step tracking
+
+        if self.save_freq != "epoch" and not isinstance(self.save_freq, int):
+            raise ValueError("Unrecognized save_freq")
```
hertschuh: Give a slightly more informative error (this is from ModelCheckpoint):
```python
            raise ValueError(
                f"Unrecognized save_freq: {self.save_freq}. "
                "Expected save_freq are 'epoch' or integer values"
            )
```

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,294 @@
...
+    def _save_checkpoint(self, step, logs=None):
+        """Save a checkpoint at the given step."""
+
+        # --- Prepare Composite State (Backend-Agnostic) ---
+        state_tree = _get_state_tree(self.model)
+
+        # Save the nested state structures directly (preserving layer
+        # names and structure)
+        composite_state = {
+            "trainable_variables": state_tree["trainable_variables"],
```
hertschuh: What about "non_trainable_variables"?

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,294 @@
...
+        }
+
+        # Include optimizer state unless save_weights_only is True
+        if not self.save_weights_only and "optimizer_variables" in state_tree:
+            composite_state["optimizer_variables"] = state_tree[
```
hertschuh: This is missing the metrics. If not save_weights_only should should just take the whole state tree:

```python
if self.save_weights_only:
    composite_state = state_tree
else:
    composite_state = {
            "trainable_variables": state_tree["trainable_variables"],
            "non_trainable_variables": state_tree["non_trainable_variables"],
    }
```

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,294 @@
...
+        """Wait for any in-progress checkpoint operations to complete.
+        This method blocks until all asynchronous checkpoint save operations
+        have completed. It should be called before attempting to load
+        checkpoints if there might be pending save operations.
+        """
+        # Wait for any async operations to complete
+        try:
+            self.checkpointer.wait()
+        except AttributeError:
+            # Fallback for older Orbax versions that don't have wait() method
```
hertschuh, line 295: Do we need to support that? Isn't it always there in V1?

keras/src/callbacks/orbax_checkpoint.py
hertschuh: So I'm ok with you removing these, but we still need one thing.

`keras.models.Model.load("orbax_path")` should work seamlessly. It should load the latest checkpoint. I'm not sure if there is an easy way to detect that it's an orbax folder rather than a tf format folder.

This support can be added in a separate PR if you prefer.

amitsrivastava78: I think we can detect the orbax checkpoint by the directory sturcture, Look for step subdirectories & Verify Orbax-specific files in subdirectories, will put it as part of separete PR, once this is merged

keras/src/callbacks/orbax_checkpoint_test.py
amitsrivastava78: But if i add orbax installation in requirements-common.txt then it will be installed every time even for users who do not want to use Orbax/Jax

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: Understood, i have removed the version check

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: Yes updated the comment and added a test case as well.

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: ok updated the error info

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: Sure added it and updated the test case as well

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: I've updated the code to simplify the logic as you suggested

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: The issue is that the V1 API import path doesn't guarantee all V1 methods are available - the V1 API evolved incrementally over time.

V1 API import: from orbax.checkpoint import v1 - this import path has existed since early V1 releases
But wait()method: Was added later in the V1 API evolution

keras/src/callbacks/orbax_checkpoint_test.py
hertschuh: Let's add it to `requirements-common.txt`.

All the requirements files today are for developers coding and running unit tests and the CI. A vast majority of things in `requirements-common.txt` are optional and many are only for testing and would never be used by users of Keras.

It is not what controls what is installed when you install Keras.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,300 @@
...
+    ModelCheckpoint.
+
+    This callback saves the model's weights and optimizer state asynchronously
+    using Orbax, allowing training to continue without blocking for I/O.
+
+    Example:
+
+    ```python
+    model.compile(loss=..., optimizer=...,
+                  metrics=['accuracy'])
```
hertschuh: Nitpick: this can fit on a single line.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,300 @@
...
+        save_best_only: if `save_best_only=True`, it only saves when the model
+            is considered the "best" based on the monitored quantity.
+        save_weights_only: if `save_weights_only=True`, only the model's
+            weights will be saved. Otherwise, the full model state
+            (weights, non-trainable variables, optimizer state, and
+            metrics state) will be saved. Defaults to False.
+        mode: one of {'auto', 'min', 'max'}. Used with `save_best_only`.
+        save_freq: `'epoch'` or integer. Frequency to save checkpoints.
+        max_to_keep: Integer, maximum number of recent checkpoints to keep.
+            If None, keeps all. Defaults to 5.
```
hertschuh: If think it would be reasonable to set the default to 1 instead of 5.

- most people will not go back and compare the quality of the checkpoints
- it will behave to same by default as `ModelCheckpoint`.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,300 @@
...
+        directory,
+        monitor="val_loss",
+        verbose=0,
+        save_best_only=False,
+        save_weights_only=False,
+        mode="auto",
+        save_freq="epoch",
+        max_to_keep=5,
+        save_on_background=True,
+        initial_value_threshold=None,
```
hertschuh, line 118: nitpick: move up between `save_freq` and `max_to_keep` for consistency with `ModelCheckpoint`.

keras/src/utils/module_utils.py
```diff
@@ -24,7 +24,14 @@ def available(self):
...
         try:
-            self.module = importlib.import_module(self.name)
+            # Special handling for orbax.checkpoint.v1
+            if self.name == "orbax.checkpoint.v1":
+                # Import the parent module and get the v1 submodule
+                parent_module = importlib.import_module("orbax.checkpoint")
+                self.module = parent_module.v1
+            else:
+                # Normal module import
+                self.module = importlib.import_module(self.name)
```
hertschuh: Hmm... let's do a subclass instead:

```python
class OrbaxLazyModule(LazyModule):
    def initialize(self):
        try:
            parent_module = importlib.import_module("orbax.checkpoint")
            self.module = parent_module.v1
        except ImportError:
            raise ImportError(self.import_error_msg)

...
ocp = OrbaxLazyModule(...)
```

Alternatively you could have an optional `submodule` in `LazyModule.__init__`.

keras/src/callbacks/orbax_checkpoint_test.py
amitsrivastava78: Sure

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: Sure

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: Sure

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: Sure

keras/src/utils/module_utils.py
amitsrivastava78: I have created the subclass that handles the special import logic removed the conditional code from the base class,

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -0,0 +1,299 @@
...
+        directory,
+        monitor="val_loss",
+        verbose=0,
+        save_best_only=False,
+        save_weights_only=False,
+        mode="auto",
+        max_to_keep=1,
+        save_on_background=True,
+        save_freq="epoch",
+        initial_value_threshold=None,
```
hertschuh, line 118: Oh, I meant that the order would be:
```
        directory,
        monitor="val_loss",
        verbose=0,
        save_best_only=False,
        save_weights_only=False,
        mode="auto",
        save_freq="epoch",
        initial_value_threshold=None,
        max_to_keep=1,
        save_on_background=True,
```

So that the 8 first arguments are the same as `ModelCheckpoint`, this makes it a drop-in replacement.

amitsrivastava78: Done

PR #21763, opened 2025-10-22
Suport keras.op.view() to view the same data bitwise at a new dtype 

wenyi-guo:
Support view() on backeds numpy, jax, tensorflow, pytorch.

### DIFF
--- a/keras/src/ops/numpy.py
+++ b/keras/src/ops/numpy.py
@@ -1183,8 +1183,7 @@ def view(x, dtype=None):
     """
     if any_symbolic_tensors((x,)):
         return View(dtype=dtype).symbolic_call(x)
-    r = backend.numpy.view(x, dtype=dtype)
-    return r
+    return backend.numpy.view(x, dtype=dtype)
 
 
 class Average(Operation):


keras/src/backend/numpy/numpy.py
```diff
@@ -294,6 +294,14 @@ def array(x, dtype=None):
     return convert_to_tensor(x, dtype=dtype)
 
 
+def view(x, dtype=None, type=None):
+    return (
+        x.view(dtype=dtype, type=type)
+        if type is not None
+        else x.view(dtype=dtype)
+    )
```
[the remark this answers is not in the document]
wenyi-guo: Need the logic

keras/src/ops/numpy.py
```diff
@@ -1124,6 +1124,75 @@ def array(x, dtype=None):
...
+@keras_export(["keras.ops.view", "keras.ops.numpy.view"])
+def view(x, dtype=None, type=None):
+    """Create a new bitwise view of the same data with the specified dtype.
+
+    Args:
+        x: Input tensor.
+        dtype: Data-type descriptor of the returned view,
+            e.g., float32 or int16.
+        type: Type of the returned view, e.g., ndarray or matrix.
+            Only support for numpy.
```
hertschuh: Let's completely remove the `type` argument everywhere since it's not supported by any backend. That's the approach we've taken for arguments that we can't support. It will simplify the code a little.

keras/src/ops/numpy.py
```diff
@@ -1124,6 +1124,75 @@ def array(x, dtype=None):
...
+
+    Returns:
+        A tensor.
```
hertschuh: Let's add a little more here: "View of `tensor` with data type `dtype`".

keras/src/ops/numpy.py
```diff
@@ -1124,6 +1124,75 @@ def array(x, dtype=None):
...
+            raise ValueError(
+                "Cannot view a scalar as a different dtype if item sizes "
+                "are different."
+            )
+
+        output_shape = list(x.shape)
+        if output_shape[-1] is None:
+            # Cannot infer shape if last dimension is dynamic
+            output_shape[-1] = None
+        else:
```
hertschuh: Oh, the branch when `output_shape[-1] = None` is a no-op. How about just doing:

```python
if output_shape[-1] is not None:
```

keras/src/backend/tensorflow/numpy.py
```diff
@@ -998,6 +998,58 @@ def array(x, dtype=None):
...
+
+    old_itemsize = old_dtype.size
+    new_itemsize = new_dtype.size
+
+    if list(x.shape)[-1] * old_itemsize % new_itemsize != 0:
+        raise ValueError(
+            f"Cannot view array of shape {x.shape} and dtype {old_dtype} "
+            f"as dtype {new_dtype} because the total number of bytes "
+            f"is not divisible by the new itemsize."
+        )
```
hertschuh, line 1018: `tf.bitcast` won't do that check for you?

keras/src/backend/tensorflow/numpy.py
```diff
@@ -998,6 +998,58 @@ def array(x, dtype=None):
...
+
+    old_shape = tf.shape(x)
+
+    if old_itemsize == new_itemsize:
+        return tf.bitcast(x, type=new_dtype)
+    elif old_itemsize > new_itemsize:
+        ratio = old_itemsize // new_itemsize
+        new_shape_list = tf.unstack(old_shape)
+        new_shape_list[-1] = new_shape_list[-1] * ratio
+        new_shape = tf.stack(new_shape_list)
```
hertschuh: `keras.ops.shape` makes it a little bit easier. It combines `x.shape` and `tf.shape(x)` and returns a tuple of ints and scalar tensors for dynamic dimensions. It can be used directly with `tf.reshape` (no need to restack).

It's imported as `shape_op`.

```python
    elif old_itemsize > new_itemsize:
        new_shape = list(shape_op(x))
        new_shape[-1] *= ratio
        flat_tensor = tf.reshape(x, [-1])
        cast_tensor = tf.bitcast(flat_tensor, type=new_dtype)
        return tf.reshape(cast_tensor, new_shape)
```

keras/src/backend/tensorflow/numpy.py
```diff
@@ -998,6 +998,58 @@ def array(x, dtype=None):
...
+        return tf.reshape(cast_tensor, new_shape)
+    else:
+        last_dim_size = old_shape[-1]
+        ratio = new_itemsize // old_itemsize
+        if last_dim_size % ratio != 0:
+            raise ValueError(
+                f"Cannot view dtype. Last dimension size ({last_dim_size}) "
+                f"must be divisible by the ratio of new/old item sizes "
+                f"({ratio})."
+            )
```
hertschuh: If the last dim is dynamic, this `if` won't work (we turn off autograph).

So change to:
```python
if isinstance(last_dim_size, int) and last_dim_size % ratio != 0:
```

In the dynamic case,  we'll let `tf.bitcast` fail for us, we won't do a dynamic check (I hear that those are expensive).

keras/src/backend/tensorflow/numpy.py
```diff
@@ -998,6 +998,58 @@ def array(x, dtype=None):
...
+        intermediate_last_two_dims = [last_dim_size // ratio, ratio]
+        intermediate_shape = tf.concat(
+            [old_shape[:-1], intermediate_last_two_dims], axis=0
+        )
+
+        reshaped_tensor = tf.reshape(x, intermediate_shape)
```
hertschuh: Same comment about using `shape_op` and not needing `concat`.

```python
old_shape = list(shape_op(x))
last_dim_size = old_shape[-1]
... 
intermediate_shape = old_shape[:-1] + [last_dim_size // ratio, ratio]
reshaped_tensor = tf.reshape(x, intermediate_shape)
```

keras/src/ops/numpy_test.py
```diff
@@ -9325,3 +9325,114 @@ def call(self, x):
         model.compile(jit_compile=jit_compile)
 
         model.predict(np.random.randn(1, 8))
+
+
+class NumpyTestView(testing.TestCase):
+    @pytest.mark.skipif(
+        keras.config.backend() == "openvino",
+        reason="OpenVINO doesn't support this change",
+    )
```
hertschuh: Instead of doing this, add to this file:
https://github.com/keras-team/keras/blob/master/keras/src/backend/openvino/excluded_concrete_tests.txt

That's how we keep track of what's missing.

keras/src/ops/numpy_test.py
```diff
@@ -9325,3 +9325,114 @@ def call(self, x):
         model.compile(jit_compile=jit_compile)
 
         model.predict(np.random.randn(1, 8))
+
+
+class NumpyTestView(testing.TestCase):
```
hertschuh: Please follow the same pattern as other ops. Add:
- NumpyOneInputOpsDynamicShapeTest::test_view to test KerasTensor with dynamic dimension
- NumpyOneInputOpsStaticShapeTest::test_view to test KerasTensor with all static dimensions
- NumpyOneInputOpsCorrectnessTest::test_view to test some numerics like you have here
- NumpyDtypeTest::test_view with the cross product of the input / output dtypes

```python
    @parameterized.named_parameters(
        named_product(dtypes=itertools.combinations(ALL_DTYPES, 2))
    )
    def test_view(self, dtypes):
        import jax.numpy as jnp

        input_dtype1, output_dtype = dtypes
        x = knp.ones((20,), dtype=input_dtype)
       ...
```

keras/src/ops/numpy.py
wenyi-guo: I see the numpy and jax view() lib supports `type`, but it's not widely supported, removed now. Let me know if still want to add for numpy and jax backend

keras/src/backend/tensorflow/numpy.py
wenyi-guo: `tf.bitcast` only supports shape[-1] == new_itemsize / old_itemsize if new_itemsize is larger, while other `view()` libs support shape[-1] = multiples of (new_itemsize / old_itemsize). So added this logic.

keras/src/ops/numpy.py
hertschuh: JAX has the argument, but it's not actually supported (fails if it's not None). So let's remove everywhere.

keras/src/ops/numpy_test.py
```diff
@@ -1858,6 +1858,14 @@ def test_angle(self):
         x = KerasTensor((None, 3))
         self.assertEqual(knp.angle(x).shape, (None, 3))
 
+    def test_view(self):
+        x = knp.array(KerasTensor((None, 3)), dtype="int32")
+        self.assertEqual(knp.view(x, dtype="uint32").shape, (None, 3))
+        self.assertEqual(knp.view(x, dtype="uint32").dtype, "uint32")
+        x = knp.array(KerasTensor((None, 3)), dtype="int32")
+        self.assertEqual(knp.view(x, dtype="int16").shape, (None, 6))
+        self.assertEqual(knp.view(x, dtype="int16").dtype, "int16")
```
hertschuh, line 1867: Also add the opposite case, i.e. "int16" to "int32" to exercise the opposite case when the destination dtype is larger than the source dtype.

keras/src/ops/numpy_test.py
```diff
@@ -2459,6 +2467,14 @@ def test_angle(self):
         x = KerasTensor((2, 3))
         self.assertEqual(knp.angle(x).shape, (2, 3))
 
+    def test_view(self):
+        x = knp.array(KerasTensor((2, 3)), dtype="int32")
+        self.assertEqual(knp.view(x, dtype="uint32").shape, (2, 3))
+        self.assertEqual(knp.view(x, dtype="uint32").dtype, "uint32")
+        x = knp.array(KerasTensor((2, 3)), dtype="int32")
+        self.assertEqual(knp.view(x, dtype="int16").shape, (2, 6))
+        self.assertEqual(knp.view(x, dtype="int16").dtype, "int16")
```
hertschuh, line 2479: Also add the opposite case, i.e. "int16" to "int32" to exercise the opposite case when the destination dtype is larger than the source dtype.

keras/src/ops/numpy_test.py
```diff
@@ -4056,6 +4072,115 @@ def test_concatenate(self):
...
+
+        import jax
+        import jax.numpy as jnp
+        import torch
+
+        input_array = knp.array([[1, 2, 3], [4, 5, 6]], dtype="int32")
+
+        # Test case 1: View with old dtype
+        if backend.backend() == "tensorflow":
+            result = knp.view(input_array, dtype="int32")
```
hertschuh: You don't need the other ones, you should always use `result = knp.view(input_array, dtype="int32")` for all backends

keras/src/ops/numpy_test.py
```diff
@@ -4056,6 +4072,115 @@ def test_concatenate(self):
             np.concatenate([x, y], axis=1),
         )
 
+    def test_view(self):
```
hertschuh, line 4081: Make this test backend independent, you don't need the `if`s, always use the string dtype which is supported across backends.

keras/src/ops/numpy_test.py
```diff
@@ -4056,6 +4072,115 @@ def test_concatenate(self):
...
+        )
+
+        # Test case 5: View int32 as int64 to a smaller bype size.
+        x = knp.array([[1, 2, 3, 4], [5, 6, 7, 8]], dtype="int32")
+        if backend.backend() == "tensorflow":
+            result = knp.view(x, dtype="int64")
+        elif backend.backend() == "jax":
+            # jax default doesn't support bit 64, need to enable it.
+            os.environ["JAX_ENABLE_X64"] = "1"
+            jax.config.update("jax_enable_x64", True)
```
hertschuh: You can't do that like that here because the config sticks and breaks other tests (that's why the JAX tests are failing). Don't worry about 64 bit values for now.

keras/src/ops/numpy_test.py
```diff
@@ -9171,6 +9296,48 @@ def test_angle(self, dtype):
...
+            return
+        x = knp.ones((2, 8), dtype=input_dtype)
+        x_jax = jnp.ones((2, 8), dtype=input_dtype)
+
+        try:
+            if backend.backend() == "torch":
+                keras_output = knp.view(x, getattr(torch, output_dtype))
+                symbolic_output = knp.View(
+                    getattr(torch, output_dtype)
+                ).symbolic_call(x)
```
hertschuh: You don't need this, the block below should work

keras/src/ops/numpy_test.py
```diff
@@ -9171,6 +9296,48 @@ def test_angle(self, dtype):
...
+            )
+            self.assertEqual(
+                standardize_dtype(symbolic_output.dtype),
+                standardize_dtype(expected_output.dtype),
+            )
+
+        except ValueError:
+            return  # Skip invalid view combinations
+        except tf.errors.InvalidArgumentError:
+            return  # Skip invalid tf bitcast combinations, e.g. int32 to bool
```
hertschuh: You shouldn't do it this way unfortunately. If all of the sudden all combinations start failing, the unit test would still pass.

Instead, you should have a specific list of combinations that you know don't work. You can either remove unsupported dtypes from the ALL_DTYPES list and/or remove pairs that don't work.

keras/src/backend/openvino/excluded_concrete_tests.txt
```diff
@@ -51,6 +51,8 @@ NumpyDtypeTest::test_trunc
 NumpyDtypeTest::test_unravel
 NumpyDtypeTest::test_var
 NumpyDtypeTest::test_vdot
+NumpyDtypeTest::test_view
+NumpyViewTest
```
hertschuh: Remove `NumpyViewTest`, it doesn't exist anymore.

keras/src/ops/numpy_test.py
```diff
@@ -4056,6 +4078,68 @@ def test_concatenate(self):
             np.concatenate([x, y], axis=1),
         )
 
+    def test_view(self):
+        input_array = knp.array([[1, 2, 3], [4, 5, 6]], dtype="int32")
+
+        # Test case 1: View with old dtype
+        result = knp.view(input_array, dtype="int32")
+        assert backend.standardize_dtype(result.dtype) == "int32"
```
hertschuh: We don't use Python asserts, we use the `unittest` ones: `self.assertEqual(backend.standardize_dtype(result.dtype), "int32")`

keras/src/ops/numpy_test.py
```diff
@@ -4056,6 +4078,68 @@ def test_concatenate(self):
...
+        self.assertAllClose(
```
hertschuh: This works, but let's make the test shorter and consistent with the other ones. What we do is that we compare against numpy so we don't have to put explicit expected values.

Also, you don't need `convert_to_tensor`. And we rarely put comments, tests are pretty self-explanatory.

The whole test can be:

```python
x = np.array([[1, 2, 3], [4, 5, 6]], dtype="int32")

self.assertEqual(backend.standardize_dtype(knp.view(x, dtype="int32")), "int32")
self.assertAllClose(knp.view(x, dtype="int32"), x.view("int32"))

self.assertEqual(backend.standardize_dtype(knp.view(x, dtype="float32")), "float32")
self.assertAllClose(knp.view(x, dtype="float32"), x.view("float32"))

self.assertEqual(backend.standardize_dtype(knp.view(x, dtype="int16")), "int16")
self.assertAllClose(knp.view(x, dtype="int16"), x.view("int16"))

    return

self.assertEqual(backend.standardize_dtype(knp.view(x, dtype="int64")), "int64")
self.assertAllClose(knp.view(x, dtype="int64"), x.view("int64"))
```

hertschuh (commented):
Last nitpick about the unit test.

keras/src/ops/numpy_test.py
wenyi-guo: Sg. Actually set input to be int16, so we can test view as a larger size int int32 for jax as well.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -998,6 +998,51 @@ def array(x, dtype=None):
...
+    new_itemsize = new_dtype.size
+
+    if list(x.shape)[-1] * old_itemsize % new_itemsize != 0:
+        raise ValueError(
+            f"Cannot view array of shape {x.shape} and dtype {old_dtype} "
+            f"as dtype {new_dtype} because the total number of bytes "
+            f"is not divisible by the new itemsize."
+        )
+
+    old_shape = list(shape_op(x))
```
hertschuh: `old_shape = list(shape_op(x))`

Move line 1031, that's the only time it's needed.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -998,6 +998,51 @@ def array(x, dtype=None):
...
+            raise ValueError(
+                f"Cannot view dtype. Last dimension size ({last_dim_size}) "
+                f"must be divisible by the ratio of new/old item sizes "
+                f"({ratio})."
+            )
+        intermediate_shape = old_shape[:-1] + [last_dim_size // ratio, ratio]
+
+        reshaped_tensor = tf.reshape(x, intermediate_shape)
+
+        return tf.bitcast(reshaped_tensor, new_dtype)
```
hertschuh, line 1041: nitpick: remove empty lines

keras/src/backend/jax/numpy.py
```diff
@@ -446,6 +446,10 @@ def array(x, dtype=None):
     return jnp.array(x, dtype=dtype)
 
 
+def view(x, dtype=None):
+    return x.view(dtype=dtype)
```
hertschuh, line 451: Add `x = convert_to_tensor(x)`, we should always do this

keras/src/ops/numpy.py
```diff
@@ -1124,6 +1124,69 @@ def array(x, dtype=None):
...
+    >>> x = keras.ops.array([1, 2, 3])
+    >>> x
+    array([1, 2, 3], dtype=int32)
+    >>> keras.ops.view(x, dtype="float32")
+    array([1.0e-45, 3.0e-45, 4.0e-45], dtype=float32)
+    """
+    if any_symbolic_tensors((x,)):
+        return View(dtype=dtype).symbolic_call(x)
+    r = backend.numpy.view(x, dtype=dtype)
+    return r
```
hertschuh: nitpick: `return backend.numpy.view(x, dtype=dtype)`

PR #21764, opened 2025-10-22
Document that `set_backend` requires re-importing keras.

hertschuh:
Both in the example and the documentation.

### DIFF
--- a/keras/src/utils/backend_utils.py
+++ b/keras/src/utils/backend_utils.py
@@ -125,7 +125,12 @@ def set_backend(backend):
     Example:
 
     ```python
+    import keras
+
     keras.config.set_backend("jax")
+
+    del keras
+    import keras
     ```
 
     ⚠️ WARNING ⚠️: Using this function is dangerous and should be done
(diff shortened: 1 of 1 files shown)


SamanehSaadat (approved):

keras/src/utils/backend_utils.py
```diff
@@ -125,7 +125,12 @@ def set_backend(backend):
...
     ```python
+    import keras
+
     keras.config.set_backend("jax")
+
+    del keras
+    import keras
     ```
 
     ⚠️ WARNING ⚠️: Using this function is dangerous and should be done
```
yufengg, line 136: Perhaps it would good to print/display some of this warning text when set_backend() is run, otherwise it looks like a silent failure (when a set_backend() is called but the backend doesn't actually change). 

If concerned about spamming text, perhaps there can be a flag to suppress the warning (but default to printing something, since this is such a "dangerous" function)

fchollet (approved):
In fact, it requires reimporting every single thing imported from Keras! And redefining any function that was using them I believe?

PR #21766, opened 2025-10-23
cleanup distillation  loss names

divyashreepathihalli:


keras/src/distillation/distiller.py
```diff
@@ -19,13 +19,13 @@ class Distiller(Model):
         teacher: A trained `keras.Model` that serves as the knowledge source.
             The teacher model is frozen during distillation.
         student: A `keras.Model` to be trained through distillation.
-        distillation_loss: List of distillation strategies to apply.
-            Can be a single strategy or a list of strategies like
+        distillation_losses: List of distillation losses to apply. Can be a
+            single distillation loss or a list of distillation losses like
             `LogitsDistillation`, `FeatureDistillation`, or custom distillation
```
fchollet: Use the full import paths, e.g. `keras.distillation.LogitsDistillation()`

keras/src/distillation/distiller.py
```diff
@@ -441,61 +459,79 @@ def compute_loss(
             teacher_features = self._extract_all_teacher_features(x)
             student_features = self._extract_all_student_features(x, y_pred)
 
-            # Apply strategies using pre-extracted features
-            for strategy, weight in zip(
-                self.distillation_loss, self.distillation_loss_weights
+            # Apply distillation losses using pre-extracted features
+            for distillation_loss_fn, weight in zip(
```
hertschuh: nitpick: just curious, why `distillation_loss_fn` here, and not `distillation_loss`?

divyashreepathihalli: looks like if i update it to distillation_loss there is a naming collision causing errors - distillation_loss is already used for the accumulated loss value.

PR #21767, opened 2025-10-23
Removes unnecessary try-except blocks and guard conditions

JyotinderSingh:
This change cleans up and removes unnecessary try-except blocks and guard conditions which reduce readability while not providing any real safety or improvements.


hertschuh:
Thanks for the cleanup!

PR #21775, opened 2025-10-24
Make confusion metrics compilable.

hertschuh:
By removing the use of `ops.nonzero` which returns an array of non-predetermined size.

Follow-up to https://github.com/keras-team/keras/pull/21765/files#r2462099871

Fixes https://github.com/keras-team/keras/issues/19376


hertschuh:
@danielenricocahall

Thanks!

danielenricocahall:
> @danielenricocahall
> 
> Thanks!

Ah I'm sorry! I was too gung ho with my fix, thank you for doing this!

PR #21776, opened 2025-10-24
update test after jax.config.jax_vjp3 is enabled

mattjj:


keras/src/quantizers/quantizers_test.py
```diff
@@ -544,10 +544,9 @@ def quantize_fn(x):
 
                 _, f_vjp = jax.vjp(quantize_fn, inputs)
 
-                # NOTE: When python version >= 3.10, the gradients are at
-                # `f_vjp.args[0].args[0][0]`. Otherwise, they are at
-                # `f_vjp.args[0].args[0][1]`.
-                if sys.version_info >= (3, 10):
+                if jax.config.jax_vjp3.value:
+                  input_gradients = f_vjp.opaque_residuals[0]
```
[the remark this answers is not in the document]
mattjj: Fixed!

keras/src/quantizers/quantizers_test.py
```diff
@@ -544,10 +544,9 @@ def quantize_fn(x):
 
                 _, f_vjp = jax.vjp(quantize_fn, inputs)
 
-                # NOTE: When python version >= 3.10, the gradients are at
-                # `f_vjp.args[0].args[0][0]`. Otherwise, they are at
-                # `f_vjp.args[0].args[0][1]`.
-                if sys.version_info >= (3, 10):
+                if jax.config.jax_vjp3:
```
hertschuh: So we use a pretty old version of JAX on CPU (0.5.0, I spent hours trying to find a combination that plays nice with both TensorFlow and Torch and failed).
Right now the test is failing with
```
FAILED keras/src/quantizers/quantizers_test.py::QuantizersTest::test_fake_quant_with_min_max_vars_wide_8bits_multi_channel - AttributeError: 'Config' object has no attribute 'jax_vjp3'
```
I guess it needs to be changed to:
`if getattr(jax.config, "jax_vjp3", False):`

mattdangerw: Went ahead and addressed this to myself to fix CI. Will pull in if all is green.

Issue #21617, opened 2025-08-27
Discrepancy in ops.correlate for tensorflow


wesselvannierop:
For the tensorflow backend, `keras.ops.correlate` does not match numpy outputs in mode='same' and mode='full'. The other backends and mode='valid' work fine.

```python
import os

os.environ["KERAS_BACKEND"] = "tensorflow"

import numpy as np
from keras import ops

x = np.array([1, 3, 5])
y = np.array([7, 9])

mode = "full"
print("Mode: ", mode)
print("numpy: ", np.correlate(x, y, mode=mode))
print("tensorflow: ", ops.convert_to_numpy(ops.correlate(x, y, mode=mode)))

mode = "same"
print("Mode: ", mode)
print("numpy: ", np.correlate(x, y, mode=mode))
print("tensorflow: ", ops.convert_to_numpy(ops.correlate(x, y, mode=mode)))

mode = "valid"
print("Mode: ", mode)
print("numpy: ", np.correlate(x, y, mode=mode))
print("tensorflow: ", ops.convert_to_numpy(ops.correlate(x, y, mode=mode)))
```

```bash
Mode:  full
numpy:  [ 9 34 66 35]
tensorflow:  [34. 66. 35.  0.]
Mode:  same
numpy:  [ 9 34 66]
tensorflow:  [34. 66. 35.]
Mode:  valid
numpy:  [34 66]
tensorflow:  [34. 66.]
```

wesselvannierop:
Besides, for jax, `jnp.correlate` supports complex values, while in torch and tensorflow it seems to cast to float.

I have an implementation that fixes both issues:

```python
def correlate(x, y, mode="full"):
    """
    Complex correlation via splitting real and imaginary parts.
    Equivalent to np.correlate(x, y, mode).

    NOTE: this function exists because tensorflow does not support complex correlation.
    NOTE: tensorflow also handles padding differently than numpy, so we manually pad the input.

    Args:
        x: np.ndarray (complex or real)
        y: np.ndarray (complex or real)
        mode: "full", "valid", or "same"
    """
    x = ops.convert_to_tensor(x)
    y = ops.convert_to_tensor(y)

    is_complex = "complex" in ops.dtype(x) or "complex" in ops.dtype(y)

    # Split into real and imaginary
    xr, xi = ops.real(x), ops.imag(x)
    yr, yi = ops.real(y), ops.imag(y)

    # Pad to do full correlation
    pad_left = ops.shape(y)[0] - 1
[...]

PR #21778, opened 2025-10-25
Fix bug with correlate for tensorflow

danielenricocahall:
Addresses https://github.com/keras-team/keras/issues/21617 by ensuring the "full" and "same" case are properly handled. For simplicity, "same" just reuses "full" and truncates. To validate, a unit test has been added which uses the original data from the issue.

### DIFF
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -3165,7 +3165,7 @@ def correlate(x1, x2, mode="valid"):
         full_corr = _full_corr(x1, x2)
         full_len = n + m - 1
         out_len = (
-            max(n, m)
+            max([n, m])
             if isinstance(n, int) and isinstance(m, int)
             else tf.maximum(n, m)
         )


keras/src/backend/tensorflow/numpy.py
```diff
@@ -3130,30 +3130,47 @@ def correlate(x1, x2, mode="valid"):
...
-            x1, paddings=[[tf.math.floor(x1_pad), tf.math.ceil(x1_pad)]]
-        )
-        x2 = tf.pad(
-            x2, paddings=[[tf.math.floor(x2_pad), tf.math.ceil(x2_pad)]]
+    def _pack(a, b):
+        # a: input [N] -> [1,N,1];
+        # b: filter [M] -> [M,1,1]
+        return (
+            tf.reshape(a, (1, tf.shape(a)[0], 1)),
+            tf.reshape(b, (tf.shape(b)[0], 1, 1)),
```
hertschuh: Please use `shape_op(a)` / `shape_op(b)` instead of `tf.shape`.

The reason is that `shape_op` returns ints when the shape is static, whereas `tf.shape` always returns a tensor, which causes all subsequent shapes to be dynamic.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -3130,30 +3130,47 @@ def correlate(x1, x2, mode="valid"):
...
         )
 
-        x1 = tf.reshape(x1, (1, full_len, 1))
-        x2 = tf.reshape(x2, (full_len, 1, 1))
+    def _full_corr(x1, x2):
+        """Compute 'full' correlation result (length = n + m - 1)."""
+        m = tf.shape(x2)[0]
```
hertschuh: Use `shape_op`.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -3130,30 +3130,47 @@ def correlate(x1, x2, mode="valid"):
...
+        pad = tf.maximum(m - 1, 0)
+        x1 = tf.pad(x1, [[pad, pad]])  # pad input with zeros
+        x1, x2 = _pack(x1, x2)
+        out = tf.nn.conv1d(x1, x2, stride=1, padding="VALID")
+        return tf.squeeze(out)
 
-        return tf.squeeze(tf.nn.conv1d(x1, x2, stride=1, padding="SAME"))
+    n = tf.shape(x1)[0]
+    m = tf.shape(x2)[0]
```
hertschuh: Use `shape_op`.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -3130,30 +3130,47 @@ def correlate(x1, x2, mode="valid"):
...
+            tf.reshape(a, (1, tf.shape(a)[0], 1)),
+            tf.reshape(b, (tf.shape(b)[0], 1, 1)),
         )
 
-        x1 = tf.reshape(x1, (1, full_len, 1))
-        x2 = tf.reshape(x2, (full_len, 1, 1))
+    def _full_corr(x1, x2):
+        """Compute 'full' correlation result (length = n + m - 1)."""
+        m = tf.shape(x2)[0]
+        pad = tf.maximum(m - 1, 0)
```
hertschuh: To perserve static shapes:
`pad = builtins.max(m - 1, 0) if isinstance(m, int) else tf.maximum(m - 1, 0)`

keras/src/backend/tensorflow/numpy.py
```diff
@@ -3130,30 +3130,47 @@ def correlate(x1, x2, mode="valid"):
...
+        x1 = tf.pad(x1, [[pad, pad]])  # pad input with zeros
+        x1, x2 = _pack(x1, x2)
+        out = tf.nn.conv1d(x1, x2, stride=1, padding="VALID")
+        return tf.squeeze(out)
```
hertschuh: I realize you didn't code that initially, but can we be specific about which dimensions to squeeze? My concern is if the length of the output is 1, we'll accidentally squeeze a dimension that we shouldn't.

Additionally, if we don't do that, it means that the rank of the output cannot always be statically determined if the outputs size is dynamic.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -3130,30 +3130,47 @@ def correlate(x1, x2, mode="valid"):
...
+        # the output is off by 1 compared to numpy, due to how
+        # tf handles centering
+        full_corr = _full_corr(x1, x2)
+        full_len = n + m - 1
+        out_len = tf.maximum(n, m)
+        start = (full_len - out_len) // 2
+        return tf.slice(full_corr, [start], [out_len])
+    elif mode == "valid":
+        x1, x2 = _pack(x1, x2)
+        return tf.squeeze(tf.nn.conv1d(x1, x2, stride=1, padding="VALID"))
```
hertschuh: I realize you didn't code that initially, but can we be specific about which dimensions to squeeze? My concern is if the length of the output is 1, we'll accidentally squeeze a dimension that we shouldn't.

Additionally, if we don't do that, it means that the rank of the output cannot always be statically determined if the outputs size is dynamic.

keras/src/ops/numpy_test.py
```diff
@@ -5133,6 +5133,26 @@ def test_correlate_different_size(self):
             knp.Correlate(mode="full")(x, y), np.correlate(x, y, mode="full")
         )
 
+    def test_correlate_bug(self):
+        # copied from https://github.com/keras-team/keras/issues/21617
```
hertschuh: Remove this comment.

keras/src/ops/numpy_test.py
```diff
@@ -5133,6 +5133,26 @@ def test_correlate_different_size(self):
             knp.Correlate(mode="full")(x, y), np.correlate(x, y, mode="full")
         )
 
+    def test_correlate_bug(self):
```
hertschuh: Since these cover a very similar scenario as `test_correlate_different_size`, the test above, let's replace `test_correlate_different_size` with this one. So remove `test_correlate_different_size` and rename this one `test_correlate_different_size`.

hertschuh (commented):
@danielenricocahall 

Thank you for fixing this! This was a tricky one.

keras/src/backend/tensorflow/numpy.py
danielenricocahall: Thank you! done

danielenricocahall:
Hey @hertschuh thank you for the feedback! I think I've addressed all comments - to confirm, for `squeeze`, we wanted to be explicit about only dropping the first and last (batch and channel) axes to ensure we don't squeeze the correlation result if it winds up being of size 1?

hertschuh:
> Hey @hertschuh thank you for the feedback! I think I've addressed all comments - to confirm, for `squeeze`, we wanted to be explicit about only dropping the first and last (batch and channel) axes to ensure we don't squeeze the correlation result if it winds up being of size 1?

Yes, exactly!

hertschuh (approved):

keras/src/backend/tensorflow/numpy.py
```diff
@@ -3130,31 +3130,54 @@ def correlate(x1, x2, mode="valid"):
...
+        return _full_corr(x1, x2)
+    elif mode == "same":
+        # unfortunately we can't leverage 'SAME' padding directly like
+        # we can with "valid"
+        # it works fine for odd-length filters, but for even-length filters
+        # the output is off by 1 compared to numpy, due to how
+        # tf handles centering
+        full_corr = _full_corr(x1, x2)
+        full_len = n + m - 1
+        out_len = tf.maximum(n, m)
```
hertschuh: Oh I missed this one: `out_len = max(n, m) if isinstance(n, int) and isinstance(m, int) else tf.maximum(n, m)`

hertschuh (approved):

keras/src/backend/tensorflow/numpy.py
```diff
@@ -3130,30 +3130,57 @@ def correlate(x1, x2, mode="valid"):
...
+        # we can with "valid"
+        # it works fine for odd-length filters, but for even-length filters
+        # the output is off by 1 compared to numpy, due to how
+        # tf handles centering
+        full_corr = _full_corr(x1, x2)
+        full_len = n + m - 1
+        out_len = (
+            max(n, m)
+            if isinstance(n, int) and isinstance(m, int)
+            else tf.maximum(n, m)
```
hertschuh, line 3170: `tf.maximum([n, m])`

danielenricocahall: ah yes it's `max([n, m])` since it's not the builtin `max`, it's the one from the module. Fixed!

Issue #21781, opened 2025-10-26
Bug in `keras.ops.quantile` when running with tensorflow graph execution (`tf.function`)


SamKnightGit:
OS: **Ubuntu 24.04.2 LTS**
Python Version: **3.12.12**
Keras Version: **3.11.3**
TF Version: **2.20.0**

### Description

When calling `keras.ops.quantile` with arguments that results in output dimension > 1, the call results in a `ValueError` instead of completing successfully.

### Reproduction

```python
import os
os.environ["KERAS_BACKEND"] = "tensorflow"

import numpy as np
import tensorflow as tf
from keras import ops

@tf.function
def run_quantile():
    x = np.array([[1, 2, 3], [4, 5, 6]])
    q = [0.5]
    return ops.quantile(x, q, axis=1)

result = run_quantile()
expected_output = np.array([[2, 5]])
assert(np.allclose(expected_output, result))
```

### Expected Result

Assert passes, program terminates without exceptions.

### Actual Result

Exception is raised in the call to `ops.quantile`:
`ValueError: Tried to convert 'perm' to a tensor and failed. Error: Expected values deque([1, 0]) to be a dense tensor with shape [], but got shape [2].`

### Root Cause

When in graph mode, python arguments are automatically converted to Tensorflow tensors. The [final transpose call](https://github.com/keras-team/keras/blob/master/keras/src/backend/tensorflow/numpy.py#L2304) in the internal `_quantile` function in `keras.ops` passes a `deque` to the perm argument. This fails to be converted to a tensor causing the crash.

PR #21782, opened 2025-10-26
Fix: `keras.ops.quantile` works with tf graph execution

SamKnightGit:
A `deque` failing to be converted to a tensor in the transpose call in `keras.ops.quantile` caused errors when running in tf graph contexts. By constructing a list from the deque before passing it to the `transpose` call we avoid this error. Test added to cover this specific case.

Fixes https://github.com/keras-team/keras/issues/21781


fchollet (approved):
Thanks for the fix!

PR #21783, opened 2025-10-26
Fix #21105: JAX Backend - Custom Gradient Variable Capture Issue in keras.ops.custom_gradient

### LINKED ISSUE #21105: custom_gradient not working with JAX backend
When trying to use the custom_gradient decorator with a JAX backend as shown in the example located here
[custom_gradient_example](https://keras.io/api/ops/core/#customgradient-function), I get an error  message saying `TypeError: 'NoneType' object is not callable`.

It works without a problem on the TensorFlow backend, but not JAX. The behaviour is seen both on keras v3.9.0 and on keras-nightly.

Mayankvlog:
Resolved issue #21105 where keras.ops.custom_gradient incorrectly captured Variable objects instead of their values when using the JAX backend. This caused gradient computation failures in custom quantization layers. The problem occurred when passing keras.Variable instances directly to functions decorated with @ops.custom_gradient, where the gradient function (roundpass_grad) captured the variable object rather than its underlying value.

Root Cause: JAX backend's custom gradient implementation didn't properly extract values from Variable objects before passing them to gradient functions.

Solution: Modified keras/src/backend/jax/core.py to automatically extract .value from Variable instances within the custom gradient decorator, eliminating the need for manual .value calls in user code. Added comprehensive tests in keras/src/ops/core_test.py::CoreOpsCorrectnessTest::test_custom_gradient_with_variable to prevent regression. This fix ensures seamless Variable handling across all backends without requiring workarounds.

### DIFF
--- a/keras/src/applications/efficientnet_v2_jit_test.py
+++ b/keras/src/applications/efficientnet_v2_jit_test.py
@@ -1,7 +1,6 @@
 """Test for Issue #21647: jit_compile=True with EfficientNetV2 on torch
 backend."""
 
-
 import numpy as np
 import pytest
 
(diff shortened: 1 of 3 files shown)


hertschuh:
@Mayankvlog 

This PR is mixing at least 3 different unrelated things. Can you split this PR? Thanks!

PR #21784, opened 2025-10-26
Fix: convert deque to list before tf.transpose in keras.ops.quantile 

### LINKED ISSUE #21781: Bug in `keras.ops.quantile` when running with tensorflow graph execution (`tf.function`)

Junead04:
Summary
Fixed a ValueError in keras.ops.quantile when running under TensorFlow graph execution (@tf.function).
The issue was caused by passing a collections.deque object as the perm argument to tf.transpose.

Fix
Fixes #21781
Convert perm to a list before passing it to TensorFlow:
result = tf.transpose(result, perm=list(perm))

Test
Added a regression test verifying correct behavior under tf.function.
CC: @keras-team


keerthanakadiri:
Hi @Junead04, Can you please sign the CLA? Thank you!

keras/api/ops/core/quantile_test.py
```diff
@@ -0,0 +1,14 @@
+import numpy as np
+import tensorflow as tf
+from keras import ops
+
+def test_quantile_graph_mode():
```
hertschuh: This test already exists: https://github.com/keras-team/keras/blob/d662ac7db416d99511fdfb3a1f823bf797af260d/keras/src/ops/numpy_test.py#L3347

Can you remove this file and modify the existing test to cover the case that was not working?

hertschuh (commented):
Thanks for the fix!

hertschuh:
You'll also need to accept the CLA.

Junead04:
> Okk I will sign into the CLA.This is my first contribution so I don't know about this.Thank you for guiding me @keerthanakadiri @hertschuh

hertschuh:
> > Okk I will sign into the CLA.This is my first contribution so I don't know about this.Thank you for guiding me @keerthanakadiri @hertschuh

@Junead04 

In the failing checks below, click on "cla/google", it will take you to a page that explains things and there will be another link to click to accept the CLA.

Junead04:
I sign in CLA sorry for late but when I rescan this PR it is showing error.

hertschuh:
Can you address the PR comments and run the code formatter? Thanks!

hertschuh:
Thanks!

PR #21791, opened 2025-10-28
Adding get_device_count function to the distribution_lib

buildwithsuhana:
This PR introduces a new utility function, keras.distribution.get_device_count, to the Keras distribution API.
This function allows a user to query the total number of available devices (like GPUs, TPUs, or CPUs) that Keras can see.

### DIFF
--- a/keras/src/distribution/distribution_lib.py
+++ b/keras/src/distribution/distribution_lib.py
@@ -41,15 +41,14 @@ def list_devices(device_type=None):
 
 @keras_export("keras.distribution.get_device_count")
 def get_device_count(device_type=None):
-    """Returns the total number of available devices.
-
+    """Returns the number of available JAX devices.
     Args:
-        device_type: Optional device type to count (e.g., "cpu",
-             "gpu", "tpu"). If `None`, it counts all available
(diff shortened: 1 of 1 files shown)


innat:
@buildwithsuhana 
Thanks for the PR. It would be useful. There are some other utiliy (similar to this), would be great to have. https://github.com/keras-team/keras/discussions/20744

keras/src/distribution/distribution_lib.py
```diff
@@ -39,6 +39,21 @@ def list_devices(device_type=None):
...
 
 
+@keras_export("keras.distribution.get_device_count")
+def get_device_count(device_type=None):
+    """Returns the total number of available devices.
+
+    Args:
+        device_type: Optional device type to count (e.g., "cpu",
+             "gpu", "tpu"). If `None`, it counts all available
+             devices.
```
hertschuh: Oh, copy the one from the jax implementation. This one is incorrect.

buildwithsuhana: Done

hertschuh (approved):

PR #21792, opened 2025-10-28
Adding Tensor_layout for Tensor parallelism for Autosharding

buildwithsuhana:
This change introduces core building blocks for tensor parallelism by adding two key components. 

First, it adds crucial collective operations, all_reduce and all_gather, to the JAX backend. These allow multiple devices to synchronize data by summing tensors (like gradients) or gathering individual slices back into a full tensor. Second, it adds the high-level tensor sharding logic (split_tensor_for_parallelism), which uses ops.array_split to intelligently slice large tensors, even unevenly, for distribution across devices. New tests confirm this new parallel logic, including the uneven splitting, works as expected.

The tests on this PR will pass after the PR https://github.com/keras-team/keras/pull/21697 gets merged

### DIFF
--- a/keras/src/distribution/tensor_parallel/coordinated_optimizer.py
+++ b/keras/src/distribution/tensor_parallel/coordinated_optimizer.py
@@ -88,7 +88,9 @@ class TensorParallelOptimizer(optimizers.Optimizer):
         self.base_optimizer.build(variables)
 
         if variables:
-            grads = [ops.zeros_like(ops.convert_to_tensor(v)) for v in variables]
+            grads = [
+                ops.zeros_like(ops.convert_to_tensor(v)) for v in variables
+            ]
             self.base_optimizer.apply_gradients(zip(grads, variables))
 
         if self.shard_optimizer_states:
--- a/keras/src/distribution/tensor_parallel/coordinated_optimizer_test.py
+++ b/keras/src/distribution/tensor_parallel/coordinated_optimizer_test.py
@@ -86,7 +86,10 @@ class TensorParallelOptimizerTest(testing.TestCase):
         model = self._get_simple_model()
         model.build((None, 10))
 
-        grads = [ops.zeros_like(ops.convert_to_tensor(v)) for v in model.trainable_variables]
+        grads = [
+            ops.zeros_like(ops.convert_to_tensor(v))
+            for v in model.trainable_variables
+        ]
         grads_and_vars = list(zip(grads, model.trainable_variables))
 
         optimizer.apply_gradients(grads_and_vars)


hertschuh:
Can you rebase to make the tests pass?

keras/src/distribution/tensor_parallel/tensor_layout.py
```diff
@@ -0,0 +1,34 @@
...
+        index: The index of the device/shard to return (e.g., 0, 1, 2...).
+        device_count: The total number of parallel devices or splits.
+        dim: The dimension along which to split the tensor. If -1, the
+            last dimension is used.
+
+    Returns:
+        A tensor slice corresponding to the given `index`.
+    """
+    if dim == -1:
+        split_dim = ops.ndim(tensor) - 1
```
hertschuh: Is `-1` the only possible case? Usually, any negative number is possible and then you do:

```python
split_dim = ops.ndim(tensor) + dim
```

keras/src/backend/jax/core_test.py
```diff
@@ -66,3 +69,78 @@ def test_keras_variable_nnx_split_merge_sync(self):
...
+
+
+@pytest.mark.skipif(
+    backend.backend() != "jax",
+    reason="JAX backend specific test for collective operations.",
+)
+@pytest.mark.skipif(
+    jax.local_device_count() < 2,
+    reason="Requires multiple local devices for testing.",
+)
```
hertschuh: These tests will never run because we don't have a setup with 2 accelerators right now (we have 1 and 8).

You should move these tests to `/Users/fhertschuh/Documents/keras/keras/src/backend/jax/distribution_lib_test.py` in which we simulate 8 devices.

keras/src/backend/jax/core.py
```diff
@@ -529,6 +530,61 @@ def remat(f):
     return jax.checkpoint(f)
 
 
+def all_reduce(x, op="sum", axis_name="model"):
```
hertschuh: Now that I see this, it's a bit weird to have these ops in `core` because all backends must implement all ops in `core` because they are core ops.

Also, this is distribution related. Can you move them to `keras/src/backend/jax/distribution_lib.py` at least for now.

Maybe we'll need a specific namespace for distribution ops, but since they're not exported, I think `distribution_lib` is fine for now.

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,262 @@
...
+        full_name = f"{prefix}.{name}" if prefix else name
+
+        _apply_layer_sharding_rules(
+            current_layer, full_name, device_count, state_rules, output_rules
+        )
+
+        children_to_add = []
+
+        if hasattr(current_layer, "layers") and current_layer.layers:
+            for sub_layer in current_layer.layers:
```
hertschuh: Again `model.layers` will return the full flat list of all the layers in the model, even if they're deeply nested.

So:
- `processed_layers` is not needed
- `stack` is not needed
- `prefix` is not needed, it will always be `""`, you will never get a deeply nested path. If you wanted that, you'll need to code this very differently
- `children_to_add` is not needed
- line 234-260 are not needed
- lines 224-232 are probably not needed

The code could be a lot shorter. This whole function `get_default_config` can be:

```python
def get_default_config(module, device_ids):
    device_count = len(device_ids)
    state_rules = {}
    output_rules = {}
    for layer in module.layers:
        _apply_layer_sharding_rules(
            layer, layer.name, device_count, state_rules, output_rules
        )
    return LayoutMap(state_rules=state_rules, output_rules=output_rules)
```

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,262 @@
...
+                children_to_add.append((sub_layer, full_name))
+
+        for specific_attr in [
+            "token_embedding",
+            "embeddings",
+            "position_embedding",
+        ]:
```
hertschuh: I don't understand why this needs a special case, can you explain?

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,262 @@
...
+                        clean_name = weight.name.split("/")[-1].split(":")[0]
+                        state_rules[f"{full_name}.{clean_name}"] = _split_rule(
+                            device_count, dim=1
+                        )
+
+            output_rules[f"{full_name}"] = {0: "no_comm"}
+
+
+def get_default_config(module, device_ids):
+    """Generates a default tensor parallelism configuration for a Keras model.
```
hertschuh: We don't use the term `module` in Keras. We have layers and models, which extend layers. I think you should just can it `model`, unless you do need to call it on layers, in which case you can call it `layer`.

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,262 @@
...
+
+    Args:
+        device_count: The total number of devices available for parallelism.
+        dim: The dimension of the tensor to split.
+
+    Returns:
+        callable: A lambda function accepting (tensor, index) that returns the
+        sharded layout.
+    """
+    return lambda x, index: _split_fn_internal(x, index, device_count, dim=dim)
```
hertschuh: I think you should just declare this `_apply_layer_sharding_rules`.

I would do it with a `function.partial` for 2 reasons:
- the code would be much shorter (no need to document this)
- you can bind `device_count` once for all, this way, you don't have to pass it to `split_rule`, which would make each call shorter and would allow them to be on 1 line instead of 3 lines, which will make them way more readable:

```python
def _apply_layer_sharding_rules(...):

    def split_rule(dim):
        return functools.partial(split_tensor_for_parallelism, device_count=device_count, dim=dim)

    ...
    state_rules[f"{full_name}.kernel"] = split_rule(dim=1)
```

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,262 @@
...
+    Returns:
+        str: One of 'up_projection', 'down_projection', or 'dense'.
+    """
+    if not isinstance(layer, layers.Dense):
+        return "dense"
+
+    input_dim = None
+    output_dim = None
+
+    if hasattr(layer, "kernel") and layer.kernel is not None:
```
hertschuh: The "real" kernel is `layer._kernel` for Dense and EinsumDense layers in Keras.

`kernel` is a property that may modify the kernel before returning it (for LoRa and quantization).

It may not matter for most of the your code, but it will matter if you want the variable and not just the value of the kernel (for instance to apply sharding to the variable itself).

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,262 @@
...
+
+        if mlp_type == "up_projection":
+            state_rules[f"{full_name}.kernel"] = _split_rule(
+                device_count, dim=1
+            )
+            if layer.use_bias:
+                state_rules[f"{full_name}.bias"] = _split_rule(
+                    device_count, dim=0
+                )
+            output_rules[f"{full_name}"] = {0: "gather"}
```
hertschuh: The `state_rules` are actual Python function.

The `output_rules` are strings, which I assume are later parsed to call the corresponding function (gather or reduce).

Why not be consistent and use either functions everywhere or strings everywhere?

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,262 @@
...
+                    ]:
+                        if getattr(layer, attr_candidate, None) is weight:
+                            state_rules[f"{full_name}.{attr_candidate}"] = (
+                                _split_rule(device_count, dim=1)
+                            )
+                            key_found = True
+                            break
+
+                    if not key_found:
+                        clean_name = weight.name.split("/")[-1].split(":")[0]
```
hertschuh: When do variable names contain ":"?

To me this was a legacy thing in Keras 2 that we don't have in Keras 3.

What about "/"? It should be in the path, not the name.

keras/src/distribution/tensor_parallel/autoconfig.py
buildwithsuhana: Done

keras/src/distribution/tensor_parallel/autoconfig.py
buildwithsuhana: Fixed

keras/src/distribution/tensor_parallel/autoconfig.py
buildwithsuhana: Modified

keras/src/distribution/tensor_parallel/autoconfig.py
buildwithsuhana: removed

keras/src/backend/jax/core.py
buildwithsuhana: Done

keras/src/distribution/tensor_parallel/tensor_layout.py
buildwithsuhana: Done

keras/src/distribution/tensor_parallel/autoconfig.py
buildwithsuhana: Actually I found that model.layers is shallow, it only returns immediate children, not a recursive flat list. For deeply nested models like OPT125m as I was testing, the simplified loop stops at the top level and never reaches the leaf layers where the weights actually live.

keras/src/backend/jax/distribution_lib.py
```diff
@@ -212,6 +213,52 @@ def process_id():
...
+        axis_name: The name of the mesh axis to reduce over.
+
+    Returns:
+        The reduced tensor.
+    """
+    if op == "sum":
+        return lax.psum(x, axis_name=axis_name)
+    elif op == "mean":
+        sum_val = lax.psum(x, axis_name=axis_name)
+        axis_size = lax.psum(1, axis_name=axis_name)
```
hertschuh: Why not `lax.pmean`? What's the difference?

keras/src/backend/jax/distribution_lib_test.py
```diff
@@ -441,6 +441,50 @@ def test_distribute_data_input(self):
         for shard in result.addressable_shards:
             self.assertEqual(shard.data.shape, (3, 4))
 
+    def test_all_reduce(self):
+        devices = jax.devices()
+        num_devices = len(devices)
+        input_data = np.ones((num_devices, 2), dtype="float32")
```
hertschuh: `input_data` needs to to sharded across the devices to make this test valid.

keras/src/backend/jax/distribution_lib_test.py
```diff
@@ -441,6 +441,50 @@ def test_distribute_data_input(self):
...
+
+        def sum_fn(x):
+            return backend_dlib.all_reduce(x, op="sum", axis_name="batch")
+
+        result_sum = jax.pmap(sum_fn, axis_name="batch")(input_data)
```
hertschuh: Shouldn't the `pmap` be part of the `all_reduce` implementation?

Same for "mean".

keras/src/backend/jax/distribution_lib_test.py
```diff
@@ -441,6 +441,50 @@ def test_distribute_data_input(self):
...
+            ValueError, "Unsupported reduction operation"
+        ):
+            backend_dlib.all_reduce(input_data[0], op="max", axis_name="batch")
+
+    def test_all_gather(self):
+        devices = jax.devices()
+        num_devices = len(devices)
+
+        input_data = np.arange(num_devices, dtype="float32").reshape(
+            num_devices, 1, 1
```
hertschuh: `input_data` needs to to sharded across the devices to make this test valid.

keras/src/backend/jax/distribution_lib_test.py
```diff
@@ -441,6 +441,50 @@ def test_distribute_data_input(self):
...
+        )
+
+        def gather_fn(x):
+            return backend_dlib.all_gather(x, axis=0, axis_name="batch")
+
+        results = jax.pmap(gather_fn, axis_name="batch")(input_data)
```
hertschuh: Shouldn't the `pmap` be part of the `all_gather` implementation?

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,241 @@
...
+        kernel_shape = layer.kernel.shape
+        if len(kernel_shape) == 2:
+            input_dim = kernel_shape[0]
+            output_dim = kernel_shape[1]
+
+    if input_dim is None or output_dim is None:
+        if hasattr(layer, "units"):
+            output_dim = layer.units
+        else:
+            return "dense"
```
hertschuh: Shouldn't this be done only if `output_dim` is None? (as opposed to if either `input_dim` or `output_dim` is None)

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,241 @@
...
+
+        if (
+            hasattr(layer, "input_shape")
+            and layer.input_shape
+            and len(layer.input_shape) > 1
+        ):
+            input_dim = layer.input_shape[-1]
+        else:
+            return "dense"
```
hertschuh: Shouldn't this be done only if `input_dim` is None? (as opposed to if either `input_dim` or `output_dim` is None)

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,241 @@
...
+    return distribution_lib.all_reduce(x, op="sum", axis_name="model")
+
+
+def _gather(x, axis):
+    """Gathers the input tensor across the model axis along the given axis."""
+    return distribution_lib.all_gather(x, axis=axis, axis_name="model")
+
+
+def _apply_layer_sharding_rules(
+    layer, full_name, device_count, state_rules, output_rules
```
hertschuh: Just call it `name` instead of `full_name`.

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,241 @@
...
+        return functools.partial(_gather, axis=axis)
+
+    if isinstance(layer, layers.Dense):
+        mlp_type = analyze_dense_layer(layer)
+
+        if mlp_type == "up_projection":
+            state_rules[f"{full_name}.kernel"] = split_rule(dim=1)
+            if layer.use_bias:
+                state_rules[f"{full_name}.bias"] = split_rule(dim=0)
+            output_rules[f"{full_name}"] = {0: gather_rule(axis=-1)}
```
hertschuh: Why `{0: gather_rule(axis=-1)}`? What does the `0` represent?

All of them use `0` always, so why not just do `output_rules[f"{full_name}"] = gather_rule(axis=-1)`?

keras/src/distribution/tensor_parallel/autoconfig.py
hertschuh: Oh sorry, instead of using `model.layers`, which is indeed not recursive, use `model._flatten_layers()` which is recursive. Then you can remove this function.

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,241 @@
...
+        return "dense"
+
+    input_dim = None
+    output_dim = None
+
+    if hasattr(layer, "_kernel") and layer._kernel is not None:
+        kernel_shape = layer._kernel.shape
+        if len(kernel_shape) == 2:
+            input_dim = kernel_shape[0]
+            output_dim = kernel_shape[1]
```
hertschuh: If it's a `keras.layers.Dense` layer, this should always work.

Do we actually need anything between lines 40-63?

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,241 @@
...
+        - Contraction: Input dimension > (Output dimension * 1.5)
+
+    Args:
+        layer (keras.layers.Layer): The layer instance to analyze.
+
+    Returns:
+        str: One of 'up_projection', 'down_projection', or 'dense'.
+    """
+    if not isinstance(layer, layers.Dense):
+        return "dense"
```
hertschuh: This condition is impossible because of line 111.

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,241 @@
...
+            state_rules[f"{full_name}.kernel"] = split_rule(dim=1)
+            if hasattr(layer, "bias") and layer.bias is not None:
+                state_rules[f"{full_name}.bias"] = split_rule(dim=0)
+            output_rules[f"{full_name}"] = {0: gather_rule(axis=-1)}
+
+    elif (
+        isinstance(layer, (layers.Embedding,))
+        or "Embedding" in layer.__class__.__name__
+    ):
+        if hasattr(layer, "weights"):
```
hertschuh: `weights` is a property of `Layer`, it cannot be missing, remove this `if`.

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,241 @@
...
+                                split_rule(dim=1)
+                            )
+                            key_found = True
+                            break
+
+                    if not key_found:
+                        clean_name = weight.name.split("/")[-1]
+                        state_rules[f"{full_name}.{clean_name}"] = split_rule(
+                            dim=1
+                        )
```
hertschuh: I'm a bit confused about this. So if it's an embedding, the rule is `split_rule(dim=1)`. Great.

Why do you need all the extra code? Why not do:

```python
state_rules[f"{full_name}.embedding"] = split_rule(dim=1)
```

And remove lines 145-164.

The `{attr_candidate}` or the `{clean_name}` is not reversible. You cannot find the weight back from the name because you removed some stuff.

Also, do you need to keep a reference to the weights in general? Like the kernel for dense?

keras/src/distribution/tensor_parallel/coordinated_optimizer.py
```diff
@@ -0,0 +1,518 @@
...
+                sharding_dim = 0
+                if self.tensor_parallel_config:
+                    norm_param_name = found_param.path.replace("/", ".")
+                    for (
+                        p,
+                        a,
+                    ) in self.tensor_parallel_config.state_rules.items():
+                        if re.search(p, norm_param_name) and hasattr(a, "dim"):
+                            sharding_dim = a.dim
+                            break
```
hertschuh: This is not safe, you're matching substrings.

For instance if `p` is `"dense.kernel"` and `norm_param_name` is `"einsum_dense.kernel"`, it will match, even though you got the wrong layer.

You want an exact match and manipulating names or paths just makes it harder. You should just identify the variables themselves directly. So you want to key your `state_rules` by `id(variable)`. For instance, in `autoconfig`, you just do `state_rules[id(layer._kernel)] = split_rule(dim=0)`. This will tremendously simplify this code and the code in `autoconfig`.

Optimizers already have a way to map optimizer variables to model variables, you don't need to recreate a different way to do this mapping.

keras/src/distribution/tensor_parallel/autoconfig.py
```diff
@@ -0,0 +1,195 @@
...
+    Returns:
+        The gathered tensor, concatenated along the specified axis.
+    """
+    return distribution_lib.all_gather(x, axis=axis, axis_name="model")
+
+
+def _get_layer_path(layer):
+    """Retrieves the unique hierarchical path of a layer.
+
+    This utilizes `layer.path` (available in Keras 3+) which provides a
```
hertschuh: This is Keras 3+, no need for a fallback.

Just remove `_get_layer_path` and use `layer.path` directly everywhere. It's guaranteed to work.

keras/src/backend/jax/distribution_lib_test.py
```diff
@@ -441,6 +441,51 @@ def test_distribute_data_input(self):
         for shard in result.addressable_shards:
             self.assertEqual(shard.data.shape, (3, 4))
 
+    def test_all_reduce(self):
+        devices = jax.devices()
+        num_devices = len(devices)
+        mesh = jax.sharding.Mesh(np.array(devices), axis_names=("batch",))
```
hertschuh, line 447: Note: as of now, this will always be running with a single device so it won't prove much. I'm hoping to have a way to fix this though.

keras/src/distribution/tensor_parallel/tensor_layout.py
```diff
@@ -0,0 +1,34 @@
...
+        tensor: The full tensor to be sharded.
+        index: The index of the device/shard to return (e.g., 0, 1, 2...).
+        device_count: The total number of parallel devices or splits.
+        dim: The dimension along which to split the tensor. Supports negative
+            indexing.
+
+    Returns:
+        A tensor slice corresponding to the given `index`.
+    """
+    if dim < 0:
```
hertschuh: please use:

```python
dim = canonicalize_axis(axis, ops.ndim(tensor))
```

You'll have to import `canonicalize_axis` from `keras.src.backend.common.backend_utils`.

keras/src/distribution/tensor_parallel/coordinated_optimizer.py
```diff
@@ -0,0 +1,542 @@
...
+            variable: The variable to update.
+            *args: Additional arguments for the update.
+            **kwargs: Additional keyword arguments for the update.
+        """
+        if hasattr(self.base_optimizer, "update_step"):
+            return self.base_optimizer.update_step(
+                gradient, variable, *args, **kwargs
+            )
+
+        return super().update_step(gradient, variable, *args, **kwargs)
```
hertschuh: Every optimizer has an `update_step`, otherwise it's not a valid optimizer. So this can be simplified to:

```python
return self.base_optimizer.update_step(gradient, variable, *args, **kwargs)
```

keras/src/distribution/tensor_parallel/coordinated_optimizer.py
```diff
@@ -0,0 +1,336 @@
+import numpy as np
+
+from keras.src import backend
+from keras.src import ops
+from keras.src import optimizers
+from keras.src.saving import serialization_lib
+
+
+class TensorParallelOptimizer(optimizers.Optimizer):
```
hertschuh: This class is similar to `LossScaleOptimizer` in that it wraps another optimizer. So I'll be referring to `LossScaleOptimizer` in a few comments.

https://github.com/keras-team/keras/blob/master/keras/src/optimizers/loss_scale_optimizer.py

keras/src/distribution/tensor_parallel/coordinated_optimizer.py
```diff
@@ -0,0 +1,336 @@
...
+        base_optimizer: The underlying Keras optimizer being wrapped.
+        device_count: Total number of accelerator devices (shards).
+        shard_optimizer_states: Whether to enable partitioning of states.
+        tensor_parallel_config: Configuration object defining how specific
+            variables should be sharded.
+    """
+
+    def __init__(
+        self,
+        base_optimizer,
```
hertschuh: Rename `base_optimizer` to `inner_optimizer`.

keras/src/distribution/tensor_parallel/coordinated_optimizer.py
```diff
@@ -0,0 +1,336 @@
...
+                mapping specific variables to axes.
+            name: String, name of the optimizer instance.
+            **kwargs: Additional arguments passed to the base optimizer.
+        """
+        if isinstance(base_optimizer, str):
+            base_optimizer = optimizers.get(base_optimizer)
+        elif isinstance(base_optimizer, dict):
+            base_optimizer = serialization_lib.deserialize_keras_object(
+                base_optimizer
+            )
```
hertschuh: The proper way to do this is to add a `from_config` method instead. See `LossScaleOptimizer`.

keras/src/distribution/tensor_parallel/coordinated_optimizer.py
```diff
@@ -0,0 +1,336 @@
...
+
+        lr = getattr(
+            base_optimizer, "_learning_rate", base_optimizer.learning_rate
+        )
+        if hasattr(lr, "numpy") and not callable(lr):
+            kwargs["learning_rate"] = float(ops.convert_to_numpy(lr))
+        else:
+            kwargs["learning_rate"] = lr
```
hertschuh: Remove everything related to learning rate here. Instead, like `LossScaleOptimizer`:
- pass `learning_rate=0.0` to `super().__init__`
- add:
```python
    @property
    def learning_rate(self):
        return self.inner_optimizer.learning_rate

    @learning_rate.setter
    def learning_rate(self, learning_rate):
        self.inner_optimizer.learning_rate = learning_rate
```

If you keep the learning rate in 2 places, you run the risk of advancing the schedule twice as fast. During deserialization, it will create 2 instances of the schedule etc.

keras/src/distribution/tensor_parallel/coordinated_optimizer.py
```diff
@@ -0,0 +1,336 @@
...
+            return
+
+        self._model_variables = variables
+        self.base_optimizer.build(variables)
+
+        if variables:
+            grads = [
+                ops.zeros_like(ops.convert_to_tensor(v)) for v in variables
+            ]
+            self.base_optimizer.apply_gradients(zip(grads, variables))
```
hertschuh: Why `applient_gradients` here? What does this do? The result is ignored.

keras/src/distribution/tensor_parallel/coordinated_optimizer.py
```diff
@@ -0,0 +1,336 @@
...
+
+        if self.shard_optimizer_states:
+            self._initialize_sharded_states()
+
+        super().build(variables)
+
+    def _initialize_sharded_states(self):
```
hertschuh: So there is already logic to apply the sharding of optimizer variable taken from the model variable. This is already done for all optimizers.

https://github.com/keras-team/keras/blob/master/keras/src/optimizers/base_optimizer.py#L324

All of this can be removed as far as I understand. This means `partition_state` and `get_sharding_dim` can also be removed.

keras/src/distribution/tensor_parallel/coordinated_optimizer.py
```diff
@@ -0,0 +1,336 @@
...
+                "`shard_models` is required for sharded gradients."
+            )
+
+        synced_grads_and_vars = self._synchronize_gradients(grads_and_vars)
+
+        for i in range(self.device_count):
+            shard_opt = shard_models[i].optimizer.base_optimizer
+            self._transfer_state(shard_opt, shard_idx=i, direction="to_local")
+            shard_opt.apply_gradients(synced_grads_and_vars[i])
+            self._transfer_state(shard_opt, shard_idx=i, direction="to_global")
```
hertschuh: This will cause the base_optimizer to do the same thing N times. There is no way to `apply_gradients` just on a subslice of gradients.

But good news is that you don't need to. The JAX magic will automatically do it across all devices.

keras/src/distribution/tensor_parallel/coordinated_optimizer.py
```diff
@@ -0,0 +1,336 @@
...
+
+    def _synchronize_gradients(self, gradients_and_vars):
+        """Averages gradients for variables that are not sharded via
+           Tensor Parallelism.
```
hertschuh: This is already done by the Keras distribution. As long as TensorParallel is a Keras distribution, there is nothing to do for this.

hertschuh (commented):
My analysis is that `CoordinateOptimizer` is not needed.

There are already a number of mechanisms in `BaseOptimizer` and `Trainer` to handle sharded variables, the sharding of optimizer variables, sharded gradients etc.

As long as the `Distribution` subclass is implemented correctly, things should just work.

Note that I wrote the comments in the code before I came to that conclusion.

hertschuh:
@buildwithsuhana 

I'm closing this old PR as I don't think this will be done this way. Feel free to reopen when the new implementation is ready.

PR #21796, opened 2025-10-29
Address bug with convolution using Tensorflow, Numpy, Jax backends

danielenricocahall:
Addressing issue detailed in https://github.com/keras-team/keras/issues/19753 where Torch and OpenVINO properly raise an exception when a kernel is bigger than the input but Tensorflow, Numpy, and Jax do not.

### DIFF
--- a/keras/src/backend/tensorflow/nn.py
+++ b/keras/src/backend/tensorflow/nn.py
@@ -318,20 +318,20 @@ def conv(
             data_format=tf_data_format,
             dilations=dilation_rate,
         )
-        result_shape = tf.shape(result)
-        assertion = tf.Assert(
-            tf.reduce_all(result_shape > 0),
-            [
+        result_shape = result.shape
+        if (
+            result_shape.is_fully_defined()
+            and math.prod(result_shape.as_list()) == 0
+        ):
+            raise ValueError(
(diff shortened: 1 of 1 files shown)


keras/src/backend/tensorflow/nn.py
```diff
@@ -310,14 +310,28 @@ def conv(
...
+                "The convolution operation resulted in an empty output. "
+                "Output shape:",
+                result_shape,
+                ". This can happen if the input is too small for the given "
+                "kernel size, strides, dilation rate, and padding mode. "
+                "Please check the input shape and convolution parameters.",
+            ],
+        )
+        with tf.control_dependencies([assertion]):
+            return tf.identity(result)
```
hertschuh: I appreciate supporting dynamic shapes too, but I hear this type of dynamic asserts can be expensive. So let's only cover the statically known shape case (which should cover the vast majority of uses).

```python
result_shape = result.shape
if result_shape.is_fully_defined() and math.prod(result_shape.as_list()) == 0:
    raise ValueError(
    ...
    )
return result
```

danielenricocahall: Done! Gemini's suggestion was good but maybe overzealous :)

hertschuh (approved):

PR #21797, opened 2025-10-29
Fix: Remove deprecated .path access in Muon optimizer for TF 2.16+ compatibility

### LINKED ISSUE #21793: keras.optimizers.Muon Fails with AttributeError on variable.path in Keras 3 / TF 2.16-2.20
### **Bug Report: `keras.optimizers.Muon` Fails with `AttributeError` on `variable.path` in Keras 3 / TF 2.16+**

**1. Bug Description**

The `keras.optimizers.Muon` optimizer is currently incompatible with modern Keras 3 and TensorFlow (2.16+) environments. When a model is compiled with the `Muon` optimizer and training begins, it fails during the first training step with the error: `AttributeError: 'ResourceVariable' object has no attribute 'path'`.

The issue stems from the internal implementation of the optimizer's `_adamw_update_step` method, which attempts to access a `.path` attribute on a TensorFlow `ResourceVariable`. This attribute no longer exists on variables in recent versions of TensorFlow when used with the Keras 3 backend abstraction layer.

This bug makes the `Muon` optimizer unusable in its current state for anyone using an up-to-date TensorFlow backend.

**2. Reproducible Example**

The following minimal code consistently reproduces the error. It uses a simple Sequential model, dummy data, and the `Muon` optimizer. The error occurs immediately upon calling `model.fit()`.

```python
import tensorflow as tf
import keras
import numpy as np

print(f"TensorFlow version: {tf.__version__}")

# 1. Define a simple model
# A named layer is included to demonstrate the `exclude_layers` argument.
model = keras.Sequential([
    keras.layers.Dense(5), 
])

# 2. Create dummy data
x_train = np.random.rand(100, 10).astype(np.float32)
y_train = np.random.rand(100, 1).astype(np.float32)

# 3. Instantiate the Muon optimizer
# The error occurs with or without `exclude_layers`.
optimizer = keras.optimizers.Muon(learning_rate=1e-3, exclude_layers=["last"])

# 4. Compile the model and attempt to fit
model.compile(optimizer=optimizer, loss='mse')

print("\nAttempting to start model.fit()... The error is expected here.")
try:
    model.fit(x_train, y_train, epochs=1, batch_size=10)
except AttributeError as e:
    print(f"\nSuccessfully reproduced the error: {e}")
```

utsab345:
### Summary
Fixes keras.optimizers.Muon failing with AttributeError: 'ResourceVariable' object has no attribute 'path' in Keras 3 / TF 2.16–2.20.

### Changes
- Replaced deprecated `.path` references with `.name` for variable identification.
- Updated `_should_use_adamw()` logic to match modern Keras 3 variable handling.
- Added robust tests covering:
  - No `.path` attribute access
  - Layer exclusion behavior
  - Compatibility with embeddings and dense layers

### Result
All 11 tests passed locally on TensorFlow 2.16+ (Windows).

### DIFF
--- a/keras/src/optimizers/muon.py
+++ b/keras/src/optimizers/muon.py
@@ -134,47 +134,57 @@ class Muon(optimizer.Optimizer):
         # any {0,1}-D parameters should all be optimized by adam
         if not 1 < len(variable.shape) < 4:
             return True
-        if self.exclude_embeddings and "embedding" in variable.path.lower():
+        # Check .path only during build (where we have keras.Variable)
+        var_path = variable.path if hasattr(variable, "path") else None
+        if var_path is None:
+            return False
+        if self.exclude_embeddings and "embedding" in var_path.lower():
             return True
-        for keyword in self.exclude_layers:
-            if re.search(keyword, variable.path):
-                return True
+        # Exclude any user-specified layer patterns
+        for pattern in self.exclude_layers:
+            try:
+                if re.search(pattern, var_path):
+                    return True
+            except (re.error, TypeError):
+                # Skip invalid regex patterns in exclude_layers
+                continue
         return False
 
     def build(self, var_list):
         """Initialize optimizer variables.
 
-        Adam optimizer has 3 types of variables: momentums, velocities and
-        velocity_hat (only set when amsgrad is applied),
+        Muon optimizer has 2 types of variables: momentums and velocities.
+        Velocities are only set when using AdamW update step.
 
         Args:
-            var_list: list of model variables to build Adam variables on.
+            var_list: list of model variables to build Muon variables on.
         """
         if self.built:
             return
         super().build(var_list)
-        self.adam_momentums = {}
-        self.adam_velocities = {}
-
-        self.muon_momentums = {}
-        self.muon_velocities = {}
+        # Initialize lists with None for all variables
+        self.adam_momentums = [None] * len(var_list)
+        self.adam_velocities = [None] * len(var_list)
 
         for var in var_list:
             if not self._overwrite_variable_with_gradient(var):
-                self.adam_momentums[var.path] = (
+                var_idx = self._get_variable_index(var)
+                self.adam_momentums[var_idx] = (
                     self.add_variable_from_reference(
                         reference_variable=var, name="momentum"
                     )
                 )
                 if self._should_use_adamw(var):
-                    self.adam_velocities[var.path] = (
+                    self.adam_velocities[var_idx] = (
                         self.add_variable_from_reference(
                             reference_variable=var, name="velocity"
                         )
                     )
 
     def update_step(self, gradient, variable, learning_rate):
-        if self._should_use_adamw(variable):
+        var_idx = self._get_variable_index(variable)
+        # Check if velocity exists to determine if we should use AdamW
+        if self.adam_velocities[var_idx] is not None:
             # It should be noted that lr is one-tenth when using adamw.
             self._adamw_update_step(
                 gradient, variable, learning_rate * self.adam_lr_ratio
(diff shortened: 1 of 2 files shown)


keras/src/optimizers/muon.py
```diff
@@ -183,7 +193,7 @@ def update_step(self, gradient, variable, learning_rate):
             self._muon_update_step(gradient, variable, learning_rate)
 
     def _muon_update_step(self, gradient, variable, lr):
-        m = self.adam_momentums[variable.path]
+        m = self.adam_momentums[variable.name]
```
hertschuh: This issue is that variable names are not unique. For instance, all `Dense` layers have a `"kernel"` variable. So multiple variables will clobber one another.

keras/src/utils/image_utils.py
```diff
@@ -463,7 +463,7 @@ def smart_resize(
     img = backend_module.image.resize(
         img, size=size, interpolation=interpolation, data_format=data_format
     )
-
+    
     if isinstance(x, np.ndarray):
         return np.array(img)
-    return img
+    return img
```
hertschuh, line 455: Revert this file

keras/src/utils/image_utils_test.py
```diff
@@ -29,8 +29,7 @@ def test_save_jpg(self, shape, name, file_format, use_explicit_format):
             save_img(path, img)  # Let it infer from path
 
         self.assertTrue(os.path.exists(path))
-
         # Verify saved image is correctly converted to RGB if needed
         loaded_img = load_img(path)
         loaded_array = img_to_array(loaded_img)
-        self.assertEqual(loaded_array.shape, (50, 50, 3))
+        self.assertEqual(loaded_array.shape, (50, 50, 3))
```
hertschuh, line 36: Revert this file

keras/src/optimizers/muon.py
```diff
@@ -129,16 +129,27 @@ def __init__(
...
+        if self.exclude_embeddings and "embedding" in var_identifier.lower():
             return True
-        for keyword in self.exclude_layers:
-            if re.search(keyword, variable.path):
-                return True
+
+        # Exclude any user-specified layer patterns
+        for pattern in self.exclude_layers:
+            try:
+                if re.search(pattern, var_identifier):
```
hertschuh: This won't work because the layer name is not part of the variable name, only the variable path.

hertschuh (commented):
@utsab345 

First, I want to clarify one thing. It's not the `.path` is deprecated. `.path` is a concept that exists in `keras.Variable` but doesn't exist in `tf.Variable`.

Second, this issue happens because there is a limitation with TensorFlow whereby the variable that is passed to `update_step` is not the `keras.Variable` but the `tf.Variable`.

However, there is a mechanism in place to workaround this limitation. Please take a look at how `Adam` is implemented. Everything has to be based on `self._get_variable_index(var)`, which is guaranteed to work with both `keras.Variable` and `tf.Variable`.

So instead of what was done:
- you should inspect `.path` in `build`, but only in `build`
- you should only use `self._get_variable_index(variable)` in `update_step`.

All the momentums and velocities should be lists, not dicts. You want to initialize them all with `[None] * len(var_list)`.
Then do:

```python
        for var in var_list:
                self.adam_momentums[self._get_variable_index(var)] = (
                        reference_variable=var, name="momentum"
                    )
                )
                if self._should_use_adamw(var):
                    self.adam_velocities[self._get_variable_index(var)] = (
                            reference_variable=var, name="velocity"
                        )
                    )
```

And the same for the muon momentums / velocities.

Then in `update_step`, you'll check:
`self.adam_velocities[self._get_variable_index(variable)]`

If it's `None`, you call `_muon_update_step`, else `_adamw_update_step`.

Does that make sense? Sorry, this is probably a bit more complicated than what you had in mind.

hertschuh:
@utsab345 

Please let me know when this is ready for re-review. Thanks!

pass-lin:
I've provided an improved bug-fix version that doesn't affect the current design of the Muon optimizer.
https://github.com/keras-team/keras/pull/21859

hertschuh:
Closing in favor of https://github.com/keras-team/keras/pull/21924

Issue #21300, opened 2025-05-18
keras.layers.Normalization#adapt() does not support keras.utils.PyDataset as input.


limzikiki:
## Expected behavior
When I create a custom PyDataset and then pass it into the Normalization adapt method, the mean and variance should be computed successfully, as for a TensorFlow dataset. 
## Actual behavior
Method adapt does not handle the PyDataset like inputs. 
```
Traceback (most recent call last):
  File "/home/user/project/keras_bug_sample.py", line 14, in <module>
    normalizer.adapt(CustomDataset)
  File "/home/user/.pyenv/versions/3.12.8/lib/python3.12/site-packages/keras/src/layers/preprocessing/normalization.py", line 230, in adapt
    self.build(input_shape)
               ^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'input_shape' where it is not associated with a value
```

## Steps to reproduce
```python
import keras

class CustomDataset(keras.utils.PyDataset):
    def __len__(self):
        return 100

    def __getitem__(self, idx):
        # Generate dummy data
        x = np.random.rand(32, 32, 3)
        y = np.random.randint(0, 10, size=(1,))
        return x, y

normalizer = keras.layers.Normalization()
normalizer.adapt(CustomDataset)
```

sonali-kumari1:
Hi @limzikiki - 
Thanks for reporting this. The error `UnboundLocalError: cannot access local variable 'input_shape' where it is not associated with a value` occurs because [adapt()](https://github.com/keras-team/keras/blob/v3.9.2/keras/src/layers/preprocessing/normalization.py#L199) does not directly support custom datasets like `keras.utils.PyDataset`. The `adapt()` method expects data to be in the form of `tf.data.Dataset`, NumPy array, or a backend-native eager tensor. Since you are passing `keras.utils.PyDataset`, `adapt()` is not able to extract the `input_shape` which leads to the `UnboundLocalError`. To resolve this you can convert your custom dataset into `tf.data.Dataset` and ensure that dataset is properly batched to compute the mean and variance correctly.

VarunS1997:
Please go ahead and make a PR, @limzikiki -- thank you!

tripathi-genius:
Is it open to contributions? I want to contribute.

hertschuh:
@tripathi-genius 

> Is it open to contributions? I want to contribute.

Yes!

[...]

PR #21817, opened 2025-11-03
Support PyDataset in Normalization layer `adapt` methods

danielenricocahall:
Addresses https://github.com/keras-team/keras/issues/21300 by adding support for `PyDataset` by converting it into a `tf.data.Dataset` per the suggestions in the Issue. Additionally, raise an exception if an unsupported type is supplied rather than having it proceed and fail on the `UnboundLocalVariable` error.

### DIFF
--- a/keras/src/layers/preprocessing/normalization.py
+++ b/keras/src/layers/preprocessing/normalization.py
@@ -6,7 +6,6 @@ from keras.src import backend
 from keras.src import ops
 from keras.src.api_export import keras_export
 from keras.src.layers.preprocessing.data_layer import DataLayer
-from keras.src.trainers.data_adapters import get_data_adapter
 from keras.src.utils.module_utils import tensorflow as tf
 from keras.utils import PyDataset
 
@@ -232,12 +231,11 @@ class Normalization(DataLayer):
                 data = data.batch(128)
             input_shape = tuple(data.element_spec.shape)
         elif isinstance(data, PyDataset):
-            adapter = get_data_adapter(data)
-            tf_dataset = adapter.get_tf_dataset()
-            # args will either be (samples,), (samples, labels) pairs,
-            # or (samples, labels, sample_weights) tuples
-            data = tf_dataset.map(lambda *args: args[0])
-            input_shape = data.element_spec.shape
+            data = data[0]
+            if isinstance(data, tuple):
+                # handling (x, y) or (x, y, sample_weight)
+                data = data[0]
+            input_shape = data.shape
         else:
             raise TypeError(
                 f"Unsupported data type: {type(data)}. `adapt` supports "
@@ -263,7 +261,7 @@ class Normalization(DataLayer):
         elif backend.is_tensor(data):
             total_mean = ops.mean(data, axis=self._reduce_axis)
             total_var = ops.var(data, axis=self._reduce_axis)
-        elif isinstance(data, tf.data.Dataset):
+        elif isinstance(data, (tf.data.Dataset, PyDataset)):
             total_mean = ops.zeros(self._mean_and_var_shape)
             total_var = ops.zeros(self._mean_and_var_shape)
             total_count = 0
--- a/keras/src/layers/preprocessing/normalization_test.py
+++ b/keras/src/layers/preprocessing/normalization_test.py
@@ -170,7 +170,7 @@ class NormalizationTest(testing.TestCase):
         layer = layers.Normalization(mean=3.0, variance=2.0)
         layer(input_data)
 
-    @parameterized.parameters([("x",), ("x_and_y",), ("x_y_and_weights")])
+    @parameterized.parameters([("x",), ("x_and_y",), ("x_y_and_weights",)])
     def test_adapt_pydataset_compat(self, pydataset_type):
         import keras
 


keras/src/layers/preprocessing/normalization.py
```diff
@@ -229,6 +231,24 @@ def adapt(self, data):
...
+                data = tf_dataset.map(lambda x: x)
+                input_shape = data.element_spec.shape
+            elif len(tf_dataset.element_spec) == 2:
+                # (x, y) pairs
+                data = tf_dataset.map(lambda x, y: x)
+                input_shape = data.element_spec.shape
+            elif len(tf_dataset.element_spec) == 3:
+                # (x, y, sample_weight) tuples
+                data = tf_dataset.map(lambda x, y, z: x)
+                input_shape = data.element_spec.shape
```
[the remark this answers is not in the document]
danielenricocahall: good call gemini

keras/src/layers/preprocessing/normalization.py
```diff
@@ -229,6 +231,24 @@ def adapt(self, data):
...
+        else:
+            raise NotImplementedError(f"Unsupported data type: {type(data)}")
```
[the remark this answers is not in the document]
danielenricocahall: done

keras/src/layers/preprocessing/normalization.py
```diff
@@ -229,6 +231,26 @@ def adapt(self, data):
...
+            if len(tf_dataset.element_spec) == 1:
+                # just x
+                data = tf_dataset.map(lambda x: x)
+            elif len(tf_dataset.element_spec) == 2:
+                # (x, y) pairs
+                data = tf_dataset.map(lambda x, y: x)
+            elif len(tf_dataset.element_spec) == 3:
+                # (x, y, sample_weight) tuples
+                data = tf_dataset.map(lambda x, y, z: x)
+            input_shape = data.element_spec.shape
```
limzikiki: Coming from your comment. 
What I did in my solution is:
```python
        elif isinstance(data, keras.utils.PyDataset):
            sample_input = data[0][0] # pydataset should return a tuple with first element being the data
            if isinstance(sample_input, np.ndarray) or backend.is_tensor(sample_input):
                input_shape = sample_input.shape
            else:
                raise ValueError(f"Unsupported data type: {type(sample_input)} returned from the PyDataset")
```
The advantage of my option lies in the fact that we don’t need to perform excessive transformations to tf tensors just for the sake of size estimation. PyDataset is also used for experimentation, and when the dataset is too large to be read into RAM, which is common for workstations and personal devices, transforming PyDataset into a TF Tensor will fail due to memmory allocation. However on contrary the drawback of my solution is that it retrieves the first batch, and during that the first batch might get changed and second retrieval of the first batch might return not the same output (If someone implemented non-idempotent PyDataset, but i think then it is a user problem then). Also why retrieval of the first batch is feasable solution is because shape of all the elements across all the batches must be identical for normalization to work correctly. 

Considering strategic direction of Keras to move away from being solely dependant on TensorFlow, adding transformation to tensorflow creates a technical debt that Keras team should later take care off.

I am open for a discussion

danielenricocahall: Thank you for the thorough response! Yes, I will defer to the core developers' judgement for this. Happy to revise and infer the shape based on sampling a batch if we think that's the better approach.

keras/src/layers/preprocessing/normalization.py
```diff
@@ -229,6 +231,26 @@ def adapt(self, data):
                 # Batch dataset if it isn't batched
                 data = data.batch(128)
             input_shape = tuple(data.element_spec.shape)
+        elif isinstance(data, PyDataset):
+            # as PyDatasets returns tuples of input/annotation pairs
+            adapter = get_data_adapter(data)
+            tf_dataset = adapter.get_tf_dataset()
+            if len(tf_dataset.element_spec) == 1:
+                # just x
+                data = tf_dataset.map(lambda x: x)
```
fchollet: This isn't doing anything, and can be removed

keras/src/layers/preprocessing/normalization.py
```diff
@@ -229,6 +231,26 @@ def adapt(self, data):
...
+            elif len(tf_dataset.element_spec) == 2:
```
fchollet: It seems fine to only support PyDataset instances that yield data samples (no targets) since that is what we expect for tf.data.Dataset instances. No?

fchollet (commented):
Thanks for the PR!

keras/src/layers/preprocessing/normalization.py
danielenricocahall: I did that because:
```python
tf_dataset.element_spec
(TensorSpec(shape=(None, 32, 3), dtype=tf.float64, name=None),)
tf_dataset.map(lambda x: x).element_spec
TensorSpec(shape=(None, 32, 3), dtype=tf.float64, name=None)
```

otherwise, there's a `E           AttributeError: 'tuple' object has no attribute 'shape'` since the former returns a tuple. Alternatively, without the `map` we could get the shape via `tf_dataset.element_spec[0].shape`.

keras/src/layers/preprocessing/normalization.py
danielenricocahall: Possibly - I was thinking there may be users who want to provide a `PyDataset` with samples and targets and compute the mean/variance from the samples, but it may not be necessary. I could remove it if we don't think that's a use-case

danielenricocahall: Updated to use `args` instead so we don't have these baked in conditionals but will support these variations - what do you think?

hertschuh (commented):
@danielenricocahall 

Thank you for adding this feature.

Your current approach falls back to wrapping the `PyDataset` into a `tf.data.Dataset`.

Ideally, we wouldn't do this to not need to depend on TensorFlow at all. It should actually be pretty easy because the code lines 267 to 298 should already work as-is (you might need minor tweaks).

You'll simply change:
```python
elif isinstance(data, (tf.data.Dataset, PyDataset)):
```

Because `PyDataset` is iterable, the for loop will work.

And in terms of what is done line 234, this might just be enough:
```python
elif isinstance(data, PyDataset):
    input_shape = data[0].shape
```

Note that the `PyDataset` should not return tuples of tensors, just tensors. The same constraint is what is done for `tf.data.Dataset`.

danielenricocahall:
> @danielenricocahall
> 
> Thank you for adding this feature.
> 
> Your current approach falls back to wrapping the `PyDataset` into a `tf.data.Dataset`.
> 
> Ideally, we wouldn't do this to not need to depend on TensorFlow at all. It should actually be pretty easy because the code lines 267 to 298 should already work as-is (you might need minor tweaks).
> 
> You'll simply change:
> 
> ```python
> elif isinstance(data, (tf.data.Dataset, PyDataset)):
> ```
> 
> Because `PyDataset` is iterable, the for loop will work.
> 
> And in terms of what is done line 234, this might just be enough:
> 
> ```python
> elif isinstance(data, PyDataset):
>     input_shape = data[0].shape
> ```
> 
> Note that the `PyDataset` should not return tuples of tensors, just tensors. The same constraint is what is done for `tf.data.Dataset`.

Thank you @hertschuh ! Updated accordingly, the tests are passing with this simpler approach.

hertschuh (approved):

PR #21818, opened 2025-11-03
Pass optional field in a few places to fix None input error.

wenyi-guo:


keras/src/models/functional.py
```diff
@@ -832,11 +832,16 @@ def clone_graph_nodes(inputs, outputs):
             kt_id_mapping[id(kt_input)] = kt_input
         else:
             # We need to create a new Keras tensor for any intermediate tensor
+            original_op = kt_input._keras_history[0]
```
hertschuh: `original_op = kt_input._keras_history.operation`

`KerasHistory` is just a namedtuple. So it's nicer to use the field names.

keras/src/models/functional.py
```diff
@@ -832,11 +832,16 @@ def clone_graph_nodes(inputs, outputs):
...
+            is_optional = False
```
hertschuh: nitpick: any reason to not call it just `optional` instead of `is_optional`?

keras/src/models/cloning.py
```diff
@@ -313,6 +314,7 @@ def _clone_sequential_model(model, clone_function, input_tensors=None):
         inputs = Input(
             tensor=input_tensors,
             name=input_name,
+            optional=ref_input_layer.optional if ref_input_layer else False,
```
hertschuh: `if ref_input_layer`

Let's be explicit and do `if ref_input_layer is not None`. I've run into many issues with this shortcut so I really avoid it. It's unclear if `__bool__` is implemented on that class and how.

keras/src/models/cloning.py
```diff
@@ -321,6 +323,7 @@ def _clone_sequential_model(model, clone_function, input_tensors=None):
                 batch_shape=input_batch_shape,
                 dtype=input_dtype,
                 name=input_name,
+                optional=ref_input_layer.optional if ref_input_layer else False,
```
hertschuh: `optional=input_optional`

keras/src/models/cloning.py
hertschuh: Actually, (different option), let's stay consistent with the current approach, i.e.
- add `input_optional = ref_input_layer.optional` line 296
- add `input_optional = False` line 300
- then use `input_optional` here and below.

hertschuh (approved):

Issue #21768, opened 2025-10-23
To limit `noise_shape` in func `keras.layers.Dropout()`


ILCSFNO:
## Bug Issue

The doc of `keras.layers.Dropout()` shows its description as below:

https://github.com/keras-team/keras/blob/465a56d7f1284f9b0d18ead3d6f87afcd3db1017/keras/src/layers/regularization/dropout.py#L27-L32

For the repro below, which is expected to raise error, I found that `noise_shape` in `keras.layers.Dropout` can be any value without raising any error, including the negative numbers.

Using `tf 2.20.0` and `keras` latest:

### Repro
```python
import tensorflow as tf
import keras
noise_shape = (1000, 2124, 6415, 461, -45616, 61515, -100, 123)
input_data = tf.random.normal((10, 5))
model = keras.models.Sequential([keras.layers.Dense(64, activation='relu', input_shape=(5,)), keras.layers.Dropout(rate=0.5, noise_shape=noise_shape), keras.layers.Dense(10, activation='softmax')])
model.compile(optimizer='adam', loss='sparse_categorical_crossentropy', metrics=['accuracy'])
model.summary()
output = model(input_data)
print(input_data.shape, output.shape)
```
### Output
```text
Model: "sequential_10"
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┓
┃ Layer (type)                    ┃ Output Shape           ┃       Param # ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━┩
│ dense_24 (Dense)                │ (None, 64)             │           384 │
├─────────────────────────────────┼────────────────────────┼───────────────┤
│ dropout_17 (Dropout)            │ (None, 64)             │             0 │
├─────────────────────────────────┼────────────────────────┼───────────────┤
│ dense_25 (Dense)                │ (None, 10)             │           650 │
└─────────────────────────────────┴────────────────────────┴───────────────┘
Total params: 1,034 (4.04 KB)
Trainable params: 1,034 (4.04 KB)
Non-trainable params: 0 (0.00 B)
(10, 5) (10, 10)
```

I think that `noise_shape` should be limited, but not sure if this is expected actually.

Thanks for noting!

dhantule:
[...]

PR #21819, opened 2025-11-03
Fix noise_shape validation in keras.layers.Dropout

Abhinavexists:
Fix: #21768

added input validation for the noise_shape parameter in `keras.layers.Dropout ` to prevent invalid values from being passed at layer initialization time.

This is an initial implementation focusing on basic validation logic. The implementation
  intentionally:
  -  Validates input types and values at initialization
  - Does not yet implement broadcasting validation against input shape (deferred for future
  iteration)
  - Does not include comprehensive tests (will be added once approach is approved)

### DIFF
--- a/keras/src/layers/regularization/dropout.py
+++ b/keras/src/layers/regularization/dropout.py
@@ -72,7 +72,8 @@ class Dropout(Layer):
             except TypeError:
                 raise ValueError(
                     f"Invalid value received for argument `noise_shape`. "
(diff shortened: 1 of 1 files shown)


Abhinavexists:
Hi @fchollet 

i have fixed the code format, also have added unit test for `_validate_noise_shape` that now covers all the expected behaviour.

also,
Added explicit string type check to reject strings before tuple conversion

image

PR #21823, opened 2025-11-05
Fix `Progbar.update` when receiving list, np arrays, and tensors.

### LINKED ISSUE #21821: Bug Report: Progress bar update fails in Keras 3.12 when a metric returns a list or numpy array (e.g. histogram, confusion matrix)
System Info:

Keras version: 3.12.0

TensorFlow version: 2.20.0

Python version: 3.12.3

OS: Ubuntu 24.04

# Bug: `ValueError` in Keras 3.12 `Progbar` when metric returns list/array

### Description
In **Keras 3.12**, `keras.utils.Progbar` crashes when a metric returns a list or NumPy array (e.g. histogram metric):

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

Worked fine in **Keras 3.11.x** and before.

---

### Source Section
```python
if abs(avg) > 1e-3:  # avg can be a list/array here
    info += f" {avg:.4f}"
```
### Cause
`avg` can be a list or array, making `if abs(avg) > 1e-3:` invalid, resulting to

> ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

### Current Workaround
Setting
`verbose = 0`
disables the progress bar and avoids the crash.

### Suggested Fix
```
if isinstance(avg, (list, np.ndarray)):
    info += " [non-scalar metric]"  # or info += str(avg)
else:
    info += f" {avg:.4f}" if abs(avg) > 1e-3 else f" {avg:.4e}"
```
or like in keras 3.11.3 mean
```
            avg = backend.numpy.mean(
                self._values[k][0] / max(1, self._values[k][1])
            )
```

### Example
avg = np.array([0.0, 0.0, ..., 1328.0, -3.0, 3.0, 200.0])  # triggers the error

### Change that caused the problem
keras 3.11.3
```
# Add metrics
for k in self._values_order:
    info += f" - {k}:"
    if isinstance(self._values[k], list):
        avg = backend.convert_to_numpy(
            backend.numpy.mean(
                self._values[k][0] / max(1, self._values[k][1])
            )
        )
        avg = float(avg)
        if abs(avg) > 1e-3:
            info += f" {avg:.4f}"
        else:
            info += f" {avg:.4e}"
    else:
        info += f" {self._values[k]}"
message += info
```

In keras 3.12.0 the             
```
backend.numpy.mean(
                self._values[k][0] / max(1, self._values[k][1])
            )
```
is removed.

james77777778:
Fix #21821 and includes a new test to verify the modification.

cc @hertschuh 

Could you please confirm if this PR breaks #21709?

### DIFF
--- a/keras/src/utils/progbar.py
+++ b/keras/src/utils/progbar.py
@@ -1,13 +1,10 @@
 import math
-import numbers
 import os
 import sys
 import time
-from collections.abc import Sequence
 
 import numpy as np
 
-from keras.src import backend
 from keras.src.api_export import keras_export
 from keras.src.utils import io_utils
 
@@ -167,7 +164,9 @@ class Progbar:
                 info += f" - {k}:"
                 if isinstance(self._values[k], list):
                     values, count = self._values[k]
-                    avg = self._compute_avg(values, count)
+                    if not isinstance(values, float):
+                        values = np.mean(values)
+                    avg = values / max(1, count)
                     if abs(avg) > 1e-3:
                         info += f" {avg:.4f}"
                     else:
@@ -195,7 +194,9 @@ class Progbar:
                 for k in self._values_order:
                     info += f" - {k}:"
                     values, count = self._values[k]
-                    avg = self._compute_avg(values, count)
+                    if not isinstance(values, float):
+                        values = np.mean(values)
+                    avg = values / max(1, count)
                     if avg > 1e-3:
                         info += f" {avg:.4f}"
                     else:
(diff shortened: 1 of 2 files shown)


hertschuh (commented):
Thanks for fixing this! This is a regression from my CL.

Because of the way `pythonify_logs` works, `values` should be either a float or a NumPy array, it should never be a tensor. So this should be enough:

```python
if not isinstance(values, float):
  values = float(np.mean(values)
avg = values / max(1, count)
```

Which can probably just be inlined in the code.

hertschuh (approved):
Thank you for the fix!

PR #21827, opened 2025-11-06
Fix CosineDecay documentation to clarify alpha is a multiplier

### LINKED ISSUE #21772: Unclear description of tf.keras.optimizers.schedules.CosineDecay
In the description of tf.keras.optimizers.schedules.CosineDecay it is written: "...taking our learning rate from warmup_target to alpha..." and "...decay will take our learning rate from initial_learning_rate to alpha." This is not correct, since the learning rate is taken to warmup_target * alpha and initial_learning_rate * alpha, respectively. The initial learning rate is multiplied by the value alpha, not taken to alpha.

yashwantbezawada:
Fixes #21772

The CosineDecay documentation was misleading about how the `alpha` parameter works.

**Current docs say:** learning rate decays "to alpha"
**Reality:** learning rate decays to `initial_learning_rate * alpha`

The parameter description correctly states alpha is "a fraction of initial_learning_rate", but the explanation text contradicted this. Updated the explanation to match the actual implementation and parameter description.

### DIFF
--- a/keras/src/optimizers/schedules/learning_rate_schedule.py
+++ b/keras/src/optimizers/schedules/learning_rate_schedule.py
@@ -584,9 +584,10 @@ class CosineDecay(LearningRateSchedule):
     schedule applies a linear increase per optimizer step to our learning rate
     from `initial_learning_rate` to `warmup_target` for a duration of
     `warmup_steps`. Afterwards, it applies a cosine decay function taking our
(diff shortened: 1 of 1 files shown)


yashwantbezawada:
Signed the CLA

fchollet (approved):
LGTM, thanks for the PR

PR #21833, opened 2025-11-07
Refactor __exit__ method in name_scope context manager to handle thread-local global state

### LINKED ISSUE #21831: [Bug] `name_scope_stack` causing exceptions
Hi, I'm currently building a tool for an EU research project and I came across a bug with `name_scope_stack`.

I'm getting two different errors when calling `predict` or when calling `load_model`,
```python
...
.../.venv/lib/python3.12/site-packages/keras/src/backend/common/name_scope.py", line 61, in __exit__
    name_scope_stack.pop()
    ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'pop'
```
or 
```python
 .../.venv/lib/python3.12/site-packages/keras/src/backend/common/name_scope.py", line 61, in __exit__
    name_scope_stack.pop()
IndexError: pop from empty list
```

If I manually add a guard clause to handle the case of `name_scope_stack` being `None` or `name_scope_stack` being empty, my code executes fine.
Any chance you could add those guards, so we can move forward with our project?

Thanks in advance.

Abhinavexists:
Fix #21831 

simple fix which just checks if name_scope_stack is empty or dosen't exist
i will build the test case once the approach is approved

### DIFF
--- a/keras/src/backend/common/name_scope.py
+++ b/keras/src/backend/common/name_scope.py
@@ -54,12 +54,13 @@ class name_scope:
         return self
 
     def __exit__(self, *args, **kwargs):
-        if self._pop_on_exit:  
-            name_scope_stack = global_state.get_global_attribute(  
-                "name_scope_stack"  
-            )  
-            if name_scope_stack:  
-                name_scope_stack.pop() 
+        if self._pop_on_exit:
+            name_scope_stack = global_state.get_global_attribute(
+                "name_scope_stack"
+            )
+            if name_scope_stack:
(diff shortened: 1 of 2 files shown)


Abhinavexists:
I have now added the test cases and have minimised the approach towards handling global_state
cc @fchollet

hertschuh:
@Abhinavexists 

Thanks for the PR. It turns out it was already done as part of https://github.com/keras-team/keras/pull/21834

PR #21803, opened 2025-10-31
OpenVINO NN Module Functions

danielenricocahall:
Support for:
- max pooling
- average pooling
- CTC loss
- one hot encoding
- CELU
- log sigmoid
- dot product attention

To ensure these tests run during CI, I have removed the `keras/src/ops/nn_test.py` from the `keras/src/backend/openvino/excluded_tests.txt` file, which was used to skip tests for the OpenVINO backend, and expanded `keras/src/backend/openvino/excluded_concrete_tests.txt` with new tests such that test omission is decided in a fixture.

PR #21836, opened 2025-11-09
Support various filtering functions in OpenVINO

danielenricocahall:
- Support `hamming` window
- Support `bartlett` window
- Support `blackman` window

Added something similar to my other PR https://github.com/keras-team/keras/pull/21803/files#diff-9ded43fac36aa350e46bcc0d2e5d9ad7853a04d54ef08882af4c2f02374207a5R1327 to support the resolution discrepancies in OV.

### DIFF
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -1215,6 +1215,10 @@ def lcm(x1, x2):
     raise NotImplementedError("`lcm` is not supported with openvino backend")
 
 
+def ldexp(x1, x2):
+    raise NotImplementedError("`ldexp` is not supported with openvino backend")
+
+
 def less(x1, x2):
     element_type = None
     if isinstance(x1, OpenVINOKerasTensor):
--- a/keras/src/ops/numpy_test.py
+++ b/keras/src/ops/numpy_test.py
@@ -254,6 +254,11 @@ class NumpyTwoInputOpsDynamicShapeTest(testing.TestCase):
         y = KerasTensor((2, None))
         self.assertEqual(knp.lcm(x, y).shape, (2, 3))
 
+    def test_ldexp(self):
+        x = KerasTensor((None, 3))
+        y = KerasTensor((1, 3))
+        self.assertEqual(knp.ldexp(x, y).shape, (None, 3))
+
     def test_less(self):
         x = KerasTensor((None, 3))
         y = KerasTensor((2, None))
@@ -836,6 +841,15 @@ class NumpyTwoInputOpsStaticShapeTest(testing.TestCase):
         x = KerasTensor((2, 3))
         y = KerasTensor((2, 3))
         self.assertEqual(knp.lcm(x, y).shape, (2, 3))
+
+    def test_ldexp(self):
+        x = KerasTensor((2, 3))
+        y = KerasTensor((2, 3))
+        self.assertEqual(knp.ldexp(x, y).shape, (2, 3))
+
+        x = KerasTensor((2, 3))
+        y = KerasTensor((1, 3))
+        self.assertEqual(knp.ldexp(x, y).shape, (2, 3))
 
     def test_less(self):
         x = KerasTensor((2, 3))
@@ -3113,6 +3127,12 @@ class NumpyTwoInputOpsCorrectnessTest(testing.TestCase):
         y = np.array([4])
         self.assertAllClose(knp.lcm(x, y), np.lcm(x, y))
         self.assertAllClose(knp.Lcm()(x, y), np.lcm(x, y))
+
+    def test_ldexp(self):
+        x = np.array([[1, 2, 3], [3, 2, 1]])
+        y = np.array([[4, 5, 6], [3, 2, 1]])
+        self.assertAllClose(knp.ldexp(x, y), np.ldexp(x, y))
+        self.assertAllClose(knp.Ldexp()(x, y), np.ldexp(x, y))
 
     def test_less(self):
         x = np.array([[1, 2, 3], [3, 2, 1]])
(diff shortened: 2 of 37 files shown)


keras/src/backend/openvino/numpy.py
```diff
@@ -545,16 +545,61 @@ def average(x, axis=None, weights=None):
 
 
 def bartlett(x):
```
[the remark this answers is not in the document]
danielenricocahall: Will keep the argument name the same as  `x` is the argument name for all other backend's `bartlett` definitions

keras/src/ops/numpy_test.py
```diff
@@ -3622,6 +3622,15 @@ def test_digitize(self):
 
 
 class NumpyOneInputOpsCorrectnessTest(testing.TestCase):
+    def assertAllClose(self, x1, x2, atol=1e-6, rtol=1e-6, msg=None):
```
hertschuh: Let's not do that, it reduces the precision for dozens of ops.

Which op(s) have the issue?

danielenricocahall: @hertschuh  Sure - with this change omitted:

-`test_bartlett`
```
E           Mismatched elements: 64 / 73 (87.7%)
E           Max absolute difference among violations: 0.00021701
E           Max relative difference among violations: 0.00035511
E            ACTUAL: array([0.      , 0.027771, 0.055542, 0.083313, 0.111084, 0.138916,
E                  0.166626, 0.194458, 0.222168, 0.25    , 0.277832, 0.305664,
E                  0.333252, 0.361084, 0.388916, 0.416748, 0.444336, 0.472168,...
E            DESIRED: array([0.      , 0.027778, 0.055556, 0.083333, 0.111111, 0.138889,
E                  0.166667, 0.194444, 0.222222, 0.25    , 0.277778, 0.305556,
E                  0.333333, 0.361111, 0.388889, 0.416667, 0.444444, 0.472222,...
```

- `test_blackman`
```
E           Mismatched elements: 40 / 46 (87%)
E           Max absolute difference among violations: 0.0002353
E           Max relative difference among violations: 0.00032871
E            ACTUAL: array([-0.      ,  0.001767,  0.007214,  0.016754,  0.031052,  0.050873,
E                   0.077087,  0.110413,  0.151611,  0.200806,  0.258057,  0.322754,
E                   0.394043,  0.470215,  0.549805,  0.629883,  0.708496,  0.782227,...
E            DESIRED: array([-1.387779e-17,  1.766901e-03,  7.213000e-03,  1.675772e-02,
E                   3.104564e-02,  5.086963e-02,  7.707242e-02,  1.104350e-01,
E                   1.515615e-01,  2.007701e-01,  2.580005e-01,  3.227451e-01,
```

- `test_hamming`
```
E           Mismatched elements: 10 / 11 (90.9%)
E           Max absolute difference among violations: 9.70361625e-05
E           Max relative difference among violations: 0.0002439
E            ACTUAL: array([0.080017, 0.167847, 0.397949, 0.682129, 0.912109, 1.      ,
E                  0.912109, 0.682129, 0.397949, 0.167847, 0.080017], dtype=float32)
E            DESIRED: array([0.08    , 0.167852, 0.397852, 0.682148, 0.912148, 1.      ,
E                  0.912148, 0.682148, 0.397852, 0.167852, 0.08    ])
```

Alternatively, there could be an `"openvino"` check in these tests before calling the assertion and updating the tolerance (such that it doesn't impact other operations to your point).

keras/src/backend/openvino/numpy.py
```diff
@@ -545,16 +545,61 @@ def average(x, axis=None, weights=None):
...
-    raise NotImplementedError(
-        "`bartlett` is not supported with openvino backend"
+    x = get_ov_output(x)
+    zero_const = ov_opset.constant(0, Type.i64)
+    one_const = ov_opset.constant(1, Type.i64)
+    two_const = ov_opset.constant(2, Type.i64)
+    two_const_f64 = ov_opset.constant(2.0, Type.f64)
+    if x.get_element_type() != Type.i64:
+        x = ov_opset.convert(x, Type.i64)
+    half = ov_opset.convert(
```
hertschuh, line 555: For `half`, I think you should not use int constants (`one_const` and `two_const`) and instead use floats of at least the same type as `x`.

keras/src/backend/openvino/numpy.py
```diff
@@ -545,16 +545,61 @@ def average(x, axis=None, weights=None):
...
+        ov_opset.divide(ov_opset.subtract(x, one_const), two_const), Type.f64
+    )
+    n = ov_opset.range(zero_const, x, one_const, Type.f64)
+    condition = ov_opset.less_equal(n, half)
+    first_half = ov_opset.divide(
+        ov_opset.multiply(two_const_f64, n),
+        ov_opset.convert(ov_opset.subtract(x, one_const), Type.f64),
```
hertschuh, line 562: `ov_opset.convert(ov_opset.subtract(x, one_const), Type.f64),`

Instead, use a float version of `one_const`.

keras/src/ops/numpy_test.py
hertschuh: This is surprising. Let's first check that the math is not introducing these differences.

keras/src/backend/openvino/numpy.py
```diff
@@ -545,16 +545,61 @@ def average(x, axis=None, weights=None):
...
-        "`hamming` is not supported with openvino backend"
+    m = get_ov_output(x)
+
+    m_i64 = (
+        m if m.get_element_type() == Type.i64 else ov_opset.convert(m, Type.i64)
+    )
+
+    start = ov_opset.constant(0, Type.i64)
+    step = ov_opset.constant(1, Type.i64)
+    n = ov_opset.range(start, m_i64, step, Type.f64)
```
hertschuh, line 581: In Tensorflow, there is some special handling for odd vs. even numbers of elements, maybe you need this:

https://github.com/tensorflow/tensorflow/blob/v2.16.1/tensorflow/python/ops/signal/window_ops.py#L234-L236

hertschuh:
@danielenricocahall 

I'm a little confused. So the tests now pass with normal tolerance? What changed?

danielenricocahall:
> @danielenricocahall
> 
> I'm a little confused. So the tests now pass with normal tolerance? What changed?

@hertschuh I'm actually thinking some of the discrepancies I'm seeing may be due to my local development environment (maybe some precision differences when using an M2...?). Since running the tests without the tolerance updates in CI/CD seems to be working. If that's the case I could undo the changes in that other PR to evaluate, similar to what I did here?

hertschuh:
> > @danielenricocahall
> > I'm a little confused. So the tests now pass with normal tolerance? What changed?
> 
> @hertschuh I'm actually thinking some of the discrepancies I'm seeing may be due to my local development environment (maybe some precision differences when using an M2...?). Since running the tests without the tolerance updates in CI/CD seems to be working. If that's the case I could undo the changes in that other PR to evaluate, similar to what I did here?

Oh great, so can you try removing the tolerance changes in the other PR?

hertschuh (approved):

danielenricocahall:
> > > @danielenricocahall
> > > I'm a little confused. So the tests now pass with normal tolerance? What changed?
> > 
> > 
> > @hertschuh I'm actually thinking some of the discrepancies I'm seeing may be due to my local development environment (maybe some precision differences when using an M2...?). Since running the tests without the tolerance updates in CI/CD seems to be working. If that's the case I could undo the changes in that other PR to evaluate, similar to what I did here?
> 
> Oh great, so can you try removing the tolerance changes in the other PR?

[done](https://github.com/keras-team/keras/pull/21803/commits/8e4aacfdcc40a4799abd77a24f8c47d66c8cf5ba)! Let's see what happens there

PR #21840, opened 2025-11-10
Implement empty_like function in keras.ops

shashaka:
Adds keras.ops.empty_like, which creates a new tensor with the same shape and dtype as the input tensor, with uninitialized values.
Not supported on OpenVINO.

### DIFF
--- a/keras/src/backend/jax/numpy.py
+++ b/keras/src/backend/jax/numpy.py
@@ -679,8 +679,6 @@ def empty(shape, dtype=None):
 
 
 def empty_like(x, dtype=None):
-    x = convert_to_tensor(x)
-    dtype = dtype or x.dtype
     return jnp.empty_like(x, dtype=dtype)
 
 
--- a/keras/src/backend/numpy/numpy.py
+++ b/keras/src/backend/numpy/numpy.py
@@ -613,8 +613,6 @@ def empty(shape, dtype=None):
 
 
 def empty_like(x, dtype=None):
-    x = convert_to_tensor(x)
-    dtype = dtype or x.dtype
     return np.empty_like(x, dtype=dtype)
 
 
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -1490,8 +1490,6 @@ def empty(shape, dtype=None):
 
 
 def empty_like(x, dtype=None):
-    x = convert_to_tensor(x)
-    dtype = dtype or x.dtype
     return tf.zeros_like(x, dtype=dtype)
 
 
--- a/keras/src/ops/numpy.py
+++ b/keras/src/ops/numpy.py
@@ -3115,12 +3115,19 @@ def empty(shape, dtype=None):
 
 
 class EmptyLike(Operation):
-    def call(self, x, dtype=None):
-        return backend.numpy.empty_like(x, dtype=dtype)
+    def __init__(self, dtype=None, *, name=None):
+        super().__init__(name=name)
+        self.dtype = None if dtype is None else backend.standardize_dtype(dtype)
 
-    def compute_output_spec(self, x, dtype=None):
-        dtype = dtype or x.dtype
-        dtype = backend.standardize_dtype(dtype)
+    def call(self, x):
+        return backend.numpy.empty_like(x, dtype=self.dtype)
+
+    def compute_output_spec(self, x):
+        dtype = (
+            backend.standardize_dtype(x.dtype)
+            if self.dtype is None
+            else self.dtype
+        )
         return KerasTensor(x.shape, dtype=dtype)
 
 
@@ -3145,7 +3152,7 @@ def empty_like(x, dtype=None):
     dtype('float32')
     """
     if any_symbolic_tensors((x,)):
-        return EmptyLike().symbolic_call(x, dtype=dtype)
+        return EmptyLike(dtype=dtype).symbolic_call(x)
     return backend.numpy.empty_like(x, dtype=dtype)
 
 
--- a/keras/src/ops/numpy_test.py
+++ b/keras/src/ops/numpy_test.py
@@ -1498,6 +1498,7 @@ class NumpyOneInputOpsDynamicShapeTest(testing.TestCase):
     def test_empty_like(self):
         x = KerasTensor((None, 3))
         self.assertEqual(knp.empty_like(x).shape, (None, 3))
+        self.assertEqual(knp.empty_like(x).dtype, x.dtype)
 
     def test_exp(self):
         x = KerasTensor((None, 3))
@@ -2146,6 +2147,7 @@ class NumpyOneInputOpsStaticShapeTest(testing.TestCase):
     def test_empty_like(self):
         x = KerasTensor((2, 3))
         self.assertEqual(knp.empty_like(x).shape, (2, 3))
+        self.assertEqual(knp.empty_like(x).dtype, x.dtype)
 
     def test_exp(self):
         x = KerasTensor((2, 3))


keras/src/backend/jax/numpy.py
```diff
@@ -678,6 +678,11 @@ def empty(shape, dtype=None):
     return jnp.empty(shape, dtype=dtype)
 
 
+def empty_like(x, dtype=None):
+    dtype = dtype or config.floatx()
+    return jnp.empty_like(x, dtype=dtype)
```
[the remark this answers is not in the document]
shashaka: updated

keras/src/backend/numpy/numpy.py
```diff
@@ -612,6 +612,11 @@ def empty(shape, dtype=None):
     return np.empty(shape, dtype=dtype)
 
 
+def empty_like(x, dtype=None):
+    dtype = dtype or config.floatx()
+    return np.empty_like(x, dtype=dtype)
```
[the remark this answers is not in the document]
shashaka: updated

keras/src/backend/tensorflow/numpy.py
```diff
@@ -1489,6 +1489,11 @@ def empty(shape, dtype=None):
     return tf.zeros(shape, dtype=dtype)
 
 
+def empty_like(x, dtype=None):
+    dtype = dtype or config.floatx()
+    return tf.zeros_like(x, dtype=dtype)
```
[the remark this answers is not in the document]
shashaka: updated

keras/src/backend/torch/numpy.py
```diff
@@ -770,6 +770,11 @@ def empty(shape, dtype=None):
     return torch.empty(size=shape, dtype=dtype, device=get_device())
 
 
+def empty_like(x, dtype=None):
+    dtype = to_torch_dtype(dtype or config.floatx())
+    return torch.empty_like(x, dtype=dtype, device=get_device())
```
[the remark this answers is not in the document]
shashaka: updated

keras/src/ops/numpy.py
```diff
@@ -3114,6 +3114,32 @@ def empty(shape, dtype=None):
...
+def empty_like(x, dtype=None):
+    """Return a new uninitialized tensor with the same shape and dtype as `x`.
+
+    Args:
+        x: Input tensor to mimic shape and dtype.
+        dtype: Optional data type. If None, uses `x.dtype`.
+
+    Returns:
+        A tensor with the same shape and dtype as `x`, with arbitrary contents.
+    """
```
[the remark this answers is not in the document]
shashaka: updated

keras/src/backend/tensorflow/numpy.py
```diff
@@ -1489,6 +1489,12 @@ def empty(shape, dtype=None):
     return tf.zeros(shape, dtype=dtype)
 
 
+def empty_like(x, dtype=None):
+    x = convert_to_tensor(x)
+    dtype = dtype or x.dtype
+    return tf.zeros_like(x, dtype=dtype)
```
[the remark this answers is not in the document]
shashaka: There is no tf.empty_like.

hertschuh:
@shashaka 

May I ask what the rationale is for adding this op. Is the hope to get better performance than `zeros_like`?

JAX implements `empty_like` as `zeros_like` per documentation. The TensorFlow implementation uses `zeros_like`. So only the Torch backend might benefit from this under a set of specific conditions.

Also note that on JAX and TensorFlow, the compiler will be smart enough under some conditions to not materialize the array at all, which is even more performant than not initializing the values.

shashaka:
@hertschuh 
Thanks for the review.
You are absolutely right. empty_like does not provide real performance improvements for most backends since both JAX and TensorFlow handle it in the same way as zeros_like.

My main motivation was API consistency with NumPy to make it easier for users porting NumPy code into keras.ops.
However, since the practical backend benefit is minimal, I agree that it is not essential at this point.

I will go ahead and close this PR. Thank you again for the detailed explanation and feedback.

hertschuh:
Actually, after I wrote this comment, I noticed that keras already implements `empty`.  So, for consistency, yes, let's also add `empty_like`.

Please re-open and sorry for the churn.

keras/src/backend/jax/numpy.py
```diff
@@ -678,6 +678,12 @@ def empty(shape, dtype=None):
     return jnp.empty(shape, dtype=dtype)
 
 
+def empty_like(x, dtype=None):
+    x = convert_to_tensor(x)
+    dtype = dtype or x.dtype
```
hertschuh: `dtype = dtype or x.dtype` is not be needed, `jnp.empty_like` will figure this out. So remove this line.

keras/src/backend/numpy/numpy.py
```diff
@@ -612,6 +612,12 @@ def empty(shape, dtype=None):
     return np.empty(shape, dtype=dtype)
 
 
+def empty_like(x, dtype=None):
+    x = convert_to_tensor(x)
+    dtype = dtype or x.dtype
```
hertschuh:
```python
    x = convert_to_tensor(x)
    dtype = dtype or x.dtype
```
These 2 lines are not needed, remove

keras/src/backend/tensorflow/numpy.py
```diff
@@ -1489,6 +1489,12 @@ def empty(shape, dtype=None):
     return tf.zeros(shape, dtype=dtype)
 
 
+def empty_like(x, dtype=None):
+    x = convert_to_tensor(x)
+    dtype = dtype or x.dtype
```
hertschuh:
```python
    x = convert_to_tensor(x)
    dtype = dtype or x.dtype
```
Remove these 2 lines for consistency with how `zeros_like` and `ones_like` are implemented.

keras/src/ops/numpy.py
```diff
@@ -3114,6 +3114,41 @@ def empty(shape, dtype=None):
...
 
 
+class EmptyLike(Operation):
+    def call(self, x, dtype=None):
+        return backend.numpy.empty_like(x, dtype=dtype)
+
+    def compute_output_spec(self, x, dtype=None):
+        dtype = dtype or x.dtype
+        dtype = backend.standardize_dtype(dtype)
+        return KerasTensor(x.shape, dtype=dtype)
```
hertschuh, line 3131: dtype is a parameter to `__init__` not `call`. Also, copying for the other ones:

```python
class EmptyLike(Operation):
    def __init__(self, dtype=None, *, name=None):
        super().__init__(name=name)
        self.dtype = None if dtype is None else backend.standardize_dtype(dtype)

    def call(self, x):
        return backend.numpy.empty_like(x, dtype=self.dtype)

    def compute_output_spec(self, x):
        dtype = (
            backend.standardize_dtype(x.dtype)
            if self.dtype is None
            else self.dtype
        )
        return KerasTensor(x.shape, dtype=dtype)
```

keras/src/ops/numpy.py
```diff
@@ -3114,6 +3114,41 @@ def empty(shape, dtype=None):
...
+    >>> from keras import ops
+    >>> x = ops.ones((2, 3), dtype="float32")
+    >>> y = ops.empty_like(x)
+    >>> y.shape
+    (2, 3)
+    >>> y.dtype
+    dtype('float32')
+    """
+    if any_symbolic_tensors((x,)):
+        return EmptyLike().symbolic_call(x, dtype=dtype)
```
hertschuh:
```python
return EmptyLike(dtype).symbolic_call(x)
```

keras/src/ops/numpy_test.py
```diff
@@ -1495,6 +1495,10 @@ def test_dot(self):
         y = KerasTensor((5,))
         self.assertEqual(knp.dot(x, y).shape, ())
 
+    def test_empty_like(self):
+        x = KerasTensor((None, 3))
+        self.assertEqual(knp.empty_like(x).shape, (None, 3))
```
hertschuh, line 1500: Also add `self.assertEqual(knp.empty_like(x).dtype, x.dtype)`

keras/src/ops/numpy_test.py
```diff
@@ -2139,6 +2143,10 @@ def test_dot(self):
             y = KerasTensor((2, 3))
             knp.dot(x, y)
 
+    def test_empty_like(self):
+        x = KerasTensor((2, 3))
+        self.assertEqual(knp.empty_like(x).shape, (2, 3))
```
hertschuh, line 2149: Also add `self.assertEqual(knp.empty_like(x).dtype, x.dtype)`

keras/src/ops/numpy.py
shashaka: Updated.

keras/src/ops/numpy_test.py
shashaka: Updated.

keras/src/ops/numpy_test.py
shashaka: Updated.

keras/src/backend/jax/numpy.py
shashaka: Updated.

keras/src/backend/tensorflow/numpy.py
shashaka: Updated.

keras/src/ops/numpy.py
shashaka: Updated.

keras/src/backend/numpy/numpy.py
shashaka: Updated.

Issue #21771, opened 2025-10-24
`TerminateOnNaN` callback triggers `on_train_end()`, causing unintended side effects in other callbacks


PhilLord0000:
The `TerminateOnNaN` callback currently terminates training by setting `self.model.stop_training = True` when NaN or Inf loss is detected. Whilst this gracefully stops the training loop, it also triggers the `on_train_end()` method for all other callbacks, which can have unintended negative consequences.
Examples of problematic interactions:
`BackupAndRestore`: Deletes its backup directory in `on_train_end()`, preventing recovery from the last good epoch when NaN occurs
`EarlyStopping`: May restore best weights based on an incomplete training run

There might be more then that, I initially ran into this issue when using `BackupAndRestore`. I've been using this custom callback to stop the training without running the other callbacks.

```
class HardTerminateOnNaN(tf.keras.callbacks.Callback):
    def on_batch_end(self, batch, logs=None):
        loss = logs.get("loss")
        if loss is not None and (tf.math.is_nan(loss) or tf.math.is_inf(loss)):
            print(f"\nNaN detected at batch {batch}. Terminating immediately.")
            raise RuntimeError("NaN loss encountered.")
```

I propose either adding a new callback, similar to the one above or adding an option to the existing `TerminateOnNaN`. I'd be happy to open a PR and implement this based on what you think would be the better option.

dhantule:
Hi @PhilLord0000, thanks for reporting this.

Can you provide standalone code to reproduce this issue ?

PhilLord0000:
Hi @dhantule, sorry that I've take a while to get back to you, I've been very busy recently.

I've created a repo with two examples, one to reproduce the issue and one with an example of my proposed callback. 
https://github.com/PhilLord0000/TerminateOnNaN_Fix

Both examples use a normal learning rate for the first epoch and then increase the learning rate to a very high value to instigate a NaN, in `reproduce_issue` you will notice that there is no backup folder after the training is terminated however the folder persists after the termination in `fix_issue`. I hope these examples provide a clear demonstration of the issue.

MalyalaKarthik66:
Hi, @PhilLord0000 @dhantule 
I created this PR to address the issue where TerminateOnNaN only stops training gracefully, causing other callbacks like BackupAndRestore to run even after a NaN loss.

This PR introduces a new callback — HardTerminateOnNaN — that immediately raises a RuntimeError when NaN or Inf losses are detected.
[...]

PR #21841, opened 2025-11-10
Add raise_error option to TerminateOnNaN for immediate termination on NaN/Inf losses

MalyalaKarthik66:
Fix: #21771 
Added a raise_error=True option to keras.callbacks.TerminateOnNaN that raises a RuntimeError immediately when NaN or Inf losses occur.

This initial implementation focuses on strict failure behavior. It intentionally:

- Preserves the existing graceful mode (raise_error=False) for backward compatibility
- Ensures immediate termination without invoking later callbacks (e.g., BackupAndRestore)
- Guarantees consistent behavior across supported backends
- Includes unit tests covering both graceful and hard termination modes

### DIFF
--- a/keras/src/callbacks/terminate_on_nan_test.py
+++ b/keras/src/callbacks/terminate_on_nan_test.py
@@ -2,6 +2,7 @@ import os
 
 import numpy as np
 import pytest
+from absl.testing import parameterized
 
 from keras.src import callbacks
 from keras.src import initializers
@@ -89,7 +90,10 @@ class TerminateOnNaNTest(testing.TestCase):
         callback = TerminateOnNaN(raise_error=True)
 
         # Training should raise RuntimeError
-        with pytest.raises(RuntimeError, match="NaN or Inf loss encountered"):
+        with self.assertRaisesRegex(
+            RuntimeError,
+            "NaN or Inf loss encountered",
+        ):
             model.fit(
                 x, y, epochs=1, batch_size=1, callbacks=[callback], verbose=0
             )
@@ -117,7 +121,7 @@ class TerminateOnNaNTest(testing.TestCase):
         raise_error_terminate_callback = TerminateOnNaN(raise_error=True)
 
         # Should raise RuntimeError
-        with pytest.raises(RuntimeError):
+        with self.assertRaises(RuntimeError):
             model.fit(
                 x,
                 y,
@@ -154,7 +158,7 @@ class TerminateOnNaNTest(testing.TestCase):
         backup_callback.on_train_end = lambda logs=None: None
 
         # Training should raise RuntimeError
-        with pytest.raises(RuntimeError):
+        with self.assertRaises(RuntimeError):
             model.fit(
                 x_nan,
                 y_nan,
@@ -173,7 +177,11 @@ class TerminateOnNaNTest(testing.TestCase):
             "Backup file missing unexpectedly.",
         )
 
-    def test_normal_training_does_not_raise(self):
+    @parameterized.named_parameters(
+        ("raise_error_false", False),
+        ("raise_error_true", True),
+    )
+    def test_normal_training_does_not_raise(self, raise_error):
         """Test that TerminateOnNaN does not raise on normal training."""
         model = models.Sequential([layers.Dense(1, input_shape=(1,))])
         model.compile(optimizer="sgd", loss="mse")
@@ -181,15 +189,13 @@ class TerminateOnNaNTest(testing.TestCase):
         x = np.array([[1.0], [2.0]])
         y = np.array([[1.0], [2.0]])
 
-        # Test both raise_error=False and raise_error=True with normal data
-        for raise_error in [False, True]:
-            callback = TerminateOnNaN(raise_error=raise_error)
+        callback = TerminateOnNaN(raise_error=raise_error)
 
-            # Should complete without raising RuntimeError
-            history = model.fit(x, y, epochs=2, callbacks=[callback], verbose=0)
+        # Should complete without raising RuntimeError
+        history = model.fit(x, y, epochs=2, callbacks=[callback], verbose=0)
 
-            # Should have completed 2 epochs
-            self.assertEqual(len(history.history["loss"]), 2)
+        # Should have completed 2 epochs
+        self.assertEqual(len(history.history["loss"]), 2)
 
     def test_raise_error_terminate_stops_on_later_batch(self):
         """Ensure TerminateOnNaN(raise_error=True) stops training
@@ -204,9 +210,9 @@ class TerminateOnNaNTest(testing.TestCase):
 
         callback = TerminateOnNaN(raise_error=True)
 
-        with pytest.raises(RuntimeError) as exc:
+        with self.assertRaises(RuntimeError) as exc:
             model.fit(
                 x, y, epochs=1, batch_size=1, callbacks=[callback], verbose=0
             )
 
-        assert any(f"batch {i}" in str(exc.value) for i in [0, 1])
+        self.assertTrue(any(f"batch {i}" in str(exc.exception) for i in [0, 1]))


fchollet:
Thanks for the PR! This sounds like it should be an option in `TerminateOnNaN` instead of being a separate callback.

MalyalaKarthik66:
@fchollet 
Thanks for the review! I’ve updated the PR.

keras/src/callbacks/hard_terminate_on_nan_test.py
```diff
@@ -0,0 +1,186 @@
...
+from keras.src import testing
+from keras.src.callbacks import BackupAndRestore
+from keras.src.callbacks import TerminateOnNaN
+
+
+@pytest.mark.skipif(
+    backend.backend() in ["numpy", "openvino"],
+    reason="TerminateOnNaN not supported for NumPy or OpenVINO backend",
+)
+class TerminateOnNaNTest(testing.TestCase):
```
hertschuh: There are some tests in `keras/src/callbacks/terminate_on_nan_test.py` already.

Please move the new tests there and combine them.

keras/src/callbacks/terminate_on_nan.py
```diff
@@ -7,14 +7,63 @@
...
+    callback = keras.callbacks.TerminateOnNaN()
+    model.fit(x, y, callbacks=[callback])
+
+    # Hard termination (strict failure)
+    callback = keras.callbacks.TerminateOnNaN(hard=True)
+    model.fit(x, y, callbacks=[callback])
+    ```
+    """
+
+    def __init__(self, hard: bool = False):
```
hertschuh: How about we call the new argument `raise_error` instead of `hard`? Hopefully that name makes it obvious how it's different.

MalyalaKarthik66:
@hertschuh 
Thanks for the review! I’ve updated the PR with the requested changes — tests are merged, the argument is renamed to raise_error.

keras/src/callbacks/terminate_on_nan.py
```diff
@@ -7,14 +7,64 @@
...
+    # raise_error termination (strict failure)
+    callback = keras.callbacks.TerminateOnNaN(raise_error=True)
+    model.fit(x, y, callbacks=[callback])
+    ```
+    """
+
+    def __init__(self, raise_error: bool = False):
+        super().__init__()
+        self.raise_error = raise_error
+        self._supports_tf_logs = True
```
hertschuh: `_supports_tf_logs` is unused. Remove.

keras/src/callbacks/terminate_on_nan_test.py
```diff
@@ -50,3 +59,156 @@ def test_TerminateOnNaN(self):
...
+            model.fit(
+                x, y, epochs=1, batch_size=1, callbacks=[callback], verbose=0
+            )
+
+    def test_raise_error_terminate_does_not_trigger_on_train_end(self):
+        """Test that on_train_end is NOT called when
+        TerminateOnNaN(raise_error=True) raises.
+        """
+
+        class TrackingCallback(keras.src.callbacks.Callback):
```
hertschuh: Use `callbacks.Callback` directly (it's already imported) instead of `keras.src.callbacks.Callback`.

keras/src/callbacks/terminate_on_nan_test.py
```diff
@@ -1,16 +1,25 @@
+import os
+import tempfile
+
 import numpy as np
 import pytest
 
+import keras
```
hertschuh: Remove `import keras`. In units tests, we import selectively from `keras.src`.

keras/src/callbacks/terminate_on_nan_test.py
```diff
@@ -50,3 +59,156 @@ def test_TerminateOnNaN(self):
...
+            )
+
+        # on_train_end should NOT have been called
+        self.assertFalse(tracking_callback.train_end_called)
+
+    def test_raise_error_terminate_preserves_backup(self):
+        """Ensure BackupAndRestore directory is preserved when
+        TerminateOnNaN(raise_error=True) triggers.
+        """
+        with tempfile.TemporaryDirectory() as tmpdir:
```
hertschuh: Please use `self.get_temp_dir()` instead. The nice thing about it is that you don't need a `with` and you don't need to indent all the code. Instead, `self.get_temp_dir()` does the cleanup automatically at the end of the test.

For example: https://github.com/keras-team/keras/blob/master/keras/src/callbacks/model_checkpoint_test.py#L55

MalyalaKarthik66:
@hertschuh 
Thanks! The PR has been updated.

keras/src/callbacks/terminate_on_nan_test.py
```diff
@@ -50,3 +57,156 @@ def test_TerminateOnNaN(self):
...
+        model = models.Sequential([layers.Dense(1, input_shape=(1,))])
+        model.compile(optimizer="sgd", loss="mse")
+
+        x = np.array([[1.0], [2.0]])
+        y = np.array([[np.inf], [np.inf]])
+
+        callback = TerminateOnNaN(raise_error=True)
+
+        # Training should raise RuntimeError
+        with pytest.raises(RuntimeError, match="NaN or Inf loss encountered"):
```
hertschuh:
```python
with self.assertRaisesRegex(RuntimeError, match="NaN or Inf loss encountered"):
```

keras/src/callbacks/terminate_on_nan_test.py
```diff
@@ -50,3 +57,156 @@ def test_TerminateOnNaN(self):
...
+        y_nan = np.array([[np.inf]])
+
+        raise_error_terminate_callback = TerminateOnNaN(raise_error=True)
+        backup_callback = BackupAndRestore(backup_dir=backup_dir)
+
+        # Monkeypatch BackupAndRestore to prevent cleanup on train_end
+        backup_callback.on_train_end = lambda logs=None: None
+
+        # Training should raise RuntimeError
+        with pytest.raises(RuntimeError):
```
hertschuh:
```python
with self.assertRaises(RuntimeError):
```

keras/src/callbacks/terminate_on_nan_test.py
```diff
@@ -50,3 +57,156 @@ def test_TerminateOnNaN(self):
...
+        model.compile(optimizer="sgd", loss="mse")
+
+        x = np.array([[1.0]])
+        y = np.array([[np.inf]])
+
+        tracking_callback = TrackingCallback()
+        raise_error_terminate_callback = TerminateOnNaN(raise_error=True)
+
+        # Should raise RuntimeError
+        with pytest.raises(RuntimeError):
```
hertschuh:
```python
with self.assertRaises(RuntimeError):
```

keras/src/callbacks/terminate_on_nan_test.py
```diff
@@ -50,3 +57,156 @@ def test_TerminateOnNaN(self):
...
+        model = models.Sequential([layers.Dense(1, input_shape=(1,))])
+        model.compile(optimizer="sgd", loss="mse")
+
+        # Batch 1: normal loss, Batch 2: NaN loss
+        x = np.array([[1.0], [2.0]])
+        y = np.array([[1.0], [np.inf]])  # NaN/Inf appears only in 2nd batch
+
+        callback = TerminateOnNaN(raise_error=True)
+
+        with pytest.raises(RuntimeError) as exc:
```
hertschuh:
```python
with self.assertRaises(RuntimeError):
```

keras/src/callbacks/terminate_on_nan_test.py
```diff
@@ -50,3 +57,156 @@ def test_TerminateOnNaN(self):
...
+    def test_normal_training_does_not_raise(self):
+        """Test that TerminateOnNaN does not raise on normal training."""
+        model = models.Sequential([layers.Dense(1, input_shape=(1,))])
+        model.compile(optimizer="sgd", loss="mse")
+
+        x = np.array([[1.0], [2.0]])
+        y = np.array([[1.0], [2.0]])
+
+        # Test both raise_error=False and raise_error=True with normal data
+        for raise_error in [False, True]:
```
hertschuh: Please use a parameterized test instead. This way we can immediately know which case fails.

Here is an example of how to do this.

https://github.com/keras-team/keras/blob/master/keras/src/trainers/compile_utils_test.py#L320

keras/src/callbacks/terminate_on_nan_test.py
```diff
@@ -50,3 +57,156 @@ def test_TerminateOnNaN(self):
...
+        y = np.array([[1.0], [np.inf]])  # NaN/Inf appears only in 2nd batch
+
+        callback = TerminateOnNaN(raise_error=True)
+
+        with pytest.raises(RuntimeError) as exc:
+            model.fit(
+                x, y, epochs=1, batch_size=1, callbacks=[callback], verbose=0
+            )
+
+        assert any(f"batch {i}" in str(exc.value) for i in [0, 1])
```
hertschuh: `self.assertTrue(any(...))`

MalyalaKarthik66:
@hertschuh 
Thanks for pointing it out! I’ve applied the changes.

hertschuh (approved):

PR #21842, opened 2025-11-10
Support jax2tf in JaxLayer for tf backend

wenyi-guo:
Support both tf and jax backend for JaxLayer

### DIFF
--- a/keras/src/utils/jax_layer_test.py
+++ b/keras/src/utils/jax_layer_test.py
@@ -185,6 +185,7 @@ if flax is not None:
     backend.backend() not in ["jax", "tensorflow"],
     reason="JaxLayer and FlaxLayer are only supported with JAX and TF backend",
 )
+@pytest.mark.skipif(testing.tensorflow_uses_gpu(), reason="GPU test failure")
 class TestJaxLayer(testing.TestCase):
     def _test_layer(
         self,


keras/src/utils/jax_layer_test.py
```diff
@@ -478,7 +487,12 @@ def create_wrapper(**kwargs):
             flax_model = flax_model_class()
             if flax_model_method:
                 kwargs["method"] = getattr(flax_model, flax_model_method)
-            return FlaxLayer(flax_model_class(), **kwargs)
+            if backend.backend() == "jax":
+                return FlaxLayer(flax_model_class(), **kwargs)
```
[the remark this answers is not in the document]
wenyi-guo: TF needs additional arg stateless_compute_output_shape

keras/src/utils/jax_layer.py
```diff
@@ -196,6 +214,9 @@ def my_haiku_module_fn(inputs, training):
         init_fn: the function to call to initialize the model. See description
             above for the list of arguments it takes and the outputs it returns.
             If `None`, then `params` and/or `state` must be provided.
+        compute_output_shape_fn: Function that takes the input shape
+            (a tuple or nested structure of tuples) and returns the output
+            shape (a tuple or nested structure of tuples).
```
hertschuh: Why was this needed? Also, it looks like it's for TensorFlow only?

If `compute_output_shape` is not implemented, Keras should trace the layer and figure it out automatically, which is a better user experience.

keras/src/layers/layer.py
```diff
@@ -1145,7 +1145,10 @@ def compute_output_spec(self, *args, **kwargs):
                 call_spec=call_spec,
                 class_name=self.__class__.__name__,
             )
-            output_shape = self.compute_output_shape(**shapes_dict)
+            try:
+                output_shape = self.compute_output_shape(**shapes_dict)
+            except NotImplementedError:
+                return super().compute_output_spec(*args, **kwargs)
```
hertschuh: This should not be needed because of the `if` line 1134. Revert.

keras/src/utils/jax_layer.py
```diff
@@ -214,15 +235,16 @@ def __init__(
...
         seed=None,
         **kwargs,
     ):
-        if backend.backend() != "jax":
+        if backend.backend() not in ["jax", "tensorflow"]:
             raise ValueError(
-                "JaxLayer is only supported with the JAX backend. Current "
-                f"backend: {backend.backend()}"
+                "JaxLayer is only supported with the JAX or Tensorflow backend"
+                f". Current backend: {backend.backend()}"
```
hertschuh: I saw you removed the equivalent error message in `FlaxLayer`. Can you do this so that it works properly for both:
```python
f"{self.__class__.__name__} is only supported with the JAX ..."
```

keras/src/utils/jax_layer.py
```diff
@@ -233,7 +255,10 @@ def __init__(
         super().__init__(**kwargs)
         self.call_fn = call_fn
         self.init_fn = init_fn
-        self.seed_generator = backend.random.SeedGenerator(seed)
+        self.compute_output_shape_fn = compute_output_shape_fn
+        if seed is None:
+            seed = random.seed_generator.make_default_seed()
+        self.jax_rng = jax.random.PRNGKey(seed)
```
hertschuh: Actually, the seed handling that was there before (with `seed_generator`) was correct. The approach with `_split_jax_rng_key` won't work once it's compiled. So revert all related code.

In fact, we'll probably need to fix the Keras 2 version.

keras/src/utils/jax_layer.py
```diff
@@ -296,14 +398,18 @@ def _create_variables(self, values, trainable):
 
         def create_variable(value):
             if backend.is_tensor(value) or isinstance(
-                value, (np.ndarray, np.generic)
+                value, (np.ndarray, np.generic, jax.Array)
```
hertschuh, line 396: `jax.Array` should not be needed because `backend.is_tensor(value)` will be True in this case.

keras/src/utils/jax_layer.py
```diff
@@ -296,14 +398,18 @@ def _create_variables(self, values, trainable):
...
                 if is_float_dtype(dtype):
                     dtype = None  # Use the layer dtype policy
                 return self.add_weight(
                     value.shape,
-                    initializer=value,
+                    initializer=(
+                        backend.convert_to_tensor(value)
+                        if value is not None
+                        else None
+                    ),
```
hertschuh: I don't think this is needed. And in particular, `None` is not possible.

keras/src/utils/jax_layer.py
```diff
@@ -328,8 +434,15 @@ def create_variable(value):
         else:
             self.state = variables
 
-        flat_variables, _ = jax.tree_util.tree_flatten(variables)
-        return flat_variables
+        if backend.backend() == "jax":
+            flat_variables, _ = jax.tree_util.tree_flatten(variables)
+            return flat_variables
+        elif backend.backend() == "tensorflow":
+            return variables
```
hertschuh: Why the different behaviors for JAX and TensorFlow here?

keras/src/utils/jax_layer.py
```diff
@@ -328,8 +434,15 @@ def create_variable(value):
...
+
+    def _split_jax_rng(self):
+        self.jax_rng, subkey = jax.random.split(self.jax_rng)
+        return subkey
```
hertschuh: Per comment above, revert

keras/src/utils/jax_layer.py
```diff
@@ -429,14 +582,39 @@ def assign_state_to_variable(value, variable):
...
+                    return call_with_fn(self.jax2tf_training_true_fn)
+                else:
+                    return call_with_fn(self.jax2tf_training_false_fn)
+
+    def compute_output_shape(self, input_shape):
+        if self.compute_output_shape_fn:
+            return self.compute_output_shape_fn(input_shape)
         else:
-            return self.call_fn(*call_args)
+            return super().compute_output_shape(input_shape)
```
hertschuh: You should be able to remove this and the default implementation should work.

keras/src/utils/jax_layer.py
```diff
@@ -429,14 +582,39 @@ def assign_state_to_variable(value, variable):
...
+        if backend.backend() == "jax":
+            return call_with_fn(self.call_fn)
+        elif backend.backend() == "tensorflow":
+            if self.jax2tf_training_true_fn is None:
+                return call_with_fn(self.jax2tf_training_false_fn)
+            else:
+                if training:
+                    return call_with_fn(self.jax2tf_training_true_fn)
+                else:
+                    return call_with_fn(self.jax2tf_training_false_fn)
```
hertschuh: nitpick: this can be done in fewer branches:

```python
if training and self.jax2tf_training_true_fn is None:
    return call_with_fn(self.jax2tf_training_true_fn)
else:
    return call_with_fn(self.jax2tf_training_false_fn)
```

keras/src/utils/jax_layer_test.py
```diff
@@ -478,7 +487,12 @@ def create_wrapper(**kwargs):
...
+            elif backend.backend() == "tensorflow":
+                return FlaxLayer(
+                    flax_model, stateless_compute_output_shape, **kwargs
+                )
```
hertschuh: Ideally we wouldn't need this branch

keras/src/utils/jax_layer_test.py
```diff
@@ -194,16 +201,18 @@ def _test_layer(
...
-            (np.random.uniform(size=(320,)) * num_classes).astype("int32")
-        ]
-        x_test = np.random.uniform(size=(32, 28, 28, 1))
+        x_train = random.uniform(shape=(320, 28, 28, 1))
+        y_train_indices = ops.cast(
+            ops.random.uniform(shape=(320,), minval=0, maxval=num_classes),
+            dtype="int32",
+        )
+        y_train = ops.one_hot(y_train_indices, num_classes, dtype="int32")
+        x_test = random.uniform(shape=(32, 28, 28, 1))
```
hertschuh, line 207: Just curious why you had to change this.

keras/src/utils/jax_layer_test.py
```diff
@@ -194,16 +201,18 @@ def _test_layer(
...
 
         def _count_params(weights):
             count = 0
             for weight in weights:
-                count = count + np.prod(weight.shape)
+                count = count + ops.prod(ops.shape(weight))
```
hertschuh: nitpick: here we only need ints, not tensors. But it still works?

it could be:

```
count = count + math.prod(ops.shape(weight))
```

keras/src/utils/jax_layer.py
```diff
@@ -429,14 +582,39 @@ def assign_state_to_variable(value, variable):
...
+                if backend.backend() == "jax":
+                    jax.tree_util.tree_map(
+                        assign_state_to_variable, new_state, self.state
+                    )
+                elif backend.backend() == "tensorflow":
+                    jax.tree_util.tree_map(
+                        assign_state_to_variable,
+                        standardize_pytree_collections(new_state),
+                        standardize_pytree_collections(self.state),
+                    )
```
hertschuh: Oh, this is the only context where `standardize_pytree_collections` is needed?

In this case try this, I think it should work and will support more use cases than `standardize_pytree_collections`:

```python
keras.tree.map_structure(assign_state_to_variable, new_state, self.state)
```

hertschuh (commented):
Thanks for the PR!

keras/src/utils/jax_layer.py
```diff
@@ -359,23 +472,23 @@ def _get_call_rng(self, training):
...
+            raise ValueError(
+                "'JaxLayer' cannot be built in tracing scope"
+                "or inside tf function"
+            )
 
         # Initialize `params` and `state` if needed by calling `init_fn`.
         def create_input(shape):
             shape = [d if d is not None else 1 for d in shape]
-            return jax.numpy.ones(shape)
+            return ops.ones(shape)
```
hertschuh: Actually, this should to be `jax.numpy.ones`, we're always calling JAX with this, even when the backend is Tensorflow.

keras/src/utils/jax_layer.py
hertschuh: Correction, `convert_to_tensor` is needed, but not the handling of `None`.

keras/src/utils/jax_layer.py
hertschuh: Correction, this is needed for the TensorFlow backend.

wenyi-guo: Actually if it's tf backend, `backend.is_tensor(value)` returns False for jax initializer initialized values. So need to add jax.Array.

keras/src/utils/jax_layer.py
wenyi-guo: I tried it. Usage wise, it works. But it would raise different exception when it happens. For test wise, if would be different from jax test for mismatch assertion exception. e.g
`FAILED keras/src/utils/jax_layer_test.py::TestJaxLayer::test_state_mismatch_during_update_mapping_instead_of_sequence - AssertionError: "Expected list, got " does not match "Expected an instance of dict, collections.OrderedDict, or collections.defaultdict, got DictWrapper({'state': DictWrapper({'foo': <Variable path=jax_layer/variable, shape=(), dtype=float32, value=0.0>})})."`

Do you want to modify the implementation or the test?

hertschuh: Hmm... it looks like something else throws an error. Where does the error come from if you comment out
```python
with self.assertRaisesRegex(ValueError, error_regex):
```
I'm talking about this error `Expected an instance of dict, collections.OrderedDict, or collections.defaultdict, got DictWrapper`

Also, is this with the JAX backend or the TF backend?

keras/src/utils/jax_layer.py
wenyi-guo: Reverted to only one branch

keras/src/utils/jax_layer_test.py
wenyi-guo: Oh I thought with tf backend, users might want to test tf data

keras/src/utils/jax_layer.py
wenyi-guo: It's the same for both JAX and TF.
For the `mismatch` test cases,  it would raise the error from op tree called by keras tree. It's just different exception messages, but same behavior.

```
keras/src/utils/jax_layer.py:571: in call_with_fn
    tree.map_structure(assign_state_to_variable, new_state, self.state)
keras/src/tree/tree_api.py:200: in map_structure
    return tree_impl.map_structure(func, *structures, none_is_leaf=none_is_leaf)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
keras/src/tree/optree_impl.py:111: in map_structure
    return optree.tree_map(

leaves, treespec = _C.flatten(tree, is_leaf, none_is_leaf, namespace)
>       flat_args = [leaves] + [treespec.flatten_up_to(r) for r in rests]
                                ^^^^^^^^^^^^^^^^^^^^^^^^^
E       ValueError: Expected an instance of list, got {'foo': <Variable path=jax_layer/variable, shape=(), dtype=float32, value=0.0>}.
```

However, for tests with None or Custom object. It has different behavior. e.g. tests from `test_with_state_none_leaves` and `test_with_state_jax_registered_node_class`, it would also raise `Structure mismatch` error though we don't want that.
```
value = None, variable = None

    def assign_state_to_variable(value, variable):
        # This exists only to make debugging this error case easier.
        if not hasattr(variable, "assign"):
>           raise ValueError(
                "Structure mismatch: the structure of the state returned "
                "by `call` does not match the structure of the state at "
                "initialization time."
            )
E           ValueError: Structure mismatch: the structure of the state returned by `call` does not match the structure of the state at initialization time.
```
```
value = <keras.src.utils.jax_layer_test.TestJaxLayer.test_with_state_jax_registered_node_class.<locals>.NamedPoint object at 0x3292e7d70>
variable = <keras.src.utils.jax_layer_test.TestJaxLayer.test_with_state_jax_registered_node_class.<locals>.NamedPoint object at 0x10e9a6b10>

    def assign_state_to_variable(value, variable):
        # This exists only to make debugging this error case easier.
        if not hasattr(variable, "assign"):
>           raise ValueError(
                "Structure mismatch: the structure of the state returned "
                "by `call` does not match the structure of the state at "
                "initialization time."
            )
E           ValueError: Structure mismatch: the structure of the state returned by `call` does not match the structure of the state at initialization time.
```

hertschuh: Doesn't replacing `jax.tree_util.tree_map` by `keras.tree.map_structure` in both the TF and the JAX case fix this ?

wenyi-guo: `keras.tree.map_structure` would cause different behavior for values of None or custom object as I shared above.

keras/src/export/saved_model.py
```diff
@@ -540,12 +540,6 @@ def write_out(self, filepath, options=None, verbose=True):
             )
 
     def _convert_to_tf_variable(self, backend_variable):
-        if not isinstance(backend_variable, backend.Variable):
-            raise TypeError(
-                "`backend_variable` must be a `backend.Variable`. "
-                f"Recevied: backend_variable={backend_variable} of type "
-                f"({type(backend_variable)})"
-            )
```
hertschuh, line 548: Undo this file. This should work again.

keras/src/utils/jax_layer.py
```diff
@@ -650,13 +807,13 @@ def _variables_to_params_and_state(self, variables):
 
     def _get_init_rng(self):
         return {
-            "params": self.seed_generator.next(),
-            "dropout": self.seed_generator.next(),
+            "params": super()._get_init_rng(),
+            "dropout": super()._get_init_rng(),
```
hertschuh: Actually, undo this. Now that the conversion is no longer done in `_get_init_rng` is makes sense to keep this as it was.

keras/src/utils/jax_layer.py
```diff
@@ -650,13 +807,13 @@ def _variables_to_params_and_state(self, variables):
...
         }
 
     def _get_call_rng(self, training):
         if training:
-            return {"dropout": self.seed_generator.next()}
+            return {"dropout": super()._get_call_rng(training)}
```
hertschuh: Undo this, this is not needed. It makes it easier to understand how to write an override.

keras/src/utils/jax_layer.py
```diff
@@ -233,7 +250,13 @@ def __init__(
         super().__init__(**kwargs)
         self.call_fn = call_fn
         self.init_fn = init_fn
-        self.seed_generator = backend.random.SeedGenerator(seed)
+        if backend.backend() == "jax":
+            dtype = jax.numpy.uint32
+        elif backend.backend() == "tensorflow":
+            dtype = tf.uint32
```
hertschuh: Replace lines 253-256 this with:

```python
from keras.src.backend.jax.core import random_seed_dtype

# We need to use the JAX seed dtype even when the backend in TensorFlow.
dtype= random_seed_dtype()
```

This will work for both backends.

keras/src/utils/jax_layer.py
```diff
@@ -342,8 +442,15 @@ def _get_init_rng(self):
         Returns:
             a JAX `PRNGKey` or structure of `PRNGKey`s that will be passed as
             the `rng` argument of `init_fn`.
+
```
hertschuh: Please update docstring of both `_get_init_rng` and `_get_call_rng`. They now returns an array of size 2, not a PRNGKey.

keras/src/utils/jax_layer.py
```diff
@@ -342,8 +442,15 @@ def _get_init_rng(self):
...
         """
-        return self.seed_generator.next()
+
+        from keras.src.backend.jax.core import convert_to_tensor
+
+        if backend.backend() == "jax":
+            return self.seed_generator.next()
+        if backend.backend() == "tensorflow":
+            return convert_to_tensor(self.seed_generator.next())
```
hertschuh: Actually undo this and do `return self.seed_generator.next()`. It's how it's documented right now.

Let's do the conversion where it's used to make it easier to override this (see my other comment).

keras/src/utils/jax_layer.py
hertschuh: Let's do the conversion here to make it easy for users to override `_get_init_rng()`.
```python
from keras.src.backend.jax.core import convert_to_tensor

init_args.append(jax.tree_util.tree_map(convert_to_tensor, self._get_init_rng()))
```

keras/src/utils/jax_layer.py
hertschuh: Here is what I don't like about the current situation. The whole point of using `jax.tree_util.tree_map` is that you can use custom data structures that are registered with `jax.tree_util`. But `standardize_pytree_collections` doesn't support those.

There is a workaround though. What we want is to leave the structures intact, we don't want `_DictWrapper`.

Right now we have this:
```python
@tracking.no_automatic_dependency_tracking
def _create_variables(self, values, trainable):
```

What we need is
```python
@tracking.no_automatic_dependency_tracking
@tf.__internal__.tracking.no_automatic_dependency_tracking
def _create_variables(self, values, trainable):
```

However, we don't want to import `tf` if the backend is JAX.

Here is what I did in KerasRS:
https://github.com/keras-team/keras-rs/blob/main/keras_rs/src/utils/keras_utils.py#L8-L27

But I think you can do something simpler here:
```python
if keras.backend.backend() == "tensorflow":
   tf_no_automatic_dependency_tracking = tf.__internal__.tracking.no_automatic_dependency_tracking
else:
   def tf_no_automatic_dependency_tracking(fn):
     return fn
...

@tracking.no_automatic_dependency_tracking
@tf_no_automatic_dependency_tracking
def _create_variables(self, values, trainable):
```

hertschuh (commented):
This is getting close!

keras/src/utils/jax_layer.py
wenyi-guo: TF `random_seed_dtype` is int64?

keras/src/utils/jax_layer.py
wenyi-guo: Thanks for the suggestion! It works. 
I'm curiosu why `@tracking.no_automatic_dependency_tracking` during TF backend doesn't skip tracking the variables as `ListWrapper` and `DictWrapper`?

hertschuh: Keras has it's own tracking in `keras/src/utils/tracking.py` that tracks via `TrackedList` and `TrackedDict`. It is disabled via `@tracking.no_automatic_dependency_tracking`. This is needed to track variables when not using the TF backend.

TensorFlow has a separate tracking mechanism that creates `ListWrapper` and `_DictWrapper`. It is disabled via `@tf.__internal__.tracking.no_automatic_dependency_tracking`.

I'm not sure how they don't interfere with one another.

keras/src/utils/jax_layer.py
hertschuh: Yes, but we don't want that. we want `uint32` always. Note that what I'm suggesting is not `keras.src.backend.random_seed_dtype`, it's `keras.src.backend.jax.core.random_seed_dtype`.

Using `keras.src.backend.jax.core.random_seed_dtype` does 2 things:
- It makes it so that it's always `uint32`, but it's not hardcoded
- It uses the string `"uint32"` instead of the backend specific dtypes `jax.numpy.uint32` / `tf.unit32` so it will work for both backends.

keras/src/utils/jax_layer.py
```diff
@@ -12,6 +16,28 @@
...
+
+def standardize_pytree_collections(pytree):
+    if isinstance(pytree, (str, bytes)):
+        return pytree
+    if isinstance(pytree, collections.abc.Mapping):
+        return {k: standardize_pytree_collections(v) for k, v in pytree.items()}
+    elif isinstance(pytree, collections.abc.Sequence):
+        return [standardize_pytree_collections(v) for v in pytree]
+    else:
+        return pytree
```
hertschuh: Remove.

keras/src/utils/jax_layer.py
```diff
@@ -333,44 +443,50 @@ def create_variable(value):
...
 
         Returns:
-            a JAX `PRNGKey` or structure of `PRNGKey`s that will be passed as
-            the `rng` argument of `init_fn`.
+            a key as an Jax or TF array of size 2 dtype uint32 will be passed
+            as the `rng` argument of `init_fn`.
         """
-        return self.seed_generator.next()
+        next = self.seed_generator.next()
+        print("!!type", type(next), next)
```
hertschuh: Remove `print`.

keras/src/utils/jax_layer.py
```diff
@@ -7,11 +10,24 @@
 from keras.src.api_export import keras_export
 from keras.src.backend.common.variables import is_float_dtype
 from keras.src.backend.common.variables import standardize_dtype
+from keras.src.backend.jax.core import (
+    random_seed_dtype as jax_random_seed_dtype,
+)
```
hertschuh: This will fail the integration tests, it needs to be a local import.

keras/src/utils/jax_layer.py
```diff
@@ -233,7 +249,14 @@ def __init__(
         super().__init__(**kwargs)
         self.call_fn = call_fn
         self.init_fn = init_fn
-        self.seed_generator = backend.random.SeedGenerator(seed)
+        if backend.backend() == "jax":
+            dtype = jax.numpy.uint32
+        elif backend.backend() == "tensorflow":
+            dtype = tf.uint32
```
hertschuh: You forgot to remove these lines:
```python
            dtype = jax.numpy.uint32
        elif backend.backend() == "tensorflow":
            dtype = tf.uint32
```

hertschuh (approved):

keras/src/utils/jax_layer.py
```diff
@@ -12,6 +15,26 @@
...
+        tf.__internal__.tracking.no_automatic_dependency_tracking
+    )
+else:
+
+    def tf_no_automatic_dependency_tracking(fn):
+        return fn
+
+
+def _convert_to_jax_key(tensor):
+    if backend.backend() == "tensorflow":
```
hertschuh, line 31: Oh, you need `else: return tensor` in the JAX backend case.

keras/src/random/seed_generator.py
```diff
@@ -84,10 +85,11 @@ def seed_initializer(*args, **kwargs):
             return self.backend.convert_to_tensor([seed, 0], dtype=dtype)
 
         with self.backend.name_scope(self.name, caller=self):
+            dtype = dtype if dtype else self.backend.random_seed_dtype()
```
hertschuh: Undo this whole file. We don't want the ability to override `dtype` based on this Tensorflow pitfall.

keras/src/utils/jax_layer.py
```diff
@@ -413,11 +560,16 @@ def unwrap_variable(variable):
                     jax.tree_util.tree_map(unwrap_variable, self.state)
                 )
             elif argument_name == "rng":
-                call_args.append(self._get_call_rng(training))
+                call_args.append(
+                    jax.tree_util.tree_map(
+                        _convert_to_jax_key, self._get_call_rng(training)
```
hertschuh, line 561: Oh, this is run in the same context as the backend. It should stay a TF tensor when the backend is tensorflow. Otherwise it will get converted twice TF -> JAX -> TF (the second conversion would be implicit).

So you need to do something different for init and call. For calling init, you need to convert to a JAX array even on Tensorflow. For call, you need the array of the right dtype and shape for JAX, but as a TF tensor with the tensorflow backend and a JAX array with the jax backend.

wenyi-guo: From the test, I see even with TF backend, it expects a jax key. jax2tf expects a JAX key before conerting to TF. Without `_convert_to_jax_key`, there's the error:
```
keras/src/utils/jax_layer.py:587: in call_with_fn
    predictions, new_state = fn(*call_args)
                             ^^^^^^^^^^^^^^
venv/lib/python3.12/site-packages/jax/experimental/jax2tf/jax2tf.py:450: in converted_fun_tf
    impl.before_conversion()
venv/lib/python3.12/site-packages/jax/experimental/jax2tf/jax2tf.py:558: in before_conversion
    self.exported = _export._export_internal(
venv/lib/python3.12/site-packages/jax/_src/export/_export.py:601: in do_export
    traced = fun_jit.trace(*args_specs, **kwargs_specs)
keras/src/utils/jax_layer.py:372: in wrapper
    return fn(*args)
           ^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

E       ValueError: The ``rngs`` argument passed to an apply function should be a ``jax.PRNGKey`` or a dictionary mapping strings to ``jax.PRNGKey``.
E       --------------------
E       For simplicity, JAX has removed its internal frames from the traceback of the following exception. Set JAX_TRACEBACK_FILTERING=off to include these.
```

hertschuh: Oh, I meant just for the call args, not for init args. And I meant that the bitcast is needed for both.

Here is what I did:
```python
def _convert_to_jax_key(tensor):
        return tf.bitcast(tensor, tf.uint32)[0]
    return tensor
...
                init_args.append(
                    jax.tree_util.tree_map(
                        lambda x: jax.numpy.array(_convert_to_jax_key(x)),
                        self._get_init_rng(),
                    )
                )
```

And the tests pass.

```diff
--- a/keras/src/utils/jax_layer.py
+++ b/keras/src/utils/jax_layer.py
@@ -29,12 +29,7 @@ else:
 
 def _convert_to_jax_key(tensor):
     if backend.backend() == "tensorflow":
-        if tensor.dtype == tf.int64:
-            key_uint32 = tf.bitcast(tensor, tf.uint32)[0]
-            if tf.is_symbolic_tensor(key_uint32):
-                return key_uint32
-            else:
-                return jax.numpy.array(key_uint32)
+        return tf.bitcast(tensor, tf.uint32)[0]
     return tensor
 
 
@@ -487,7 +482,8 @@ class JaxLayer(Layer):
             if argument_name == "rng":
                 init_args.append(
                     jax.tree_util.tree_map(
-                        _convert_to_jax_key, self._get_init_rng()
+                        lambda x: jax.numpy.array(_convert_to_jax_key(x)),
+                        self._get_init_rng(),
                     )
                 )
             elif argument_name == "inputs":
```

Init needs JAX arrays because it's run with JAX.

But call if fully TensorFlow end to end, it just needs the RNG in the right dtype to work.

PR #21847, opened 2025-11-13
📝docs: clarify shuffle behavior and example in PyDataset

maitry63:
This PR updates the `PyDataset` class docstring inside `keras/src/trainers/data_adapters/py_dataset_adapter.py` to include the  documentation for the `shuffle´ argument.

### Context
Addresses issue #20142.

### Notes
- Adds clarification about `shuffle´  default behavior (False).
- Keeps the existing example intact, adding only the `shuffle´  usage line.
- No functional code changes.

### DIFF
--- a/keras/src/trainers/data_adapters/py_dataset_adapter.py
+++ b/keras/src/trainers/data_adapters/py_dataset_adapter.py
@@ -42,7 +42,7 @@ class PyDataset:
             your dataset. Defaults to 10.
         shuffle: Whether to shuffle the sample ordering at the end of
             each epoch. This argument is passed to `model.fit()`. When
-            `model.fit(..., shuffle=True)`, the training loop
+            calling `model.fit(..., shuffle=True)`, the training loop
             automatically calls `on_epoch_end()` at each epoch
             boundary, allowing datasets to implement custom
             shuffling logic. Defaults to `False`.
@@ -80,7 +80,12 @@ class PyDataset:
             self.x, self.y = x_set, y_set
             self.batch_size = batch_size
             self.shuffle = shuffle
+            # create index array for shuffling
             self.indices = np.arange(len(self.x))
+            # Shuffle once at initialization when shuffle=True
+            if self.shuffle:
+                np.random.shuffle(self.indices)
+
 
         def __len__(self):
             # Return number of batches.
@@ -92,7 +97,7 @@ class PyDataset:
             # Cap upper bound at array length; the last batch may be smaller
             # if the total number of items is not a multiple of batch size.
             high = min(low + self.batch_size, len(self.x))
-            # Retrieve a batch of data by index 
+            # Retrieve a batch using shuffled indices
             batch_indices = self.indices[low:high]
             batch_x = self.x[batch_indices]
             batch_y = self.y[batch_indices]
@@ -100,7 +105,7 @@ class PyDataset:
             return np.array([
                 resize(imread(file_name), (200, 200))
                    for file_name in batch_x]), np.array(batch_y)
-        
+
         def on_epoch_end(self):
             # Shuffle indices at the end of each epoch if enabled
             if self.shuffle:


keras/src/trainers/data_adapters/py_dataset_adapter.py
```diff
@@ -66,10 +75,12 @@ class PyDataset:
 
     class CIFAR10PyDataset(keras.utils.PyDataset):
 
-        def __init__(self, x_set, y_set, batch_size, **kwargs):
+        def __init__(self, x_set, y_set, batch_size, shuffle=False, **kwargs):
             super().__init__(**kwargs)
             self.x, self.y = x_set, y_set
             self.batch_size = batch_size
+            self.shuffle = shuffle
+            self.indices = np.arange(len(self.x))
```
fchollet, line 84: In the example, also do `np.random.shuffle(self.indices)` here when `shuffle` is True, since the best practice when shuffling is to do it for every epoch, not just epoch >= 1

keras/src/trainers/data_adapters/py_dataset_adapter.py
```diff
@@ -40,6 +40,12 @@ class PyDataset:
             multiprocessed setting.
             Reduce this value to reduce the CPU memory consumption of
             your dataset. Defaults to 10.
+        shuffle: Whether to shuffle the sample ordering at the end of
+            each epoch. This argument is passed to `model.fit()`. When
+            `model.fit(..., shuffle=True)`, the training loop
```
fchollet: Add verb: "When calling"

keras/src/trainers/data_adapters/py_dataset_adapter.py
```diff
@@ -40,6 +40,12 @@ class PyDataset:
             multiprocessed setting.
             Reduce this value to reduce the CPU memory consumption of
             your dataset. Defaults to 10.
+        shuffle: Whether to shuffle the sample ordering at the end of
+            each epoch. This argument is passed to `model.fit()`. When
+            calling `model.fit(..., shuffle=True)`, the training loop
+            automatically calls `on_epoch_end()` at each epoch
+            boundary, allowing datasets to implement custom
+            shuffling logic. Defaults to `False`.
```
hertschuh, line 48: The `shuffle` argument is not part of `PyDataset.__init__`, so it should be removed from here. It's just an argument in your example.

keras/src/trainers/data_adapters/py_dataset_adapter.py
```diff
@@ -52,6 +58,9 @@ class PyDataset:
         over the dataset. They are not being used by the `PyDataset` class
         directly. When you are manually iterating over a `PyDataset`,
         no parallelism is applied.
+    - `shuffle=False` keeps the sample order fixed across epochs.
+        For distributed or deterministic training prefer
+        `shuffle=False` and manage the order externally.
```
hertschuh, line 63: The `shuffle` argument is not part of `PyDataset.__init__`, so it should be removed from here. It's just an argument in your example.

keras/src/trainers/data_adapters/py_dataset_adapter.py
```diff
@@ -66,10 +75,17 @@ class PyDataset:
 
     class CIFAR10PyDataset(keras.utils.PyDataset):
 
-        def __init__(self, x_set, y_set, batch_size, **kwargs):
+        def __init__(self, x_set, y_set, batch_size, shuffle=False, **kwargs):
```
hertschuh, line 78: The `shuffle` argument is not part of `PyDataset.__init__`. I understand that you can do it this way here like in your example.

But in Keras, we add shuffling via `PyDatasetAdapter`. So it makes it confusing because if you do:
- `CIFAR10PyDataset(shuffle=True)` and `model.fit(shuffle=False)` it will be shuffled
- `CIFAR10PyDataset(shuffle=False)` and `model.fit(shuffle=True)` it will be shuffled

I guess my question is, what is the reason to add `shuffle` directly in the `PyDataset`?

brainbytes42: Hi there,

I'm sorry, but I'm quite confused about sorting a PyDataset.

> But in Keras, we add shuffling via `PyDatasetAdapter`.

The PyDatasetAdapter seems to shuffle the order of the _batches_, see https://github.com/keras-team/keras/blob/df0ec013296ceb830de8ae79738e06cebeab2546/keras/src/trainers/data_adapters/py_dataset_adapter.py#L257-L261 but not the samples inside the batches - right?

So if I have samples [1, ..., 9] and batches of size 3 I'd get something like [4,5,6], [1,2,3], [7,8,9] if shuffeling via model.fit(...)?

I'd think to shuffle the _entire_ dataset, one would have to shuffle in `on_epoch_begin()` to have randomized samples from which the batches are generated? Like [8,3,6], [9,2,1], [5,4,7]. Or do I miss something that `model.fit(shuffle=True)` would do this? Is there any best practice to get all samples shuffled?

Thank you for clarification / advice on this.

hertschuh: Oh, I see. Yes, indeed, `PyDatasetAdapter` shuffles batches, not samples.

So you can shuffle them the way you did.

Issue #21655, opened 2025-09-10
ValueError: Invalid dtype: property object - `ops.prod` with `layers.Dense`


innat:
The output of `ops.prod` doesn't work in layer's argument. To make it work, it's required to cast, i.e. `int(ops.prod)`. Now, still its an issue to use `ops.prod` in `call` method anyway.

```python
import keras
from keras import layers, ops
import numpy as np

class ProdDenseLayer(layers.Layer):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        
    def build(self, input_shape):
        self.input_prod = ops.prod(input_shape[1:])
        self.dense = layers.Dense(self.input_prod, activation='relu')
        self.dense.build(input_shape)
        
    def call(self, inputs):
        scale_factor = ops.prod(ops.shape(inputs)[1:])
        scaled_inputs = inputs * ops.cast(scale_factor, inputs.dtype)
        return self.dense(scaled_inputs)

# Main model using ops.prod
class ProdModel(keras.Model):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.prod_layer1 = ProdDenseLayer()
        
    def build(self, input_shape):
        self.prod_layer1.build(input_shape)
        
    def call(self, inputs):
        batch_size = ops.shape(inputs)[0]
        total_elements = ops.prod(ops.shape(inputs)[1:])
        normalized_inputs = inputs / ops.cast(total_elements, inputs.dtype)
        x = self.prod_layer1(normalized_inputs)
        return x

# Create dummy data
batch_size = 32
input_shape = (10,)
X_train = np.random.randn(batch_size * 10, *input_shape).astype(np.float32)
y_train = np.random.randint(0, 2, (batch_size * 10, 1)).astype(np.float32)
model = ProdModel()
model.compile(
    optimizer='adam',
    loss='binary_crossentropy',
    metrics=['accuracy']
)
history = model.fit(
    X_train, y_train,
    epochs=3,
    batch_size=batch_size,
    validation_split=0.2,
    verbose=1
)
model.summary()
```

```bash
Epoch 1/3
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/tmp/ipykernel_36/1432196851.py in <cell line: 0>()
     45     metrics=['accuracy']
     46 )
---> 47 history = model.fit(
     48     X_train, y_train,
     49     epochs=3,

[...]

PR #21848, opened 2025-11-13
Add _maybe_convert_to_int utility to handle symbolic tensor dimensions safely

MalyalaKarthik66:
Fix: #21655 
This PR adds a helper utility _maybe_convert_to_int to safely handle cases where shape dimensions or layer units are symbolic tensors. It ensures that numeric values are properly converted to integers when possible, improving backend compatibility.

### DIFF
--- a/keras/src/testing/test_ops_int_cast.py
+++ b/keras/src/testing/test_ops_int_cast.py
@@ -1,18 +1,24 @@
 import numpy as np
+import pytest
 
 import keras
+from keras import backend
 from keras import layers
 from keras import ops
-from keras.src.utils.arg_casts import _maybe_convert_to_int  # import helper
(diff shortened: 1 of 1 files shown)


hertschuh (commented):
@MalyalaKarthik66 

Thanks for your effort. But actually, this utility already exists:

https://github.com/keras-team/keras/blob/6fd26b43798c901ee1cd93ad7671fb655c3dbfc8/keras/src/backend/common/variables.py#L591

I modified it recently to cast anything that is castable to int:

https://github.com/keras-team/keras/blob/6fd26b43798c901ee1cd93ad7671fb655c3dbfc8/keras/src/backend/common/variables.py#L627

So it will work with eager tensors, etc.

PR #21850, opened 2025-11-14
Update keras3 Softmax mask handling to be more numerically robust.

yangliyl-g:
The original impl of masking is essentially `(1.0 - mask) * -1e-9 + inputs`; this can be sensitive to numerical noise on `mask` (imagine if it is slightly off from either 1 or 0; then we would be adding a very large perturbation to the inputs).

Using comparison and where ops are much more numerically robust. Also for cases where we get floating point masks coming in, we add a binarization step to make it compatible w/ the `where` op.

### DIFF
--- a/keras/src/layers/activations/softmax.py
+++ b/keras/src/layers/activations/softmax.py
@@ -56,7 +56,8 @@ class Softmax(Layer):
             # other (masked) positions to -1e.9.
             if backend.standardize_dtype(mask.dtype) != "bool":
                 mask = backend.numpy.greater(
-                    mask, backend.cast(0.5, dtype=mask.dtype))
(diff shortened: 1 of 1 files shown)


hertschuh (approved):
Thanks for the improvement!

hertschuh:
Oh, you'll need to reformat the code.

`ruff --config pyproject.toml format .` should do the job.

Otherwise the real way is:

hertschuh (approved):

yangliyl-g:
This PR was based on a deep and thorough investigation led by [Jaswanth Sreeram](https://www.linkedin.com/in/jaswanth-sreeram-b9117763) who prototyped the solution. the identification and analysis of where this issue came from is largely thanks to Jaswanth & other folks who contributed to the difficult investigation.

Issue #20949, opened 2025-02-23
Conv3DTranspose issue with torch backend padding arguments (padding=same) and output_padding=1) lead to a Torch output_padding (1) that is greater than strides (1).


pure-rgb:
I ended up with the following issue https://github.com/keras-team/keras-core/issues/774

Currently a warning is given but what is the solution or workaround? 

ValueError: Exception encountered when calling Conv3DTranspose.call().

The padding arguments (padding=same) and output_padding=1) lead to a Torch output_padding (1) that is greater than strides (1). This is not supported. You can change the padding arguments, kernel or stride, or run on another backend. 

Arguments received by Conv3DTranspose.call():
  • inputs=torch.Tensor(shape=torch.Size([16, 96, 6, 6, 128]), dtype=float32)

dhantule:
Hi @pure-rgb, thanks for reporting this. Could you provide some reproducible code?

pure-rgb:
@dhantule 

Sample code.

```python
import  os
os.environ["KERAS_BACKEND"] = "torch"

import numpy as np
import keras

a = np.ones((1, 8, 8, 8, 16), dtype=np.float32)
x = keras.layers.Conv3DTranspose(
    32,
    kernel_size=2,
    strides=1,
    padding='same',
)
output = x(a)
print(output.shape)
```

This works in tf and jax, output: `(1, 8, 8, 8, 32)`. But in torch backend, it gives

```bash
/opt/conda/lib/python3.10/site-packages/keras/src/backend/common/backend_utils.py:91: UserWarning: You might experience inconsistencies across backends when calling conv transpose with kernel_size=2, stride=1, dilation_rate=1, padding=same, output_padding=1.
  warnings.warn(
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[2], line 8
      1 a = np.ones((1, 8, 8, 8, 16), dtype=np.float32)
      2 x = keras.layers.Conv3DTranspose(
      3     32,
      4     kernel_size=2,
      5     strides=1,
      6     padding='same',
      7 )
----> 8 output = x(a)
      9 print(output.shape)

File /opt/conda/lib/python3.10/site-packages/keras/src/utils/traceback_utils.py:122, in filter_traceback.<locals>.error_handler(*args, **kwargs)
    119     filtered_tb = _process_traceback_frames(e.__traceback__)
    120     # To get the full stack trace, call:
    121     # `keras.config.disable_traceback_filtering()`
[...]

PR #21852, opened 2025-11-15
Fix Torch output_padding constraint for ConvTranspose layers

MalyalaKarthik66:
Fix: #20949 
This PR fixes the Torch backend error where output_padding >= stride results in invalid shapes for ConvTranspose ops.
Changes

- Added clamping logic in _convert_conv_transpose_padding_args_from_keras_to_torch to ensure output_padding < stride (Torch requirement).

- Added clear UserWarning when clamping occurs.

- Added dedicated test: test_output_padding_clamped_for_torch_constraint.

- This ensures consistent behavior across backends and prevents Torch runtime failures.


fchollet (approved):
LGTM, thanks for the fix!

Issue #21808, opened 2025-11-01
Failed to apply Splash kernel for flash attention.


telecomadm1145:
ERROR:absl:Failed to apply Splash kernel for flash attention. Falling back to JAX native dot_product_attention.
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/keras/src/backend/jax/nn.py", line 1336, in dot_product_attention
    output = wrap_flash_attention(
             ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/keras/src/backend/jax/nn.py", line 1188, in wrap_flash_attention
    splash_kernel = splash_attention_kernel.make_splash_mha(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/jax/experimental/pallas/ops/tpu/splash_attention/splash_attention_kernel.py", line 2557, in _make_splash_attention
    fwd_mask_info, mask_function_fwd = process_mask_fn(
                                       ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/jax/experimental/pallas/ops/tpu/splash_attention/splash_attention_mask.py", line 222, in __hash__
    return hash((type(self),) + tuple(hash(mask) for mask in self.masks))
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/jax/experimental/pallas/ops/tpu/splash_attention/splash_attention_mask.py", line 222, in <genexpr>
    return hash((type(self),) + tuple(hash(mask) for mask in self.masks))
                                      ^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/jax/experimental/pallas/ops/tpu/splash_attention/splash_attention_mask.py", line 516, in __hash__
    return hash((type(self), self.array.tobytes()))
                             ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/jax/_src/core.py", line 935, in tobytes
    raise ConcretizationTypeError(self,
jax.errors.ConcretizationTypeError: Abstract tracer value encountered where concrete value is expected: traced array with shape bool[1024,1024]
The tobytes() method was called on traced array with shape bool[1024,1024].
[...]

PR #21853, opened 2025-11-15
Fix ConcretizationTypeError in dot_product_attention with traced masks

Abhinavexists:
Fix: #21808 

using `dot_product_attention` with a mask inside JAX control flow operations (e.g., `while_loop`, `scan`) on TPU, the code raised a `ConcretizationTypeError`

Added a traced mask detection check before attempting to use TPU flash attention

### DIFF
--- a/keras/src/backend/jax/nn.py
+++ b/keras/src/backend/jax/nn.py
@@ -1356,7 +1356,6 @@ def dot_product_attention(
         except jax.errors.ConcretizationTypeError:
             # Mask is traced
             # Fall back to native attention
-            flash_attention = False
             logging.exception(
                 "Failed to apply Splash kernel for flash attention. "
                 "Falling back to JAX native dot_product_attention."


keras/src/backend/jax/nn.py
```diff
@@ -1276,6 +1276,10 @@ def dot_product_attention(
         # use flash attention
         _can_use_flash_attention(query, key, value, bias, raise_error=True)
 
+    # On TPU, traced masks cause ConcretizationTypeError with flash attention.
+    if is_tpu and mask is not None and isinstance(mask, jax.core.Tracer):
+        flash_attention = False
```
hertschuh: Because on JAX we compile by default, this means that `flash_attention` is now always disabled. The whole flash attention implementation was for TPU specifically. So I don't think this is the right fix.

Am I misunderstanding something?

Abhinavexists: Hi @hertschuh 

You're absolutely right to question this - I need to clarify.

The key distinction is that `mask` being a `Tracer` doesn't happen during normal compilation. In typical usage, the mask is a concrete array that gets traced, but the mask *value itself* isn't a Tracer.

However, when `dot_product_attention` is called inside JAX control flow primitives like `scan` or `while_loop`, the mask becomes a Tracer because these primitives abstract over the loop-carried values. This is when flash attention fails with ConcretizationTypeError.

So the check `isinstance(mask, jax.core.Tracer)` should only trigger in the specific case where we're inside control flow operations, not during normal compilation.

That said, I want to verify this is actually the case. ?

Abhinavexists: @hertschuh You're absolutely right! Thank you for catching that.

My initial fix using `isinstance(mask, jax.core.Tracer)` was too broad and would have disabled flash attention even during normal JIT compilation, not just inside control flow operations.

I've revised the fix to instead:

1. **Removed the Tracer check**
2. **Improved the exception handling** to specifically catch `jax.errors.ConcretizationTypeError`

The fix now correctly handles traced masks without disabling flash attention for normal use cases.

keras/src/backend/jax/nn.py
```diff
@@ -1353,7 +1353,10 @@ def dot_product_attention(
             )
             # Transpose output back to Keras layout
             return jnp.transpose(output, axes=(0, 2, 1, 3))
-        except Exception:
+        except jax.errors.ConcretizationTypeError:
+            # Mask is traced
+            # Fall back to native attention
+            flash_attention = False
```
hertschuh: `flash_attention = False`

This is already line 1364, remove.

hertschuh (approved):

hertschuh (commented):
So the TPU tests fail.

https://github.com/keras-team/keras/actions/runs/19870652602/job/56945198925?pr=21853

It look like the original code was correct. It's not just `ConcretizationTypeError`s that can happen. Other types of errors can happen and we need to turn off splash attention.

divyashreepathihalli (approved):
Thank you for the fix!

hertschuh:
Closing in favor of https://github.com/keras-team/keras/pull/21891 which adds unit tests.

PR #21854, opened 2025-11-16
Fix Muon optimizer to use variable.name instead of deprecated variable.path

MalyalaKarthik66:
Fix: #21793 

This PR fixes an AttributeError in the Muon optimizer caused by the deprecated  `variable.path` attribute. All references are now updated to use `variable.name`,  ensuring compatibility with TensorFlow 2.16+.

- Updated `_should_use_adamw`, `_muon_update_step`, and `_adamw_update_step`  to reference `variable.name`.
- Added regression test `test_exclude_layers_with_variable_name` in muon_test.py  to ensure `exclude_layers` works correctly with current TF versions.

### DIFF
--- a/keras/src/optimizers/muon.py
+++ b/keras/src/optimizers/muon.py
@@ -161,13 +161,13 @@ class Muon(optimizer.Optimizer):
 
         for var in var_list:
             if not self._overwrite_variable_with_gradient(var):
-                self.adam_momentums[var.name] = (
+                self.adam_momentums[var.path] = (
                     self.add_variable_from_reference(
                         reference_variable=var, name="momentum"
                     )
                 )
                 if self._should_use_adamw(var):
-                    self.adam_velocities[var.name] = (
+                    self.adam_velocities[var.path] = (
                         self.add_variable_from_reference(
                             reference_variable=var, name="velocity"
                         )


keras/src/optimizers/muon.py
```diff
@@ -161,13 +161,13 @@ def build(self, var_list):
 
         for var in var_list:
             if not self._overwrite_variable_with_gradient(var):
-                self.adam_momentums[var.path] = (
+                self.adam_momentums[var.name] = (
```
fchollet: These code paths are intended for Keras variables, for which we mean to use the `path` attribute, which is different from the `name` attribute.

MalyalaKarthik66:
@fchollet 
Updated the code as suggested, restoring var.path for Keras variables. Thanks for the review!

pass-lin:
There are issues with this fix:

1. There are many duplicates in `variable.name`. For two dense layers, their weights are both named `'kernel'` and `'bias'`. If we only want to use Adam for one of the dense layers, how can we distinguish between the two dense layers?

2. Similarly, due to the naming conflict issue:
   ```python
   if variable.name not in self.adam_momentums:
       self.adam_momentums[variable.name] = (
               reference_variable=variable, name="momentum"
           )
       )
   if variable.name not in self.adam_velocities:
       self.adam_velocities[variable.name] = (
               reference_variable=variable, name="velocity"
           )
       )
   ```
   Multiple variables might end up using the same cache, which would cause errors.

You can easily see the difference between the `name` and `path` properties in the following code.
```python
import keras
model = keras.Sequential([
    keras.layers.Dense(5), 
    keras.layers.Dense(10), 
])
for w in model.weights:
    print(w.path,w.name) 
```

pass-lin:
I've provided an improved bug-fix version that doesn't affect the current design of the Muon optimizer.
https://github.com/keras-team/keras/pull/21859

keras/src/optimizers/muon.py
hertschuh: Please see the detailed comment here: https://github.com/keras-team/keras/pull/21797#pullrequestreview-3436406654

For how to fix this properly.

hertschuh:
Closing in favor of https://github.com/keras-team/keras/pull/21924

PR #21855, opened 2025-11-17
Add note about label noise in CIFAR-10 dataset documentation

### LINKED ISSUE #21631: Possible label inconsistencies in CIFAR-10 dataset
While training a CNN model using CIFAR-10 via "keras.datasets.cifar10.load_data()", I noticed that some images seem to have incorrect ground-truth labels. This leads to misleading training results and incorrect evaluation.

import numpy as np
import keras

(X_train, Y_train), (X_test, Y_test) = keras.datasets.cifar10.load_data()
X_train = X_train / 255.0
X_test = X_test / 255.0

class_labels = ['airplane','automobile','bird','cat','deer',
                'dog','frog','horse','ship','truck']

plt.figure(figsize=(12,8))
for j, i in enumerate(np.random.randint(0, 10000, 6)):
    plt.subplot(2,3,j+1)
    plt.imshow(X_test[i])
    plt.axis('off')
    plt.title(f"Actual={class_labels[Y_test[i][0]]} ({Y_test[i][0]})")
plt.show()

Some images appear to be mislabeled. For example:

     An airplane image labeled as frog

     A horse image labeled as cat

    A ship image labeled as truck

Image

ssam18:
I was working with the CIFAR-10 dataset and noticed that the documentation doesn't mention the known issue of label noise. I kept running into images that seemed mislabeled - airplanes labeled as frogs, horses as cats, etc. At first I thought something was wrong with my code, but after researching, I learned this is a well-known characteristic of the original CIFAR-10 dataset.

I've added a brief note to the `load_data()` docstring to inform users about this label noise issue. This should help new users understand that some label inconsistencies are expected and not due to bugs in their code or the Keras implementation.

The note is concise and placed right after the dataset description, so users will see it before they start working with the data. It mentions that:
- The dataset has a small percentage of mislabeled samples
- This is inherent to the original dataset
- It may impact training and evaluation

Fixes #21631

### DIFF
--- a/keras/src/datasets/cifar10.py
+++ b/keras/src/datasets/cifar10.py
@@ -59,6 +59,11 @@ def load_data():
     assert y_train.shape == (50000, 1)
     assert y_test.shape == (10000, 1)
     ```
+
+    **Note**: The CIFAR-10 dataset is known to have a small percentage of
+    mislabeled samples, which is inherent to the original dataset. This label
+    noise may impact training and evaluation. For more details, refer to
+    discussions in the research literature on CIFAR-10 label quality.
     """
     dirname = "cifar-10-batches-py-target"
     origin = "https://www.cs.toronto.edu/~kriz/cifar-10-python.tar.gz"


ssam18:
This is a simple documentation improvement that should help avoid confusion for users who encounter mislabeled images in CIFAR-10. The note is factual, concise, and doesn't require any code changes - just making the existing dataset behavior more transparent in the docs.

PR #21859, opened 2025-11-18
Modify Muon optimizer 

pass-lin:
In this PR, we have introduced three improvements to Muon:

1. In the Muon optimizer, we often designate a subset of variables to be optimized with Adam. However, since different optimizers should not be assumed to have the same weight decay parameter, we addressed this by adding an `adam_weight_decay` parameter.

2. The current implementation of Muon mainly references the [KellerJordan version](https://github.com/KellerJordan/Muon). However, the [Moonlight version](https://github.com/MoonshotAI/Moonlight/blob/master/examples/toy_train.py) is now widely recognized as superior. Compared to the KellerJordan version, the Moonlight version adjusts the learning rate from `max(d_out/d_in, 1)**0.5` to `max(d_out, d_in) * rate`. The KellerJordan version assumes that the second dimension is the output dimension and the first dimension is the input dimension. As a general-purpose optimizer, we should not make such assumptions.  
   Additionally, the Moonlight version allows Muon and Adam to maintain the same weight decay and learning rate. We have added an `rms_rate` parameter to enable this feature, with a default value of 0.2. This parameter can be disabled by setting it to `None`. We have also adjusted some default parameters based on the Moonlight version.

3. We have fixed the issue [https://github.com/keras-team/keras/issues/21793](https://github.com/keras-team/keras/issues/21793). The main cause of this bug was the loss of the `var.path` attribute during the conversion to `tf.Variable`. Therefore, we performed an assignment to ensure that `tf.Variable` can also retain the `path` attribute from `keras.variable`.

4. When we initially submitted Muon optimizer, our understanding of Muon was not deep enough. As our research progressed, we discovered that Muon was designed with the assumption that the model is a Transformer. For 3D weights, it is necessary to assume that one dimension is d_in, and the other dimensions are reshaped to d_out. However, unlike the 2D case, the 3D scenario does not always have a clear distinction between d_in and d_out. Therefore, out of caution, we only use the Adam optimizer for cases other than 2D.

### DIFF
--- a/keras/api/_tf_keras/keras/ops/__init__.py
+++ b/keras/api/_tf_keras/keras/ops/__init__.py
@@ -215,6 +215,7 @@ from keras.src.ops.numpy import isreal as isreal
 from keras.src.ops.numpy import kaiser as kaiser
 from keras.src.ops.numpy import kron as kron
 from keras.src.ops.numpy import lcm as lcm
+from keras.src.ops.numpy import ldexp as ldexp
 from keras.src.ops.numpy import left_shift as left_shift
 from keras.src.ops.numpy import less as less
 from keras.src.ops.numpy import less_equal as less_equal
--- a/keras/api/_tf_keras/keras/ops/numpy/__init__.py
+++ b/keras/api/_tf_keras/keras/ops/numpy/__init__.py
@@ -101,6 +101,7 @@ from keras.src.ops.numpy import isreal as isreal
 from keras.src.ops.numpy import kaiser as kaiser
 from keras.src.ops.numpy import kron as kron
 from keras.src.ops.numpy import lcm as lcm
+from keras.src.ops.numpy import ldexp as ldexp
 from keras.src.ops.numpy import left_shift as left_shift
 from keras.src.ops.numpy import less as less
 from keras.src.ops.numpy import less_equal as less_equal
--- a/keras/api/ops/__init__.py
+++ b/keras/api/ops/__init__.py
@@ -215,6 +215,7 @@ from keras.src.ops.numpy import isreal as isreal
 from keras.src.ops.numpy import kaiser as kaiser
 from keras.src.ops.numpy import kron as kron
 from keras.src.ops.numpy import lcm as lcm
+from keras.src.ops.numpy import ldexp as ldexp
 from keras.src.ops.numpy import left_shift as left_shift
 from keras.src.ops.numpy import less as less
 from keras.src.ops.numpy import less_equal as less_equal
--- a/keras/api/ops/numpy/__init__.py
+++ b/keras/api/ops/numpy/__init__.py
@@ -101,6 +101,7 @@ from keras.src.ops.numpy import isreal as isreal
 from keras.src.ops.numpy import kaiser as kaiser
 from keras.src.ops.numpy import kron as kron
 from keras.src.ops.numpy import lcm as lcm
+from keras.src.ops.numpy import ldexp as ldexp
 from keras.src.ops.numpy import left_shift as left_shift
 from keras.src.ops.numpy import less as less
 from keras.src.ops.numpy import less_equal as less_equal
--- a/keras/src/backend/jax/numpy.py
+++ b/keras/src/backend/jax/numpy.py
@@ -845,6 +845,19 @@ def lcm(x1, x2):
     return jnp.lcm(x1, x2)
 
 
+def ldexp(x1, x2):
+    x1 = convert_to_tensor(x1)
+    x2 = convert_to_tensor(x2)
+
+    if standardize_dtype(x2.dtype) not in dtypes.INT_TYPES:
+        raise TypeError(
+            f"ldexp exponent must be an integer type. "
+            f"Received: x2 dtype={x2.dtype}"
+        )
+
+    return jnp.ldexp(x1, x2)
+
+
 def less(x1, x2):
     x1 = convert_to_tensor(x1)
     x2 = convert_to_tensor(x2)
--- a/keras/src/backend/numpy/numpy.py
+++ b/keras/src/backend/numpy/numpy.py
@@ -773,6 +773,19 @@ def lcm(x1, x2):
     return np.lcm(x1, x2).astype(dtype)
 
 
+def ldexp(x1, x2):
+    x1 = convert_to_tensor(x1)
+    x2 = convert_to_tensor(x2)
+    dtype = dtypes.result_type(x1.dtype, x2.dtype, float)
+
+    if standardize_dtype(x2.dtype) not in dtypes.INT_TYPES:
+        raise TypeError(
+            f"ldexp exponent must be an integer type. "
+            f"Received: x2 dtype={x2.dtype}"
+        )
+    return np.ldexp(x1, x2).astype(dtype)
+
+
 def less(x1, x2):
     return np.less(x1, x2)
 
--- a/keras/src/backend/openvino/excluded_concrete_tests.txt
+++ b/keras/src/backend/openvino/excluded_concrete_tests.txt
@@ -29,6 +29,7 @@ NumpyDtypeTest::test_isin
 NumpyDtypeTest::test_isreal
 NumpyDtypeTest::test_kron
 NumpyDtypeTest::test_lcm
+NumpyDtypeTest::test_ldexp
 NumpyDtypeTest::test_logaddexp2
 NumpyDtypeTest::test_matmul_
 NumpyDtypeTest::test_maximum_python_types
@@ -108,6 +109,7 @@ NumpyTwoInputOpsCorrectnessTest::test_inner
 NumpyTwoInputOpsCorrectnessTest::test_isin
 NumpyTwoInputOpsCorrectnessTest::test_kron
 NumpyTwoInputOpsCorrectnessTest::test_lcm
+NumpyTwoInputOpsCorrectnessTest::test_ldexp
 NumpyTwoInputOpsCorrectnessTest::test_quantile
 NumpyTwoInputOpsCorrectnessTest::test_tensordot
 NumpyTwoInputOpsCorrectnessTest::test_vdot
@@ -131,11 +133,13 @@ NumpyTwoInputOpsDynamicShapeTest::test_hypot
 NumpyTwoInputOpsDynamicShapeTest::test_isin
 NumpyTwoInputOpsDynamicShapeTest::test_kron
 NumpyTwoInputOpsDynamicShapeTest::test_lcm
+NumpyTwoInputOpsDynamicShapeTest::test_ldexp
 NumpyTwoInputOpsStaticShapeTest::test_gcd
 NumpyTwoInputOpsStaticShapeTest::test_hypot
 NumpyTwoInputOpsStaticShapeTest::test_isin
 NumpyTwoInputOpsStaticShapeTest::test_kron
 NumpyTwoInputOpsStaticShapeTest::test_lcm
+NumpyTwoInputOpsStaticShapeTest::test_ldexp
 CoreOpsBehaviorTests::test_associative_scan_invalid_arguments
 CoreOpsBehaviorTests::test_scan_invalid_arguments
 CoreOpsCallsTests::test_associative_scan_basic_call
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -1149,6 +1149,10 @@ def lcm(x1, x2):
     raise NotImplementedError("`lcm` is not supported with openvino backend")
 
 
+def ldexp(x1, x2):
+    raise NotImplementedError("`ldexp` is not supported with openvino backend")
+
+
 def less(x1, x2):
     element_type = None
     if isinstance(x1, OpenVINOKerasTensor):
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -1844,6 +1844,23 @@ def lcm(x1, x2):
     return result
 
 
+def ldexp(x1, x2):
+    x1 = convert_to_tensor(x1)
+    x2 = convert_to_tensor(x2)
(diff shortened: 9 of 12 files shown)


pass-lin:
@fchollet 
Hello! I’ve been using Muon to train my model recently and encountered some issues that I hope to resolve. Since the Keras version of Muon was submitted in its initial form, I would like to help maintain it as much as possible.

keras/src/backend/tensorflow/optimizer.py
```diff
@@ -111,10 +111,16 @@ def weight_decay_fn(variable):
...
-        trainable_variables = [
-            v.value if isinstance(v, backend.Variable) else v
-            for v in trainable_variables
-        ]
+        def _prepare_var(v):
+            if isinstance(v, backend.Variable):
+                new_v = v.value
+                new_v.path = v.path
+            else:
+                new_v = v
```
fchollet: This statement does nothing -- it doesn't create a copy

keras/src/backend/tensorflow/optimizer.py
```diff
@@ -111,10 +111,16 @@ def weight_decay_fn(variable):
...
+                new_v.path = v.name
```
fchollet: This seems dangerous, path and name are different concepts and you can't just overwrite the path attribute in this way

fchollet (commented):
Thanks for the PR

keras/src/backend/tensorflow/optimizer.py
pass-lin: > This statement does nothing -- it doesn't create a copy

This is just my personal coding habit to keep things consistent, even though it doesn’t actually serve a practical purpose. It’s just a small part of my style. Do you think it needs to be changed?

keras/src/backend/tensorflow/optimizer.py
pass-lin: > This seems dangerous, path and name are different concepts and you can't just overwrite the path attribute in this way

If we don’t do this, TensorFlow will still report an error for `tf.variable`. In the TensorFlow backend environment, assuming that all input data are Keras variables (`keras.variable`) seems to be a reasonable assumption. According to the current handling method, TensorFlow’s `tf.variable` no longer uses the `path` attribute but instead uses the `name` attribute as the path (`path`). In fact, in the current version of TensorFlow, this approach has been widely adopted.

For example, running the following code:

```python
import keras
model = keras.Sequential([
    keras.layers.Dense(5),
    keras.layers.Dense(10),
])
for w in model.weights:
    print(w.path, w.value.name)
```

The output is as follows:

```
sequential_1/dense_2/kernel sequential_1/dense_2/kernel:0
sequential_1/dense_2/bias sequential_1/dense_2/bias:0
sequential_1/dense_3/kernel sequential_1/dense_3/kernel:0
sequential_1/dense_3/bias sequential_1/dense_3/bias:0
sequential_1/last/kernel sequential_1/last/kernel:0
sequential_1/last/bias sequential_1/last/bias:0
```

From the code and output above, it can be seen that the `name` attribute of TensorFlow variables has actually taken on the role of a path identifier. Therefore, assuming that all input data are Keras variables is reasonable in the current implementation of TensorFlow and is consistent with TensorFlow’s existing behavior.

hertschuh: Please see the detailed comment here: https://github.com/keras-team/keras/pull/21797#pullrequestreview-3436406654

This describes how to reliably code this by inspecting `path` within `build`. In `update_step`, what should be used is `self._get_variable_index(var)`. This utility was created specially to overcome this problem.

You should not overwrite `name` because some users have workflows that depend on it.

pass-lin:
@fchollet @hertschuh
Thank you for your feedback. I’ve revised the implementation to keep the changes minimal.

First, unlike weight decay, we can’t replace the variable path with `_var_key`, because the TensorFlow backend turns `keras.Variable` into `tf.Variable`, which makes the `id` change.

```python
import keras
model = keras.Sequential([
    keras.layers.Dense(5),
    keras.layers.Dense(10),
])
for w in model.weights:
    print(id(w) == id(w.value))
```

Output:
```
False
False
False
False
False
False
```

Therefore, during the build phase we attach two new attributes to every variable: `_muon_path_id` (used as an index) and `_muon_use_adam_flag` (used to decide whether to use Adam).  
We also modified `_prepare_var` so that these two attributes are copied to the new variable:

```python
def _prepare_var(v):
    new_v = v.value if isinstance(v, backend.Variable) else v
    new_v._muon_use_adam_flag = v._muon_use_adam_flag
    new_v._muon_path_id = v._muon_path_id
    return new_v
```

What do you think of this approach?

hertschuh:
> @fchollet @hertschuh Thank you for your feedback. I’ve revised the implementation to keep the changes minimal.
> 
> First, unlike weight decay, we can’t replace the variable path with `_var_key`, because the TensorFlow backend turns `keras.Variable` into `tf.Variable`, which makes the `id` change.
> 
> ```python
> import keras
> model = keras.Sequential([
>     keras.layers.Input(shape=(10,)),
>     keras.layers.Dense(5),
>     keras.layers.Dense(10),
>     keras.layers.Dense(1, name="last")
> ])
> for w in model.weights:
>     print(id(w) == id(w.value))
> ```
> 
> Output:
> 
> ```
> False
> False
> False
> False
> False
> False
> ```

Yes, this is exactly why the `self._get_variable_index(var)` mechanism was put in place. Please take a look at how Adam is implemented. Everything has to be based on this, which is guaranteed to work with both `keras.Variable` and `tf.Variable`.

> Therefore, during the build phase we attach two new attributes to every variable: `_muon_path_id` (used as an index) and `_muon_use_adam_flag` (used to decide whether to use Adam). We also modified `_prepare_var` so that these two attributes are copied to the new variable:
> 
> ```python
> def _prepare_var(v):
>     new_v = v.value if isinstance(v, backend.Variable) else v
>     new_v._muon_use_adam_flag = v._muon_use_adam_flag
>     new_v._muon_path_id = v._muon_path_id
>     return new_v
> ```
> 
> What do you think of this approach?

Let's not do that. Let's use `self._get_variable_index(var)`.

This part is actually correct in this PR: https://github.com/keras-team/keras/pull/21797/files#diff-0bd128978356cf21bbad03c1012b512942b0eb5c6c4dd3641b62dbe8d5d21744

Specifically, the correct approach is for `build`:

```python
    def build(self, var_list):
        if self.built:
            return
        super().build(var_list)

        # Initialize lists with None for all variables
        self.adam_momentums = [None] * len(var_list)
        self.adam_velocities = [None] * len(var_list)

        for var in var_list:
                var_idx = self._get_variable_index(var)
                self.adam_momentums[var_idx] = (
                        reference_variable=var, name="momentum"
                    )
                )
                if self._should_use_adamw(var):
                    self.adam_velocities[var_idx] = (
                            reference_variable=var, name="velocity"
                        )
                    )
```

And then for `update_step`:

```python
    def update_step(self, gradient, variable, learning_rate):
        if self._should_use_adamw(variable):
        var_idx = self._get_variable_index(variable)
        # Check if velocity exists to determine if we should use AdamW
        if self.adam_velocities[var_idx] is not None:
            # It should be noted that lr is one-tenth when using adamw.
            self._adamw_update_step(
                gradient, variable, learning_rate * self.adam_lr_ratio
            )
        else:
            self._muon_update_step(gradient, variable, learning_rate)
```

No need to introduce a different mechanism.

pass-lin:
> > 感谢你的反馈。我已经修改了实现，尽量减少改动。
> > 首先，与权重衰减不同，我们不能用 替换变量路径，因为 TensorFlow 后端会变成 ，从而产生变化。`_var_key``keras.Variable``tf.Variable``id`
> > ```python
> > import keras
> > model = keras.Sequential([
> >     keras.layers.Input(shape=(10,)),
> >     keras.layers.Dense(5),
> >     keras.layers.Dense(10),
> >     keras.layers.Dense(1, name="last")
> > ])
> > for w in model.weights:
> >     print(id(w) == id(w.value))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 输出：
> > ```
> > False
> > False
> > False
> > False
> > False
> > False
> > ```
> 
> 是的，这正是设立机制的原因。请看看亚当是如何实现的。一切都必须基于这个，这保证能同时配合和。`self._get_variable_index(var)``keras.Variable``tf.Variable`
> 
> > 因此，在构建阶段，我们会为每个变量附加两个新属性：（用作索引）和（用于决定是否使用Adam）。我们还修改了这两个属性，将这两个属性复制到新变量中：`_muon_path_id``_muon_use_adam_flag``_prepare_var`
> > ```python
> > def _prepare_var(v):
> >     new_v = v.value if isinstance(v, backend.Variable) else v
> >     new_v._muon_use_adam_flag = v._muon_use_adam_flag
> >     new_v._muon_path_id = v._muon_path_id
> >     return new_v
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 你怎么看这种做法？
> 
> 别这样了。我们用 。`self._get_variable_index(var)`
> 
> 这部分内容在这篇宣传中其实是正确的：https://github.com/keras-team/keras/pull/21797/files#diff-0bd128978356cf21bbad03c1012b512942b0eb5c6c4dd3641b62dbe8d5d21744
> 
> 具体来说，正确的方法是 ：`build`
> 
> ```python
>     def build(self, var_list):
>         if self.built:
>             return
>         super().build(var_list)
> 
>         # Initialize lists with None for all variables
>         self.adam_momentums = [None] * len(var_list)
>         self.adam_velocities = [None] * len(var_list)
> 
>         for var in var_list:
>             if not self._overwrite_variable_with_gradient(var):
>                 var_idx = self._get_variable_index(var)
>                 self.adam_momentums[var_idx] = (
>                     self.add_variable_from_reference(
>                         reference_variable=var, name="momentum"
>                     )
>                 )
>                 if self._should_use_adamw(var):
>                     self.adam_velocities[var_idx] = (
>                         self.add_variable_from_reference(
>                             reference_variable=var, name="velocity"
>                         )
>                     )
> ```
> 
> 然后 ：`update_step`
> 
> ```python
>     def update_step(self, gradient, variable, learning_rate):
>         if self._should_use_adamw(variable):
>         var_idx = self._get_variable_index(variable)
>         # Check if velocity exists to determine if we should use AdamW
>         if self.adam_velocities[var_idx] is not None:
>             # It should be noted that lr is one-tenth when using adamw.
>             self._adamw_update_step(
>                 gradient, variable, learning_rate * self.adam_lr_ratio
>             )
>         else:
>             self._muon_update_step(gradient, variable, learning_rate)
> ```
> 
> 没必要引入不同的机制。



> > @fchollet @hertschuh Thank you for your feedback. I’ve revised the implementation to keep the changes minimal.
> > First, unlike weight decay, we can’t replace the variable path with `_var_key`, because the TensorFlow backend turns `keras.Variable` into `tf.Variable`, which makes the `id` change.
> > ```python
> > import keras
> > model = keras.Sequential([
> >     keras.layers.Input(shape=(10,)),
> >     keras.layers.Dense(5),
> >     keras.layers.Dense(10),
> >     keras.layers.Dense(1, name="last")
> > ])
> > for w in model.weights:
> >     print(id(w) == id(w.value))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Output:
> > ```
> > False
> > False
> > False
> > False
> > False
> > False
> > ```
> 
> Yes, this is exactly why the `self._get_variable_index(var)` mechanism was put in place. Please take a look at how Adam is implemented. Everything has to be based on this, which is guaranteed to work with both `keras.Variable` and `tf.Variable`.
> 
> > Therefore, during the build phase we attach two new attributes to every variable: `_muon_path_id` (used as an index) and `_muon_use_adam_flag` (used to decide whether to use Adam). We also modified `_prepare_var` so that these two attributes are copied to the new variable:
> > ```python
> > def _prepare_var(v):
> >     new_v = v.value if isinstance(v, backend.Variable) else v
> >     new_v._muon_use_adam_flag = v._muon_use_adam_flag
> >     new_v._muon_path_id = v._muon_path_id
> >     return new_v
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What do you think of this approach?
> 
> Let's not do that. Let's use `self._get_variable_index(var)`.
> 
> This part is actually correct in this PR: https://github.com/keras-team/keras/pull/21797/files#diff-0bd128978356cf21bbad03c1012b512942b0eb5c6c4dd3641b62dbe8d5d21744
> 
> Specifically, the correct approach is for `build`:
> 
> ```python
>     def build(self, var_list):
>         if self.built:
>             return
>         super().build(var_list)
> 
>         # Initialize lists with None for all variables
>         self.adam_momentums = [None] * len(var_list)
>         self.adam_velocities = [None] * len(var_list)
> 
>         for var in var_list:
>             if not self._overwrite_variable_with_gradient(var):
>                 var_idx = self._get_variable_index(var)
>                 self.adam_momentums[var_idx] = (
>                     self.add_variable_from_reference(
>                         reference_variable=var, name="momentum"
>                     )
>                 )
>                 if self._should_use_adamw(var):
>                     self.adam_velocities[var_idx] = (
>                         self.add_variable_from_reference(
>                             reference_variable=var, name="velocity"
>                         )
>                     )
> ```
> 
> And then for `update_step`:
> 
> ```python
>     def update_step(self, gradient, variable, learning_rate):
>         if self._should_use_adamw(variable):
>         var_idx = self._get_variable_index(variable)
>         # Check if velocity exists to determine if we should use AdamW
>         if self.adam_velocities[var_idx] is not None:
>             # It should be noted that lr is one-tenth when using adamw.
>             self._adamw_update_step(
>                 gradient, variable, learning_rate * self.adam_lr_ratio
>             )
>         else:
>             self._muon_update_step(gradient, variable, learning_rate)
> ```
> 
> No need to introduce a different mechanism.

You're right. I'll open a new PR and only keep the improvements to the muon algorithm. I'll hold off on fixing the TF bug for now.
https://github.com/keras-team/keras/pull/21885

PR #21860, opened 2025-11-19
Fix ops.tile shape inference issue on TensorFlow backend

### LINKED ISSUE #20914: Inconsistent behaviours between backends on ops.tile
I have the following function to broadcast two arrays to compute the multiplication for all possible mutations. In this function I used both tile and repeat function and I found tile shows inconsistent behaviours between backends.

- Jax: 
```
(<KerasTensor shape=(None, 6, 2, 2), dtype=float32, sparse=False, name=keras_tensor_2>, <KerasTensor shape=(None, 6, 2, 2), dtype=float32, sparse=False, name=keras_tensor_3>)
(<KerasTensor shape=(None, 6, 2, 2), dtype=float32, sparse=False, name=keras_tensor_4>, <KerasTensor shape=(None, 6, 2, 2), dtype=float32, sparse=False, name=keras_tensor_5>)
```

- TensorFlow:
```
(<KerasTensor shape=(None, None, None, None), dtype=float32, sparse=False, name=keras_tensor_2>, <KerasTensor shape=(None, None, None, None), dtype=float32, sparse=False, name=keras_tensor_3>)
(<KerasTensor shape=(None, 6, 2, 2), dtype=float32, sparse=False, name=keras_tensor_4>, <KerasTensor shape=(None, 6, 2, 2), dtype=float32, sparse=False, name=keras_tensor_5>)
```

It seems that TensorFlow could not properly infer the shape of the resulting symbolic tensor. 

Another issue is that when using `repeats` for `tile` based on the shape of symbolic tensor, TensorFlow still works (although with shapes all None), but Jax raises an error: "'str' object has no attribute '_error_repr'". This issue can be reproduced by replacing `repeats` with the commented command.

*Reproduction Code*
```python
import os

from keras import ops, layers
from keras import Input

# %%
def broadcast(x1, x2):
    """
    Broadcast the shapes of x1 and x2 to allow the computation of cross-interation. 
    
    - repeating input1: (a, b) -> (a, b, a, b)
    - repeating input2: (c, d) -> (c, c, d, d)
    - result: (a, b) * (c, d) = (a * c, b * c, a * d, b * d)
    
    Args:
        x1: nD array in shape (..., n1, ny1, nx1) to be broadcasted
        x2: nD array in shape (..., n2, ny2, nx2) to be broadcasted

    Returns:
        Broadcasted nD arrays in shape (..., n1 * n2, ...)

    Examples:
        >>> import numpy as np
        >>> x1 = np.array([[[[0., 1., 2.]],
        ...                 [[3., 4., 5.]]]])
        >>> x2 = np.array([[[[0., 1., 2.]],
        ...                 [[3., 4., 5.]]]])
        >>> x1, x2 = broadcast((x1, np.zeros(np.shape(x1))), (x2, np.zeros(np.shape(x2))))
       
        >>> np.array(x1[0]) + 1j * np.array(x1[1])
        array([[[[0.+0.j, 1.+0.j, 2.+0.j]],
                [[3.+0.j, 4.+0.j, 5.+0.j]],
                [[0.+0.j, 1.+0.j, 2.+0.j]],
                [[3.+0.j, 4.+0.j, 5.+0.j]]]], dtype=complex64)
        >>> np.array(x2[0]) + 1j * np.array(x2[1])
        array([[[[0.+0.j, 1.+0.j, 2.+0.j]],
                [[0.+0.j, 1.+0.j, 2.+0.j]],
                [[3.+0.j, 4.+0.j, 5.+0.j]],
                [[3.+0.j, 4.+0.j, 5.+0.j]]]], dtype=complex64)
    """

    x1real, x1imag = x1
    x2real, x2imag = x2
    
    x1shape = ops.shape(x1real)[-3] # spatial mode dimension
    x2shape = ops.shape(x2real)[-3] # spatial mode dimension
    
    x1dims = len(ops.shape(x1real))
    # repeats = ops.scatter_update(ops.cast(ops.ones(x1dims), dtype="int32"), [[-3 + x1dims]], [x2shape])
    repeats = [1, 2, 1, 1]
    x1real = ops.tile(x1real, repeats)
    x1imag = ops.tile(x1imag, repeats)

    x2real = ops.repeat(x2real, x1shape, axis=-3)
    x2imag = ops.repeat(x2imag, x1shape, axis=-3)

    return ((x1real, x1imag), (x2real, x2imag))

# %%

class Test(layers.Layer):
    
    def call(self, inputs1, inputs2):

        return broadcast(inputs1, inputs2)

test = Test()

# %%

x1 = Input(shape=(3, 2, 2))
x2 = Input(shape=(2, 2, 2))

y1, y2 = test((x1, x1), (x2, x2))

print(y1)
print(y2)
```

*Environment*
jax                          0.5.0
jaxlib                       0.5.0
keras                        3.8.0
tensorboard                  2.18.0

ssam18:
## Summary
Fixes issue #20914 where `ops.tile` returns all-None shapes when called inside a Layer's call method with concrete integer repeats on the TensorFlow backend.

## Problem
When `ops.tile` was called inside a Layer with concrete integer repeats like `[1, 2, 1, 1]`, the TensorFlow backend was converting those repeats to a tensor, which prevented TensorFlow's shape inference from properly determining the output shape. This resulted in all dimensions being `None`.

For example:
```python
class TileLayer(layers.Layer):
    def call(self, x):
        return ops.tile(x, [1, 2, 1, 1])

x = Input(shape=(3, 2, 2))
result = TileLayer()(x)
# Before fix: result.shape = (None, None, None, None)
# After fix:  result.shape = (None, 6, 2, 2)
```

## Solution
1. **TensorFlow Backend**: Modified `tile()` to detect when repeats contains only concrete integer values and pass them directly to `tf.tile` as a Python list instead of converting to a tensor. This allows TensorFlow's shape inference to work correctly.

2. **Symbolic Ops**: Enhanced `Tile.compute_output_spec()` to handle symbolic repeat values more gracefully by checking if each repeat is a concrete integer before attempting multiplication.

3. **Tests**: Added regression tests to verify shape inference works correctly both in direct `ops.tile` calls and when used inside `Layer.call()`.

## Testing
- Added test case in `NumpyOneInputOpsDynamicShapeTest.test_tile` for multi-dimensional inputs
- Added new test `test_tile_shape_inference_in_layer` specifically for the Layer use case
- Verified both TensorFlow and JAX backends work correctly
- All existing tests continue to pass

Fixes #20914

### DIFF
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -2744,7 +2744,10 @@ def tile(x, repeats):
     x = convert_to_tensor(x)
 
     # Convert repeats to a list (works for both sequences and 1D tensors)
-    repeats = [v for v in repeats]
+    if isinstance(repeats, int):
+        repeats = [repeats]
+    else:
+        repeats = [v for v in repeats]
 
     # Process list elements: convert concrete scalar tensors to Python ints
     processed_repeats = []


ssam18:
> Thanks for the PR. Please fix the code format first, and remove changes that aren't related to the fix in the PR.

Done

ssam18:
I am not sure what am i missing for code format. @fchollet Can you help ?

ssam18:
There was confusion pre-commit hook runs shell/api_gen.sh NOT api_gen.py and I was using this python script for checking code format. Now it is fixed.

keras/src/ops/numpy.py
```diff
@@ -6411,17 +6411,31 @@ def compute_output_spec(self, x):
...
         if isinstance(repeats, int):
             repeats = [repeats]
+
+        # Convert repeats to list if it's a tuple or other iterable
+        # and extract concrete integer values
+        if not isinstance(repeats, list):
+            try:
+                repeats = list(repeats)
+            except TypeError:
+                repeats = [repeats]
```
hertschuh: I'm not following what this is fixing. The `int` case is already covered lines 6412-6413.

Can you just do this line 6414?
```python
else:
    repeats = list(repeats)
```

keras/src/ops/numpy.py
```diff
@@ -6411,17 +6411,31 @@ def compute_output_spec(self, x):
...
+            # Check if repeat is a concrete integer value
+            # If it's a symbolic tensor or unknown, we can't infer the size
             if x_size is None:
                 output_shape.append(None)
-            else:
+            elif isinstance(repeat, int):
                 output_shape.append(x_size * repeat)
+            else:
+                # repeat is symbolic (e.g., KerasTensor, tf.Tensor, etc.)
+                output_shape.append(None)
```
hertschuh: Let's simplify:

```python
if isinstance(repeat, int):
    output_shape.append(x_size * repeat)
else:
    output_shape.append(None)
```

keras/src/backend/tensorflow/numpy.py
```diff
@@ -2742,6 +2742,45 @@ def round(x, decimals=0):
...
+                            d if d is not None else -1
+                            for d in x.shape.as_list()
+                        ]
+                        x = tf.reshape(x, x_shape_list)
+                return tf.tile(x, repeats)
+    except (AttributeError, TypeError, ValueError):
+        # If anything goes wrong, fall back to original implementation
+        pass
+
+    # Original dynamic implementation for non-concrete repeats
```
hertschuh: This can be simplified. And also, there doesn't need to be 2 separate implementations, they can be combined.

The key is that you're allowed to mix TF scalar tensors and plain Python ints when you're manipulating shapes or tensors. That's what we do with `keras.ops.shape`, we return a tuple with ints for static dimensions and scalar tensors for dynamic dimensions.

So the flow for this should be as follows:
- turn `repeats` into a list. If repeats is a tensor, you also want a list. Note that `repeats = [v for v in repeats]` will also work for a 1D tensor.
- you can process the list and do the `if hasattr(r, "numpy"):  something here (int(r.numpy()))` thing
- if you need to add values to `repeats` always do `repeats = [1] * (x_rank - len(repeats)) + repeats`, remove the use of `tf.pad` for this
- if you need to add dimensions to `x` do `tf.expand_dims(x, 0)` in a loop. This works out much better than `reshape` to add dimensions without losing the static dimensions. So remove the use of `tf.pad` for this.
- finally do `tf.tile(x, repeats)`

Remove `repeats = tf.reshape(convert_to_tensor(repeats, dtype="int32"), [-1])`, non-1D tensors are not allowed.

keras/src/ops/numpy_test.py
```diff
@@ -9507,3 +9511,24 @@ def call(self, x):
         model.compile(jit_compile=jit_compile)
 
         model.predict(np.random.randn(1, 8))
+
+    def test_tile_shape_inference_in_layer(self):
```
hertschuh: Right now, this is part of the `HistogramTest` class, which is not relevant. Please add a separate class.

keras/src/ops/numpy_test.py
```diff
@@ -9507,3 +9511,24 @@ def call(self, x):
...
+        """Test that ops.tile properly infers output shape in a Layer.
+
+        This is a regression test for issue #20914 where TensorFlow backend
+        would return all-None shapes when tile was called inside a Layer's
+        call method with concrete integer repeats.
+        """
+
```
hertschuh: Remove this comment. While it's relevant in the context of the PR, if we did this for every PR, it would really add up and make the code less readable.

We can always recover the context of this test by using blame / history.

keras/src/ops/numpy_test.py
```diff
@@ -9507,3 +9511,24 @@ def call(self, x):
...
+            def call(self, x):
+                # Use concrete integer repeats
+                repeats = [1, 2, 1, 1]
+                return knp.tile(x, repeats)
+
+        inputs = keras.Input(shape=(3, 2, 2))
+        output = TileLayer()(inputs)
+
+        # With the fix, output shape should be (None, 6, 2, 2)
+        # Before the fix, it was (None, None, None, None)
```
hertschuh: Remove this comment. While it's relevant in the context of the PR, "after the fix" won't make sense to whoever is looking at this code in the future.

We can always recover the context of this test by using blame / history.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -2742,19 +2742,32 @@ def round(x, decimals=0):
...
+        if hasattr(r, "numpy") and r.shape == ():
+            processed_repeats.append(int(r.numpy()))
+        else:
+            processed_repeats.append(r)
+    repeats = processed_repeats
+
+    # Get x rank
+    x_rank = x.shape.rank
+    if x_rank is None:
+        x_rank = tf.rank(x)
```
hertschuh: You can remove these 2 lines. Just in general, we don't support dynamic ranks. This will fail anyway line 2764-2765 with the current implementation.

hertschuh (commented):
Thanks for the update. This is much simpler. Last tweak:

ssam18:
Done. Let's see

keras/src/ops/numpy.py
```diff
@@ -6411,17 +6411,20 @@ def compute_output_spec(self, x):
...
             repeats = [1] * (len(x_shape) - len(repeats)) + repeats
         else:
             x_shape = [1] * (len(repeats) - len(x_shape)) + x_shape
 
         output_shape = []
         for x_size, repeat in zip(x_shape, repeats):
-            if x_size is None:
-                output_shape.append(None)
-            else:
+            if isinstance(repeat, int):
```
hertschuh: Sorry, I was incorrect in the previous review, this should be `if isinstance(x_size, int):`

keras/src/backend/tensorflow/numpy.py
```diff
@@ -2742,19 +2742,30 @@ def round(x, decimals=0):
...
-    )
-    x_shape = tf.pad(
-        tf.shape(x),
-        [[tf.maximum(repeats_size - x.shape.rank, 0), 0]],
-        constant_values=1,
-    )
-    x = tf.reshape(x, x_shape)
+
+    # Convert repeats to a list (works for both sequences and 1D tensors)
+    repeats = [v for v in repeats]
```
hertschuh:
```python
if isinstance(repeats, int):
    repeats = [repeats]
else:
    repeats = [v for v in repeats]

ssam18: done

hertschuh (approved):

PR #21863, opened 2025-11-20
Implement ldexp function in keras.ops

shashaka:
Adds keras.ops.ldexp, which multiplies a floating-point tensor by 2 raised to the power of an integer tensor, element-wise.
Not supported on OpenVINO.

### DIFF
--- a/keras/src/backend/numpy/numpy.py
+++ b/keras/src/backend/numpy/numpy.py
@@ -783,7 +783,6 @@ def ldexp(x1, x2):
             f"ldexp exponent must be an integer type. "
             f"Received: x2 dtype={x2.dtype}"
         )
-
     return np.ldexp(x1, x2).astype(dtype)
 
 


keras/src/backend/tensorflow/numpy.py
```diff
@@ -1844,6 +1844,22 @@ def lcm(x1, x2):
...
+
+    if standardize_dtype(x2.dtype) not in dtypes.INT_TYPES:
+        raise TypeError(
+            f"ldexp exponent must be an integer type. "
+            f"Received: x2 dtype={x2.dtype}"
+        )
+
+    x1 = tf.cast(x1, tf.float32)
+    x2 = tf.cast(x2, tf.float32)
+    return tf.cast(x1 * tf.pow(2.0, x2), dtype)
```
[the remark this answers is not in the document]
shashaka: I couldn't find tf.math.ldexp

keras/src/backend/torch/numpy.py
```diff
@@ -975,6 +975,20 @@ def lcm(x1, x2):
...
+    x2 = convert_to_tensor(x2)
+    dtype = dtypes.result_type(x1.dtype, x2.dtype, float)
+
+    if standardize_dtype(x2.dtype) not in dtypes.INT_TYPES:
+        raise TypeError(
+            f"ldexp exponent must be an integer type. "
+            f"Received: x2 dtype={x2.dtype}"
+        )
+
+    return cast(torch.ldexp(x1, x2), dtype)
```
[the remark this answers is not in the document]
shashaka: All dtypes tested.

keras/src/backend/numpy/numpy.py
```diff
@@ -773,6 +773,20 @@ def lcm(x1, x2):
...
+def ldexp(x1, x2):
+    x1 = convert_to_tensor(x1)
+    x2 = convert_to_tensor(x2)
+    dtype = dtypes.result_type(x1.dtype, x2.dtype, float)
+
+    if standardize_dtype(x2.dtype) not in dtypes.INT_TYPES:
+        raise TypeError(
+            f"ldexp exponent must be an integer type. "
+            f"Received: x2 dtype={x2.dtype}"
+        )
```
hertschuh, line 785: Is this needed on NumPy? Isn't the type promotion already consistent with `jax.numpy`?

keras/src/backend/tensorflow/numpy.py
```diff
@@ -1844,6 +1844,25 @@ def lcm(x1, x2):
...
+
+    if standardize_dtype(x2.dtype) not in dtypes.INT_TYPES:
+        raise TypeError(
+            f"ldexp exponent must be an integer type. "
+            f"Received: x2 dtype={x2.dtype}"
+        )
+
+    x1 = tf.cast(x1, dtypes.result_type(x1.dtype, float))
+
+    x1 = tf.cast(x1, tf.float32 if not x1.dtype.is_floating else x1.dtype)
```
hertschuh: You just cast `x1` to a float already line 1858, this is not needed.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -1844,6 +1844,25 @@ def lcm(x1, x2):
...
+    x2 = convert_to_tensor(x2)
+    dtype = dtypes.result_type(x1.dtype, x2.dtype, float)
+
+    if standardize_dtype(x2.dtype) not in dtypes.INT_TYPES:
+        raise TypeError(
+            f"ldexp exponent must be an integer type. "
+            f"Received: x2 dtype={x2.dtype}"
+        )
+
+    x1 = tf.cast(x1, dtypes.result_type(x1.dtype, float))
```
hertschuh: Why not use the `dtype` you computed line 1850?

keras/src/backend/tensorflow/numpy.py
```diff
@@ -1844,6 +1844,25 @@ def lcm(x1, x2):
...
+
+    x1 = tf.cast(x1, tf.float32 if not x1.dtype.is_floating else x1.dtype)
+    x2 = tf.cast(x2, x1.dtype)
+    result = x1 * tf.pow(tf.constant(2.0, dtype=x1.dtype), x2)
+    return tf.cast(tf.where(tf.math.is_inf(x1) | (x1 == 0), x1, result), dtype)
```
hertschuh, line 1861: Why is the `tf.where` needed here? What happens otherwise?

keras/src/backend/tensorflow/numpy.py
shashaka: I removed this.

keras/src/backend/tensorflow/numpy.py
shashaka: Thanks for review. I changed the code to re-use that.

keras/src/backend/numpy/numpy.py
shashaka: I may not have fully understood the review at first, 
but if the suggestion was to simply return np.ldexp(x1, x2), 
I tried that approach and it caused dtype mismatches with JAX in DtypeTest. So I think the explicit dtype handling is still required.

keras/src/backend/tensorflow/numpy.py
shashaka: Actually, I referred to the JAX implementation for this and tested the behavior. The where ensures correct results in edge cases. Please see the examples below.

```
import tensorflow as tf

x1 = tf.constant([0], dtype=tf.float16)
x2 = tf.constant([999], dtype=tf.int32)

result_normal = x1 * (tf.constant(2.0, dtype=tf.float16) ** tf.cast(x2, tf.float16))
result_where = tf.where(tf.math.is_inf(x1) | (x1 == 0), x1, result_normal)

tf.print("Without _where:", result_normal)
tf.print("With _where:", result_where)

Without _where: [-nan]
With _where: [0]
```

```
import tensorflow as tf

x1 = tf.constant([np.inf], dtype=tf.float16)
x2 = tf.constant([-99999], dtype=tf.int32)

result_normal = x1 * (tf.constant(2.0, dtype=tf.float16) ** tf.cast(x2, tf.float16))
result_where = tf.where(tf.math.is_inf(x1) | (x1 == 0), x1, result_normal)

tf.print("Without _where:", result_normal)
tf.print("With _where:", result_where)

Without _where: [-nan]
With _where: [inf]
```

keras/src/backend/tensorflow/numpy.py
```diff
@@ -1844,6 +1844,23 @@ def lcm(x1, x2):
...
+    if standardize_dtype(x2.dtype) not in dtypes.INT_TYPES:
+        raise TypeError(
+            f"ldexp exponent must be an integer type. "
+            f"Received: x2 dtype={x2.dtype}"
+        )
+
+    x1 = tf.cast(x1, dtype)
+    x2 = tf.cast(x2, x1.dtype)
+    result = x1 * tf.pow(tf.constant(2.0, dtype=x1.dtype), x2)
+    return tf.cast(tf.where(tf.math.is_inf(x1) | (x1 == 0), x1, result), dtype)
```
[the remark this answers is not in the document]
shashaka: I couldn't find native tensorflow method for ldexp.

hertschuh (approved):

PR #21864, opened 2025-11-21
Fix assigning a value to a variable within an autocast scope.

hertschuh:
Previously `assign` would incorrectly cast the value to assign to the autocast dtype instead of the true dtype of the variable.

Because on JAX and OpenVino variables are just a reference to an array, this would cause the variable value to change dtypes.


hertschuh:
@yangliyl-g

divyashreepathihalli (approved):

Issue #21861, opened 2025-11-19
[Torch backend] Layer loses keyword arguments when wrapped by RematScope


pass-lin:
**Description**  
When the Torch backend is used together with `keras.RematScope`, any layer whose `call` method is called with **keyword arguments** loses those arguments and receives the default values instead.

Other backends (JAX/TF) are not affected

---

**Reproducible example**  
```python
import os
os.environ["KERAS_BACKEND"] = "torch"

import keras
dim = 4
class RwkvBlock(keras.layers.Layer):
    def call(self, x, v_first=None):
        print(f"[{self.name}]  v_first={v_first is None}")
        v_first  = x
        return x, v_first 
    def compute_output_shape(self,**kwargs):
        return [[None,dim],[None,dim]]

def build_model():
    inputs = keras.Input(shape=(dim,))
    x = inputs
    v_first = None

    for i in range(5):
        x, v_first = RwkvBlock(name=f"block{i}")(x, v_first=v_first)

    model = keras.Model(inputs, x)
    model.build((None, dim))
    return model



model = build_model()
_ = model(keras.ops.ones((1, dim)))


with keras.RematScope(mode="list_of_layers",
    layer_names=[f"block{i}" for i in range(5)]):
    model = build_model()
_ = model(keras.ops.ones((1, dim)))
```

**Output**  
```
----- normal call -----
[block0]  v_first=True
[block1]  v_first=False
[block2]  v_first=False
[block3]  v_first=False
[block4]  v_first=False
----- inside RematScope -----
[block0]  v_first=True
[block1]  v_first=True          # <- should be False
[block2]  v_first=True
[block3]  v_first=True
[block4]  v_first=True
```

Calling the same layer with **positional arguments** (`layer(x, v_first)`) works correctly even under remat.

---

**Expected behaviour**  
The layer should receive the same arguments regardless of whether rematerialisation is enabled or which backend is used.

---

**Environment**  
- Keras 3.x (current master)  
- Torch 2.x  
- Python 3.10+  

---

dhantule:
Hi @pass-lin, thanks for reporting this. 

[...]

PR #21865, opened 2025-11-21
Fix: torch layer losses keyword arguments in rematscope

Abhinavexists:
Fix: #21861

wrapped the layer call in a small function that accepts only positional arguments. That wrapper remembers the original keyword arguments.
passed this wrapper to `torch.utils.checkpoint`, so replay uses positional args but the wrapper restores the missing kwargs before calling the layer.

### DIFF
--- a/keras/src/backend/torch/core.py
+++ b/keras/src/backend/torch/core.py
@@ -675,7 +675,7 @@ def remat(f):
     def wrapped(*args, **kwargs):
         return torch.utils.checkpoint.checkpoint(
             f, *args, use_reentrant=False, **kwargs
-            )
+        )
 
     return wrapped
 


keras/src/backend/torch/core.py
```diff
@@ -673,7 +674,13 @@ def remat(f):
...
 
     def wrapped(*args, **kwargs):
-        return torch.utils.checkpoint.checkpoint(f, *args, use_reentrant=False)
+        if not kwargs:
+            return checkpoint(f, *args, use_reentrant=False)
+
+        def positional_wrapper(*pos_args):
+            return f(*pos_args, **kwargs)
+
+        return checkpoint(positional_wrapper, *args, use_reentrant=False)
```
hertschuh: Looking at the documentation, it looks like you can just do:

```python
return torch.utils.checkpoint.checkpoint(f, *args, use_reentrant=False, **kwargs)
```

Is that not the case?

My concern with your approach is that I think it statically binds the kwargs so they cannot be tensors.

Abhinavexists: @hertschuh 
PyTorch's checkpoint() doesn't support passing kwargs to the checkpointed function as i tried to use `return torch.utils.checkpoint.checkpoint(f, *args, use_reentrant=False, **kwargs)` because it was my initial approach but it looses there kwargs and only accept positional args.

Regarding the static binding concern: you're right that this could be problematic. Let me add a test case with tensor kwargs to verify gradient tracking works correctly.

if that works here ?

hertschuh: Hmm... it works for me: https://github.com/keras-team/keras/pull/21883

Abhinavexists: okay.. i was having issue with my type checker which i interpreted for the `return torch.utils.checkpoint.checkpoint(f, *args, use_reentrant=False, **kwargs)` not working , which is now working fine

hertschuh (approved):

PR #21874, opened 2025-11-27
`keras.utils.set_random_seed` clear the global `SeedGenerator`.

hertschuh:
This is needed to get reproducible results with `keras.random` ops.

 Also introduced a constant for `"global_seed_generator"`.

### DIFF
--- a/keras/src/utils/rng_utils.py
+++ b/keras/src/utils/rng_utils.py
@@ -5,6 +5,7 @@ import numpy as np
 from keras.src import backend
 from keras.src.api_export import keras_export
 from keras.src.backend.common import global_state
+from keras.src.random import seed_generator
 from keras.src.utils.module_utils import tensorflow as tf
 
 GLOBAL_RANDOM_SEED = "global_random_seed"
@@ -20,7 +21,7 @@ def set_random_seed(seed):
     sources of randomness, or when certain non-deterministic cuDNN ops are
     involved.
 
-    Calling this utility is equivalent to the following:
+    Calling this utility does the following:
 
     ```python
     import random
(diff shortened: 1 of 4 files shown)


keras/src/utils/rng_utils.py
```diff
@@ -52,6 +52,8 @@ def set_random_seed(seed):
 
     # Store seed in global state so we can query it if set.
     global_state.set_global_attribute(GLOBAL_RANDOM_SEED, seed)
+    # Remove global SeedGenerator, it will be recreated from the seed.
+    global_state.set_global_attribute("global_seed_generator", None)
```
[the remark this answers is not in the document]
hertschuh: Fixed

keras/src/utils/rng_utils.py
```diff
@@ -52,6 +53,10 @@ def set_random_seed(seed):
 
     # Store seed in global state so we can query it if set.
     global_state.set_global_attribute(GLOBAL_RANDOM_SEED, seed)
+    # Remove global SeedGenerator, it will be recreated from the seed.
```
divyashreepathihalli, line 59: maybe the docstring can be updated to mention that  it now resets the Keras global random generator?

keras/src/utils/rng_utils_test.py
```diff
@@ -23,11 +17,17 @@ def get_model_output():
...
-            return model.predict(ds)
+            return model.predict(x, batch_size=16)
 
         rng_utils.set_random_seed(42)
         y1 = get_model_output()
         rng_utils.set_random_seed(42)
         y2 = get_model_output()
         self.assertAllClose(y1, y2)
+
+    def test_set_random_seed_with_global_seed_generator(self):
```
divyashreepathihalli, line 39: maybe also add a test to check that different seed produces different results.  To make sure the reset mechanism doesn't accidentally lock the generator to a single state.

keras/src/utils/rng_utils.py
hertschuh: Done

keras/src/utils/rng_utils_test.py
hertschuh: Good idea, added

PR #21882, opened 2025-12-01
Implement vander function in keras.ops

shashaka:
Adds keras.ops.vander, which constructs a Vandermonde matrix from a 1-D input tensor.
Not supported on OpenVINO.

### DIFF
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -3098,7 +3098,7 @@ def vander(x, N=None, increasing=False):
     result_dtype = dtypes.result_type(x.dtype)
 
     if N is None:
-        N = tf.shape(x)[0]
+        N = shape_op(x)[0]
 
     if increasing:
         powers = tf.range(N)
--- a/keras/src/ops/numpy.py
+++ b/keras/src/ops/numpy.py
@@ -7334,8 +7334,7 @@ class Vander(Operation):
         else:
             N = self.N
 
-        out_shape = list(x.shape)
-        out_shape.append(N)
+        out_shape = x.shape + (N,)
         return KerasTensor(tuple(out_shape), dtype=x.dtype)
 
 
@@ -7369,15 +7368,15 @@ def vander(x, N=None, increasing=False):
         )
 
     if N is not None:
-        if N < 0:
-            raise ValueError(
-                f"Argument 'N' must be nonnegative. Received: N={N}"
-            )
-
         if not isinstance(N, int):
             raise TypeError(
                 f"Argument `N` must be of type `int`. "
                 f"Received: N={N} of type {type(N)}"
+            )
+
+        if N < 0:
+            raise ValueError(
+                f"Argument 'N' must be nonnegative. Received: N={N}"
             )
 
     if not isinstance(increasing, bool):
--- a/keras/src/ops/numpy_test.py
+++ b/keras/src/ops/numpy_test.py
@@ -3766,20 +3766,23 @@ class NumpyOneInputOpsCorrectnessTest(testing.TestCase):
     def test_vander(self):
         x = np.random.random((3,))
         N = 6
-        increasing = True
 
         self.assertAllClose(knp.vander(x), np.vander(x))
         self.assertAllClose(knp.vander(x, N=N), np.vander(x, N=N))
         self.assertAllClose(
-            knp.vander(x, N=N, increasing=increasing),
-            np.vander(x, N=N, increasing=increasing),
+            knp.vander(x, N=N, increasing=True),
+            np.vander(x, N=N, increasing=True),
         )
 
         self.assertAllClose(knp.Vander().call(x), np.vander(x))
         self.assertAllClose(knp.Vander(N=N).call(x), np.vander(x, N=N))
         self.assertAllClose(
-            knp.Vander(N=N, increasing=increasing).call(x),
-            np.vander(x, N=N, increasing=increasing),
+            knp.Vander(N=N, increasing=True).call(x),
+            np.vander(x, N=N, increasing=True),
+        )
+        self.assertAllClose(
+            knp.Vander(N=N, increasing=False).call(x),
+            np.vander(x, N=N, increasing=False),
         )
 
     def test_var(self):


keras/src/backend/numpy/numpy.py
```diff
@@ -1359,6 +1359,36 @@ def trapezoid(y, x=None, dx=1.0, axis=-1):
...
+
+    if not isinstance(increasing, bool):
+        raise TypeError(
+            "Argument 'increasing' must be bool. "
+            f"Received: dtype={type(increasing)}"
+        )
+
+    result_dtype = dtypes.result_type(x.dtype)
+    x = x.astype(config.floatx())
+    return np.vander(x, N=N, increasing=increasing).astype(result_dtype)
```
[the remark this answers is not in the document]
shashaka: applied

keras/src/backend/tensorflow/numpy.py
```diff
@@ -3080,6 +3080,50 @@ def _move_axis_to_last(tensor, axis):
...
+        powers = tf.range(N)
+    else:
+        powers = tf.range(N - 1, -1, -1)
+
+    x_exp = tf.expand_dims(x, axis=-1)
+
+    x_exp = tf.cast(x_exp, tf.float32)
+    powers = tf.cast(powers, tf.float32)
+    vander = tf.math.pow(x_exp, powers)
+    return tf.cast(vander, result_dtype)
```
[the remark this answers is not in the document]
shashaka: applied

keras/src/backend/jax/numpy.py
```diff
@@ -1385,6 +1385,34 @@ def trapezoid(y, x=None, dx=1.0, axis=-1):
...
+        if not isinstance(N, int):
+            raise TypeError(
+                f"Argument 'N' must be integer. Received: dtype={type(N)}"
+            )
+
+    if not isinstance(increasing, bool):
+        raise TypeError(
+            "Argument 'increasing' must be bool. "
+            f"Received: dtype={type(increasing)}"
+        )
```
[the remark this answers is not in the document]
shashaka: applied

keras/src/backend/torch/numpy.py
```diff
@@ -1819,6 +1819,35 @@ def trapezoid(y, x=None, dx=1.0, axis=-1):
...
+        if not isinstance(N, int):
+            raise TypeError(
+                f"Argument 'N' must be integer. Received: dtype={type(N)}"
+            )
+
+    if not isinstance(increasing, bool):
+        raise TypeError(
+            "Argument 'increasing' must be bool. "
+            f"Received: dtype={type(increasing)}"
+        )
```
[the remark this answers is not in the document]
shashaka: applied

keras/src/ops/numpy.py
```diff
@@ -7279,6 +7279,53 @@ def mean(x, axis=None, keepdims=False):
...
+        else:
+            N = self.N
+
+        out_shape = list(x.shape)
+        out_shape.append(N)
+        return KerasTensor(tuple(out_shape), dtype=x.dtype)
+
+
+@keras_export(["keras.ops.vander", "keras.ops.numpy.vander"])
+def vander(x, N=None, increasing=False):
```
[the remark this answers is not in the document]
shashaka: applied

keras/src/ops/numpy.py
```diff
@@ -301,33 +301,6 @@ def all(x, axis=None, keepdims=False):
     return backend.numpy.all(x, axis=axis, keepdims=keepdims)
 
 
-class Any(Operation):
```
shashaka, line 304: Because the Any class was misplaced, I moved it.

fchollet: Sounds good.

keras/src/ops/numpy.py
```diff
@@ -7319,6 +7319,77 @@ def mean(x, axis=None, keepdims=False):
...
+        return KerasTensor(tuple(out_shape), dtype=x.dtype)
+
+
+@keras_export(["keras.ops.vander", "keras.ops.numpy.vander"])
+def vander(x, N=None, increasing=False):
+    """Generate a Vandermonde matrix.
+
+    Args:
+        x: 1D input tensor.
+        N: Number of columns. If None, `N` = `len(x)`.
```
fchollet: "If `None`, it will default to `len(x)`." (don't forget backticks around None)

keras/src/ops/numpy.py
```diff
@@ -7319,6 +7319,77 @@ def mean(x, axis=None, keepdims=False):
...
+        increasing: Order of powers. If True, powers increase left to right.
+
+    Returns:
+        Output tensor, vandermonde matrix of shape `(len(x), N)`.
```
fchollet: Capitalize Vandermonde

keras/src/ops/numpy.py
```diff
@@ -7319,6 +7319,77 @@ def mean(x, axis=None, keepdims=False):
...
+
+        if not isinstance(N, int):
+            raise TypeError(
+                f"Argument `N` must be of type `int`. Received: dtype={type(N)}"
+            )
+
+    if not isinstance(increasing, bool):
+        raise TypeError(
+            f"Argument `increasing` must be of type `bool`. "
+            f"Received: dtype={type(increasing)}"
```
fchollet: "Received: increasing={increasing} of type {type(increasing)}"

keras/src/ops/numpy.py
fchollet: We should avoid capital N as a variable name.

keras/src/ops/numpy.py
```diff
@@ -7319,6 +7319,77 @@ def mean(x, axis=None, keepdims=False):
...
+
+    if N is not None:
+        if N < 0:
+            raise ValueError(
+                f"Argument 'N' must be nonnegative. Received: N={N}"
+            )
+
+        if not isinstance(N, int):
+            raise TypeError(
+                f"Argument `N` must be of type `int`. Received: dtype={type(N)}"
```
fchollet: "Received: N={N} of type {type(N)}"

keras/src/ops/numpy.py
fchollet: On second thought, we went with the capitalized-name NumPy API for `ops.eye`, so we can do the same here.

shashaka: Updated.

keras/src/ops/numpy.py
shashaka: Updated.

keras/src/ops/numpy.py
shashaka: Updated.

keras/src/ops/numpy.py
shashaka: Updated.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -3097,6 +3097,27 @@ def _move_axis_to_last(tensor, axis):
     return result
 
 
+def vander(x, N=None, increasing=False):
+    x = convert_to_tensor(x)
+    result_dtype = dtypes.result_type(x.dtype)
+
+    if N is None:
+        N = tf.shape(x)[0]
```
hertschuh: Please use `shape_op`.

```python
B = shape_op(x)[0]
```

This way, if the shape is static, N will be an int and not a tensor and in turn the result will have a static second dimension.

keras/src/ops/numpy.py
```diff
@@ -7319,6 +7319,78 @@ def mean(x, axis=None, keepdims=False):
...
+        raise ValueError(
+            "Input tensor must be 1-dimensional. "
+            f"Received: input.shape={x.shape}"
+        )
+
+    if N is not None:
+        if N < 0:
+            raise ValueError(
+                f"Argument 'N' must be nonnegative. Received: N={N}"
+            )
```
hertschuh: nitpick: move after the `if not instance(N, int)` block. If `N` is not an int, it might not be comparable, in which case this will raise an error that may not be clear. It would be nicer to say that the issue is that `N` is not an int.

keras/src/ops/numpy_test.py
```diff
@@ -3755,6 +3763,25 @@ def test_trapezoid(self):
...
+            knp.vander(x, N=N, increasing=increasing),
+            np.vander(x, N=N, increasing=increasing),
+        )
+
+        self.assertAllClose(knp.Vander().call(x), np.vander(x))
+        self.assertAllClose(knp.Vander(N=N).call(x), np.vander(x, N=N))
+        self.assertAllClose(
+            knp.Vander(N=N, increasing=increasing).call(x),
+            np.vander(x, N=N, increasing=increasing),
+        )
```
hertschuh, line 3786: Add one for `increasing=False`

keras/src/ops/numpy_test.py
```diff
@@ -3755,6 +3763,25 @@ def test_trapezoid(self):
             np.trapezoid(y, x=x, axis=1),
         )
 
+    def test_vander(self):
+        x = np.random.random((3,))
+        N = 6
+        increasing = True
```
hertschuh: It's actually simpler to follow the test if you write `knp.vander(x, N=N, increasing=True)`

keras/src/ops/numpy.py
```diff
@@ -7319,6 +7319,78 @@ def mean(x, axis=None, keepdims=False):
...
+        return backend.numpy.vander(x, self.N, self.increasing)
+
+    def compute_output_spec(self, x):
+        if self.N is None:
+            N = x.shape[0]
+        else:
+            N = self.N
+
+        out_shape = list(x.shape)
+        out_shape.append(N)
```
hertschuh: FYI, you can assume x.shape is a tuple and simply do:
`output_shape = x.shape + (N,)`

keras/src/backend/tensorflow/numpy.py
shashaka: Applied.

keras/src/ops/numpy_test.py
shashaka: Applied.

keras/src/ops/numpy_test.py
shashaka: Applied.

keras/src/ops/numpy.py
shashaka: Thank you for the helpful tip.

keras/src/ops/numpy.py
shashaka: Thanks for the suggestion.
I reordered the checks so that the isinstance(N, int) validation runs before the non-negativity check.

hertschuh (approved):

shashaka:
@hertschuh
Since I don’t have merge access, could you merge this PR when you have a moment?

PR #21890, opened 2025-12-03
Remove `log.log`.

james77777778:
I believe this file was wrongly committed in #21842


hertschuh:
Thanks!

This was already addressed in https://github.com/keras-team/keras/pull/21901

PR #21891, opened 2025-12-03
Fix TPU tests - for splash attention

divyashreepathihalli:


keras/src/backend/jax/nn.py
```diff
@@ -1353,7 +1353,7 @@ def dot_product_attention(
             )
             # Transpose output back to Keras layout
             return jnp.transpose(output, axes=(0, 2, 1, 3))
-        except Exception:
+        except (jax.errors.ConcretizationTypeError, Exception):
```
hertschuh: `jax.errors.ConcretizationTypeError` is a subclass of `Exception`, so you don't need this change.

keras/src/ops/nn_test.py
```diff
@@ -1324,6 +1324,37 @@ def test_polar(self):
...
+        try:
+            if jax.devices()[0].platform != "tpu":
+                self.skipTest("TPU-specific test")
+        except:
+            self.skipTest("TPU-specific test")
+
+        import jax.numpy as jnp
+
+        from keras.src.backend.jax import nn as jax_nn
+
```
hertschuh: Why import this? Because you know the backend is jax, `knn` is the right one. Why not just do `knn.dot_product_attention` below.

keras/src/ops/nn_test.py
```diff
@@ -1324,6 +1324,37 @@ def test_polar(self):
 
 
 class NNOpsCorrectnessTest(testing.TestCase):
+    def test_dot_product_attention_inside_scan(self):
+        if backend.backend() != "jax":
+            self.skipTest("JAX-specific test")
```
hertschuh: We normally use:

```python    
   @pytest.mark.skipif(
        backend.backend() != "jax", reason="JAX only"
    )
```

hertschuh (approved):

jeffcarp (approved):

PR #21896, opened 2025-12-04
Introduces QuantizationConfig for fine-grained quantization control

JyotinderSingh:
## Description

This PR introduces `keras.quantizers.QuantizationConfig`, a configuration class that allows users to customize the quantization behavior for weights and activations during Post-Training Quantization (PTQ).

Previously, calling `model.quantize(mode="int8")` applied a fixed quantization strategy (typically `AbsMaxQuantizer` for both weights and activations). With this change, users can instantiate a configuration object (e.g. `Int8QuantizationConfig`) to define specific quantizers for weights and activations, or disable activation quantization entirely for weight-only quantization.

#### `QuantizationConfig` Class

* Introduced `keras.src.quantizers.quantization_config.py`.
* Added subclasses for specific modes: `Int8QuantizationConfig`, `Int4QuantizationConfig`, and `Float8QuantizationConfig`.
* Holds `weight_quantizer` and `activation_quantizer` instances with serialization support (`get_config`/`from_config`).

#### Model & Layer Updates

* Model: Updated `Model.quantize()` to accept an optional `QuantizationConfig` object, allowing custom configurations beyond the default string modes.
* Layers: Refactored `Dense`, `EinsumDense`, `Embedding`, and `ReversibleEmbedding`.
    * `quantized_build` and `quantize` methods now accept and parse the provided `config`.
    * Added logic to handle `activation_quantizer=None`, enabling **weight-only quantization**.
    * Updated `grad_fn` in `Dense` and `EinsumDense` to correctly handle cases where inputs are not quantized (i.e., when `inputs_scale` is effectively 1.0).

#### Quantizer Updates

* Updated `AbsMaxQuantizer.__call__` to accept `to_numpy=True` to support direct weight manipulation during the quantization step.

### Usage Examples

1. Default Quantization (Int8)
Applies the default `AbsMaxQuantizer` to both weights and activations.

```python
model.quantize("int8")
```

2. Weight-Only Quantization (Int8)
Disable activation quantization by setting the activation quantizer to `None`.

```python
from keras.quantizers import Int8QuantizationConfig, AbsMaxQuantizer

config = Int8QuantizationConfig(
    weight_quantizer=AbsMaxQuantizer(axis=0),
    activation_quantizer=None 
)

model.quantize(config=config)
```

3. Custom Quantization Parameters
Customize the value range or other parameters for specific quantizers.

```python
config = Int8QuantizationConfig(
    # Restrict range for symmetric quantization
    weight_quantizer=AbsMaxQuantizer(axis=0, value_range=(-127, 127)),
    activation_quantizer=AbsMaxQuantizer(axis=-1, value_range=(-127, 127))
)

model.quantize(config=config)
```

### Tests

  * Added `QuantizationConfigTest` for serialization/deserialization.
  * Updated layer tests (`DenseTest`, `EinsumDenseTest`, `EmbeddingTest`, `ReversibleEmbeddingTest`) to verify behavior with `QuantizationConfig`, specifically testing weight-only scenarios.
  * Updated `GPTQTest` to ensure config validation logic remains correct.

### DIFF
--- a/keras/src/layers/core/dense.py
+++ b/keras/src/layers/core/dense.py
@@ -772,7 +772,7 @@ class Dense(Layer):
             x = self.activation(x)
         return x
 
-    def quantize(self, mode, type_check=True, config=None):
+    def quantize(self, mode=None, type_check=True, config=None):
         # Prevent quantization of the subclasses
         if type_check and (type(self) is not Dense):
             raise self._not_implemented_error(self.quantize)
--- a/keras/src/layers/core/einsum_dense.py
+++ b/keras/src/layers/core/einsum_dense.py
@@ -975,7 +975,7 @@ class EinsumDense(Layer):
             x = self.activation(x)
         return x
 
-    def quantize(self, mode, type_check=True, config=None):
+    def quantize(self, mode=None, type_check=True, config=None):
         # Prevent quantization of the subclasses
         if type_check and (type(self) is not EinsumDense):
             raise self._not_implemented_error(self.quantize)
--- a/keras/src/layers/core/embedding.py
+++ b/keras/src/layers/core/embedding.py
@@ -409,7 +409,7 @@ class Embedding(Layer):
             )
         return outputs
 
-    def quantize(self, mode, type_check=True, config=None):
+    def quantize(self, mode=None, type_check=True, config=None):
         # Prevent quantization of the subclasses.
         if type_check and (type(self) is not Embedding):
             raise self._not_implemented_error(self.quantize)
--- a/keras/src/layers/core/reversible_embedding.py
+++ b/keras/src/layers/core/reversible_embedding.py
@@ -289,7 +289,7 @@ class ReversibleEmbedding(layers.Embedding):
                 )
             return logits
 
-    def quantize(self, mode, type_check=True, config=None):
+    def quantize(self, mode=None, type_check=True, config=None):
         if type_check and type(self) is not ReversibleEmbedding:
             raise self._not_implemented_error(self.quantize)
 
--- a/keras/src/layers/layer.py
+++ b/keras/src/layers/layer.py
@@ -1277,7 +1277,7 @@ class Layer(BackendLayer, Operation):
     def quantized_build(self, input_shape, mode):
         raise self._not_implemented_error(self.quantized_build)
 
-    def quantize(self, mode, type_check=True, config=None):
+    def quantize(self, mode=None, type_check=True, config=None):
         raise self._not_implemented_error(self.quantize)
 
     def _check_quantize_args(self, mode, compute_dtype):


keras/src/quantizers/quantization_config.py
```diff
@@ -0,0 +1,166 @@
...
+class Float8QuantizationConfig(QuantizationConfig):
+    def __init__(self, weight_quantizer=None, activation_quantizer=None):
+        super().__init__(weight_quantizer, activation_quantizer)
+
+    @property
+    def mode(self):
+        return "float8"
+
+
+def validate_and_resolve_config(mode, config, name=None):
```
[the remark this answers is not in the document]
JyotinderSingh: Fixed.

keras/src/quantizers/quantization_config.py
```diff
@@ -0,0 +1,213 @@
...
+            config.get("activation_quantizer")
+        )
+        return cls(
+            weight_quantizer=weight_quantizer,
+            activation_quantizer=activation_quantizer,
+        )
+
+    @staticmethod
+    def weight_quantizer_or_default(config, default):
+        if config and config.weight_quantizer:
```
hertschuh: I've come to conclusion that we should check for `None` explicitly. I've come across too many bugs caused by either having objects be "falsy" or because `__bool__` raises.

```python
if config is not None and config.weight_quantizer is not None:
```

keras/src/quantizers/quantization_config.py
```diff
@@ -0,0 +1,213 @@
...
+            return config.weight_quantizer
+        return default
+
+    @staticmethod
+    def activation_quantizer_or_default(config, default):
+        if config and config.activation_quantizer:
+            return config.activation_quantizer
+        elif config and config.activation_quantizer is None:
+            return None
```
hertschuh: You can simplify these 4 lines in:
```python
if config is not None:
    return config.activation_quantizer
```

keras/src/quantizers/quantization_config.py
```diff
@@ -0,0 +1,213 @@
...
+            AbsMaxQuantizer with axis=-1.
+    """
+
+    def __init__(self, weight_quantizer=None, activation_quantizer="default"):
+        from keras.src.quantizers.quantizers import AbsMaxQuantizer
+
+        if activation_quantizer == "default":
+            activation_quantizer = AbsMaxQuantizer(axis=-1)
+        super().__init__(weight_quantizer, activation_quantizer)
+        if self.weight_quantizer:
```
hertschuh: `if  self.weight_quantizer is not None:`.

keras/src/quantizers/quantization_config.py
```diff
@@ -0,0 +1,213 @@
...
+            if hasattr(self.weight_quantizer, "value_range"):
+                if self.weight_quantizer.value_range != (-127, 127):
+                    raise ValueError(
+                        "Int8QuantizationConfig requires a weight_quantizer "
+                        "with value_range=(-127, 127). Received: "
+                        f"value_range={self.weight_quantizer.value_range}"
+                    )
```
hertschuh: There something a bit weird with this test. I think it's in the wrong place.

Should the quantizer itself verify that the `output_dtype` and `value_range` are compatible? (meaning this test would move to `AbsMaxQuantizer.__init__`).

However, shouldn't you verify here that `quantizer.output_dtype` is `"int8"`?

keras/src/quantizers/quantization_config.py
```diff
@@ -0,0 +1,213 @@
...
+            activation_quantizer = AbsMaxQuantizer(axis=-1)
+        super().__init__(weight_quantizer, activation_quantizer)
+        if self.weight_quantizer:
+            if hasattr(self.weight_quantizer, "value_range"):
+                if self.weight_quantizer.value_range != (-8, 7):
+                    raise ValueError(
+                        "Int4QuantizationConfig requires a weight_quantizer "
+                        "with value_range=(-8, 7). Received: "
+                        f"value_range={self.weight_quantizer.value_range}"
+                    )
```
hertschuh, line 118: Same comments:
- move this to `AbsMaxQuantizer`
- add check of `quantizer.output_dtype`

keras/src/quantizers/quantization_config.py
```diff
@@ -0,0 +1,213 @@
...
+    if config is None:
+        if mode == "int8":
+            config = Int8QuantizationConfig()
+        elif mode == "int4":
+            config = Int4QuantizationConfig()
+        elif mode == "float8":
+            config = Float8QuantizationConfig()
+        elif mode == "gptq":
+            raise ValueError(
+                "For GPTQ, you must pass a GPTQConfig object explicitly."
```
hertschuh: Mention the `config` argument in this message for clarity. Maybe something like

```
"For GPTQ, you must pass a `GPTQConfig` object in the `config` argument."
```

keras/src/quantizers/quantization_config.py
```diff
@@ -0,0 +1,213 @@
...
+        )
+
+    # Ensure mode is consistent
+    mode = config.mode
+
+    if mode not in QUANTIZATION_MODES:
+        raise ValueError(
+            "Invalid quantization mode. "
+            f"Expected one of {QUANTIZATION_MODES}. Received: mode={mode}"
+        )
```
hertschuh: Move this up line 159. The reason is that if config is None, people will see the less specific error message line 175. We want this message instead as it has more details.

keras/src/quantizers/__init__.py
```diff
@@ -13,7 +17,14 @@
...
 from keras.src.utils.naming import to_snake_case
 
-ALL_OBJECTS = {Quantizer, AbsMaxQuantizer}
+ALL_OBJECTS = {
+    Quantizer,
+    AbsMaxQuantizer,
+    QuantizationConfig,
+    Int8QuantizationConfig,
+    Int4QuantizationConfig,
+    Float8QuantizationConfig,
```
hertschuh, line 26: This is now mixing quantizers and quantization configs. That seems weird.

The truth is that `serialize` and `deserialize` are pretty useless.

keras/src/models/model.py
```diff
@@ -447,25 +448,76 @@ def get_quantization_layer_structure(self, mode):
...
+            **kwargs: Additional keyword arguments.
+
+        Example:
 
-        from keras.src.dtype_policies import QUANTIZATION_MODES
+        Quantize a model to int8 with default configuration:
 
+        ```python
+        # Build the model
+        >>> model = keras.Sequential([
```
hertschuh: Either you do `>>>` or you do ` ```python ` but not both

keras/src/models/model.py
```diff
@@ -447,25 +448,76 @@ def get_quantization_layer_structure(self, mode):
...
+        ```
+
+        Quantize a model to int8 with a custom configuration:
+
+        ```python
+        from keras.quantizers import Int8QuantizationConfig
+        from keras.quantizers import AbsMaxQuantizer
+
+        # Build the model
+        >>> model = keras.Sequential([
```
hertschuh: Either you do `>>>` or you do ` ```python ` but not both

keras/src/layers/core/dense.py
```diff
@@ -753,33 +771,41 @@ def quantize(self, mode, type_check=True, config=None):
...
-            kernel_value_int4, kernel_scale = quantizers.abs_max_quantize(
-                self._kernel,
-                axis=0,
-                value_range=(-8, 7),
-                dtype="int8",
-                to_numpy=True,
+            weight_quantizer = QuantizationConfig.weight_quantizer_or_default(
+                config,
+                quantizers.AbsMaxQuantizer(
+                    axis=0, value_range=(-8, 7), output_dtype="int8"
```
hertschuh, line 802: `output_dtype="int4"`, which shows the need for validation (other comment in this PR).

keras/src/layers/core/einsum_dense.py
```diff
@@ -938,19 +986,27 @@ def quantize(self, mode, type_check=True, config=None):
...
-                axis=self._kernel_reduced_axes,
-                value_range=(-8, 7),
-                dtype="int8",
-                to_numpy=True,
+            weight_quantizer = QuantizationConfig.weight_quantizer_or_default(
+                config,
+                quantizers.AbsMaxQuantizer(
+                    axis=self._kernel_reduced_axes,
+                    value_range=(-8, 7),
+                    output_dtype="int8",
```
hertschuh, line 1005: `output_dtype="int4"`

keras/src/layers/core/embedding.py
```diff
@@ -412,34 +414,45 @@ def quantize(self, mode, type_check=True, config=None):
...
-                axis=-1,
-                value_range=(-8, 7),
-                dtype="int8",
-                to_numpy=True,
+            weight_quantizer = QuantizationConfig.weight_quantizer_or_default(
+                config,
+                quantizers.AbsMaxQuantizer(
+                    axis=-1,
+                    value_range=(-8, 7),
+                    output_dtype="int8",
```
hertschuh, line 443: `output_dtype="int4"`

keras/src/layers/core/reversible_embedding.py
```diff
@@ -272,43 +290,57 @@ def _int4_call(self, inputs, reverse=False):
...
-                axis=-1,
-                value_range=(-8, 7),
-                dtype="int8",
-                to_numpy=True,
+            weight_quantizer = QuantizationConfig.weight_quantizer_or_default(
+                config,
+                quantizers.AbsMaxQuantizer(
+                    axis=-1,
+                    value_range=(-8, 7),
+                    output_dtype="int8",
```
hertschuh, line 339: `output_dtype="int4"`

keras/src/layers/core/reversible_embedding.py
```diff
@@ -317,13 +349,19 @@ def quantize(self, mode, type_check=True, config=None):
             )
             del self._embeddings
             if not self.tie_weights:
+                reverse_weight_quantizer = (
+                    QuantizationConfig.weight_quantizer_or_default(
+                        config,
+                        quantizers.AbsMaxQuantizer(
+                            axis=0,
+                            value_range=(-8, 7),
+                            output_dtype="int8",
```
hertschuh, line 358: `output_dtype="int4"`

hertschuh (commented):
Thanks for the PR!

Should all the `quantize` methods in the layers now have `mode=None`?

keras/src/layers/core/reversible_embedding.py
JyotinderSingh: `"int4"` dtype is only supported by JAX, therefore we use `"int8"` dtype as the output and then pack the numbers manually in a later step.

keras/src/layers/core/reversible_embedding.py
JyotinderSingh: `"int4"` dtype is only supported by JAX, therefore we use `"int8"` dtype as the output and then pack the numbers manually in a later step.

keras/src/layers/core/embedding.py
JyotinderSingh: `"int4"` dtype is only supported by JAX, therefore we use `"int8"` dtype as the output and then pack the numbers manually in a later step.

keras/src/layers/core/einsum_dense.py
JyotinderSingh: `"int4"` dtype is only supported by JAX, therefore we use `"int8"` dtype as the output and then pack the numbers manually in a later step.

keras/src/layers/core/dense.py
JyotinderSingh: `"int4"` dtype is only supported by JAX, therefore we use `"int8"` dtype as the output and then pack the numbers manually in a later step.

keras/src/quantizers/quantization_config.py
JyotinderSingh: Done

keras/src/quantizers/quantization_config.py
JyotinderSingh: Looks like we require a check at both the places. The first one at line 159 to avoid invalid modes passed as string, and one at the existing location to validate mode derived from the config (which may have been passed directly to the quantize call)

keras/src/models/model.py
JyotinderSingh: Fixed.

keras/src/models/model.py
JyotinderSingh: Fixed.

keras/src/quantizers/quantization_config.py
JyotinderSingh: Done

keras/src/quantizers/quantization_config.py
JyotinderSingh: Done.

keras/src/quantizers/quantization_config.py
JyotinderSingh: Fixed.

keras/src/quantizers/quantization_config.py
JyotinderSingh: Done.

keras/src/quantizers/quantization_config.py
JyotinderSingh: I'm continuing to check for the (-8, 7) value range here, since the check inside the AbsMaxQuantizer is too broad (since it checks for valid range of int8 dtype)

keras/src/quantizers/quantization_config.py
```diff
@@ -0,0 +1,227 @@
...
+    """
+
+    def __init__(self, weight_quantizer=None, activation_quantizer="default"):
+        from keras.src.quantizers.quantizers import AbsMaxQuantizer
+
+        if activation_quantizer == "default":
+            activation_quantizer = AbsMaxQuantizer(axis=-1)
+        super().__init__(weight_quantizer, activation_quantizer)
+        if self.weight_quantizer is not None:
+            if hasattr(self.weight_quantizer, "output_dtype"):
```
hertschuh, line 82: All quantizers have an `output_dtype`, you don't need this check

JyotinderSingh: Will fix this in the next PR which adds serialization support (since the test suite has now successfully completed on this version of the PR)

PR #21908, opened 2025-12-09
Fix: Remove redundant epsilon in loss mask weight calculation

Abhinavexists:
This PR fixes incorrect use of epsilon in mask weight calculation that caused numerical instability when all mask values are False.

In `keras/src/losses/loss.py`, the mask weight calculation used:

```
valid = ops.sum(mask)  # May be 0!
mask *= total/ valid + backend.epsilon()
```

This is problematic because:
- Adding backend.epsilon() i feel defeats this purpose by ensuring b is never actually zero
- When valid = 0 (all mask values are False), this computes total / epsilon ≈ 10^7, causing mask weights to explode to huge values instead of becoming 0

### DIFF
--- a/keras/src/losses/loss_test.py
+++ b/keras/src/losses/loss_test.py
@@ -287,17 +287,3 @@ class LossTest(testing.TestCase):
         loss_fn = ExampleLoss()
         loss = loss_fn(y_true, y_pred)
         self.assertDType(loss, backend.floatx())
-
-    def test_divide_no_nan(self):
-        x = np.array([1.0, 2.0, 3.0, 4.0])
-        y = np.zeros_like(x)
-
-        # divide_no_nan should return 0, not inf, when dividing by 0
-        self.assertAllClose(
-            ops.divide_no_nan(x, y),
-            y,
-        )
-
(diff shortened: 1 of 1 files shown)


keras/src/losses/loss_test.py
```diff
@@ -287,3 +287,17 @@ def test_default_dtype(self):
         loss_fn = ExampleLoss()
         loss = loss_fn(y_true, y_pred)
         self.assertDType(loss, backend.floatx())
+
+    def test_divide_no_nan(self):
```
hertschuh: This is already tested: https://github.com/keras-team/keras/blob/master/keras/src/ops/numpy_test.py#L2681-L2690

So no need to test this here.

Additionally, it looks like this case of all zeros mask is already covered:
https://github.com/keras-team/keras/blob/master/keras/src/losses/loss_test.py#L125-L129

hertschuh (commented):
Thanks for the PR!

keras/src/losses/loss_test.py
Abhinavexists: The test have been removed

PR #21911, opened 2025-12-10
Fix openvino thresholds for export

pctablet505:
This pull request updates the OpenVINO export tests in `keras/src/export/openvino_test.py` to use explicit tolerances when comparing model outputs. This ensures that small numerical differences between reference and OpenVINO outputs are tolerated, making the tests more robust to minor floating-point discrepancies.

**Testing improvements:**

* Updated all `self.assertAllClose` calls in the test cases to include `atol=1e-3` and `rtol=1e-3` for output comparisons, allowing for small absolute and relative differences between reference and OpenVINO outputs. [[1]](diffhunk://#diff-d4c65f893ea09fee392e95efda27ccd3cfe974f73ca82f675cf703e0ba5d04cbL122-R122) [[2]](diffhunk://#diff-d4c65f893ea09fee392e95efda27ccd3cfe974f73ca82f675cf703e0ba5d04cbL181-R181) [[3]](diffhunk://#diff-d4c65f893ea09fee392e95efda27ccd3cfe974f73ca82f675cf703e0ba5d04cbL246-R246)

### DIFF
--- a/keras/src/layers/core/dense.py
+++ b/keras/src/layers/core/dense.py
@@ -94,6 +94,12 @@ class Dense(Layer):
         lora_alpha=None,
         **kwargs,
     ):
+        if not isinstance(units, int) or units <= 0:
+            raise ValueError(
+                "Received an invalid value for `units`, expected a positive "
+                f"integer. Received: units={units}"
+            )
+
         super().__init__(activity_regularizer=activity_regularizer, **kwargs)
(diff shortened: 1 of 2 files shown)


hertschuh:
The tests are not failing at HEAD, so these are not needed.

Maybe you're seeing this on your local machine only.

PR #21922, opened 2025-12-12
OpenVino `device_scope` and data adapters tests

hertschuh:
The `device` scope feature was not working because of a missing import as well as incorrect device name logic. The existing device name logic had been copied from the torch implementation but is not applicable to OpenVino.

Also turned on all the data adapter related tests for OpenVino, which happen to exercise the `device` scope. Data adapters are used with `predict` and `evaluate` and are useful with OpenVino even without training support.

### DIFF
--- a/keras/src/backend/openvino/core.py
+++ b/keras/src/backend/openvino/core.py
@@ -13,7 +13,6 @@ from openvino import compile_model
 from keras.src import tree
 from keras.src.backend.common import KerasVariable
 from keras.src.backend.common import dtypes
-from keras.src.backend.common import global_state
 from keras.src.backend.common import standardize_dtype
 from keras.src.backend.common.dtypes import result_type
 from keras.src.backend.common.keras_tensor import KerasTensor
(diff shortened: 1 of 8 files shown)


keras/src/backend/openvino/excluded_tests.txt
```diff
@@ -36,5 +36,4 @@ keras/src/quantizers
 keras/src/random/seed_generator_test.py
 keras/src/regularizers
 keras/src/saving
-keras/src/trainers
```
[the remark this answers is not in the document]
hertschuh: Only including data adapters

keras/src/backend/openvino/core.py
```diff
@@ -530,31 +529,11 @@ def ov_to_keras_type(ov_type):
 
 @contextlib.contextmanager
 def device_scope(device_name):
-    current_device = _parse_device_input(device_name)
-    global_state.set_global_attribute("openvino_device", current_device)
+    yield
```
abheesht17, line 532: Why yield? That would make it a generator, right? Did you mean to write `pass` here?

abheesht17 (commented):
Left a comment

keras/src/backend/openvino/core.py
hertschuh: That how the `@contextlib.contextmanager` decorator works.

https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager

> The function being decorated must return a [generator](https://docs.python.org/3/glossary.html#term-generator)-iterator when called.

fchollet: Yeah it's because you use it via `with ... as ...`

PR #21924, opened 2025-12-12
Fix Muon optimizer with TensorFlow backend.

hertschuh:
Fixes https://github.com/keras-team/keras/issues/21793

The variable path cannot be accessed in the `update_step` as the variable passed is the underlying `tf.Variable` and not the `keras.Variable`. Instead, we rely on the variable index.


hertschuh:
FYI @pass-lin

PR #21929, opened 2025-12-16
Add RandomResizedCrop preprocessing layer

### LINKED ISSUE #21822: Add RandomResizedCrop
https://github.com/keras-team/keras-cv/issues/131

MalyalaKarthik66:
This PR adds `keras.layers.RandomResizedCrop`, a preprocessing layer that
randomly crops and resizes images to a fixed target size.

The layer:
- samples random crop area and aspect ratio during training
- applies a deterministic center crop during inference
- supports batched image inputs across all backends

This implementation supports image tensors only. For bounding boxes and
segmentation masks, users can rely on `keras_cv.layers.RandomResizedCrop`.

Closes #21822

### DIFF
--- a/keras/src/layers/preprocessing/image_preprocessing/random_resized_crop.py
+++ b/keras/src/layers/preprocessing/image_preprocessing/random_resized_crop.py
@@ -200,12 +200,17 @@ class RandomResizedCrop(BaseImagePreprocessingLayer):
 
     def compute_output_shape(self, input_shape):
         input_shape = list(input_shape)
-        if len(input_shape) == 4 and input_shape[-1] in (1,3,4):
-            input_shape[1] = self.height
-            input_shape[2] = self.width
-        else:
-            input_shape[self.height_axis] = self.height
-            input_shape[self.width_axis] = self.width
+        rank = len(input_shape)
+        if rank < 3:
+            return tuple(input_shape)
+        ha = (
+            self.height_axis
+            if self.height_axis >= 0
+            else rank + self.height_axis
+        )
+        wa = self.width_axis if self.width_axis >= 0 else rank + self.width_axis
+        input_shape[ha] = self.height
+        input_shape[wa] = self.width
         return tuple(input_shape)
 
     def get_config(self):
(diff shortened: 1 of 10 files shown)


MalyalaKarthik66:
@hertschuh  Gentle reminder on this PR when convenient.

There is still one CI failure on the Torch backend, although the same test passes locally. I may be missing something subtle here—could you please take a look when you have time? Thanks!

hertschuh (commented):
First sorry for the delay on this.

For consistency with all the other image preprocessing layers, this should also support bounding boxes and masks.

Note that once this PR is in: https://github.com/keras-team/keras/pull/22192

It will be quite easy to add support for masks.

keras/src/layers/preprocessing/image_preprocessing/random_resized_crop.py
```diff
@@ -0,0 +1,227 @@
...
+        if transformation is None:
+            transformation = self.get_random_transformation(images, training)
+        h, w, ch, cw = transformation
+        images = self._slice_images(images, h, w, ch, cw)
+        return backend.image.resize(
+            images,
+            size=(self.height, self.width),
+            interpolation=self.interpolation,
+            data_format=self.data_format,
+        )
```
hertschuh: If you rebase to master to get https://github.com/keras-team/keras/pull/22192 then you can rename this to:
```python
def _transform_images(self, images, transformation, interpolation):
```
And you'll get support for masks automatically. But you still need to support bounding boxes.

keras/src/layers/preprocessing/image_preprocessing/random_resized_crop_test.py
```diff
@@ -0,0 +1,230 @@
...
+SKIP_BACKENDS = ["openvino"]
+
+pytestmark = pytest.mark.skipif(
+    backend.backend() in SKIP_BACKENDS,
+    reason=(
+        "RandomResizedCrop tests not supported for backend: {}".format(
+            backend.backend()
+        )
+    ),
+)
```
hertschuh, line 18: Instead of this, please add a line `RandomResizedCropTest` to this file: https://github.com/keras-team/keras/blob/master/keras/src/backend/openvino/excluded_concrete_tests.txt#L148

hertschuh (commented):
Sorry again for the delay.

Also, the tests are failing on Torch. The tests need to work with `channels_first` instead of `channels_last`. You can look at how the other tests do this.

PR #21930, opened 2025-12-16
Fix Normalization.adapt() for supervised tf.data.Dataset inputs

MalyalaKarthik66:
This PR fixes `Normalization.adapt()` for supervised `tf.data.Dataset` inputs that yield `(x, y)` or `(x, y, sample_weight)` tuples.

The layer:

- correctly infers `input_shape` when `data.element_spec` is a tuple
- computes normalization statistics using only the feature tensor `x`
- continues to support NumPy arrays, backend tensors, PyDataset, and unlabeled `tf.data.Dataset`

This change also adds a regression test, `test_adapt_tf_dataset_with_labels`, which verifies that `adapt()` runs successfully on a supervised `tf.data.Dataset` and produces the expected mean and variance.

### DIFF
--- a/keras/src/layers/preprocessing/normalization.py
+++ b/keras/src/layers/preprocessing/normalization.py
@@ -264,20 +264,20 @@ class Normalization(DataLayer):
         if isinstance(data, np.ndarray) or backend.is_tensor(data):
             input_shape = data.shape
         elif isinstance(data, tf.data.Dataset):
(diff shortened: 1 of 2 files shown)


MalyalaKarthik66:
@hertschuh  Gentle reminder on this PR when convenient — it adds support for supervised `tf.data.Dataset` inputs in `Normalization.adapt()`. Thanks!

PR #21935, opened 2025-12-18
Update feature_space.py

orbin123:
Allow `adapt()` to accept Python iterables in FeatureSpace.

* FeatureSpace: Updated `adapt()` to detect dictionary inputs (of lists/arrays) and automatically convert them to tf.data.Dataset. Previously, it yielded a ValueError demanding a Dataset object.

This change enable users to pass standard Python data structures directly to the preprocessing layer, improving usability and consistency with other layers like `IntegerLookup`.

### DIFF
--- a/keras/src/layers/preprocessing/normalization.py
+++ b/keras/src/layers/preprocessing/normalization.py
@@ -236,14 +236,11 @@ class Normalization(DataLayer):
                 # handling (x, y) or (x, y, sample_weight)
                 data = data[0]
             input_shape = data.shape
-        elif isinstance(data, (list, tuple)):
-            data = np.array(data)
-            input_shape = data.shape
         else:
             raise TypeError(
                 f"Unsupported data type: {type(data)}. `adapt` supports "
-                f"`np.ndarray`, backend tensors, `tf.data.Dataset`, "
-                f"`keras.utils.PyDataset`, `list`, and `tuple`."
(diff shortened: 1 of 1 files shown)


orbin123:
Could gemini explain the issue. Is the issue saying there should be a human reviewer to push or is there any other configurations

keras/src/layers/preprocessing/normalization.py
```diff
@@ -236,11 +236,14 @@ def adapt(self, data):
                 # handling (x, y) or (x, y, sample_weight)
                 data = data[0]
             input_shape = data.shape
+        elif isinstance(data, (list, tuple)):
+            data = np.array(data)
+            input_shape = data.shape
```
hertschuh: I don't think we can do that. There is some ambiguity because the data itself can be a tuple/list of arrays (see line 235). There is no way to disambiguate between a tuple/list of what should be arrays vs. an array of one more dimension.

Users should be explicit and do `layer.adapt(np.array(python_array)))`

hertschuh (commented):
Hi @orbin123 

Thanks for the PR.

Can you add unit tests covering the changes in `feature_space.py`?

orbin123:
Thankyou for the opportunity
I have reverted the changes in normalisation file as you have suggested and added the unit test function for the update.

hertschuh (approved):
Thanks for the update.

Issue #21942, opened 2025-12-19
Training stateful RNN fails due to symbolic tensor


fhchl:
```python
import keras
import numpy as np

keras.config.disable_traceback_filtering()

model = keras.Sequential(
    [
        keras.layers.Input(shape=(10, 10), batch_size=2),
        keras.layers.GRU(5, stateful=True),
    ]
)
model.compile(optimizer="adam", loss="mse")

x = np.random.random((2, 10, 10))
y = np.random.random((2, 10, 5))

model.fit(x, y, epochs=1)
```

gives

```
Traceback (most recent call last):
  File "/home/fheuchel/projects/nxnx/shape_bug.py", line 17, in <module>
    model.fit(x, y, epochs=1)
  File "/home/fheuchel/projects/nxnx/.venv/lib/python3.11/site-packages/keras/src/utils/traceback_utils.py", line 113, in error_handler
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/fheuchel/projects/nxnx/.venv/lib/python3.11/site-packages/keras/src/backend/tensorflow/trainer.py", line 399, in fit
    logs = self.train_function(iterator)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fheuchel/projects/nxnx/.venv/lib/python3.11/site-packages/keras/src/backend/tensorflow/trainer.py", line 241, in function
    opt_outputs = multi_step_on_iterator(iterator)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fheuchel/projects/nxnx/.venv/lib/python3.11/site-packages/tensorflow/python/util/traceback_utils.py", line 153, in error_handler
    raise e.with_traceback(filtered_tb) from None
  File "/home/fheuchel/projects/nxnx/.venv/lib/python3.11/site-packages/keras/src/backend/tensorflow/trainer.py", line 154, in multi_step_on_iterator
    one_step_on_data(iterator.get_next())
  File "/home/fheuchel/projects/nxnx/.venv/lib/python3.11/site-packages/keras/src/backend/tensorflow/trainer.py", line 125, in wrapper
    result = step_func(converted_data)
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fheuchel/projects/nxnx/.venv/lib/python3.11/site-packages/keras/src/backend/tensorflow/trainer.py", line 134, in one_step_on_data
[...]

PR #21945, opened 2025-12-20
Fix handling of symbolic Tensor in RNN

danielenricocahall:
Per bug reported in https://github.com/keras-team/keras/issues/21942, the batch size check for RNNs added in https://github.com/keras-team/keras/pull/21742 does not for support symbolic tensors, which happens when we use `ops.shape` (@hertschuh actually mentioned this in the original PR that 21742 was based upon [here](https://github.com/keras-team/keras/pull/21249#discussion_r2264087272)). To mitigate this, the `ops.shape` check has been updated to use the `.shape` field. A test has been added based on the original reproduction case - for simplicity, I have made it only for the tensorflow backend.

### DIFF
--- a/keras/src/layers/rnn/gru_test.py
+++ b/keras/src/layers/rnn/gru_test.py
@@ -408,17 +408,7 @@ class GRUTest(testing.TestCase):
         "for TensorFlow backend.",
     )
     def test_stateful_with_symbolic_batch_size(self):
-        from keras.src import backend
-
-        if backend.backend() != "tensorflow":
-            self.skipTest(
-                "Test only applicable to fixing a bug for TensorFlow backend."
-            )
-
-        layer = layers.GRU(
-            5,
-            stateful=True,
-        )
+        layer = layers.GRU(5, stateful=True)
 
         x_concrete = np.ones((2, 10, 10), dtype=np.float32)
         _ = layer(x_concrete, training=True)


keras/src/layers/rnn/gru_test.py
```diff
@@ -400,3 +400,33 @@ def test_legacy_implementation_argument(self):
...
+
+    def test_stateful_with_symbolic_batch_size(
+        self,
+    ):
+        from keras.src import backend
+
+        if backend.backend() != "tensorflow":
+            self.skipTest(
+                "Test only applicable to fixing a bug for TensorFlow backend."
+            )
```
hertschuh: Use `@pytest.mark.skipif(` to decorate the test instead of skipping within the test

keras/src/layers/rnn/gru_test.py
```diff
@@ -400,3 +400,33 @@ def test_legacy_implementation_argument(self):
             tpu_atol=1e-3,
             tpu_rtol=1e-3,
         )
+
+    def test_stateful_with_symbolic_batch_size(
+        self,
+    ):
```
hertschuh: Rename `test_stateful_with_dynamic_batch_size`.
Also remove the comma after `self` and let it be a single line.

keras/src/layers/rnn/gru_test.py
```diff
@@ -400,3 +400,33 @@ def test_legacy_implementation_argument(self):
...
+        @tf.function(
+            input_signature=[
+                tf.TensorSpec(shape=(None, 10, 10), dtype=tf.float32)
+            ]
+        )
+        def f(x):
+            return layer(x, training=True)
+
+        y = f(x_concrete)
+        self.assertEqual(y.shape, (2, 5))
```
hertschuh: I think this test can be re-written this way:

```python
model = models.Sequential([layers.GRU(5, stateful=True)])
model.jit_compile = True

x = np.ones((5, 10, 10), dtype=np.float32)
model(x)

y = model.predict(x, batch_size=2)
self.assertEqual(y.shape, (2, 5))
```

Which I like better because you may be able to run it with other backends since there's no TF specific code. The trick is that `predict` will use a dynamic batch size because it has to.

Let me know if this doesn't work or it doesn't exercise the issue you're fixing.

hertschuh (commented):
Thanks for the fix!

Some comments about the unit test.

keras/src/layers/rnn/gru_test.py
danielenricocahall: Done, thank you!

keras/src/layers/rnn/gru_test.py
danielenricocahall: Renamed and reformatted

keras/src/layers/rnn/gru_test.py
danielenricocahall: I actually had a similar idea and the original test I had was written to that effect. Unfortunately it didn't exercise what's being tested (as in, I could revert the change and the test still passes)

keras/src/layers/rnn/gru_test.py
```diff
@@ -400,3 +401,36 @@ def test_legacy_implementation_argument(self):
...
+        reason="Test only applicable to fixing a bug with symbolic batch size "
+        "for TensorFlow backend.",
+    )
+    def test_stateful_with_symbolic_batch_size(self):
+        from keras.src import backend
+
+        if backend.backend() != "tensorflow":
+            self.skipTest(
+                "Test only applicable to fixing a bug for TensorFlow backend."
+            )
```
hertschuh: Remove, this is now a decorator

keras/src/layers/rnn/gru_test.py
```diff
@@ -400,3 +401,36 @@ def test_legacy_implementation_argument(self):
...
+
+        layer = layers.GRU(
+            5,
+            stateful=True,
+        )
```
hertschuh: Combine these on a single line.

keras/src/layers/rnn/gru_test.py
danielenricocahall: Ah sorry I missed this, removed

keras/src/layers/rnn/gru_test.py
danielenricocahall: done

hertschuh (approved):

PR #21953, opened 2025-12-25
 Add `CLAHE` preprocessing layer

### LINKED ISSUE #21898: Add CLAHE
CLAHE is a variant of Adaptive histogram equalization (AHE) which takes care of over-amplification of the contrast.

Initial ticket: [keras-cv/#359](https://github.com/keras-team/keras-cv/issues/359)
Initial PR: [keras-cv/#420](https://github.com/keras-team/keras-cv/pull/420)

**pseudo code**

```python
# https://github.com/tensorflow/models
@tf.function
def equalize(image, mode='grayscale'):
    def scale_channel(im, c):
        """Scale the data in the channel to implement equalize."""
        im = tf.cast(im[..., c], tf.int32)
        # Compute the histogram of the image channel.
        histo = tf.histogram_fixed_width(im, [0, 255], nbins=256)

        # For the purposes of computing the step, filter out the nonzeros.
        nonzero = tf.where(tf.not_equal(histo, 0))
        nonzero_histo = tf.reshape(tf.gather(histo, nonzero), [-1])
        step = (tf.reduce_sum(nonzero_histo) - nonzero_histo[-1]) // 255

        def build_lut(histo, step):
            # Compute the cumulative sum, shifting by step // 2
            # and then normalization by step.
            lut = (tf.cumsum(histo) + (step // 2)) // step
            # Shift lut, prepending with 0.
            lut = tf.concat([[0], lut[:-1]], 0)
            # Clip the counts to be in range.  This is done
            # in the C code for image.point.
            return tf.clip_by_value(lut, 0, 255)

        # If step is zero, return the original image.  Otherwise, build
        # lut from the full histogram and step and then index from it.
        result = tf.cond(
            tf.equal(step, 0), lambda: im,
            lambda: tf.gather(build_lut(histo, step), im))
        return tf.cast(result, tf.uint8)

    if mode == 'grayscale':
        image = scale_channel(image, 0)
        return tf.cast(image, tf.float32)
    elif mode == 'rgb':
        s1 = scale_channel(image, 0)
        s2 = scale_channel(image, 1)
        s3 = scale_channel(image, 2)
        image = tf.stack([s1, s2, s3], -1)
        return tf.cast(image, tf.float32)
```

Example:

AlanPonnachan:
## Add `CLAHE` (Contrast Limited Adaptive Histogram Equalization) preprocessing layer

This PR adds the `CLAHE` layer to `keras.layers`. CLAHE is a variant of Adaptive Histogram Equalization (AHE) that prevents over-amplification of contrast by limiting the histogram amplification in local tiles.

Fixes #21898

**Inference on X-Rays**

*Generated using `clip_limit=3.0`, `tile_grid_size=(8, 8)`.*

**1. Chest X-Ray**
chest

**2. Hand X-Ray**
hand

**3. Oxford Flowers Dataset**
flower1

flower2

**Implementation Details**

1.  **Tiling Strategy**: Uses `ops.reshape` and `ops.transpose` to split input images into `(GridH, GridW)` tiles vectorially.
2.  **Histogram Calculation**: Uses a vectorized approach with `ops.one_hot` and `ops.sum`. 
3.  **Interpolation**: Implements bilinear interpolation by flattening the computed CDFs and using `ops.take`. 
4.  **Padding**: Uses `ops.pad` with `mode="symmetric"` to handle boundary tiles correctly.

**Changes**

*   Added `CLAHE` class in `keras/src/layers/preprocessing/image_preprocessing/clahe.py`
*   Added unit tests in `clahe_test.py` covering value ranges, dtypes, and shapes.

**Colab Notebook**

[CLAHE test](https://colab.research.google.com/drive/1SepbTQOIwb0fd-byPiT1_hW63WwdWVCH?usp=sharing)

### DIFF
--- a/keras/src/layers/preprocessing/image_preprocessing/clahe_test.py
+++ b/keras/src/layers/preprocessing/image_preprocessing/clahe_test.py
@@ -44,7 +44,7 @@ class ContrastLimitedAdaptiveHistogramEqualizationTest(testing.TestCase):
             np.float32
         )
         layer = layers.ContrastLimitedAdaptiveHistogramEqualization(
-            value_range=(0, 255)
+            value_range=(0, 255), data_format="channels_last"
         )
         xs_out = layer(xs)
         self.assertAllInRange(ops.convert_to_numpy(xs_out), 0, 255)
@@ -58,7 +58,7 @@ class ContrastLimitedAdaptiveHistogramEqualizationTest(testing.TestCase):
             dtype
         )
         layer = layers.ContrastLimitedAdaptiveHistogramEqualization(
-            value_range=(0, 255)
+            value_range=(0, 255), data_format="channels_last"
         )
         xs_out = ops.convert_to_numpy(layer(xs))
         self.assertAllInRange(xs_out, 0, 255)
@@ -69,7 +69,7 @@ class ContrastLimitedAdaptiveHistogramEqualizationTest(testing.TestCase):
             size=(2, 32, 32, 3), low=lower, high=upper
         ).astype(np.float32)
         layer = layers.ContrastLimitedAdaptiveHistogramEqualization(
-            value_range=(lower, upper)
+            value_range=(lower, upper), data_format="channels_last"
         )
         xs_out = ops.convert_to_numpy(layer(xs))
         self.assertAllInRange(xs_out, lower, upper)
@@ -85,7 +85,7 @@ class ContrastLimitedAdaptiveHistogramEqualizationTest(testing.TestCase):
 
     def test_tf_data_compatibility(self):
         layer = layers.ContrastLimitedAdaptiveHistogramEqualization(
-            value_range=(0, 255)
+            value_range=(0, 255), data_format="channels_last"
         )
         input_data = np.random.random((2, 16, 16, 3)) * 255
         ds = tf_data.Dataset.from_tensor_slices(input_data).batch(2).map(layer)


keras/src/layers/preprocessing/image_preprocessing/clahe.py
```diff
@@ -0,0 +1,271 @@
...
+    # For normalized images [0, 1]
+    clahe_normalized = keras.layers.CLAHE(value_range=(0.0, 1.0))
+    norm_image = np.random.rand(1, 256, 256, 3)
+    output_norm = clahe_normalized(norm_image)
+    ```
+    """
+
+    def __init__(
+        self,
+        value_range=(0, 255),
```
innat, line 77: Just curious, I know other preocessing layer in keras use this param (value range 0 to 255). I wonder, what if the samples are not all scale uniformly, i.e. medical image nor non-matrual image.

Instead, how about `value_range=auto`- computing large and min value per sample and compute accordingly.

AlanPonnachan: Hi @innat, thanks for the suggestion! 

I think it might be better to keep `value_range` fixed (requiring explicit input) for a few key reasons:

1.  The sibling layer `keras.layers.Equalization` strictly requires a fixed `value_range`. Keeping `CLAHE` aligned with that pattern makes the image preprocessing API more predictable for users.
2.  Keras already provides `keras.layers.AutoContrast` for dynamic range adjustment. If a user has unscaled inputs, the better way would be to chain them: `x = AutoContrast()(x); x = CLAHE()(x)`. Embedding that logic inside CLAHE might complicate the layer's single responsibility.
3.   For X-rays or CT scans , "auto" scaling can sometimes be risky. For example, if an X-ray is intentionally underexposed (dark), automatically stretching it to 0-255 before CLAHE would massively amplify background noise, potentially creating artifacts that look like tissue. Requiring an explicit range ensures the user knows exactly how their data is being interpreted

Let me know your thoughts!

AlanPonnachan:
@gbaned any updates?

innat:
@hertschuh 
Could you please take a look into this PR. Thanks.

AlanPonnachan:
@hertschuh 
- unit tests are passing now!
- the layer is renamed to `ContrastLimitedAdaptiveHistogramEqualization`

keras/src/layers/preprocessing/image_preprocessing/clahe.py
```diff
@@ -0,0 +1,298 @@
+from keras.src import backend
+from keras.src import ops
```
hertschuh: Actually, you should not use `keras.ops` in preprocessing layers.

In your code below, replace all occurrences of `ops.` with `self.backend.` or `self.backend.numpy.` etc.

See this for an example:
https://github.com/keras-team/keras/blob/master/keras/src/layers/preprocessing/image_preprocessing/auto_contrast.py#L66-L68

keras/src/layers/preprocessing/image_preprocessing/clahe.py
```diff
@@ -0,0 +1,298 @@
...
+        c_idx = ops.arange(channels, dtype="int32")[None, None, None, :]
+
+        y0_e = y0[None, :, None, None]
+        y1_e = y1[None, :, None, None]
+
+        x0_e = x0[None, None, :, None]
+        x1_e = x1[None, None, :, None]
+
+        wy_e = wy[None, :, None, None]
+        wx_e = wx[None, None, :, None]
```
hertschuh, line 279: I'm surprised this trick works with all backends.

keras/src/layers/preprocessing/image_preprocessing/clahe_test.py
```diff
@@ -0,0 +1,103 @@
...
+        self.assertEqual(out.shape[-1], 1)
+        self.assertAllInRange(out, 0, 255)
+
+    @pytest.mark.skipif(
+        backend.backend() != "tensorflow",
+        reason=(
+            "tf.data compatibility map strictly requires the "
+            "tensorflow backend."
+        ),
+    )
```
hertschuh: Remove this, once you switch from `ops.` to `self.backend.` it should pass for all backends.

keras/src/layers/preprocessing/image_preprocessing/clahe.py
AlanPonnachan: Thanks for pointing this out!

keras/src/layers/preprocessing/image_preprocessing/clahe.py
AlanPonnachan: I think all backends have standardized their indexing APIs to match NumPy's behavior! Because of this, it seamlessly translates to an expand_dims operation under the hood across all the backends without crashing.

AlanPonnachan:
@hertschuh 
all unit tests are passing now!


> Everything looks good to me, but do you understand why it's failing on torch with the tf dataset? It seems unrelated to torch.

I think PyTorch CI environment defaults to channels_first. In test_tf_data_compatibility, the input array was hardcoded to (2, 16, 16, 3), but the layer instantiation was missing data_format="channels_last".

Because of this, the layer interpreted the input as BCHW, making width=3. To fit the (8, 8) grid, it calculated a required width padding of 5. Since this executes inside `tf.data`, it delegates to TensorFlow's `MirrorPad` op.
From the [mirrorpad doc](https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/mirror-pad), I think `MirrorPad` op strictly forbids padding a dimension by an amount larger than its own size (5 > 3), causing the crash. 

I've pushed a fix that explicitly sets `data_format="channels_last"` across all the unit tests so the CI environments don't misinterpret the test arrays anymore!

PR #21956, opened 2025-12-26
Use `kokoro:force-run` label for TPU tests too.

hertschuh:
The `tpu_tests_labeler` workflow doesn't work because the `pull_request_review` workflow is run with read-only token permissions. This is a restriction imposed on workflow triggered from PRs from forks.
    
For now, we will reuse the `kokoro:force-run` label, which is handled by the `google-ml-butler` application.


hertschuh (commented):
Testing

PR #21959, opened 2025-12-27
fix image.extract_patches strides handling

calad0i:
In the `keras.ops.image.extract_patches` function, the current codebase has reference to `strides` that is undefined in `compute_output_spec` when it is actually provided instead of left as `None` by default. This PR fixes the issue and added a few lines in tests for it.


hertschuh (approved):
Thank you for the fix!

PR #21961, opened 2025-12-28
Fix ReversibleEmbedding mask error when using reverse=True

MarcosAsh:
## Summary
Using ReversibleEmbedding with mask_zero=True and reverse=True causes a shape mismatch error. 

## Problem
The mask from the parent Embedding (Embedding.compute_mask()) class has shape (batch, seq_len) but the reverse output has shape (batch, seq_len, vocab_size). Causing a boradcasting error in the loss calculation.

## Solution
Added compute_mask() to return None when inputs are 3D (reverse mode). This prevents incompatible mask from being propagated.

## Test
```python
embedding = keras.layers.ReversibleEmbedding(100, 32, mask_zero=True)
inputs = keras.Input(shape=(None,))
x = embedding(inputs)
x = keras.layers.GRU(32, return_sequences=True)(x)
logits = embedding(x, reverse=True)  # This now works
```

Fixes keras-team/keras-hub#2498

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -31,7 +31,7 @@ jobs:
       PYTHON: ${{ matrix.python-version }}
       KERAS_HOME: .github/workflows/config/${{ matrix.backend }}
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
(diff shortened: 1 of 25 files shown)


MarcosAsh:
Thanks for the feedback just implemented the changes.

PR #21966, opened 2025-12-31
Add simple example for keras.layers.Resizing

shashaka:
Adds a minimal example showing how to use keras.layers.Resizing, using a built-in Keras dataset and no external assets.

If this style of example is acceptable, I can add similar minimal examples for other preprocessing layers that currently do not have examples.

### DIFF
--- a/keras/src/layers/preprocessing/image_preprocessing/resizing.py
+++ b/keras/src/layers/preprocessing/image_preprocessing/resizing.py
@@ -70,17 +70,11 @@ class Resizing(BaseImagePreprocessingLayer):
     Example:
 
     ```python
-    import keras
-
     (x_train, y_train), _ = keras.datasets.cifar10.load_data()
-
     image = x_train[0]
-
     resizer = keras.layers.Resizing(128, 128)
     resized_image = resizer(image)
-
-    print("Original:", image.shape)
-    print("Resized:", resized_image.shape)
+    print("original:", image.shape, "resized:", resized_image.shape)
     ```
     """
 


keras/src/layers/preprocessing/image_preprocessing/resizing.py
```diff
@@ -66,6 +66,22 @@ class Resizing(BaseImagePreprocessingLayer):
             `~/.keras/keras.json`. If you never set it, then it will be
             `"channels_last"`.
         **kwargs: Base layer keyword arguments, such as `name` and `dtype`.
+
+    Example:
+
+    ```python
+    import keras
```
hertschuh: Remove, we don't import keras in examples, it's implied.

keras/src/layers/preprocessing/image_preprocessing/resizing.py
```diff
@@ -66,6 +66,22 @@ class Resizing(BaseImagePreprocessingLayer):
...
+
+    (x_train, y_train), _ = keras.datasets.cifar10.load_data()
+
```
hertschuh: nitpick: remove this extra line break to make the example more compact.

keras/src/layers/preprocessing/image_preprocessing/resizing.py
```diff
@@ -66,6 +66,22 @@ class Resizing(BaseImagePreprocessingLayer):
...
+    image = x_train[0]
+
+    resizer = keras.layers.Resizing(128, 128)
+    resized_image = resizer(image)
+
+    print("Original:", image.shape)
+    print("Resized:", resized_image.shape)
```
hertschuh: nitpick: combine in a single line to make the example more compact:
```python
print("original:", image.shape, "resized:", resized_image.shape)
```

keras/src/layers/preprocessing/image_preprocessing/resizing.py
shashaka: Removed.

keras/src/layers/preprocessing/image_preprocessing/resizing.py
shashaka: Removed.

keras/src/layers/preprocessing/image_preprocessing/resizing.py
shashaka: Updated.

hertschuh (approved):

PR #21967, opened 2025-12-31
[OpenVINO backend] Fix and enable numpy.rot90

PES2UG23CS205:
This PR fixes and enables numpy.rot90 support for the OpenVINO backend in Keras.

The implementation follows NumPy semantics and aligns with existing OpenVINO backend patterns by expressing rot90 using supported OpenVINO ops.

What was done

Implemented numpy.rot90 in keras/src/backend/openvino/numpy.py

Normalized and validated k and axes to match NumPy behavior

Implemented rotation using transpose and reverse OpenVINO ops

Removed NumPyTestRot90 from excluded_concrete_tests.txt

Fixed OpenVINO test failures related to rot90

### DIFF
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -1179,13 +1179,13 @@ def rot90(array, k=1, axes=(0, 1)):
     result = array
 
     for _ in range(k):
-        # 1️⃣ Transpose axis1 <-> axis2
+        # 1️ Transpose axis1 <-> axis2
         perm = list(range(ndim))
         perm[axis1], perm[axis2] = perm[axis2], perm[axis1]
         perm_const = ov_opset.constant(perm, Type.i32).output(0)
         result = ov_opset.transpose(result, perm_const).output(0)
 
-        # 2️⃣ Reverse along axis1 using StridedSlice
+        # 2️ Reverse along axis1 using StridedSlice
         begin = [0] * ndim
         end = [0] * ndim
         strides = [1] * ndim


hertschuh:
@PES2UG23CS205 

Something weird happened here. Github shows that both files are fully changed. My guess is that the line breaks changed from Unix still to DOS style. Can you undo this and rebase? Thanks!

PES2UG23CS205:
Thanks for pointing this out.

You’re right, this looks like a line-ending (CRLF vs LF) issue.
I’ll undo the line-ending changes, rebase on top of master, and push an updated commit shortly.

PES2UG23CS205:
Thanks for pointing this out  it was a CRLF vs LF issue.
I’ve reset the files, rebased on top of master, and force-pushed a clean commit with only the intended changes.

PES2UG23CS205:
I attempted to run api_gen.py locally, but hit a known TensorFlow/JAX dependency conflict (ml_dtypes version mismatch).
Since CI is already generating the API changes successfully, could a maintainer please apply the api_gen updates or advise on the preferred workflow here?

keras/src/backend/openvino/numpy.py
```diff
@@ -1143,6 +1143,42 @@ def flip(x, axis=None):
         end_mask=all_ones_mask,
     )
     return OpenVINOKerasTensor(result.output(0))
+def rot90(x, k=1, axes=(0, 1)):
```
hertschuh: Rename `x` to `array`. This is what's making the tests fail.

hertschuh:
> I attempted to run api_gen.py locally, but hit a known TensorFlow/JAX dependency conflict (ml_dtypes version mismatch). Since CI is already generating the API changes successfully, could a maintainer please apply the api_gen updates or advise on the preferred workflow here?

I can run it myself. Just curious though, can't you upgrade TensorFlow?

keras/src/backend/openvino/numpy.py
PES2UG23CS205: Thanks for the review!

I’ve renamed the argument from `x` to `array` to match NumPy’s signature and pushed an updated commit.

PES2UG23CS205:
I’ve fixed the OpenVINO `flip` implementation to derive the rank from the underlying OpenVINO tensor instead of `OpenVINOKerasTensor`, and pushed the update.

hertschuh (commented):
@PES2UG23CS205 

The code looks good thanks!

But you need to remove line 1 of `keras/src/backend/openvino/excluded_concrete_tests.txt` to enable the testing of `rot90`.

Thanks!

PES2UG23CS205:
Thanks for the review!
I’ve removed the rot90 entry from excluded_concrete_tests.txt to enable the tests and pushed the update.
Please let me know if anything else is needed.

keras/src/backend/openvino/numpy.py
```diff
@@ -1145,6 +1151,61 @@ def flip(x, axis=None):
...
+    result = array
+
+    for _ in range(k):
+        # 1️⃣ Transpose axis1 <-> axis2
+        perm = list(range(ndim))
+        perm[axis1], perm[axis2] = perm[axis2], perm[axis1]
+        perm_const = ov_opset.constant(perm, Type.i32).output(0)
+        result = ov_opset.transpose(result, perm_const).output(0)
+
+        # 2️⃣ Reverse along axis1 using StridedSlice
```
hertschuh: Sorry, super small nitpick, but can you remove the emojis for the numbers in the comments?

We generally don't do that, I think it's distracting.

PES2UG23CS205:
Thanks for pointing that out, and sorry about that.
I’ve removed the emojis from the comments and pushed the update

hertschuh (approved):

PR #21977, opened 2026-01-03
Add fill_mode="crop" to RandomRotation

### LINKED ISSUE #21954: Usages of Keras Augmentaiton Layers
I was trying to use RandomRotation in a custom class based setup with tf.data API.

```python
class RandRotateKeras3D:
    def __init__(self, keys, factor=0.1, prob=0.8, fill_value=0.0):
        self.keys = keys
        self.prob = prob
        self.fill_value = fill_value
        self.seed_val = np.random.randint(0, 10**6)

        self.img_rotate = layers.RandomRotation(
            factor=factor, 
            fill_mode="constant",
            interpolation="bilinear", 
            fill_value=fill_value,
            seed=self.seed_val
        )
        
        self.label_rotate = layers.RandomRotation(
            factor=factor, 
            fill_mode="constant",
            interpolation="nearest",
            fill_value=0.0,
            seed=self.seed_val
        )
        
    def __call__(self, inputs):
        image_key, label_key = self.keys
        img = inputs.data[image_key]
        lbl = inputs.data[label_key]
        
        img_shape = tf.shape(img)
        D, H, W = img_shape[0], img_shape[1], img_shape[2]

        # Padding
        pad_h, pad_w = H // 2, W // 2
        img_padded = tf.pad(
            img, 
            [[0, 0], [pad_h, pad_h], [pad_w, pad_w], [0, 0]], 
            constant_values=tf.cast(self.fill_value, img.dtype)
        )
        lbl_padded = tf.pad(
            lbl, 
            [[0, 0], [pad_h, pad_h], [pad_w, pad_w], [0, 0]], 
            constant_values=tf.cast(0, lbl.dtype)
        )

        # Transpose to (Batch=1, H, W, Channels=D)
        img_trans = tf.transpose(img_padded, [3, 1, 2, 0]) 
        lbl_trans = tf.transpose(lbl_padded, [3, 1, 2, 0])

        img_rot = self.img_rotate(img_trans)
        lbl_rot = self.label_rotate(lbl_trans)

        # Transpose back and crop
        img_final = tf.transpose(img_rot, [3, 1, 2, 0])
        lbl_final = tf.transpose(lbl_rot, [3, 1, 2, 0])
        
        img_final = tf.image.resize_with_crop_or_pad(img_final, H, W)
        lbl_final = tf.image.resize_with_crop_or_pad(lbl_final, H, W)

        return ...
```
I got

```bash
ValueError: in user code:

    File "/tmp/ipykernel_47/2758838664.py", line 48, in train_transformation  *
        result = pipeline(data)
    File "/usr/local/lib/python3.11/dist-packages/medicai/transforms/base.py", line 53, in __call__  *
        x = transform(x)
    File "/tmp/ipykernel_47/285916447.py", line 138, in __call__  *
        img_rot = self.img_rotate(img_trans)
    File "/usr/local/lib/python3.11/dist-packages/keras/src/layers/preprocessing/data_layer.py", line 119, in __call__  **
        outputs = super().__call__(inputs, **kwargs)
    File "/usr/local/lib/python3.11/dist-packages/keras/src/utils/traceback_utils.py", line 122, in error_handler

    ValueError: Exception encountered when calling RandomRotation.call().
    
    tf.function only supports singleton tf.Variables created on the first call. Make sure the tf.Variable is only created once or created outside tf.function. See https://www.tensorflow.org/guide/function#creating_tfvariables for more information.
    
    Arguments received by RandomRotation.call():
      • data=<tf.Tensor 'transpose:0' shape=(1, None, None, None) dtype=float32>
      • training=True
```

## Queries

1. In the setup described above, is it considered invalid or discouraged to instantiate random augmentation layers once and invoke them inside the call method, particularly when used within a `tf.data` pipeline or `tf.function`? I tested the code outside the tf.data and it works well. Furthermore, while checking out the output, I noticed that, due to rotation, some border effect was obvious. To remedy that, would it be nice to add `cropping` after the rotation, i.e. see the following output (`A` is not desired but `B`) - taken from [here](https://github.com/kobiso/Image-Rotation-and-Cropping-tensorflow/tree/master).

Image

2. I aimed to implement a custom random-rotation layer. However, at present, there is no equivalent operation in Keras 3 to `tfa.image.rotate`. It would be highly beneficial if such a stateless, low-level rotation primitive were available natively in Keras 3. Hopefully, it will behave nicely in tf.data API regardless of the backend.
3. Regarding the existing [Keras image augmentation layers](https://keras.io/api/layers/preprocessing_layers/image_augmentation/):     - These layers do not expose a probability (**prob**) argument to explicitly control whether a random transformation is applied. What is the motivation behind omitting such a parameter, given that probabilistic control is closely tied to augmentation behavior and is important for analyzing experimental impact? I thought it was added in `keras-cv` with `rate` flag, not sure if it is still true.
     - The current API is designed to operate on a single input tensor, which is appropriate for image classification. However, for segmentation tasks, how is one expected to reliably ensure that identical random transformations are applied to both the input image and its corresponding label or mask? Additionally, since labels are discrete, even minor geometric distortions can introduce spatial noise—how is this concern intended to be addressed within the current design?
     - Lastly, the official documentation for several Keras augmentation layers does not provide concrete code examples. For instance, layers such as [RandomShear](https://keras.io/api/layers/preprocessing_layers/image_augmentation/random_shear/) and [RandomTranslation](https://keras.io/api/layers/preprocessing_layers/image_augmentation/random_translation/) are described only at the API level, without minimal, runnable examples demonstrating their usage, inputs, and expected outputs. This makes the behavior of these layers unintuitive to understand. Including simple, illustrative examples would significantly improve usability, user engagement, and confidence in these APIs.

MalyalaKarthik66:
This PR adds `fill_mode="crop"` to `keras.layers.RandomRotation`.

### What’s new
- Introduces a new `fill_mode="crop"` that removes border artifacts by  
  **rotating on the original canvas with constant fill → angle-dependent center crop → resize**,  
  while preserving the input shape.
- Uses exact **angle-dependent largest inscribed rectangle (LRR)** math so the effective zoom
  scales smoothly with rotation angle  
  (e.g. for square inputs: ~1 at 0° → ~√2 at 45°), avoiding over-cropping for small rotations.
- Fully backend-agnostic and batched, with consistent behavior across TensorFlow, JAX, NumPy,
  and Torch.
- Supports structured inputs (`images`, `segmentation_masks`, `bounding_boxes`) with
  synchronized transformations.
- Adds clear documentation and runnable examples directly in the layer docstring, including
  a visual comparison of crop vs constant fill.

Fixes #21954.

### DIFF
--- a/keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
+++ b/keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
@@ -1,90 +1,37 @@
 from keras.src.api_export import keras_export
 from keras.src.layers.preprocessing.image_preprocessing.base_image_preprocessing_layer import (  # noqa: E501
-    BaseImagePreprocessingLayer,
 )
 from keras.src.layers.preprocessing.image_preprocessing.base_image_preprocessing_layer import (  # noqa: E501
     base_image_preprocessing_transform_example,
 )
 from keras.src.layers.preprocessing.image_preprocessing.bounding_boxes import (
-    converters,
-)
+                fill_mode=transformation.get("fill_mode", self.fill_mode),
+                fill_value=transformation.get("fill_value", self.fill_value),
 from keras.src.random.seed_generator import SeedGenerator
 
-
-@keras_export("keras.layers.RandomRotation")
-class RandomRotation(BaseImagePreprocessingLayer):
-    """A preprocessing layer which randomly rotates images during training.
-
-    This layer will apply random rotations to each image, filling empty space
-    according to `fill_mode`.
-
-    By default, random rotations are only applied during training.
-    At inference time, the layer does nothing. If you need to apply random
-    rotations at inference time, pass `training=True` when calling the layer.
-
-    Input pixel values can be of any range (e.g. `[0., 1.)` or `[0, 255]`) and
-    of integer or floating point dtype.
-    By default, the layer will output floats.
-
-    **Note:** This layer is safe to use inside a `tf.data` or `grain` pipeline
-    (independently of which backend you're using).
-
-    Input shape:
-        3D (unbatched) or 4D (batched) tensor with shape:
-        `(..., height, width, channels)`, in `"channels_last"` format
-
-    Output shape:
-        3D (unbatched) or 4D (batched) tensor with shape:
-        `(..., height, width, channels)`, in `"channels_last"` format
-
-    Args:
-        factor: a float represented as fraction of 2 Pi, or a tuple of size 2
-            representing lower and upper bound for rotating clockwise and
-            counter-clockwise. A positive values means rotating
-            counter clock-wise,
-            while a negative value means clock-wise.
-            When represented as a single
-            float, this value is used for both the upper and lower bound.
-            For instance, `factor=(-0.2, 0.3)`
-            results in an output rotation by a random
-            amount in the range `[-20% * 360, 30% * 360]`.
-            `factor=0.2` results in an
-            output rotating by a random amount
-            in the range `[-20% * 360, 20% * 360]`.
-        fill_mode: Points outside the boundaries of the input are filled
-            according to the given mode
-            (one of `{"constant", "reflect", "wrap", "nearest"}`).
-            - *reflect*: `(d c b a | a b c d | d c b a)`
-                The input is extended by reflecting about
-                the edge of the last pixel.
-            - *constant*: `(k k k k | a b c d | k k k k)`
-                The input is extended by
-                filling all values beyond the edge with
-                the same constant value k = 0.
-            - *wrap*: `(a b c d | a b c d | a b c d)` The input is extended by
-                wrapping around to the opposite edge.
-            - *nearest*: `(a a a a | a b c d | d d d d)`
-                The input is extended by the nearest pixel.
-        interpolation: Interpolation mode. Supported values: `"nearest"`,
-            `"bilinear"`.
-        seed: Integer. Used to create a random seed.
-        fill_value: a float represents the value to be filled outside
-            the boundaries when `fill_mode="constant"`.
-        data_format: string, either `"channels_last"` or `"channels_first"`.
-            The ordering of the dimensions in the inputs. `"channels_last"`
-            corresponds to inputs with shape `(batch, height, width, channels)`
-            while `"channels_first"` corresponds to inputs with shape
-            `(batch, channels, height, width)`. It defaults to the
-            `image_data_format` value found in your Keras config file at
-            `~/.keras/keras.json`. If you never set it, then it will be
-            `"channels_last"`.
+            - `"reflect"`: Reflects values at the edge.
+            - `"constant"`: Fills with the constant value `fill_value`.
+            - `"wrap"`: Wraps around to the opposite edge.
+            - `"nearest"`: Extends the nearest edge value.
+            - `"crop"`: Rotates with an angle-dependent zoom to remove border
+              artifacts without explicit cropping or resizing.
+        interpolation: Interpolation mode. Supported values are `"nearest"` and
+            `"bilinear"`. Segmentation masks always use `"nearest"`
+            interpolation.
+        seed: Optional integer used to create a deterministic random seed.
+        fill_value: Float value used when `fill_mode="constant"`.
+        data_format: One of `"channels_last"` or `"channels_first"`. Defaults to
+            the global Keras image data format.
+        bounding_box_format: String specifying the format of bounding boxes when
+            `"bounding_boxes"` are provided (e.g., `"xyxy"`, `"xywh"`). Required
+            when using bounding box inputs.
 
     Example:
 
     {{base_image_preprocessing_transform_example}}
     """
 
-    _SUPPORTED_FILL_MODE = ("reflect", "wrap", "constant", "nearest")
+    _SUPPORTED_FILL_MODE = ("reflect", "wrap", "constant", "nearest", "crop")
     _SUPPORTED_INTERPOLATION = ("nearest", "bilinear")
 
     def __init__(
@@ -95,9 +42,15 @@ class RandomRotation(BaseImagePreprocessingLayer):
         seed=None,
         fill_value=0.0,
         data_format=None,
+        bounding_box_format=None,
         **kwargs,
     ):
-        super().__init__(factor=factor, data_format=data_format, **kwargs)
+        super().__init__(
+            factor=factor,
+            data_format=data_format,
+            bounding_box_format=bounding_box_format,
+            **kwargs,
+        )
         self.seed = seed
         self.generator = SeedGenerator(seed)
         self.fill_mode = fill_mode
@@ -107,12 +60,12 @@ class RandomRotation(BaseImagePreprocessingLayer):
 
         if self.fill_mode not in self._SUPPORTED_FILL_MODE:
             raise NotImplementedError(
-                f"Unknown `fill_mode` {fill_mode}. Expected of one "
+                f"Unknown `fill_mode` {fill_mode}. Expected one of "
                 f"{self._SUPPORTED_FILL_MODE}."
             )
         if self.interpolation not in self._SUPPORTED_INTERPOLATION:
             raise NotImplementedError(
-                f"Unknown `interpolation` {interpolation}. Expected of one "
+                f"Unknown `interpolation` {interpolation}. Expected one of "
                 f"{self._SUPPORTED_INTERPOLATION}."
             )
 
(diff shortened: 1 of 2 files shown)


shashaka:
It would be very helpful to understand the logic if you could add a gist for this PR. You may refer to my sample gist below.

https://colab.research.google.com/drive/1n_0OzVKFHz5g0bFNOjQNulAAIjBuXvOT#scrollTo=lNIsN7fVPl3P

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
```diff
@@ -109,79 +220,236 @@ def __init__(
...
+            return ops.expand_dims(x, axis=0), True
+        return x, False
+
+    def _maybe_unbatch(self, x, was_unbatched):
+        if was_unbatched:
+            return ops.squeeze(x, axis=0)
+        return x
+
+    def _pad_for_crop(self, images, height, width):
+        """Pad so rotation can be center-cropped without border artifacts."""
```
hertschuh: Do you need to pad at all if all the padding is supposed to be removed in the end?

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
```diff
@@ -240,6 +507,7 @@ def get_config(self):
         config = {
             "factor": self.factor,
             "data_format": self.data_format,
+            "bounding_box_format": self.bounding_box_format,
```
hertschuh: This is already done by `super.get_config()`, you don't need it here.

hertschuh (commented):
Thanks for all the work on this!

My main comment is this actually: https://github.com/keras-team/keras/issues/21954#issuecomment-3781339047

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
MalyalaKarthik66: Thanks — padding is no longer needed, so I’ve removed it and now rotate on the original canvas followed by crop + resize. I’ve updated the code.

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
MalyalaKarthik66: Thanks — removed the redundant fields.

MalyalaKarthik66:
> It would be very helpful to understand the logic if you could add a gist for this PR. You may refer to my sample gist below.
> 
> https://colab.research.google.com/drive/1n_0OzVKFHz5g0bFNOjQNulAAIjBuXvOT#scrollTo=lNIsN7fVPl3P

@shashaka I’ve added a Colab gist that walks through the crop logic both visually and numerically (angle-dependent safe crop / zoom), and compares `fill_mode="constant"` vs `fill_mode="crop"` side-by-side.

Colab gist:
https://colab.research.google.com/drive/1qymnDOJ8ku2vXfwzSMrOFLtu12b7-A9F

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
```diff
@@ -1,3 +1,4 @@
+from keras.src import ops
```
hertschuh: Oh... you can't do that in this layer.

You have to change every `ops.xxx` to `self.backend.xxx` in the code. This is a preprocessing layer that may run TF in tf.data even if the current backend is something else.

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
```diff
@@ -10,74 +11,166 @@
...
-    By default, the layer will output floats.
+    can be integer or floating-point. The output is always floating-point.
 
-    **Note:** This layer is safe to use inside a `tf.data` or `grain` pipeline
-    (independently of which backend you're using).
+    **Note:** This layer is safe to use inside `tf.data` and `grain` input
+    pipelines. When used in a `tf.data` pipeline, the layer correctly handles
+    symbolic tensors across supported backends.
+
+    **Structured inputs**
```
hertschuh: This is how we do these kinds of sections:
- Use `##` instead of `**`
- Leave an empty line after

This way it create links in the table of contents on the right [example](https://keras.io/api/layers/backend_specific_layers/jax_layer/)

```
symbolic tensors across supported backends.

## Structured inputs

This layer...
```

Please change this one and the other ones below.

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
```diff
@@ -10,74 +11,166 @@
...
+    This layer supports synchronized augmentation of images and associated data
```
hertschuh: Change `synchronized` to `combined`, otherwise it makes you think of sync vs async.

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
```diff
@@ -10,74 +11,166 @@
...
+        `(..., height, width, channels)` when `data_format="channels_last"`, or
+        `(..., channels, height, width)` when `data_format="channels_first"`.
 
     Output shape:
-        3D (unbatched) or 4D (batched) tensor with shape:
-        `(..., height, width, channels)`, in `"channels_last"` format
+        Same as the input shape.
+
+    Examples:
+    **Basic usage with crop mode:**
```
hertschuh: I think what is done here will work and will be enough of an example:

https://github.com/keras-team/keras/pull/21978/changes#diff-3a44984f30aac97ce019e13e9871b1f5c6400ec7c2849dee689f62f36b839412

So just rebase so that you get the `{{base_image_preprocessing_transform_example}}` thing.

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
```diff
@@ -10,74 +11,166 @@
...
+        factor: A float or a tuple of two floats representing a fraction of a
+            full rotation (360 degrees). If a single float is provided, the
+            rotation angle is sampled uniformly from
+            `[-factor * 360, factor * 360]`. If a tuple `(lower, upper)` is
+            provided, the angle is sampled from
+            `[lower * 360, upper * 360]`.
+        fill_mode: Points outside the input boundaries are filled according to
+            the given mode (one of
+            `{"constant", "reflect", "wrap", "nearest", "crop"}`).
+
```
hertschuh: Remove empty line

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
```diff
@@ -109,79 +208,234 @@ def __init__(
...
+        )
+
+        cropped = self._center_crop_to(rotated, crop_h, crop_w)
+
+        resized = self._resize_back(
+            cropped,
+            orig_h,
+            orig_w,
+            interpolation,
+        )
```
hertschuh: Oh my idea was that you would do the math so that `_apply_affine` would already have the zoom factor from `_resize_back` combined into it so that all you have to do is `_apply_affine`. No cropping needed and no resizing needed. Does that make sense?

It would be faster and the image quality would be better.

And what's great about it too is that you wouldn't need to special case `fill_mode="crop"` in the rest of the code (`transform_images` etc.).

MalyalaKarthik66:
@hertschuh Thanks for the review — I’ve addressed all comments, rebased on master, and updated the PR with the angle-dependent zoom embedded directly in the affine transform.

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
```diff
@@ -107,88 +131,141 @@ def __init__(
...
+
+        out = self._apply_affine(
+            images=segmentation_masks,
+            rotation_matrix=transformation["rotation_matrix"],
+            interpolation="nearest",
+            fill_mode=transformation.get("fill_mode", self.fill_mode),
+            fill_value=transformation.get("fill_value", 0.0),
         )
+        return self._maybe_unbatch(out, was_unbatched)
```
hertschuh: Well, now it's identical to `transform_images` so you can go back to just calling `transform_images`.

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
```diff
@@ -206,50 +283,80 @@ def get_random_transformation(self, data, training=True, seed=None):
...
+
         return {
             "angle": angle,
             "rotation_matrix": rotation_matrix,
-            "image_height": image_height,
-            "image_width": image_width,
-            "batch_size": batch_size,
+            "image_height": self.backend.cast(image_height, "int32"),
+            "image_width": self.backend.cast(image_width, "int32"),
+            "batch_size": self.backend.cast(batch_size, "int32"),
```
hertschuh: Why did you add the `cast` here?

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
```diff
@@ -206,50 +283,80 @@ def get_random_transformation(self, data, training=True, seed=None):
...
+            "fill_mode": fill_mode,
+            "fill_value": fill_value,
         }
 
-    def compute_output_shape(self, input_shape):
-        return input_shape
```
hertschuh, line 234: `compute_output_shape` got removed. Can you add it back?

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
```diff
@@ -206,50 +283,80 @@ def get_random_transformation(self, data, training=True, seed=None):
...
+        angles_rad = self.backend.numpy.deg2rad(angles)
+        sin_a = self.backend.numpy.abs(self.backend.numpy.sin(angles_rad))
+        cos_a = self.backend.numpy.abs(self.backend.numpy.cos(angles_rad))
+
+        denom_w = width * cos_a + height * sin_a
+        denom_h = width * sin_a + height * cos_a
+
+        scale_w = width / denom_w
+        scale_h = height / denom_h
+        scale = self.backend.numpy.minimum(scale_w, scale_h)
```
hertschuh: I don't think the math is correct here. For one thing, it gives you a scale factor less than 1 when it should always be greater or equal to 1.

`denom_w` and `denom_h` are the width and height of the rotated image with fill, but getting to the scale factor is more complicated and requires more math.

This answer has the right illustration: https://stackoverflow.com/a/21347469

What you have with `denom_w` and `denom_h` is `W1` and `H1`. But what we need is the scaling factor related to `W2` and `H2`, which are more complicated to calculate. Unfortunately the formula in the answer is incorrect AFAICT. Somehow, I couldn't find the correct formula anywhere.

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
```diff
@@ -206,50 +283,80 @@ def get_random_transformation(self, data, training=True, seed=None):
...
+
+        min_scale = self.backend.numpy.min(scale)
```
hertschuh: Why this? `scale` is already a scalar, isn't it?

hertschuh (commented):
Thanks for the various fixes you did along the way!

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
MalyalaKarthik66: this cast isn’t required anymore. It was added during debugging and can be safely removed. I’ll drop it.

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
hertschuh: Actually you where super close. Everything up to `denom_w` / `denom_h` was correct. Your code was computing the size of an *inscribed* rotated image, but we want the reverse, the *circumscribed* rotated mage.

So, simply do:
```python
scale = self.backend.numpy.maximum(denom_w / width, denom_h / height)
return scale
```

Except that for clarity, I would rename `denom_w` to `minimum_width` and `denom_h` to `minimum_height`.

hertschuh (changes requested):
Do you mind finishing this?

I figured out the formula for the scale.

Then, if you rebase, things will be simpler. You only have to override:
```python
    def _transform_images(self, images, transformation, interpolation):
```

The default implementation of `transform_images` and `transform_segmentation_masks` will work and don't need to be overriden.

MalyalaKarthik66:
@hertschuh Sorry for the late reply, and thanks for reopening the PR.

I've rebased the branch on the latest `master` and simplified the implementation accordingly. The layer now overrides `_transform_images` only and relies on the base class implementations of `transform_images` and `transform_segmentation_masks`, as suggested.

I've also updated the PR accordingly. Please let me know if anything else should be adjusted.

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
MalyalaKarthik66: You're right — I renamed `denom_w/denom_h` to `minimum_width/minimum_height` and removed the earlier `min_scale` reduction.

In the current implementation I kept `minimum(width / minimum_width, height / minimum_height)` so that the affine transform scale remains ≤ 1, which zooms out the rotated image and avoids fill artifacts in crop mode.

Happy to switch to the `maximum(minimum_width / width, minimum_height / height)` formulation if you'd prefer matching that exactly.

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
```diff
@@ -13,78 +13,96 @@
...
+    explicit cropping or resizing.
+
+    This preserves output shape while avoiding fill regions. For batched
+    inputs, the zoom uses the maximal scale across the batch for uniform
+    processing. This mode is particularly useful for dense prediction tasks
+    such as segmentation.
+
+    When using structured inputs, the zoomed rotation is applied to images
+    and segmentation masks. Bounding boxes are rotated and clipped
+    (zoom intentionally not applied).
```
hertschuh, line 64: Why would the zoom not be applied to bounding boxes?

You need to apply the same transformation other the bounding boxes are not placed correctly on the image.

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
```diff
@@ -121,64 +145,74 @@ def _transform_images(self, images, transformation, interpolation):
...
+            boxes=boxes,
+            angle=angle,
+            translate_x=zeros,
+            translate_y=zeros,
+            scale=ones,
+            shear_x=zeros,
+            shear_y=zeros,
+            height=height,
+            width=width,
+        )
```
hertschuh, line 188: You need to apply `transformation["rotation_matrix"],` instead.

keras/src/layers/preprocessing/image_preprocessing/random_rotation_test.py
```diff
@@ -75,3 +76,209 @@ def test_tf_data_compatibility(self):
...
+        output = ops.convert_to_numpy(layer(images, training=True))
+
+        self.assertEqual(output.shape, shape)
+
+        fill_ratio = np.mean(output < 1e-3)
+        self.assertLess(
+            fill_ratio,
+            0.06,
+            f"Batch crop still contains fill artifacts: {fill_ratio:.2%}",
+        )
```
hertschuh, line 265: These 2 tests are already covered by `test_random_rotation_crop_avoids_fill_artifacts` in my opinion.

hertschuh (changes requested):
Thanks for the update!

Can you rebase and address the bounding boxes?

PR #21979, opened 2026-01-04
Fix: config keys for `chain_depth` and `num_chains` in `aug_mix.py`

Abhinavexists:
Fixed bug in `AugMix.get_config()` where `num_chains` and `chain_depth` were swapped


hertschuh (approved):
Good catch, thanks!

PR #21980, opened 2026-01-04
Unify extract_patches to support both 2D and 3D patches

MarcosAsh:
## Summary
Unified extract_patches to handle both 2D and 3D patch extraction based on the size argument, similar to how keras.ops.conv handles different dimensions.

## Changes
size=(h, w) -> 2D patch extraction. This already existed.
size=(d, h, w) -> 3D patch extraction 

## Why
As suggested in #21624, having a unified API is more consistent with other Keras ops like conv.

## Tests
Updated test_extract_patches_invalid_size to reflect new valid 3D sizes.
Added test_extract_patches_unified_3d  for the new functionality.
All existing extract_patches tests pass.

Relates to #21624

### DIFF
--- a/keras/src/ops/image.py
+++ b/keras/src/ops/image.py
@@ -627,7 +627,7 @@ def extract_patches(
             For 2D patches: 3D `(H, W, C)` or 4D `(N, H, W, C)`.
             For 3D patches: 4D `(D, H, W, C)` or 5D `(N, D, H, W, C)`.
         size: Patch size as int or tuple.
-            Length 2 tuple `(patch_height, patch_width)` for 2D patches.
+            Length 2 tuple `(patch_height, patch_width)` or int for 2D patches.
             Length 3 tuple `(patch_depth, patch_height, patch_width)` for
             3D patches.
         strides: Strides for patch extraction. If not specified, defaults


keras/src/ops/image.py
```diff
@@ -616,42 +616,98 @@ def extract_patches(
...
-            if `None`, it defaults to the same value as `size`.
-        dilation_rate: This is the input stride, specifying how far two
-            consecutive patch samples are in the input. For value other than 1,
-            strides must be 1. NOTE: `strides > 1` is not supported in
-            conjunction with `dilation_rate > 1`
+        images: Input image/volume or batch of images/volumes.
+            For 2D patches: 3D `(H, W, C)` or 4D `(N, H, W, C)`.
+            For 3D patches: 4D `(D, H, W, C)` or 5D `(N, D, H, W, C)`.
+        size: Patch size as int or tuple.
+            Length 2 tuple `(patch_height, patch_width)` for 2D patches.
```
hertschuh: Small tweak: mention that if the size is an int it will do 2D patches:

```
    size: Patch size as int or tuple.
        Length 2 tuple `(patch_height, patch_width)` or int for 2D patches.
```

MarcosAsh:
Hey just committed the changes thanks for the help!

PR #21986, opened 2026-01-06
benchmarks: add RandomRotation tf.data performance benchmark

PES2UG23CS205:
This PR adds a simple benchmark script for measuring the performance of
keras.layers.RandomRotation when used inside a tf.data pipeline.

The benchmark compares eager execution vs tf.function wrapping and
measures end-to-end dataset iteration time.

Motivation:
There is currently no reference benchmark for Keras preprocessing layers.
This script helps evaluate performance characteristics and potential
regressions.

Related issue:
https://github.com/keras-team/keras/issues/21954

cc @mattdangerw

### DIFF
--- a/benchmarks/layer_benchmark/random_rotation_benchmark.py
+++ b/benchmarks/layer_benchmark/random_rotation_benchmark.py
@@ -3,7 +3,7 @@
 from absl import app
 from absl import flags
 
-from keras.benchmarks.layer_benchmark.base_benchmark import LayerBenchmark
+from benchmarks.layer_benchmark.base_benchmark import LayerBenchmark
 
 FLAGS = flags.FLAGS
 


PES2UG23CS205:
Thanks for the feedback!

I’ve moved the benchmark to `keras/benchmarks/layer_benchmark/random_rotation_benchmark.py` and refactored it to use the standard `LayerBenchmark` machinery, consistent with the other layer benchmarks.

I also removed the `use_tf_function` comparison. In my local testing I didn’t observe any meaningful difference in performance, and since it shouldn’t matter here, I’ve kept the benchmark focused on the standard execution path.

Please let me know if this looks aligned with expectations.

keras/benchmarks/layer_benchmark/random_rotation_benchmark.py
```diff
@@ -0,0 +1,19 @@
...
+            layer=layers.RandomRotation(0.1),
+            input_shape=(256, 224, 224, 3),
+            num_batches=100,
+        )
+
+    def make_dataset(self):
+        images = tf.random.uniform(self.input_shape, dtype=tf.float32)
+        ds = tf.data.Dataset.from_tensor_slices(images)
+        ds = ds.batch(32)
+        return ds
```
hertschuh: This is dead code, nothing will call `make_dataset`.

keras/benchmarks/layer_benchmark/random_rotation_benchmark.py
```diff
@@ -0,0 +1,19 @@
+import tensorflow as tf
+
+from keras import layers
+from keras.benchmarks.layer_benchmark.layer_benchmark import LayerBenchmark
+
+
+class RandomRotationBenchmark(LayerBenchmark):
```
hertschuh: I don't think you need to subclass.

keras/benchmarks/layer_benchmark/random_rotation_benchmark.py
```diff
@@ -0,0 +1,19 @@
+import tensorflow as tf
```
hertschuh: Can you follow the exact structure of the other files? For instance: https://github.com/keras-team/keras/blob/master/benchmarks/layer_benchmark/core_benchmark.py

- imports
- a function called `benchmark_random_rotation` which creates a `LayerBenchmark` and calls `benchmark_predict` and `benchmark_train` (although predict is a no-op).
- define `BENCHMARK_NAMES`
- add the `main` and related logic

PES2UG23CS205:
Thanks for the clarification!

I’ve updated the RandomRotation benchmark to follow the exact structure used by the other layer benchmarks (mirroring `core_benchmark.py`), including the function-based setup, BENCHMARK_NAMES mapping, and standard CLI boilerplate.

Please let me know if this now looks aligned.

keras/benchmarks/layer_benchmark/random_rotation_benchmark.py
```diff
@@ -0,0 +1,65 @@
+"""Benchmark RandomRotation layer."""
+
+from absl import app
+from absl import flags
+
+from keras.benchmarks.layer_benchmark.base_benchmark import LayerBenchmark
```
hertschuh: Oh, you put the file in the wrong folder. It should be `benchmarks/layer_benchmark/random_rotation_benchmark.py` with the other benchmark files instead of `keras/benchmarks/layer_benchmark/random_rotation_benchmark.py`.

Then, the import here should be from `benchmarks.layer_benchmark.base_benchmark import LayerBenchmark` like the other ones.

PES2UG23CS205: Thanks for catching that , you’re right.

I’ve moved the benchmark to `benchmarks/layer_benchmark/random_rotation_benchmark.py` alongside the other layer benchmarks and updated the import to use `benchmarks.layer_benchmark.base_benchmark.LayerBenchmark`, consistent with the existing files.

Please let me know if this now looks correct.

hertschuh (approved):

hertschuh:
Yes, that was correct. Thanks!

PR #21997, opened 2026-01-11
Fixed enable dynamic shape handling in CenterCrop

MarcosAsh:
## Summary

Support dynamic spatial dimensions in CenterCrop. Previously the layer could raise a ValueError when height/width were unknown at build time.

## Changes

- Compute runtime spatial shape and branch between crop vs resize.
- Keep existing behavior for inputs smaller than the crop box by resizing with crop_to_aspect_ratio=True.
- Add test coverage for dynamic spatial dims.

## Why

CenterCrop should work in Functional/graph models and tf.data pipelines where spatial dims may be dynamic.

## Tests

- Added test_dynamic_spatial_dims for (None, None, 3) inputs (crop + resize cases).
- pytest keras/src/layers/preprocessing/image_preprocessing/center_crop_test.py -v

## Reproducibility of error
This is how I checked the function doesn't accept dynamic shape handling. 
```
python
import numpy as np
import tensorflow as tf
from keras import layers, models

class OldCenterCrop(layers.Layer):
    def __init__(self, height, width, **kwargs):
        super().__init__(**kwargs)
        self.height = height
        self.width = width

    def call(self, inputs):
        h, w = inputs.shape[1], inputs.shape[2]
        if h is None or w is None:
            raise ValueError(
                f"At this time, CenterCrop can only process images with a static "
                f"spatial shape. Received: inputs.shape={inputs.shape}"
            )
        top = (h - self.height) // 2
        left = (w - self.width) // 2
        return inputs[:, top:top + self.height, left:left + self.width, :]

inputs = layers.Input(shape=(None, None, 3))
model = models.Model(inputs, OldCenterCrop(10, 12)(inputs))
model(np.random.random((2, 25, 30, 3)).astype("float32"))
```

### DIFF
--- a/keras/src/layers/preprocessing/image_preprocessing/center_crop.py
+++ b/keras/src/layers/preprocessing/image_preprocessing/center_crop.py
@@ -1,4 +1,3 @@
-from keras.src.utils import image_utils
 from keras.src.api_export import keras_export
 from keras.src.layers.preprocessing.image_preprocessing.base_image_preprocessing_layer import (  # noqa: E501
     BaseImagePreprocessingLayer,
@@ -9,6 +8,7 @@ from keras.src.layers.preprocessing.image_preprocessing.bounding_boxes.converter
 from keras.src.layers.preprocessing.image_preprocessing.bounding_boxes.converters import (  # noqa: E501
     convert_format,
 )
+from keras.src.utils import image_utils
 
 
 @keras_export("keras.layers.CenterCrop")


hertschuh:
@MarcosAsh 

I noticed two things:
- When the image dimensions are dynamic, you skip the cropping and only do the resizing
- The new unit test as-is works with the unmodified `CenterCrop` code, so it doesn't actually exercise the case when the dimensions are dynamic

It's a bit tricky to explain how to fix these so I ended up just putting up a PR: https://github.com/keras-team/keras/pull/22020

Thanks for your work on this though!

PR #21998, opened 2026-01-11
Add uninitialized() function to support prim::Uninitialized in OpenVI…

Vbhatt03:
dd `uninitialized()` function to support `prim::Uninitialized` operation in OpenVINO backend.

Since OpenVINO doesn't support truly uninitialized data, the function returns a zeros tensor (following the same approach used in OpenVINO's PyTorch Frontend).

Fixes #29727

### DIFF
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -1083,6 +1083,16 @@ def empty_like(x, dtype=None):
     return zeros_like(x, dtype=dtype)
 
 
+def uninitialized(shape, dtype=None):
+    """Return a tensor of given shape and type filled with zeros.
+
+    In OpenVINO, uninitialized data is not supported, so we create a tensor
+    filled with zeros with the given shape and type. This implements support
+    for prim::Uninitialized operation.
+    """
+    return empty(shape, dtype=dtype)
+
+
 def equal(x1, x2):
     element_type = None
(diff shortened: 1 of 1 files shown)


Vbhatt03:
.take

hertschuh (commented):
This code is not connected to anything and is therefore unreachable.

What is it supposed to do?

hertschuh:
#29727 is not in this repo, this must be a mistake.

Feel free to reopen if otherwise.

Vbhatt03:
Hi @hertschuh , thank you for reviewing!

This code is connected to Keras through the backend selection mechanism.

When users set KERAS-BACKEND=OPENVINO (via environment variable or keras config), the OpenVINO backend becomes active and all keras.ops.* operations use these implementations.

Regarding issue #29727 - this references the OpenVINO repository's tracking issue for the Keras backend integration work (https://github.com/openvinotoolkit/openvino/issues/29727) , not a Keras issue. This PR is part of that effort to add OpenVINO as an official Keras backend alongside TensorFlow, JAX, PyTorch, and NumPy.

Would be happy to clarify anything further!

hertschuh:
The fix needs to happen the other way around:
- https://github.com/openvinotoolkit/openvino/issues/29727 needs to be fixed in the OpenVino repo
- then, and only then, Keras's `numpy.empty` OpenVino implementation needs to make use of the new uninitialized feature

PR #22002, opened 2026-01-13
Implement complete Keras-Orbax checkpoint integration

amitsrivastava78:
- Add compile config saving to OrbaxCheckpoint for full model restoration
- Update saving_api.py to use compile_from_config for proper model loading
- Replace direct Orbax API tests with Keras saving API tests
- Remove unused Orbax imports from test file
- Add comprehensive test coverage for all model state components
- Ensure cross-backend compatibility (TensorFlow, JAX, PyTorch)

This enables users to save and load complete model state (weights, optimizer, metrics) using keras.saving.load_model() with Orbax checkpoints.

### DIFF
--- a/keras/src/callbacks/orbax_checkpoint_test.py
+++ b/keras/src/callbacks/orbax_checkpoint_test.py
@@ -7,8 +7,11 @@ from absl.testing import parameterized
 from keras.src import backend
 from keras.src import layers
 from keras.src import models
+from keras.src import saving
 from keras.src import testing
+from keras.src import utils
 from keras.src.callbacks.orbax_checkpoint import OrbaxCheckpoint
+from keras.src.saving import register_keras_serializable
 
 # Import advanced Orbax functionality directly from the LazyModule
 
@@ -416,8 +419,6 @@ class OrbaxCheckpointTest(testing.TestCase, parameterized.TestCase):
     @pytest.mark.requires_trainable_backend
     def test_checkpoint_loading_via_saving_api(self):
         """Test model loading via saving API."""
-        from keras.src import saving
-
         model = self._create_test_model()
         x, y = self._create_dummy_data()
 
@@ -449,97 +450,20 @@ class OrbaxCheckpointTest(testing.TestCase, parameterized.TestCase):
         with self.assertRaises(ValueError):
             saving.load_model(weights_only_dir)
 
+    @parameterized.parameters(
+        {"save_on_background": False},
+        {"save_on_background": True},
+    )
     @pytest.mark.requires_trainable_backend
-    def test_checkpoint_loading_full_state_via_saving_api(self):
-        """Test loading checkpoints with optimizer and metrics state
-        via saving API."""
-        from keras.src import saving
-
-        model = self._create_test_model()
-        model.compile(optimizer="adam", loss="mse", metrics=["mae"])
-        x, y = self._create_dummy_data(num_samples=100)
-
-        checkpoint_dir = os.path.join(
-            self.get_temp_dir(), "test_full_state_loading"
-        )
-        callback = OrbaxCheckpoint(
-            directory=checkpoint_dir, save_freq="epoch", save_weights_only=False
-        )
-
-        model.fit(x, y, epochs=1, callbacks=[callback], verbose=0)
-
-        original_state_tree = model.get_state_tree()
-        loaded_model = saving.load_model(checkpoint_dir)
-        loaded_state_tree = loaded_model.get_state_tree()
-
-        # Helper function to compare state trees
-        def compare_state_components(orig_dict, loaded_dict, component_name):
-            """Compare state components with cross-backend array handling."""
-            for key in orig_dict:
-                if key not in loaded_dict:
-                    # Skip missing metrics keys for non-JAX backends
-                    # (known issue)
-                    if component_name == "metrics_variables" and key != "loss":
-                        continue
-                    self.fail(f"Key {key} missing in loaded {component_name}")
-
-                orig_val, loaded_val = orig_dict[key], loaded_dict[key]
-
-                if isinstance(orig_val, dict):
-                    compare_state_components(
-                        orig_val, loaded_val, f"{component_name}.{key}"
-                    )
-                else:
-                    # Convert to numpy for comparison
-                    def to_numpy(val):
-                        if hasattr(val, "numpy"):
-                            try:
-                                return val.detach().cpu().numpy()  # PyTorch
-                            except AttributeError:
-                                return val.numpy()  # TensorFlow
-                        return val  # JAX array or numpy
-
-                    self.assertAllClose(
-                        to_numpy(orig_val),
-                        to_numpy(loaded_val),
-                        msg=f"Mismatch in {component_name}.{key}",
-                    )
-
-        # Compare all state components
-        for component in [
-            "trainable_variables",
-            "non_trainable_variables",
-            "optimizer_variables",
-        ]:
-            compare_state_components(
-                original_state_tree[component],
-                loaded_state_tree[component],
-                component,
-            )
-
-        # Compare metrics (with backend-specific handling)
-        if "metrics_variables" in original_state_tree:
-            compare_state_components(
-                original_state_tree["metrics_variables"],
-                loaded_state_tree["metrics_variables"],
-                "metrics_variables",
-            )
-
-    @pytest.mark.requires_trainable_backend
-    @pytest.mark.requires_trainable_backend
-    def test_comprehensive_model_state_restoration(self):
+    def test_comprehensive_model_state_restoration(self, save_on_background):
         """Test comprehensive model state restoration with exact weight
         matching.
 
-        Tests async saving, exact weight matching, and complete state
+        Tests sync/async saving, exact weight matching, and complete state
         restoration including trainable/non-trainable variables, optimizer
         state, and custom layers.
         """
-        import keras
-        from keras.src import saving
-        from keras.src.saving import register_keras_serializable
-
-        keras.utils.set_random_seed(42)
+        utils.set_random_seed(42)
 
         # Create model with custom layer having non-trainable variables
         @register_keras_serializable(package="test")
@@ -568,10 +492,17 @@ class OrbaxCheckpointTest(testing.TestCase, parameterized.TestCase):
         model.compile(optimizer="adam", loss="mse", metrics=["mae"])
 
         x, y = self._create_dummy_data(num_samples=100)
-        checkpoint_dir = os.path.join(self.get_temp_dir(), "test_comprehensive")
+        checkpoint_dir = os.path.join(
+            self.get_temp_dir(),
+            f"test_comprehensive_{save_on_background}_{id(self)}",
+        )
 
-        # Test async saving with exact weight matching
-        callback = OrbaxCheckpoint(directory=checkpoint_dir, save_freq="epoch")
+        # Test saving with exact weight matching
+        callback = OrbaxCheckpoint(
+            directory=checkpoint_dir,
+            save_freq="epoch",
+            save_on_background=save_on_background,
+        )
         model.fit(x, y, epochs=2, verbose=0, callbacks=[callback])
 
         # Verify exact weight matching functionality
@@ -594,90 +525,8 @@ class OrbaxCheckpointTest(testing.TestCase, parameterized.TestCase):
         ):
             self.assertAllClose(saved, loaded, msg=f"Weight {i} mismatch")
 
-        # State tree verification (JAX only due to backend limitations)
-        if keras.backend.backend() == "jax":
-            original_state = model.get_state_tree()
-            loaded_state = loaded_model.get_state_tree()
-
-            # Helper function to compare nested state dictionaries
-            def compare_state_dict(orig_dict, loaded_dict, path=""):
-                """Recursively compare nested dictionaries containing arrays."""
-                for key in orig_dict:
-                    if key not in loaded_dict:
-                        continue  # Skip missing keys
-
-                    orig_val, loaded_val = orig_dict[key], loaded_dict[key]
-                    current_path = f"{path}.{key}" if path else key
-
-                    if isinstance(orig_val, dict):
-                        compare_state_dict(orig_val, loaded_val, current_path)
-                    else:
-                        # Compare array values
-                        self.assertAllClose(
-                            orig_val,
-                            loaded_val,
-                            msg=f"Mismatch in {current_path}",
-                        )
-
-            # Compare optimizer variables with nested structure handling
-            compare_state_dict(
-                original_state["optimizer_variables"],
-                loaded_state["optimizer_variables"],
-                "optimizer_variables",
-            )
-
-    @pytest.mark.requires_trainable_backend
-    def test_exact_weight_matching_with_sync_save(self):
-        """Test exact weight matching using synchronous vs asynchronous
-        saving."""
-        import keras
-        from keras.src import saving
-
-        keras.utils.set_random_seed(42)
-
-        model = self._create_test_model()
-        model.compile(optimizer="adam", loss="mse")
-        x, y = self._create_dummy_data(num_samples=50)
-
-        # Test 1: Synchronous saving (exact precision expected)
-        sync_dir = os.path.join(self.get_temp_dir(), "test_sync_exact")
-        sync_callback = OrbaxCheckpoint(
-            directory=sync_dir, save_freq="epoch", save_on_background=False
-        )
-        model.fit(x, y, epochs=1, verbose=0, callbacks=[sync_callback])
-
-        sync_saved_weights = model.get_weights()
-        sync_loaded_model = saving.load_model(sync_dir)
-        sync_loaded_weights = sync_loaded_model.get_weights()
-
-        # Synchronous should have exact matches
+        # Verify optimizer variables
         for i, (saved, loaded) in enumerate(
-            zip(sync_saved_weights, sync_loaded_weights)
+            zip(model.optimizer.variables, loaded_model.optimizer.variables)
         ):
-            self.assertAllClose(
-                saved, loaded, msg=f"Sync exact mismatch: weight {i}"
-            )
-
-        # Test 2: Asynchronous saving (should also work with final
-        # checkpoint strategy)
-        async_dir = os.path.join(self.get_temp_dir(), "test_async_exact")
-        async_callback = OrbaxCheckpoint(
-            directory=async_dir, save_freq="epoch", save_on_background=True
-        )
-        model.fit(x, y, epochs=1, verbose=0, callbacks=[async_callback])
-
-        async_saved_weights = model.get_weights()
-        async_loaded_model = saving.load_model(async_dir)
-        async_loaded_weights = async_loaded_model.get_weights()
-
-        # Async should also match due to final checkpoint strategy
-        for i, (saved, loaded) in enumerate(
-            zip(async_saved_weights, async_loaded_weights)
-        ):
-            self.assertAllClose(
-                saved, loaded, msg=f"Async exact mismatch: weight {i}"
-            )
-
-        # Verify both models are compiled
-        self.assertTrue(sync_loaded_model.compiled)
-        self.assertTrue(async_loaded_model.compiled)
+            self.assertAllClose(saved, loaded, msg=f"Weight {i} mismatch")
--- a/keras/src/saving/saving_api.py
+++ b/keras/src/saving/saving_api.py
@@ -203,6 +203,7 @@ def load_model(filepath, custom_objects=None, compile=True, safe_mode=True):
             filepath,
             custom_objects=custom_objects,
             compile=compile,
+            safe_mode=safe_mode,
         )
 
     elif str(filepath).endswith(".keras"):
@@ -350,7 +351,7 @@ def load_weights(model, filepath, skip_mismatch=False, **kwargs):
 
 
 def _load_model_from_orbax_checkpoint(
-    filepath, custom_objects=None, compile=True
+    filepath, custom_objects=None, compile=True, safe_mode=True
 ):
     """Load a model from an Orbax checkpoint directory."""
 
@@ -381,7 +382,7 @@ def _load_model_from_orbax_checkpoint(
         composite_state["model_config"],
         custom_objects=custom_objects,
         compile=compile,
-        safe_mode=True,
+        safe_mode=safe_mode,
     )
 
     # Prepare state tree with only variable keys for set_state_tree


keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -172,6 +176,14 @@ def __init__(
             preservation_policy=preservation_policy,
         )
 
+    def __del__(self):
+        """Destructor to ensure checkpointer cleanup on garbage collection."""
+        try:
+            if hasattr(self, "checkpointer") and self.checkpointer is not None:
+                self.checkpointer.close()
+        except Exception:
+            pass  # Ignore cleanup errors during destruction
```
hertschuh: Any particular reason to add this? Do you know when it's needed?

The issue with this pattern is that it can lead to deadlocks, which is worse than not closing the resource.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -238,23 +250,79 @@ def _should_save_on_batch(self, batch):
...
         state_tree = _get_state_tree(self.model)
 
         # Save the nested state structures directly (preserving layer
         # names and structure)
-        composite_state = state_tree
+        if self.save_weights_only:
+            composite_state = {
+                "trainable_variables": state_tree["trainable_variables"],
+            }
+            if "non_trainable_variables" in state_tree:
```
hertschuh: No need for an `if` here: `"non_trainable_variables"` is guaranteed to be in the tree.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -238,23 +250,79 @@ def _should_save_on_batch(self, batch):
...
+            # Include compile config if compiled
+            if self.model.compiled:
+                composite_state["compile_config"] = (
+                    self.model.get_compile_config()
+                )
+                from keras.src import optimizers
+
+                composite_state["optimizer_config"] = optimizers.serialize(
+                    self.model.optimizer
+                )
```
hertschuh: The optimizer is already part of `get_compile_config()`.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -238,23 +250,79 @@ def _should_save_on_batch(self, batch):
...
+                def _convert_to_numpy(obj):
+                    """Convert arrays to numpy format for exact matching."""
+                    if hasattr(obj, "shape"):  # Array-like objects
+                        return np.asarray(obj)
+                    return obj
+
+                # Use tree.map_structure for efficient deep conversion
+                state_for_exact_matching = tree.map_structure(
+                    _convert_to_numpy, composite_state
+                )
```
hertschuh: `_get_state_tree` already does that. Why re-do it here?

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -238,23 +250,79 @@ def _should_save_on_batch(self, batch):
...
+                if force_sync is not None
+                else not self.save_on_background
+            )
+
+            if use_sync:
+                # Synchronous save
                 self.checkpointer.save_pytree(step, composite_state)
+            else:
+                # Async save
+                self.checkpointer.save_pytree_async(step, composite_state)
```
hertschuh: Can you rebase with master? You'll need to add the future to the list of futures here.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -238,23 +250,79 @@ def _should_save_on_batch(self, batch):
...
+
+            # Store the exact state that should match what's saved to disk
+            self._last_saved_state = state_for_exact_matching
```
hertschuh: What's going to happen is one of two things:
1. It's going to use twice as much HBM as needed
2. It's going to be deallocated and invalid because we donate the buffers

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -306,11 +374,51 @@ def on_epoch_end(self, epoch, logs=None):
 
         if should_save:
             # Use epoch number as the step for Orbax save
-            # Keras has already made the save decision - Checkpointer will
-            # save unconditionally
             self._save_checkpoint(step=epoch, logs=logs)
 
     def on_train_end(self, logs=None):
+        # Set flag to indicate training is ending (for any remaining callbacks)
+        self._training_ending = True
```
hertschuh: Unused, remove.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -306,11 +374,51 @@ def on_epoch_end(self, epoch, logs=None):
...
+
+        # For exact weight matching, save a final checkpoint that will be
+        # loaded. Ensure this checkpoint will be preserved and found.
+        if self.model:
```
hertschuh: `self.model` cannot be None, and if it is, we definetly want an exception to be raised

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -306,11 +374,51 @@ def on_epoch_end(self, epoch, logs=None):
...
+                    state_tree = _get_state_tree(self.model)
+
+                    if backend.backend() == "jax":
+                        import copy
+
+                        self._final_saved_state = copy.deepcopy(state_tree)
+                    else:
+                        import copy
+
+                        self._final_saved_state = copy.deepcopy(state_tree)
```
hertschuh: ???

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -306,11 +374,51 @@ def on_epoch_end(self, epoch, logs=None):
...
+                # Fall back to current saved state
+                if (
+                    hasattr(self, "_last_saved_state")
+                    and self._last_saved_state is not None
+                ):
+                    self._final_saved_state = self._last_saved_state
+                else:
+                    from keras.src.callbacks.orbax_checkpoint import (
+                        _get_state_tree,
+                    )
```
hertschuh: ???

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -306,11 +374,51 @@ def on_epoch_end(self, epoch, logs=None):
...
+                # Use a step preserved by the max_to_keep policy
+                final_step = max(getattr(self, "_current_epoch", 0), 999999)
+
+                self._save_checkpoint(
+                    step=final_step, logs=logs, force_sync=True
+                )
+                self._final_saved_state = self._last_saved_state
+            except Exception as e:
+                print(f"Warning: Could not save final checkpoint: {e}")
+                # Fall back to current saved state
```
hertschuh: Why would this happen?

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -306,11 +374,51 @@ def on_epoch_end(self, epoch, logs=None):
...
-            # Keras has already made the save decision - Checkpointer will
-            # save unconditionally
             self._save_checkpoint(step=epoch, logs=logs)
 
     def on_train_end(self, logs=None):
+        # Set flag to indicate training is ending (for any remaining callbacks)
+        self._training_ending = True
+
+        # For exact weight matching, save a final checkpoint that will be
+        # loaded. Ensure this checkpoint will be preserved and found.
```
hertschuh: All the changes in `on_train_end` should be reverted. `on_epoch_end` was just called and would have done the last checkpoint already.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -330,3 +438,58 @@ def wait_until_finished(self):
...
+        """
+        # If training has ended, use the final saved state for exact matching
+        if (
+            hasattr(self, "_final_saved_state")
+            and self._final_saved_state is not None
+        ):
+            return self._extract_weights_from_state(self._final_saved_state)
+
+        # During training, use the last saved state
+        return self._extract_weights_from_state(self._last_saved_state)
```
hertschuh: What is all this code? Why is it here? It's never called.

keras/src/saving/saving_api.py
```diff
@@ -337,3 +349,136 @@ def load_weights(model, filepath, skip_mismatch=False, **kwargs):
...
+    )
+
+    # Build model if needed using first layer's batch_shape
+    if not model.built and model.layers:
+        first_layer = model.layers[0]
+        if hasattr(first_layer, "batch_shape") and first_layer.batch_shape:
+            input_shape = first_layer.batch_shape
+            if input_shape[0] is None:  # Replace None batch with actual shape
+                input_shape = (None,) + input_shape[1:]
+            model.build(input_shape)
```
hertschuh: You don't need this. The config has a `build_config` in it which will build each layer.

keras/src/saving/saving_api.py
```diff
@@ -337,3 +349,136 @@ def load_weights(model, filepath, skip_mismatch=False, **kwargs):
...
+        compile_config = composite_state.get("compile_config")
+        if compile_config:
+            model.compile_from_config(compile_config)
+        elif "optimizer_config" in composite_state:
+            # Fallback for backwards compatibility
+            optimizer = optimizers.deserialize(
+                composite_state["optimizer_config"],
+                custom_objects=custom_objects,
+            )
+            model.compile(optimizer=optimizer)
```
hertschuh: Remove this block.

keras/src/saving/saving_api.py
```diff
@@ -337,3 +349,136 @@ def load_weights(model, filepath, skip_mismatch=False, **kwargs):
...
+    # Prepare state tree efficiently
+    state_tree = {
+        key: composite_state.get(key, {})
+        for key in [
+            "trainable_variables",
+            "non_trainable_variables",
+            "optimizer_variables",
+            "metrics_variables",
+        ]
+    }
```
hertschuh, line 399: Why do you need to do this?

keras/src/saving/saving_api.py
```diff
@@ -337,3 +349,136 @@ def load_weights(model, filepath, skip_mismatch=False, **kwargs):
...
+
+    # Convert to numpy for non-JAX backends
+    if backend.backend() != "jax":
+
+        def convert_to_numpy(obj):
+            if hasattr(obj, "__array__"):
+                return np.asarray(obj)
+            return obj
+
+        state_tree = tree.map_structure(convert_to_numpy, state_tree)
```
hertschuh: Why do you need to do this? Is it to reload a JAX saved checkpoint with a different backend?

keras/src/saving/saving_api.py
```diff
@@ -337,3 +349,136 @@ def load_weights(model, filepath, skip_mismatch=False, **kwargs):
...
+                else:
+                    # Fallback: use same shape as input
+                    dummy_target = np.zeros_like(dummy_input)
+
+                # Initialize metrics through evaluation
+                model.evaluate(dummy_input, dummy_target, verbose=0)
+
+        except Exception:
+            # Gracefully handle any initialization failures
+            pass
```
hertschuh: I don't understand any of this? Why would this be needed?

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: Removed

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: removed

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: 1. Synchronous saves: Store the state since there's no buffer donation issue
2. Async saves: Don't store the state to avoid either doubling HBM usage or dealing with invalid donated buffers
3. get_weights() fallback: When no saved state is available (due to async saves), fall back to current model state
This resolves both potential issues you mentioned while maintaining functionality.

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: I've removed the redundant if/else block since both branches were doing exactly the same thing

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: I've removed the unnecessary import

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: Removed the exception

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: You're correct that on_epoch_end handles the last checkpoint save. However, for exact weight matching to work properly with async saves, we need the final checkpoint to be synchronous so we can capture and store its exact state.
With async saves, _last_saved_state is set to None (to avoid memory issues), which means get_last_saved_weights() falls back to the current model state. But the current model state may have changed slightly since the checkpoint was actually written to disk, breaking exact weight matching for tests.
The solution is to either:
1. Force the final epoch save to be synchronous, or
2. Save one final synchronous checkpoint in on_train_end specifically for exact weight matching

keras/src/saving/saving_api.py
amitsrivastava78: Done

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: Actually, these methods ARE being called! The test file has three calls to get_last_saved_weights():
So this is not dead code - it's actually being used by the tests for exact weight matching. The methods should stay.

keras/src/saving/saving_api.py
amitsrivastava78: Done

keras/src/saving/saving_api.py
amitsrivastava78: This filtering is necessary because model.set_state_tree() is strict about the keys it accepts. It only handles variable-related keys (trainable_variables, non_trainable_variables, optimizer_variables, metrics_variables) and throws a ValueError for unknown keys.
The composite_state also contains config keys like model_config and compile_config which are handled separately (for model reconstruction and compilation). If we pass the full composite_state directly to set_state_tree(), it fails with 'Unknown variable name: compile_config'.
So we need to filter out the config keys and only pass the variable keys that set_state_tree() expects.

keras/src/saving/saving_api.py
amitsrivastava78: Removed

keras/src/saving/saving_api.py
amitsrivastava78: Removed

amitsrivastava78:
> The whole asset saving / loading part of it is missing.

yes will raise a separate PR for that

keras/src/callbacks/orbax_checkpoint.py
hertschuh: Ok, this is test code. So it is dead code in the Pip wheel. Please move to the test file.

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: looks like there is expected behaviour with orbax that multiply async save is going then only one step is saved causing the issue so checked with orbax team and they suggested to use save_decision_policy=ocp.training.save_decision_policies.FixedIntervalPolicy which saves every step

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -268,15 +264,19 @@ def test_save_on_background_sync(self):
             save_on_background=False,  # Synchronous saving
         )
 
-        # Train and ensure it completes (synchronous save should not block)
-        model.fit(x, y, epochs=2, callbacks=[callback], verbose=0)
-        callback.wait_until_finished()
+        try:
+            # Train and ensure it completes (synchronous save should not block)
+            model.fit(x, y, epochs=2, callbacks=[callback], verbose=0)
+            callback.wait_until_finished()
```
hertschuh: You should not need that. `model.fit` should call `on_train_end` which should wait. If it doesn't, this unit test should fail.

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -268,15 +264,19 @@ def test_save_on_background_sync(self):
...
-            len(checkpoint_files), 0, "Should have checkpoint files"
-        )
+            # Check that checkpoints were created
+            checkpoint_files = os.listdir(checkpoint_dir)
+            self.assertGreater(
+                len(checkpoint_files), 0, "Should have checkpoint files"
+            )
+        finally:
+            # Ensure proper cleanup to prevent file descriptor leaks
+            callback.on_train_end()
```
hertschuh: So this is just so that if the unit test fails and the assert line 274 is called some cleanup happens?

The truth is that `on_train_end` was already called as part of `model.fit`. So this isn't needed. If it didn't work in `model.fit`, it's not going to work here.

And just in general, you have to assume that unit tests succeed by default. I appreciate that you want some cleanup when the test fails, but then we would have `try / finally` blocks like that all over the unit tests, which would make them longer and harder to read.

So remove the `try / finally`.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -246,15 +255,41 @@ def _save_checkpoint(self, step, logs=None):
...
                 self.checkpointer.save_pytree(step, composite_state)
+            else:
+                # Async save
+                future = self.checkpointer.save_pytree_async(
+                    step, composite_state
+                )
+                self._async_futures.append(future)
+
+            # Store the checkpoint path for reference
+            self._last_checkpoint_path = os.path.join(self.directory, str(step))
```
hertschuh: This is never used. Remove.

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -400,15 +401,19 @@ def test_save_on_background_async(self):
...
-            len(checkpoint_files), 0, "Should have checkpoint files"
-        )
+            # Check that checkpoint was created
+            checkpoint_files = os.listdir(checkpoint_dir)
+            self.assertGreater(
+                len(checkpoint_files), 0, "Should have checkpoint files"
+            )
+        finally:
+            # Ensure proper cleanup to prevent file descriptor leaks
+            callback.on_train_end()
```
hertschuh: Remove `try / finally`.

keras/src/saving/saving_api.py
```diff
@@ -337,3 +347,59 @@ def load_weights(model, filepath, skip_mismatch=False, **kwargs):
...
+    if "model_config" not in composite_state:
+        raise ValueError(
+            "Checkpoint does not contain model configuration. "
+            "This checkpoint may have been saved with save_weights_only=True."
+        )
+
+    # Create and build model from config
+    model = models.Model.from_config(
+        composite_state["model_config"], custom_objects=custom_objects
+    )
```
hertschuh: Related to my other comment, you should use `saving_lib._model_from_config` instead to handle shared objects and `compile_config`.

https://github.com/keras-team/keras/blob/master/keras/src/saving/saving_lib.py#L422

It'll take the config as a JSON string, which is what you got on the saving side.

keras/src/saving/saving_api.py
```diff
@@ -337,3 +347,59 @@ def load_weights(model, filepath, skip_mismatch=False, **kwargs):
...
+
+    # Handle compilation efficiently
+    if compile:
+        compile_config = composite_state.get("compile_config")
+        if compile_config:
+            model.compile_from_config(compile_config)
```
hertschuh: This is no longer needed, the `compile_config` is part of the overall config.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -246,15 +255,41 @@ def _save_checkpoint(self, step, logs=None):
...
+            }
+        else:
+            composite_state = state_tree
+            # Include model configuration for full model restoration
+            composite_state["model_config"] = self.model.get_config()
+            # Include compile config if compiled
+            if self.model.compiled:
+                composite_state["compile_config"] = (
+                    self.model.get_compile_config()
+                )
```
hertschuh: So I realized that we should use some utilities that already exist for consistency and to support object sharing. Instead of doing this, use `saving_lib._serialize_model_as_json(model)`.
https://github.com/keras-team/keras/blob/master/keras/src/saving/saving_lib.py#L145

- it will contain the `compile_config` already.
- it will return the JSON as a string and that's fine, that's consistent with the other formats.

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -538,3 +543,279 @@ def test_distributed_checkpoint_functionality(self):
...
+        callback.wait_until_finished()
+
+        original_weights = model.get_weights()
+        loaded_model = saving.load_model(checkpoint_dir)
+        loaded_weights = loaded_model.get_weights()
+
+        # Compare weights and verify compilation
+        self.assertEqual(len(original_weights), len(loaded_weights))
+        for orig, loaded in zip(original_weights, loaded_weights):
+            np.testing.assert_array_almost_equal(orig, loaded)
```
hertschuh: Use `self.assertAllClose`, it takes care of the conversion etc.

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -9,14 +9,8 @@
...
 
 # Import advanced Orbax functionality directly from the LazyModule
-Checkpointer = ocp.training.Checkpointer
-save_pytree = ocp.save_pytree
-load_pytree = ocp.load_pytree
-preservation_policies = ocp.training.preservation_policies
-save_decision_policies = ocp.training.save_decision_policies
 
 
 class OrbaxCheckpointTest(testing.TestCase):
```
hertschuh: I still have a hard time following the structure of the tests. It looks like many of them verify the same thing in the end. I think the test file would be much shorter if you used parameterized tests to verify the different options.

For instance, all these tests verify the exact same thing and could be a parameterized tests:
- frequency is batch
- frequency is epoch
- save best only
- save with threshold
- async vs sync

keras/src/callbacks/orbax_checkpoint_test.py
hertschuh: You don't need this one, all the other tests verify that there is a checkpoint which implies there is a folder.

keras/src/callbacks/orbax_checkpoint_test.py

keras/src/callbacks/orbax_checkpoint_test.py

keras/src/callbacks/orbax_checkpoint_test.py

keras/src/callbacks/orbax_checkpoint.py

keras/src/saving/saving_api.py
amitsrivastava78: Done

keras/src/saving/saving_api.py
amitsrivastava78: Done

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint_test.py

keras/src/callbacks/orbax_checkpoint_test.py
amitsrivastava78: 1. Created test_checkpoint_saving_basic - Parameterized test combining batch and epoch frequency tests
2. Created test_save_best_only - Parameterized test for different modes (min/max)
3. Created test_async_vs_sync_saving - Parameterized test for synchronous vs asynchronous saving
4. Kept individual tests for:
    * test_directory_creation - Simple directory creation test
    * test_max_to_keep - Max checkpoints test
    * test_invalid_save_freq - Error handling test
    * test_initial_value_threshold - Threshold test
    * test_load_weights_from_orbax_checkpoint - Consolidated loading test
    * test_checkpoint_loading_comprehensive - Parameterized for sync/async
    * test_checkpoint_loading_via_saving_api - Simplified saving API test
    * Complex state restoration tests remain as they test specific functionality

keras/src/callbacks/orbax_checkpoint_test.py

amitsrivastava78:
> I realized that the config saving and loading should reuse some utilities that already exist and take care of some non-trivial stuff.

yes, the implementation now is perfectly aligned with the recommendation to reuse existing utilities

keras/src/saving/saving_api.py
```diff
@@ -195,6 +196,15 @@ def load_model(filepath, custom_objects=None, compile=True, safe_mode=True):
...
             safe_mode=safe_mode,
         )
+
+    # Check for Orbax checkpoint directory using utility function
+    if is_orbax_checkpoint(filepath):
+        return _load_model_from_orbax_checkpoint(
+            filepath,
+            custom_objects=custom_objects,
+            compile=compile,
+        )
```
hertschuh, line 207: Also pass `safe_mode=safe_mode`.

keras/src/saving/saving_api.py
```diff
@@ -337,3 +347,56 @@ def load_weights(model, filepath, skip_mismatch=False, **kwargs):
             "`.weights.h5` files, legacy H5 format files "
             "(`.h5` extension), or Orbax checkpoints."
         )
+
+
+def _load_model_from_orbax_checkpoint(
+    filepath, custom_objects=None, compile=True
+):
```
hertschuh, line 355: Add a `safe_mode=True` argument.

keras/src/saving/saving_api.py
```diff
@@ -337,3 +347,56 @@ def load_weights(model, filepath, skip_mismatch=False, **kwargs):
...
+            "This checkpoint may have been saved with save_weights_only=True."
+        )
+
+    # Create and build model from config using saving_lib helper
+    # This properly handles shared objects and compile_config
+    model = saving_lib._model_from_config(
+        composite_state["model_config"],
+        custom_objects=custom_objects,
+        compile=compile,
+        safe_mode=True,
```
hertschuh: `safe_mode=safe_mode`.

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -538,3 +412,272 @@ def test_distributed_checkpoint_functionality(self):
...
+        # Compare metrics (with backend-specific handling)
+        if "metrics_variables" in original_state_tree:
+            compare_state_components(
+                original_state_tree["metrics_variables"],
+                loaded_state_tree["metrics_variables"],
+                "metrics_variables",
+            )
+
+    @pytest.mark.requires_trainable_backend
+    @pytest.mark.requires_trainable_backend
```
hertschuh: ???  Remove one.

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -538,3 +412,272 @@ def test_distributed_checkpoint_functionality(self):
...
+        """Test comprehensive model state restoration with exact weight
+        matching.
+
+        Tests async saving, exact weight matching, and complete state
+        restoration including trainable/non-trainable variables, optimizer
+        state, and custom layers.
+        """
+        import keras
+        from keras.src import saving
+        from keras.src.saving import register_keras_serializable
```
hertschuh: Move to top.

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -538,3 +412,272 @@ def test_distributed_checkpoint_functionality(self):
...
+    @pytest.mark.requires_trainable_backend
+    def test_comprehensive_model_state_restoration(self):
+        """Test comprehensive model state restoration with exact weight
+        matching.
+
+        Tests async saving, exact weight matching, and complete state
+        restoration including trainable/non-trainable variables, optimizer
+        state, and custom layers.
+        """
+        import keras
```
hertschuh: Don't `import keras` in unit tests, import the specific bits you need.

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -538,3 +412,272 @@ def test_distributed_checkpoint_functionality(self):
...
+                            loaded_val,
+                            msg=f"Mismatch in {current_path}",
+                        )
+
+            # Compare optimizer variables with nested structure handling
+            compare_state_dict(
+                original_state["optimizer_variables"],
+                loaded_state["optimizer_variables"],
+                "optimizer_variables",
+            )
```
hertschuh: You don't need all this. For one thing, we have `keras.tree` that does this.

But if you're just trying to verify the optimizer variables, you can just do this:

```python
        for i, (saved, loaded) in enumerate(
            zip(model.optimizer.variables, loaded_model.optimizer.variables)
        ):
            self.assertAllClose(saved, loaded, msg=f"Weight {i} mismatch")
```

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -538,3 +412,272 @@ def test_distributed_checkpoint_functionality(self):
...
+
+    @pytest.mark.requires_trainable_backend
+    def test_exact_weight_matching_with_sync_save(self):
+        """Test exact weight matching using synchronous vs asynchronous
+        saving."""
+        import keras
+        from keras.src import saving
```
hertschuh: Same comment about moving to top and not importing keras.

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -538,3 +412,272 @@ def test_distributed_checkpoint_functionality(self):
...
+            # Compare optimizer variables with nested structure handling
+            compare_state_dict(
+                original_state["optimizer_variables"],
+                loaded_state["optimizer_variables"],
+                "optimizer_variables",
+            )
+
+    @pytest.mark.requires_trainable_backend
+    def test_exact_weight_matching_with_sync_save(self):
+        """Test exact weight matching using synchronous vs asynchronous
```
hertschuh: If you took the test right above and parametrized it to do sync=True and sync=False, you wouldn't need this test at all. The one above does everything and more (except it's async).

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -538,3 +412,272 @@ def test_distributed_checkpoint_functionality(self):
...
+                                return val.detach().cpu().numpy()  # PyTorch
+                            except AttributeError:
+                                return val.numpy()  # TensorFlow
+                        return val  # JAX array or numpy
+
+                    self.assertAllClose(
+                        to_numpy(orig_val),
+                        to_numpy(loaded_val),
+                        msg=f"Mismatch in {component_name}.{key}",
+                    )
```
hertschuh: You don't need all this. For one thing, we have `keras.tree`. But you don't even need this.

Also `assertAllClose` already does the numpy conversion and handles torch correctly.

You can just compare the model weights in a for loop like you did line 592. And for the optimizer, I explained how to do it line 627.

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -538,3 +412,272 @@ def test_distributed_checkpoint_functionality(self):
...
+            save_weights_only=True,
+        )
+        model.fit(x, y, epochs=1, callbacks=[weights_callback], verbose=0)
+
+        with self.assertRaises(ValueError):
+            saving.load_model(weights_only_dir)
+
+    @pytest.mark.requires_trainable_backend
+    def test_checkpoint_loading_full_state_via_saving_api(self):
+        """Test loading checkpoints with optimizer and metrics state
```
hertschuh: What's the difference between this one and `test_comprehensive_model_state_restoration` line 530. It looks to me like `test_comprehensive_model_state_restoration` does everything here plus some more. Do we need both?

keras/src/saving/saving_api.py
amitsrivastava78: Done

keras/src/saving/saving_api.py
amitsrivastava78: Done

keras/src/callbacks/orbax_checkpoint_test.py

keras/src/callbacks/orbax_checkpoint_test.py

keras/src/callbacks/orbax_checkpoint_test.py

keras/src/saving/saving_api.py
amitsrivastava78: Done

keras/src/callbacks/orbax_checkpoint_test.py

keras/src/callbacks/orbax_checkpoint_test.py

keras/src/callbacks/orbax_checkpoint_test.py

keras/src/callbacks/orbax_checkpoint_test.py

keras/src/callbacks/orbax_checkpoint_test.py
amitsrivastava78: removed test_checkpoint_loading_full_state_via_saving_api

PR #22003, opened 2026-01-13
Add uninitialized() function to support prim::Uninitialized in OpenVI…

Vbhatt03:
dd `uninitialized()` function to support `prim::Uninitialized` operation in OpenVINO backend.

Since OpenVINO doesn't support truly uninitialized data, the function returns a zeros tensor (following the same approach used in OpenVINO's PyTorch Frontend).

Fixes related to: openvinotoolkit/openvino#29727

### DIFF
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -1083,6 +1083,16 @@ def empty_like(x, dtype=None):
     return zeros_like(x, dtype=dtype)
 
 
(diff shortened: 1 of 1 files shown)


hertschuh:
https://github.com/openvinotoolkit/openvino/issues/29727 has nothing to do with Keras. It needs to be addressed in OpenVino. Once it is fixed in OpenVino, then Keras can make use of it in the implementation of `numpy.empty`. It's the other way around.

PR #22007, opened 2026-01-14
Adds Asymmetric INT4 Sub-Channel Quantization Support

JyotinderSingh:
This PR introduces support for INT4 Sub-channel (Grouped) Quantization to Keras.

Previously, Keras `int4` quantization operated in "per-channel" mode (one scale factor per output channel). This PR extends the functionality to support fine-grained quantization by grouping weights (e.g., blocks of 64 or 128 parameters) along the input dimension. This significantly reduces quantization error compared to per-channel approaches.

## Features
1.  The config `Int4QuantizationConfig` class now accepts a `block_size` argument (defaulting to `128`).
    * `block_size=128`: Enables sub-channel quantization with 128-element groups.
    * `block_size=None` or `-1`: Preserves legacy per-channel behavior.
2.  Sub-channel quantization implements asymmetric quantization, introducing a `zero_point` quantization variable alongside the `scale`.
3.  Updates `Dense`, `EinsumDense`, `Embedding`, and `ReversibleEmbedding` to support the new variable layout:
    * `kernel` / `embeddings`: Packed int4 weights.
    * `kernel_scale`: Shape `(n_groups, output_dim)`.
    * `kernel_zero`: New variable for zero-points, shape `(n_groups, output_dim)`.
    * `g_idx`: New variable mapping input indices to group indices.
4.  Adds a specific policy class to handle strings like `"int4/128"` or `"int4/64"`, ensuring configuration is preserved during serialization/deserialization of KerasHub models.

## API

To use grouped quantization, users specify the `block_size` in the `Int4QuantizationConfig`.

```python
import keras
from keras.quantizers import Int4QuantizationConfig

# 1. Define configuration with specific block size.
# block_size=128 is a common default for LLMs
config = Int4QuantizationConfig(block_size=128)

# 2. Apply quantization
# This triggers weight packing, scale calculation, and zero-point calculation.
# Note: model.quantize("int4") triggers the same behaviour, defaulting to
# a block size of 128.
model.quantize(config=config)

# 3. Save the quantized model.
# The policy "int4/128" will be saved with the model.
model.save("grouped_int4_model.keras")
```

To use the legacy per-channel mode (no zero-point overhead):

```python
config = Int4QuantizationConfig(block_size=None)
model.quantize(config=config)
```

## Verification
1. Unit Tests: Added comprehensive tests in quantizers_test.py verifying numerical correctness of abs_max_quantize_grouped_with_zero_point and round-trip accuracy.
2. Layer Tests: Updated Dense, EinsumDense, and Embedding tests to cover:
    * Shape verification for scales and zero-points based on block_size.
    * Serialization/Deserialization (ensuring g_idx and block_size persist).
    * Integration with LoRA (merging grouped quantized weights).
3. Backward Compatibility: Verified that block_size=None produces the exact same variable structure as the previous implementation.

### Configuration

| Parameter | Value |
|-----------|-------|
| GPU | NVIDIA RTX 4070 Ti Super |
| Dataset | WikiText-2 |
| Sequence Length | 128 |
| Evaluation Batches | 50 |
| Calibration Samples | 128 |
| Grid Points | 20 |
| Runs per Model | 1-3 |

### Results
Source: [benchmark.csv](https://github.com/user-attachments/files/24623948/int4_benchmarks_with_zero_point_no_embedding.csv)

| Model | Pre PPL | Post PPL | PPL Δ | Disk Size (Pre→Post) | Disk Δ | GPU Mem (Pre→Post) | Mem Δ | Throughput (Pre→Post) | Throughput Δ | Quant Time |
|-------|---------|----------|-------|----------------------|--------|-------------------|-------|----------------------|--------------|------------|
| Qwen3 1.7B | 37.65 | 45.59 | +21.11% | 6.42→1.88 GB | -70.8% | 6.95→2.07 GB | -70.2% | 182.5→154.5 tok/s | -15.4% | 20.2s |
| Gemma3 1B | 172.45 | 182.12 | +5.60% | 3.73→1.48 GB | -60.2% | 3.88→1.59 GB | -59.1% | 400.3→316.1 tok/s | -21.0% | 16.4s |
| Qwen3 0.6B | 53.71 | 63.87 | +18.91% | 2.23→0.81 GB | -63.7% | 2.61→0.96 GB | -63.2% | 307.5→300.8 tok/s | -2.2% | 17.2s |
| OPT 125M | 77.06 | 84.24 | +9.32% | 0.47→0.19 GB | -58.4% | 0.63→0.37 GB | -41.7% | 506.0→551.9 tok/s | +9.1% | 5.7s |

## Observations

- **Disk size reduction**: 58–71% across models (INT4 quantization from FP32/BF16 weights)
- **GPU memory reduction**: 42–70% during inference
- **Perplexity degradation**: +5.6% (Gemma3 1B) to +21.1% (Qwen3 1.7B)
- **Throughput**: -2% to -21% depending on model; OPT 125M shows +9% average with high variance
- **Quantization time**: Fast quantization (~5.7s for 125M, ~16.4s for 1B, ~20.2s for 1.7B)

int4_benchmark_analysis

### DIFF
--- a/keras/src/layers/core/einsum_dense.py
+++ b/keras/src/layers/core/einsum_dense.py
@@ -309,8 +309,10 @@ class EinsumDense(Layer):
         self._tracker.unlock()
         # Determine the appropriate (unpacked) kernel shape for LoRA.
         if self.quantization_mode == "int4":
-            # For INT4 with GPTQ layout, use original_kernel_shape since
-            # the packed kernel is transposed
+            # INT4 weights are stored in a flattened 2D layout that loses
+            # the original N-dimensional structure required by the einsum
+            # equation. We use `original_kernel_shape`` to ensure LoRA adapters
+            # operate in the correct logical dimension space.
             kernel_shape_for_lora = tuple(self.original_kernel_shape)
         else:
             kernel_shape_for_lora = self.kernel.shape


keras/src/layers/core/reversible_embedding.py
```diff
@@ -272,23 +372,83 @@ def _int4_call(self, inputs, reverse=False):
...
+                # Untied weights with asymmetric grouped quantization
+                # Must dequantize embeddings before matmul for correctness
+                # unpacked_embeddings shape: (output_dim, input_dim)
+                # scale shape: (n_groups, input_dim)
+                # reverse_embeddings_zero shape: (n_groups, input_dim)
+
+                # Compute g_idx to map each output position to its group
+                g_idx = ops.floor_divide(
+                    ops.arange(self.output_dim, dtype="int32"), block_size
+                )
```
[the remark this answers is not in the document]
JyotinderSingh: Fixed.

keras/src/layers/core/reversible_embedding.py
```diff
@@ -180,8 +183,84 @@ def variable_serialization_spec(self):
...
+                self, "reverse_embeddings_zero"
+            ):
+                # reverse_embeddings_zero only exists for sub-channel int4
+                continue
+            elif name == "embeddings_scale" and mode in ("int4", "int8"):
+                store[str(idx)] = merged_kernel_scale
+            else:
+                store[str(idx)] = getattr(self, name)
+            idx += 1
+
```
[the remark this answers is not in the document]
JyotinderSingh: Fixed.

keras/src/quantizers/quantizers.py
```diff
@@ -118,6 +120,190 @@ def abs_max_quantize(
...
+    input_shape = ops.shape(inputs)
+    input_dim = input_shape[0]
+    output_dim = input_shape[1]
+    qmin, qmax = value_range
+
+    # Infer bit-width from quantization range (e.g., [-8, 7] -> 4 bits)
+    num_levels = qmax - qmin + 1
+    bits = int(math.log2(num_levels))
+
+    n_groups = int(math.ceil(int(input_dim) / block_size))
```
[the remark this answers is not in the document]
JyotinderSingh: not required.

keras/src/quantizers/quantizers.py
```diff
@@ -118,6 +120,190 @@ def abs_max_quantize(
...
+        epsilon=epsilon,
+        signed=True,
+    )
+
+    # Transpose results back to (n_groups, output_dim)
+    scale = ops.transpose(scale_t)
+    zero_point = ops.transpose(zero_point_t)
+
+    # Zero-pad rows so input_dim is divisible by block_size
+    pad_size = padded_input_dim - int(input_dim)
```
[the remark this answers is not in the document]
JyotinderSingh: invalid case.

keras/src/layers/core/dense.py
```diff
@@ -630,37 +654,74 @@ def _awq_call(self, inputs, training=False):
...
 
-        # Kernel is stored *packed*: each int8 byte contains two int4 values.
+        # kernel is packed along last axis (output dimension)
+        # Stored shape: [input_dim, ceil(output_dim/2)]
+        packed_cols = (output_dim + 1) // 2
+
         self._kernel = self.add_weight(
             name="kernel",
-            shape=(packed_rows, output_dim),
+            shape=(input_dim, packed_cols),
```
hertschuh, line 676: Does this mean this is a breaking change? I.e. people who had done int4 quantization (with no groups) and try to reload their model won't be able too.

Not that I think we should fix it necessarily, but I'm curious.

keras/src/layers/core/dense.py
```diff
@@ -630,37 +654,74 @@ def _awq_call(self, inputs, training=False):
...
+                name="g_idx",
+                shape=(input_dim,),
+                initializer="zeros",
+                dtype="float32",
+                trainable=False,
+            )
+            self.g_idx.assign(
+                ops.floor_divide(
+                    ops.arange(input_dim, dtype="float32"), block_size
+                )
```
hertschuh: nitpick: You know you can pass a tensor as `initializer` in `add_weight`, so you could pass this directly.

You could also do:
```python
def idx_initializer(shape, dtype):
    return ops.floor_divide(ops.arange(input_dim[0], dtype=dtype), block_size)
```

keras/src/layers/core/dense_test.py
```diff
@@ -1215,3 +1221,213 @@ def test_dense_int4_custom_quantizer(self):
...
+        # Verify block_size is stored
+        self.assertEqual(layer._int4_block_size, block_size)
+
+        # Verify kernel_scale shape
+        if block_size is None or block_size == -1:
+            # Per-channel: one scale per output unit
+            expected_scale_shape = (output_dim,)
+        else:
+            # Sub-channel: (n_groups, out_features)
+            import math
```
hertschuh: Just import at the top

keras/src/layers/core/einsum_dense.py
```diff
@@ -306,16 +309,9 @@ def enable_lora(
...
-            # the LoRA matrices operate in the full-precision space.
-            kernel_shape_for_lora = list(self._kernel.shape)
-            pack_axis = getattr(self, "_int4_pack_axis", 0)
-            orig_len = getattr(self, "_orig_length_along_pack_axis", None)
-            if orig_len is not None:
-                kernel_shape_for_lora[pack_axis] = orig_len
-            kernel_shape_for_lora = tuple(kernel_shape_for_lora)
+            # For INT4 with GPTQ layout, use original_kernel_shape since
+            # the packed kernel is transposed
+            kernel_shape_for_lora = tuple(self.original_kernel_shape)
```
hertschuh: I don't understand why this changed. Is this also a breaking change?

I also don't understand the comment, you say it's transposed but you don't swap the dimensions.

keras/src/layers/core/einsum_dense_test.py
```diff
@@ -1370,3 +1389,415 @@ def test_einsum_dense_int8_custom_quantizer(self):
...
+        self.assertIsNotNone(kernel_scale)
+        # (n_groups, columns) where n_groups is at index 0
+        self.assertEqual(kernel_scale.shape[0], 2)
+
+    def test_int4_multi_reduced_axes_serialization(self):
+        """Test serialization with multi-reduced-axis equations."""
+        if testing.tensorflow_uses_gpu():
+            self.skipTest("Segfault on TF GPU")
+
+        import tempfile
```
hertschuh: Remove, use `get_temp_dir()`. You won't need a `with`, it cleans up automatically.

keras/src/layers/core/einsum_dense_test.py
```diff
@@ -1370,3 +1389,415 @@ def test_einsum_dense_int8_custom_quantizer(self):
...
+
+        import keras
```
hertschuh: Remove. We never `import keras` in unit tests. Instead we import from keras.src the parts that are needed.

Oh, I move it to the top.

keras/src/layers/core/embedding_test.py
```diff
@@ -654,3 +663,207 @@ def test_embedding_int8_custom_quantizer(self):
...
+
+    @parameterized.named_parameters(
+        ("grouped_block_64", 64),
+        ("grouped_block_128", 128),
+        ("per_channel_none", None),
+        ("per_channel_neg1", -1),
+    )
+    def test_int4_quantization_block_size(self, block_size):
+        """Test int4 quantization with different block_size configurations."""
+        import math
```
hertschuh: Move to top

keras/src/layers/core/embedding_test.py
```diff
@@ -654,3 +663,207 @@ def test_embedding_int8_custom_quantizer(self):
...
+
+        x = np.random.randint(0, input_dim, size=(4, 8))
+
+        # Should run without error
+        y = layer(x)
+        self.assertEqual(y.shape, (4, 8, output_dim))
+
+    def test_int4_grouped_vs_perchannel_scale_shapes(self):
+        """Test that grouped and per-channel have different scale shapes."""
+        import math
```
hertschuh: Move to top.

keras/src/layers/core/reversible_embedding.py
```diff
@@ -180,8 +183,88 @@ def variable_serialization_spec(self):
                 variable_spec.append("reverse_embeddings")
                 if mode in ("int4", "int8"):
                     variable_spec.append("reverse_embeddings_scale")
+                if mode == "int4":
+                    # reverse_embeddings_zero only exists for sub-channel
+                    variable_spec.append("reverse_embeddings_zero")
         return _spec
 
+    def save_own_variables(self, store):
```
hertschuh: How was it working before? Or was there a bug?

Also, can you call "super()" first and only deal with the reverse embedding here?

keras/src/layers/core/reversible_embedding.py
```diff
@@ -180,8 +183,88 @@ def variable_serialization_spec(self):
...
+                continue
+            elif name == "embeddings_scale" and mode in ("int4", "int8"):
+                # For int4/int8, the merged LoRA scale (if any) comes from
+                # `_get_embeddings_with_merged_lora()`
+                store[str(idx)] = merged_embeddings_scale
+            else:
+                store[str(idx)] = getattr(self, name)
+            idx += 1
+
+    def load_own_variables(self, store):
```
hertschuh: How was it working before? Or was there a bug?

Also, can you call "super()" first and only deal with the reverse embedding here?

keras/src/layers/core/reversible_embedding_test.py
```diff
@@ -271,3 +274,241 @@ def test_masking(self):
...
+
+    @parameterized.named_parameters(
+        named_product(
+            block_size=(64, 128, None, -1),
+            tie_weights=(True, False),
+        )
+    )
+    def test_int4_quantization_block_size(self, block_size, tie_weights):
+        """Test int4 quantization with different block_size configurations."""
+        import math
```
hertschuh: Move to top.

keras/src/layers/core/reversible_embedding_test.py
```diff
@@ -271,3 +274,241 @@ def test_masking(self):
...
+        ("tie_grouped", True, 64),
+        ("tie_perchannel", True, None),
+        ("untie_grouped", False, 64),
+        ("untie_perchannel", False, None),
+    )
+    def test_int4_grouped_vs_perchannel_scale_shapes(
+        self, tie_weights, block_size
+    ):
+        """Test that grouped and per-channel have different scale shapes."""
+        import math
```
hertschuh: Move to top.

keras/src/dtype_policies/dtype_policy.py
```diff
@@ -288,6 +288,79 @@ def get_config(self):
...
+                "Invalid block_size in mode. Supported values are "
+                "-1 (per-channel) or a positive integer (sub-channel), "
+                f"but got {block_size} from '{mode}'."
+            )
+
+        base_mode = parts[0]
+        super().__init__(
+            mode=base_mode,
+            source_name=source_name,
+        )
```
hertschuh, line 345: Not to address in this PR, but I find it weird that:
- the mode that you pass in `__init__` is the full mode
- but then the mode property returns the base mode
- the mode from get_config is the full mode

It forces you to override `get_config` always to reconstruct the full mode.

Instead:
- the base class should take the full mode
- the mode property should be the full mode
- we can add a base_mode property
- the get_config from the super class works for the base class

Can you address this in a separate PR?

keras/src/layers/core/reversible_embedding_test.py
JyotinderSingh: Done.

keras/src/layers/core/embedding_test.py
JyotinderSingh: Done.

keras/src/layers/core/embedding_test.py
JyotinderSingh: Done.

keras/src/layers/core/einsum_dense_test.py
JyotinderSingh: Done.

keras/src/layers/core/einsum_dense_test.py
JyotinderSingh: Done.

keras/src/layers/core/dense_test.py
JyotinderSingh: Done.

keras/src/layers/core/dense.py
JyotinderSingh: Done.

keras/src/layers/core/reversible_embedding_test.py
JyotinderSingh: Done.

keras/src/layers/core/reversible_embedding.py
JyotinderSingh: It worked before since we did not have any conditional variables in any of the quantization modes.

With the addition of INT4 sub-channel, `ReversibleEmbedding` required a separate implementation because it defines a conditional variable (`reverse_embeddings_zero`) in its serialization spec that doesn't exist in all configurations (i.e. per-channel INT4 config does not require a zero-point), causing the original parent method to crash with an `AttributeError` when blindly accessing it.

But yes, we can reduce duplication by simply allowing the parent `Embedding` method implementations to generically handle the conditional variables of its subclasses. I've made the change and removed the implementation from `ReversibleEmbedding`.

keras/src/layers/core/reversible_embedding.py
JyotinderSingh: Same explanation and fix as the `save_own_variables` implementation (https://github.com/keras-team/keras/pull/22007#discussion_r2730232380)

keras/src/layers/core/dense.py
JyotinderSingh: Yes this is a breaking change, but it helps us consolidate the quantization implementations and share the quantizers for different configurations. I think it should be acceptable to break compatibility here since the existing INT4 quantization may not have been widely used anyway due to lower accuracy numbers.

However, I can add logic to the (de)serialization hooks to handle this more gracefully, and dynamically reshape the kernel to the new layout when trying to load an older model:
```py
def load_own_variables(self, store):
    ...
    idx = 0
    for name in self.variable_serialization_spec[mode]:
        if name == "kernel":
            loaded_kernel = store[str(idx)]
                
            if mode == "int4":
                current_shape = self._kernel.shape
                loaded_shape = loaded_kernel.shape
                    
                # Check if loaded kernel uses the old layout (packed on axis 0)
                # Old layout: (ceil(input/2), output)
                # New layout: (input, ceil(output/2))
                if current_shape != loaded_shape:
                    # check if dimensions swapped match
                    # the math based on self._orig_input_dim
                    expected_old_rows = (self._orig_input_dim + 1) // 2
                    if loaded_shape == (expected_old_rows, self.units):
                        # This is a legacy checkpoint. Migrate it.
                        # 1. Unpack using old logic (axis=0)
                        unpacked = quantizers.unpack_int4(
                            loaded_kernel, 
                            self._orig_input_dim, 
                            axis=0
                        )
                        # 2. Repack using new logic (axis=1)
                        # Note: We assume per-channel quantization for legacy weights
                        # so we don't need to worry about sub-channel blocking here.
                        repacked, _, _ = quantizers.pack_int4(
                            unpacked, 
                            axis=1
                        )
                        loaded_kernel = repacked

            self._kernel.assign(loaded_kernel)
           ...
```

I felt like the benefits of backwards compatibility here were not worth the added complexity. But I can do that if you think that's the right choice.

keras/src/layers/core/einsum_dense.py
JyotinderSingh: This change does break backwards compatibility (and we can choose to maintain it by adding adapter logic into the serialization hooks, similar to what I've described in another comment)

This change is required because this PR changes how the INT4 quantized kernel is stored in memory for `EinsumDense` layers.

Previously, the kernel retained its original N-dimensional structure, and quantization packing happened along the reduction axis.

The `_int4_build` method now flattens the kernel into a 2D matrix and performs packing along the last axis (columns/output). With this change the original N-dimensional shape cannot be inferred solely from the stored variable's shape.

The LoRA adapters operate in the original full-precision space. The code now uses `self.original_kernel_shape` (which is now explicitly saved) to ensure the LoRA matrices are created with the correct dimensions, rather than the flattened dimensions of the stored quantized kernel.

I've improved the comment to better explain this behavior.

Personally, I feel it should be okay to break compatibility here since INT4 per-channel is a less widely used feature.

keras/src/dtype_policies/dtype_policy.py
JyotinderSingh: Sure I can make a separate fix for this.

JyotinderSingh:
> Question:
> 
> * do all these tests really crash on TF with GPU? Why were they not crashing before?
> 
> One general comment (but not for this PR). This quantization logic is getting quite complicated. Now a vast majority of the code in `Dense`, `EinsumDense`, `Embedding` has to do with quantization. There seems to be a bit of code duplication too between these. I wonder if there is a way to restructure the code. Just something to think about.

1. The skip conditions were initially added for the GPTQ tests in the past, and since we now share some of the INT4 utils with GPTQ I added the skip conditions to the int4 sub-channel tests as well. The tests never crash on my personal machine, when I run them with TF GPU. It looks like some weird CI behavior, which I've been unable to debug.
2. I agree, the quantization logic is becoming quite complex, I'll spend some time thinking about how we can structure it better.

JyotinderSingh:
~JAX failure in unrelated distillation test (passed in several previous gpu runs, so most likely an intermittent failure)~

Re-ran the tests and they now pass.

keras/src/layers/core/dense.py
hertschuh: No, let's not do that

keras/src/layers/core/dense_test.py
hertschuh: Move import at top.

keras/src/layers/core/einsum_dense.py
hertschuh: Ok

hertschuh (commented):
Thanks!

PR #21978, opened 2026-01-04
feat: Add documentation examples for image preprocessing augmentation layers

Abhinavexists:
In reference to #21954
Added documentation examples across 20 image preprocessing augmentation layers.

### Updated Argumentation Layers
- [AugMix](https://keras.io/api/layers/preprocessing_layers/image_augmentation/aug_mix/)
- [CutMix](https://keras.io/api/layers/preprocessing_layers/image_augmentation/cut_mix/) 
- [MaxNumBoundingBox](https://keras.io/api/layers/preprocessing_layers/image_augmentation/max_num_bounding_boxes/)
- [RandAugment](https://keras.io/api/layers/preprocessing_layers/image_augmentation/rand_augment/)
- [RandomColorDegeneration](https://keras.io/api/layers/preprocessing_layers/image_augmentation/random_color_degeneration/)
- [RandomColorJitter](https://keras.io/api/layers/preprocessing_layers/image_augmentation/random_color_jitter/)
- [RandomContrast](https://keras.io/api/layers/preprocessing_layers/image_augmentation/random_contrast/)
- [RandomCrop](https://keras.io/api/layers/preprocessing_layers/image_augmentation/random_crop/)
- [RandomElasticTransform](https://keras.io/api/layers/preprocessing_layers/image_augmentation/random_elastic_transform/)
- [RandomErasing](https://keras.io/api/layers/preprocessing_layers/image_augmentation/random_erasing/)
- [RandomFlip](https://keras.io/api/layers/preprocessing_layers/image_augmentation/random_flip/)
- [RandomGaussianBlur](https://keras.io/api/layers/preprocessing_layers/image_augmentation/random_gaussian_blur/) 
- [RandomGrayscale](https://keras.io/api/layers/preprocessing_layers/image_augmentation/random_grayscale/)
- [RandomInvert](https://keras.io/api/layers/preprocessing_layers/image_augmentation/random_invert/)
- [RandomPerspective](https://keras.io/api/layers/preprocessing_layers/image_augmentation/random_perspective/)
- [RandomPosterization](https://keras.io/api/layers/preprocessing_layers/image_augmentation/random_posterization/)
- [RandomRotation](https://keras.io/api/layers/preprocessing_layers/image_augmentation/random_rotation/)
- [RandomSharpness](https://keras.io/api/layers/preprocessing_layers/image_augmentation/random_sharpness/)
- [RandomShear](https://keras.io/api/layers/preprocessing_layers/image_augmentation/random_shear/)
[...]

PR #22011, opened 2026-01-15
Add RandomRotation segmentation example (fill_mode constant vs crop)

PES2UG23CS205:
This PR adds a visual keras-io example demonstrating how
keras.layers.RandomRotation behaves for segmentation tasks.

It compares fill_mode="constant" vs fill_mode="crop" using
structured dict inputs to ensure synchronized transforms
for images and segmentation masks.

Related to keras-team/keras#21954

cc @innat @mattdangerw

### DIFF
--- a/examples/vision/random_rotation_segmentation.py
+++ b/examples/vision/random_rotation_segmentation.py
@@ -1,0 +1,83 @@
+"""
+RandomRotation with segmentation masks (Keras 3).
+
+This example demonstrates how keras.layers.RandomRotation behaves
+when applied to images and segmentation masks using structured
+dictionary inputs.
+
+We compare:
+- fill_mode="constant": introduces border fill values
+- fill_mode="crop": removes border artifacts by cropping the valid region
+"""
+
+import numpy as np
+import matplotlib.pyplot as plt
+from keras import layers, ops
+
+
+def create_sample():
+    image = np.zeros((128, 128, 1), dtype="float32")
+    image[32:96, 32:96, 0] = 1.0
+
+    mask = np.zeros((128, 128, 1), dtype="int32")
+    mask[32:96, 32:96, 0] = 1
+
+    return image, mask
+
+
+def apply_rotation(layer, image, mask):
+    inputs = {
+        "images": ops.expand_dims(image, axis=0),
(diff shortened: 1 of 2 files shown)


PES2UG23CS205:
Hi @innat 

I’ve opened a PR adding a keras-io example for RandomRotation focusing on
segmentation masks and comparing fill_mode="constant" vs "crop", with
visual before/after outputs.

PR: <https://github.com/keras-team/keras/pull/22011>

Thanks for the suggestion ,happy to iterate if you’d like any changes.

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
```diff
@@ -75,6 +75,38 @@ class RandomRotation(BaseImagePreprocessingLayer):
             `image_data_format` value found in your Keras config file at
             `~/.keras/keras.json`. If you never set it, then it will be
             `"channels_last"`.
+    Examples:
```
hertschuh, line 78: Oh, this part is already addressed here: https://github.com/keras-team/keras/pull/21978

examples/vision/random_rotation_segmentation.py
```diff
@@ -0,0 +1,83 @@
+"""
```
hertschuh, line 1: We don't add examples in the keras repo like this because people won't find them and they won't see the pictures. The right place for discoverability is [keras.io](https://keras.io/) which you can contribute to here: https://github.com/keras-team/keras-io where images get rendered for guides and examples.

The thing is, we normally do end-to-end examples, having an example just about RandomRotation would be very narrow. However, we could really use a guide showing all the image augmentation layers.

Is that something you're willing to work on?

PES2UG23CS205:
Hi @hertschuh,

Thanks for the guidance ,that makes sense.

Yes, I’m willing to work on a keras-io guide covering image augmentation layers.
I’ll close/move this example and start working on a proper keras-io guide instead.

Please let me know if there are any preferred structure or examples you'd like included.

hertschuh:
@PES2UG23CS205 

Great, thanks!

I don't really have a preferred structure.

PR #22013, opened 2026-01-15
Refactor ExtractPatches to handle both 2D and 3D

MarcosAsh:
Follow-up to #21980 as requested by hertschuh.

## Summary
Consolidates ExtractPatches and ExtractPatches3D into a single unified class that handles both 2D and 3D patch extraction.

## Changes
- Renamed _extract_patches to _extract_patches_2d
- Created new _extract_patches dispatcher that routes to _extract_patches_2d or _extract_patches_3d based on size argument
- Updated ExtractPatches class to handle both 2D and 3D via is_3d flag
- Simplified extract_patches function to use single code path
- Removed ExtractPatches3D class

## Tests
All existing extract_patches tests pass.

### DIFF
--- a/keras/src/ops/image.py
+++ b/keras/src/ops/image.py
@@ -622,6 +622,15 @@ class ExtractPatches(Operation):
                 out_shape = out_shape[1:]
         return KerasTensor(shape=out_shape, dtype=images.dtype)
 
+    def get_config(self):
+        return {
+            "size": self.size,
+            "strides": self.strides,
+            "dilation_rate": self.dilation_rate,
+            "padding": self.padding,
+            "data_format": self.data_format,
+        }
+
 
 @keras_export("keras.ops.image.extract_patches")
 def extract_patches(


keras/src/ops/image.py
```diff
@@ -602,8 +613,13 @@ def compute_output_spec(self, images):
...
-        if original_ndim == 3:
-            out_shape = out_shape[1:]
+
+        if self.is_3d:
+            if original_ndim == 4:
+                out_shape = out_shape[1:]
+        else:
+            if original_ndim == 3:
+                out_shape = out_shape[1:]
         return KerasTensor(shape=out_shape, dtype=images.dtype)
```
hertschuh, line 624: Oh, I just noticed, this class is missing a `get_config`. Can you add it? It needs to return `size`, `strides`, `dilation_rate`, `padding`, `data_format`, but not `is_3d`.

MarcosAsh:
Just did the changes. Tell me if you need anything else done.

hertschuh (approved):
Thanks for the follow-up!

PR #22015, opened 2026-01-15
docs: Add guide on device placement using keras.device()

### LINKED ISSUE #21988: Allow manual choice of device (mesh) for new Tensor allocation on all backends
As far as I can see, it is currently not possible to specify in a backend-agnostic way which device or device mesh a tensor should be allocated on. The only way is to use the backend-specific allocation methods.

Since GPU memory is often at a premium, I think it would be beneficial to add the functionality to allocate on CPU, or choose which device or mesh a newly allocated tensor uses, ideally via both an argument and a context manager - defaulting to the current behavior if not specified.

If the functionality is not implemented for sake of complexity with respect to the differences in backends for multi-device utilities, I think it should at least be possible to specify a single device for now, at the user's risk of course.

Sikandar1310291:
This PR adds documentation for the keras.device() context manager, which provides backend-agnostic device placement for tensor allocation.

Changes
Added a new section "Controlling device placement" to 
guides/training_with_built_in_methods.py
 that covers:

Using the device context manager - Basic usage with code examples
Device naming conventions - Explains the device_type:device_index format
Practical examples - CPU preprocessing with GPU training pattern
Motivation
This addresses issue #21988 where a user requested backend-agnostic device placement. The functionality already exists via keras.device() but was not documented in the guides.

Related Issues
Closes #21988

### DIFF
--- a/keras/src/backend/__init__.py
+++ b/keras/src/backend/__init__.py
@@ -82,9 +82,10 @@ def device(device_name):
     multi-device setups.
 
     Args:
-        device_name: String specifying the device in format `"device_type:device_index"`.
-            For example: `"cpu:0"`, `"gpu:0"`, `"gpu:1"`.
-            For PyTorch backend, `"gpu"` is automatically converted to `"cuda"`.
+        device_name: String specifying the device in format
+            `"device_type:device_index"`. For example: `"cpu:0"`, `"gpu:0"`,
+            `"gpu:1"`. For the PyTorch backend, `"gpu"` is automatically
+            converted to `"cuda"`.
 
     Example:
 
@@ -127,8 +128,8 @@ def device(device_name):
     Use cases:
 
     - **Memory management**: Keep large tensors on CPU to save GPU memory
-    - **Data preprocessing**: Process data on CPU before moving to GPU for training
-    - **GPU / TPU setups**: Control what runs on GPU / TPU vs CPU 
+    - **Data preprocessing**: Process data on CPU before training on GPU
+    - **GPU / TPU setups**: Control what runs on GPU / TPU vs CPU
     - **Multi-device setups**: Control which device receives which tensors
 
     Device naming conventions:


hertschuh:
@Sikandar1310291 

I see you pushed some code to this branch, but the example is still in the wrong file. Did something go wrong?

Sikandar1310291:
> @Sikandar1310291
> 
> I see you pushed some code to this branch, but the example is still in the wrong file. Did something go wrong?

@hertschuh Apologies for the confusion! You're right - my previous commits were only attempting to regenerate API files without actually moving the documentation.

I've now corrected this in commit e70a57513. The device placement documentation has been:

Added as a comprehensive docstring to the 
device()
 function in 
keras/src/backend/
init
.py
 (lines 77-147)
Removed entirely from 
guides/training_with_built_in_methods.py
The docstring includes self-contained examples and will now appear on keras.io/api/. Please let me know if this looks correct!

keras/src/backend/__init__.py
```diff
@@ -74,4 +74,74 @@ class name_scope(backend_name_scope):
...
+            For example: `"cpu:0"`, `"gpu:0"`, `"gpu:1"`.
+            For PyTorch backend, `"gpu"` is automatically converted to `"cuda"`.
+
+    Example:
+
+    Basic usage with CPU and GPU:
+
+    ```python
+    import keras
+    import numpy as np
```
hertschuh: Remove these imports. We don't put them in code snippets in docstring, they're implied.

keras/src/backend/__init__.py
```diff
@@ -74,4 +74,74 @@ class name_scope(backend_name_scope):
 
 @keras_export("keras.device")
 def device(device_name):
+    """Context manager for backend-agnostic device placement.
+
+    Use this context manager to control which device tensors are allocated on across
+    all backends (TensorFlow, JAX, PyTorch). This is useful for memory management,
+    data preprocessing, and multi-device setups.
```
hertschuh: Let's phrase it a bit differently.

```
Use this context manager to control on which device operations are performed and tensors
are allocated. This works across all backends (TensorFlow, JAX, PyTorch).
```

keras/src/backend/__init__.py
```diff
@@ -74,4 +74,74 @@ class name_scope(backend_name_scope):
...
+    # Allocate tensors on GPU (if available)
+    with keras.device("gpu:0"):
+        gpu_tensor = keras.ops.ones((2, 2))
+    ```
+
+    Practical example with CPU preprocessing and GPU training:
+
+    ```python
+    import keras
+    import numpy as np
```
hertschuh: Remove imports.

keras/src/backend/__init__.py
```diff
@@ -74,4 +74,74 @@ class name_scope(backend_name_scope):
...
+    # Train on GPU (if available)
+    with keras.device("gpu:0"):
+        model.fit(x_processed, y_train, epochs=2)
+    ```
+
+    Use cases:
+
+    - **Memory management**: Keep large tensors on CPU to save GPU memory
+    - **Data preprocessing**: Process data on CPU before moving to GPU for training
+    - **Multi-GPU setups**: Explicitly control which GPU receives which tensors
```
hertschuh: I feel like it's missing a key element (number 3 out of 4 in the list), which is that you want to control where things are executed, you want to run your model on GPU for instance.

hertschuh (commented):
Thanks for the documentation!

keras/src/backend/__init__.py
Sikandar1310291: Thank you for the review @hertschuh! I've addressed all your feedback in the latest commit:

✅ Rephrased the intro to: "Use this context manager to control on which device operations are performed and tensors are allocated. This works across all backends..."
✅ Removed all imports from code snippets
✅ Added "Device execution" use case: "Run your model on GPU for faster computation"
Please take another look! 🙏

hertschuh:
FYI, you made the changes in the other PR instead of this one.

keras/src/backend/__init__.py
hertschuh: @Sikandar1310291 

Sorry, this completely slipped out. I don't see the changes that you mention in this checklist. Did you forget to push them?

hertschuh:
Thanks!

PR #22020, opened 2026-01-17
Allow `CenterCrop` layer to handle dynamic image sizes.

hertschuh:
Previously the layer would raise when dynamically sized images where passed.


hertschuh:
@MarcosAsh

MarcosAsh:
This fix makes a lot more sense. Ill keep it in mind for future fixes.

Issue #22018, opened 2026-01-16
Activity Regularizer Not Normalized by Batch Size


fwesel:
The documentation in `keras/src/regularizers/regularizers.py` lines 29-53 states:

> "The value returned by the `activity_regularizer` is divided by the input batch size"

This is not true. The example in the docs claims `ops.sum(layer.losses)` equals `5.25`, but it actually returns `25.25`.

**Minimal reproduction following the example in the docs in `keras/src/regularizers/regularizers.py`:**

```python
import os
os.environ["KERAS_BACKEND"] = "jax"
from keras.layers import Dense
from keras.regularizers import L1, L2
from keras import ops

layer = Dense(5, input_dim=5, kernel_initializer='ones',
              kernel_regularizer=L1(0.01), activity_regularizer=L2(0.01))
tensor = ops.ones(shape=(5, 5)) * 2.0
out = layer(tensor)

print(f"Total: {float(ops.sum(layer.losses))}")
# Expected: 5.25 (with normalization)
# Actual: 25.25 (without normalization)
```

The activity regularizer returns `25.0` instead of `5.0` (not divided by batch_size=5). This means activity regularization strength scales with batch size, making hyperparameter tuning inconsistent across different batch sizes.

**Environment:** Keras 3.x, JAX backend

Abhinavexists:
Hi @fwesel , i was able to reproduce this error and it was a calculation miss in the `layer.py` where division was not happening  via batch size , fix was relatively small.

fwesel:
Hi @Abhinavexists, great work with the PR! I have one doubt: I think batch scaling should affect all types of regularization, not just activity regularization. From a maximum likelihood point of view, when training any deep learning model we maximize the likelihood. Activity regularization is then equivalent to placing a prior on the outputs of a layer, and kernel regularization is equivalent to placing a prior on the weights of a layer. These priors multiply the likelihood and are jointly optimized (so we maximize the posterior). In practice we maximize the log-likehood, or log-posterior if there are any priors, and when doing so, we normalize the log-likehood or log-posterior by the batch size. This means that all terms in the loss (which is the log-posterior) should be normalized by batch size, not just the likelihood and activity regularizer, but also kernel or any other type of layer regularization.

Abhinavexists:
[...]

PR #22021, opened 2026-01-17
Fix: activity regularizer not normalized by batch size

Abhinavexists:
fixes: #22018 

The documentation in `keras/src/regularizers/regularizers.py` states that activity regularizer losses are divided by batch size to ensure consistent regularization strength across different batch sizes. 

However, the implementation was not performing this normalization, causing activity regularization strength to scale with batch size.

### DIFF
--- a/keras/src/layers/layer.py
+++ b/keras/src/layers/layer.py
@@ -982,9 +982,7 @@ class Layer(BackendLayer, Operation):
                                 batch_size = ops.cast(
                                     ops.shape(output)[0], dtype=loss.dtype
                                 )
-                                loss = ops.where(
-                                    batch_size > 0, loss / batch_size, 0.0
-                                )
+                                loss = ops.divide_no_nan(loss, batch_size)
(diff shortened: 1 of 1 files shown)


keras/src/layers/layer.py
```diff
@@ -974,7 +975,17 @@ def maybe_convert(x):
...
+                            loss = self.activity_regularizer(output)
+                            if output.ndim > 0:
+                                # Normalize by batch size to ensure consistent
+                                # regularization strength across batch sizes
+                                batch_size = ops.cast(
+                                    ops.shape(output)[0], dtype=loss.dtype
+                                )
+                                loss = ops.where(
+                                    batch_size > 0, loss / batch_size, 0.0
+                                )
```
hertschuh: Just do `ops.divide_no_nan(loss, batch_size)`, it should be faster.

Abhinavexists: updated with divide_no_nan

PR #22022, opened 2026-01-17
Increase test coverage for IntegerLookup layer

MarcosAsh:
## Description
Addresses TODO in integer_lookup_test.py to increase test coverage.
## Changes
Added tests for:
- max_tokens - vocabulary truncation
- mask_token - masking behavior
- invert - inverse lookup
- pad_to_max_tokens - output padding
- num_oov_indices - multiple OOV buckets
- num_oov_indices=0 - no OOV handling
- get_vocabulary() - vocabulary retrieval after adapt
- adapt() with steps parameter
- sparse=True - sparse tensor output
- vocabulary from file path
- Invalid argument validation (max_tokens, num_oov_indices, vocabulary_dtype)

### DIFF
--- a/keras/src/layers/preprocessing/integer_lookup_test.py
+++ b/keras/src/layers/preprocessing/integer_lookup_test.py
@@ -3,8 +3,8 @@ import os
 import numpy as np
 from tensorflow import data as tf_data
 
-from keras.src import layers
 from keras.src import backend
+from keras.src import layers
 from keras.src import testing
 
 


keras/src/layers/preprocessing/integer_lookup_test.py
```diff
@@ -153,3 +153,104 @@ def test_tf_idf_output_shape(self):
...
+
+    def test_num_oov_indices(self):
+        layer = layers.IntegerLookup(
+            vocabulary=[1, 2, 3],
+            num_oov_indices=2,
+            output_mode="int",
+        )
+        output = layer([1, 2, 3, 999, 1000])
+        self.assertAllClose(output[:3], np.array([2, 3, 4]))
+        self.assertTrue(all(o in [0, 1] for o in np.array(output[3:])))
```
hertschuh: This line is failing on GPU with Torch. I think you need to change it to:

```python
self.assertTrue(all(o in [0, 1] for o in backend.convert_to_numpy(output[3:])))
```

MarcosAsh:
Hi, thanks for the feedback. Just pushed the changes

keras/src/layers/preprocessing/integer_lookup_test.py
```diff
@@ -153,3 +153,104 @@ def test_tf_idf_output_shape(self):
...
+
+    def test_num_oov_indices(self):
+        layer = layers.IntegerLookup(
+            vocabulary=[1, 2, 3],
+            num_oov_indices=2,
+            output_mode="int",
+        )
+        output = layer([1, 2, 3, 999, 1000])
+        self.assertAllClose(output[:3], np.array([2, 3, 4]))
+        self.assertTrue(all(o in [0, 1] for o in backend.convert_to_numpy(output[3:])))
```
hertschuh: This line is failing on GPU with Torch. I think you need to change it to:

```python
self.assertTrue(all(o in [0, 1] for o in backend.convert_to_numpy(output[3:])))
```

hertschuh: Now the code formatter is complaining about this line, which is too long.

```python
      self.assertTrue(
          all(o in [0, 1] for o in backend.convert_to_numpy(output[3:]))
      )
```

You can always run `ruff --config pyproject.toml format .` locally.

PR #22027, opened 2026-01-18
Torch backend: allow explicit device selection and guard DirectML usage

### LINKED ISSUE #21228: Keras 3 may not work with PyTorch DirectML
I think this might be a tech issue so I opened issue from [disscutions](https://github.com/keras-team/keras/discussions/21218) to here.

Issue description: 
My test on Keras 3 may cannot set selected device correctly with DirectML based PyTorch backend. I looked #21190 and still can't find some useful anwers.

Backgrounds:
I have a Windows 11 Machine, contains 2 cards with NVIDIA GeForce RTX 4070ti Super and AMD Radeon RX 7800XT respectively.
Then I deployed a Python 3.11.9 env installed `torch-directml`, by using DirectML backend to use PyTorch. In PyTorch we can manually use `.to(device)` to set GPU device, and can do check like this:
```
import torch_directml

# List all available DML devices
device_count = torch_directml.device_count()
print(f"Available DirectML devices: {device_count}")

# Loop over devices and print info
for i in range(device_count):
    dml_device = torch_directml.device(i)
    print(f"Device {i}: {dml_device}")

# Select device 1 (example)
dml = torch_directml.device(1)
```
```
Available DirectML devices: 2
Device 0: privateuseone:0
Device 1: privateuseone:1
```
From the above codes and the result, we can know that device 1 is the AMD card.
How can I manage/add preprocess "on DirectML device select" in Keras? If we just do nothing, the model will be run on CPU only.

Hardware and Software/Environ:
```
Intel Xeon W-3175X
NVIDIA GeForce RTX 4070ti Super
AMD Radeon RX 7800XT

Windows 11 24H2
Python 3.13.2 ---> PyTorch 2.6.0+cu126, Keras 3            # The NVIDIA CUDA one
Python 3.11.9 ---> PyTorch-DirectML, Keras 3                 # The DirectML for AMD card environment
```

PES2UG23CS205:
### Summary
This PR adds support for explicit device selection in the Torch backend via the
`KERAS_TORCH_DEVICE` environment variable.

### Motivation
The Torch backend currently assumes CUDA-only GPU support. This causes issues
for non-CUDA backends such as PyTorch DirectML (`privateuseone` devices), leading
to silent CPU fallback or cryptic errors.

### What this PR does
- Allows users to explicitly specify a Torch device using `KERAS_TORCH_DEVICE`
- Preserves existing CUDA and CPU behavior
- Guards DirectML usage with a clear error if `torch-directml` is unavailable
- Adds a warning that some operations may fall back to CPU on DirectML devices

### Notes
- DirectML is Windows-only. On non-Windows systems, requesting a DirectML device
  will raise a clear error if `torch-directml` is not available.
- This PR does not add a hard dependency on `torch-directml`.

Fixes #21228

### DIFF
--- a/keras/src/backend/torch/core.py
+++ b/keras/src/backend/torch/core.py
@@ -1,9 +1,7 @@
 import builtins
 import contextlib
 import functools
-import importlib.util
 import os
-import warnings
 
 import ml_dtypes
 import numpy as np
@@ -30,30 +28,16 @@ IS_THREAD_SAFE = True
 # are not currently implemented for the MPS device.
 # check https://github.com/pytorch/pytorch/issues/77764.
 if "KERAS_TORCH_DEVICE" in os.environ:
-    device = os.environ["KERAS_TORCH_DEVICE"]
-
-    # Guard DirectML usage
-    if device.startswith("privateuseone"):
-        if importlib.util.find_spec("torch_directml") is None:
-            raise ImportError(
-                "KERAS_TORCH_DEVICE is set to a DirectML device "
-                f"({device}), but torch-directml is not available. "
-                "Please install torch-directml and ensure it can be imported."
-            )
-
-    DEFAULT_DEVICE = device
+    DEFAULT_DEVICE = os.environ["KERAS_TORCH_DEVICE"]
+elif torch.backends.mps.is_available():
+    DEFAULT_DEVICE = "mps"
+elif torch.cuda.is_available():
+    DEFAULT_DEVICE = "cuda"
+elif hasattr(torch, "xpu") and torch.xpu.is_available():
+    DEFAULT_DEVICE = "xpu"
 else:
-    if torch.cuda.is_available():
-        DEFAULT_DEVICE = "cuda"
-    else:
-        DEFAULT_DEVICE = "cpu"
+    DEFAULT_DEVICE = "cpu"
 
-
-if DEFAULT_DEVICE.startswith("privateuseone"):
-    warnings.warn(
-        "Using PyTorch DirectML device. "
-        "Some operations may fall back to CPU due to limited backend support."
-    )
 TORCH_DTYPES = {
     "float16": torch.float16,
     "float32": torch.float32,


PES2UG23CS205:
I’ve opened a PR that adds explicit Torch device selection via
`KERAS_TORCH_DEVICE` and guards DirectML usage with a clear error when
`torch-directml` is unavailable. This removes CUDA-only assumptions and avoids
silent CPU fallback.

PES2UG23CS205:
Is using an environment variable (KERAS_TORCH_DEVICE) the preferred long-term approach for Torch device selection, or would you prefer a higher-level Keras API in the future?

keras/src/backend/torch/core.py
```diff
@@ -26,14 +28,34 @@
...
+                "Please install torch-directml and ensure it can be imported."
+            ) from e
+
+    DEFAULT_DEVICE = device
 else:
-    DEFAULT_DEVICE = "cpu"
+    if torch.cuda.is_available():
+        DEFAULT_DEVICE = "cuda"
+    else:
+        DEFAULT_DEVICE = "cpu"
```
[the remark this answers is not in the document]
hertschuh: +1 please re-add those.

Also, combine `else: if` to `elif` so that everything is unindented.

examples/vision/random_rotation_segmentation.py
```diff
@@ -0,0 +1,83 @@
+"""
+RandomRotation with segmentation masks (Keras 3).
```
hertschuh: This file is unrelated to this PR, please revert

keras/src/layers/preprocessing/image_preprocessing/random_rotation.py
```diff
@@ -75,6 +75,38 @@ class RandomRotation(BaseImagePreprocessingLayer):
             `image_data_format` value found in your Keras config file at
             `~/.keras/keras.json`. If you never set it, then it will be
             `"channels_last"`.
+    Examples:
```
hertschuh: This file is unrelated to this PR, please revert

keras/src/backend/torch/core.py
```diff
@@ -26,15 +29,31 @@
...
+    device = os.environ["KERAS_TORCH_DEVICE"]
+
+    # Guard DirectML usage
+    if device.startswith("privateuseone"):
+        if importlib.util.find_spec("torch_directml") is None:
+            raise ImportError(
+                "KERAS_TORCH_DEVICE is set to a DirectML device "
+                f"({device}), but torch-directml is not available. "
+                "Please install torch-directml and ensure it can be imported."
+            )
```
hertschuh: `privateuseone` does not imply DirectML, it could be any new backend. So we cannot make this assumption.

keras/src/backend/torch/core.py
```diff
@@ -26,15 +29,31 @@
...
+    DEFAULT_DEVICE = device
 else:
-    DEFAULT_DEVICE = "cpu"
+    if torch.cuda.is_available():
+        DEFAULT_DEVICE = "cuda"
+    else:
+        DEFAULT_DEVICE = "cpu"
 
+
+if DEFAULT_DEVICE.startswith("privateuseone"):
```
hertschuh: `privateuseone` does not imply DirectML, it could be any new backend. So we cannot make this assumption.

hertschuh (commented):
I agree with the concept of adding `KERAS_TORCH_DEVICE`.

But we can't assume `privateuseone` is DirectML.

PES2UG23CS205:
Thanks for the detailed review, that makes sense.

You’re right: privateuseone is a generic PyTorch mechanism and should not be interpreted as DirectML. I’ll remove all DirectML-specific assumptions and let Keras treat KERAS_TORCH_DEVICE as a pure pass-through to Torch.

I’ll also:

revert the unrelated RandomRotation changes,

restore mps/xpu auto-detection in the fallback logic,

keep this PR strictly scoped to explicit device selection only.

I’ll push an updated revision shortly.

One quick design question for alignment: is treating KERAS_TORCH_DEVICE as a pure pass-through to Torch the intended long-term approach, or do you envision Keras eventually providing a higher-level device abstraction for the Torch backend?

hertschuh:
> One quick design question for alignment: is treating KERAS_TORCH_DEVICE as a pure pass-through to Torch the intended long-term approach, or do you envision Keras eventually providing a higher-level device abstraction for the Torch backend?

This may not be a quick question :)

We don't have any plan for a higher-level device abstraction for the Torch backend right now, but do you have something in mind?

PES2UG23CS205:
That makes sense , I agree a higher-level device abstraction would be a much larger design effort and not something to introduce lightly.

I don’t have a concrete proposal for a new abstraction right now. My intent with the question was mostly to confirm that, for the foreseeable future, treating KERAS_TORCH_DEVICE as a thin pass-through to Torch (without Keras interpreting device semantics) is the right direction.

If anything, the only incremental follow-up I could imagine , and only if it’s useful  would be small UX improvements like clearer documentation or comments around expected device strings, rather than new APIs or abstractions. But I’m happy to keep this PR focused purely on explicit device selection and leave broader design questions aside.

hertschuh:
> But I’m happy to keep this PR focused purely on explicit device selection and leave broader design questions aside.

Yes, let's do that.

PR #22030, opened 2026-01-19
Fix orbax checkpoint loading

pctablet505:
This pull request improves the handling of asynchronous checkpoint saves and enhances the detection logic for Orbax checkpoints in Keras. The main changes include tracking pending asynchronous saves to ensure proper cleanup and completion, and making the checkpoint detection more robust by checking for both marker files and step subdirectories.

**Checkpoint save management improvements:**

* Added a `_pending_saves` list to the `OrbaxCheckpoint` callback to track asynchronous save operations for better management and cleanup.
* Modified the async save logic to store the returned future in `_pending_saves` and ensure all pending saves are waited on and cleared at the end of training and in `wait_until_finished()`. [[1]](diffhunk://#diff-e9f4b69b04c99695dc3a9f46eca20f073f95fbc4e14fa7ec1308339b376c1621L255-R257) [[2]](diffhunk://#diff-e9f4b69b04c99695dc3a9f46eca20f073f95fbc4e14fa7ec1308339b376c1621R322-R324) [[3]](diffhunk://#diff-e9f4b69b04c99695dc3a9f46eca20f073f95fbc4e14fa7ec1308339b376c1621L329-R336)

**Orbax checkpoint detection enhancements:**

* Updated `is_orbax_checkpoint` in `orbax_util.py` to check for the presence of the `orbax.checkpoint` file and for step subdirectories named with digits, making the detection more reliable even if the Orbax API is unavailable.


keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -326,7 +335,9 @@ def wait_until_finished(self):
         have completed across all hosts in a multi-host setup.
         """
         # Wait for any async operations to complete on this host
-        self.checkpointer.wait()
+        for future in self._pending_saves:
+            future.result()
+        self._pending_saves.clear()
```
hertschuh: So `wait_until_finished` is only called by unit tests, it's not part of the `Callback` protocol, so it's not actually use in real world scenarios.

1. We should delete it
2. We should change the unit tests to not need it. Calling `on_train_end` instead should work hopefully.

hertschuh (commented):
Thanks for the fix!

One more change:

keras/src/saving/orbax_util.py
```diff
@@ -10,11 +10,23 @@ def is_orbax_checkpoint(filepath):
...
+        has_step_subdirs = any(
+            os.path.isdir(os.path.join(filepath, item)) and item.isdigit()
+            for item in items
+        )
+        if has_step_subdirs:
+            return True
+    except OSError:
+        pass
+
+    return False
```
hertschuh: I had not noticed earlier, but Amit put up a PR for this that seems more thorough. I'll take his PR. Can you revert this part and then rebase?

hertschuh:
@pctablet505 

It looks like you did a merge from master instead of a rebase: 
https://github.com/keras-team/keras/pull/22030/changes

Can you rebase?

pctablet505:
https://github.com/keras-team/keras/pull/22036
was unable to make changes in this branch, so created new PR for it.

PR #22035, opened 2026-01-20
Disallow TFSMLayer deserialization in safe_mode to prevent external SavedModel execution

0xManan:
TFSMLayer currently loads external TensorFlow SavedModels during deserialization
without respecting Keras `safe_mode`. Although TensorFlow does not execute
SavedModel functions at load time, the attacker-controlled graph is registered
during deserialization and executed during normal model invocation, violating
the security guarantees of `safe_mode=True`.

This change disallows instantiation of `TFSMLayer` when `safe_mode` is enabled,
both during direct construction and during deserialization via `from_config()`.
This matches the existing security model used by other potentially unsafe Keras
components (e.g. Lambda layers) and prevents loading of untrusted executable
graph artifacts without explicit user opt-in.

Specifically:
- Add a `safe_mode` check in `TFSMLayer.__init__` to prevent loading external
  SavedModels when unsafe deserialization is not explicitly enabled.
- Override `from_config()` to block deserialization of `TFSMLayer` when
  `safe_mode=True`, preventing config-based gadget abuse.
- Provide clear error messages guiding users to explicitly opt out via
  `safe_mode=False` or `keras.config.enable_unsafe_deserialization()` when the
  source is trusted.

This change preserves backward compatibility for trusted workflows while
closing a safe_mode bypass that could otherwise lead to attacker-controlled
graph execution during inference.

Security report:
https://huntr.com/bounties/7e78d6f1-6977-4300-b595-e81bdbda331c

### DIFF
--- a/keras/src/export/tfsm_layer_test.py
+++ b/keras/src/export/tfsm_layer_test.py
@@ -124,9 +124,7 @@ class TestTFSMLayer(testing.TestCase):
         temp_model_filepath = os.path.join(self.get_temp_dir(), "m.keras")
         model.save(temp_model_filepath, save_format="keras_v3")
         reloaded_model = saving_lib.load_model(
-            temp_model_filepath,
-            custom_objects={"TFSMLayer": tfsm_layer.TFSMLayer},
-            safe_mode=False,
+            temp_model_filepath, safe_mode=False
         )
         self.assertAllClose(reloaded_model(ref_input), ref_output, atol=1e-7)
 
@@ -153,11 +151,7 @@ class TestTFSMLayer(testing.TestCase):
             saving_lib.load_model(model_path)
 
         # Explicit opt-out should allow loading
-        loaded_model = saving_lib.load_model(
-            model_path,
-            custom_objects={"TFSMLayer": tfsm_layer.TFSMLayer},
-            safe_mode=False,
-        )
+        loaded_model = saving_lib.load_model(model_path, safe_mode=False)
 
         x = tf.random.normal((2, 10))
         self.assertAllClose(loaded_model(x), wrapper_model(x))


yomananchill:
@hertschuh I have made changes in current implementation for the fix. Please review it. I have tested on my end. It will throw this error.
image

0xManan:
@hertschuh The test "[Tests / Run tests (3.11, tensorflow) (pull_request)](https://github.com/keras-team/keras/actions/runs/21172127983/job/60891247152?pr=22035)" Failed because the code is implemented correctly and it prevents model to load unsafe external models. 

My suggestion would be to change test cases for all files or keep it this way for now.

keras/src/export/tfsm_layer.py
```diff
@@ -146,3 +147,37 @@ def get_config(self):
...
+        )
+
+        if effective_safe_mode is not False:
+            raise ValueError(
+                "Loading a TFSMLayer from config is disallowed when "
+                "`safe_mode=True` because it loads an external SavedModel "
+                "that may contain attacker-controlled executable graph code. "
+                "If you trust the source, pass `safe_mode=False` to the "
+                "loading function, or call "
+                "`keras.config.enable_unsafe_deserialization()`."
```
hertschuh: Let's make the error message a bit more consistent with the Lambda layer one:
```python
                "Requested the deserialization of a `TFSMLayer`, which "
                "loads an external SavedModel. This carries a potential risk "
                "of arbitrary code execution and thus it is disallowed by "
                "default. If you trust the source of the artifact, you can "
                "override this error by passing `safe_mode=False` to the "
                "loading function, or calling "
                "`keras.config.enable_unsafe_deserialization()."

File: keras/src/export/tfsm_layer.py [hunk 2]
```diff
@@ -1,38 +1,32 @@
...
 from keras.src.api_export import keras_export
 from keras.src.export.saved_model import _list_variables_used_by_fns
+from keras.src.saving import serialization_lib
 from keras.src.utils.module_utils import tensorflow as tf
 
 
 @keras_export("keras.layers.TFSMLayer")
 class TFSMLayer(layers.Layer):
     """Reload a Keras model/layer that was saved via SavedModel / ExportArchive.
-
```
hertschuh, line 11: Undo all the spurious changes here: these empty lines that were removed should stay.

keras/src/export/tfsm_layer_test.py
```diff
@@ -28,10 +28,16 @@ def test_reloading_export_archive(self):
 
         saved_model.export_saved_model(model, temp_filepath)
         reloaded_layer = tfsm_layer.TFSMLayer(temp_filepath)
-        self.assertAllClose(reloaded_layer(ref_input), ref_output, atol=1e-7)
+
```
hertschuh: Almost all of the changes in this file are spurious changes (line re-wrapping, comments being removed).
Can you undo all of those?

keras/src/export/tfsm_layer.py
hertschuh: Please undo all the empty line removal in this file. This is what is making the "code format" test fail.

keras/src/export/tfsm_layer_test.py
```diff
@@ -124,9 +126,42 @@ def test_serialization(self):
...
+        with self.assertRaisesRegex(
+            ValueError,
+            "arbitrary code execution",
+        ):
+            saving_lib.load_model(model_path)
+
+        # Explicit opt-out should allow loading
+        loaded_model = saving_lib.load_model(
+            model_path,
+            custom_objects={"TFSMLayer": tfsm_layer.TFSMLayer},
```
hertschuh: You don't need the `custom_objects` argument.

yomananchill:
> Can you check why the tests are failing?
> 
> You can test locally with:
> 
> ```
> KERAS_BACKEND=tensorflow pytest keras/src/export/tfsm_layer_test.py
> ```

Will check and get back to you shortly.

hertschuh (approved):
Thanks!

I allowed myself to fix the tests and formatting.

yomananchill:
Sorry for the delayed follow-up on my side - I got pulled into other work.
Thanks a lot for stepping in to fix the remaining test and formatting issues and for pushing this through. I really appreciate your help in getting this merged.

PR #22036, opened 2026-01-20
Fix async save handling in OrbaxCheckpoint

pctablet505:
- Add _pending_saves list to track async save futures
- Wait for all pending saves to complete in on_train_end
- Update tests to use on_train_end() instead of wait_until_finished()

This ensures that async checkpoint saves complete before training ends, preventing potential data loss or incomplete checkpoints.


pctablet505:
@amitsrivastava78 Do we still need this PR or shall I close it?

PR #22039, opened 2026-01-21
Fix timedistributed mask validation

### LINKED ISSUE #22037: TimeDistributed silently accepts mask with mismatched timesteps
## Description
In Keras 3.13.1 (also in 3.9.0), layers.TimeDistributed does not raise an error when the provided mask has mismatched timesteps. The layer still runs without any warnings or exceptions. In Keras 3.8.0, this correctly raises a ValueError, which seems like the expected behavior. This looks like a regression, possibly introduced in gh-20765.

### Minimal example
```
from tensorflow.keras import layers, models
import numpy as np


batch = 4
timesteps = 5
features = 3
  
td = layers.TimeDistributed(layers.Dense(units=5, activation="softmax"))
inputs = np.zeros((batch, timesteps, features))
mask_timemismatch = np.ones((batch, timesteps + 1), dtype=bool)


td(
   inputs,
   mask=mask_timemismatch
) 
)
```

## Behaviors
Keras 3.13.1 & 3.9.0: No error is raised
Keras 3.8.0: Raises a ValueError
```
ValueError: Exception encountered when calling TimeDistributed.call().
`TimeDistributed` Layer should be passed a `mask` of shape (4, 5, ...), received: mask.shape=(4, 6)
Arguments received by TimeDistributed.call():
  • inputs=tf.Tensor(shape=(4, 5, 3), dtype=float32)
  • training=None
  • mask=tf.Tensor(shape=(4, 6), dtype=bool)
```

## Environment
Keras: 3.13.1, 3.9.0
Python: 3.11
OS: macOS

pctablet505:
# Fix TimeDistributed Mask Validation

## Problem

The `TimeDistributed` layer was silently accepting masks with mismatched timesteps (regression introduced after Keras 3.8.0). When a mask's shape doesn't match the expected `(batch_size, timesteps, ...)` format, the layer should raise a `ValueError`, but it was processing the invalid mask without error.

## Solution

This PR restores and strengthens mask validation in the `TimeDistributed` layer:

- **Enhanced Mask Validation Logic**: Improved shape checks to ensure mask dimensions align with input dimensions. The validation correctly handles both TensorFlow (eager and graph modes) and non-TensorFlow backends and raises a clear `ValueError` when mismatches are detected.
- **Comprehensive Tests**: Added a regression test to ensure that a `ValueError` is raised if a `TimeDistributed` layer receives a mask with a mismatched number of timesteps.

## Changes

- **Refactored mask validation logic** in `time_distributed.py`: Consolidated backend-specific and execution-mode checks into a single validation block, eliminating code duplication while maintaining correctness for TensorFlow (eager/graph modes) and non-TensorFlow backends.
- Added `time_distributed_test.py` regression test covering mismatched mask timesteps.

## Testing

The PR verifies that:

- ✅ Correctly shaped masks `(batch_size, timesteps, ...)` are allowed.
- ✅ Masks with mismatched timesteps raise a `ValueError` with a clear error message.
- ✅ Behavior is correct for both TensorFlow and non-TensorFlow backends and in eager/graph execution modes.

## Related

Fixes #22037

### DIFF
--- a/requirements-tensorflow-cuda.txt
+++ b/requirements-tensorflow-cuda.txt
@@ -13,4 +13,3 @@ torch
 jax[cpu]
 
 -r requirements-common.txt
-psutil


pctablet505:
@gemini-code-assist review

pctablet505:
@gemini-code-assist review

keras/src/layers/rnn/time_distributed.py
```diff
@@ -76,31 +76,77 @@ def call(self, inputs, training=None, mask=None):
...
-            )
+        # Validate mask shape
+        if mask_shape is not None:
+            # For TF backend with graph mode and `partial_batch_size`, skip
+            # evaluation of `batch_size` as it can be a `strided_slice` and
+            # not a constant. In eager mode or non-TF backends, check both dims.
+            if backend.backend() == "tensorflow":
+                from keras.src.utils.module_utils import tensorflow as tf
+
+                if tf.executing_eagerly():
```
hertschuh: I think the check for eager mode vs graph mode came from a misunderstanding of dynamic dimensions, I don't think we need any of that.

I believe you only need the code starting at line 124, meaning you can remove all the Tensorflow specific block.

pctablet505:
@gemini-code-assist review

pctablet505:
@gemini-code-assist review

requirements-tensorflow-cuda.txt
```diff
@@ -13,3 +13,4 @@ torch
 jax[cpu]
 
 -r requirements-common.txt
+psutil
```
hertschuh: Remove this, `psutil` is already installed beforehand separately: https://github.com/keras-team/keras/blob/master/.kokoro/github/ubuntu/gpu/build.sh#L25

PR #22040, opened 2026-01-21
Implement nanmin function in keras.ops

shashaka:
Adds keras.ops.nanmin, which computes the minimum of elements along a specified axis while ignoring NaN values in the input tensor.

### DIFF
--- a/keras/src/backend/openvino/excluded_concrete_tests.txt
+++ b/keras/src/backend/openvino/excluded_concrete_tests.txt
@@ -16,7 +16,6 @@ NumpyDtypeTest::test_inner
 NumpyDtypeTest::test_isreal
 NumpyDtypeTest::test_lcm
 NumpyDtypeTest::test_ldexp
-NumpyDtypeTest::test_logaddexp2
 NumpyDtypeTest::test_nanmin
 NumpyDtypeTest::test_nansum
 NumpyDtypeTest::test_matmul_
@@ -41,7 +40,6 @@ NumpyDtypeTest::test_view
 NumpyDtypeTest::test_vstack
 HistogramTest
 NumpyOneInputOpsCorrectnessTest::test_angle
-NumpyOneInputOpsCorrectnessTest::test_argpartition
 NumpyOneInputOpsCorrectnessTest::test_array
 NumpyOneInputOpsCorrectnessTest::test_hanning
 NumpyOneInputOpsCorrectnessTest::test_kaiser
@@ -52,7 +50,6 @@ NumpyOneInputOpsCorrectnessTest::test_diagflat
 NumpyOneInputOpsCorrectnessTest::test_exp2
 NumpyOneInputOpsCorrectnessTest::test_imag
 NumpyOneInputOpsCorrectnessTest::test_isreal
-NumpyOneInputOpsCorrectnessTest::test_logaddexp2
 NumpyOneInputOpsCorrectnessTest::test_nanmin
 NumpyOneInputOpsCorrectnessTest::test_nansum
 NumpyOneInputOpsCorrectnessTest::test_real
--- a/keras/src/ops/numpy_test.py
+++ b/keras/src/ops/numpy_test.py
@@ -5504,7 +5504,7 @@ class NumpyOneInputOpsCorrectnessTest(testing.TestCase):
         self.assertEqual(out[1].shape, (2, 4))
 
     def test_nanmin(self):
-        x = np.array([[1.0, np.nan, 3.0], [np.nan, 2.0, 1.0]])
+        x = np.array([[1.0, np.nan, 3.0], [np.nan, 2.0, np.inf]])
 
         self.assertAllClose(knp.nanmin(x), np.nanmin(x))
         self.assertAllClose(knp.nanmin(x, axis=()), np.nanmin(x, axis=()))
@@ -5542,7 +5542,7 @@ class NumpyOneInputOpsCorrectnessTest(testing.TestCase):
         )
 
     def test_nansum(self):
-        x = np.array([[1.0, np.nan, 3.0], [np.nan, 2.0, np.inf]])
+        x = np.array([[1.0, np.nan, 3.0], [np.nan, 2.0, 1.0]])
 
         self.assertAllClose(knp.nansum(x), np.nansum(x))
         self.assertAllClose(knp.nansum(x, axis=()), np.nansum(x, axis=()))


keras/src/backend/torch/numpy.py
```diff
@@ -1272,6 +1272,31 @@ def moveaxis(x, source, destination):
...
+
+    out = torch.amin(x_clean, dim=axis, keepdim=keepdims)
+
+    if axis is not None:
+        all_nan = torch.isnan(x).all(dim=axis, keepdim=keepdims)
+        out = torch.where(all_nan, torch.nan, out)
+    elif torch.isnan(x).all():
+        out = torch.tensor(float("nan"), dtype=x.dtype)
+
+    return out
```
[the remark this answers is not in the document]
shashaka: 1. torch.amin does NOT return (values, indices) — it always returns a tensor. Only torch.min does that.
2. torch.nanmin does NOT exist in PyTorch (including 2.9.1).

keras/src/backend/tensorflow/numpy.py
```diff
@@ -2125,6 +2125,35 @@ def moveaxis(x, source, destination):
...
+    out = tf.reduce_min(x_clean, axis=axis, keepdims=keepdims)
+
+    if axis is not None:
+        all_nan = tf.reduce_all(tf.math.is_nan(x), axis=axis, keepdims=keepdims)
+        out = tf.where(all_nan, tf.constant(float("nan"), dtype=x.dtype), out)
+    else:
+        if tf.reduce_all(tf.math.is_nan(x)):
+            out = tf.constant(float("nan"), dtype=x.dtype)
+
+    return out
```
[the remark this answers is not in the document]
shashaka: TensorFlow does not have tf.math.reduce_nanmin, so the implementation cannot be simplified that way.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -2125,6 +2125,35 @@ def moveaxis(x, source, destination):
...
+
+    if not x.dtype.is_floating:
+        dtype = standardize_dtype(x.dtype)
+        if dtype == "bool":
+            return cast(
+                tf.reduce_min(
+                    cast(x, config.floatx()), axis=axis, keepdims=keepdims
+                ),
+                dtype,
+            )
```
hertschuh: I think it would be easier and more efficient to just do:

```python
return tf.reduce_all(x, axis=axis, keepdims=keepdims)
```

keras/src/backend/tensorflow/numpy.py
```diff
@@ -2125,6 +2125,35 @@ def moveaxis(x, source, destination):
...
+    inf = tf.constant(float("inf"), dtype=x.dtype)
+    x_clean = tf.where(tf.math.is_nan(x), inf, x)
+
+    out = tf.reduce_min(x_clean, axis=axis, keepdims=keepdims)
+
+    if axis is not None:
+        all_nan = tf.reduce_all(tf.math.is_nan(x), axis=axis, keepdims=keepdims)
+        out = tf.where(all_nan, tf.constant(float("nan"), dtype=x.dtype), out)
+    else:
+        if tf.reduce_all(tf.math.is_nan(x)):
```
hertschuh: You cannot do an `if` on a tensor result, it won't be compilable (we turn off Autograph).

You'll need to use a `tf.where` or `tf.cond` here.

keras/src/backend/torch/numpy.py
```diff
@@ -1272,6 +1272,31 @@ def moveaxis(x, source, destination):
...
+    x = convert_to_tensor(x)
+    if not torch.is_floating_point(x):
+        return torch.amin(x, dim=axis, keepdim=keepdims)
+
+    if axis == () or axis == []:
+        return x
+
+    x_clean = torch.where(
+        torch.isnan(x),
+        torch.full((), float("inf"), device=x.device, dtype=x.dtype),
```
hertschuh: You don't need `torch.full` here, you can just have a scalar tensor for `float("inf")`.

keras/src/backend/torch/numpy.py
```diff
@@ -1272,6 +1272,31 @@ def moveaxis(x, source, destination):
...
+        x,
+    )
+
+    out = torch.amin(x_clean, dim=axis, keepdim=keepdims)
+
+    if axis is not None:
+        all_nan = torch.isnan(x).all(dim=axis, keepdim=keepdims)
+        out = torch.where(all_nan, torch.nan, out)
+    elif torch.isnan(x).all():
```
hertschuh: Similar comment here about `if` with a tensor. While it should typically work on torch, I don't know what happens if you try to use a compiler.

keras/src/ops/numpy_test.py
```diff
@@ -5473,6 +5503,45 @@ def test_slogdet(self):
         self.assertEqual(out[0].shape, ())
         self.assertEqual(out[1].shape, (2, 4))
 
+    def test_nanmin(self):
+        x = np.array([[1.0, np.nan, 3.0], [np.nan, 2.0, 1.0]])
```
hertschuh: Can you change it to 
```
x = np.array([[1.0, np.nan, 3.0], [np.nan, 2.0, np.inf]])
```
to exercise the infinite case too?

keras/src/backend/tensorflow/numpy.py
shashaka: Thanks, I’ve updated the code.

keras/src/backend/tensorflow/numpy.py
shashaka: I’ve updated the code.

keras/src/backend/torch/numpy.py
shashaka: Thanks, I’ve updated it to use a scalar tensor for float("inf") instead of torch.full.

keras/src/backend/torch/numpy.py
shashaka: I've updated.

keras/src/ops/numpy_test.py
shashaka: I've applied.

keras/src/backend/torch/numpy.py
```diff
@@ -1272,6 +1272,24 @@ def moveaxis(x, source, destination):
     return torch.moveaxis(x, source=source, destination=destination)
 
 
+def nanmin(x, axis=None, keepdims=False):
+    x = convert_to_tensor(x)
+    if not torch.is_floating_point(x):
+        return torch.amin(x, dim=axis, keepdim=keepdims)
+
+    if axis == () or axis == []:
+        return x
```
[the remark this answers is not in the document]
shashaka: To align its behavior with NumPy, I added this code.

keras/src/ops/numpy_test.py
```diff
@@ -5473,9 +5503,47 @@ def test_slogdet(self):
...
+            ]
+        )
+        self.assertAllClose(knp.nanmin(x_3d), np.nanmin(x_3d))
+        self.assertAllClose(
+            knp.nanmin(x_3d, axis=(1, 2)),
+            np.nanmin(x_3d, axis=(1, 2)),
+        )
+
+    def test_nansum(self):
+        x = np.array([[1.0, np.nan, 3.0], [np.nan, 2.0, np.inf]])
```
hertschuh: Oh, you changed `test_nansum` instead of `nanmin`.

keras/src/backend/openvino/excluded_concrete_tests.txt
```diff
@@ -16,6 +16,8 @@ NumpyDtypeTest::test_inner
 NumpyDtypeTest::test_isreal
 NumpyDtypeTest::test_lcm
 NumpyDtypeTest::test_ldexp
+NumpyDtypeTest::test_logaddexp2
```
hertschuh: Remove

keras/src/backend/openvino/excluded_concrete_tests.txt
```diff
@@ -39,6 +41,7 @@ NumpyDtypeTest::test_view
 NumpyDtypeTest::test_vstack
 HistogramTest
 NumpyOneInputOpsCorrectnessTest::test_angle
+NumpyOneInputOpsCorrectnessTest::test_argpartition
```
hertschuh: Remove

keras/src/backend/openvino/excluded_concrete_tests.txt
```diff
@@ -49,6 +52,8 @@ NumpyOneInputOpsCorrectnessTest::test_diagflat
 NumpyOneInputOpsCorrectnessTest::test_exp2
 NumpyOneInputOpsCorrectnessTest::test_imag
 NumpyOneInputOpsCorrectnessTest::test_isreal
+NumpyOneInputOpsCorrectnessTest::test_logaddexp2
```
hertschuh: Remove

shashaka: Removed.

keras/src/backend/openvino/excluded_concrete_tests.txt
shashaka: Removed.

keras/src/backend/openvino/excluded_concrete_tests.txt
shashaka: Removed.

keras/src/ops/numpy_test.py
shashaka: Thanks for catching that. I corrected.

hertschuh (approved):

PR #22041, opened 2026-01-22
Torch dynamic shapes

MarcosAsh:
## Description

Implements dynamic shape support for the torch backend export, addressing the TODO in keras/src/backend/torch/export.py.

Previously, exporting models with the torch backend would bake in fixed shapes, causing tests with different batch sizes to fail.

## Changes

- Modified track_and_add_endpoint in keras/src/backend/torch/export.py to:
  - Build dynamic_shapes from input_signature where shape has None
  - Use torch.export.Dim to mark dynamic dimensions
  - Changed replace_none_number from 1 to 2 to avoid torch dimension specialization
- Removed early returns in tests that were skipping dynamic shape verification:
  - keras/src/export/saved_model_test.py
  - keras/src/export/onnx_test.py

### DIFF
--- a/keras/src/backend/torch/export.py
+++ b/keras/src/backend/torch/export.py
@@ -36,15 +36,31 @@ class TorchExportArchive:
             )
 
         sample_inputs = tree.map_structure(
-            lambda x: convert_spec_to_tensor(x, replace_none_number=1),
+            lambda x: convert_spec_to_tensor(x, replace_none_number=2),
             input_signature,
         )
         sample_inputs = tuple(sample_inputs)
 
+        # Build dynamic_shapes from input_signature where shape has None
+        # Use a shared "batch" dim for dimension 0 across all inputs
+        batch_dim = torch.export.Dim("batch", min=1)
+        dynamic_shapes = []
+        for spec in input_signature:
+            dim_spec = {}
+            for dim_idx, dim_val in enumerate(spec.shape):
+                if dim_val is None:
+                    if dim_idx == 0:
+                        dim_spec[dim_idx] = batch_dim
+                    else:
+                        dim_spec[dim_idx] = torch.export.Dim(
+                            f"dim_{len(dynamic_shapes)}_{dim_idx}", min=1
+                        )
+            dynamic_shapes.append(dim_spec if dim_spec else None)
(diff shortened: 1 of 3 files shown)


keras/src/backend/torch/export.py
```diff
@@ -36,15 +36,30 @@ def track_and_add_endpoint(self, name, resource, input_signature, **kwargs):
...
+        # Build dynamic_shapes from input_signature where shape has None
+        dynamic_shapes = []
+        for spec_idx, spec in enumerate(input_signature):
+            shape = getattr(spec, "shape", None)
+            if shape is not None:
+                dim_spec = {}
+                for dim_idx, dim_val in enumerate(shape):
+                    if dim_val is None:
+                        dim_spec[dim_idx] = torch.export.Dim(
+                            f"dim_{spec_idx}_{dim_idx}", min=1
```
hertschuh: I think dimension 0 should be shared between the various inputs and should be named `batch`. Otherwise you may run into issues.

https://github.com/keras-team/keras/blob/master/keras/src/backend/jax/export.py#L158-L159

keras/src/backend/torch/export.py
```diff
@@ -36,15 +36,30 @@ def track_and_add_endpoint(self, name, resource, input_signature, **kwargs):
...
-            lambda x: convert_spec_to_tensor(x, replace_none_number=1),
+            lambda x: convert_spec_to_tensor(x, replace_none_number=2),
             input_signature,
         )
         sample_inputs = tuple(sample_inputs)
 
+        # Build dynamic_shapes from input_signature where shape has None
+        dynamic_shapes = []
+        for spec_idx, spec in enumerate(input_signature):
+            shape = getattr(spec, "shape", None)
```
hertschuh: You can assume that `spec.shape` is defined, no need for `getattr`.

keras/src/backend/torch/export.py
```diff
@@ -36,15 +36,30 @@ def track_and_add_endpoint(self, name, resource, input_signature, **kwargs):
...
+            if shape is not None:
```
hertschuh: You can assume that `shape` is not `None`, no need for an if

hertschuh (commented):
Thanks for fixing this gap!

MarcosAsh:
Hey just did the changes! thanks for the feedback

PR #22047, opened 2026-01-23
Add Batch Renormalisation

abheesht17:
Verified outputs with the Keras 2 implementation: https://colab.research.google.com/gist/abheesht17/f43fd12378d6abb9691d7ed9ab4938a6/batch-renorm-verify-keras-3-implementation.ipynb.

Reference: https://arxiv.org/abs/1702.03275.

### DIFF
--- a/keras/src/layers/normalization/batch_normalization.py
+++ b/keras/src/layers/normalization/batch_normalization.py
@@ -280,11 +280,8 @@ class BatchNormalization(Layer):
             # paper. The stddev is initialized as sqrt of the variance
             # initializer.
             def moving_stddev_initializer(shape, dtype=None):
-                return ops.sqrt(
-                    _clone_initializer(
-                        self.moving_variance_initializer(shape, dtype=dtype)
-                    )
-                )
+                cloned = _clone_initializer(self.moving_variance_initializer)
+                return ops.sqrt(cloned(shape, dtype=dtype))
 
             self.moving_stddev = self.add_weight(
                 shape=shape,


keras/src/layers/normalization/batch_normalization_test.py
```diff
@@ -239,3 +241,198 @@ def test_masked_broadcast_normalization(self):
...
+        actual_output_np = backend.convert_to_numpy(actual_output)
+        self.assertAllClose(actual_output_np, expected_output, atol=1e-5)
+
+        # Verify moving statistics.
+        expected_renorm_mean = (
+            init_renorm_mean * renorm_momentum
+            + batch_mean * (1 - renorm_momentum)
+        )
+        self.assertAllClose(
+            backend.convert_to_numpy(layer.renorm_mean),
```
hertschuh: nitpick: `assertAllClose` does `convert_to_numpy` for you, so you don't need that.

Same lines 402, 410, 418, 424

keras/src/layers/normalization/batch_normalization.py
```diff
@@ -208,6 +254,41 @@ def build(self, input_shape):
...
         )
 
+        if self.renorm:
+            # In batch renormalization we track the inference moving stddev
+            # instead of the moving variance to more closely align with the
+            # paper. The stddev is initialized as sqrt of the variance
+            # initializer.
+            def moving_stddev_initializer(shape, dtype=None):
+                return ops.sqrt(
+                    self.moving_variance_initializer(shape, dtype=dtype)
```
hertschuh: Don't we have the issue that we cannot reuse the initializers, we have to clone them? (same issue lines 270, 280 and 287)

abheesht17: Ah, right. We did that for KerasRS. Fixed

keras/src/layers/normalization/batch_normalization_test.py
abheesht17: My bad

keras/src/layers/normalization/batch_normalization.py
```diff
@@ -208,6 +274,43 @@ def build(self, input_shape):
...
+        if self.renorm:
+            # In batch renormalization we track the inference moving stddev
+            # instead of the moving variance to more closely align with the
+            # paper. The stddev is initialized as sqrt of the variance
+            # initializer.
+            def moving_stddev_initializer(shape, dtype=None):
+                return ops.sqrt(
+                    _clone_initializer(
+                        self.moving_variance_initializer(shape, dtype=dtype)
+                    )
```
hertschuh: You're trying to clone the resulting tensor, not the initializer (which will be a no-op).

```python
            def moving_stddev_initializer(shape, dtype=None):
                return ops.sqrt(
                    _clone_initializer(self.moving_variance_initializer)(shape, dtype=dtype)
                )
```

Although this would probably be more readable:

```python
            def moving_stddev_initializer(shape, dtype=None):
                cloned = _clone_initializer(self.moving_variance_initializer)
                return ops.sqrt(cloned(shape, dtype=dtype))
```

Issue #22038, opened 2026-01-20
Add `euclidean_dist_transform`


innat:
This methods becomes very crucial in certian cases, i.e. loss calculation etc. 

- [Euclidean distance transform in tensorflow](https://stackoverflow.com/questions/47140842/euclidean-distance-transform-in-tensorflow)
- [tf-addons adaption](https://github.com/tensorflow/addons/blob/d208d752e98c310280938efa939117bf635a60a8/tensorflow_addons/image/distance_transform.py#L29)
- `from scipy.ndimage import distance_transform_edt`
- https://github.com/tensorflow/tensorflow/issues/24410

IMvision12:
I can work on this!

innat:
@IMvision12 Thansk for your interest. Please note, there might be some complication, i.e. gpu support or gradient etc. 

This method is typically used to transform a binary ground-truth segmentation mask. In this case, it must be implemented either in pure TensorFlow or using the Keras Ops API. However, if implemented with the Keras Ops API and later used within a `tf.data` pipeline running on a non-tf backend, it may lead to compatibility issues. Conversely, if implemented in pure TensorFlow and later used inside a `keras.loss` class with a non-tf backend, similar problems can arise. Generally, it is generally recommended to apply this transformation at the data-loading stage. https://github.com/keras-team/keras/issues/21785 cc. @mattdangerw

IMvision12:
Hi @innat,

I've implemented the Euclidean Distance Transform using the Felzenszwalb-Huttenlocher algorithm across all backends (TensorFlow, JAX, PyTorch, and NumPy). Here's my approach:

Implements the exact linear-time EDT algorithm (same results as scipy.ndimage.distance_transform_edt)
Supports rank 2, 3, and 4 tensors with both channels_first and channels_last formats
Handles batched processing with per-channel distance computation

The PyTorch and JAX implementations natively support GPU execution. The TensorFlow implementation uses NumPy for computation (similar to preprocessing operations), which is appropriate since EDT is typically applied during data loading rather than in training loops.

PR #22049, opened 2026-01-23
Add `euclidean_dist_transform`

IMvision12:
#22038

### DIFF
--- a/keras/api/ops/image/__init__.py
+++ b/keras/api/ops/image/__init__.py
@@ -7,6 +7,9 @@ since your modifications would be overwritten.
 from keras.src.ops.image import affine_transform as affine_transform
 from keras.src.ops.image import crop_images as crop_images
 from keras.src.ops.image import elastic_transform as elastic_transform
+from keras.src.ops.image import (
+    euclidean_dist_transform as euclidean_dist_transform,
+)
 from keras.src.ops.image import extract_patches as extract_patches
 from keras.src.ops.image import extract_patches_3d as extract_patches_3d
 from keras.src.ops.image import gaussian_blur as gaussian_blur


innat:
@IMvision12 Thanks for the PR. Would it be possible to create a collab gist, showing comparision between scipy dist transform and this?

IMvision12:
Hi @innat here's the notebook : https://colab.research.google.com/drive/1WCb656Ojz7I3j0w1uALuBlHnL3n-Qt0n?usp=sharing

But this implementation is very slow!! because of python for loops, scipy's implementation is c/c++ compiled so its faster

innat:
@IMvision12 Thanks for your effort. This is indeed a good implementation. Howerver, the performance as slow as shown is quite problematic. While googling, I found that, others use `from scipy.ndimage import distance_transform_edt` and wrap ot to GPU supported wrapper. For example:

1. MONAI provides a wrapper for `scipy.ndimage.distance_transform_edt` that handles tensor conversions and supports GPU acceleration via `cuCIM`. 

```python
import torch
from monai.transforms import DistanceTransformEDT

# 1 represents foreground, 0 represents background
mask = torch.randint(low=0, high=2, size=(1, 1, 64, 64, 64)).float().cuda()

# Initialize transform
edt_transform = DistanceTransformEDT(sigmoid=False) 
dist_map = edt_transform(mask)
````

2. CuPy to handle the scipy.ndimage call directly on GPU is faster than moving data back and forth to CPU. 

```python
import torch
import cupy as cp
from scipy.ndimage import distance_transform_edt

# Move torch tensor to numpy, then to cupy
mask_gpu = cp.array(torch_mask.detach().cpu().numpy())
dist_map_gpu = distance_transform_edt(mask_gpu)
# Convert back to torch
dist_map_torch = torch.from_numpy(cp.asnumpy(dist_map_gpu)).cuda()
```

Will it be doable and acceptable in Keras? cc. @fchollet

Issue #18428, opened 2023-07-28
no ssim metrics for images


anas-rz:
Some networks [e.g.](https://keras.io/examples/vision/basnet_segmentation/) require computing SSIM between images. Keras core doesn't have cross-backend ops for this. Can I open a PR for this?

dhantule:
@anas-rz, thanks for reporting this. Is this issue still relevant to you ?

PR #22053, opened 2026-01-24
Add SSIM  to keras.ops.image

MarcosAsh:
## Description
Adds keras.ops.image.ssim to compute the Structural Similarity Index (SSIM) between images, addressing #18428.
## Changes

Added ssim function to keras.ops.image
Uses Gaussian filter (11x11, sigma=1.5) matching TensorFlow's implementation
Supports both channels_last and channels_first data formats
Added tests for dynamic shapes and correctness against tf.image.ssim


MarcosAsh:
I think the reason its not passing the tests is a flaky test issue. But please tell me if its an actual issue with my code.

PR #22054, opened 2026-01-24
Fix gaussian_blur padding calculation for even kernel sizes

### LINKED ISSUE #22046: keras.src.backend.numpy.image.gaussian_blur crashes on even-sized images
### Description

`gaussian_blur` throws a `ValueError` when the input image has even spatial dimensions. The error seems to come from a shape mismatch between the output of `scipy.signal.convolve2d` and the expected blurred_images array.
This does not happen if the same function is called through the JAX backend (`keras.src.backend.numpy.jax.image.gaussian_blur`), so the issue appears to be specific to the NumPy backend.

Please use this example to replicate the problem:
```
import numpy as np
from keras.src.backend.numpy.image import gaussian_blur

rng = np.random.RandomState(12345)
img = rng.uniform(size=(2, 3, 32, 32)).astype(np.float64)

gaussian_blur(
    img,
    sigma=(0.8, 1.2),
    kernel_size=(4, 4),
    data_format="channels_first",
)
```
Error Message:
```
  File "test.py", line 9, in <module>
    gaussian_blur(img, sigma=(0.8, 1.2), kernel_size=(4, 4), data_format="channels_first")
  File "miniconda3/envs/keras/lib/python3.11/site-packages/keras/src/backend/numpy/image.py", line 1033, in gaussian_blur
    blurred_images[b, :, :, ch] = scipy.signal.convolve2d(
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^
ValueError: could not broadcast input array from shape (33,33) into shape (32,32)

```
This looks like a regression introduced in gh-20974. The older implementation from gh-20943 works fine with these images and does not raise this error.

### Environment
- Keras: 3.13.1 (also tested with 3.10.0)
- Python: 3.11 
- OS: macOS

SamareshSingh:
Fixes #22046

## Description
This PR fixes a bug in the NumPy backend's gaussian_blur function where it crashes when processing images with even-sized kernel dimensions.

The gaussian_blur function was using incorrect symmetric padding that caused shape mismatches when convolving with even-sized kernels.

### DIFF
--- a/keras/src/ops/image_test.py
+++ b/keras/src/ops/image_test.py
@@ -755,18 +755,14 @@ def gaussian_blur_np(
     )
 
     blurred_images = np.zeros_like(images)
-    kernel_reshaped = kernel.reshape(
-        (1, kernel_h, kernel_w, num_channels)
-    )
+    kernel_reshaped = kernel.reshape((1, kernel_h, kernel_w, num_channels))
 
     for b in range(batch_size):
         image_patch = padded_images[b : b + 1, :, :, :]
 
     for i in range(height):
         for j in range(width):
-            patch = image_patch[
-                :, i : i + kernel_h, j : j + kernel_w, :
-            ]
+            patch = image_patch[:, i : i + kernel_h, j : j + kernel_w, :]
             blurred_images[b, i, j, :] = np.sum(
                 patch * kernel_reshaped, axis=(1, 2)
             )


ssam18:
hmm. need to check whey 4 tests are failing.

ssam18:
Do I need to push the regenerated keras/api directory as api_ben hook is failing due to modified files?

hertschuh:
> Do I need to push the regenerated keras/api directory as api_ben hook is failing due to modified files?

This indicates that something is off, try `pip install -U ruff`.

The code formatting check fails, meaning you need to run `./shell/api_gen.sh`.

SamareshSingh:
image

SamareshSingh:
it ran all good.

hertschuh:
> it ran all good.

Have you done `pre-commit install` in the past?

Then, to format the code, you can run `pre-commit run --all-files`

SamareshSingh:
yes, I did that on this machine but not in this branch. Just now I ran locally and didn't observe any issue. See below screenshot 
image

hertschuh:
Thank you for the fix!

PR #22055, opened 2026-01-25
Implement nanmean function in keras.ops

shashaka:
Adds keras.ops.nanmean, which computes the mean of elements along a specified axis while ignoring NaN values in the input tensor.

### DIFF
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -2151,29 +2151,19 @@ def nanmean(x, axis=None, keepdims=False):
     if axis == () or axis == []:
         return x
 
+    if not x.dtype.is_floating:
+        return tf.reduce_mean(
+            tf.cast(x, "float32"), axis=axis, keepdims=keepdims
+        )
+
     dtype = dtypes.result_type(standardize_dtype(x.dtype), float)
-
-    nan_sum = cast(nansum(x, axis=axis, keepdims=keepdims), dtype)
-
-    is_nan_mask = (
-        tf.math.is_nan(x)
-        if x.dtype.is_floating
-        else tf.zeros_like(x, dtype=tf.bool)
-    )
-
+    total_sum = cast(nansum(x, axis=axis, keepdims=keepdims), dtype)
     normalizer = tf.reduce_sum(
-        cast(~is_nan_mask, dtype),
+        cast(~tf.math.is_nan(x), dtype),
         axis=axis,
         keepdims=keepdims,
     )
-
-    out = tf.math.divide_no_nan(nan_sum, normalizer)
-
-    return tf.where(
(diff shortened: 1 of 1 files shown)


keras/src/backend/tensorflow/numpy.py
```diff
@@ -2145,6 +2145,30 @@ def nanmax(x, axis=None, keepdims=False):
...
+        return x
+
+    dtype = dtypes.result_type(standardize_dtype(x.dtype), float)
+
+    nan_sum = cast(nansum(x, axis=axis, keepdims=keepdims), dtype)
+    normalizer = tf.reduce_sum(
+        cast(~tf.math.is_nan(cast(x, config.floatx())), dtype),
+        axis=axis,
+        keepdims=keepdims,
+    )
```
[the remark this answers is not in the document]
shashaka: Updated

keras/src/backend/tensorflow/numpy.py
```diff
@@ -2145,6 +2145,37 @@ def nanmax(x, axis=None, keepdims=False):
...
+    if axis == () or axis == []:
+        return x
+
+    dtype = dtypes.result_type(standardize_dtype(x.dtype), float)
+
+    nan_sum = cast(nansum(x, axis=axis, keepdims=keepdims), dtype)
+
+    is_nan_mask = (
+        tf.math.is_nan(x)
+        if x.dtype.is_floating
```
hertschuh: I think you should just do what you did for the other ones by doing this line 2155:
```python
if not x.dtype.is_floating:
    return tf.reduce_mean(tf.cast(x, "float32"),  axis=axis, keepdims=keepdims)
```

keras/src/backend/tensorflow/numpy.py
```diff
@@ -2145,6 +2145,37 @@ def nanmax(x, axis=None, keepdims=False):
...
+        keepdims=keepdims,
+    )
+
+    out = tf.math.divide_no_nan(nan_sum, normalizer)
+
+    return tf.where(
+        normalizer > 0,
+        out,
+        tf.fill(tf.shape(out), tf.constant(float("nan"), out.dtype)),
+    )
```
hertschuh: Wait, can't you just do:
```
return tf.divide(nan_sum, normalizer)
```
so that you get nan where all the numbers where nans (i.e. where normalizer is zero)?

keras/src/backend/tensorflow/numpy.py
shashaka: Thanks. I updated it.

keras/src/backend/tensorflow/numpy.py
shashaka: Thank you for pointing that out. I’ve updated it to use that directly.

PR #22056, opened 2026-01-25
Increase test coverage for StringLookup preprocessing layer

MalyalaKarthik66:
Addresses the TODO in `string_lookup_test.py` by adding 7 new tests.

New tests:
- `test_invert_lookup_basic`: Basic inversion with OOV (TensorFlow backend).
- `test_output_mode_count_shape`: Shape matches vocabulary size in `count` mode.
- `test_output_mode_multi_hot_binary`: Outputs are binary in `multi_hot` mode.
- `test_mask_token_basic`: Mask token maps to index 0.
- `test_adapt_with_python_iterable`: `adapt()` works on Python iterables and updates vocabulary.
- `test_sparse_output_in_multi_hot`: Sparse `multi_hot` output returns a `SparseTensor` (TF).
- `test_get_vocabulary_include_special_tokens_false`: Excludes special tokens from vocabulary.

### DIFF
--- a/keras/src/layers/preprocessing/string_lookup_test.py
+++ b/keras/src/layers/preprocessing/string_lookup_test.py
@@ -9,7 +9,6 @@ from keras.src import layers
 from keras.src import models
 from keras.src import saving
 from keras.src import testing
-from keras.src.ops import convert_to_numpy
 from keras.src.ops import convert_to_tensor
 
 
@@ -169,22 +168,17 @@ class StringLookupTest(testing.TestCase):
             invert=True,
         )
         output = layer([1, 2, 0])
-
-        if backend.is_tensor(output):
-            output = convert_to_numpy(output)
-            output = [
-                x.decode("utf-8") if isinstance(x, (bytes, bytearray)) else x
-                for x in output
-            ]
-
-        self.assertEqual(output, ["a", "b", "[UNK]"])
+        self.assertAllEqual(
+            backend.convert_to_numpy(output).astype(str),
+            ["a", "b", "[UNK]"],
+        )
 
     def test_output_mode_count_shape(self):
         layer = layers.StringLookup(
             vocabulary=["a", "b"],
             output_mode="count",
         )
-        output = layer(["a", "b", "a"])
+        output = layer(["a", "a", "a", "b", "b"])
         self.assertEqual(output.shape[-1], len(layer.get_vocabulary()))
 
     def test_output_mode_multi_hot_binary(self):
@@ -193,11 +187,7 @@ class StringLookupTest(testing.TestCase):
             output_mode="multi_hot",
         )
         output = layer(["a", "b"])
-
-        if backend.is_tensor(output):
-            output = convert_to_numpy(output)
-
-        self.assertTrue(np.all((output == 0) | (output == 1)))
+        self.assertAllClose(output, [0, 1, 1])
 
     def test_mask_token_basic(self):
         layer = layers.StringLookup(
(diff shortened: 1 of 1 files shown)


keras/src/layers/preprocessing/string_lookup_test.py
```diff
@@ -157,3 +158,88 @@ def test_torch_backend_compatibility(self):
...
+        reason="invert=True requires TensorFlow string tensors",
+    )
+    def test_invert_lookup_basic(self):
+        layer = layers.StringLookup(
+            vocabulary=["a", "b", "c"],
+            invert=True,
+        )
+        output = layer([1, 2, 0])
+
+        if backend.is_tensor(output):
```
hertschuh: You can skip the `if`, the output is going to be a backend tensor.

keras/src/layers/preprocessing/string_lookup_test.py
```diff
@@ -157,3 +158,88 @@ def test_torch_backend_compatibility(self):
...
+
+        self.assertEqual(output, ["a", "b", "[UNK]"])
+
+    def test_output_mode_count_shape(self):
+        layer = layers.StringLookup(
+            vocabulary=["a", "b"],
+            output_mode="count",
+        )
+        output = layer(["a", "b", "a"])
+        self.assertEqual(output.shape[-1], len(layer.get_vocabulary()))
```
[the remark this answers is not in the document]
hertschuh: I kind of agree here.

Moreover, as it happens, the vocabulary size is 3 and in the input size is also 3. Pick a different number for the input size so that you know for sure it's working.

keras/src/layers/preprocessing/string_lookup_test.py
```diff
@@ -157,3 +158,88 @@ def test_torch_backend_compatibility(self):
...
+        layer = layers.StringLookup(
+            vocabulary=["a", "b"],
+            output_mode="multi_hot",
+        )
+        output = layer(["a", "b"])
+
+        if backend.is_tensor(output):
+            output = convert_to_numpy(output)
+
+        self.assertTrue(np.all((output == 0) | (output == 1)))
```
hertschuh: Replace all of this with:
```python
self.assertAllClose(output, [0, 1, 1])
```
It will also do the conversion to numpy for you.

keras/src/layers/preprocessing/string_lookup_test.py
```diff
@@ -157,3 +158,88 @@ def test_torch_backend_compatibility(self):
...
+        layer = layers.StringLookup(
+            vocabulary=["a"],
+            mask_token="[MASK]",
+        )
+        output = layer(["[MASK]", "a"])
+
+        if backend.is_tensor(output):
+            output = convert_to_numpy(output)
+
+        self.assertEqual(int(output[0]), 0)
```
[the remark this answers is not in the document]
hertschuh: +1 and remove the convert_to_numpy.

keras/src/layers/preprocessing/string_lookup_test.py
```diff
@@ -157,3 +158,88 @@ def test_torch_backend_compatibility(self):
...
+
+    def test_adapt_with_python_iterable(self):
+        layer = layers.StringLookup()
+        layer.adapt(["a", "a", "b", "c"])
+        vocab = layer.get_vocabulary()
+
+        for token in ["a", "b", "c"]:
+            self.assertIn(token, vocab)
```
hertschuh: This test doesn't cover anything that `def test_adapt_flow(self):` didn't already cover. I'd say remove it.

MalyalaKarthik66:
@hertschuh Thanks for the review! Tests have been updated accordingly.

PR #22064, opened 2026-01-27
Fix arctan2 NaN propagation in OpenVINO backend

### LINKED ISSUE #22061: `layers.TimeDistributed` produce non-nan results for nan inputs with openvino backend
### Description

With `x2` is nan and `openvino` backend, `arctan2` still returns non-nan results though it should be `nan` (the results are nan with other backend and `np.arctan2`). 
Please use this example to replicate the problem:

```
import os


from keras.src.ops import numpy as knp
import numpy as np


out1 = knp.arctan2(1.0, np.nan)
print(out1.numpy())
# 1.5703125


out2 = np.arctan2(1.0, np.nan)
print(out2)
# NaN 
```

### Environment
- Keras: 3.13.1 
- Python: 3.11 
- OS: macOS

MalyalaKarthik66:
This PR fixes incorrect NaN propagation in the OpenVINO backend for arctan2.

Fixes #22061

### DIFF
--- a/keras/src/ops/numpy_test.py
+++ b/keras/src/ops/numpy_test.py
@@ -2888,6 +2888,11 @@ class NumpyTwoInputOpsCorrectnessTest(testing.TestCase):
         self.assertAllClose(knp.arctan2(m, n), np.arctan2(m, n))
         self.assertAllClose(knp.Arctan2()(m, n), np.arctan2(m, n))
 
+        x = np.array([1.0, 2.0, np.nan])
+        y = np.array([3.0, np.nan, 4.0])
+        self.assertAllClose(knp.arctan2(x, y), np.arctan2(x, y))
+        self.assertAllClose(knp.Arctan2()(x, y), np.arctan2(x, y))
+
     def test_bitwise_and(self):
         x = np.array([2, 5, 255])
         y = np.array([3, 14, 16])
@@ -3821,11 +3826,6 @@ class NumpyTwoInputOpsCorrectnessTest(testing.TestCase):
         self.assertTrue(
             standardize_dtype(knp.Digitize()(x, bins).dtype) == "int32"
         )
-
-    def test_arctan2_nan_propagation(self):
-        out = knp.arctan2(1.0, np.nan)
-        out_np = np.asarray(out)
-        self.assertTrue(np.isnan(out_np))
 
 
 class NumpyOneInputOpsCorrectnessTest(testing.TestCase):


keras/src/backend/openvino/numpy.py
```diff
@@ -393,7 +397,11 @@ def arctan2(x1, x2):
 
     out_not_pos = ov_opset.select(cond_x2_lt0, out_x2_lt0, out_x2_zero)
 
-    final_out = ov_opset.select(cond_x2_gt0, y, out_not_pos)
+    value_out = ov_opset.select(cond_x2_gt0, y, out_not_pos)
+
+    # Generate NaN safely for all floating dtypes (including bf16)
+    nan_value = ov_opset.divide(zero_const, zero_const)
```
hertschuh, line 403: I think you can just do `ov_opset.constant(np.nan, result_type)`. Does that not work for bfloat16?

keras/src/ops/numpy_test.py
```diff
@@ -3792,6 +3792,11 @@ def test_digitize(self):
             standardize_dtype(knp.Digitize()(x, bins).dtype) == "int32"
         )
 
+    def test_arctan2_nan_propagation(self):
```
hertschuh: Please add this to the existing `test_arctan2` test:
https://github.com/keras-team/keras/blob/master/keras/src/ops/numpy_test.py#L2889

That is, add more lines to the existing test, don't create a separate test.

Then write it like this for consistency:

```python
        x = np.array([1.0, 2.0, np.nan])
        y = np.array([3.0, np.nan, 4.0])
        self.assertAllClose(knp.arctan2(x, y), np.arctan2(x, y))
        self.assertAllClose(knp.Arctan2()(x, y), np.arctan2(x, y))
```

keras/src/backend/openvino/numpy.py
MalyalaKarthik66: Thanks for the suggestion! I tried using `ov_opset.constant(np.nan, result_type)`, but it fails for `bfloat16` with OpenVINO (`Cannot fill constant data. Values is outside the range`).
Using `0.0 / 0.0` reliably produces NaN across all floating dtypes, including `bf16`, and passes the dtype matrix tests. That’s why I kept the current approach.

keras/src/ops/numpy_test.py
MalyalaKarthik66: Done and updated.

hertschuh (approved):
Thanks!

PR #22067, opened 2026-01-27
Adding Torch support for model parallel and data parallel

buildwithsuhana:
This PR is adds support for running model parallel and data parallel with torch backend. We have added a distribution_lib.py for torch here.

Mapping distribution_lib.py with the JAX and Torch backend: https://docs.google.com/document/d/1LVJn5MrWnxnjk9gPPxv4EMjKsqqV6_LX8P7tpg6Hno8/edit?resourcekey=0-bnpxvYggmL7LqaFdKHKlaw&tab=t.0

Implementation: https://www.kaggle.com/code/buildwithsuhana/model-parallel-data-parallel-for-torch


keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,581 @@
...
+            mesh_dim = device_mesh.shape[axis_idx]
+
+            if tensor.shape[axis_idx] >= mesh_dim:
+                if tensor.shape[axis_idx] % mesh_dim == 0:
+                    shard_size = tensor.shape[axis_idx] // mesh_dim
+                    result = list(torch.split(tensor, shard_size, dim=axis_idx))
+                    return result
+                else:
+                    result = list(torch.chunk(tensor, mesh_dim, dim=axis_idx))
+                    return result
```
hertschuh, line 129: So this splits the tensors, but it doesn't send each shard to a device, they're still all on the same device.

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,581 @@
...
+    )
+
+    sharded_value._distributed_layout = backend_layout
+    sharded_value._full_shape = value.shape
+    sharded_value._sharding_axis = getattr(
+        backend_layout, "axes", [None] * len(value.shape)
+    )[0]
+    sharded_value._is_sharded = True
+
+    return sharded_value
```
hertschuh, line 293: So this returns a single shard of a tensor, not a "sharded tensor" that represents all the shards on all the devices. How does it work afterwards?

hertschuh (commented):
Let's say you're running this with `DataParallel`.

What it looks is happening is that the variable are sharded, each process runs indepdently and each will compute the gradients and update the variables independently.

How do you reconcile the variables between all the processes?

Issue #22065, opened 2026-01-27
keras.layers.Normalization crashes with multi-dim mean and variance


thanhlecongg:
`keras.layers.Normalization` crash with valid multi-dim mean and variance
Please use this example to replicate the problem:
```
import keras
import numpy as np
in1 = keras.Input(shape=(2, 3, 4))
mean = np.array([[0, 1., 2., 3.],
                [4., 5., 6., 7.]]
           )
variance = np.array([[1., 1., 1., 1.],
                    [2., 2., 2., 2.]])


normLayer = keras.layers.Normalization(axis=(1,3), mean=mean, variance=variance)
out1 = normLayer(in1)
print(out1)
```

Error Message:
```
tensorflow.python.framework.errors_impl.InvalidArgumentError: {{function_node __wrapped__BroadcastTo_device_/job:localhost/replica:0/task:0/device:CPU:0}} Unable to broadcast tensor of shape [2,4] to tensor of shape [1,2,1,4] [Op:BroadcastTo]
```
I believe we can reshape the tensor of [2, 4] to tensor of shape [1,2,1,4] given 2x4 = 1x2x1x4. In keras 3.7.0, keras.layers.Normalization worked well with these inputs. However, from keras 3.8.0, this error happens. I found that this may be a regression introduced in gh-20626. I think we can just seperate two cases, one for float, i.e., the case of gh-20626 where we will use `broadcast_to` and one for matrices where we use `reshape` same as previous implementation. If this approach make senses to you, I can make a PR.

Ankitaghavate:
Hi @sachinprasadhs and maintainers — thanks for the quick investigation and for pointing to gh-20626. I’m happy to take this on and open a PR.

Summary of proposed fix
Root cause

The change introduced in gh-20626 uses tf.broadcast_to when preparing mean / variance for the Normalization layer. This fails in cases where mean / variance are provided as lower-rank numeric arrays intended to be reshaped (not strictly broadcasted) into the specified axes.

In these cases, tf.broadcast_to raises an InvalidArgumentError when it cannot match the target shape, even though an equivalent reshape would be valid. This worked correctly prior to 3.8.

Proposal

Restore the previous behavior for the “matrix-like” stats case, while preserving the improvements from gh-20626, by distinguishing between two scenarios:

Reshape case (restore pre-3.8 behavior)
[...]

PR #22068, opened 2026-01-27
Fix Normalization broadcasting for scalar and multidim mean and variance

maitry63:
This PR fixes a regression in `keras.layers.Normalization` where providing multi-dimensional `mean` and `variance` values causes a crash during layer build.

Fixes: https://github.com/keras-team/keras/issues/22065

### DIFF
--- a/keras/src/layers/preprocessing/normalization.py
+++ b/keras/src/layers/preprocessing/normalization.py
@@ -142,13 +142,6 @@ class Normalization(DataLayer):
                     f"mean shape {np.shape(mean)} for axis {self.axis}"
                 )
 
-        self.input_mean = mean
-        self.input_variance = variance
-        self.invert = invert
-        self.supports_masking = True
-        self._build_input_shape = None
-        self.mean = None
-
     def build(self, input_shape):
         if input_shape is None:
             return


keras/src/layers/preprocessing/normalization.py
```diff
@@ -193,11 +193,21 @@ def build(self, input_shape):
...
         else:
             # In the no adapt case, make constant tensors for mean and variance
-            # with proper broadcast shape for use during call.
+
             mean = ops.convert_to_tensor(self.input_mean)
             variance = ops.convert_to_tensor(self.input_variance)
-            mean = ops.broadcast_to(mean, self._broadcast_shape)
-            variance = ops.broadcast_to(variance, self._broadcast_shape)
+
+            # Broadcast scalar mean/variance to the target shape
```
hertschuh: This fix covers 2 cases correctly:
- when `mean` and `variance` are scalars
- when `mean` and `variance` cover all the kept axes, i.e. they have the same shape as the inputs in all the kept dimensions.

However it does not cover the case when `mean` and `variance` are not scalars _and_ a broadcast is needed.

To cover this general case, I think what's needed is (when it's not a scalar) to do an `expand_dims` on the non-kept axes (i.e. `self._reduce_axis`). But it needs to be done from right to left and it needs to stop once you run out of dimensions for `mean` or `variance`. Then you should do the `broadcast_to` to cover the missing dimensions.

Does that make sense?

Also, you'll want to add a unit test for this.

keras/src/layers/preprocessing/normalization_test.py
```diff
@@ -202,3 +203,60 @@ def __getitem__(self, idx):
...
         output = normalizer(sample_input)
         self.assertEqual(output.shape, (1, 32, 32, 3))
+
+    def test_broadcast_non_scalar_middle_axis(self):
+        """
+        Tests mean/variance that are not scalars and require
+        expanding dims on non-kept axes (the 'general case').
+        """
+        # Input shape: (Batch=2, Height=4, Width=4, Channels=3)
+        input_shape = (2, 4, 4, 3)
```
hertschuh: Please use unique dimension values so that we know for sure the axes are working.

i.e. `input_shape = (2, 4, 5, 3)` so that there's no ambiguity between the two `4` values.

keras/src/layers/preprocessing/normalization_test.py
```diff
@@ -202,3 +203,60 @@ def __getitem__(self, idx):
...
+        layer = layers.Normalization(
+            axis=axis, mean=custom_mean, variance=custom_var
+        )
+        layer.build(input_shape)
+
+        # The expected broadcast shape should be (1, 1, 4, 1)
+        expected_shape = (1, 1, 4, 1)
+
+        assert layer.mean.shape == expected_shape
+        assert layer.variance.shape == expected_shape
```
hertschuh: Please use `self.assertEqual(...)` for these.

keras/src/layers/preprocessing/normalization_test.py
```diff
@@ -202,3 +203,60 @@ def __getitem__(self, idx):
...
+        assert layer.mean[0, 0, 1, 0] == 20.0
```
hertschuh: please use `self.assertAllClose(layer.mean, [20.0, ...])` for this one.

keras/src/layers/preprocessing/normalization_test.py
```diff
@@ -202,3 +203,60 @@ def __getitem__(self, idx):
...
+
+    def test_broadcast_multiple_axes(self):
+        """
+        Tests keeping multiple axes, e.g., (Height, Width) but not Channels.
+        """
+        input_shape = (None, 10, 10, 3)
```
hertschuh: Please use unique dimension values so that we know for sure the axes are working. There is ambiguity on the `10` right now.

keras/src/layers/preprocessing/normalization_test.py
```diff
@@ -202,3 +203,60 @@ def __getitem__(self, idx):
...
+        custom_mean = np.zeros((10, 10), dtype="float32")
+        custom_var = np.ones((10, 10), dtype="float32")
+
+        layer = layers.Normalization(
+            axis=axis, mean=custom_mean, variance=custom_var
+        )
+        layer.build(input_shape)
+
+        # The expected broadcast shape should be (1, 10, 10, 1)
+        assert layer.mean.shape == (1, 10, 10, 1)
```
hertschuh: `self.assertEqual`

keras/src/layers/preprocessing/normalization_test.py
```diff
@@ -202,3 +203,60 @@ def __getitem__(self, idx):
...
+
+    def test_scalar_broadcast(self):
+        """
+        Ensures the scalar case still broadcasts to the full rank.
+        """
+        input_shape = (2, 2)
```
hertschuh: Please use unique dimension values so that we know for sure the axes are working. There is ambiguity on the `2` right now.

keras/src/layers/preprocessing/normalization_test.py
```diff
@@ -202,3 +203,60 @@ def __getitem__(self, idx):
...
+        layer = layers.Normalization(axis=-1, mean=5.0, variance=1.0)
+        layer.build(input_shape)
+
+        # The expected broadcast shape should be (1, 2)
+        assert layer.mean.shape == (1, 2)
```
hertschuh: `self.assertEqual`

keras/src/layers/preprocessing/normalization_test.py
```diff
@@ -202,3 +203,60 @@ def __getitem__(self, idx):
...
+
+        mean_val = ops.convert_to_numpy(layer.mean)
+        assert np.all(mean_val == 5.0)
```
hertschuh: These 2 lines can be replaced with:
```python
self.assertAllClose(layer.mean, [[5.0, 5.0]])
```

keras/src/layers/preprocessing/normalization_test.py
```diff
@@ -202,3 +203,60 @@ def __getitem__(self, idx):
         sample_input = np.random.rand(1, 32, 32, 3)
         output = normalizer(sample_input)
         self.assertEqual(output.shape, (1, 32, 32, 3))
+
+    def test_broadcast_non_scalar_middle_axis(self):
```
hertschuh, line 206: I think you need to add a test where:

- axis is a tuple of 2 values
- but mean only has 1 dimension meaning it will cover the last of the 2 axes but will be broadcast in the other axis. The input should have a different size in these 2 axes to make sure the broadcast is done correctly.

keras/src/layers/preprocessing/normalization.py
```diff
@@ -191,15 +191,34 @@ def build(self, input_shape):
...
+                # Case 1: Scalar mean/variance
+                mean = ops.broadcast_to(mean, self._broadcast_shape)
+                variance = ops.broadcast_to(variance, self._broadcast_shape)
+            else:
+                # Case 2: General broadcasting (handles partial dimensions).
+                # We need to expand mean/variance so they have 1s on the
+                # _reduce_axis and align with the rank of the input_shape.
+                expanded_shape = [1] * ndim
+                mean_shape = ops.shape(mean)
+                for i, axis_idx in enumerate(self._keep_axis):
```
hertschuh: This makes the assumption that `self._keep_axis` (i.e. `axis`) has exactly the same number of values as `mean` has dimensions.

However, in theory, `mean` can have fewer dimensions than there are axes, in which case it should be broadcast. In fact, you need a new unit test for this case.

Two things:
- if `mean` has more dimensions than there are axes in `axis` or `self._keep_axis`, it should raise an error saying that much, in fact, you can make this verification in `__init__`, it's better to fail early.
- otherwise, this for loop needs to change to:
    - go in reverse order of `self._keep_axis` and `mean_shape`
    - stop when the dimensions of `mean_shape` are exhausted

keras/src/layers/preprocessing/normalization.py
```diff
@@ -191,15 +191,34 @@ def build(self, input_shape):
...
+                    expanded_shape[axis_idx] = mean_shape[i]
+
+                mean = ops.reshape(mean, expanded_shape)
+                variance = ops.reshape(variance, expanded_shape)
```
hertschuh, line 233: This makes the assumption that `mean` and `variance` have the same shape.
This is fine, but it should be verified and raise a `ValueError` if they don't. In fact, you can make this verification in `__init__`, it's better to fail early.

I understand that this was missing before.

keras/src/layers/preprocessing/normalization.py
```diff
@@ -125,6 +125,29 @@ def __init__(
...
+                    f"number of axes ({len(self.axis)}). Received: "
+                    f"mean shape {np.shape(mean)} for axis {self.axis}"
+                )
+
+        self.input_mean = mean
+        self.input_variance = variance
+        self.invert = invert
+        self.supports_masking = True
+        self._build_input_shape = None
+        self.mean = None
```
hertschuh: These lines are already lines 115-120. You don't need both, remove one of them.

hertschuh (commented):
Great! Thanks for the update.

One last thing I noticed:

hertschuh (approved):
Thanks!

PR #22073, opened 2026-01-28
Fix TODO: enforce static scan dimension in `associative_scan`

MalyalaKarthik66:
This PR resolves the TODO in `associative_scan` by enforcing that the scan dimension is statically known. Without this check, dynamic or unknown dimensions can trigger deep recursion and runtime failures that are difficult to debug.

The validation is added early to fail fast with a clear error message, improving robustness without changing the intended behavior of the operation.


hertschuh:
Apparently https://github.com/keras-team/keras/pull/22072 will add support for dynamic scan dimensions

hertschuh:
Dynamic scan dimension use case was address by https://github.com/keras-team/keras/pull/22072

PR #22079, opened 2026-01-29
Validate positive height and width in image resize

sharpenteeth:
**Problem**
keras.ops.image.resize() and the Resize operation accept size=(height, width) without checking that height and width are positive. When users pass zero or negative dimensions (e.g. size=(0, 10), size=(-1, 10), or size=(0, 0)), backends (NumPy, TensorFlow, JAX, PyTorch) divide by target_height and target_width in the crop_to_aspect_ratio and pad_to_aspect_ratio paths. That leads to division-by-zero or invalid output shapes, with unclear stack traces instead of a clear validation error.

**Solution**
- In resize(): after the existing len(size) == 2 check, require size[0] > 0 and size[1] > 0, and raise ValueError with a message that size must have positive height and width.
- In Resize.__init__: validate length and positivity of size so that direct use (e.g. Resize(size=(0, 10))) raises the same clear error, keeping behavior consistent for both the function and the op.

**Tests**
- Add test_resize_invalid_size_zero_or_negative in ImageOpsCorrectnessTest for size=(0, 10), (10, 0), (0, 0), (-1, 10), (10, -1) for resize(), and Resize(size=(0, 10)) for the op.
- All cases assert ValueError with a message containing "size must have positive height and width".

### DIFF
--- a/keras/src/ops/image.py
+++ b/keras/src/ops/image.py
@@ -232,7 +232,20 @@ class Resize(Operation):
         name=None,
     ):
         super().__init__(name=name)
-        self.size = tuple(size)
+        size = tuple(size)
+        if len(size) != 2:
+            raise ValueError(
+                "Expected `size` to be a tuple of 2 integers. "
+                f"Received: size={size}"
+            )
+        if size[0] <= 0 or size[1] <= 0:
+            raise ValueError(
+                "`size` must have positive height and width. "
+                "Zero or negative dimensions would cause division by zero or "
+                "invalid output shape. "
+                f"Received: size={size}"
+            )
+        self.size = size
         self.interpolation = interpolation
         self.antialias = antialias
         self.crop_to_aspect_ratio = crop_to_aspect_ratio
@@ -340,6 +353,14 @@ def resize(
     if len(size) != 2:
         raise ValueError(
             "Expected `size` to be a tuple of 2 integers. "
+            f"Received: size={size}"
+        )
+    height, width = size[0], size[1]
+    if height <= 0 or width <= 0:
+        raise ValueError(
+            "`size` must have positive height and width. "
+            "Zero or negative dimensions would cause division by zero or "
+            "invalid output shape. "
             f"Received: size={size}"
         )
     if len(images.shape) < 3 or len(images.shape) > 4:
--- a/keras/src/ops/image.py
+++ b/keras/src/ops/image.py
@@ -355,8 +355,7 @@ def resize(
             "Expected `size` to be a tuple of 2 integers. "
             f"Received: size={size}"
         )
-    height, width = size[0], size[1]
-    if height <= 0 or width <= 0:
+    if size[0] <= 0 or size[1] <= 0:
         raise ValueError(
             "`size` must have positive height and width. "
             "Zero or negative dimensions would cause division by zero or "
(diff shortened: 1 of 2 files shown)


keras/src/ops/image.py
```diff
@@ -232,7 +232,20 @@ def __init__(
...
+                f"Received: size={size}"
+            )
+        if size[0] <= 0 or size[1] <= 0:
+            raise ValueError(
+                "`size` must have positive height and width. "
+                "Zero or negative dimensions would cause division by zero or "
+                "invalid output shape. "
+                f"Received: size={size}"
+            )
+        self.size = size
```
hertschuh: You can undo all of this. The only public API you can use to instantiate a `Resize` object is to call the `resize` function below.  The `resize` function does this validation already, and we don't want to duplicate the validation for static parameters.

keras/src/ops/image.py
```diff
@@ -342,6 +355,13 @@ def resize(
             "Expected `size` to be a tuple of 2 integers. "
             f"Received: size={size}"
         )
+    if size[0] <= 0 or size[1] <= 0:
+        raise ValueError(
+            "`size` must have positive height and width. "
+            "Zero or negative dimensions would cause division by zero or "
```
hertschuh: nitpick: no need to explain that much, zero or negative dimensions are just invalid. I would remove this part:
```
"Zero or negative dimensions would cause division by zero or "
"invalid output shape. "
```

keras/src/ops/image_test.py
```diff
@@ -1395,6 +1395,27 @@ def test_resize_with_pad(self, fill_value):
...
+        with self.assertRaisesRegex(
+            ValueError,
+            "`size` must have positive height and width",
+        ):
+            kimage.resize(x, size=invalid_size)
+        with self.assertRaisesRegex(
+            ValueError,
+            "`size` must have positive height and width",
+        ):
+            kimage.Resize(size=invalid_size)(x)
```
hertschuh: Remove this part, it won't work anymore.

keras/src/ops/image_test.py
```diff
@@ -2246,6 +2267,27 @@ def test_rgb_to_hsv_invalid_dtype(self):
...
+
```
hertschuh: I think you accidentally added the test twice. Remove this one.

sharpenteeth:
Hi, @hertschuh 
Please review my PR, again.
Thanks.

hertschuh (approved):
Thanks!

ahasselbring:
With the TensorFlow backend this breaks models that have variable shape inputs (such that size is a symbolic tensor):

> tensorflow.python.framework.errors_impl.OperatorNotAllowedInGraphError: Exception encountered when calling UpSampling2D.call().
>
> Using a symbolic `tf.Tensor` as a Python `bool` is not allowed. You can attempt the following resolutions to the problem: If you are running in Graph mode, use Eager execution mode or decorate this function with @tf.function. If you are using AutoGraph, you can try decorating this function with @tf.function. If that does not work, then you may be using an unsupported feature or your source code may not be visible to AutoGraph. See https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/autograph/g3doc/reference/limitations.md#access-to-source-code for more information.

PR #22082, opened 2026-01-30
Add initial steps/epoches arguments to orbax checkpointer

SamanehSaadat:
Add initial steps/epoches arguments to orbax checkpointer so the checkpointer can resume from a previous step.

### DIFF
--- a/keras/src/callbacks/orbax_checkpoint.py
+++ b/keras/src/callbacks/orbax_checkpoint.py
@@ -89,6 +89,21 @@ class OrbaxCheckpoint(MonitorCallback):
         save_freq=100)  # Save every 100 batches
 
     model.fit(epochs=EPOCHS, callbacks=[orbax_checkpoint_callback])
+
+    # To load a checkpoint from a previous training run and resume training:
+    # Scan the checkpoint directory to find the latest checkpoint.
+    latest_checkpoint_path = get_latest_checkpoint(checkpoint_dir)
+    # You can get the step numbers from the saved latest checkpoints.
+    # "checkpoint_dir/1000" is the checkpoint saved at step 1000
+    step = get_step_from_checkpoint_path(latest_checkpoint_path)
+    initial_epoch = step // steps_per_epoch
+    initial_batch = step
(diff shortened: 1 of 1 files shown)


keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -119,6 +119,8 @@ def __init__(
         max_to_keep=1,
         save_on_background=True,
         save_weights_only=False,
+        initial_batch=0,
+        initial_epoch=0,
```
hertschuh, line 142: Please add the docstring explaining these arguments.

hertschuh (commented):
So how does it work when you resume training. Do you have so set those values? How do you know what values to set? I know the optimizer has the step count.

Can you add an example of resuming training in the `Example:` section of the docstring?

SamanehSaadat:
> So how does it work when you resume training. Do you have so set those values? How do you know what values to set? I know the optimizer has the step count.

You can read the step/batch from the last saved checkpoint. And we know the number of training samples so we can compute the epoch, right?

> Can you add an example of resuming training in the `Example:` section of the docstring?

Will do

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -135,9 +156,11 @@ def __init__(
...
         self.save_weights_only = save_weights_only
         self._batches_seen_since_last_saving = 0
-        self._last_batch_seen = 0
-        self._current_epoch = 0  # Keep track of epoch
-        self._total_batches_seen = 0  # Global batch counter for step tracking
+        self._last_batch_seen = initial_batch
+        self._current_epoch = initial_epoch  # Keep track of epoch
+        self._total_batches_seen = (
+            initial_batch  # Global batch counter for step tracking
+        )
```
hertschuh, line 163: We can automate this. You can revert these changes and instead do this:
```python
    def set_model(self, model):
        super().set_model(model).
        if hasattr(model, "optimizer") and model.optimizer is not None:
            self._total_batches_seen = int(model.optimizer.iterations)
```

PR #22086, opened 2026-01-30
fix: pytorch onnx export with dynamic shapes

### LINKED ISSUE #22076: Symbolic shapes ignored when exporting to ONNX with Torch backend
When trying to export model containing symbolic / non-fixed shapes, the exporter bakes in the lowest possible shape (i.e., 1).

**Colab to reproduce:**
https://colab.research.google.com/drive/1H81pAhOWr42YFnzrAT4zGI3XooXvM8oe

**Suggestions:**
1) `model.export()` already supports `**kwargs`, and torch specifies `dynamic_shapes` / `dynamic_axes` for `torch.onnx.export()`, so just expose those, and it would work fine as is
2) (better ux:) calculate `dynamic_shapes` / `dynamic_axes` automatically based off the input spec
3) or all of the above

bittoby:
# Fix PyTorch ONNX Export Dynamic Shapes - Issue #22076
## Description
_Fixes PyTorch ONNX export to properly preserve dynamic dimensions from input signatures instead of converting them to fixed size 1._

## Issue
When exporting Keras models to ONNX with PyTorch backend, dynamic dimensions (specified as None in input signatures) were being converted to fixed dimensions of size 1. This made exported models unusable with variable input sizes at inference time.

## Example Problem:

Input signature: (None, None, None, 1)
Exported ONNX shape: [1, 1, 1, 1] ❌
Expected ONNX shape: ['batch_size', 'seq_len_or_height', 'width_or_depth', 1] ✅

## Solution
Generate Dynamic Shape Metadata: Extract dynamic dimensions from input signatures and create proper dynamic_shapes (TorchDynamo) and dynamic_axes (TorchScript) parameters.

Closes #22076

### DIFF
--- a/keras/src/export/onnx.py
+++ b/keras/src/export/onnx.py
@@ -76,13 +76,19 @@ def export_onnx(
     if input_signature is None:
         input_signature = get_input_signature(model)
         if not input_signature or not model._called:
-            raise ValueError(
-                "The model provided has never called. "
-                "It must be called at least once before export."
-            )
+            raise ValueError("The model provided has never called. ")
+
+    # Extract specs for proper input name generation
+    if len(input_signature) == 1 and isinstance(input_signature[0], list):
+        # Multi-input case: input_signature = [[spec1, spec2, ...]]
+        specs_for_names = input_signature[0]
+    else:
+        # Single input case: input_signature = [spec]
+        specs_for_names = input_signature
+
     input_names = [
         getattr(spec, "name", None) or f"input_{i}"
-        for i, spec in enumerate(input_signature)
+        for i, spec in enumerate(specs_for_names)
     ]
 
     if backend.backend() in ("tensorflow", "jax"):
@@ -105,10 +111,37 @@ def export_onnx(
     elif backend.backend() == "torch":
         import torch
 
+        """Generate dynamic_axes format for ONNX export."""
+        dynamic_axes = {}
+
+        for input_idx, spec in enumerate(specs_for_names):
+            if not hasattr(spec, "shape"):
+                continue
+
+            shape = spec.shape
+            dynamic_dims = {}
+
+            for dim_idx, dim_size in enumerate(shape):
+                if dim_size is None:
+                    if dim_idx == 0:
+                        dim_name = "batch"
+                    else:
+                        dim_name = f"dim_{input_idx}_{dim_idx}"
+                    dynamic_dims[dim_idx] = dim_name
+
+            if dynamic_dims:
+                input_name = (
+                    input_names[input_idx]
+                    if input_idx < len(input_names)
+                    else f"input_{input_idx}"
+                )
+                dynamic_axes[input_name] = dynamic_dims
+
         sample_inputs = tree.map_structure(
             lambda x: convert_spec_to_tensor(x, replace_none_number=1),
             input_signature,
         )
+
         sample_inputs = tuple(sample_inputs)
         # TODO: Make dict model exportable.
         if any(isinstance(x, dict) for x in sample_inputs):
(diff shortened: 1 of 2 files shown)


keras/src/export/onnx.py
```diff
@@ -105,10 +106,81 @@ def export_onnx(
...
+            for input_idx, spec in enumerate(specs):
+                if not hasattr(spec, "shape"):
+                    continue
+
+                shape = spec.shape
+                dynamic_dims = {}
+
+                for dim_idx, dim_size in enumerate(shape):
+                    if dim_size is None:
+                        dim_name = DIM_NAMES.get(dim_idx, f"dim_{dim_idx}")
```
hertschuh: This is making too strong assumptions.

Yes dim 0 can be assumed to be the batch size, that's fine.

But if you do this:
```
            1: "seq_len_or_height",
            2: "width_or_depth",
            3: "channels",
```

and you have multiple inputs. Let's say one input is an image, and one input is a sequence of tokens, you're forcing the image height to be the same as the sequence length for the tokens. But they're completely independent, you have to assume everything is independent.

This is how it's done for saved model:
https://github.com/keras-team/keras/blob/master/keras/src/backend/torch/export.py#L46-L58

Crystora: [hertschuh](https://github.com/hertschuh) Thanks for your comment. I fixed dimension names like below.
```
dimension:0   =>   batch
others        =>    dim_{index of input}_{index of dimension}
```

Crystora:
[hertschuh](https://github.com/hertschuh) Plz review my PR. Thanks.

keras/src/export/onnx.py
```diff
@@ -105,10 +106,76 @@ def export_onnx(
...
+                    if input_idx < len(input_names)
+                    else f"input_{input_idx}"
+                )
+                dynamic_axes[input_name] = dynamic_dims
+
+            return dynamic_axes if dynamic_axes else None
+
+        # Generate both formats
+        dynamic_shapes = generate_dynamic_shapes()
+        dynamic_axes = generate_dynamic_axes()
```
hertschuh: I feel like there's an opportunity to greatly simplify the code here:
- `dynamic_shapes` is used only once and as boolean to check if there is any dynamic shape. You can simply use `dynamic_axes` for that, it will work the same. So you can remove `dynamic_shapes` and the `generate_dynamic_shapes` function.
- `generate_dynamic_axes` has a loop which enumerates from `_generate_dynamic_dimensions` which itself has a loop. You can easily combine the two functions and the two loops in a single function and loop. The code will be a lot easier to follow.

keras/src/export/onnx.py
```diff
@@ -140,34 +207,109 @@ def export_onnx(
...
+            # For TorchScript (dynamo=False), use dynamic_axes parameter
+            if dynamic_axes:
+                export_kwargs["dynamic_axes"] = dynamic_axes
+
+            torch.onnx.export(model, sample_inputs, filepath, **export_kwargs)
+
+        # When dynamic shapes are present, prefer TorchScript over
+        # TorchDynamo because TorchDynamo has constraint inference issues
+        # with dynamic dimensions
+        if dynamic_shapes:
```
hertschuh: There is quite a bit of code duplication due to the fact that the call to `_export_with_torchscript` is duplicated.

Instead of doing:
```python
if dynamic_shapes:
    _export_with_torchscript(...)
    return

try:
    onnx_program = torch.onnx.export(...)
    ...
except Exception:
    _export_with_torchscript(...)
```

You can reorganize the flow:
```python
if not dynamic_shapes:
    try:
        onnx_program = torch.onnx.export(...)
        ...
        return
    except Exception:
        pass

_export_with_torchscript(...)
```

Then you don't even need the `def _export_with_torchscript` function definition since it's only used once and it adds a lot of lines of code, you can just inline the code right there.

Crystora:
[hertschuh](https://github.com/hertschuh) I corrected the issues you mentioned. Please review my PR.

keras/src/export/onnx.py
```diff
@@ -105,10 +114,53 @@ def export_onnx(
...
+            """Generate dynamic_axes format for ONNX export."""
+            # Extract specs from input_signature
+            if len(input_signature) == 1 and isinstance(
+                input_signature[0], list
+            ):
+                # Multi-input case: input_signature = [[spec1, spec2, ...]]
+                specs = input_signature[0]
+            else:
+                # Single input case: input_signature = [spec]
+                specs = input_signature
```
hertschuh: This code duplicates 84-90. Can't you just use `specs_for_names` directly?

keras/src/export/onnx.py
```diff
@@ -105,10 +114,53 @@ def export_onnx(
     elif backend.backend() == "torch":
         import torch
 
+        def generate_dynamic_axes():
```
hertschuh: You don't need a function here. You can just remove `def generate_dynamic_axes():`, unintend the code and remove `dynamic_axes = generate_dynamic_axes()` at the end.

keras/src/export/onnx_test.py
```diff
@@ -295,3 +295,172 @@ def test_export_with_input_names(self):
...
+
+        Tests two scenarios:
+        - batch_only: Only batch dimension is dynamic, spatial dims fixed
+        - height_width: Batch, height, width are dynamic, channels fixed
+        """
+        if backend.backend() != "torch":
+            self.skipTest(
+                "Dynamic shapes are currently only supported with torch "
+                "backend."
+            )
```
hertschuh: This test should definitely work with tensorflow, and probably with JAX too. Can you remove this and try?

keras/src/export/onnx_test.py
```diff
@@ -295,3 +295,172 @@ def test_export_with_input_names(self):
...
+        elif model_type == "functional":
+            input_layer = layers.Input(shape=input_shape)
+            output = input_layer
+            for layer in layer_list:
+                output = layer(output)
+            model = models.Model(inputs=input_layer, outputs=output)
+
+        # Build model with initial input
+        import torch
```
hertschuh: Please keep the test code backend agnostic (especially now that you're adding the other backends).

You can use `np.random.normal`.

keras/src/export/onnx_test.py
```diff
@@ -295,3 +295,172 @@ def test_export_with_input_names(self):
...
+            self.assertEqual(result[0].shape[0], expected_batch_size)
+            self.assertEqual(result[0].shape[1], 10)  # Number of classes
+
+    def test_multi_input_dynamic_shapes(self):
+        """Test ONNX export with multi-input model having dynamic shapes."""
+        if backend.backend() != "torch":
+            self.skipTest(
+                "Dynamic shapes are currently only supported with torch "
+                "backend."
+            )
```
hertschuh: Same comment, remove this, it should at least work with tensorflow and hopefully JAX.

keras/src/export/onnx_test.py
```diff
@@ -295,3 +295,172 @@ def test_export_with_input_names(self):
...
+        combined = layers.Concatenate()([text_pooled, image_pooled])
+        output = layers.Dense(5, activation="softmax")(combined)
+
+        model = models.Model(inputs=[text_input, image_input], outputs=output)
+
+        # Build model
+        import torch
+
+        sample_text = torch.randn(1, 20, 64)
+        sample_image = torch.randn(1, 32, 32, 3)
```
hertschuh: Same comment about using `np.random.normal` instead.

Crystora:
[hertschuh](https://github.com/hertschuh) I just fixed. Thanks. Plz review my PR.

hertschuh (commented):
Thanks for implementing this!

Crystora:
[hertschuh](https://github.com/hertschuh) Thanks for you help!

PR #22089, opened 2026-01-31
Implement nanprod function in keras.ops

shashaka:
Adds keras.ops.nanprod, which computes the product of elements along a specified axis while ignoring NaN values in the input tensor.

### DIFF
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -2192,7 +2192,7 @@ def nanprod(x, axis=None, keepdims=False):
     if not x.dtype.is_floating:
         return prod(x, axis=axis, keepdims=keepdims)
 
-    x_safe = tf.where(tf.math.is_nan(x), tf.ones_like(x), x)
+    x_safe = tf.where(tf.math.is_nan(x), tf.ones((), dtype=x.dtype), x)
     return prod(x_safe, axis=axis, keepdims=keepdims)
 
 
--- a/keras/src/backend/torch/numpy.py
+++ b/keras/src/backend/torch/numpy.py
@@ -1331,7 +1331,7 @@ def nanprod(x, axis=None, keepdims=False):
         return prod(x, axis=axis, keepdims=keepdims)
 
     return prod(
-        torch.where(torch.isnan(x), torch.ones_like(x), x),
+        torch.where(torch.isnan(x), torch.ones((), dtype=x.dtype), x),
         axis=axis,
         keepdims=keepdims,
     )


keras/src/backend/torch/numpy.py
```diff
@@ -1318,6 +1318,25 @@ def nanmin(x, axis=None, keepdims=False):
...
 
 
+def nanprod(x, axis=None, keepdims=False):
+    x = convert_to_tensor(x)
+
+    if axis == () or axis == []:
+        return torch.nan_to_num(x, nan=1)
+
+    if isinstance(axis, (list, tuple)):
+        axis = sorted(axis, reverse=True)
```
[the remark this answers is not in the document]
shashaka: There is no sorting on prod function.

keras/src/backend/numpy/numpy.py
```diff
@@ -973,6 +973,19 @@ def nanmin(x, axis=None, keepdims=False):
...
+    axis = standardize_axis_for_numpy(axis)
+
+    x = convert_to_tensor(x)
+
+    dtype = dtypes.result_type(x.dtype)
+    if dtype in ("bool", "int8", "int16"):
+        dtype = "int32"
+    elif dtype in ("uint8", "uint16"):
+        dtype = "uint32"
+    return np.nanprod(x, axis=axis, keepdims=keepdims).astype(dtype)
```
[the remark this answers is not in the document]
shashaka: I now pass the target dtype to np.nanprod so the product is computed in the widened type and avoids integer overflow.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -2186,6 +2186,16 @@ def nanmin(x, axis=None, keepdims=False):
     )
 
 
+def nanprod(x, axis=None, keepdims=False):
+    x = convert_to_tensor(x)
+
+    if not x.dtype.is_floating:
+        return prod(x, axis=axis, keepdims=keepdims)
+
+    x_safe = tf.where(tf.math.is_nan(x), tf.ones_like(x), x)
```
hertschuh: nitpick: like for `nansum`, you don't need the full array of ones, just a scalar, so you can replace `tf.ones_like(x)` with `tf.ones((), dtype=dtype)`

keras/src/backend/torch/numpy.py
```diff
@@ -1318,6 +1318,25 @@ def nanmin(x, axis=None, keepdims=False):
...
+
+    if not torch.is_floating_point(x):
+        return prod(x, axis=axis, keepdims=keepdims)
+
+    return prod(
+        torch.where(torch.isnan(x), torch.ones_like(x), x),
```
hertschuh: Same here, use `torch.ones((), dtype=x.dtype)`.

hertschuh (commented):
Thanks for adding this!

keras/src/backend/tensorflow/numpy.py
shashaka: Agreed. I’ve replaced tf.ones_like(x) with tf.ones((), dtype=x.dtype).

keras/src/backend/torch/numpy.py
shashaka: I updated.

hertschuh (approved):
Thanks!

PR #22092, opened 2026-02-01
Fail fast on invalid convolution output shapes during symbolic execution

### LINKED ISSUE #22077: `Conv1D` silently returns zero-dimension output in symbolic mode instead of raising an error for invalid parameters.
**Related Keras Version**
keras==3.13.1, keras==3.10.0
The `Conv1D` layer shows inconsistent behavior between eager and symbolic execution. When given parameters that are invalid for the input shape (e.g., a `dilation_rate` that makes the effective kernel larger than the input), it correctly raises an error in eager mode.

However, in symbolic mode, it fails silently and returns a `KerasTensor` with a zero-dimension shape (e.g., `(None, 0, 1)`). This can mask configuration errors during model construction, leading to downstream issues that are difficult to debug. The symbolic path should perform the same validation and fail fast.

This issue is observed across different backends (TensorFlow==2.20.0, PyTorch==2.10.0+cu128).

**Code to Reproduce**

```python
import keras

# Eager execution
try:
    eager_input = keras.random.normal(shape=(2, 10, 3))
    layer = keras.layers.Conv1D(
        filters=1, 
        kernel_size=2, 
        strides=1, 
        padding='valid',
        dilation_rate = 10,
        activation="relu"
    )
    eager_output = layer(eager_input)
    print("Eager output shape:", eager_output.shape) 
except Exception as e:
    print("Error during eager execution:", e)

# Symbolic execution
symbolic_input = keras.Input(shape=(10, 3)) 
symbolic_output = layer(symbolic_input)
print("Symbolic output shape:", symbolic_output.shape)
```

**Actual Behavior**

Eager execution fails with a descriptive error, but symbolic execution proceeds and produces a tensor with a zero dimension.

```
Keras version: 3.13.1
Error during eager execution: "The convolution operation resulted in an empty output. Output shape: (2, 0, 1)..."
Symbolic output shape: (None, 0, 1)
```
*(Note: The exact error message varies by backend, but the eager failure is consistent.)*

**Expected Behavior**

Symbolic execution should also raise an error during graph construction, mirroring the eager behavior. This would provide immediate feedback that the layer configuration is invalid.

```
Keras version: 3.13.1
Error during eager execution: "The convolution operation resulted in an empty output..."
# Followed by a similar error during the symbolic call:
# layer(symbolic_input)
# -> raises ValueError: "The convolution parameters are invalid for the given input shape..."
```

MalyalaKarthik66:
This PR fixes an inconsistency between eager and symbolic execution for convolution layers.

When convolution parameters produce an empty spatial output (e.g. large dilation), eager execution raises an error while symbolic execution previously returned a zero-sized KerasTensor. This change adds validation in compute_conv_output_shape to fail fast during symbolic shape inference.

Includes regression tests for Conv1D, Conv2D, and Conv3D.

Fixes #22077

### DIFF
--- a/keras/src/ops/operation_utils.py
+++ b/keras/src/ops/operation_utils.py
@@ -219,7 +219,7 @@ def compute_conv_output_shape(
         for i in range(len(output_spatial_shape)):
             if i not in none_dims and output_spatial_shape[i] <= 0:
                 raise ValueError(
-                    "Computed output size would be negative. Received "
+                    "Computed output size would be zero or negative. Received "
                     f"`inputs shape={input_shape}`, "
                     f"`kernel shape={kernel_shape}`, "
                     f"`dilation_rate={dilation_rate}`, "


keras/src/layers/convolutional/base_conv.py
```diff
@@ -264,7 +264,7 @@ def call(self, inputs):
         return outputs
 
     def compute_output_shape(self, input_shape):
-        return compute_conv_output_shape(
+        output_shape = compute_conv_output_shape(
```
hertschuh: You you move the validation inside `compute_conv_output_shape` instead? (and revert this file)

https://github.com/keras-team/keras/blob/master/keras/src/ops/operation_utils.py#L237

This way, other convolutions will benefit from it too (separable convs, depthwise convs).

MalyalaKarthik66: Thanks for the review!

Initially, I considered placing the validation directly in `compute_conv_output_shape`, but due to some uncertainty around the broader impact, I temporarily moved it into `BaseConv`. After your clarification, it makes sense that `compute_conv_output_shape` is the right place so all convolution variants (including depthwise and separable convs) benefit consistently.

I’ve now moved the validation into `compute_conv_output_shape`.

keras/src/ops/operation_utils.py
```diff
@@ -217,14 +217,17 @@ def compute_conv_output_shape(
             + 1
         )
         for i in range(len(output_spatial_shape)):
-            if i not in none_dims and output_spatial_shape[i] < 0:
+            if i not in none_dims and output_spatial_shape[i] <= 0:
                 raise ValueError(
                     "Computed output size would be negative. Received "
```
hertschuh: Can you change the message from `Computed output size would be negative.` to `Computed output size would be zero or negative`?

MalyalaKarthik66: Updated the error message.

PR #22096, opened 2026-02-02
Implement cross product operation for OpenVINO backend

Shi-pra-19:
Implemented cross product ops for OpenVINO backend

Tests are now included and passing

### DIFF
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -813,21 +813,22 @@ def cross(x1, x2, axisa=-1, axisb=-1, axisc=-1, axis=None):
     shape1 = x1.get_partial_shape()
     shape2 = x2.get_partial_shape()
 
-    if shape1.rank.is_dynamic or shape2.rank.is_dynamic:
-        raise ValueError("Input ranks must be static for cross product")
-
     # Rank Normalization
     rank1 = shape1.rank.get_length()
     rank2 = shape2.rank.get_length()
 
-    axisa %= rank1
-    axisb %= rank2
+    axisa = canonicalize_axis(axisa, rank1)
+    axisb = canonicalize_axis(axisb, rank2)
+    axisc = canonicalize_axis(axisc, rank1 if rank1 > rank2 else rank2)
 
     d1 = shape1[axisa].get_length()
     d2 = shape2[axisb].get_length()
 
     if d1 not in (2, 3) or d2 not in (2, 3):
-        raise ValueError("Dimension must be 2 or 3 for cross product")
+        raise ValueError(
+            "Dimension of vectors for cross product must be 2 or 3. "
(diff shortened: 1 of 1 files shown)


keras/src/backend/openvino/numpy.py
```diff
@@ -802,7 +802,77 @@ def count_nonzero(x, axis=None):
...
+    x1 = get_ov_output(x1)
+    x2 = get_ov_output(x2)
+
+    x1, x2 = _align_operand_types(x1, x2, "cross()")
+
+    shape1 = x1.get_partial_shape()
+    shape2 = x2.get_partial_shape()
+
+    if shape1.rank.is_dynamic or shape2.rank.is_dynamic:
+        raise ValueError("Input ranks must be static for cross product")
```
hertschuh: For future reference: I know Gemini keeps commenting about this and I know this check is in a lot of places in this file, but the truth is that nothing in Keras supports dynamic ranks, so we don't even check anywhere I think it fails much earlier anyway.

All that to say that you don't need to add this check.

keras/src/backend/openvino/numpy.py
```diff
@@ -802,7 +802,77 @@ def count_nonzero(x, axis=None):
...
+
+    # Rank Normalization
+    rank1 = shape1.rank.get_length()
+    rank2 = shape2.rank.get_length()
+
+    axisa %= rank1
+    axisb %= rank2
```
hertschuh: Use this instead, which will perform a check on the axes:
```python
axisa = canonicalize_axis(axisa, rank1)
axisb = canonicalize_axis(axisa, rank2)
```

keras/src/backend/openvino/numpy.py
```diff
@@ -802,7 +802,77 @@ def count_nonzero(x, axis=None):
...
+    res_z = ov_opset.subtract(
+        ov_opset.multiply(u[0], v[1]), ov_opset.multiply(u[1], v[0])
+    )
+
+    # If dim was 2D, we remove the padded zero component.
+    if d1 == 2 and d2 == 2:
+        result = res_z
+        result = ov_opset.squeeze(result, ov_opset.constant([axisc], Type.i32))
+    else:
+        result = ov_opset.concat([res_x, res_y, res_z], axisc)
```
hertschuh, line 874: You should probably do `canonicalize_axis` with `axisc` to make sure it has a valid value.

Shi-pra-19: Updated, thanks for the suggestion.

keras/src/backend/openvino/numpy.py
Shi-pra-19: Updated 👍🏻

keras/src/backend/openvino/numpy.py
Shi-pra-19: Understood. I'll skip adding this check. 👍🏻

PR #22099, opened 2026-02-02
Add assets support to OrbaxCheckpoint using checkpointables API

amitsrivastava78:
- Implement assets saving/loading using Orbax checkpointables API
- Use DiskIOStore to preserve Keras hierarchical path structure for assets
- Modified _save_checkpoint() to detect and save assets with save_checkpointables
- Updated saving_api.py to handle both pytree and checkpointables formats
- Modified saving_lib.py to allow None for weights_store in assets-only operations
- Added comprehensive test_checkpoint_with_assets with custom LayerWithAssets
- Test is parameterized for both sync and async modes
- Works on TensorFlow, JAX, and PyTorch backends (18 tests pass on each)
- Disabled JIT compilation for JAX/Torch to avoid compilation issues in tests

### DIFF
--- a/keras/src/callbacks/orbax_checkpoint.py
+++ b/keras/src/callbacks/orbax_checkpoint.py
@@ -272,9 +272,7 @@ class OrbaxCheckpoint(MonitorCallback):
         # Collect assets if saving full model (not just weights)
         assets_dict = None
         if not self.save_weights_only:
-            from keras.src.saving.saving_lib import _save_assets_to_dict
-
-            assets_dict = _save_assets_to_dict(self.model)
+            assets_dict = saving_lib._save_assets_to_dict(self.model)
 
         # Use a single with statement. If context_options is empty,
         # Context() uses defaults.
--- a/keras/src/callbacks/orbax_checkpoint_test.py
+++ b/keras/src/callbacks/orbax_checkpoint_test.py
@@ -583,9 +583,7 @@ class OrbaxCheckpointTest(testing.TestCase, parameterized.TestCase):
         callback.checkpointer.close()
 
         # Load the model back through the Orbax load path
-        from keras.src.saving import saving_api
-
-        loaded_model = saving_api.load_model(checkpoint_dir)
+        loaded_model = saving.load_model(checkpoint_dir)
 
         # Verify model structure
         self.assertEqual(model.name, loaded_model.name)


keras/src/saving/saving_api.py
```diff
@@ -258,6 +260,8 @@ def save_weights(
 
 @keras_export("keras.saving.load_weights")
 def load_weights(model, filepath, skip_mismatch=False, **kwargs):
+    from keras.src.saving import saving_lib as slib
+
```
hertschuh: `saving_lib` is already imported at the top. And just in general, no late imports unless absolutely needed.

Remove all the renames from `saving_lib` to `slib`.

keras/src/saving/saving_api.py
```diff
@@ -336,8 +339,34 @@ def load_weights(model, filepath, skip_mismatch=False, **kwargs):
             # It's a step directory, use it directly
             checkpoint_path = filepath
 
-        # Load checkpoint
-        loaded_state = ocp.load_pytree(checkpoint_path)
+        # Check if checkpoint has assets by inspecting metadata
+        try:
+            # Try to peek at the checkpoint structure without full loading
+            import os as _os
```
hertschuh: Why??? it's already imported as `os`

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -16,6 +16,54 @@
 # Import advanced Orbax functionality directly from the LazyModule
 
 
+# Custom layer with assets for testing (JAX-compatible)
+@register_keras_serializable(package="test")
+class LayerWithAssets(layers.Layer):
```
hertschuh: We already talked about the fact that this is not needed. You can use `StringLookup`, which will do exactly the same.

I'd rather use `StringLookup` for 2 reasons:
- no extra test code needed
- `LayerWithAssets` might accidentally work with Orbax while `StringLookup` won't because `LayerWithAssets` was written for the orbax test

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -269,20 +270,75 @@ def _save_checkpoint(self, step, logs=None):
...
             composite_state["model_config"] = config_json
 
+        # Collect assets if saving full model (not just weights)
+        # Use saving_lib._save_state to properly handle recursive
+        # asset collection
+        assets_dict = None
+        if not self.save_weights_only:
+            # Let DiskIOStore create its own temp directory structure
+            # Pass a simple name, not a full path
+            assets_root = f"keras_assets_{id(self)}_{step}"
```
hertschuh: You can't use a local path like. You have to assume the local path is not writeable. You need to use a temp folder.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -269,20 +270,75 @@ def _save_checkpoint(self, step, logs=None):
             config_json, _ = saving_lib._serialize_model_as_json(self.model)
             composite_state["model_config"] = config_json
 
+        # Collect assets if saving full model (not just weights)
+        # Use saving_lib._save_state to properly handle recursive
+        # asset collection
+        assets_dict = None
+        if not self.save_weights_only:
+            # Let DiskIOStore create its own temp directory structure
+            # Pass a simple name, not a full path
```
hertschuh: DiskIOStore only uses its own temp directory structure when extracting an archive. When saving, it will write in the path you pass.

In fact, it's a bug that it creates a temp dir when it's not extracting an archive, it's never used.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -269,20 +270,75 @@ def _save_checkpoint(self, step, logs=None):
...
+                saving_lib._save_state(
+                    self.model,
+                    weights_store=None,  # Only save assets, not weights
+                    assets_store=assets_store,
+                    inner_path="",
+                    visited_saveables=set(),
+                )
+
+                # Get the working directory where assets were saved
+                working_dir = assets_store.working_dir
```
hertschuh: No need to peek in `assets_store`, this is `assets_root`

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -269,20 +270,75 @@ def _save_checkpoint(self, step, logs=None):
...
+                    "pytree": composite_state,
+                    "assets": assets_dict,
+                }
+                save_fn = self.checkpointer.save_checkpointables
+                save_async_fn = self.checkpointer.save_checkpointables_async
+            else:
+                # No assets, use simpler pytree API
+                payload = composite_state
+                save_fn = self.checkpointer.save_pytree
+                save_async_fn = self.checkpointer.save_pytree_async
```
hertschuh: Is this a good idea to use two completely different API based on the presence of assets or not?

keras/src/saving/saving_api.py
```diff
@@ -336,8 +339,34 @@ def load_weights(model, filepath, skip_mismatch=False, **kwargs):
...
+
+            has_assets = _os.path.exists(
+                _os.path.join(checkpoint_path, "assets")
+            )
+        except:
+            has_assets = False
```
hertschuh: This is making assumption about how Orbax implements their checkpointable feature. We shouldn't do this in case they change their implementation.

How can we detect this using only Orbax APIs?

keras/src/saving/saving_api.py
amitsrivastava78: Done

keras/src/saving/saving_api.py
amitsrivastava78: Done

keras/src/callbacks/orbax_checkpoint_test.py
amitsrivastava78: Done! I've replaced the custom LayerWithAssets class with StringLookup. The test now:
1. Uses StringLookup layer which has real vocabulary assets (stored in asset files)
2. Creates a simple model with StringLookup, saves it, loads it, and verifies the vocabulary is correctly restored

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: 1. Done! The code now explicitly uses tempfile.gettempdir() to ensure assets are written to the system's temp directory

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: Yes you are right , modified the code to always use save_checkpointables

keras/src/saving/saving_api.py
amitsrivastava78: Replaced filesystem-based asset detection with Orbax's proper metadata API

keras/src/saving/saving_lib.py
```diff
@@ -707,10 +707,12 @@ def _save_state(
 ):
     from keras.src.saving.keras_saveable import KerasSaveable
 
-    if not isinstance(weights_store, (H5IOStore, ShardedH5IOStore, NpzIOStore)):
+    if not isinstance(
+        weights_store, (H5IOStore, ShardedH5IOStore, NpzIOStore, type(None))
+    ):
```
hertschuh, line 712: I'm confused about this, you allow `None` here, now. But later, `weights_store` is used.
1. It should fail, the unit tests should not pass, something is off
2. you need to add `if`s to skip the weights part of it

keras/src/saving/saving_lib.py
```diff
@@ -773,10 +775,12 @@ def _load_state(
...
```
hertschuh, line 780: Some comment about allowing `None` but not adding the necessary `if`s.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -269,19 +272,73 @@ def _save_checkpoint(self, step, logs=None):
...
+        # asset collection
+        assets_dict = None
+        if not self.save_weights_only:
+            # Create a unique path in temp directory for assets
+            # Use timestamp to ensure uniqueness across multiple saves
+            assets_root = os.path.join(
+                tempfile.gettempdir(),
+                f"keras_assets_{step}_{id(self)}_{int(time.time() * 1e9)}",
+            )
+            assets_store = saving_lib.DiskIOStore(assets_root, mode="w")
```
hertschuh: You created `_load_assets_from_dict` as a separate utility. I feel like you should do the same for saving, i.e. have a utility function to save assets and collect them as a dict.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -269,19 +272,73 @@ def _save_checkpoint(self, step, logs=None):
...
+
+                # Check if any assets were actually saved
+                assets_dict = {}
+                for root, dirs, files in os.walk(assets_root):
+                    for file in files:
+                        file_path = os.path.join(root, file)
+                        rel_path = os.path.relpath(file_path, assets_root)
+                        with open(file_path, "rb") as f:
+                            # Convert bytes to numpy uint8 array for Orbax
+                            assets_dict[rel_path] = np.frombuffer(
```
hertschuh: So there is one thing I don't like about this, which is that you use the relative path as a key and then, when reloading, you use the key and a relative path. There are several issues:
- This won't work on Windows
- This introduces a vulnerability, which is called "zip slip". Somebody can craft an archive with `../../../../etc/passwd` for the path and change your root passwords.

I think the clean alternative is to create a nested structure of dicts which mimics the nested folders.

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -269,19 +272,73 @@ def _save_checkpoint(self, step, logs=None):
...
+        # Use saving_lib._save_state to properly handle recursive
+        # asset collection
+        assets_dict = None
+        if not self.save_weights_only:
+            # Create a unique path in temp directory for assets
+            # Use timestamp to ensure uniqueness across multiple saves
+            assets_root = os.path.join(
+                tempfile.gettempdir(),
+                f"keras_assets_{step}_{id(self)}_{int(time.time() * 1e9)}",
+            )
```
hertschuh: just do `assets_root = saving_lib.get_temp_dir()`. It will take of creating a unique folder.

keras/src/saving/saving_api.py
```diff
@@ -350,6 +375,44 @@ def load_weights(model, filepath, skip_mismatch=False, **kwargs):
         )
 
 
+def _load_assets_from_dict(model, assets_dict):
```
hertschuh: Move `_load_assets_from_dict` to `saving_lib.py`.

My understanding is that `saving_lib.py` is more the private utilities.

keras/src/saving/saving_api.py
```diff
@@ -350,6 +375,44 @@ def load_weights(model, filepath, skip_mismatch=False, **kwargs):
...
+    Args:
+        model: The model to load assets into.
+        assets_dict: Dictionary mapping relative paths to numpy uint8 arrays
+            containing asset file contents.
+    """
+    if not assets_dict:
+        return
+
+    temp_assets_dir = tempfile.mkdtemp()
+    try:
```
hertschuh: Use  `with tempfile.TemporaryDirectory() as tmp_dir:` instead of try / finally.

keras/src/saving/saving_api.py
```diff
@@ -336,8 +339,30 @@ def load_weights(model, filepath, skip_mismatch=False, **kwargs):
...
+            loaded_checkpointables = ocp.load_checkpointables(
+                checkpoint_path, dict(pytree=None)
+            )
+            assets_dict = None
+
+        loaded_state = loaded_checkpointables["pytree"]
+
+        # Load assets if present
+        if assets_dict:
+            _load_assets_from_dict(model, assets_dict)
```
hertschuh: This is `load_weights`. You should not look for assets, and even if they are present, you should ignore them.

keras/src/saving/saving_api.py
```diff
@@ -366,8 +429,28 @@ def _load_model_from_orbax_checkpoint(
 
     # Load the composite state efficiently
     checkpointer = ocp.training.Checkpointer(directory=filepath)
+
+    # Check if checkpoint has assets using Orbax metadata API
+    try:
+        checkpoint_metadata = ocp.checkpointables_metadata(checkpoint_path)
+        has_assets = "assets" in checkpoint_metadata.metadata
```
hertschuh: I'm confused. You don't have the symmetrical code to write this metadata.

keras/src/saving/saving_api.py
```diff
@@ -366,8 +429,28 @@ def _load_model_from_orbax_checkpoint(
...
+    except Exception:
+        has_assets = False
```
hertschuh: I believe this broad catch is hiding the fact that this system doesn't work.

Do we have reload assets?

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -530,3 +528,65 @@ def call(self, inputs):
...
+    def test_checkpoint_with_assets(self, save_on_background):
+        """Test checkpoint saving/loading with layers that have assets.
+
+        Tests that models with preprocessing layers that have vocab assets
+        can be saved and loaded correctly through Orbax checkpoints.
+        """
+        # Create a model with StringLookup which has vocabulary assets
+        # StringLookup stores its vocabulary as an asset file
+        string_lookup = layers.StringLookup(
+            vocabulary=["cat", "dog", "bird", "fish"],
```
hertschuh: Oh... now it makes sense.

So you you pass the vocabulary like that, it's serialized in `get_config`, not in the assets. So this is not exercising the assets case.

You need the create a separate text file for the vocabulary and pass the path.

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -530,3 +528,65 @@ def call(self, inputs):
...
+        outputs = layers.Embedding(input_dim=10, output_dim=8)(x)
+        model = models.Model(inputs, outputs, name="model_with_assets")
+
+        # Get original vocabulary
+        original_vocab = string_lookup.get_vocabulary()
+
+        checkpoint_dir = os.path.join(
+            self.get_temp_dir(), f"test_assets_{save_on_background}_{id(self)}"
+        )
+        os.makedirs(checkpoint_dir, exist_ok=True)
```
hertschuh: You don't need to create a nested path. `self.get_temp_dir()` does that for you.

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -530,3 +528,65 @@ def call(self, inputs):
...
+        # Verify vocabulary (assets) was restored correctly
+        loaded_string_lookup = loaded_model.get_layer("string_lookup_layer")
+        loaded_vocab = loaded_string_lookup.get_vocabulary()
+
+        self.assertEqual(
+            len(original_vocab), len(loaded_vocab), "Vocabulary length mismatch"
+        )
+
+        for i, (orig, loaded) in enumerate(zip(original_vocab, loaded_vocab)):
+            self.assertEqual(orig, loaded, f"Vocabulary mismatch at index {i}")
```
hertschuh: Just do `self.assertEquals(orig, loaded)`.

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: Two small helpers in saving_api.py:378:
* _split_path_components(path) — platform-safe decomposition of a relative path into components
* _write_nested_dict_to_dir(tree, base_dir) — reconstructs the directory tree from the nested dict, only ever joining one component at a time

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: done as part of above

keras/src/saving/saving_api.py
amitsrivastava78: Done

keras/src/saving/saving_api.py
amitsrivastava78: Done

keras/src/saving/saving_api.py
amitsrivastava78: Done

keras/src/saving/saving_api.py
amitsrivastava78: The old code used ocp.checkpointables_metadata(checkpoint_path) (module-level function with a path) which wasn't symmetric with the save side. Now it uses checkpointer.checkpointables_metadata(step) — the same Checkpointer instance already created — to inspect .metadata.keys() and only requests checkpointable keys that actually exist.

keras/src/callbacks/orbax_checkpoint_test.py
amitsrivastava78: 1. Vocabulary is now passed as a file path — creates a vocab.txt file and passes its path to StringLookup(vocabulary=vocab_file).
2. Actually round-trips through Orbax — saves via OrbaxCheckpoint._save_checkpoint() and loads via saving_api.load_model(checkpoint_dir), exercising the real Orbax assets path end-to-end.

keras/src/callbacks/orbax_checkpoint_test.py

keras/src/callbacks/orbax_checkpoint_test.py

keras/src/saving/saving_api.py
amitsrivastava78: yes, asset reload works end-to-end. The test_checkpoint_with_assets test verifies the full Orbax round-trip

keras/src/saving/saving_lib.py
amitsrivastava78: You're right to be concerned — but the necessary if guards already exist in the current code. They're just not visible in the diff because they weren't changed by this PR.
in _save_state
if hasattr(saveable, "save_own_variables") and weights_store:
And in load_state
if hasattr(saveable, "load_own_variables") and weights_store:
The and weights_store is a truthiness check — None is falsy, so the entire block is skipped.

keras/src/saving/saving_lib.py
amitsrivastava78: Same as above — the guard already exists a few lines below (not visible in the diff since it's unchanged code):

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -269,19 +269,37 @@ def _save_checkpoint(self, step, logs=None):
             config_json, _ = saving_lib._serialize_model_as_json(self.model)
             composite_state["model_config"] = config_json
 
+        # Collect assets if saving full model (not just weights)
+        assets_dict = None
+        if not self.save_weights_only:
+            from keras.src.saving.saving_lib import _save_assets_to_dict
```
hertschuh: Please stop doing local imports. All the imports you need are already there:

`assets_dict = saving_lib.save_assets_to_dict(self.model)`

keras/src/callbacks/orbax_checkpoint.py
```diff
@@ -269,19 +269,37 @@ def _save_checkpoint(self, step, logs=None):
...
+            assets_dict = _save_assets_to_dict(self.model)
+
         # Use a single with statement. If context_options is empty,
         # Context() uses defaults.
         with ocp.Context():
             # Determine sync vs async based on save_on_background setting
             use_sync = not self.save_on_background
 
+            # Always use checkpointables API for consistency
+            # If no assets, just pass pytree alone
```
hertschuh, line 284: Is this a breaking change for people already using `OrbaxCheckpoint`?

keras/src/callbacks/orbax_checkpoint_test.py
```diff
@@ -530,3 +528,71 @@ def call(self, inputs):
...
+        )
+
+        # We can't easily train with string inputs, so invoke the
+        # save path directly.
+        callback.set_model(model)
+        callback._save_checkpoint(step=0)
+        callback.checkpointer.close()
+
+        # Load the model back through the Orbax load path
+        from keras.src.saving import saving_api
```
hertschuh: Remove local import, you already have an import that works:

```python
loaded_model = saving.load_model(checkpoint_dir)
```

keras/src/callbacks/orbax_checkpoint.py

keras/src/callbacks/orbax_checkpoint_test.py

keras/src/callbacks/orbax_checkpoint.py
amitsrivastava78: No this will not be the breaking change.

amitsrivastava78:
> Last nitpicks.
> 
> Again, local imports should never be used unless there is a circular import.

Sure will keep this in mind for future as well

hertschuh (approved):

PR #22100, opened 2026-02-02
feat: Use JAX's AbstractMesh in distribution lib

bitloi:


bitloi:
@hertschuh Can you check the pr and give me your opinion?

keras/src/backend/jax/distribution_lib.py
```diff
@@ -240,5 +295,10 @@ def _to_backend_layout(tensor_layout):
             "for TensorLayout."
         )
     partition_spec = jax.sharding.PartitionSpec(*tensor_layout.axes)
+    # In tracing scope use AbstractMesh for stable JIT cache; else concrete.
```
hertschuh: Why this hook? In what context will this be used?

Your code adds explicit use of AbstractMesh in the context of `jit(..., out_shardings=...)`. Do we also need this?

keras/src/distribution/distribution_lib.py
```diff
@@ -273,6 +281,9 @@ def device_mesh(self, device_mesh):
 
     @property
     def backend_layout(self):
+        # JAX: context-dependent layout (abstract vs concrete mesh).
+        if backend() == "jax":
+            return distribution_lib._to_backend_layout(self)
```
hertschuh: The `backend_layout` property was added to provide some caching. Some testing showed that not caching the layout significantly slowed down training.

This hook makes it so that on JAX we no longer cache. I'm concerned that we creating a regression in performance here. But this ties to my other question in the other comment.

keras/src/backend/jax/distribution_lib.py
bitloi: The hook was only for the `relayout` path (`distribute_tensor`), which runs both:
- Inside JIT: `with_sharding_constraint`
- Eagerly: `device_put` / `make_array_from_process_local_data`

I refactored so it’s no longer needed in `_to_backend_layout`:
- The Trainer’s `jit(..., out_shardings=...)` already uses `_to_abstract_shardings()`.
- The only other place that needs “abstract when tracing” is `relayout`, so that logic now lives there: we call `_to_abstract_shardings(layout)` before `with_sharding_constraint`.

keras/src/distribution/distribution_lib.py
bitloi: To avoid a regression I refactored so that:
1. `_to_backend_layout` always returns a concrete `NamedSharding`.
2. `TensorLayout.backend_layout` is cached again for JAX (same as other backends).
3. Only `relayout` converts to abstract when in a tracing scope.

So we keep a stable JIT cache where it matters and avoid the uncached-layout performance hit.

bitloi:
@hertschuh Please check my changes again

keras/src/distribution/distribution_lib.py
```diff
@@ -217,6 +217,13 @@ def backend_mesh(self):
             self._backend_mesh = distribution_lib._to_backend_mesh(self)
         return self._backend_mesh
 
+    @property
+    def backend_mesh_abstract(self):
+        """Backend mesh for JIT-stable cache keys (e.g. JAX AbstractMesh)."""
+        if hasattr(distribution_lib, "_to_backend_abstract_mesh"):
+            return distribution_lib._to_backend_abstract_mesh(self)
+        return self.backend_mesh
```
hertschuh: Let's remove this. It's not used, and it's unclear if the concept of "abstract mesh" will ever make sense for other backends besides JAX.

keras/src/backend/jax/distribution_lib.py
```diff
@@ -225,9 +225,64 @@ def _to_backend_mesh(device_mesh):
...
+    jax_mesh = device_mesh.backend_mesh
+    abstract = getattr(jax_mesh, "abstract_mesh", None)
+    if abstract is not None:
+        return abstract
+    if not hasattr(jax.sharding, "AbstractMesh"):
+        return None
+    # Fallback for older JAX: build AbstractMesh(axis_sizes, axis_names).
+    shape = device_mesh.devices.shape
+    axis_names = device_mesh.axis_names
+    return jax.sharding.AbstractMesh(tuple(shape), tuple(axis_names))
```
hertschuh: Let's remove this, it's not used right now.

keras/src/backend/jax/distribution_lib.py
```diff
@@ -225,9 +225,64 @@ def _to_backend_mesh(device_mesh):
...
+    def convert_one(s):
+        if not isinstance(s, jax.sharding.NamedSharding):
+            return s
+        mesh = s.mesh
+        abstract = getattr(mesh, "abstract_mesh", None)
+        if abstract is None:
+            return s
+        return jax.sharding.NamedSharding(abstract, s.spec)
+
+    return jax.tree_util.tree_map(convert_one, shardings)
```
hertschuh: For consistency with the rest of the functions in this file, let's make this handle a single sharding. The tree part of it can be done where needed (trainer.py). In particular, you don't need line 85.

- rename `def _to_abstract_sharding(shardings:`
- remove `convert_one` and `tree_map`
- in `trainer._get_state_sharding_spec`, you'll have to do `_to_abstract_sharding` in each list comprehension.

keras/src/backend/jax/trainer.py
```diff
@@ -265,6 +265,10 @@ def make_train_function(self, force=False):
             out_shardings = None
             if distribution_lib.distribution() is not None:
                 state_shardings = self._get_state_sharding_spec()
+                # Use AbstractMesh for out_shardings so JIT cache key is stable.
+                state_shardings = jax_distribution_lib._to_abstract_shardings(
+                    state_shardings
+                )
```
hertschuh: Let's move this inside `_get_state_sharding_spec` to avoid duplicating this. `_get_state_sharding_spec` is purely for `out_shardings`.

keras/src/backend/jax/trainer.py
```diff
@@ -301,6 +305,9 @@ def make_test_function(self, force=False):
                     non_trainable_shardings,
                     metrics_shardings,
                 )
+                state_shardings = jax_distribution_lib._to_abstract_shardings(
+                    state_shardings
+                )
```
hertschuh: Move to `_get_state_sharding_spec`.

keras/src/backend/jax/trainer.py
```diff
@@ -341,6 +348,9 @@ def predict_step(state, data):
                     trainable_shardings,
                     non_trainable_shardings,
                 )
+                state_shardings = jax_distribution_lib._to_abstract_shardings(
+                    state_shardings
+                )
```
hertschuh: Move to `_get_state_sharding_spec`.

hertschuh (commented):
Let's keep this to a minimum and remove `_to_backend_abstract_mesh`, which is not used right now.

But besides that, it's really close. Thanks!

keras/src/distribution/distribution_lib.py
bitloi: Please check again, I just fixed.

keras/src/backend/jax/distribution_lib.py
bitloi: Please check again, I just fixed.

keras/src/backend/jax/distribution_lib.py
bitloi: Please check again, I just fixed.

keras/src/backend/jax/trainer.py
bitloi: Please check again, I just fixed.

keras/src/backend/jax/trainer.py
bitloi: Please check again, I just fixed.

keras/src/backend/jax/trainer.py
bitloi: Please check again, I just fixed.

hertschuh:
@bitloi 

What makes you think you need to do something different on TPU?

The crash on TPU is unrelated AFAICT, conv transpose crash randomly.

hertschuh:
FYI https://github.com/keras-team/keras/pull/22117

hertschuh:
> FAILED keras/src/trainers/trainer_test.py::JAXTrainerCorrectnessTest::test_jit_fit_with_out_shardings_logic_distributed - RuntimeError: A jitted computation cannot contain AbstractMesh in in_shardings and out_shardings during compilation. You can use `jax.export` to  lower with an AbstractMesh and later compile with concrete devices.

It's not actually working. This errors calls into question the whole concept.

bitloi:
> > FAILED keras/src/trainers/trainer_test.py::JAXTrainerCorrectnessTest::test_jit_fit_with_out_shardings_logic_distributed - RuntimeError: A jitted computation cannot contain AbstractMesh in in_shardings and out_shardings during compilation. You can use `jax.export` to  lower with an AbstractMesh and later compile with concrete devices.
> 
> It's not actually working. This errors calls into question the whole concept.

The TPU failure was unrelated; the real issue was JAX not allowing AbstractMesh in jit’s in/out_shardings during compilation. I’ve switched to using the variables’ concrete shardings for `out_shardings` and pushed the fix.

bitloi:
I’ve fixed it. Would you mind reviewing it?

hertschuh:
@bitloi 

I cannot merge this.

First, the changes in `keras/src/backend/jax/trainer.py` are a no-op.

The changes in `keras/src/backend/jax/distribution_lib.py` amount to this:
```python
    if jax_utils.is_in_jax_tracing_scope():
        layout = _to_abstract_sharding(layout)
        return jax.lax.with_sharding_constraint(tensor, layout)
```
But this cannot work. It will fail in the same way that it failed in `trainer.py`. The only reason why nothing is failing right now is that we don't have a unit test to cover this code (which is a problem).

This issue https://github.com/keras-team/keras/issues/21027 was very much of an investigation, which is why it was not tagged "contributions welcome". I had no idea what `AbstractMesh` could help with in Keras. Now that I learned a bit more about it, it sounds like at this stage we cannot use it in any part of Keras.

I'm sorry for the time you spent on this.

bitloi:
> @bitloi
> 
> I cannot merge this.
> 
> First, the changes in `keras/src/backend/jax/trainer.py` are a no-op.
> 
> The changes in `keras/src/backend/jax/distribution_lib.py` amount to this:
> 
> ```python
>     if jax_utils.is_in_jax_tracing_scope():
>         layout = _to_abstract_sharding(layout)
>         return jax.lax.with_sharding_constraint(tensor, layout)
> ```
> 
> But this cannot work. It will fail in the same way that it failed in `trainer.py`. The only reason why nothing is failing right now is that we don't have a unit test to cover this code (which is a problem).
> 
> This issue #21027 was very much of an investigation, which is why it was not tagged "contributions welcome". I had no idea what `AbstractMesh` could help with in Keras. Now that I learned a bit more about it, it sounds like at this stage we cannot use it in any part of Keras.
> 
> I'm sorry for the time you spent on this.

@hertschuh Thanks for the review. You're right about the limitations. I found that JAX's explicit sharding mode (`jax.make_mesh` with `AxisType.Explicit`) might be the proper path forward, but that's a bigger architectural change. Can I continue working on it?

hertschuh:
> @hertschuh Thanks for the review. You're right about the limitations. I found that JAX's explicit sharding mode (`jax.make_mesh` with `AxisType.Explicit`) might be the proper path forward, but that's a bigger architectural change. Can I continue working on it?

Yes, if you want to. Can you explain a bit:
- what's the benefit?
- how it is a bigger architectural change?

hertschuh:
> **Benefit**: `jax.make_mesh()` uses `AbstractMesh` by default, so JIT cache stability comes for free — no wrapper needed.

But we just established that we can't use `AbstractMesh` in the Keras code right now.

> More importantly, explicit sharding propagates shardings through ops at trace time, so the Trainer wouldn't need `out_shardings` at all. JAX just infers them from inputs. It also means sharding mismatches get caught as trace-time errors instead of silently producing wrong layouts.

Well we did the reverse change: https://github.com/keras-team/keras/pull/21559/changes

The reasons were that it's cleaner and more readable.
 
> **Why it's bigger**: This PR only touches 3 call sites in the Trainer + `distribute_tensor` — we just wrap existing shardings with `AbstractMesh` at JIT boundaries. Explicit sharding would mean replacing `jax.sharding.Mesh` with `jax.make_mesh`, dropping `out_shardings` from all `jit()` calls, switching `device_put` to `reshard`, and reworking `DeviceMesh`/`TensorLayout` since `make_mesh` handles device assignment internally. There's also a maturity concern — some JAX ops (scatter, gather) still lack sharding rules and need `auto_axes` fallbacks.

What will switching from `device_put` to `reshard` do for us?

hertschuh:
> Anyway, can you let me know if you'd like me to focus on anything else for this PR? To be honest, I don't want to close this PR because it's my first contribution. :)

I'm not sure it really makes sense to reuse this PR for something fundamentally different though.

bitloi:
@hertschuh could you please allow me to implement another issue on this PR or let me know the fix I can work?

hertschuh:
> @hertschuh could you please allow me to implement another issue on this PR or let me know the fix I can work?

Ok. Go ahead and find another issue in the backlog.

bitloi:
> > @hertschuh could you please allow me to implement another issue on this PR or let me know the fix I can work?
> 
> Ok. Go ahead and find another issue in the backlog.

@hertschuh Can I implement this issue - https://github.com/keras-team/keras/issues/21150 ?

PR #22101, opened 2026-02-03
Implement MultiOptimizer for discriminative layer training

### LINKED ISSUE #21741: [Feature Request] Implement multiple optimizer
I would like to propose the addition of multi-optimizer setting with the option of changing the learning rate via callbacks for discriminative layer training purposes.

### MOTIVATION

There is currently no direct way of implementing discriminative training without going low level by customizing what happens in `model.fit()`;Even with that, it is still unclear whether it will work out of the box of if the learning rate can be changed via callbacks. I believe providing a way to use discriminative learning would greatly enhance the usability of keras as a whole

### PROPOSED CHANGE

Adding an optimizer that can be passed as an argument to `model.compile()`

```
    model = keras.Sequential([
        keras.layers.Dense(64, activation='relu', name='layer1'),
        keras.layers.Dense(32, activation='relu', name='layer2'),
        keras.layers.Dense(10, activation='softmax', name='output')
    ], name='example_model')
    
    model.build(input_shape=(None, 20))
    
    adam_opt = keras.optimizers.Adam(learning_rate=0.001)
    sgd_opt = keras.optimizers.SGD(learning_rate=0.01, momentum=0.9)
    
    layer1_vars = model.layers[0].trainable_variables
    layer2_and_output_vars = (
        model.layers[1].trainable_variables + 
        model.layers[2].trainable_variables
    )
    
    multi_optimizer = MultiOptimizer([
        (adam_opt, layer1_vars),
        (sgd_opt, layer2_and_output_vars)
    ])
    
    model.compile(
        optimizer=multi_optimizer,
        loss='sparse_categorical_crossentropy',
        metrics=['accuracy']
    )
```

This feature would make it so that, one can easily experiment with different training regimen easily without having to switch frameworks constantly.

timon0305:
Add MultiOptimizer class that enables using different optimizers for different variables/layers. This is useful for discriminative layer training where different learning rates or optimization algorithms are needed for different parts of a model.

Resolve #21741

### DIFF
--- a/keras/api/_tf_keras/keras/optimizers/__init__.py
+++ b/keras/api/_tf_keras/keras/optimizers/__init__.py
@@ -21,6 +21,7 @@ from keras.src.optimizers.lion import Lion as Lion
 from keras.src.optimizers.loss_scale_optimizer import (
     LossScaleOptimizer as LossScaleOptimizer,
 )
+from keras.src.optimizers.multi_optimizer import MultiOptimizer as MultiOptimizer
 from keras.src.optimizers.muon import Muon as Muon
 from keras.src.optimizers.nadam import Nadam as Nadam
 from keras.src.optimizers.optimizer import Optimizer as Optimizer
--- a/keras/api/optimizers/__init__.py
+++ b/keras/api/optimizers/__init__.py
@@ -21,6 +21,7 @@ from keras.src.optimizers.lion import Lion as Lion
 from keras.src.optimizers.loss_scale_optimizer import (
     LossScaleOptimizer as LossScaleOptimizer,
 )
+from keras.src.optimizers.multi_optimizer import MultiOptimizer as MultiOptimizer
 from keras.src.optimizers.muon import Muon as Muon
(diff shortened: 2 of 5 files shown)


hertschuh (changes requested):
Thanks for working on this!

In terms of how you configure the mapping from optimizers to variable, I don't think it should be done in an explicit way like this:
- This is too low level, as a user I don't know how to create that mapping or what all the variables are.
- At the time when I create my optimizer, the variables may not all have been created yet.

I can see 2 ways to configure this differently:
1. The regex config way like https://keras.io/api/distribution/layout_map/
2. Which a function that makes the decision for each variable like https://github.com/keras-team/keras/blob/5b7ab6f0de7d8f19f71301dc15967034b8d4801a/keras/src/optimizers/muon.py#L141

I think we should do both the following way:
- MultiOptimizer take a map like the layout map above from regex as string to optimizer instance (instances can be reused between multiple regexes)
- MultiOptimizer has a method that by default applies the regexes, but can be overridden to be anything you want

PR #22104, opened 2026-02-03
JAX trainer: respect trainable state during fit()

MalyalaKarthik66:
This PR addresses the TODO in the JAX trainer to respect the compiled trainable state during `model.fit()`.
The trainer now captures the effective set of trainable variables at the start of training and ensures that only those variables receive gradient updates.

This brings JAX behavior in line with the TensorFlow backend and common Keras workflows such as freezing layers after `compile()`.
JAX-specific regression tests are added to validate the fix.

### DIFF
--- a/keras/src/trainers/trainer.py
+++ b/keras/src/trainers/trainer.py
@@ -159,9 +159,8 @@ class Trainer:
             model.fit(x_train, y_train)
             ```
 
-            This behavior applies to all Keras backends and matches the behavior
-            of Keras 2 (`tf.keras`), where trainable variables are captured at
-            compile time (as also documented in the transfer learning guide).
+            This behavior applies to all Keras backends and is also documented
+            in the transfer learning guide.
         """
         optimizer = optimizers.get(optimizer)
         self.optimizer = optimizer


MalyalaKarthik66:
@hertschuh  Thanks for the detailed feedback — this is very helpful.

Regarding the TensorFlow comment: I was referring to tf.keras behavior rather than the current Keras 3 TensorFlow backend behavior.

You're right that if layers are frozen/unfrozen while keeping the same number of trainable variables, an index-based mask can silently become incorrect. I also agree that filtering gradients cannot fully address cases where new variables become trainable — at that point we would need gradients to be computed for a different variable set.

Given this, your suggestion about warning users instead of trying to emulate dynamic behavior makes sense. My main goal was to address the TODO and reduce surprising behavior for users.

Would you prefer that this PR pivot toward:
• Adding a warning if `trainable` flags change after compile without recompiling, or  
• Leaving the current behavior and documenting the expected compile → fit → modify → compile → fit workflow?

hertschuh:
> Would you prefer that this PR pivot toward:
> • Adding a warning if `trainable` flags change after compile without recompiling, or
> • Leaving the current behavior and documenting the expected compile → fit → modify → compile → fit workflow?

I think we should document "compile → fit → modify → compile → fit" regardless of whether the other bullet point is implemented.

What does Keras 2 do in this scenario?

MalyalaKarthik66:
Thanks — I’ve pivoted the PR to documentation only and explicitly documented the
`compile → fit → modify trainable → compile → fit` workflow in the `compile()` docstring.

Regarding Keras 2 (`tf.keras`): trainable variables are captured at `compile()` time, and changes to `layer.trainable` made after compilation are not honored until the model is recompiled. The documented workflow matches that behavior.

Please let me know if you’d prefer this note to live in a different location or be phrased differently.

keras/src/trainers/trainer.py
```diff
@@ -138,6 +138,30 @@ def compile(
...
+            layer.trainable = False  # or True
+
+            # Recompile for the change to take effect
+            model.compile(optimizer="adam", loss="mse")
+            model.fit(x_train, y_train)
+            ```
+
+            This behavior applies to all Keras backends and matches the behavior
+            of Keras 2 (`tf.keras`), where trainable variables are captured at
+            compile time (as also documented in the transfer learning guide).
```
hertschuh: Just remove the mention of Keras 2. A lot things match the behavior of Keras 2, but we don't mention it.

MalyalaKarthik66: Done — I’ve removed the mention of Keras 2.

PR #22106, opened 2026-02-03
fix: pytorch export pooling reshape layer

### LINKED ISSUE #22102: Pool>Project>Reshape layer stack produces frozen dimensions on export with PyTorch
For some reason, when used with unfixed/symbolic shapes, the combination of AveragePooling2D, Conv2D, and Reshape causes PyTorch's tracer to freeze dimensions, making the export with unfixed shapes impossible. If the dims are marked as static, the export will go through.
I believe, this is a Keras-side issue as an identical model in pure Pytorch will be converted just fine, retaining the unfixed nature of the dimensions.

Colab to reproduce:
https://colab.research.google.com/drive/1VGox3Uw4RrcPDSlcgHIMZgBz_JYEjIRm

P.S. here I use a pure call to `torch.onnx.export` as the Keras wrapper is being fixed at this moment: https://github.com/keras-team/keras/issues/22076

bittoby:
# Fix ONNX export failures with PyTorch backend for Pool→Project→Reshape layer stacks

## Summary

Resolves GitHub issue #22102 where Keras models with AveragePooling2D→Conv2D→Reshape layer combinations fail ONNX export with dynamic shape conflicts when using the PyTorch backend.

## Problem

When exporting Keras models to ONNX with `dynamic_shapes` specified, PyTorch's TorchDynamo tracer encounters conflicts between user-specified dynamic dimensions and statically inferred shapes. This occurs when:

Closes #22102

### DIFF
--- a/keras/src/backend/common/variables_test.py
+++ b/keras/src/backend/common/variables_test.py
@@ -1,4 +1,5 @@
 import itertools
+from unittest.mock import create_autospec
 
 import numpy as np
 import pytest
@@ -1197,8 +1198,6 @@ class TestStandardizeShapeWithTorch(test_case.TestCase):
         This validates the fix for GitHub issue #22102 where torch.SymInt
         objects from torch.export were causing "Constraints violated" errors.
         """
-        from unittest.mock import create_autospec
-
         import torch
 
         # Create a mock SymInt object
@@ -1214,10 +1213,6 @@ class TestStandardizeShapeWithTorch(test_case.TestCase):
         shape_with_multiple_sym_ints = (sym_int, sym_int2, 64)
         result = standardize_shape(shape_with_multiple_sym_ints)
         self.assertEqual(result, (None, None, 64))
-
-        # Verify torch.SymInt is imported and checked in the actual code
-        # by checking that torch backend is active
-        self.assertEqual(backend.backend(), "torch")
 
 
 @pytest.mark.skipif(
--- a/keras/src/layers/reshaping/reshape_test.py
+++ b/keras/src/layers/reshaping/reshape_test.py
@@ -5,6 +5,7 @@ from keras.src import Sequential
 from keras.src import backend
 from keras.src import layers
 from keras.src import ops
+from keras.src import random
 from keras.src import testing
 from keras.src.backend.common.keras_tensor import KerasTensor
 
@@ -122,8 +123,6 @@ class ReshapeTest(testing.TestCase):
 
     @pytest.mark.requires_trainable_backend
     def test_reshape_model_fit_with_varying_input_size_and_minus_one(self):
-        from keras.src import random
-
         def generator():
             yield (
                 ops.ones((1, 12, 2), dtype="float32"),
--- a/keras/src/ops/numpy_test.py
+++ b/keras/src/ops/numpy_test.py
@@ -3877,44 +3877,6 @@ class NumpyOneInputOpsCorrectnessTest(testing.TestCase):
         x = np.array([65504, 65504, 65504], dtype="float16")
         self.assertAllClose(knp.mean(x), np.mean(x))
 
-    @pytest.mark.skipif(
-        backend.backend() != "torch",
-        reason="Test for torch mean ONNX compatibility",
-    )
-    def test_mean_torch_onnx_compatibility(self):
-        """Test mean() works with ONNX export (fix for GlobalAveragePooling2D).
-
-        Validates that mean() doesn't pass dtype kwarg to torch.mean(),
-        which causes ONNX export errors.
-        """
-        import torch
-
-        # Test that it works during torch.export (ONNX export simulation)
-        try:
-            # Test mean over multiple axes (GlobalAveragePooling2D use case)
-            x = np.random.randn(2, 8, 8, 64).astype(np.float32)
-            result = knp.mean(x, axis=[1, 2])
-            expected = np.mean(x, axis=(1, 2))
-            self.assertAllClose(result, expected)
-
-            class MeanModel(torch.nn.Module):
-                def forward(self, x):
-                    # Use torch.mean directly (as ops.mean does internally)
-                    return torch.mean(x, dim=[1, 2], keepdim=False)
-
-            model = MeanModel()
-            sample = torch.randn(1, 8, 8, 64)
-
-            # This should not raise dtype kwarg error
-            exported = torch.export.export(model, (sample,), strict=False)
-            self.assertIsNotNone(exported)
-        except TypeError as e:
-            if "dtype" in str(e):
-                self.fail(f"mean() failed with dtype kwarg error: {e}")
-            pytest.skip(f"torch.export not available: {e}")
-        except Exception as e:
-            pytest.skip(f"torch.export not available: {e}")
-
     def test_array_split(self):
         x = np.array([[1, 2, 3], [4, 5, 6]])
 


keras/src/backend/torch/onnx_export_patch.py
```diff
@@ -0,0 +1,166 @@
...
+This module provides a patched version of torch.onnx.export that automatically
+falls back to TorchScript export when TorchDynamo export fails with dynamic
+shape conflicts, specifically addressing GitHub issue #22102.
+"""
+
+import contextlib
+import warnings
+
+
+def patched_onnx_export(
```
hertschuh: I'm very confused about this.

If you're changing `torch.onnx.export`, why not upstream it and change it in `torch` instead?

But it doesn't look like you're really changing it. You're calling the original one with different parameters. So why patch? Why not just call the original?

keras/src/layers/reshaping/reshape.py
```diff
@@ -61,6 +61,36 @@ def compute_output_spec(self, inputs):
...
+
+                unknown_dim = total_elements // known_elements
+                resolved_shape = [
+                    unknown_dim if dim == -1 else dim
+                    for dim in self.target_shape
+                ]
+            else:
+                resolved_shape = list(self.target_shape)
+
+            return torch.reshape(inputs, [batch_size] + resolved_shape)
```
hertschuh: This is exactly what the "keras fallback" below does.

In particular `operation_utils.compute_reshape_output_shape` does exactly that resolved shape calculation that you're doing. AFAICT this is a no-op change.

Crystora:
@hertschuh I fixed the issue with possible cases. Please review my PR. Thanks.

Crystora:
@hertschuh Please review my PR. Thanks.

keras/src/ops/numpy_test.py
```diff
@@ -3877,6 +3877,45 @@ def test_mean(self):
...
+        Validates that mean() doesn't pass dtype kwarg to torch.mean(),
+        which causes ONNX export errors.
+        """
+        import torch
+
+        # Test mean over multiple axes (GlobalAveragePooling2D use case)
+        x = np.random.randn(2, 8, 8, 64).astype(np.float32)
+        result = knp.mean(x, axis=[1, 2])
+        expected = np.mean(x, axis=(1, 2))
+        self.assertAllClose(result, expected)
```
hertschuh: This is indeed a gap in the current unit tests, can you move this to here:

https://github.com/keras-team/keras/blob/master/keras/src/ops/numpy_test.py#L3901

keras/src/ops/numpy_test.py
```diff
@@ -3877,6 +3877,45 @@ def test_mean(self):
...
+
+            # This should not raise dtype kwarg error
+            exported = torch.export.export(model, (sample,), strict=False)
+            self.assertIsNotNone(exported)
+        except TypeError as e:
+            if "dtype" in str(e):
+                self.fail(f"mean() failed with dtype kwarg error: {e}")
+            pytest.skip(f"torch.export not available: {e}")
+        except Exception as e:
+            pytest.skip(f"torch.export not available: {e}")
```
hertschuh: I'm confused about this test:
- it doesn't use keras at all, just pure torch
- if anything goes wrong, it skips anyway

I don't think this belongs here. So please remove.

keras/src/layers/reshaping/reshape.py
```diff
@@ -61,6 +61,29 @@ def compute_output_spec(self, inputs):
...
+            if is_tracing:
+                # Don't extract batch_size separately - build shape directly
+                # This mimics what pure PyTorch does: x.reshape(n, -1, c)
+                # where n and c come from x.shape
+
+                # Build output shape using inputs.shape[0] directly in the list
+                output_shape = [inputs.shape[0]] + list(self.target_shape)
+
+                # Use torch.reshape
+                return torch.reshape(inputs, output_shape)
```
hertschuh: I don't think any of this should be needed:
1. Can you test without?
2. In general, there should not be any backend specific code in layers. What this usually means is that something else (usually an op) needs fixing.
3. Here is why I don't think this is needed:

```python
                # Don't extract batch_size separately - build shape directly
                # This mimics what pure PyTorch does: x.reshape(n, -1, c)
                # where n and c come from x.shape

                # Build output shape using inputs.shape[0] directly in the list
                output_shape = [inputs.shape[0]] + list(self.target_shape)
```

> This mimics what pure PyTorch does: x.reshape(n, -1, c) where n and c come from x.shape

But `c` does *not* come from `x.shape`, it comes from `self.target_shape` in your code. So it's not going to propagate `SymInt`s outside of the batch size.

```python
                # Use torch.reshape
                return torch.reshape(inputs, output_shape)
```

`keras.ops.reshape` just calls `torch.reshape`, so that wouldn't fix anything.

(the original code below)

```python
        return ops.reshape(
            inputs, (ops.shape(inputs)[0],) + potentially_resolved_target_shape
        )
```
`keras.ops.shape` on torch *will* return the `SymInt`s, so the `ops.shape(inputs)[0]` will be a `SymInt`, which will passed to `keras.ops.reshape` and therefore `torch.reshape`.

keras/src/layers/reshaping/reshape_test.py
```diff
@@ -136,3 +136,43 @@ def generator():
         model = Sequential([layer])
         model.compile(loss="mean_squared_error")
         model.fit(generator(), steps_per_epoch=2, epochs=1)
+
+    @pytest.mark.skipif(
+        backend.backend() != "torch",
+        reason="Test for torch.SymInt detection during ONNX export",
+    )
```
hertschuh: Make this test backend independent.

All you need to do is:
- rename something like `test_reshape_with_dynamic_dimensions`
- remove `import torch`
- add `from keras.src import random`
- replace `torch.randn` with `keras.random.normal` with the shape as a tuple.

That being said, the test right above already tests with one dynamic dimension. So what is different about this test that was not covered above? Can the test above be augmented?

keras/src/backend/common/variables_test.py
```diff
@@ -1181,6 +1181,36 @@ def test_standardize_shape_with_torch_size(self):
...
+
+    @parameterized.named_parameters(
+        ("all_dynamic", (None, None, None, 64), (None, None, None, 64)),
+        ("mixed", (None, 224, 224, 3), (None, 224, 224, 3)),
+        ("all_static", (1, 224, 224, 3), (1, 224, 224, 3)),
+    )
+    def test_standardize_shape_preserves_none(self, input_shape, expected):
+        """Test that None dimensions are preserved correctly."""
+        result = standardize_shape(input_shape)
+        self.assertEqual(result, expected)
```
hertschuh: This test is not torch specific, please move to line 374.

keras/src/backend/common/variables_test.py
```diff
@@ -1181,6 +1181,36 @@ def test_standardize_shape_with_torch_size(self):
...
+        # but we can verify the code path exists and handles None correctly
+
+        # Test that the fix preserves None (which SymInt becomes)
+        shape_with_none = (None, None, 64)
+        result = standardize_shape(shape_with_none)
+        self.assertEqual(result, (None, None, 64))
+
+        # Verify torch.SymInt is imported and checked in the actual code
+        # by checking that torch backend is active
+        self.assertEqual(backend.backend(), "torch")
```
hertschuh: I understand that it's not easy to instantiate a `SymInt`.

If you can't instantiate a `SymInt`, then this test is the exact same test as `test_standardize_shape_preserves_none` below, so you can just remove it.

But you can try this:
```python
from unittest.mock import create_autospec
...
sym_int = create_autospec(torch.SymInt)
shape_with_sym_int = (sym_int, 64)
...
```

Crystora:
@hertschuh I have fixed what you mentioned. Please check my PR again. Thanks.

keras/src/backend/common/variables_test.py
```diff
@@ -1181,6 +1191,34 @@ def test_standardize_shape_with_torch_size(self):
         for d in standardized_shape:
             self.assertIsInstance(d, int)
 
+    def test_standardize_shape_with_torch_symint(self):
+        """Test that torch.SymInt dimensions are converted to None.
+
+        This validates the fix for GitHub issue #22102 where torch.SymInt
+        objects from torch.export were causing "Constraints violated" errors.
+        """
+        from unittest.mock import create_autospec
```
hertschuh: Move this import `from unittest.mock import create_autospec` to the top.

keras/src/backend/common/variables_test.py
```diff
@@ -1181,6 +1191,34 @@ def test_standardize_shape_with_torch_size(self):
...
+
+        # Test with multiple SymInts
+        sym_int2 = create_autospec(torch.SymInt, instance=True)
+        shape_with_multiple_sym_ints = (sym_int, sym_int2, 64)
+        result = standardize_shape(shape_with_multiple_sym_ints)
+        self.assertEqual(result, (None, None, 64))
+
+        # Verify torch.SymInt is imported and checked in the actual code
+        # by checking that torch backend is active
+        self.assertEqual(backend.backend(), "torch")
```
hertschuh: Remove this code, this will always succeed base on the `@pytest.mark.skipif(backend.backend() != "torch" ...)` decorator on the class.

keras/src/layers/reshaping/reshape_test.py
```diff
@@ -122,6 +122,8 @@ def test_reshape_sets_static_shape(self):
 
     @pytest.mark.requires_trainable_backend
     def test_reshape_model_fit_with_varying_input_size_and_minus_one(self):
+        from keras.src import random
```
hertschuh: Move this import `from keras.src import random` to the top, or just use `ops.ones` instead of `random.normal`.

keras/src/ops/numpy_test.py
```diff
@@ -3925,6 +3877,44 @@ def test_mean(self):
...
+
```
hertschuh: Remove this test. This is not a Keras test, this is a pure torch export test.

Crystora:
@hertschuh Can you check my PR again? Thanks.

hertschuh (approved):
Thanks!

Crystora:
@hertschuh  Thank you for your help.

PR #22110, opened 2026-02-04
Fix bounding box docstring references

Kh9705:
PR description:

Replaced placeholder “keras-io” references with the correct Keras API reference for bounding box formats.
Cleaned up docstring text 
No functional changes; docstring-only.


Kh9705:
Thanks for the review and the suggestion!

I’ve updated the docstring to include examples of supported bounding box formats for better clarity and consistency.

I’ve also signed the CLA,please let me know if anything else is needed.

hertschuh:
FYI I'm adding documentation for `keras.utils.bounding_boxes.convert_format` so that your docs can point to it.

https://github.com/keras-team/keras-io/pull/2282

PR #22112, opened 2026-02-05
feat: add depth_to_space and space_to_depth ops

### LINKED ISSUE #20622: Key space/depth operations commonly used in SISR models are missing
Hi! I would like to ask if it would be possible to add support for the `depth to space` and `space to depth` operations commonly found on many SISR models.

While you can achieve the same result with some clever usage of `keras.ops.reshape` and `keras.ops.transpose`, both TF and Pytorch offer them as easy to use functions:

https://www.tensorflow.org/api_docs/python/tf/nn/depth_to_space
https://www.tensorflow.org/api_docs/python/tf/nn/space_to_depth

https://pytorch.org/docs/stable/generated/torch.nn.PixelShuffle.html
https://pytorch.org/docs/stable/generated/torch.nn.PixelUnshuffle.html

I've been able to find the same operations in the PIX library as well:

https://dm-pix.readthedocs.io/en/latest/api.html#depth-and-space-transformations

This exact same depth to space transformation can be found in the following example:

https://keras.io/examples/vision/super_resolution_sub_pixel/

Thanks in advance =)

timon0305:
## Summary

- Adds `depth_to_space` (pixel shuffle) and `space_to_depth` (pixel unshuffle) operations to `keras.ops.nn`
- These operations rearrange data between depth and spatial dimensions, commonly used in super-resolution models (ESPCN-style upscaling)
- Implemented for all backends: TensorFlow, PyTorch, JAX, NumPy, and OpenVINO

## Implementation Details

- `depth_to_space`: Rearranges `(batch, height, width, depth)` to `(batch, height*block_size, width*block_size, depth/(block_size^2))`
- `space_to_depth`: Rearranges `(batch, height, width, depth)` to `(batch, height/block_size, width/block_size, depth*(block_size^2))`
- Both operations support `channels_last` (NHWC) and `channels_first` (NCHW) data formats

Fixes #20622

### DIFF
--- a/keras/src/ops/nn.py
+++ b/keras/src/ops/nn.py
@@ -3316,7 +3316,7 @@ def _unfold(x, kernel_size, dilation=1, padding=0, stride=1):
 
 
 class DepthToSpace(Operation):
-    def __init__(self, block_size, data_format=None, *, name=None):
+    def __init__(self, block_size, data_format="channels_last", *, name=None):
         super().__init__(name=name)
         self.block_size = block_size
         self.data_format = standardize_data_format(data_format)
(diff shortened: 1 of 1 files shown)


hertschuh (approved):
Thank you for adding this feature!

PR #22113, opened 2026-02-05
Add ScheduleFreeAdamW optimizer

### LINKED ISSUE #19454: Add AdamWScheduleFree
Is possible add the new optimizer detailed in https://github.com/facebookresearch/schedule_free?

timon0305:
## Summary
- Implements the Schedule-Free AdamW optimizer from Facebook Research ([arXiv:2405.15682](https://arxiv.org/abs/2405.15682))
- Eliminates the need for learning rate schedules by maintaining interpolation and averaging of parameter sequences
- Supports train/eval mode switching, warmup, and weight decay
- Compatible with all Keras backends (TensorFlow, JAX, PyTorch)

## Test plan
- [x] Unit tests for config serialization
- [x] Single step and multiple steps training
- [x] Weight decay with exclusion
- [x] Train/eval mode switching
- [x] Warmup functionality
- [x] Integration with Keras model
- [x] Gradient clipping (norm and value)
- [x] Tested on all backends (TensorFlow, JAX, PyTorch)

Closes #19454

### DIFF
--- a/keras/src/optimizers/schedule_free_adamw.py
+++ b/keras/src/optimizers/schedule_free_adamw.py
@@ -111,7 +111,7 @@ class ScheduleFreeAdamW(optimizer.Optimizer):
         # Initialize momentum to match the initial parameter values
         for momentum, var in zip(self._momentums, var_list):
             if momentum is not None:
-                self.assign(momentum, var.value)
+                self.assign(momentum, ops.copy(var))
 
     def update_step(self, gradient, variable, learning_rate):
         """Update step given gradient and the associated model variable."""


keras/src/optimizers/schedule_free_adamw.py
```diff
@@ -0,0 +1,298 @@
...
+        self.swap_to_train()
+
+    def eval(self):
+        """Alias for swap_to_eval()."""
+        self.swap_to_eval()
+
+    @property
+    def is_training_mode(self):
+        """Returns True if optimizer is in training mode."""
+        return self._is_training_mode
```
hertschuh: These don't exist in Keras, they will never be called. Remove.

keras/src/optimizers/schedule_free_adamw.py
```diff
@@ -0,0 +1,298 @@
...
+            ema_overwrite_frequency=ema_overwrite_frequency,
+            loss_scale_factor=loss_scale_factor,
+            gradient_accumulation_steps=gradient_accumulation_steps,
+            **kwargs,
+        )
+        self.beta_1 = beta_1
+        self.beta_2 = beta_2
+        self.epsilon = epsilon
+        self.warmup_steps = warmup_steps
+        self._is_training_mode = True
```
hertschuh: Remove `self._is_training_mode`. Optimizers are only called in training mode.

keras/src/optimizers/schedule_free_adamw.py
```diff
@@ -0,0 +1,298 @@
...
+
+        Args:
+            var_list: list of model variables to build optimizer variables on.
+        """
+        if self.built:
+            return
+        super().build(var_list)
+        # z: auxiliary variable (gradient steps applied here)
+        # v: second moment estimate (velocity)
+        # The model variables act as 'y' during training and 'x' during eval
```
hertschuh: Remove this comment:
```
# The model variables act as 'y' during training and 'x' during eval
```
Optimizers are not used during eval.

keras/src/optimizers/schedule_free_adamw.py
```diff
@@ -0,0 +1,298 @@
...
+                self._v.append(
+                    self.add_variable_from_reference(
+                        reference_variable=var,
+                        name="velocity",
+                        initializer="zeros",
+                    )
+                )
+            else:
+                self._z.append(None)
+                self._v.append(None)
```
hertschuh: All of this can be replaced with:

```python
        self._momentums, self._velocities = self.add_optimizer_variables(
            var_list, ["momentum", "velocity"]
        )
```

Also, I'd like you to rename `_z` to `_momentum` and `_v` to `_velocities`. I believe that's what they are. It improves readability and consistency with the current `AdamW` code.

keras/src/optimizers/schedule_free_adamw.py
```diff
@@ -0,0 +1,298 @@
...
+
+        # Initialize z to match the initial parameter values
+        # We use ops.copy to ensure no aliasing issues with JAX
```
hertschuh: I'm not sure what you mean by "aliasing issues". But arrays are immutable in JAX (and Tensorflow). So you can never have any issue.

keras/src/optimizers/schedule_free_adamw.py
```diff
@@ -0,0 +1,298 @@
...
+        for i, var in enumerate(var_list):
+            z = self._z[i]
+            if z is not None:
+                self.assign(z, ops.copy(var))
```
hertschuh:
```python
for momentum, var in zip(self._momentums, var_list):
    momentum.assign(var.value)
```

keras/src/optimizers/schedule_free_adamw.py
```diff
@@ -0,0 +1,298 @@
...
+            warmup_steps = ops.cast(self.warmup_steps, variable.dtype)
+            warmup_factor = ops.minimum(local_step / warmup_steps, 1.0)
+            lr = lr * warmup_factor
+
+        var_index = self._get_variable_index(variable)
+        z = self._z[var_index]
+        v = self._v[var_index]
+
+        # Store z_old before any updates (use copy to avoid aliasing in JAX)
+        z_old = ops.copy(z)
```
hertschuh: I don't think you need this at all. But if there's a concern, you should do this instead:
```python
z_old = z.value
```

keras/src/optimizers/schedule_free_adamw.py
hertschuh: These don't exist in Keras, they will never be called. Remove.

keras/src/optimizers/schedule_free_adamw.py
```diff
@@ -0,0 +1,298 @@
...
+
+        # Apply warmup
+        if self.warmup_steps > 0:
+            warmup_steps = ops.cast(self.warmup_steps, variable.dtype)
+            warmup_factor = ops.minimum(local_step / warmup_steps, 1.0)
+            lr = lr * warmup_factor
+
+        var_index = self._get_variable_index(variable)
+        z = self._z[var_index]
+        v = self._v[var_index]
```
hertschuh: Rename `momentum` and `velocity`.

hertschuh (commented):
Thanks for adding this!

keras/src/optimizers/schedule_free_adamw.py
timon0305: Kept `self.assign(momentum, ops.copy(var))` instead of `momentum.assign(var.value)` due to JAX buffer donation issues

keras/src/optimizers/schedule_free_adamw.py
```diff
@@ -0,0 +1,198 @@
...
+            return
+        super().build(var_list)
+        self._momentums, self._velocities = self.add_optimizer_variables(
+            var_list, ["momentum", "velocity"]
+        )
+
+        # Initialize momentum to match the initial parameter values
+        for momentum, var in zip(self._momentums, var_list):
+            if momentum is not None:
+                self.assign(momentum, ops.copy(var))
```
hertschuh: Last nitpick: replace `ops.copy(var)` with `var.value`, it's more explicit.

hertschuh (approved):
Thanks!

keras/src/optimizers/schedule_free_adamw.py
```diff
@@ -0,0 +1,198 @@
...
+            return
+        super().build(var_list)
+        self._momentums, self._velocities = self.add_optimizer_variables(
+            var_list, ["momentum", "velocity"]
+        )
+
+        # Initialize momentum to match the initial parameter values
+        for momentum, var in zip(self._momentums, var_list):
+            if momentum is not None:
+                self.assign(momentum, var.value)
```
hertschuh: Now I understand, you actually do need a copy here. So please change it back and sorry for the misunderstanding.

It's actually very unintuitive, so please also add a comment, something like: `because the buffers for both the variable itself and momentum will be donated in the training loop, they need to be separate arrays`.

PR #22115, opened 2026-02-05
Implement GRU for PyTorch backend

MarcosAsh:
Implement `gru()` for PyTorch backend.

- Add `gru()` following the existing `lstm()` pattern
- Add `prepare_gru_weights()` to handle Keras→PyTorch gate reordering ([z,r,h] → [r,z,h])
- Add `_cudnn_gru()` for cuDNN-accelerated execution
- Support masking, `go_backwards`, `return_sequences`

Only `reset_after=True` supported (cuDNN semantics).

### DIFF
--- a/keras/src/backend/torch/rnn.py
+++ b/keras/src/backend/torch/rnn.py
@@ -821,12 +821,8 @@ def prepare_gru_params(kernel, recurrent_kernel, bias, device):
     z_r, r_r, h_r = torch.chunk(recurrent_kernel, 3, dim=1)
 
     # Reorder to PyTorch format [r, z, h] and transpose
-    weight_ih = (
-        torch.cat([r_k, z_k, h_k], dim=1).T.contiguous().to(device)
-    )
-    weight_hh = (
-        torch.cat([r_r, z_r, h_r], dim=1).T.contiguous().to(device)
-    )
+    weight_ih = torch.cat([r_k, z_k, h_k], dim=1).T.contiguous().to(device)
+    weight_hh = torch.cat([r_r, z_r, h_r], dim=1).T.contiguous().to(device)
 
     if bias is not None:
         # bias shape is (2, 3*units) for reset_after=True
@@ -835,12 +831,8 @@ def prepare_gru_params(kernel, recurrent_kernel, bias, device):
         z_bh, r_bh, h_bh = torch.chunk(bias[1], 3)
 
         # Reorder to [r, z, h]
-        bias_ih = (
-            torch.cat([r_bi, z_bi, h_bi]).contiguous().to(device)
-        )
-        bias_hh = (
-            torch.cat([r_bh, z_bh, h_bh]).contiguous().to(device)
-        )
+        bias_ih = torch.cat([r_bi, z_bi, h_bi]).contiguous().to(device)
+        bias_hh = torch.cat([r_bh, z_bh, h_bh]).contiguous().to(device)
     else:
         hidden_size = recurrent_kernel.shape[0]
         bias_ih = torch.zeros(
@@ -869,9 +861,7 @@ def _cudnn_gru(
     elif initial_state.dim() == 3 and initial_state.shape[1] == 1:
         initial_state = initial_state.permute(1, 0, 2)
 
-    params = prepare_gru_params(
-        kernel, recurrent_kernel, bias, device
-    )
+    params = prepare_gru_params(kernel, recurrent_kernel, bias, device)
 
     # Use functional GRU to maintain gradient flow through weight tensors
     outputs, h_n = torch._VF.gru(
@@ -881,7 +871,7 @@ def _cudnn_gru(
         bias is not None,  # has_biases
         1,  # num_layers
         0.0,  # dropout
-        False,  # training (cuDNN dropout, not Keras dropout)
+        torch.is_grad_enabled(),  # training: must be True for backward pass
         False,  # bidirectional
         True,  # batch_first
     )


keras/src/backend/torch/rnn.py
```diff
@@ -741,5 +741,308 @@ def _cudnn_lstm(
...
+    device = get_device()
+
+    from keras.src.backend.torch import Variable
+
+    if isinstance(kernel, Variable):
+        kernel = kernel.value
+    if isinstance(recurrent_kernel, Variable):
+        recurrent_kernel = recurrent_kernel.value
+    if isinstance(bias, Variable):
+        bias = bias.value
```
hertschuh: I understand this was done this way for `lstm`. But I would do it this way, which is more general (`convert_to_tensor` unwraps variables):
```python
kernel = convert_to_tensor(kernel)
recurrent_kernel = convert_to_tensor(recurrent_kernel)
bias = convert_to_tensor(bias)
```

Can you also change it for `lstm()`?

MarcosAsh:
Hi, just did the changes thanks for the review!

MarcosAsh:
I also did an extension for when reset_after=True. I did it on another pr as I wasn't sure if it would be useful.

hertschuh (dismissed):
Thanks for the implementing this!

hertschuh:
Oh, the tests are failing wth CUDA:
```
>           raise ValueError(error_message)
E           ValueError: You are passing a RNN mask that does not correspond to right-padded sequences, while using cuDNN, which is not supported. With cuDNN, RNN masks can only be used for right-padding, e.g. `[[True, True, False, False]]` would be a valid mask, but any mask that isn't just contiguous `True`'s on the left and contiguous `False`'s on the right would be invalid. You can pass `use_cudnn=False` to your RNN layer to stop using cuDNN (this may be slower).

keras/src/backend/torch/rnn.py:451: ValueError
```

MarcosAsh:
I fixed the cuda error

hertschuh:
> I also did an extension for when reset_after=True. I did it on another pr as I wasn't sure if it would be useful.

So you integrated the changes from the other PR here?

MarcosAsh:
No, the reset_after=False changes are in a separate PR. This PR only supports reset_after=True. Should I integrate them here or keep them separate? I closed the pr as i couldn't get the tests to work and didn't want to fill up the pr requests.

hertschuh:
> No, the reset_after=False changes are in a separate PR. This PR only supports reset_after=True. Should I integrate them here or keep them separate? I closed the pr as i couldn't get the tests to work and didn't want to fill up the pr requests.

Ok. Please make it a separate PR, but after this one is submitted so that I can see the diff.

MarcosAsh:
Ok will do thanks for your patience

hertschuh:
Can you look into the errors again?

MarcosAsh:
Just pushed the fixes. The implementation now uses native PyTorch ops instead of cuDNN, which should resolve all the failing tests.

hertschuh:
Some tests failed.

MarcosAsh:
I added a check in the GRU and LSTM layers to skip it when zero_output_for_mask=True with a mask, falling back to the       
  generic rnn() path which handles it correctly.

hertschuh:
There are still a couple of errors on CUDA, but it looks like they should be fairly easy to fix.

Thanks!

MarcosAsh:
Hey just completed the changes I think that should be it! Thanks for all the help on this pr.

MarcosAsh:
Hey all tests pass. Tell me if i should make any more changes. Thanks!

keras/src/backend/torch/rnn.py
```diff
@@ -579,65 +561,84 @@ def lstm(
     unroll=False,
     batch_first=True,
 ):
-    cudnn_supported = cudnn_ok(
-        activation,
-        recurrent_activation,
-        unroll,
-        use_bias=bias is not None,
-    )
+    """Native PyTorch LSTM with pre-computed input projections.
```
hertschuh: So this CL is disabling the CUDNN implementation of LSTM. In fact `_cudnn_lstm` is all dead code now.

Why is that? It seems like it was working, and it was probably a lot faster.

keras/src/layers/rnn/gru.py
```diff
@@ -549,7 +549,10 @@ def inner_loop(self, sequences, initial_state, mask, training=False):
             initial_state = initial_state[0]
         if tree.is_nested(mask):
             mask = mask[0]
-        if self.use_cudnn in ("auto", True):
+        # Skip the fast path when zero_output_for_mask is enabled with a
+        # mask, since the backend function does not support it.
+        use_fast_path = not (self.zero_output_for_mask and mask is not None)
+        if use_fast_path and self.use_cudnn in ("auto", True):
```
hertschuh: The comment seems to be contradicting the actual code.

Also, what do you call the "fast path"?

MarcosAsh:
Sorry about that, I didn't mean to disable cuDNN for LSTM. That happened while I was debugging the CUDA failures. I'll revert all the LSTM changes and keep this PR scoped to just adding GRU.  For the comment, I'll replace fast path with clearer wording. I'll also drop the lstm.py change since it's not related to GRU.

keras/src/backend/torch/rnn.py
```diff
@@ -741,5 +741,330 @@ def _cudnn_lstm(
...
+    initial_state,
+    kernel,
+    recurrent_kernel,
+    bias,
+    mask,
+    batch_first,
+    go_backwards,
+    return_sequences,
+    device,
+):
```
hertschuh, line 856: I'm confused. This is not used.

keras/src/layers/rnn/gru.py
```diff
@@ -549,7 +549,11 @@ def inner_loop(self, sequences, initial_state, mask, training=False):
             initial_state = initial_state[0]
         if tree.is_nested(mask):
             mask = mask[0]
-        if self.use_cudnn in ("auto", True):
+        # When zero_output_for_mask is enabled with a mask, fall back to
+        # the generic rnn() loop because the backend GRU does not zero
+        # outputs for masked timesteps.
+        skip_backend_gru = self.zero_output_for_mask and mask is not None
+        if not skip_backend_gru and self.use_cudnn in ("auto", True):
```
hertschuh:
```python
        # When zero_output_for_mask is enabled with a mask, fall back to
        # the generic rnn() loop because the backend GRU does not zero
        # outputs for masked timesteps.
```

Is this true for Tensorflow too?

The way it worked in the past is that:
- if you pass `use_cudnn=True` and it's not supported (because of recurrent dropout) it fails telling you it cannot honor `use_cudnn=True`
- if you pass `use_cudnn="auto"` then it automatically selects the implementation that works

But now, with the new code, you silently turn off cudnn without telling the user about it. It will probably be slower and it would take a lot of time for somebody to figure out why.

We should keep the same approach as before. But the first question is whether the criteria is the same for TF and Torch.

keras/src/backend/torch/rnn.py
MarcosAsh: I restructured gru() to actually call _cudnn_gru

keras/src/layers/rnn/gru.py
MarcosAsh: Yes, TF handles it the same way its backend gru() doesn't have zero_output_for_mask either, it just raises NotImplementedError and the layer falls back to the generic rnn() loop. Removed the skip_backend_gru check so the layer now matches LSTM's inner_loop exactly. The criteria are the same for both backends now.

MarcosAsh:
I made some changes to address the issues

hertschuh:
So the tests fail with CUDA. But you remove the check for the mask. Is that why? Do you need to add the test back to skip the CUDA implementation if there is a mask?

MarcosAsh:
Yeah you're right, I'll add the mask check back. The cuDNN path can't handle non right padded masks so it needs to fall back to the generic loop when there's a mask.

hertschuh:
@MarcosAsh 

So the tests fail on GPU.

For these two, we can simply change the atol / rtol to 1e-5 I think:
- keras/src/layers/rnn/gru_test.py::GRUTest::test_correctness0
- keras/src/layers/rnn/gru_test.py::GRUTest::test_correctness1

But this other one is kind of way off:
keras/src/layers/rnn/bidirectional_test.py::SimpleRNNTest::test_correctness

And than there are some other errors.

MarcosAsh:
Bumped the tolerance to 1e-5 for test_correctness. The bidirectional issue was the cuDNN path flipping outputs back to original time order on go_backwards, but the generic loop returns them in reversed order so the merge was off. Removed that flip. Ill fix any errors after the CI runs.

MarcosAsh:
After the tests failed on GPU I think I figured out the issues. So I switched from creating a torch.nn.GRU module to calling torch._VF.gru directly. My previous approach copied weights into a fresh module with copy_() under torch.no_grad(), which severed the autograd graph and caused No gradients provided for any variable during training. torch._VF.gru is the functional API that torch.nn.GRU uses internally it still gets cuDNN acceleration but the weight tensors stay in the computation graph so gradients flow back to the Keras variables. I also bumped the bidirectional GRU test tolerance to 1e-5 (max diff was ~4.2e-6).

hertschuh:
There is one last failure:

```
FAILED keras/src/layers/rnn/gru_test.py::GRUTest::test_basics - RuntimeError: cudnn RNN backward can only be called in training mode
```

MarcosAsh:
Ok I fixed it!

hertschuh:
The tests pass, which is great.

Can you reformat the code? Thanks!

hertschuh:
The code format check still fails, do you understand why?

MarcosAsh:
Just fixed it i think it was ruff 0.9.2 wants to reformat the actual source file the same as the issue on the geomspace pr

MarcosAsh:
Thanks for taking the time to review this PR so many times!

hertschuh (approved):
Thanks!

PR #22118, opened 2026-02-06
Fix custom layer *_mask parameters incorrectly treated as masking arguments

### LINKED ISSUE #21154: Custom layer with parameter ending in _mask is dysfunctional
Repro Colab: https://colab.research.google.com/drive/14n21B0Y6WE3EUj28igcAqKT4T_44LULs?usp=sharing

If a parameter in a custom layer is called *_mask, then it will not work with the layer's compute_output_shapes.
The solution is either to rename the parameter or to remove it altogether from the  params of compute_output_shapes function (which is allowed even though params of compute_output_shapes are supposed to shadow those of the layer).

This is a very surprising behavior and kind of odd not to be able to name layer arguments in a certain way. If the behavior was introduced for internal reasons (maybe handling of keras masking), see if it still needed and maybe if it could be relaxed to just "mask" instead of *_mask

MalyalaKarthik66:
This PR fixes an issue where custom layer parameters ending with `_mask` were incorrectly treated as special masking arguments.

The change restricts special masking behavior to only the exact argument name `mask`, allowing parameters like `attention_mask` and `padding_mask` to be used as regular layer arguments.

Includes regression tests to ensure custom *_mask parameters work correctly while preserving existing masking behavior.

Fixes #21154

### DIFF
--- a/keras/src/layers/layer_test.py
+++ b/keras/src/layers/layer_test.py
@@ -1854,39 +1854,20 @@ class LayerTest(testing.TestCase):
         y2 = model(sample_input)
         self.assertAllClose(y2, sample_input)
 
-    def test_functional_model_with_attention_mask_arg(self):
-        class CustomLayer(layers.Layer):
-            def call(self, x, attention_mask=None):
-                if attention_mask is not None:
-                    return x * ops.cast(attention_mask, x.dtype)
-                return x
-
-        x = Input(shape=(3,))
-        mask = Input(shape=(1,))
-        y = CustomLayer()(x, attention_mask=mask)
-        model = Model([x, mask], y)
-        out = model(
-            [
-                np.ones((2, 3), dtype="float32"),
-                np.ones((2, 1), dtype="float32"),
-            ]
-        )
-        self.assertEqual(out.shape, (2, 3))
-
     def test_layer_build_with_attention_mask_arg(self):
+        test = self
+
         class CustomLayer(layers.Layer):
-            def call(self, x, attention_mask=None):
+            def call(self, inputs, attention_mask=None):
                 if attention_mask is not None:
-                    return x * ops.cast(attention_mask, x.dtype)
-                return x
+                    return inputs * ops.cast(attention_mask, x.dtype)
+                return inputs
+
+            def build(self, inputs_shape, attention_mask_shape=None):
+                test.assertIsNotNone(attention_mask_shape)
+                self.built = True
 
         layer = CustomLayer()
-        layer.build(
-            {
-                "x": (None, 3),
-                "attention_mask": (None, 1),
-            }
-        )
         x = np.ones((2, 3), dtype="float32")
         mask = np.ones((2, 1), dtype="float32")
         y = layer(x, attention_mask=mask)


keras/src/layers/layer_test.py
```diff
@@ -1853,3 +1853,96 @@ def call(self, x):
...
+        self.assertEqual(y1.shape, (3, 8))
+
+        # Test with padding_mask
+        padding_mask = np.array([[1.0], [1.0], [0.0]], dtype="float32")
+        y2 = layer(x, padding_mask=padding_mask)
+        self.assertEqual(y2.shape, (3, 8))
+
+        # Verify compute_output_shape works
+        output_shape = layer.compute_output_shape((3, 5))
+        self.assertEqual(output_shape, (3, 8))
```
hertschuh: The only difference between this test and the previous one is the name of the `_mask` argument. I don't think it's worth the trouble. Just remove it.

keras/src/layers/layer_test.py
```diff
@@ -1853,3 +1853,96 @@ def call(self, x):
...
+        self.assertEqual(y.shape, (2, 4))
+        self.assertLen(layer.weights, 2)  # kernel and bias from Dense
+
+        # Test 2: Forward pass with attention_mask
+        attention_mask = np.array([[1.0], [0.5]], dtype="float32")
+        y_masked = layer(x, attention_mask=attention_mask)
+        self.assertEqual(y_masked.shape, (2, 4))
+
+        # Test 3: compute_output_shape works correctly
+        output_shape = layer.compute_output_shape((2, 3))
```
hertschuh: I'm not following what these tests are trying to prove.

First, they will pass without your change to `layer.py`, so they don't show that something was fixed.

You explicitly define `compute_output_shape`, you call `compute_output_shape`, nothing can fail.

What https://github.com/keras-team/keras/issues/21154 was about, is the fact that the workaround is to implement `compute_output_shape`. But you want it to work without that workaround.

What the tests should do is:
- not define `compute_output_shape`
- try to call `build` with both `x` and `attention_mask` shapes
- separately, create a `Functional` model with this layer and see if it works

MalyalaKarthik66: Thanks for the review. Updated the tests to remove the `compute_output_shape` workaround and instead validate behavior via `build()` with both input and `attention_mask` shapes, and via Functional model usage.

keras/src/layers/layer_test.py
```diff
@@ -1853,3 +1853,41 @@ def call(self, x):
         # foo_mode omitted -> foo_mode defaults to False -> no change
         y2 = model(sample_input)
         self.assertAllClose(y2, sample_input)
+
+    def test_functional_model_with_attention_mask_arg(self):
```
hertschuh: I tried these 2 new unit tests and they both pass *without* the change you made to `layer.py`. So these do not highlight the issue.

You can replace them with this one test, which *does* require the change. Specifically, it calls `build` indirectly by invoking `__call__`.

```python
    def test_layer_build_with_attention_mask_arg(self):
        test = self

        class CustomLayer(layers.Layer):
            def call(self, inputs, attention_mask=None):
                if attention_mask is not None:
                    return inputs * ops.cast(attention_mask, x.dtype)
                return inputs

            def build(self, inputs_shape, attention_mask_shape=None):
                test.assertIsNotNone(attention_mask_shape)
                self.built = True

        layer = CustomLayer()
        x = np.ones((2, 3), dtype="float32")
        mask = np.ones((2, 1), dtype="float32")
        y = layer(x, attention_mask=mask)
        self.assertEqual(y.shape, (2, 3))
```

MalyalaKarthik66: Updated the tests as suggested.

PR #22119, opened 2026-02-06
Fix vocabulary reload corruption caused by trailing newline handling in IndexLookup.load_assets

MalyalaKarthik66:
- Fixes an issue where trailing newlines in `vocabulary.txt` could introduce unintended empty tokens during asset reload.
- This ensures vocabulary (including special tokens) is reconstructed exactly as saved.
- Adds tests covering save/load consistency and trailing newline edge cases.

### DIFF
--- a/keras/src/layers/preprocessing/index_lookup_test.py
+++ b/keras/src/layers/preprocessing/index_lookup_test.py
@@ -1,5 +1,4 @@
 import os
-import tempfile
 
 import numpy as np
 import pytest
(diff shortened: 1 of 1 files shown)


keras/src/layers/preprocessing/index_lookup_test.py
```diff
@@ -1,4 +1,5 @@
 import os
+import tempfile
```
hertschuh: Please use `self.get_temp_dir()` in the tests instead.
- you don't need a `with` so the code is not indented, which is nicer
- it will cleanup at the end of the test

keras/src/layers/preprocessing/index_lookup_test.py
```diff
@@ -616,3 +617,99 @@ def test_ensure_known_vocab_size_without_vocabulary(self):
         ):
             input_data = ["sample", "data"]
             layer(input_data)
+
+    def test_save_and_load_assets_string_vocab(self):
```
hertschuh, line 620: Just for my understanding. This test was *not* failing before the fix you added in `index_lookup.py`, right? This is just adding coverage.

keras/src/layers/preprocessing/index_lookup_test.py
```diff
@@ -616,3 +617,99 @@ def test_ensure_known_vocab_size_without_vocabulary(self):
...
+            )
+
+            self.assertEqual(vocab_before, vocab_after)
+            self.assertEqual(vocab_before_no_special, vocab_after_no_special)
+
+            # Verify outputs match
+            output_after = layer2(sample_input).numpy()
+            np.testing.assert_array_equal(output_before, output_after)
+
+    def test_save_and_load_assets_with_multiple_oov_indices(self):
```
hertschuh, line 659: Just for my understanding. This test was *not* failing before the fix you added in `index_lookup.py`, right? This is just adding coverage.

hertschuh (commented):
Thank you for the fix!

keras/src/layers/preprocessing/index_lookup_test.py
MalyalaKarthik66: Updated to use `self.get_temp_dir()`.

keras/src/layers/preprocessing/index_lookup_test.py
MalyalaKarthik66: Yes, these add coverage to ensure save/load behavior (including special tokens) remains stable and prevents regressions.

PR #22123, opened 2026-02-07
Add train/val shuffle coverage for image_dataset_from_directory

Kh9705:
Adds tests for train/val splitting with shuffle and seed determinism.
Verifies train/val file paths are disjoint and cover the full dataset.
No functional changes; tests only.
Testing:
PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest [image_dataset_utils_test.py](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

### DIFF
--- a/keras/src/utils/image_dataset_utils_test.py
+++ b/keras/src/utils/image_dataset_utils_test.py
@@ -671,6 +671,7 @@ class ImageDatasetFromDirectoryTest(testing.TestCase):
                 **common_args,
             )
         )
+
         def _count_batches_and_samples(dataset):
             batch_count = 0
             sample_count = 0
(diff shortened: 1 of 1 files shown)


keras/src/utils/image_dataset_utils_test.py
```diff
@@ -653,3 +652,44 @@ def test_image_dataset_from_directory_shuffle(self, format):
...
+        train_dataset, val_dataset = (
+            image_dataset_utils.image_dataset_from_directory(
+                directory,
+                validation_split=0.2,
+                subset="both",
+                **common_args,
+            )
+        )
+        train_paths = train_dataset.file_paths
+        val_paths = val_dataset.file_paths
```
hertschuh: This additional test fully relies on `file_paths`, which is added for reference but is typically not used (and it's not even clear if they're shuffled or not).

I'd rather have the test verifies the actual tensors returned by the dataset, like the test above. You can count the batches for verification. I would also verify that the split was applied correctly (i.e. the number of train batches vs. val batches are what you expect).

hertschuh (commented):
Thanks for adding this!

keras/src/utils/image_dataset_utils_test.py
Kh9705: Updated to check batch/sample counts and expected split sizes, removed file_paths usage.

PR #22124, opened 2026-02-07
fix: correct typo 'occurences' to 'occurrences' in function name

thecaptain789:
Fixes a typo in the file_editor module where the local function 'count_occurences' should be 'count_occurrences'.


thecaptain789:
Thanks for catching that bug! I've pushed a fix implementing your suggested refactor. The function now correctly counts occurrences in nested structures.

hertschuh:
Thanks for the fix. Could you approve the Google CLA?

Issue #22128, opened 2026-02-09
draw_bounding_boxes casts float32 [0,1] images to uint8 without scaling, causing data loss


amadhan882:
## Summary
`keras.visualization.draw_bounding_boxes` silently converts float32 images to `uint8` using direct casting instead of proper scaling.

For images in the common `[0.0, 1.0]` range, this results in catastrophic data loss:

- 1.0 → 1  
- 0.0 → 0  
- All intermediate values → truncated to 0  

This causes the entire image to become nearly black, making bounding box visualization unusable.

This is a correctness and data-integrity bug.

---

## Reproduction Script

```python
import numpy as np
from keras.src.visualization.draw_bounding_boxes import draw_bounding_boxes

# A pure white float32 image in [0, 1] range
images = np.ones((1, 100, 100, 3), dtype="float32")
boxes = {
    "boxes": np.array([[[10, 10, 50, 50]]]),
    "labels": np.array([[0]])
}

# Apply bounding box drawing
out = draw_bounding_boxes(images, boxes, "xyxy")

# Inspect background value
print("Background pixel value:", out[0, 0, 0, 0])
# Expected: 255 (white)
# Actual:   1 (almost black)
```

### Actual Behavior
```
Background pixel value: 1
```

### Expected Behavior
```
Background pixel value: 255
```

---

## Root Cause
Inside `draw_bounding_boxes`, the image is cast to uint8:

```python
images = ops.convert_to_numpy(images).astype("uint8")
```

No scaling is applied, so float32 values in `[0,1]` become almost entirely zeros.

---

### Minimal fix
```python
if ops.dtype(images) == "float32":
    images = images * 255.0
images = ops.convert_to_numpy(images).astype("uint8")
```

### Recommended
- Scale only if float values are in `[0, 1]`
- Preserve images already in `[0, 255]` float format
- Consider warning users when ranges are ambiguous

---

## Environment
- **OS:** Windows 11  
- **Python:** 3.13.5  
- **Keras:** 3.14.0  

---

## Conclusion
`draw_bounding_boxes` currently destroys pixel information for float32 inputs.  
Handling the conversion properly restores expected visualization behavior.

amadhan882:
Hi @hertschuh, @Abhinavexists, and @maitry63,

Thank you all for the quick fix and for addressing the float32 handling issue in draw_bounding_boxes. 
Appreciate the fast turnaround and the improvement to preserve image data correctly.

Glad to see it resolved.

PR #22129, opened 2026-02-09
Fix: draw_bounding_boxes float32 to uint8 conversion

Abhinavexists:
fixes: #22128 

Scale float32 images in [0, 1] range by 255 before casting to uint8.
Previously, direct casting truncated all values to near-zero, making
the output image almost entirely black.

Now the conversion checks if the image dtype is floating-point:
- If max <= 1.0: scale by 255, then cast to uint8
- If max > 1.0: cast directly to uint8
- Non-float dtypes: cast directly to uint8


hertschuh (approved):
Thanks for the fix!

Issue #22127, opened 2026-02-08
Missing validation for output_padding < strides in Conv2DTranspose leading to backend abort


amadhan882:
## Summary
`keras.layers.Conv2DTranspose` does not validate the relationship between `output_padding` and `strides`.  
Passing an illegal configuration (e.g., `output_padding >= strides`) results in a **backend-level AbortedError** from TensorFlow instead of a Keras `ValueError`.

This makes the operator backend-dependent and inconsistent with expected Keras API behavior.

---

## Reproduction Script

```python
import os
os.environ["KERAS_BACKEND"] = "tensorflow"
import keras
import numpy as np

print("\n[TEST] Testing ConvTranspose shape validation...")

try:
    layer = keras.layers.Conv2DTranspose(
        filters=3,
        kernel_size=3,
        strides=2,
        padding="valid",
        output_padding=3,   # Illegal: must be < strides
    )

    x = np.random.rand(1, 10, 10, 3).astype("float32")
    out = layer(x)
    print("BUG: Layer allowed invalid output_padding. Shape:", out.shape)

except Exception as e:
    print("Caught:", type(e).__name__)
```

### Actual Behavior (TensorFlow)
```
AbortedError: could not create primitive descriptor …
```

### Expected Behavior (Keras frontend)
```
ValueError: output_padding must be < strides
```

---

## Root Cause

`Conv2DTranspose` does not check:

```python
if output_padding >= strides:
    raise ValueError(...)
```

It relies entirely on backend shape inference, which is undefined across TF / JAX / Torch backends.

---

## Suggested Fix

Add simple validation in the constructor:

```python
if output_padding is not None:
    if output_padding >= strides:
        raise ValueError(
            f"`output_padding` must be < `strides`. "
            f"Received: output_padding={output_padding}, strides={strides}"
        )
```

---

## Environment
- OS: Windows 11
- Python: 3.13.5
- Keras: 3.14.0
- Backend: TensorFlow

---

## Conclusion

This is a correctness issue:  
Keras should validate `output_padding` before dispatching the op to backend frameworks.

amadhan882:
Hi @hertschuh, @Abhinavexists, @sachinprasadhs, and @maitry63,

Thank you all for reviewing and fixing this so quickly. 
Appreciate the update to the validation checks and the fast turnaround.

Happy to contribute further if needed.

PR #22130, opened 2026-02-09
Fix: add missing validation for output padding < strides

Abhinavexists:
fixes: #22127 

Add validation for output_padding < strides in ConvTranspose layers

Conv2DTranspose (and other ConvTranspose variants) did not validate the relationship between `output_padding` and `strides`. Passing `output_padding >= strides` resulted in backend-specific errors

Added a per-dimension check in BaseConvTranspose.__init__ to raise a ValueError when any output_padding value is not strictly less than the corresponding strides value.


hertschuh (approved):
Thanks for the fix!

PR #22135, opened 2026-02-09
Add reduction parameter to scatter_update for accumulating updates

### LINKED ISSUE #20061: Support advanced ND scatter/reduce operations
## Proposal
Support operations that scatter updates into a tensor with the following requirements:
- Support a reduction with overlapping values (e.g. min, max, sum)
- Supports default value `tensor` to also consider in the reduction
- Supports ND updates & indices

Essentially, support tensorflow's `tensor_scatter_nd_*` operations, like: 
- [`tf.tensor_scatter_nd_update`](https://www.tensorflow.org/api_docs/python/tf/tensor_scatter_nd_update)
- [`tf.tensor_scatter_nd_max`](https://www.tensorflow.org/api_docs/python/tf/tensor_scatter_nd_max)
- [`tf.tensor_scatter_nd_min`](https://www.tensorflow.org/api_docs/python/tf/tensor_scatter_nd_min)

### API design
Currently, there is already a [`keras.ops.scatter_update`](https://github.com/keras-team/keras/blob/v3.3.3/keras/src/ops/core.py#L75) function. This only supports a default value of 0 for the `tensor`, and doesn't support reductions.

**Option 1:** Upgrade `keras.ops.scatter_update` to take in optional `tensor` and `reduction` arguments. 

**Option 2:** Create a dedicated `tensor_scatter_nd` api for these new functions.

### Technical implementation
Overall the technical implementation is straightfoward, just flatten indices to the 1D case and invoke the segmentation functions of each backend (e.g. `segment_max` for TF and `scatter_reduce` for torch).

## Implementation plan

I have already implemented a `tensor_scatter_nd` function for Nuro's internal use and it works well for all our cases (and is XLA compatible for all backends). Once we settle on a design, I can upstream the implementation.

timon0305:
Extend scatter_update to support reduction operations when multiple updates target the same index. This enables use cases like:
- Accumulating gradients at sparse indices
- Computing max/min pooling with sparse updates
- Multiplicative updates for attention mechanisms

Supported reductions:
- None (default): Updates replace existing values
- "add": Updates are added to existing values
- "max": Maximum of updates and existing values
- "min": Minimum of updates and existing values
- "mul": Updates are multiplied with existing values

Implementation for all backends:
- TensorFlow: tensor_scatter_nd_add/max/min
- JAX: jnp.ndarray.at[].add/max/min/multiply
- PyTorch: index_put_ with accumulate, scatter_reduce_
- NumPy: np.add.at, np.maximum.at, etc.

Fixes #20061

### DIFF
--- a/keras/src/ops/core_test.py
+++ b/keras/src/ops/core_test.py
@@ -1128,34 +1128,6 @@ class CoreOpsCorrectnessTest(testing.TestCase):
         result = core.ScatterUpdate(reduction="add")(inputs, indices, updates)
         self.assertAllClose(result, [3.0, 3.0, 0.0, 0.0])
 
-        # Test add reduction with slice updates and duplicate indices
-        inputs = np.zeros((2, 2))
-        indices = [[0], [0]]
-        updates = np.array([[1.0, 2.0], [3.0, 4.0]])
-        result = core.scatter_update(inputs, indices, updates, reduction="add")
-        self.assertAllClose(result, [[4.0, 6.0], [0.0, 0.0]])
-
-        # Test max reduction with slice updates and duplicate indices
-        inputs = np.zeros((2, 2))
-        indices = [[0], [0]]
-        updates = np.array([[1.0, 5.0], [3.0, 2.0]])
-        result = core.scatter_update(inputs, indices, updates, reduction="max")
-        self.assertAllClose(result, [[3.0, 5.0], [0.0, 0.0]])
-
-        # Test min reduction with slice updates and duplicate indices
-        inputs = np.full((2, 2), 10.0)
-        indices = [[0], [0]]
-        updates = np.array([[5.0, 8.0], [3.0, 9.0]])
-        result = core.scatter_update(inputs, indices, updates, reduction="min")
-        self.assertAllClose(result, [[3.0, 8.0], [10.0, 10.0]])
-
-        # Test mul reduction with slice updates and duplicate indices
-        inputs = np.ones((2, 2))
-        indices = [[0], [0]]
-        updates = np.array([[2.0, 3.0], [4.0, 5.0]])
-        result = core.scatter_update(inputs, indices, updates, reduction="mul")
-        self.assertAllClose(result, [[8.0, 15.0], [1.0, 1.0]])
-
     def test_shape(self):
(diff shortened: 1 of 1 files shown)


keras/src/ops/core_test.py
```diff
@@ -1084,6 +1084,78 @@ def test_scatter_update(self):
...
+        inputs = np.array([10.0, 10.0, 10.0, 10.0])
+        indices = [[0], [0], [1]]
+        updates = np.array([3.0, 5.0, 2.0])
+        result = core.scatter_update(inputs, indices, updates, reduction="min")
+        self.assertAllClose(result, [3.0, 2.0, 10.0, 10.0])
+
+        # Test mul reduction
+        inputs = np.array([2.0, 3.0, 1.0, 1.0])
+        indices = [[0], [1]]
+        updates = np.array([3.0, 4.0])
```
hertschuh: Can you do a similar example for "mul" as for the previous ones:
```python
        indices = [[0], [0], [1]]
        updates = np.array([3.0, 5.0, 2.0])
```
So that it exercise the case when the same value is multipled multiple times.

keras/src/ops/core_test.py
```diff
@@ -1084,6 +1084,78 @@ def test_scatter_update(self):
...
+        updates = np.array([[5.0, 8.0], [3.0, 9.0]])
+        result = core.scatter_update(inputs, indices, updates, reduction="min")
+        self.assertAllClose(result, [[3.0, 8.0], [10.0, 10.0]])
+
+        # Test mul reduction with slice updates and duplicate indices
+        inputs = np.ones((2, 2))
+        indices = [[0], [0]]
+        updates = np.array([[2.0, 3.0], [4.0, 5.0]])
+        result = core.scatter_update(inputs, indices, updates, reduction="mul")
+        self.assertAllClose(result, [[8.0, 15.0], [1.0, 1.0]])
```
hertschuh: As far as I can tell, all of these are already covered above. So these can be removed.

hertschuh (commented):
Thank you for adding this feature!

hertschuh:
Any reason why this PR is still a draft? Or can I submit it?

PR #22136, opened 2026-02-09
implement vsplit operation for all backends

Shi-pra-19:
Added new `vsplit` ops, which split a tensor into multiple sub-tensors vertically. 
Supported across all backends (NumPy, TensorFlow, Torch, JAX, OpenVINO).

### DIFF
--- a/keras/src/backend/jax/numpy.py
+++ b/keras/src/backend/jax/numpy.py
@@ -1355,6 +1355,7 @@ def vstack(xs):
 
 
 def vsplit(x, indices_or_sections):
+    x = convert_to_tensor(x)
     return jnp.vsplit(x, indices_or_sections)
 
 
--- a/keras/src/backend/jax/numpy.py
+++ b/keras/src/backend/jax/numpy.py
@@ -1354,6 +1354,10 @@ def vstack(xs):
     return jnp.vstack(xs)
 
 
+def vsplit(x, indices_or_sections):
+    return jnp.vsplit(x, indices_or_sections)
+
+
 def vectorize(pyfunc, *, excluded=None, signature=None):
     if excluded is None:
         excluded = set()
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -3030,6 +3030,10 @@ def vstack(xs):
     return tf.concat(xs, axis=0)
 
 
+def vsplit(x, indices_or_sections):
+    return tf.experimental.numpy.vsplit(x, indices_or_sections)
+
+
 def _vmap_fn(fn, in_axes=0):
     if in_axes != 0:
         raise ValueError(
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -3031,7 +3031,7 @@ def vstack(xs):
 
 
 def vsplit(x, indices_or_sections):
-    return tf.experimental.numpy.vsplit(x, indices_or_sections)
+    return split(x, indices_or_sections, axis=0)
 
 
 def _vmap_fn(fn, in_axes=0):
--- a/keras/src/ops/numpy.py
+++ b/keras/src/ops/numpy.py
@@ -7268,6 +7268,98 @@ def vstack(xs):
     return backend.numpy.vstack(xs)
 
 
+class Vsplit(Operation):
+    def __init__(self, indices_or_sections, *, name=None):
+        super().__init__(name=name)
+        if not isinstance(indices_or_sections, int):
+            indices_or_sections = tuple(indices_or_sections)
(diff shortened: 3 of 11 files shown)


keras/src/ops/numpy.py
```diff
@@ -7194,6 +7194,98 @@ def vstack(xs):
...
+            end = all_indices[i + 1]
+            if start is None or end is None:
+                output_size = None
+            else:
+                output_size = end - start
+            output_shape = list(x_shape)
+            output_shape[0] = output_size
+            outputs.append(KerasTensor(output_shape, dtype=x.dtype))
+
+        return outputs
```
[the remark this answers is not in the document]
Shi-pra-19: Done

keras/src/backend/tensorflow/numpy.py
```diff
@@ -3009,6 +3009,10 @@ def vstack(xs):
     return tf.concat(xs, axis=0)
 
 
+def vsplit(x, indices_or_sections):
+    return tf.experimental.numpy.vsplit(x, indices_or_sections)
```
hertschuh: Actually, I would prefer you do `return split(x, indices_or_sections, axis=0)`.

Early on, we tried using `tf.experimental.numpy` but ran into issues.

keras/src/ops/numpy_test.py
```diff
@@ -2608,6 +2616,14 @@ def test_split(self):
         with self.assertRaises(ValueError):
             knp.split(x, 2, axis=1)
 
+    def test_vsplit(self):
+        x = KerasTensor((2, 3))
+        self.assertEqual(knp.vsplit(x, 2)[0].shape, (1, 3))
```
hertschuh: Can you also do `self.assertEqual(knp.vsplit(x, 2)[1].shape, (1, 3))`

keras/src/ops/numpy_test.py
```diff
@@ -2608,6 +2616,14 @@ def test_split(self):
...
+        self.assertEqual(len(knp.vsplit(x, [1, 2])), 3)
+        self.assertEqual(knp.vsplit(x, [1, 2])[0].shape, (1, 3))
+        self.assertEqual(knp.vsplit(x, [1, 2])[1].shape, (1, 3))
+        self.assertEqual(knp.vsplit(x, [1, 2])[2].shape, (0, 3))
```
hertschuh: Tensors of 0 size (i.e. where any dimension is 0) are generally poorly supported. Can you change the example to use (5, 3) as the input shape for instance?

keras/src/ops/numpy_test.py
```diff
@@ -5291,6 +5307,34 @@ def fn(x, indices, axis):
...
+
+        # test indices_or_sections as tensor
+        x = knp.array([[1, 2, 3], [3, 2, 1]])
+        indices_or_sections = knp.array([1])
+        x_np = np.array([[1, 2, 3], [3, 2, 1]])
+        indices_or_sections_np = np.array([1])
+        self.assertAllClose(
+            knp.vsplit(x, indices_or_sections),
+            np.vsplit(x_np, indices_or_sections_np),
+        )
```
hertschuh: For the indices split examples, can you use a slightly more complicated examples with several indices instead of just one?

keras/src/backend/jax/numpy.py
```diff
@@ -1350,6 +1350,10 @@ def vstack(xs):
     return jnp.vstack(xs)
 
 
+def vsplit(x, indices_or_sections):
+    return jnp.vsplit(x, indices_or_sections)
```
hertschuh, line 1359: Add `x = convert_to_tensor(x)`.

hertschuh (commented):
Thanks for the adding this!

hertschuh:
You'll need to rebase too.

keras/src/backend/tensorflow/numpy.py
Shi-pra-19: Updated 👍🏻

keras/src/backend/jax/numpy.py
Shi-pra-19: Updated 👍🏻

Shi-pra-19:
I’ve refactored and improved the tests as suggested and rebased on the latest master. Looking forward to your feedback!

keras/src/ops/numpy_test.py
Shi-pra-19: Updated

keras/src/ops/numpy_test.py
Shi-pra-19: Updated

keras/src/ops/numpy_test.py
Shi-pra-19: Updated

hertschuh (approved):
Thanks!

PR #22140, opened 2026-02-09
fix: support axis parameter in sparse_categorical_crossentropy

### LINKED ISSUE #21097: `sparse_categorical_crossentropy` ignores feature channel (and axis parameter) when checking tensor dimensions [torch backend])

timon0305:
## Summary
Fixes #21097 - `sparse_categorical_crossentropy` now properly supports the `axis` parameter across all backends.

Previously, the function ignored the `axis` parameter when checking tensor dimensions, causing errors when using `channels_first` data format (axis=1) for segmentation tasks.

## Changes
- Fixed `compute_output_spec` in `keras/src/ops/nn.py` to use axis parameter
- Added full axis support in TensorFlow backend (was previously axis=-1 only)
- Fixed shape checking in JAX, PyTorch, and NumPy backends to use axis parameter
- Fixed squeeze and shape handling in `keras/src/losses/losses.py`
- Updated tests to verify correct behavior across all backends

## Test plan
- [x] Tested with PyTorch, TensorFlow, and JAX backends
- [x] Verified axis=1 works for channels_first segmentation use case
- [x] All existing SparseCategoricalCrossentropy tests pass
- [x] Pre-commit checks pass

### DIFF
--- a/keras/src/backend/jax/nn.py
+++ b/keras/src/backend/jax/nn.py
@@ -977,8 +977,6 @@ def categorical_crossentropy(target, output, from_logits=False, axis=-1):
 def sparse_categorical_crossentropy(target, output, from_logits=False, axis=-1):
     target = jnp.array(target, dtype="int32")
     output = jnp.array(output)
-    if len(target.shape) == len(output.shape) and target.shape[-1] == 1:
-        target = jnp.squeeze(target, axis=-1)
 
     if len(output.shape) < 1:
         raise ValueError(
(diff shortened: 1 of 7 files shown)


keerthanakadiri:
Hi @timon0305, Can you please try to resolve the conflicts? Thank you!

PR #22142, opened 2026-02-10
Fix TrackedDict constructor to support iterable (key, value) inputs

MalyalaKarthik66:
- Fixes an issue where `TrackedDict.__init__` assumed mapping inputs and failed when given iterable (key, value) pairs (e.g. `zip(keys, values)`), raising `AttributeError: 'zip' object has no attribute 'items'`.

- The constructor now accepts both mappings and iterable pairs, aligning with standard Python dict constructor behavior and preventing failures during container reconstruction (e.g. tree unflatten flows).

- Adds regression tests covering OrderedDict input and iterable pair input.

### DIFF
--- a/keras/src/utils/tracking.py
+++ b/keras/src/utils/tracking.py
@@ -226,7 +226,7 @@ class TrackedDict(dict):
             if hasattr(values, "items"):
                 items_iter = values.items()
             else:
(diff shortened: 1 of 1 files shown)


hertschuh (approved):
Thanks for the fix!

PR #22145, opened 2026-02-10
Add TrackedOrderedDict support in Tracker.track()

MalyalaKarthik66:
- Adds support for OrderedDict inputs in Tracker.track() by introducing TrackedOrderedDict, preserving ordering semantics while maintaining tracking behavior.

- Ensures mutation operations (setitem, update, pop, etc.) correctly track and untrack values, consistent with existing tracked collection implementations.

- Keeps the change localized to tracking.py, aligning with the existing TODO comment (OrderedDict?) without modifying backend behavior.

- Adds minimal unit tests to verify type preservation, order preservation, and basic tracking behavior for TrackedOrderedDict.

### DIFF
--- a/keras/src/utils/tracking.py
+++ b/keras/src/utils/tracking.py
@@ -242,14 +242,23 @@ class TrackedDict(dict):
             mapping = {k: self.tracker.track(v) for k, v in mapping.items()}
         super().update(mapping)
 
-    def pop(self, key, default=None):
-        if self.tracker:
-            value = super().pop(key, default)
-            if value is not default:
-                self.tracker.untrack(value)
+    def pop(self, key, *args):
+        if len(args) > 1:
+            raise TypeError(
+                f"pop expected at most 2 arguments, got {1 + len(args)}"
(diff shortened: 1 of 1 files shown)


keras/src/utils/tracking.py
```diff
@@ -289,6 +291,85 @@ def torchtree_flatten_with_keys(self):
...
+        if not self.tracker:
+            return super().pop(key, *args)
+        try:
+            value = super().pop(key)
+            self.tracker.untrack(value)
+            return value
+        except KeyError:
+            if args:
+                return args[0]
+            raise
```
hertschuh, line 348: This is actually a more correct approach than what was done for `TrackedDict`. Can you fix `TrackedDict` too?

MalyalaKarthik66: Updated `TrackedDict.pop` to match the corrected logic and rebased onto latest master.

hertschuh (approved):
Thanks!

PR #22146, opened 2026-02-10
OpenVINO backend: add bitwise_left_shift and bitwise_right_shift

Rahuldrabit:
Implements bitwise_left_shift and bitwise_right_shift for the OpenVINO backend. 
Tests passed for numpy bitshift tests. 
Related OpenVINO issue: https://github.com/openvinotoolkit/openvino/issues/34036. CC @rkazants

### DIFF
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -642,7 +642,7 @@ def bitwise_left_shift(x, y):
     x = get_ov_output(x, element_type)
     y = get_ov_output(y, element_type)
     x, y = _align_operand_types(x, y, "bitwise_left_shift()")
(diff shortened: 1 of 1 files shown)


goyaladitya05:
Please remove pytest.ini from the PR. That file is for local testing adjustments and shouldn't be committed to the repo.

Rahuldrabit:
Now I remove the pytest,ini from pr , thank you .

keras/src/backend/openvino/numpy.py
```diff
@@ -1,5 +1,6 @@
 import numpy as np
 import openvino.opset14 as ov_opset
+import openvino.opset15 as ov_opset15
```
hertschuh: Please rebase, this was upgraded to opset15.

PR #22155, opened 2026-02-11
Implement numpy.gcd using Euclidean algorithm for OpenVINO backend

andersendsa:
**Details:**
Implemented numpy.gcd in keras/src/backend/openvino/numpy.py using OpenVINO's loop operation (Euclidean algorithm).
Enabled corresponding tests in keras/src/backend/openvino/excluded_concrete_tests.txt.
Verified that the implementation works correctly with various dtypes and element-wise operations.
closes issue : https://github.com/openvinotoolkit/openvino/issues/34056
ticket : #34017


andersendsa:
@rkazants  , @hertschuh  This PR is ready for review, please have a look when you have the time.

PR #22160, opened 2026-02-11
Functional: validate None inputs against InputLayer(optional=True) and improve error handling

MalyalaKarthik66:
- Implements the TODO in `Functional._convert_inputs_to_tensors` to validate `None` inputs using the `InputLayer(optional=...)` flag.

- Previously, `None` values were passed through without validation and could lead to downstream errors (e.g., AttributeError during structure mismatch handling). This change ensures `None` is only allowed for optional inputs, while required inputs raise a `ValueError`. Existing behavior is preserved for tensors not originating from `InputLayer`.

- Also improves structure mismatch message safety for `None` and ensures user-facing `ValueError` is not masked during symbolic build.

- Added tests covering required inputs receiving `None`, mixed optional/required structures, and the `fit()` path. Verified across all backends.

### DIFF
--- a/keras/src/models/functional.py
+++ b/keras/src/models/functional.py
@@ -214,6 +214,21 @@ class Functional(Function, Model):
         return output_shapes
 
     def _assert_input_compatibility(self, *args):
+        flat_inputs = tree.flatten(args[0])
+        for x, input_tensor in zip(flat_inputs, self._inputs):
+            if x is None:
+                input_layer = input_tensor._keras_history.operation
+                if (
+                    isinstance(input_layer, InputLayer)
+                    and not input_layer.optional
+                ):
+                    raise ValueError(
+                        (
+                            f"The input '{input_tensor.name}' is not optional, "
+                            "but None was passed. "
+                            "Please provide a valid tensor."
+                        )
+                    )
         return super(Model, self)._assert_input_compatibility(*args)
 
     def _maybe_warn_inputs_struct_mismatch(self, inputs, raise_exception=False):
@@ -225,26 +240,14 @@ class Functional(Function, Model):
                 tree.lists_to_tuples(self._inputs_struct),
             )
         except:
-
-            def _safe_name(x):
-                if x is None:
-                    return "None"
-                if isinstance(x, backend.KerasTensor):
-                    return x.name
-                return repr(x)
-
-            def _safe_inputs_repr(x):
-                if x is None:
-                    return "None"
-                if isinstance(x, backend.KerasTensor):
-                    return f"Tensor(shape={x.shape})"
-                return repr(x)
-
             model_inputs_struct = tree.map_structure(
-                _safe_name,
+                lambda x: "None" if x is None else x.name,
                 self._inputs_struct,
             )
-            inputs_struct = tree.map_structure(_safe_inputs_repr, inputs)
+            inputs_struct = tree.map_structure(
+                lambda x: "None" if x is None else f"Tensor(shape={x.shape})",
+                inputs,
+            )
             msg = (
                 "The structure of `inputs` doesn't match the expected "
                 f"structure.\nExpected: {model_inputs_struct}\n"
@@ -270,9 +273,8 @@ class Functional(Function, Model):
                     input_layer = input_tensor._keras_history.operation
 
                 if isinstance(input_layer, InputLayer):
-                    is_optional = getattr(input_layer, "optional", False)
-                    if not is_optional:
-                        input_name = getattr(input_tensor, "name", "unknown")
+                    if not input_layer.optional:
+                        input_name = input_tensor.name
                         raise ValueError(
                             (
                                 f"The input '{input_name}' is not optional, "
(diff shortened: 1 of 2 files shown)


keras/src/models/functional_test.py
```diff
@@ -687,6 +687,62 @@ def test_disconnected_output(self):
...
+            def compute_output_shape(self, x_shape):
+                return x_shape
+
+        i1 = Input((2,), name="input1")
+        i2 = Input((2,), name="input2", optional=False)
+        outputs = OptionalInputLayer()(i1, i2)
+        model = Model({"input1": i1, "input2": i2}, outputs)
+
+        # Passing None for a required input must raise
+        with pytest.raises(ValueError):
```
hertschuh: Please change to `self.assertRaisesRegex` with a short part of the message. The truth is that without your change, it would already raise, but with a message that was not very clear.

keras/src/models/functional_test.py
```diff
@@ -687,6 +687,62 @@ def test_disconnected_output(self):
...
+        # Small trainable model where input is required. fit should error
+        x = layers.Input((3,), name="x")
+        y = layers.Dense(1)(x)
+        model = Model(x, y)
+        model.compile(optimizer="sgd", loss="mse")
+
+        y_data = np.ones((4, 1))
+
+        # Passing None for required input should raise during fit
+        with pytest.raises(ValueError):
```
hertschuh: Use `self.assertRaisesRegex` with a short part of the message.

keras/src/models/functional.py
```diff
@@ -225,12 +225,26 @@ def _maybe_warn_inputs_struct_mismatch(self, inputs, raise_exception=False):
                 tree.lists_to_tuples(self._inputs_struct),
             )
         except:
+
+            def _safe_name(x):
+                if x is None:
+                    return "None"
+                if isinstance(x, backend.KerasTensor):
+                    return x.name
+                return repr(x)
```
hertschuh: I don't think you need 3 cases here, it's either `None` or has a name.

So you can do:
```python
            model_inputs_struct = tree.map_structure(
                lambda x: "None" if x is None else x.name, self._inputs_struct
            )
```

keras/src/models/functional.py
```diff
@@ -225,12 +225,26 @@ def _maybe_warn_inputs_struct_mismatch(self, inputs, raise_exception=False):
...
+
+            def _safe_inputs_repr(x):
+                if x is None:
+                    return "None"
+                if isinstance(x, backend.KerasTensor):
+                    return f"Tensor(shape={x.shape})"
+                return repr(x)
```
hertschuh: Same here, this should work:

```python
            inputs_struct = tree.map_structure(
                lambda x: "None" if x is None else f"Tensor(shape={x.shape})", inputs
            )
```

keras/src/models/functional.py
```diff
@@ -242,14 +256,39 @@ def _maybe_warn_inputs_struct_mismatch(self, inputs, raise_exception=False):
...
+                input_layer = None
+                if (
+                    hasattr(input_tensor, "_keras_history")
+                    and input_tensor._keras_history
+                ):
+                    input_layer = input_tensor._keras_history.operation
+
+                if isinstance(input_layer, InputLayer):
+                    is_optional = getattr(input_layer, "optional", False)
+                    if not is_optional:
```
hertschuh: The attribute will be there, you don't need `getattr`:
```python
if not input_layer.optional:
```

keras/src/models/functional.py
```diff
@@ -242,14 +256,39 @@ def _maybe_warn_inputs_struct_mismatch(self, inputs, raise_exception=False):
...
+                        input_name = getattr(input_tensor, "name", "unknown")
+                        raise ValueError(
+                            (
+                                f"The input '{input_name}' is not optional, "
+                                "but None was passed. "
+                                "Please provide a valid tensor."
+                            )
+                        )
```
hertschuh: Same thing, the `name` attribute will be there, no need for `getattr`.

keras/src/models/functional.py
```diff
@@ -242,14 +256,39 @@ def _maybe_warn_inputs_struct_mismatch(self, inputs, raise_exception=False):
...
                 converted.append(x)
             else:
+                # Preserve dtype and sparse settings from input_tensor
+                dtype = getattr(input_tensor, "dtype", None)
+                sparse = getattr(input_tensor, "sparse", False)
                 converted.append(
-                    ops.convert_to_tensor(
-                        x, dtype=input.dtype, sparse=input.sparse
-                    )
+                    ops.convert_to_tensor(x, dtype=dtype, sparse=sparse)
```
hertschuh: I don't think these changes are needed. `dtype` and `sparse` are always there.

hertschuh (commented):
Thank you for fixing this!

MalyalaKarthik66:
@hertschuh Thanks for the feedback — I've addressed the review comments and pushed the updates.

PR #22164, opened 2026-02-13
Refactors LayoutMap to use strict regex matching

JyotinderSingh:
This PR updates `keras.distribution.LayoutMap` to use strict regular expression matching via `re.fullmatch` instead of standard pattern searching via `re.search`. This change aligns the behavior of `LayoutMap` with `keras.dtype_policies.DTypePolicyMap` (fixed in https://github.com/keras-team/keras/pull/21608), establishing consistency across the two map APIs that support both exact and fallback prefix/regex string matching. 

Previously, a key like `"conv2d.*kernel"` in `LayoutMap` would implicitly match `"my_model/conv2d_123/kernel"`. Now, users must be explicit and specify `".*conv2d.*kernel"` in order to fully match the tensor's path.


hertschuh (approved):
Thanks for the fix!

PR #22170, opened 2026-02-13
Add backend-agnostic type aliases module (keras.types)

### LINKED ISSUE #19230: Backend-Agnostic Types
It would be nice if `keras` could provide backend-agnostic types that we can use to type-hint libraries built on top of `keras`. Some that come to mind are:

- Tensor
- Shape
- Distribution

The following snippet works for the `Tensor` type, even from outside `keras`:

```py3
# this is ugly, but:
# 1. it is recognized by static type checkers (not possible with if-else branching)
# 2. it does not leave the Tensor type possibly undefined (not possible without nesting)
try:
    import jax
    Tensor = jax.Array
except ModuleNotFoundError:
    try:
        import tensorflow as tf
        Tensor = tf.Tensor
    except ModuleNotFoundError:
        import torch
        Tensor = torch.Tensor
```

timon0305:
## Summary

Adds a `keras.types` module with backend-agnostic type aliases for use in type annotations:

- **`Tensor`** — Under `TYPE_CHECKING`, resolves to `Union[KerasTensor, np.ndarray, jax.Array, tf.Tensor, torch.Tensor]`
- **`Shape`** — Under `TYPE_CHECKING`, resolves to `Tuple[Optional[int], ...]`
- **`DType`** — Under `TYPE_CHECKING`, resolves to `str`

These types are **annotation-only** — they raise `TypeError` on instantiation and cannot be subclassed. At runtime, Keras operations accept native backend tensors, plain tuples for shapes, and plain strings for dtypes. The value of this module is providing proper type narrowing for static type checkers (mypy, PyCharm) without requiring any specific backend to be installed.

Closes #19230.

### DIFF
--- a/keras/src/types.py
+++ b/keras/src/types.py
@@ -95,9 +95,7 @@ else:
             )
 
         def __init_subclass__(cls, **kwargs):
-            raise TypeError(
-                "keras.types.Tensor cannot be subclassed."
-            )
+            raise TypeError("keras.types.Tensor cannot be subclassed.")
 
     @keras_export("keras.types.Shape")
     class Shape:
@@ -127,9 +125,7 @@ else:
             )
 
         def __init_subclass__(cls, **kwargs):
-            raise TypeError(
-                "keras.types.Shape cannot be subclassed."
-            )
+            raise TypeError("keras.types.Shape cannot be subclassed.")
 
     @keras_export("keras.types.DType")
     class DType:
@@ -160,6 +156,4 @@ else:
             )
 
         def __init_subclass__(cls, **kwargs):
-            raise TypeError(
-                "keras.types.DType cannot be subclassed."
-            )
+            raise TypeError("keras.types.DType cannot be subclassed.")


keras/src/types.py
```diff
@@ -0,0 +1,131 @@
...
+    Tensor = Union[KerasTensor, np.ndarray, _JaxArray, _TfTensor, _TorchTensor]
+
+    Shape = Tuple[Optional[int], ...]
+
+    DType = str
+
+else:
+
+    @keras_export("keras.types.Tensor")
+    class Tensor:
```
hertschuh, line 64: But this can't work though, we never accept this type in any of our operations or layers.

keras/src/types.py
```diff
@@ -0,0 +1,131 @@
...
+
+        def my_function(x: Tensor) -> Tensor:
+            return x
+        ```
+        """
+
+        pass
+
+    @keras_export("keras.types.Shape")
+    class Shape:
```
hertschuh, line 101: But this can't work though, we only accept and return tuples.

keras/src/types.py
```diff
@@ -0,0 +1,131 @@
...
+
+        def reshape(x, shape: Shape):
+            return keras.ops.reshape(x, shape)
+        ```
+        """
+
+        pass
+
+    @keras_export("keras.types.DType")
+    class DType:
```
hertschuh, line 131: But this can't work, we only use strings for dtypes.

keras/src/types.py
timon0305: Good point. I've updated all three types to be **annotation-only** — they now raise `TypeError` on instantiation and cannot be subclassed. The docstrings explicitly state they exist only for type hints and that Keras operations accept native backend tensors directly. The `TYPE_CHECKING` branch still resolves to the proper `Union` for static type checkers.

keras/src/types.py
timon0305: Agreed — updated. `Shape` now raises `TypeError` on instantiation with a message directing users to use plain tuples instead.

keras/src/types.py
timon0305: Agreed — updated. `DType` now raises `TypeError` on instantiation with a message directing users to use plain strings instead.

Issue #21078, opened 2025-03-21
How Do I Track a List of Objects for Export?


rlcauvin:
I define an `EnsembleModel` class that is constructed from a list of other Keras models.
```
class EnsembleModel(keras.Model):

  def __init__(
    self,
    models: Iterable[keras.Model],
    reduce_fn: Callable = keras.ops.mean,
    **kwargs):
    
    super(EnsembleModel, self).__init__(**kwargs)
    
    self.models = models
    # self.model0 = models[0]
    # self.model1 = models[1]
    self.reduce_fn = reduce_fn

  @tf.function(input_signature=[input_signature])
  def call(
    self,
    input: Dict[Text, Any]) -> Any:
        
    all_outputs = [keras.ops.reshape(model(input), newshape=(-1,)) for model in self.models]
    output = self.reduce_fn(all_outputs, axis=0)
    
    return output

averaging_model = EnsembleModel(models=[model0, model1])
```
I then wish to export the ensemble model:
```
averaging_model.export("export/1/", input_signature=[input_signature])
```
But I get an error on the export:
```
AssertionError: Tried to export a function which references an 'untracked' resource. TensorFlow objects (e.g. 
tf.Variable) captured by functions must be 'tracked' by assigning them to an attribute of a tracked object or 
assigned to an attribute of the main object directly. See the information below:
        Function name = b'__inference_signature_wrapper___call___10899653'
        Captured Tensor = <ResourceHandle(name="10671455", device="/job:localhost/replica:0/task:0/device:CPU:0", 
container="localhost", type="tensorflow::lookup::LookupInterface", dtype and shapes : "[  ]")>
        Trackable referencing this tensor = <tensorflow.python.ops.lookup_ops.StaticHashTable object at 
0x7fd62d126990>
        Internal Tensor = Tensor("10899255:0", shape=(), dtype=resource)
```
If I explicitly assign the models to variables in the constructor:
```
    self.model0 = models[0]
    self.model1 = models[1]
```
It works fine (even if I don't reference those variables anywhere else). But I want an instance of the `EnsembleModel` class to support an arbitrary list of models. How can I ensure the models are "tracked" so that I don't get an error on export?

rlcauvin:
Adding this kludge to the `EnsembleModel` constructor seems to have worked to "track" the models in the list and avoid export errors:
```
    # Register each model so that it is "tracked" for export.
    for i, model in enumerate(self.models):
      self.__setattr__(f"model_{i}", model)
```
[...]

PR #22172, opened 2026-02-14
Fix support of tracked collection in saved model export.

hertschuh:
The `TFLayer` implements a `_trackable_children` which is supposed to add support for tracked collections. However, this mechanism was bypassed because `tf.train.TrackableView` reports a `save_type` of "checkpoint" and a `save_type` of "savedmodel" was expected.

https://github.com/keras-team/keras/blob/master/keras/src/backend/tensorflow/layer.py#L48

Fixes https://github.com/keras-team/keras/issues/21078

### DIFF
--- a/keras/src/export/saved_model_test.py
+++ b/keras/src/export/saved_model_test.py
@@ -773,6 +773,36 @@ class ExportArchiveTest(testing.TestCase):
         revived_model = tf.saved_model.load(temp_filepath)
         self.assertAllClose(ref_output, revived_model.serve(ref_input))
 
+    @pytest.mark.skipif(
+        backend.backend() != "tensorflow",
+        reason="String lookup requires TensorFlow backend",
+    )
+    def test_model_with_tracked_collection(self):
+        temp_filepath = os.path.join(self.get_temp_dir(), "exported_model")
+        text_vectorization = layers.TextVectorization()
+        text_vectorization.adapt(["one two", "three four", "five six"])
+
+        # CustomModel has a list of layers. The `TrackedList` that Keras uses is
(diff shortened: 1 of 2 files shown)


keras/src/export/saved_model_test.py
```diff
@@ -773,6 +773,36 @@ def test_model_with_lookup_table(self):
         revived_model = tf.saved_model.load(temp_filepath)
         self.assertAllClose(ref_output, revived_model.serve(ref_input))
 
+    @pytest.mark.skipif(
+        backend.backend() != "tensorflow",
+        reason="String lookup requires TensorFlow backend",
```
abheesht17, line 778: Do we need to use `TextVectorization` here? Can we pass in an tensor of floats, with the model being a list of embedding and dense layers? That way, we don't need to skip tests for other backends?

hertschuh: Well, first, the bug is very TF specific. It has to to with how `tf.__internal__.tracking.AutoTrackable` are recursed, which is only used with the TF backend.

- if we use a tensor that is not a variable, it will just be inlined in the graph, it doesn't need to be tracked as a resource, there won't be any error.
- if we use a variable, it's already tracked correctly via `model.variables`, which already discovers all variables of all tracked layers etc. using the Keras tracking (not the TF tracking).

`TextVectorization` is the easiest example I can think of. That's because it uses a lookup table, which is a resource that is not a variable, and it's not a tensor that's inlined in the graph. That's why the test above also uses `TextVectorization`.

abheesht17: Gotcha, thanks!

abheesht17 (approved):
LGTM

PR #22173, opened 2026-02-14
Implement norm function for OpenVINO backend with support for various…

aaishwarymishra:
This pull request implements the `norm` function for the OpenVINO backend, replacing the previous placeholder that raised a `NotImplementedError`. The new implementation supports several common norms and handles different values for the `ord` argument.

Math operations support:

* Implemented the `norm` function in `keras/src/backend/openvino/math.py`, supporting L2, L1, max (infinity), and min (-infinity) norms, with handling for different `axis` and `keepdims` values.
* Added an import for `numpy` as `np` to support norm calculations.… orders

Fixes: openvinotoolkit/openvino#34127

### DIFF
--- a/keras/src/backend/openvino/excluded_concrete_tests.txt
+++ b/keras/src/backend/openvino/excluded_concrete_tests.txt
@@ -189,7 +189,8 @@ LinalgOpsCorrectnessTest::test_eigh
 LinalgOpsCorrectnessTest::test_inv
 LinalgOpsCorrectnessTest::test_lstsq
 LinalgOpsCorrectnessTest::test_lu_factor
-LinalgOpsCorrectnessTest::test_norm_2
+LinalgOpsCorrectnessTest::test_norm_2_-2
+LinalgOpsCorrectnessTest::test_norm_2_2
 LinalgOpsDynamicShapeTest::test_qr
 NNOpsDynamicShapeTest::test_binary_crossentropy
 NNOpsDynamicShapeTest::test_categorical_crossentropy
--- a/keras/src/backend/openvino/linalg.py
+++ b/keras/src/backend/openvino/linalg.py
@@ -75,26 +75,30 @@ def norm(x, ord=None, axis=None, keepdims=False):
             # L2 norm: sqrt(sum(x * conj(x)))
             x_conj = x_ov
             x_sq = ov_opset.multiply(x_conj, x_conj).output(0)
-            axis_const = ov_opset.constant(axis, Type.i32).output(0)
+            axis_for_const = list(axis) if isinstance(axis, tuple) else axis
+            axis_const = ov_opset.constant(axis_for_const, Type.i32).output(0)
             norm_result = ov_opset.reduce_sum(
                 x_sq, axis_const, keepdims
             ).output(0)
             norm_result = ov_opset.sqrt(norm_result).output(0)
         elif ord == float("inf"):
-            axis_const = ov_opset.constant(axis, Type.i32).output(0)
+            axis_for_const = list(axis) if isinstance(axis, tuple) else axis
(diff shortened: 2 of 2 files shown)


keras/src/backend/openvino/math.py
```diff
@@ -125,4 +126,28 @@ def solve(a, b):
...
+    elif ord == 1:
+        norm = ov_opset.reduce_l1(x, axis_tensor, keepdims).output(0)
+    elif ord == np.inf or ord == "inf" or ord == float("inf"):
+        norm = ov_opset.reduce_max(
+            ov_opset.abs(x).output(0), axis_tensor, keepdims
+        ).output(0)
+    elif ord == -np.inf or ord == "-inf" or ord == float("-inf"):
+        norm = ov_opset.reduce_min(
+            ov_opset.abs(x).output(0), axis_tensor, keepdims
+        ).output(0)
```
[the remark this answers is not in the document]
aaishwarymishra: Oh, I am not sure how to handle it, I searched numpy,tensorflow and torch backends all of them passed the ord directly as they natively support norm , but for openvino there are different methods for each so I tried to cover all possible values, I can remove them if needed.

hertschuh:
Oh, and you'll need to approve the CLA too.

aaishwarymishra:
> Thank you for implementing this!
> 
> Can you:
> 
> 1. Move this code to `keras/src/backend/openvino/linalg.py`, this is definition in this file is a mistake.
> 2. Delete this line to enable the testing: https://github.com/keras-team/keras/blob/master/keras/src/backend/openvino/excluded_tests.txt#L32

So I removed the line from excluded_tests.py and ran the tests and they passed, the tests are language agnostic right after removing that line it should work with openvino or I have to do something else?

hertschuh:
> So I removed the line from excluded_tests.py and ran the tests and they passed, the tests are language agnostic right after removing that line it should work with openvino or I have to do something else?

So not all of linalg is implemented ord=nuc norm is not supported on OpenVino. Can add the following lines to
https://github.com/keras-team/keras/blob/master/keras/src/backend/openvino/excluded_concrete_tests.txt#L182 ?

```
LinalgOpsCorrectnessTest::test_norm_2_nuc
LinalgOpsCorrectnessTest::test_qr
LinalgOpsCorrectnessTest::test_solve
LinalgOpsCorrectnessTest::test_solve_triangular
LinalgOpsCorrectnessTest::test_svd
QrOpTest
```

aaishwarymishra:
I have implemented it similar to tensorflow norm implementation

keras/src/backend/openvino/excluded_concrete_tests.txt
```diff
@@ -178,6 +178,19 @@ TestMathErrors::test_stft_invalid_window
...
+LinalgOpsCorrectnessTest::test_solve
+LinalgOpsCorrectnessTest::test_solve_triangular
+LinalgOpsCorrectnessTest::test_svd
+LinalgOpsCorrectnessTest::test_det
+LinalgOpsCorrectnessTest::test_eig
+LinalgOpsCorrectnessTest::test_eigh
+LinalgOpsCorrectnessTest::test_inv
+LinalgOpsCorrectnessTest::test_lstsq
+LinalgOpsCorrectnessTest::test_lu_factor
+LinalgOpsCorrectnessTest::test_norm_2
```
hertschuh: Please remove this line: `LinalgOpsCorrectnessTest::test_norm_2`

The norm tests with 2 dimensions should pass, except of the `test_norm_2_nuc` ones (which are excluded above line 181).

aaishwarymishra: They were failing because the test_norm_2 was run with 2 and -2, which were for nuc so i just added them.

PR #22176, opened 2026-02-14
Implement numpy.lcm for OpenVINO backend

andersendsa:
**Details:**
Implemented numpy.lcm in keras/src/backend/openvino/numpy.py.
The implementation uses the existing gcd function and the formula lcm(a, b) = (|a| / gcd(a, b)) * |b|.
It handles the edge case where gcd(a, b) is 0 (which implies both inputs are 0) by replacing 0 with 1 in the divisor to avoid division by zero.
Enabled tests for lcm in keras/src/backend/openvino/excluded_concrete_tests.txt.
Verified tests pass with KERAS_BACKEND=openvino.
**Closes** : https://github.com/openvinotoolkit/openvino/issues/34055
**Ticket** : #34017

### DIFF
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -1713,7 +1713,7 @@ def lcm(x1, x2):
     x1_abs = ov_opset.abs(x1).output(0)
     x2_abs = ov_opset.abs(x2).output(0)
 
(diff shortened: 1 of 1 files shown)


andersendsa:
Hello @rkazants , @hertschuh This PR is ready for review, please have a look when you have the time.

PR #22178, opened 2026-02-14
Refactor ExportArchive to be a dispatcher for different export backends.

jerryxyj:
The existing ExportArchive is renamed to BackendSavedModelExportArchive. A new ExportArchive class is introduced, which acts as a factory. Its __new__ method instantiates either BackendSavedModelExportArchive or raises an error for other export types, based on the `export_model` argument. This change prepares for integrating other export backends.


keras/src/export/saved_model.py
```diff
@@ -34,13 +34,13 @@
...
     )
 
-
 DEFAULT_ENDPOINT_NAME = "serve"
 
 
-@keras_export("keras.export.ExportArchive")
-class ExportArchive(BackendExportArchive):
-    """ExportArchive is used to write SavedModel artifacts (e.g. for inference).
+@keras_export("keras.export.BackendSavedModelExportArchive")
```
hertschuh: Remove `@keras_export("keras.export.BackendSavedModelExportArchive")`, this should not be exported. (Exporting means making it part of the public API).

keras/src/export/saved_model.py
```diff
@@ -691,3 +691,21 @@ def _list_variables_used_by_fns(fns):
...
+
+@keras_export("keras.export.ExportArchive")
+class ExportArchive:
+    """This is the wrapper class for the backend-specific export archives.
+
+    It's for transitional purposes during the migration to the new export API.
+    """
+
+    def __new__(cls, export_model="backend_saved_model", **kwargs):
+        if export_model == "backend_saved_model":
```
hertschuh: It should be called `format` and have a default value of `"saved_model"` to be consistent with the `model.export` API. And for Orbax, the value should probably be `"orbax_model"`.

keras/src/export/saved_model.py
```diff
@@ -691,3 +691,21 @@ def _list_variables_used_by_fns(fns):
...
                     non_trainable_variables_ids.add(id(v))
     return trainable_variables, non_trainable_variables
+
+
+@keras_export("keras.export.ExportArchive")
+class ExportArchive:
+    """This is the wrapper class for the backend-specific export archives.
+
+    It's for transitional purposes during the migration to the new export API.
+    """
```
hertschuh: The issue is that this is the class that will be used to generate the documentation and the class that people will see in their IDE. So you need to:
- Move the docstring (line 43-106) to here
- Add empty versions of the methods: `track`, `add_endpoint`, `track_and_add_endpoint`, `add_variable_collection`, `writeout`, they should raise `NotImplementedError`, but they should have all the docstrings (moved from above).

hertschuh (commented):
The inheritance will also need to be inverted and the base class will need to extend this new `ExportArchive` class, which means you'll probably have to move this to a different file.

Will all of this be a separate PR?

keras/src/export/saved_model.py
jerryxyj: I see, I will use `format`, but will use different export model via getting from kwargs, since Orbax can export either saved model or orbax model.

jerryxyj:
Addressed all the comment

hertschuh (approved):
Thanks!

PR #22179, opened 2026-02-14
Define contiguous replica-group data shard ID convention in ModelParallel

MalyalaKarthik66:
- This PR defines the data sharding ID convention in `ModelParallel.distribute_dataset`, resolving the existing TODO.

- Previously, shard IDs were computed using modulo (%), which resulted in round-robin assignment of dataset shards across processes. This caused processes within the same replica group to receive different data when model replicas were sharded across multiple processes.

- This change replaces the modulo operation with integer division (//), which groups contiguous process ranks into the same replica group. This aligns with the row-major `DeviceMesh` layout semantics and ensures processes within a replica group consume identical shards.

- Additionally, a validation check is added to ensure that `num_process` is divisible by `num_model_replicas`, preventing out-of-range shard indices in misconfigured multi-process setups.

- An integration test is included to verify correct shard grouping behavior in a simulated multi-process environment.

### Contributor Agreement

### DIFF
--- a/keras/src/distribution/distribution_lib_test.py
+++ b/keras/src/distribution/distribution_lib_test.py
@@ -489,9 +489,7 @@ class DataShardingIntegrationTest(testing.TestCase):
         self.assertGreaterEqual(
             num_devices, 4, "Number of devices must be at least 4"
         )
-        self.assertEqual(
-            num_devices % 2, 0, "Number of devices must be even"
-        )
+        self.assertEqual(num_devices % 2, 0, "Number of devices must be even")
 
         num_model_replicas = num_devices // 2
         device_mesh = distribution_lib.DeviceMesh(


keras/src/distribution/distribution_lib.py
```diff
@@ -746,9 +746,16 @@ def distribute_dataset(self, dataset):
                 )
             per_process_batch_size = global_batch_size // num_model_replicas
             distributed_dataset = dataset.rebatch(per_process_batch_size)
+            if self._num_process % num_model_replicas != 0:
+                raise ValueError(
+                    "`num_process` must be divisible by `num_model_replicas`. "
+                    f"Got num_process={self._num_process}, "
+                    f"num_model_replicas={num_model_replicas}."
+                )
```
hertschuh: This is a good test to have, but I believe we can move it earlier so that people know they setup an invalid mesh for the number of processes. This should be done at the end of `__init__` above line 662.

keras/src/distribution/distribution_lib_test.py
```diff
@@ -11,6 +11,7 @@
 from keras.src import testing
 from keras.src.backend import distribution_lib as backend_dlib
 from keras.src.distribution import distribution_lib
+from keras.src.utils.module_utils import tensorflow as tf_mod
```
hertschuh: Tensorflow is already imported as tf, remove this.

keras/src/distribution/distribution_lib_test.py
```diff
@@ -477,6 +478,65 @@ def test_iter(self):
...
 
 
+class DataShardingIntegrationTest(testing.TestCase):
+    def test_distribute_dataset_sharding_behavior(self):
+        if not tf_mod.available:
+            self.skipTest(
+                "TensorFlow not available for dataset integration test"
+            )
+
+        tf = tf_mod
```
hertschuh: Remove

MalyalaKarthik66:
@hertschuh Thanks for the review — I’ve addressed all comments and updated the PR.

keras/src/distribution/distribution_lib.py
```diff
@@ -749,8 +764,9 @@ def distribute_dataset(self, dataset):
             per_process_batch_size = global_batch_size // num_model_replicas
             distributed_dataset = dataset.rebatch(per_process_batch_size)
             processes_per_replica = self._num_process // num_model_replicas
-            # TODO: Figure out what the convention is for data sharding id.
-            data_shard_id = self._process_id % processes_per_replica
+            # Each process belongs to a replica group. Determine which replica
+            # this process group belongs to.
+            data_shard_id = self._process_id // processes_per_replica
```
hertschuh, line 769: > This change replaces the modulo operation with integer division (//), which groups contiguous process ranks into the same replica group. This aligns with the row-major DeviceMesh layout semantics and ensures processes within a replica group consume identical shards.

About the `processes within a replica group consume identical shards`. But the whole point is that you want the replicated model to consume *different* data. Otherwise, if you're passing the same data to the same replicated model, you're just wasting time recomputing the same thing N times.

Am I missing something?

keras/src/distribution/distribution_lib_test.py
```diff
@@ -477,6 +477,59 @@ def test_iter(self):
...
 
 
+class DataShardingIntegrationTest(testing.TestCase):
+    def test_distribute_dataset_sharding_behavior(self):
+        devices = [f"cpu:{i}" for i in range(4)]
+        device_mesh = distribution_lib.DeviceMesh(
+            (2, 2),
+            ["data", "model"],
+            devices,
+        )
```
hertschuh, line 498: Please rebase and look at: https://github.com/keras-team/keras/pull/22484/changes#diff-519ae9e883cdaf7062167794507d3493a6641d90dea3bacc73d763d127546f01

In particular:
- add `@pytest.mark.multi_device` on the class
- don't hardcode the number of devices, use `distribution_lib.get_device_count()` and adapt the code to handle any number of devices
- don't mock anything, assume it will actually run with multiple devices

MalyalaKarthik66:
@hertschuh  Thanks for calling this out — that makes sense.

You are right that different data-parallel replicas should consume different data.

In this case, my understanding is that a single model replica can be sharded across multiple processes (model-parallel axis), so processes within the same replica group collaborate on one model replica and should consume the same data shard. Different replica groups (data-parallel axis) would still receive different shards.

Please let me know if this interpretation is incorrect or if the grouping should follow a different axis — happy to adjust the logic accordingly.

I’ve also rebased and updated the integration test to run as a real multi-device test (no mocks, dynamic device count, and @pytest.mark.multi_device).

PR #22180, opened 2026-02-14
[legacy] Resolve function strings using module_objects in serialization

MalyalaKarthik66:
- Passes `module_objects` to `get_registered_object` when resolving string-referenced functions in legacy serialization, completing the existing TODO.

- Ensures function lookup behavior is consistent with class resolution logic in `class_and_config_for_serialized_keras_object`.

- Keeps the change minimal and localized to `legacy/saving/serialization.py`, without affecting modern `.keras` serialization paths.

- Adds focused unit tests covering both successful resolution from `module_objects` and fallback behavior when the function is not present.


keras/src/legacy/saving/serialization_test.py
```diff
@@ -0,0 +1,56 @@
...
+    return x**2
+
+
+class DummyLayer:
+    """A dummy layer class for testing."""
+
+    pass
+
+
+class SerializationTest(testing.TestCase):
```
hertschuh: It appears that this legacy serialization code is all tested here:
https://github.com/keras-team/keras/blob/master/keras/src/legacy/saving/legacy_h5_format_test.py#L116

Can you move these tests there? I think what you need to do is very similar to this:
https://github.com/keras-team/keras/blob/master/keras/src/legacy/saving/legacy_h5_format_test.py#L206-L217

Except that you *don't* `register_keras_serializable` `MyDense` and that you pass a `custom_objects` in reload.

MalyalaKarthik66: Thanks for your review — I’ve updated the test to follow the legacy H5 reload pattern using `custom_object_scope` as suggested.

hertschuh (approved):
Thanks!

PR #22182, opened 2026-02-14
Add missing API methods to OpenVINOKerasTensor

### LINKED ISSUE #21544: OpenVINOKerasTensor does not implement the full expected API of KerasTensor
A variety of keras ops have paths that return objects of type `OpenVINOKerasTensor`. For example

```python
import keras.ops as ops
from typing import reveal_type

result = keras.ops.concatenate(...)
reveal_type(result) # OpenVINOKerasTensor
```

This is because the paths of this op route to `backend.numpy.concatenate(...)`

And `backend.numpy.concatenate` has the following implementation:

```python
def concatenate(xs, axis=0):
    assert isinstance(xs, list), "`concatenate` is supported only for `x` list"
    elems = []
    for elem in xs:
        elem = get_ov_output(elem)
        elems.append(elem)
    res = ov_opset.concat(elems, axis).output(0)
    return OpenVINOKerasTensor(res)
```

However, when inspecting the definition of `OpenVINOKerasTensor` [here](https://github.com/keras-team/keras/blob/b990e5467dcd232a9c60a93ba6dbdbefce4254af/keras/src/backend/openvino/core.py#L140), it doesn't implement a variety of the `__dunder__` methods expected (we see KerasTensor and frameworks like PyTorch, Jax, Numpy, and TFs objects all implement a much larger superset). Furthermore, it is missing an `__array__` implementation, which is surprising of the return of any op coming from `backend.numpy.some_operation`. 

When working with framework code that is supposed to cleanly be interoperable with other frameworks, which is a large selling point of Keras 3.x, the minimum expectation should be to abide by the same interface defined by `KerasTensor`, including overloads that are expected from non-symbolic tensors like `__array__` with errors just like `KerasTensor` has. The reason for this is that it makes it much easier to write properly typed downstream code with `Protocol`s for things like `ArrayLike` and `TensorLike`. Right now, all such code will get typing errors for any `ops` that have implementations through `backend.numpy`, which is not ideal.

timon0305:
## Summary
- Adds missing methods to `OpenVINOKerasTensor` for API parity with `KerasTensor`
- Arithmetic: `__matmul__`, `__rmatmul__`, `__div__`, `__rdiv__`, `__rmod__`
- Logical: `__and__`, `__rand__`, `__or__`, `__ror__`, `__xor__`, `__rxor__`
- Type conversion: `__int__`, `__float__`, `__bool__`, `__iter__`, `__repr__`
- Other: `__round__`, `reshape()`, `squeeze()`, `__jax_array__`, `__tf_tensor__`
- Resolves #21544

### DIFF
--- a/keras/src/backend/openvino/core.py
+++ b/keras/src/backend/openvino/core.py
@@ -602,18 +602,22 @@ class OpenVINOKerasTensor:
         return OpenVINOKerasTensor(ov_opset.logical_xor(other, first).output(0))
 
     def __int__(self):
-        raise ValueError(
-            "An OpenVINOKerasTensor is symbolic: it's a placeholder "
-            "for a shape and a dtype. It doesn't have any actual "
-            "numerical value. You cannot convert it to an int."
-        )
+        arr = self.output.get_node().data
+        if arr.ndim > 0:
+            raise TypeError(
+                "Only scalar arrays can be converted to Python scalars. "
+                f"Got: shape={arr.shape}"
+            )
+        return int(arr)
 
     def __float__(self):
-        raise ValueError(
-            "An OpenVINOKerasTensor is symbolic: it's a placeholder "
-            "for a shape and a dtype. It doesn't have any actual "
-            "numerical value. You cannot convert it to a float."
-        )
+        arr = self.output.get_node().data
+        if arr.ndim > 0:
+            raise TypeError(
+                "Only scalar arrays can be converted to Python scalars. "
+                f"Got: shape={arr.shape}"
+            )
+        return float(arr)
 
     def __repr__(self):
         return f"<OpenVINOKerasTensor shape={self.shape}, dtype={self.dtype}>"


keras/src/backend/openvino/core.py
```diff
@@ -519,6 +519,149 @@ def __array__(self, dtype=None):
...
+            "dtype, used during inference with the OpenVINO backend."
+        )
+
+    def __tf_tensor__(self, dtype=None, name=None):
+        raise ValueError(
+            "An OpenVINOKerasTensor cannot be used as input to a "
+            "TensorFlow function. It is a symbolic placeholder for a "
+            "shape and dtype, used during inference with the OpenVINO "
+            "backend."
+        )
```
hertschuh: Remove these, they make sense for `KerasTensor`, not for `OpenVINOKerasTensor`.

keras/src/backend/openvino/core.py
```diff
@@ -519,6 +519,149 @@ def __array__(self, dtype=None):
...
+            "for a shape and a dtype. It doesn't have any actual "
+            "numerical value. You cannot convert it to an int."
+        )
+
+    def __float__(self):
+        raise ValueError(
+            "An OpenVINOKerasTensor is symbolic: it's a placeholder "
+            "for a shape and a dtype. It doesn't have any actual "
+            "numerical value. You cannot convert it to a float."
+        )
```
hertschuh: These two actually do make sense for `OpenVINOKerasTensor` (they don't for `KerasTensor`).

But the underlying tensor needs to be a scalar.

hertschuh (commented):
Thanks for adding these!

PR #22185, opened 2026-02-16
Ops with a `dtype` argument use `standardize_dtype` in non-symbolic…

hertschuh:
…cases.

Previously, `standardize_dtype` was only used in the op class, causing the op to behave differently in a functional model vs. when called on tensors.

Also fixed `cumprod` and `cumsum`, which didn't handle symbolic calls correctly.


abheesht17 (approved):
Thanks!

PR #22191, opened 2026-02-16
fix(preprocessing): preserve segmentation mask dtype in augmentation layers

### LINKED ISSUE #20857: Several Problems when using Keras Image Augmentation Layers to Augment Images and Masks for a Semantic Segmentation Dataset
In the following, I will try to tell you a bit of my “story” about how I read through the various layers and (partly outdated) tutorials on image augmentation for semantic segmentation, what problems and bugs I encountered and how I was able to solve them, at least for my use case. I hope that I am reporting these issue(s) in the right place and that I can help the developers responsible with my “story”.

I am currently using the latest versions of Keras (3.8) and Tensorflow (2.18.0) on an Ubuntu 23.10 machine, but have also observed the issues with other setups.

My goal was to use some of the Image Augmentation Layers from Keras to augment the images AND masks in my semantic segmentation dataset. To do this I followed the [segmentation tutorial from Tensarflow](https://www.tensorflow.org/tutorials/images/segmentation) where a custom `Augment` class is used for augmentation.

This code works fine because the `Augment` class uses only a single augmentation layer. But, for my custom code I wanted to use more augmentations. Therefore, I "combined" several of the augmentation layers trying both the normal [Pipeline layer](https://keras.io/api/layers/preprocessing_layers/image_augmentation/pipeline/) but also my preferred [RandomAugmentationPipeline layer from keras_cv](https://github.com/keras-team/keras-cv/blob/master/keras_cv/src/layers/preprocessing/random_augmentation_pipeline.py).

I provide an example implementation in [this gist](https://colab.research.google.com/gist/RabJon/9f27d600301dcf6a734e8b663c802c7f/modification_of_segmentation.ipynb). The gist is a modification of the aforementioned [image segmentation tutorial from Tensorflow](https://www.tensorflow.org/tutorials/images/segmentation) where I intended to only change the `Augment` class slightly to fit my use case and to remove all the unnecessary parts. Unfortunately, I also had to touch the code that loads the dataset, because the dataset version used in the notebook was no longer supported. But, this is another issue. 

Anyway, at the bottom of the notebook you can see some example images and masks from the data pipeline and if you inspect the in more detail you can see that for some examples images and masks do not match because they were augmented differently, which is obviously not the expected behaviour. You can also see that these mismatches already happen from the first epoch on. On my system with my custom use case (where I was using the RandomAugmentationPipeline layer) it happend only from the second epoch onward, which made debugging much more difficult.

I first assumed that it was one specific augmentation layer that caused the problems, but after trying them one-by-one I found out that it is the combination of layers that makes the problems. So, I started to think about possible solutions and I also tried to use the Sequential model from keras to combine the layers instead of the Pipeline layer, but the result remained the same.

I found that the random seed is the only parameter which can be used to "control and sync" the augmentation of the images and masks, but I already had ensured that I always used the same seed. So, I started digging into the source code of the different augmentation layers and I found that most of the ones that I was using implement a `transform_segmentation_masks()` method which is however not used if the layers are called as described in the tutorial. This is because, in order to enforce the call of this method, images and masks must be passed to the same augmentation layer object as a dictionary with the keys “images” and “segmentation_masks” instead of using two different augmentation layer objects for images and masks. However, I had not seen this type of call using a dictionary in any tutorial, neither for Tensorflow nor for Keras. Nevertheless, I decided to change my code to make use of the `transform_segmentation_masks()` method, as I hoped that if such a method already existed, it would also process the images and masks correctly and thus avoid the mismatches. 
Unfortunately, this was not directly the case, because some of the augmentation layers changed the masks to float32 data types although the input was uint8. Even with layers such as “RandomFlip”, which should not change or interpolate the data at all, but only shift it. So, I had to wrap all layers again in a custom layer, which castes the data back to the input data type before returning it:

```python
class ApplyAndCastAugmenter(BaseImageAugmentationLayer):
    
    def __init__(self, other_augmenter, **kwargs):
        super().__init__(**kwargs)
        
        self.other_augmenter = other_augmenter
    
    def call(self, inputs):
        
        output_dtypes = {}
        is_dict = isinstance(inputs, dict)
        if is_dict:
            for key, value in inputs.items():
                output_dtypes[key] = value.dtype
        else:
            output_dtypes = inputs.dtype
        
        outputs = self.other_augmenter(inputs)
        
        if is_dict:
            for key in outputs.keys():
                outputs[key] = keras.ops.cast(outputs[key], output_dtypes[key])
        else:
            outputs = keras.ops.cast(outputs, output_dtypes)
        
        return outputs
```

With this (admittedly unpleasant) work-around, I was able to fix all the remaining problems. Images and masks finally matched perfectly after augmentation and this was immediately noticeable in a significant performance improvement in the training of my segmentation model.

For the future I would now like to see that wrapping with my custom `ApplyAndCastAugmenter` is no longer necessary but is handled directly by the `transform_segmentation_masks()` method and it would also be good to have a proper tutorial on image augmentation for semantic segmentation or to update the old tutorials.

timon0305:
## Summary

Multiple image augmentation layers in `keras.layers.preprocessing` cast segmentation masks to `float32` when they should preserve the original dtype (e.g., `uint8`). This happens because `transform_segmentation_masks()` delegates to `transform_images()`, which internally casts inputs to the layer's `compute_dtype`.

This PR fixes 13 augmentation layers to preserve the original segmentation mask dtype by:
1. Saving the input dtype before calling `transform_images()`
2. Rounding the output for integer dtypes (to handle interpolation artifacts)
3. Casting back to the original dtype

### Affected layers
- `AugMix`
- `CenterCrop`
- `CutMix`
- `MaxNumBoundingBoxes`
- `RandAugment`
- `RandomCrop`
- `RandomElasticTransform`
- `RandomFlip`
- `RandomPerspective`
- `RandomRotation`
- `RandomShear`
- `RandomTranslation`
- `RandomZoom`

### Reproduction

```python
import numpy as np
import keras
from keras import layers

masks = np.random.randint(0, 5, (2, 32, 32, 1), dtype=np.uint8)
data = {
    'images': np.random.rand(2, 32, 32, 3).astype('float32'),
    'segmentation_masks': masks,
}

layer = layers.RandomFlip()
output = layer(data, training=True)
print(output['segmentation_masks'].dtype)  # Before: float32, After: uint8
```

Fixes #20857

### DIFF
--- a/keras/src/layers/preprocessing/image_preprocessing/aug_mix.py
+++ b/keras/src/layers/preprocessing/image_preprocessing/aug_mix.py
@@ -1,6 +1,7 @@
 import random as py_random
 
 import keras.src.layers as layers
+from keras.src import backend
 from keras.src.api_export import keras_export
 from keras.src.layers.preprocessing.image_preprocessing.base_image_preprocessing_layer import (  # noqa: E501
     BaseImagePreprocessingLayer,
@@ -313,9 +314,15 @@ class AugMix(BaseImagePreprocessingLayer):
     def transform_segmentation_masks(
         self, segmentation_masks, transformation, training=True
     ):
-        return self.transform_images(
+        original_dtype = segmentation_masks.dtype
+        output = self.transform_images(
             segmentation_masks, transformation, training=training
         )
+        if output.dtype == original_dtype:
+            return output
+        if backend.is_int_dtype(original_dtype):
+            output = self.backend.numpy.round(output)
+        return self.backend.cast(output, original_dtype)
 
     def compute_output_shape(self, input_shape):
         return input_shape
--- a/keras/src/layers/preprocessing/image_preprocessing/center_crop.py
+++ b/keras/src/layers/preprocessing/image_preprocessing/center_crop.py
@@ -1,3 +1,4 @@
+from keras.src import backend
 from keras.src.api_export import keras_export
 from keras.src.layers.preprocessing.image_preprocessing.base_image_preprocessing_layer import (  # noqa: E501
     BaseImagePreprocessingLayer,
@@ -177,9 +178,15 @@ class CenterCrop(BaseImagePreprocessingLayer):
     def transform_segmentation_masks(
         self, segmentation_masks, transformation, training=True
     ):
-        return self.transform_images(
+        original_dtype = segmentation_masks.dtype
+        output = self.transform_images(
             segmentation_masks, transformation, training=training
         )
+        if output.dtype == original_dtype:
+            return output
+        if backend.is_int_dtype(original_dtype):
+            output = self.backend.numpy.round(output)
+        return self.backend.cast(output, original_dtype)
 
     def transform_images(self, images, transformation=None, training=True):
         inputs = self.backend.cast(images, self.compute_dtype)
--- a/keras/src/layers/preprocessing/image_preprocessing/cut_mix.py
+++ b/keras/src/layers/preprocessing/image_preprocessing/cut_mix.py
@@ -1,3 +1,4 @@
+from keras.src import backend
 from keras.src.api_export import keras_export
 from keras.src.layers.preprocessing.image_preprocessing.base_image_preprocessing_layer import (  # noqa: E501
     BaseImagePreprocessingLayer,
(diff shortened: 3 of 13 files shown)


hertschuh (dismissed):
Thanks for the fix!

hertschuh:
This was addressed as part of https://github.com/keras-team/keras/pull/22192

PR #22203, opened 2026-02-18
Add test cases for OrbaxCheckpoint sharding feature

amitsrivastava78:
Added test cases for Sharding & multihost feature with virtual CPU devices


keras/src/callbacks/orbax_checkpoint_distributed_test.py
```diff
@@ -0,0 +1,851 @@
...
+
+        Kernel shapes: dense_1 (16, 8), dense_2 (8, 8), output (8, 4).
+        """
+        inputs = layers.Input(shape=(16,), name="input_layer")
+        x = layers.Dense(8, activation="relu", name="dense_1")(inputs)
+        x = layers.Dense(8, activation="relu", name="dense_2")(x)
+        outputs = layers.Dense(4, name="output_layer")(x)
+        model = models.Model(inputs, outputs, name="deep_sharding_model")
+        model.compile(optimizer="adam", loss="mse", jit_compile=False)
+        return model
```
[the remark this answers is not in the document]
amitsrivastava78: Done

keras/src/callbacks/orbax_checkpoint_distributed_test.py
```diff
@@ -0,0 +1,851 @@
...
+    )
+    def test_resharding(
+        self,
+        save_devices,
+        load_devices,
+        save_axis,
+        load_axis,
+        save_layout,
+        load_layout,
+    ):
```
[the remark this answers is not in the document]
amitsrivastava78: Done

amitsrivastava78:
> After looking at the changes in `orbax_utils.py` and `saving_api.py`, here is my thinking:
> 
> Wouldn't it be a lot simpler to make the `model_config` a separate checkpointable and not part of the `pytree`?
> 
> I think it would be better because:
> 
> * `assets` is also a separate checkpointable, it makes it more consistent because both `assets` and `model_config` only exist when you use `saved_weights_only=False`.
> * `pytree` is purely arrays, no strings or anything funny, which is cleaner.
> * `pytree` is the same regardless of whether you use `saved_weights_only=True` or `saved_weights_only=False`.
> 
> What do you think?

Done — model_config is now a separate checkpointable, no longer part of the pytree

keras/src/saving/saving_api.py
```diff
@@ -358,10 +363,18 @@ def load_weights(model, filepath, skip_mismatch=False, **kwargs):
...
-    """Load a model from an Orbax checkpoint directory."""
+    """Load a model from an Orbax checkpoint directory.
 
-    from keras.src.utils.module_utils import ocp
+    ``model_config`` is stored as its own checkpointable (separate from
+    ``pytree``), so loading proceeds in two simple steps:
 
+      1. Load the ``model_config`` checkpointable to obtain the model
+         configuration string and rebuild the model.
+      2. Load the ``pytree`` checkpointable (all arrays).  When a JAX
```
hertschuh: The backquotes have to be single backquotes. These double backquotes won't render correctly.

keras/src/saving/orbax_util.py
```diff
@@ -52,3 +54,83 @@ def find_latest_orbax_checkpoint(checkpoint_dir):
...
+        checkpoint_path: Path to a specific Orbax checkpoint step
+            directory (e.g. ``<root>/2``).
+        ref_state: Optional reference state tree (from
+            ``model.get_state_tree()``) whose variables carry the
+            target shardings. If ``None``, shardings default to
+            ``None`` per leaf (Orbax uses saved shardings).
+
+    Returns:
+        A pytree of ``jax.ShapeDtypeStruct`` matching the checkpoint
+        structure, or ``None`` when resharding is not needed.
```
hertschuh: The backquotes have to be single backquotes. These double backquotes won't render correctly.

keras/src/saving/orbax_util.py
```diff
@@ -52,3 +54,83 @@ def find_latest_orbax_checkpoint(checkpoint_dir):
...
+    """
+    if backend.backend() != "jax":
+        return None
+
+    if get_distribution() is None:
+        return None
+
+    import jax
+
+    pytree_meta = ocp.pytree_metadata(checkpoint_path).metadata
```
hertschuh, line 89: My first inclination was that line 90-136 could be simplified like this:

```python
def to_shape_dtype_struct(ref, meta):
    if hasattr(meta, "shape") and hasattr(meta, "dtype"):
        sharding = getattr(ref, "sharding", None)
        return jax.ShapeDtypeStruct(x.shape, x.dtype, sharding=sharding)
    return None

tree.map_structures(ref_state, pytree_meta)
```

But more comments below

keras/src/saving/orbax_util.py
```diff
@@ -52,3 +54,83 @@ def find_latest_orbax_checkpoint(checkpoint_dir):
...
+                out.update(_collect_shardings(v, path + (k,)))
+            return out
+        return {}
+
+    ref_shardings = {}
+    if ref_state:
+        for group in (
+            "trainable_variables",
+            "non_trainable_variables",
+        ):
```
hertschuh: Why are optimizer variables and metrics excluded here? I think they should be included.

Optimizers are allowed to do whatever they want in terms of sharding. Sure, a common scenario is the apply the same sharding to let's say momentums as for the model variable, but that's not guaranteed.

For instance, DistributedEmbedding uses a very custom sharding (to SparseCore) that does not apply to the optimizer.

keras/src/saving/orbax_util.py
```diff
@@ -52,3 +54,83 @@ def find_latest_orbax_checkpoint(checkpoint_dir):
...
+    def _to_abstract(meta, ref=None, path=()):
+        """Convert metadata leaf → jax.ShapeDtypeStruct with sharding."""
+        if hasattr(meta, "shape") and hasattr(meta, "dtype"):
+            # Array leaf: use ref sharding when available, otherwise
+            # try matching an optimizer slot path to a model variable
+            # path (strip the optimizer-name prefix and slot-name
+            # suffix: path[1:-1]).
+            sharding = getattr(ref, "sharding", None)
+            if sharding is None and len(path) >= 3:
+                sharding = ref_shardings.get(path[1:-1])
```
hertschuh: I don't think we should do that. Optimizers can do whatever they want. We should instead use the original sharding from the optimizers.

Also this `path[1:-1]` makes me nervous that we could randomly pickup random shardings that happen to match a subpart of the path.

What was the reason for this?

keras/src/saving/saving_api.py
amitsrivastava78: Fixed

keras/src/saving/orbax_util.py
amitsrivastava78: Fixed

keras/src/saving/orbax_util.py
amitsrivastava78: Agreed — removed the path[1:-1] inheritance logic. Each variable (including optimizer and metrics) now uses its own sharding directly from the live model's state tree. No cross-path matching.

keras/src/saving/orbax_util.py
amitsrivastava78: Agreed — this code has been removed entirely. The simplified version now walks ref_state and pytree_meta in parallel, so every variable group (trainable, non-trainable, optimizer, metrics) gets its own sharding directly from the live model

keras/src/saving/orbax_util.py
amitsrivastava78: Good idea — I've simplified the implementation along similar lines. The _collect_shardings helper, path  tracking, and path[1:-1] optimizer inheritance are all removed. The function now just recursively walks ref_state and pytree_meta in parallel.

amitsrivastava78:
Updated the comments please check

Issue #22199, opened 2026-02-17
[Bug] Missing validation for sequence_stride in stft leads to negative stride acceptance and backend crashes


amadhan882:
### Description

The `keras.ops.stft` operation lacks proper input validation for the `sequence_stride` parameter. This leads to two specific issues across all backends:

1.  **Negative Stride Failure:** A negative `sequence_stride` causes the operation to return an unexpected object type (tuple) instead of a tensor, leading to `'tuple' object has no attribute 'shape'` errors during downstream processing. This indicates a silent internal failure.
2.  **Zero Stride Crash:** A zero `sequence_stride` leaks to the backend and causes a raw `InvalidArgumentError` (Integer division by zero) instead of a unified Keras `ValueError`.

### Minimal Reproducible Example
```python
import os
import numpy as np
import keras
import keras.ops as ops

# Set backend before running
# os.environ["KERAS_BACKEND"] = "tensorflow"

def reproduce_stft_bug():
    print(f"Testing Backend: {keras.backend.backend()}")
    x = ops.convert_to_tensor(np.random.rand(1, 1024).astype("float32"))

    print("\n--- Test 1: Negative Stride ---")
    try:
        # Expected: ValueError | Observed: Returns a tuple, causing AttributeError on .shape
        res_neg = ops.stft(x, sequence_length=256, sequence_stride=-5, fft_length=256)
        print(f"Result shape: {res_neg.shape}")
    except Exception as e:
        print(f"Caught negative stride issue: {e}")

    print("\n--- Test 2: Zero Stride ---")
    try:
        # Expected: ValueError | Observed: Raw Backend Crash
        res_zero = ops.stft(x, sequence_length=256, sequence_stride=0, fft_length=256)
    except Exception as e:
        print(f"Observed: Caught by backend with {type(e).__name__}: {e}")

if __name__ == "__main__":
    reproduce_stft_bug()
```
### Observed Output
```
Testing Backend: tensorflow

--- Test 1: Negative Stride ---
SUCCESS: Caught negative stride: 'tuple' object has no attribute 'shape'

--- Test 2: Zero Stride ---
OBSERVED: Caught by backend with InvalidArgumentError: {{function_node __wrapped__FloorDiv_device_/job:localhost/replica:0/task:0/device:GPU:0}} Integer division by zero [Op:FloorDiv] name: 
NOTE: This should be caught by Keras validation as a ValueError, not a raw backend error.


```
### Observed Behavior
[...]

PR #22205, opened 2026-02-18
Add validation for sequence_stride to prevent negative or zero values.

maitry63:
This PR adds validation for sequnce_stride to prevent negative or zero values in stft.

Fixes : [22199](https://github.com/keras-team/keras/issues/22199)


maitry63:
Already raised PR for this issue.

PR #22209, opened 2026-02-18
[OpenVINO Backend] included tests for bitwise_left_shift and bitwise_right_shift

goyaladitya05:
bitwise_left_shift and bitwise_right_shift for the OpenVINO backend, was previously implemented in #22146, but some tests were left excluded. 
 
 Those tests have been included here.


goyaladitya05:
@hertschuh Re-enabled the two previously excluded tests. Would appreciate a quick look when you get a chance.

PR #22212, opened 2026-02-18
fix(docs): correct docstring method name in Layer

timon0305:
## Description

Fixes a docstring typo in `keras/src/layers/layer.py`.

The example code referenced `seq.register_call_context_args("foo_mode")` but the correct method name is `seq._register_call_context_args("foo_mode")`.

## Files Changed
- `keras/src/layers/layer.py` — docstring method name fix (`register_call_context_args` → `_register_call_context_args`)

### DIFF
--- a/keras/src/backend/jax/trainer.py
+++ b/keras/src/backend/jax/trainer.py
@@ -52,7 +52,6 @@ class JAXTrainer(base_trainer.Trainer):
         kwargs = {}
         if self._call_has_training_arg:
             kwargs["training"] = training
-        kwargs.update(self._call_context_kwargs)
 
         # Run stateless forward pass
         y_pred, non_trainable_variables, losses = self.stateless_call(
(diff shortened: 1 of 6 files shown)


keras/src/layers/layer.py
```diff
@@ -1821,7 +1821,7 @@ def call(self, x):
 
         # Tell the Sequential model to propagate foo_mode down
         # the call-stack
-        seq.register_call_context_args("foo_mode")
+        seq._register_call_context_args("foo_mode")
```
hertschuh, line 1824: @timon0305 

Sorry, I should have closed this bug. This feature is not intended to be used this way. And it's not intended to be used at all actually.

However, you did catch a mistake in the docstring though.

Can you revert all the files except this one?

timon0305: Done

PR #22134, opened 2026-02-09
[OpenVINO backend] Support numpy.trunc

goyaladitya05:
Implements `numpy.trunc` operation for the OpenVINO backend.

Ref: openvinotoolkit/openvino/issues/34025

PR #22214, opened 2026-02-19
[OpenVINO Backend] included tests for numpy.trunc

goyaladitya05:
These tests were included previously in #22134, but got excluded again most likely due to a faulty rebase.

Those tests have been included again here.


goyaladitya05:
@hertschuh Re-enabled two previously excluded tests. Would really appreciate a quick look when you get a chance.

Issue #22213, opened 2026-02-18
[Feature]Integration Keras with Orbax Export


jerryxyj:
Currently, Keras exports JAX models by utilizing the jax2tf function to convert them into TensorFlow models, which are then persisted via SavedModel. However, Orbax has emerged as the modern standard for persisting JAX models and more. To enable native support for the Orbax export format, we need to integrate the Orbax library into the Keras ecosystem."

jerryxyj:
@sachinprasadhs  I'm working on this for Keras, please free to reassign this to me. This integration is approved internally at Google with @hertschuh

dataCenter430:
Hey, @jerryxyj 
are you still on this implementation?
otherwise, I'd love to help you with this feature as a contributor.
may I pick this up?

jerryxyj:
> Hey, [@jerryxyj](https://github.com/jerryxyj) are you still on this implementation? otherwise, I'd love to help you with this feature as a contributor. may I pick this up?

Hi @dataCenter430,
Thanks for offering this! Currently Orbax Model is in the transition from repository: https://github.com/google/orbax/tree/main/model

to a new repository, while the new one is still private and under the migration.

I talked to some Keras folks @hertschuh , and we agreed to pause this development until Orbax Model(future name is Neptune Model) is ready then we directly integrate with new Orbax Model repository.

dataCenter430:
ok, very good.
I'd love to have the chance to contribute to Keras.
please let me know if you have another issues, so I can pick it up.
Thanks

PR #22215, opened 2026-02-19
[Refactor]Inverse inheritance of ExportArchive(#22213)

jerryxyj:
Refactored ExportArchive to resolve an inverted inheritance pattern; backend-specific implementations now reside in child classes instead of the parent base class.


keras/src/export/saved_model.py
```diff
@@ -770,114 +242,6 @@ def add_endpoint(self, name, fn, input_signature=None, **kwargs):
...
-        model has a single input argument:
-
-        ```python
-        export_archive = ExportArchive()
-        export_archive.track(model)
-        export_archive.add_endpoint(
-            name="serve",
-            fn=model.call,
-            input_signature=[keras.InputSpec(shape=(None, 3), dtype="float32")],
-        )
```
hertschuh, line 788: Why was all of this documentation removed? It should stay.

keras/src/export/saved_model.py
```diff
@@ -921,28 +285,6 @@ def add_variable_collection(self, name, variables):
...
-        # Save a variable collection
-        export_archive.add_variable_collection(
-            name="optimizer_variables", variables=model.optimizer.variables)
-        export_archive.write_out("path/to/location")
-
-        # Reload the object
-        revived_object = tf.saved_model.load("path/to/location")
-        # Retrieve the variables
-        optimizer_variables = revived_object.optimizer_variables
-        ```
```
hertschuh, line 947: Why was all of this documentation removed? It should stay.

keras/src/export/saved_model_export_archive.py
```diff
@@ -0,0 +1,467 @@
...
+        self._endpoint_signatures[name] = input_signature
+        setattr(self._tf_trackable, name, decorated_fn)
+        self._endpoint_names.append(name)
+        return decorated_fn
+
+    def _add_endpoint_helper(self, name, fn, input_signature, **kwargs):
+        """Backend-specific endpoint creation. Override in subclasses."""
+        raise NotImplementedError(
+            "_add_endpoint_helper() must be implemented in backend subclasses."
+        )
```
hertschuh: Let's call it `_backend_add_endpoint` for consistency.

keras/src/export/saved_model_export_archive.py
```diff
@@ -0,0 +1,467 @@
...
+                "Invalid resource type. Expected a Keras `Layer` or `Model` "
+                "or a TensorFlow `Trackable` object. "
+                f"Received object {resource} of type '{type(resource)}'. "
+            )
+
+    def _track_layer(self, layer):
+        """Track variables of a layer. Override in backend subclasses."""
+        raise NotImplementedError(
+            "_track_layer() must be implemented in backend subclasses."
+        )
```
hertschuh: Let's rename it `_backend_track_layer` for consistency.

keras/src/export/saved_model_export_archive.py
```diff
@@ -0,0 +1,467 @@
...
+
+    def add_endpoint(self, name, fn, input_signature=None, **kwargs):
+        """Register a new serving endpoint.
+
+        Args:
+            name: `str`. The name of the endpoint.
+            fn: A callable. It should only leverage resources
```
hertschuh: Note that the docstrings in this class here won't be used for anything. I think we should remove them to avoid confusion (people might modify these instead of the other ones that are actually used).

On the other hand, the methods that need an override should have a docstring:
- _track_layer (document the `layer` argument)
- _add_endpoint_helper (document the arguments)

jerryxyj:
@hertschuh can you please review this pr?

keras/src/export/saved_model_export_archive.py
```diff
@@ -0,0 +1,467 @@
...
+        # special handling (e.g., Torch backend).
+        decorated_fn = self._track_and_add_endpoint_helper(
+            name, resource, input_signature, **kwargs
+        )
+        self._endpoint_signatures[name] = input_signature
+        setattr(self._tf_trackable, name, decorated_fn)
+        self._endpoint_names.append(name)
+        return decorated_fn
+
+    def _track_and_add_endpoint_helper(
```
hertschuh: Let's do something simpler here for `track_and_add_endpoint` (which we couldn't do before because of the inverted inheritance):

- `track_and_add_endpoint` should just be this here:
```python
    def track_and_add_endpoint(self, name, resource, input_signature, **kwargs):
        self.track(resource)
        return self.add_endpoint(
            name, resource.__call__, input_signature, **kwargs
        )
```
- remove `_track_and_add_endpoint_helper`
- in `torch/export.py`, override the actual `track_and_add_endpoint` instead of `_track_and_add_endpoint_helper`
- move the code from lines 224-257 to `torch/export.py` within `track_and_add_endpoint`

This way the default implementation is clean. And the logic lines 224-257, which is only really needed for torch lives in the torch file only.

PR #22222, opened 2026-02-19
[OpenVINO backend] Improve numpy.concatenate implemenation

goyaladitya05:
This PR improves the concatenate operation in the OpenVINO backend to handle type promotion and flattening, aligning it with NumPy behavior and ensuring compatibility with Keras ops tests.

## Key Changes
- OpenVINO's `Concat` operation requires all input tensors to share the exact same element type. Added logic to calculate the common result type using `dtypes.result_type` and cast all inputs to that target type.
- Added support for the `axis=None` case by flattening all input tensors before concatenation.
- Removed the strict list type check to allow tuples and other iterables, consistent with other Keras backends.
- Re-enabled relevant tests in `excluded_concrete_tests.txt` that were previously failing due to type mismatches.

Closes: openvinotoolkit/openvino/issues/34214


hertschuh (approved):
Thank you for the improvements!

PR #22224, opened 2026-02-20
implemented numpy.view for openvino backend.

geeky33:
### Details:
Implement numpy.view for OpenVINO backend

### Description:
Implements `numpy.view()` for the OpenVINO backend, enabling bitcast-style `dtype` reinterpretation of tensor data. Uses NumPy's .view() on the underlying node data since OpenVINO has no native bitcast op.

Closes issue : https://github.com/openvinotoolkit/openvino/issues/34217

### DIFF
--- a/keras/src/backend/openvino/excluded_concrete_tests.txt
+++ b/keras/src/backend/openvino/excluded_concrete_tests.txt
@@ -2,7 +2,6 @@ NumpyDtypeTest::test_add_
 NumpyDtypeTest::test_angle
 NumpyDtypeTest::test_array
 NumpyDtypeTest::test_heaviside
-NumpyDtypeTest::test_kaiser
 NumpyDtypeTest::test_concatenate
 NumpyDtypeTest::test_einsum
 HistogramTest::test_histogram_predict_jit_compile_false
@@ -21,7 +20,6 @@ NumpyDtypeTest::test_subtract
 NumpyDtypeTest::test_var
 NumpyOneInputOpsCorrectnessTest::test_angle
 NumpyOneInputOpsCorrectnessTest::test_array
-NumpyOneInputOpsCorrectnessTest::test_kaiser
 NumpyOneInputOpsCorrectnessTest::test_conj
 NumpyOneInputOpsCorrectnessTest::test_imag
 NumpyOneInputOpsCorrectnessTest::test_isreal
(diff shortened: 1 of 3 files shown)


geeky33:
Hi @rkazants , @hertschuh This PR pr passes all the tests and it is ready for the review.

please have a look at this whenever you get time :)
waiting for your response!

thanks!!

keras/src/backend/openvino/numpy.py
goyaladitya05, line 1: Please enable the relevent tests for this function from `keras\src\backend\openvino\excluded_concrete_tests.txt`

goyaladitya05:
Hi @geeky33, could you please link the correct issue?

keras/src/backend/openvino/numpy.py
geeky33: enabled it now
thanks for the help!! :)

goyaladitya05:
> Hi @geeky33, could you please link the correct issue?

Please link this issue https://github.com/openvinotoolkit/openvino/issues/34217.
The maintainers were using proper milestoning for each sub-issue, so it would be better to create a issue and then link the PR.
Thanks for the contribution!

geeky33:
Hi @rkazants , @hertschuh This PR pr passes all the tests and it is ready for the review.

please have a look at this whenever you get time :)
waiting for your response!

thanks!!

keras/src/backend/openvino/numpy.py
```diff
@@ -542,7 +543,18 @@ def array(x, dtype=None):
...
-    raise NotImplementedError("`view` is not supported with openvino backend")
+    x = get_ov_output(x)
+    old_ov_type = x.get_element_type()
+    old_dtype = ov_to_keras_type(old_ov_type)
+
+    if dtype is None:
+        dtype = old_dtype
+    new_dtype = standardize_dtype(dtype)
+    np_array = convert_to_numpy(OpenVINOKerasTensor(x))
+    result = np_array.view(np.dtype(new_dtype))
```
hertschuh, line 554: Oh, but you can't do that. This is an OpenVino only reimplementation of the numpy API, so you can't use Numpy (otherwise we would just use Numpy everywhere here).

keras/src/backend/openvino/excluded_concrete_tests.txt
```diff
@@ -30,13 +29,14 @@ NumpyOneInputOpsCorrectnessTest::test_reshape
...
-NumpyOneInputOpsCorrectnessTest::test_view
 NumpyTwoInputOpsCorrectnessTest::test_einsum
 NumpyTwoInputOpsCorrectnessTest::test_nextafter
 NumpyTwoInputOpsCorrectnessTest::test_quantile
 NumpyOneInputOpsDynamicShapeTest::test_angle
+NumpyOneInputOpsDynamicShapeTest::test_kaiser
 NumpyOneInputOpsDynamicShapeTest::test_view
 NumpyOneInputOpsStaticShapeTest::test_angle
+NumpyOneInputOpsStaticShapeTest::test_isreal
 NumpyOneInputOpsStaticShapeTest::test_view
```
hertschuh, line 40: You should remove all of these:
NumpyOneInputOpsDynamicShapeTest::test_kaiser
NumpyOneInputOpsDynamicShapeTest::test_view
NumpyOneInputOpsStaticShapeTest::test_isreal
NumpyOneInputOpsStaticShapeTest::test_view

hertschuh (commented):
Thanks for the contribution!

keras/src/backend/openvino/numpy.py
geeky33: OpenVINO opset15 doesn't have a bitcast/reinterpret_cast op. The only way I found to implement byte-level view is through NumPy. Should I raise NotImplementedError instead?"

PR #22228, opened 2026-02-20
Add numpy.signbit implementation for OpenVINO backend

Sujanian1304:
Add numpy.signbit implementation for OpenVINO backend

ref : https://github.com/openvinotoolkit/openvino/issues/34054


Sujanian1304:
Hi @rkazants @hertschuh ,

The PR for numpy.signbit is ready for review.
CI has passed, and tests were already included.

hertschuh:
> ### Critical: Unreachable Code
> The function still starts with a raise NotImplementedError. This will cause the function to exit immediately, and the logic you wrote below it will never execute (Dead Code). You need to remove the raise block at the start of the signbit function.

Just for the record, no it doesn't.

keras/src/backend/openvino/numpy.py
```diff
@@ -3073,9 +3073,18 @@ def sign(x):
...
-    raise NotImplementedError(
-        "`signbit` is not supported with openvino backend"
-    )
+    x = get_ov_output(x)
+    x_type = x.get_element_type()
+    zero = ov_opset.constant(0, dtype=x_type).output(0)
+    is_negative = ov_opset.less(x, zero).output(0)
+    if x_type.is_real():
+        one = ov_opset.constant(1.0, dtype=x_type).output(0)
+        recip = ov_opset.divide(one, x).output(0)
```
hertschuh, line 3408: Doing a `divide` seems line an expensive way of extracting the sign.

Can't you just use `ov_opset.sign`?
https://docs.openvino.ai/2025/api/ie_python_api/_autosummary/openvino.runtime.opset13.sign.html

hertschuh (commented):
Thank you for your contribution!

keras/src/backend/openvino/numpy.py
Sujanian1304: @hertschuh doesn't ov_opset.sign(-0.0) returns 0 (per IEEE 754, -0.0 == 0.0), so it can't distinguish -0.0 from +0.0. The divide trick exploits the fact that 1.0 / -0.0 = -inf to detect negative zero. The test at line 5361 specifically includes -0.0 in the input and expects True. Without the divide, that test would fail.

PR #22234, opened 2026-02-20
Fix Grain import to fix internal tests

abheesht17:
Fix Grain import to fix internal tests


amadhan882 (commented):
Hi @abheesht17, thanks for addressing the internal Grain import changes! The try-except fallback logic is the right approach here to maintain compatibility.

One small observation:
It looks like there is a duplicate import of SharedMemoryArrayMetadata right before the try block starts (around line 158 in the diff). Since that direct import isn't wrapped in a try-except, it might still trigger an ImportError before reaching your fallback logic. 

Suggesting to remove the direct import above the try block.

PR #22235, opened 2026-02-20
Fix plain callable rejection for constraint and regularizer

### LINKED ISSUE #22221: Passing a plain callable to add_weight(constraint=...) raises ValueError in Keras 3 (torch backend)
**Bug description**

When using Keras 3 with the PyTorch backend, passing a plain Python callable to the constraint argument of `Layer.add_weight()` raises a ValueError, stating that the constraint must be an instance of `keras.constraints.Constraint`.

Historically (and in many examples), constraints are simple callables of the form f(w) -> w'. However, with Keras 3.12.1 (torch backend), plain callables are rejected, even though a proper subclass of keras.constraints.Constraint works correctly.

This appears to be either:

- A regression in callable constraint support, or
- A backend-specific inconsistency (if other backends accept callables)
- A documentation mismatch if callables are no longer supported but examples imply otherwise.

A control test confirms that the constraint mechanism itself works when using a proper Constraint subclass.

**Reproduction script**

```
import os

import keras
from keras import ops
import torch

def masked_constraint(mask):
    # Plain callable constraint
    def _apply(w):
        return w * mask
    return _apply

class CustomLayer(keras.layers.Layer):
    def build(self, input_shape):
        mask = torch.ones((2, 2), dtype=torch.float32)
        c = masked_constraint(mask)

        # This line raises ValueError
        self.w = self.add_weight(
            name="w",
            shape=(2, 2),
            initializer="ones",
            trainable=True,
            constraint=c,  # <-- plain callable
        )

    def call(self, inputs):
        return inputs

x = ops.ones((1, 2), dtype="float32")
layer = CustomLayer()
_ = layer(x)

```

**Commands**
`KERAS_BACKEND=torch python keras_testcase.py
`

**Actual behavior**

The script raises a` ValueError:`
`Invalid value for attribute constraint. Expected an instance of keras.constraints.Constraint, or None. Received: constraint=<function ...>`

**Traceback (full)**
```
`ValueError`: Invalid value for attribute `constraint`. Expected an instance of `keras.constraints.Constraint`, or `None`. 
Received: constraint=<function main.<locals>.masked_constraint.<locals>._apply at 0x75e5ccdf13f0>

```

**Expected behavior**

One of the following:

- Plain callables should be accepted as valid constraints (if intended behavior),
OR
- The documentation should clearly state that only `keras.constraints.Constraint` subclasses are supported and plain callables are not allowed.

If this is an intentional API restriction in Keras 3, clarification in docs would help prevent confusion.

**Environment**

```
- Python: 3.10.19
- Keras: 3.12.1
- PyTorch: 2.10.0+cu128
- CUDA available: True
- Backend: torch
```

**GPU / Driver**

```
- GPU: NVIDIA RTX 3090 (x4 system)
- CUDA Runtime: 12.8 (from PyTorch build)
- torch.cuda.is_available(): True
```

timon0305:
## Summary
- Wrap plain callables in adapter classes (`_CallableConstraint` / `_CallableRegularizer`) in `constraints.get()` and `regularizers.get()` so they pass `isinstance` checks in `Variable` setters
- Update `Variable.constraint` and `Variable.regularizer` setters to also wrap callables via `get()` as a defensive fallback
- Adapters raise `NotImplementedError` on `get_config()` to clearly indicate serialization is unsupported for plain callables

## Test plan
- [x] Unit tests for `constraints.get()` with plain callable
- [x] Unit tests for `regularizers.get()` with plain callable
- [x] Unit tests for `Variable.constraint` and `Variable.regularizer` setters with plain callables
- [x] Non-callable values still raise `ValueError`
- [x] Integration test: `add_weight(constraint=<plain callable>)` works end-to-end
- [x] Existing constraint and regularizer tests still pass

Fixes #22221

### DIFF
--- a/keras/src/backend/common/variables.py
+++ b/keras/src/backend/common/variables.py
@@ -371,13 +371,17 @@ class Variable:
     @regularizer.setter
     def regularizer(self, value):
         from keras.src.regularizers import Regularizer
+        from keras.src.regularizers import get as get_regularizer
 
         if value is not None and not isinstance(value, Regularizer):
-            raise ValueError(
-                "Invalid value for attribute `regularizer`. Expected an "
-                "instance of `keras.regularizers.Regularizer`, or `None`. "
-                f"Received: regularizer={value}"
-            )
+            if callable(value):
+                value = get_regularizer(value)
+            else:
+                raise ValueError(
+                    "Invalid value for attribute `regularizer`. Expected an "
+                    "instance of `keras.regularizers.Regularizer`, a callable, "
+                    f"or `None`. Received: regularizer={value}"
+                )
         self._regularizer = value
 
     @property
@@ -387,13 +391,17 @@ class Variable:
     @constraint.setter
     def constraint(self, value):
         from keras.src.constraints import Constraint
+        from keras.src.constraints import get as get_constraint
 
         if value is not None and not isinstance(value, Constraint):
-            raise ValueError(
-                "Invalid value for attribute `constraint`. Expected an "
-                "instance of `keras.constraints.Constraint`, or `None`. "
-                f"Received: constraint={value}"
-            )
+            if callable(value):
+                value = get_constraint(value)
+            else:
+                raise ValueError(
+                    "Invalid value for attribute `constraint`. Expected an "
+                    "instance of `keras.constraints.Constraint`, a callable, "
+                    f"or `None`. Received: constraint={value}"
+                )
         self._constraint = value
 
     def __repr__(self):
--- a/keras/src/constraints/__init__.py
+++ b/keras/src/constraints/__init__.py
@@ -53,8 +53,27 @@ def get(identifier):
     if callable(obj):
         if inspect.isclass(obj):
             obj = obj()
+        if not isinstance(obj, Constraint):
+            obj = _CallableConstraint(obj)
         return obj
     else:
         raise ValueError(
             f"Could not interpret constraint identifier: {identifier}"
+        )
+
+
+class _CallableConstraint(Constraint):
+    """Wraps a plain callable as a `Constraint` instance."""
+
+    def __init__(self, fn):
+        self.fn = fn
+
+    def __call__(self, w):
+        return self.fn(w)
+
+    def get_config(self):
+        raise NotImplementedError(
+            "A plain callable used as a constraint cannot be serialized. "
+            "Please subclass `keras.constraints.Constraint` and implement "
+            "`get_config()` for serialization support."
         )
--- a/keras/src/constraints/constraints_test.py
+++ b/keras/src/constraints/constraints_test.py
@@ -99,3 +99,21 @@ class ConstraintsTest(testing.TestCase):
             "axis": 1,
         }
         self.assertEqual(config, expected_config)
+
+    def test_plain_callable_constraint(self):
+        def my_constraint(w):
+            return w * 0.5
+
+        obj = constraints.get(my_constraint)
+        self.assertIsInstance(obj, constraints.Constraint)
+        x = np.array([2.0, 4.0, 6.0])
+        output = obj(x)
+        self.assertAllClose(output, [1.0, 2.0, 3.0])
+
+    def test_plain_callable_constraint_get_config_raises(self):
+        def my_constraint(w):
+            return w
+
+        obj = constraints.get(my_constraint)
+        with self.assertRaises(NotImplementedError):
+            obj.get_config()
--- a/keras/src/regularizers/__init__.py
+++ b/keras/src/regularizers/__init__.py
@@ -53,8 +53,27 @@ def get(identifier):
     if callable(obj):
         if inspect.isclass(obj):
             obj = obj()
+        if not isinstance(obj, Regularizer):
+            obj = _CallableRegularizer(obj)
         return obj
     else:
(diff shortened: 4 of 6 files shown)


keras/src/backend/common/variables.py
```diff
@@ -371,13 +371,17 @@ def regularizer(self):
...
+        from keras.src.regularizers import get as get_regularizer
 
         if value is not None and not isinstance(value, Regularizer):
-            raise ValueError(
-                "Invalid value for attribute `regularizer`. Expected an "
-                "instance of `keras.regularizers.Regularizer`, or `None`. "
-                f"Received: regularizer={value}"
-            )
+            if callable(value):
+                value = get_regularizer(value)
```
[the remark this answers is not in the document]
timon0305: This is a false positive. The `regularizers.get()` function (`@keras_export("keras.regularizers.get")`) has always accepted arbitrary callables — this is by design in the public Keras API (see [upstream line 53](https://github.com/keras-team/keras/blob/master/keras/src/regularizers/__init__.py#L53)). Our change doesn't introduce callable acceptance; it wraps them in `_CallableRegularizer` so they satisfy `isinstance(value, Regularizer)` checks that the variable setter enforces. The `get_config()` method explicitly raises `NotImplementedError`, which actually *prevents* serialization of arbitrary callables.

keras/src/backend/common/variables.py
```diff
@@ -387,13 +391,17 @@ def constraint(self):
...
+        from keras.src.constraints import get as get_constraint
 
         if value is not None and not isinstance(value, Constraint):
-            raise ValueError(
-                "Invalid value for attribute `constraint`. Expected an "
-                "instance of `keras.constraints.Constraint`, or `None`. "
-                f"Received: constraint={value}"
-            )
+            if callable(value):
+                value = get_constraint(value)
```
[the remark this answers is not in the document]
timon0305: Same as above — false positive. `constraints.get()` has always accepted arbitrary callables by design. This PR wraps them in `_CallableConstraint` for type safety, with `get_config()` raising `NotImplementedError` to prevent serialization.

hertschuh:
@timon0305 

Is this ready for review? It's marked as a Draft right now.

keras/src/constraints/__init__.py
```diff
@@ -53,8 +53,27 @@ def get(identifier):
...
+
+    def __call__(self, w):
+        return self.fn(w)
+
+    def get_config(self):
+        raise NotImplementedError(
+            "A plain callable used as a constraint cannot be serialized. "
+            "Please subclass `keras.constraints.Constraint` and implement "
+            "`get_config()` for serialization support."
+        )
```
hertschuh, line 79: Why did you add this?
- For calling, the function itself works.
- It prevents function serialization. Arbitrary functions can be saved when decorated with `@register_keras_serializable`.

keras/src/constraints/constraints_test.py
```diff
@@ -99,3 +99,21 @@ def test_min_max_norm_get_config(self):
...
+        output = obj(x)
+        self.assertAllClose(output, [1.0, 2.0, 3.0])
+
+    def test_plain_callable_constraint_get_config_raises(self):
+        def my_constraint(w):
+            return w
+
+        obj = constraints.get(my_constraint)
+        with self.assertRaises(NotImplementedError):
+            obj.get_config()
```
hertschuh, line 119: You should actually test the opposite. Functions are serializable without a `get_config` with the `@register_keras_serializable` decorator.

Example: https://github.com/keras-team/keras/blob/master/keras/src/ops/core_test.py#L1695

You can test by using a `Dense` layer with a constraint and do a`get_config`, `from_config` on the layer to verify serialization.

keras/src/regularizers/regularizers_test.py
```diff
@@ -132,6 +132,21 @@ def test_orthogonal_regularizer_get_config(self):
...
+        obj = regularizers.get(my_regularizer)
+        self.assertIsInstance(obj, regularizers.Regularizer)
+
+    def test_plain_callable_regularizer_get_config_raises(self):
+        def my_regularizer(x):
+            return 0.0
+
+        obj = regularizers.get(my_regularizer)
+        with self.assertRaises(NotImplementedError):
+            obj.get_config()
```
hertschuh, line 148: You should actually test the opposite. Functions are serializable without a `get_config` with the `@register_keras_serializable` decorator.

Example: https://github.com/keras-team/keras/blob/master/keras/src/ops/core_test.py#L1695

You can test by using a `Dense` layer with a regularizer and do a`get_config`, `from_config` on the layer to verify serialization.

keras/src/regularizers/__init__.py
```diff
@@ -53,8 +53,27 @@ def get(identifier):
...
+
+    def __call__(self, x):
+        return self.fn(x)
+
+    def get_config(self):
+        raise NotImplementedError(
+            "A plain callable used as a regularizer cannot be serialized. "
+            "Please subclass `keras.regularizers.Regularizer` and implement "
+            "`get_config()` for serialization support."
+        )
```
hertschuh, line 79: Why did you add this?
- For calling, the function itself works.
- It prevents function serialization. Arbitrary functions can be saved when decorated with `@register_keras_serializable`.

PR #22236, opened 2026-02-20
Add numpy.nextafter implementation for OpenVINO backend

Sujanian1304:
Add numpy.nextafter implementation for OpenVINO backend

ref : https://github.com/openvinotoolkit/openvino/issues/34225


Sujanian1304:
Hi @rkazants @hertschuh ,

The PR for numpy.nextafter is ready for review.
CI has passed, and tests were already included.

keras/src/backend/openvino/excluded_concrete_tests.txt
hertschuh: Also remove `NumpyDtypeTest::test_nextafter`

Issue #21800, opened 2025-10-30
RuntimeError in `test_step` with multi-GPU: Model input shape mismatch


innat:
Say, 

- I've dataloader gives input shape `batch_size, 2 * model_input_shape`
- Build model `batch_size, model_input_shape`
- In order to run `model.evaluate(dataloader)`, we override `test_step`.

This works in single GPU. But doesn't on multi-GPU (possibly will fail in TPU as well). I've tested the code with keras-nightly.

```python
strategy = tf.distribute.MirroredStrategy()

class CustomModel(keras.Model):
    def __init__(self, model, split=1, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.model = model
        self.split = split

    def call(self, inputs, training=None):
        return self.model(inputs, training=None)
        
    def test_step(self, data):
        x, y = data
        x_list = ops.split(
            x, indices_or_sections=self.split, axis=-1
        )

        results = []
        for x_in_list in x_list:
            y_pred = self(x_in_list, training=False)
            results.append(y_pred)
        y_pred = ops.mean(
            ops.stack(results, axis=0), axis=0
        )

        loss = self.compute_loss(y=y, y_pred=y_pred)
        self.compiled_metrics.update_state(
            y, y_pred
        )
        return {m.name: m.result() for m in self.metrics}

inputs = keras.Input(shape=(32,))
outputs = keras.layers.Dense(1)(inputs)

with strategy.scope():
    model = keras.Model(inputs, outputs)
    custom_model = CustomModel(model, split=2)
    custom_model.compile(loss="mse", metrics=["mae"])

x = np.random.random((10, 64))
y = np.random.random((10, 1))
custom_model.evaluate(x, y)
```

```bash
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
/tmp/ipykernel_37/3196480519.py in <cell line: 0>()
     42 x = np.random.random((10, 64))
     43 y = np.random.random((10, 1))
---> 44 custom_model.evaluate(x, y)

/usr/local/lib/python3.11/dist-packages/keras/src/utils/traceback_utils.py in error_handler(*args, **kwargs)
    120             # To get the full stack trace, call:
    121             # `keras.config.disable_traceback_filtering()`
--> 122             raise e.with_traceback(filtered_tb) from None
    123         finally:
    124             del filtered_tb

[...]

PR #22237, opened 2026-02-20
fix : evaluate build fallback for custom test_step

darthvader58:
## Title
Fix symbolic `evaluate()` build fallback for distributed models with custom `test_step`

## Summary
This PR fixes a regression reported in #21800  where `model.evaluate()` can fail under distribution strategy when `test_step` customizes input structure before calling the model.

## Root cause
During the build, Keras always attempted to build from `backend.compute_output_spec(self, x, training=False)`.  
For certain custom `test_step` implementations, this direct `self(x)` path is invalid, while `test_step(data)` itself is valid.

## Changes
- Updated `Trainer._symbolic_build()` to:
  - First try the existing model-call symbolic build path.
  - If that fails, attempt a fallback symbolic build through `self.test_step(data_batch)` for 1-arg `test_step` signatures.
  - Preserve existing error behavior when both paths fail.
- Added a TensorFlow distributed regression test that reproduces the issue pattern:
  - Custom `test_step` transforms `x` (split into two tensors) and then delegates to `super().test_step(...)`.
  - Verifies `evaluate(..., return_dict=True)` succeeds under `tf.distribute.MirroredStrategy`.

## Files changed
- `keras/src/trainers/trainer.py`
- `keras/src/trainers/trainer_test.py`

## Testing
- Added targeted regression test:
  - `test_evaluate_distributed_with_custom_test_step_input_transform`

## Checklist
- [x] Bug fix scoped to issue behavior
- [x] Regression test added
- [x] No public API changes
- [x] Full test run in supported pytest keras

Requesting feedback from @dhantule @mehtamansi29 @gbaned


hertschuh (changes requested):
This is a very ad-hoc fix that assumes that `_symbolic_build` is called within `test_step`. But that's not always true.

darthvader58:
Imo, In most cases, it would be called within the test_step only. But I see your point here.

darthvader58 (commented):
looks good. much more concrete.

hertschuh:
> Imo, In most cases, it would be called within the test_step only. But I see your point here.

But if `model.fit` is called, it will build `test_step` instead of `train_step`.

PR #22239, opened 2026-02-20
Fix progress bar double-averaging stateful metrics during evaluate()

### LINKED ISSUE #21301: evaluate() shows incorrect metric values in progress bar vs returned results
When using model.evaluate(), the metric values displayed in the progress bar differ from the values returned by the method. There appears to be double averaging happening - the batch values are already averaged, but the progress bar shows an additional average of these averages.
Code to reproduce:

**Code to reproduce:**
```python
import tensorflow as tf
import numpy as np

# Modify the model to output the same as the input
model = tf.keras.Sequential([
    tf.keras.layers.Lambda(lambda x: x)  # Lambda layer to pass input directly to output
])

# Compile the model with MAE as the metric
model.compile(optimizer='adam', loss='mse', metrics=['mae'])

# Dummy data for evaluation
x = np.array([1, 2, 3, 4, 5, 6, 7, 8, 9, 10], dtype=float)
y = np.zeros_like(x)  # Dummy target values

results = model.evaluate(x, y, verbose=1, batch_size=1)
print("Evaluation results:", results)
```

**Output:**
```
10/10 ━━━━━━━━━━━━━━━━━━━━ 3s 4ms/step - loss: 17.8182 - mae: 3.4545
Evaluation results: [38.5, 5.5]
```

**Expected behavior:**
The metric values shown in the progress bar should match the final returned results (or at least be clearly documented if this difference is intentional).

**Issue:**

Progress bar shows loss: 17.8182, MAE: 3.4545

Returned values show loss: 38.5, MAE: 5.5

The correct values should be the returned ones (38.5 and 5.5 respectively), as these match manual calculations

The progress bar seems to be averaging already-averaged batch values

**Environment:**
TensorFlow 2.19 

**Additional notes:**
This discrepancy can be confusing for users who rely on the progress bar metrics during evaluation.

timon0305:
## Summary

All compiled Keras metrics are stateful — they maintain running totals internally and return already-aggregated values. The `ProgbarLogger` callback was not informing the `Progbar` widget of this, causing it to average the values a second time. This led to incorrect metric values being displayed in the progress bar during `evaluate()`, `fit()`, and `predict()`, while the returned dictionary contained the correct values.

- Dynamically mark every log key as a stateful metric in `ProgbarLogger._update_progbar()` and `_finalize_progbar()`, so the progress bar displays them as-is
- Remove a prior workaround in `Progbar.update()` that special-cased the `finalize=True` path for non-stateful metrics — this was masking the underlying issue
- Add unit tests for stateful / non-stateful metric display in the `Progbar` widget
- Add a regression test that verifies `evaluate()` progress-bar values match the returned metric dict

Fixes https://github.com/keras-team/keras/issues/21301

### DIFF
--- a/keras/src/callbacks/progbar_logger.py
+++ b/keras/src/callbacks/progbar_logger.py
@@ -91,6 +91,13 @@ class ProgbarLogger(Callback):
         self._maybe_init_progbar()
         self.seen = batch + 1  # One-indexed.
 
+        # All metrics from compiled Keras models are stateful: they maintain
+        # running totals internally and return already-aggregated values.
+        # Mark every log key as stateful so the progress bar displays them
+        # as-is instead of averaging them a second time.
+        if logs and self.progbar is not None:
+            self.progbar.stateful_metrics.update(logs.keys())
+
         if self.verbose == 1:
             self.progbar.update(self.seen, list(logs.items()), finalize=False)
 
@@ -99,4 +106,6 @@ class ProgbarLogger(Callback):
         if self.target is None:
             self.target = self.seen
             self.progbar.target = self.target
+        if logs and self.progbar is not None:
+            self.progbar.stateful_metrics.update(logs.keys())
         self.progbar.update(self.target, list(logs.items()), finalize=True)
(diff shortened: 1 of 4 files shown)


hertschuh:
Closing in favor of https://github.com/keras-team/keras/pull/23030

PR #22240, opened 2026-02-20
Accept integer-like types for layer dimension parameters

timon0305:
## Summary

`Dense(units=...)` and `RepeatVector(n=...)` use strict `isinstance(x, int)` checks that reject valid integer-like types such as `numpy.int32`, `numpy.int64`, and backend tensors returned by `ops.prod()`. This makes it impossible to write `Dense(ops.prod(input_shape[1:]))` without an explicit `int()` cast.

- Replace `isinstance(units, int)` with `operator.index()` in `Dense.__init__()`, which accepts any type implementing the `__index__` protocol (Python ints, numpy integers, 0-d backend tensors) while still rejecting floats, strings, and None
- Apply the same fix to `RepeatVector.__init__()`
- Add tests for numpy integer acceptance in both layers

Fixes https://github.com/keras-team/keras/issues/21655

### DIFF
--- a/keras/src/layers/core/dense.py
+++ b/keras/src/layers/core/dense.py
@@ -1,4 +1,5 @@
 import math
+import operator
 
 import ml_dtypes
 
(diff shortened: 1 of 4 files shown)


hertschuh:
I'm against this approach just in general:
- if we were to have this `try / except` block everywhere there's an int, this would bloat the code
- it's very easy to do `Dense(units=int(my_units))`
- we had a bug where an int-like made it into a shape and later was flowing through the operation, it resulted in a cryptic error message very far from the injection point. It's just much easier to gate ints from the get-go.

PR #22245, opened 2026-02-21
[OpenVINO Backend] Improve numpy.heaviside implementation 

goyaladitya05:
This PR improves the implementation for `numpy.heaviside` for the OpenVINO backend.

The heaviside implementation is currently failing tests due to lack of operand alignment (e.g. mixing integer and float dtypes).

### Changes

- Added the missing _align_operand_types call to ensure that both input tensor arguments (x1 and x2) are properly typed for OpenVINO operations.
- Added explicit type casting loops to ensure that integer types and booleans are properly cast to standard floatx() matching the behavior of NumPy heaviside
- int64 specifically is handled by mapping out into "float64".

Closes: openvinotoolkit/openvino/issues/34242


goyaladitya05:
I've rebased it on master. Ready for review.

PR #22249, opened 2026-02-21
[OpenVINO backend] Implement CrossEntropy loss operations for NumPy support

CityBoy-Claude:
This PR implements the categorical_crossentropy(), sparse_categorical_crossentropy(), and binary_crossentropy() operations for the OpenVINO backend and enables related tests. This may help unblock the models using cross-entropy loss.

Note: `test_logit_recovery_binary_crossentropy` is not enabled because OpenVINO does not support the test model evaluation

Closes https://github.com/openvinotoolkit/openvino/issues/34248

### DIFF
--- a/keras/src/backend/openvino/nn.py
+++ b/keras/src/backend/openvino/nn.py
@@ -677,11 +677,9 @@ def binary_crossentropy(target, output, from_logits=False):
     minus_log_prob = ov_opset.log(minus_output).output(0)
     result = ov_opset.multiply(target, log_prob).output(0)
     minus_result = ov_opset.multiply(minus_target, minus_log_prob).output(0)
-    result = ov_opset.add(result, minus_result).output(0)
-    loss = ov_opset.reduce_mean(result, -1).output(0)
-    loss = ov_opset.negative(loss).output(0)
-    loss = ov_opset.unsqueeze(loss, -1).output(0)
-    return OpenVINOKerasTensor(loss)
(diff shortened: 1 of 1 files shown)


CityBoy-Claude:
Hi @hertschuh This PR passes the CI tests. Would you like to have a look when you are available? Thanks in advance!

keras/src/backend/openvino/excluded_concrete_tests.txt
hertschuh, line 179: Can you also try to remove `NNOpsBehaviorTest::test_logit_recovery_binary_crossentropy`? It should pass now.

CityBoy-Claude: I've rebase but this test still fail:
1. It fails `self.assertAllClose(loss_value, 2.682124)`, nut sure if it is due to my implementation or underlying calculation errors
```
ACTUAL: array(2.681924, dtype=float32)
DESIRED: array(2.682124)
```
2. Even if I comment out this assert, it also fails with
```
keras/src/backend/openvino/trainer.py:239: NotImplementedError
```

Also mentioned this in PR description.

CityBoy-Claude: Hi @hertschuh , please ignore point 1 I mentioned above, there may be some setup issues on my macbook for the openvino backend 😢 , the precision is good on my Linux setup. 

However, the un-implemented error of point 2 still exists, so I am afraid this test cannot be enabled at this time.

PR #22250, opened 2026-02-21
Fix RandomZoom zero-scale validation and missing labels KeyError

### LINKED ISSUE #22247: [Bug] RandomZoom crashes with KeyError: 'labels' and lacks validation for zero-scale factor (-1.0)
### Summary
`keras.layers.RandomZoom` lacks sufficient validation when transforming bounding boxes and handling extreme factor ranges. 

1. **KeyError:** The `transform_bounding_boxes` method expects a `'labels'` key to be present in the bounding boxes dictionary, even when only coordinate transformation is intended.
2. **Zero-Scale Zoom:** The layer accepts a `height_factor` or `width_factor` of `-1.0`, which results in a zero-scale transformation matrix ($1.0 + (-1.0) = 0$), leading to potential singular matrix errors or NaN outputs.

This is a **usability and reliability bug** that violates the "Fail-Fast" principle of the Keras API.

### Reproduction Script
```python
import os
import keras
import numpy as np

print(f"Testing Keras version: {keras.__version__}")

# --- Case 1: Missing 'labels' Key ---
print("\n--- Testing RandomZoom with missing labels ---")
try:
    layer = keras.layers.RandomZoom(height_factor=0.2, width_factor=0.2, bounding_box_format="xyxy")
    img = np.ones((1, 100, 100, 3))
    # Dictionary with only 'boxes', missing 'labels'
    bboxes = {"boxes": np.array([[[0, 0, 10, 10]]])}
    
    transformation = layer.get_random_transformation(img)
    out_bbox = layer.transform_bounding_boxes(bboxes, transformation)
    print("[OK] Bounding boxes transformed successfully.")
except Exception as e:
    print(f"[ISSUE] RandomZoom crashed with KeyError: {e}")

# --- Case 2: Zero-Scale Zoom (-1.0) ---
print("\n--- Testing RandomZoom with factor -1.0 ---")
try:
    # Constructor accepts -1.0 which leads to 0.0 scale
    layer_zero = keras.layers.RandomZoom(height_factor=-1.0, width_factor=-1.0)
    print("[ISSUE] RandomZoom accepted -1.0 factor without validation error!")
except Exception as e:
    print(f"[OK] RandomZoom caught extreme factor.")
```

### Observed Result
```text
Testing Keras version: 3.10.0
--- Testing RandomZoom with missing labels ---
[ISSUE] RandomZoom crashed with KeyError: 'labels'

--- Testing RandomZoom with factor -1.0 ---
[ISSUE] RandomZoom accepted -1.0 factor without validation error!
```

### Actual Behavior
The `transform_bounding_boxes` method attempts to access the `'labels'` key from the input dictionary without checking for its existence, causing a crash. Additionally, the constructor allows a factor of `-1.0`, which creates a mathematically invalid zero-scale state during the affine transformation.

### Expected Behavior
The constructor and transformation methods should:
- Raise a `ValueError` if the factor is `-1.0` (zero scale).
- Allow `transform_bounding_boxes` to handle dictionaries containing only `"boxes"` by treating `"labels"` as optional.

### Root Cause
In `keras/src/layers/preprocessing/image_preprocessing/random_zoom.py`, the bounding box logic assumes a fixed dictionary structure. Furthermore, the `_check_factor_range` utility does not exclude the `-1.0` boundary which leads to non-invertible affine matrices.

### Suggested Fix

**For the KeyError:**
```python
if "labels" not in bounding_boxes:
    # Handle box-only transformation logic or provide default
```

**For the Factor Range:**
```python
if factor <= -1.0:
    raise ValueError("The `factor` argument for RandomZoom must be greater than -1.0.")
```

### Environment
* **OS:** Linux / Colab (Ubuntu)
* **Python:** 3.12.12
* **Keras:** 3.10.0
* **Backend:** TensorFlow 2.19.0

### Conclusion
Implementing strict input validation and flexible dictionary handling ensures that architecture inconsistencies are caught early, preventing silent failures and crashes in the augmentation pipeline.

rstar327:
## Summary

Fixes two bugs in `RandomZoom`:

1. **Zero-scale factor validation**: `_check_factor_range` now rejects `-1.0` (changed `<` to `<=`), which previously created a zero-scale transformation matrix (`1.0 + (-1.0) = 0`), leading to singular matrix errors or NaN outputs.

2. **Missing labels KeyError**: `clip_to_image_size` in `BoundingBox` now treats the `labels` key as optional using `.get()`, so bounding box dicts with only `boxes` no longer crash with `KeyError: 'labels'`.

Fixes #22247

### DIFF
--- a/keras/src/layers/preprocessing/image_preprocessing/bounding_boxes/bounding_box.py
+++ b/keras/src/layers/preprocessing/image_preprocessing/bounding_boxes/bounding_box.py
@@ -118,7 +118,8 @@ class BoundingBox:
             )
 
         ops = self.backend
-        boxes, labels = bounding_boxes["boxes"], bounding_boxes["labels"]
+        boxes = bounding_boxes["boxes"]
+        labels = bounding_boxes.get("labels", None)
         if width is not None:
             width = ops.cast(width, boxes.dtype)
         if height is not None:
@@ -132,9 +133,10 @@ class BoundingBox:
             y2 = ops.numpy.clip(y2, 0, height)
             boxes = ops.numpy.concatenate([x1, y1, x2, y2], axis=-1)
 
-            areas = self._compute_area(boxes)
-            areas = ops.numpy.squeeze(areas, axis=-1)
-            labels = ops.numpy.where(areas > 0, labels, -1)
+            if labels is not None:
+                areas = self._compute_area(boxes)
+                areas = ops.numpy.squeeze(areas, axis=-1)
+                labels = ops.numpy.where(areas > 0, labels, -1)
         elif bounding_box_format == "rel_xyxy":
             x1, y1, x2, y2 = ops.numpy.split(boxes, 4, axis=-1)
             x1 = ops.numpy.clip(x1, 0.0, 1.0)
@@ -143,13 +145,15 @@ class BoundingBox:
             y2 = ops.numpy.clip(y2, 0.0, 1.0)
             boxes = ops.numpy.concatenate([x1, y1, x2, y2], axis=-1)
 
-            areas = self._compute_area(boxes)
-            areas = ops.numpy.squeeze(areas, axis=-1)
-            labels = ops.numpy.where(areas > 0, labels, -1)
+            if labels is not None:
+                areas = self._compute_area(boxes)
+                areas = ops.numpy.squeeze(areas, axis=-1)
+                labels = ops.numpy.where(areas > 0, labels, -1)
 
         result = bounding_boxes.copy()
         result["boxes"] = boxes
-        result["labels"] = labels
+        if labels is not None:
+            result["labels"] = labels
         return result
 
     def affine(
--- a/keras/src/layers/preprocessing/image_preprocessing/random_zoom.py
+++ b/keras/src/layers/preprocessing/image_preprocessing/random_zoom.py
@@ -100,7 +100,7 @@ class RandomZoom(BaseImagePreprocessingLayer):
     _FACTOR_VALIDATION_ERROR = (
         "The `height_factor` and `width_factor` arguments "
         "should be a number (or a list of two numbers) "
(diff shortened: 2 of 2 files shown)


amadhan882 (commented):
### Technical Review by Issue Author

Hi @rstar327, thanks for the quick PR to address the issues I reported in #22247! 

I've reviewed your changes in `random_zoom.py` and `bounding_box.py`. While the core fixes are solid, I noticed some areas in `bounding_box.py` that could be optimized to prevent redundant operations and code duplication.

#### Technical Observations:

1. **Redundant Logic in `bounding_box.py`**:
In the updated `clip_to_image_size` logic, the area calculation and label filtering (mapping invalid boxes to -1) are repeated within both the `xyxy` and `rel_xyxy` blocks.
```python
if labels is not None:
    areas = self._compute_area(boxes)
    areas = ops.numpy.squeeze(areas, axis=-1)
    labels = ops.numpy.where(areas > 0, labels, -1)
```
**Suggestion**: Since this filtering logic is independent of the coordinate format (once clipped), it would be much cleaner to move it to a single point at the end of the method, right before creating the `result` dictionary. This avoids running the same ops twice in different branches.

2. **Factor Range Validation**:
The change in `random_zoom.py` to `input_number <= -1.0` correctly implements the "Fail-Fast" principle for the zero-scale scenario. This is a robust fix for the second part of the reported bug.

3. **KeyError Resolution**:
Using `bounding_boxes.get("labels", None)` and conditionally adding it back to the `result` dictionary effectively resolves the `KeyError` when labels are missing.

#### Suggested Action:
I recommend refactoring `bounding_box.py` to consolidate the label-filtering logic to avoid redundancy. Other than that, the fix correctly addresses the stability issues.

Looking forward to seeing this merged!

hertschuh (approved):
Thanks for the fix!

PR #22251, opened 2026-02-21
Fix TorchModuleWrapper deserialization when module config is a dict

### LINKED ISSUE #21350: Cannot properly save the model using TorchModuleWrapper and load the model
# environment
```python
pip install keras==3.7.0
```

# Save the model test
https://colab.research.google.com/drive/1b-fkujd3At131GGX4w5oj5XRWxbZwYXB?usp=sharing

Image

# Load the model test
```python
import keras
keras.models.load_model("/tmp/my_model/test.keras")
```

error info
```text
  File "/demo-python/test_load.py", line 3, in <module>
    keras.models.load_model("/demo-python/test_keras/my_model/test.keras")
  File "/demo-python/.venv/lib/python3.9/site-packages/keras/src/saving/saving_api.py", line 189, in load_model
    return saving_lib.load_model(
  File "/demo-python/.venv/lib/python3.9/site-packages/keras/src/saving/saving_lib.py", line 370, in load_model
    return _load_model_from_fileobj(
  File "/demo-python/.venv/lib/python3.9/site-packages/keras/src/saving/saving_lib.py", line 447, in _load_model_from_fileobj
    model = _model_from_config(
  File "/demo-python/.venv/lib/python3.9/site-packages/keras/src/saving/saving_lib.py", line 436, in _model_from_config
    model = deserialize_keras_object(
  File "/demo-python/.venv/lib/python3.9/site-packages/keras/src/saving/serialization_lib.py", line 720, in deserialize_keras_object
    raise TypeError(
TypeError: <class 'keras.src.models.functional.Functional'> could not be deserialized properly. Please ensure that components that are Python object instances (layers, models, etc.) returned by `get_config()` are explicitly deserialized in the model's `from_config()` method.

config={'build_config': {'input_shape': None}, 'class_name': 'Functional', 'config': {}, 'module': 'keras.src.models.functional', 'registered_name': 'Functional'}.

Exception encountered: <class 'keras.src.utils.torch_utils.TorchModuleWrapper'> could not be deserialized properly. Please ensure that components that are Python object instances (layers, models, etc.) returned by `get_config()` are explicitly deserialized in the model's `from_config()` method.

config={'class_name': 'TorchModuleWrapper', 'config': {'dtype': {'class_name': 'DTypePolicy', 'config': {'name': 'mixed_float16'}, 'module': 'keras', 'registered_name': None}, 'module': {'class_name': '__bytes__', 'config': {'value': 'xxx'}}, 'name': 'torch_module_wrapper', 'trainable': True}, 'inbound_nodes': [{'args': [{'class_name': '__keras_tensor__', 'config': {'dtype': 'float16', 'keras_history': ['cast', 0, 0], 'shape': [None, 1, 28, 28]}}], 'kwargs': {}}], 'module': 'keras.layers', 'name': 'torch_module_wrapper', 'registered_name': None}.

Exception encountered: a bytes-like object is required, not 'dict'
```

# suggest
https://github.com/keras-team/keras/blob/c7b66fc23fbeb7197d7c4a6491d7fda6d3502c6b/keras/src/utils/torch_utils.py#L166

```
if type(config["module"]) == str:
    buff = bytes.fromhex(config["module"].replace("\\x", ""))
elif type(config["module"]) == bytes:
    buff = io.BytesIO(config["module"])
```

rstar327:
## Summary

Fixes `from_config` in `TorchModuleWrapper` to handle the case where `config["module"]` is a dictionary instead of a base64 string.

When a model using `TorchModuleWrapper` is saved to `.keras` format and loaded back, the Keras serialization pipeline may convert the base64 string into a dict format (`{'class_name': '__bytes__', 'config': {'value': ...}}`). The previous code called `.encode("ascii")` directly on this dict, causing `a bytes-like object is required, not 'dict'`.

The fix:
- Detects the `__bytes__` dict wrapper and extracts the inner value
- Handles both `str` and `bytes` types for the module data
- Raises clear errors for unexpected formats

Fixes #21350

### DIFF
--- a/keras/src/utils/torch_utils_test.py
+++ b/keras/src/utils/torch_utils_test.py
@@ -248,7 +248,6 @@ class TorchUtilsTest(testing.TestCase):
         Since from_config receives the raw serialized inner config (not
         recursively deserialized), it must handle this dict format.
         """
-        import base64
 
         module = torch.nn.Sequential(torch.nn.Linear(2, 4))
         mw = TorchModuleWrapper(module)


hertschuh:
> When a model using `TorchModuleWrapper` is saved to `.keras` format and loaded back, the Keras serialization pipeline may convert the base64 string into a dict format (`{'class_name': '__bytes__', 'config': {'value': ...}}`). The previous code called `.encode("ascii")` directly on this dict, causing `a bytes-like object is required, not 'dict'`.

Under what circumstances would that happen?

This is the code from `get_config`:
```python
        buffer_b64 = base64.b64encode(buffer.getvalue()).decode("ascii")
        config = {
            "module": buffer_b64,
            "output_shape": self.output_shape,
        }
```

The `module` is not a dict.

Can you add a unit test demonstrating the issue?

keras/src/utils/torch_utils_test.py
```diff
@@ -238,6 +238,34 @@ def test_from_config(self):
...
+
+        # Simulate the legacy __bytes__ dict format that
+        # serialize_keras_object produces for bytes values
+        b64_str = config["module"]  # current format: base64 string
+        config["module"] = {
+            "class_name": "__bytes__",
+            "config": {"value": b64_str},
+        }
+
+        keras.config.enable_unsafe_deserialization()
```
hertschuh, line 264: Instead of this use:
```python
with serialization_lib.SafeModeScope(safe_mode=False):
```
Otherwise this can impact the other tests.

keras/src/utils/torch_utils_test.py
```diff
@@ -238,6 +238,34 @@ def test_from_config(self):
...
+        mw = TorchModuleWrapper(module)
+        config = mw.get_config()
+
+        # Simulate the legacy __bytes__ dict format that
+        # serialize_keras_object produces for bytes values
+        b64_str = config["module"]  # current format: base64 string
+        config["module"] = {
+            "class_name": "__bytes__",
+            "config": {"value": b64_str},
+        }
```
hertschuh, line 262: Wait, but this is not simulating what was going on in Keras 3.11 and earlier. The string was NOT base64 encoded, it was UTF8 decoded only, which was failing most of the time.

https://github.com/keras-team/keras/pull/21505 https://github.com/keras-team/keras/issues/19226

keras/src/utils/torch_utils.py
```diff
@@ -178,8 +178,30 @@ def from_config(cls, config):
...
+            if isinstance(module_data, str):
+                module_data = module_data.encode("ascii")
+
+            if not isinstance(module_data, bytes):
+                raise TypeError(
+                    "Expected 'module' to be a str or bytes, "
+                    f"got {type(module_data)}"
+                )
+
+            buffer_bytes = base64.b64decode(module_data)
```
hertschuh, line 203: Wait, but this is not handling what was going on in Keras 3.11 and earlier. The string was NOT base64 decoded, it was UTF8 decoded only, which was failing most of the time.

https://github.com/keras-team/keras/pull/21505 https://github.com/keras-team/keras/issues/19226

hertschuh:
@rstar327 

I looked at https://github.com/keras-team/keras/issues/21350 and it turns out it was already fixed by https://github.com/keras-team/keras/pull/21505 

The bug is not about cross-version compatibility, it is about the fact that with 3.11 and earlier, the module would almost never reload.

Sorry for the work you put into this.

PR #22256, opened 2026-02-22
Warn on mixed local/distributed variable shardings in JAX trainer

amitsrivastava78:
Add _check_sharding_consistency() helper to _get_state_sharding_spec() that detects when a model has a mix of SingleDeviceSharding (local) and mesh-aware variables. This happens when the model is built outside distribution.scope() and then compiled/trained inside it, leading to 'ValueError: Received incompatible devices for jitted computation'.
Also add parameterized tests covering DataParallel and ModelParallel for both the warning and no-warning cases.


keras/src/backend/jax/trainer.py
```diff
@@ -916,13 +916,87 @@ def _get_state_sharding_spec(self):
...
+        has_mesh = False
+        for v, s in var_shard_pairs:
+            if isinstance(s, jax.sharding.SingleDeviceSharding):
+                local_names.append(v.path)
+            else:
+                has_mesh = True
+            # Early exit: we know there is a mix as soon as we have
+            # seen at least one of each kind.
+            if local_names and has_mesh:
+                break
```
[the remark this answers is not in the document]
amitsrivastava78: Done

keras/src/backend/jax/trainer.py
```diff
@@ -916,13 +916,88 @@ def _get_state_sharding_spec(self):
...
+            "distributed (mesh-aware) variables. This will cause "
+            "a 'ValueError: Received incompatible devices for "
+            "jitted computation' when JAX tries to compile the "
+            "training step.\n\n"
+            f"First local variable found: {first_local_var_path!r}\n\n"
+            "This typically happens when the model is built or "
+            "weights are loaded outside `distribution.scope()`. "
+            "To fix this, create and build your model inside the "
+            "distribution scope:\n\n"
+            "    with distribution.scope():\n"
```
hertschuh, line 1001: Oh, honestly, I think it's easier to tell them to use `keras.distribution.set_distribution()` before creating any Keras object.

I'm not sure why we have scope since you have to put everything inside the scope for things to work properly.

Alternatively, we can mention both ways.

keras/src/backend/jax/trainer_sharding_fix_test.py
```diff
@@ -0,0 +1,147 @@
+"""Tests for mixed-sharding warning in JAX trainer.
+
+Verifies that _get_state_sharding_spec() emits a warning when it detects
+a mix of SingleDeviceSharding (local) and mesh-aware (NamedSharding)
+variables, and stays silent when shardings are consistent.
+
+Run with:
+    KERAS_BACKEND=jax python -m pytest <this_file> -v
+"""
```
hertschuh: Remove this, this is true for any test file.

keras/src/backend/jax/trainer_sharding_fix_test.py
```diff
@@ -0,0 +1,147 @@
...
+
```
hertschuh: About the name of this file, let's call it something else.

Test files should map exactly with the implementation file. The question the be able to answer is, is this test case already covered? The only clue you have on where to look is the file name.

So let's call it `trainer_test.py` in this folder (up until now there was no backend specific test).

keras/src/backend/jax/trainer_sharding_fix_test.py
```diff
@@ -0,0 +1,147 @@
...
+            model._get_state_sharding_spec()
+
+        mixed = [w for w in caught if "mix of local" in str(w.message)]
+        assert len(mixed) > 0, (
+            f"[{dist_name}] Expected a mixed-sharding warning "
+            f"but none was raised"
+        )
+        msg = str(mixed[0].message)
+        assert "SingleDeviceSharding" in msg
+        assert "distribution.scope()" in msg
```
hertschuh: Use `self.assertXXX` instead.

keras/src/backend/jax/trainer_sharding_fix_test.py
```diff
@@ -0,0 +1,147 @@
...
+)
+def test_warns_when_model_built_outside_scope(dist_factory, dist_name):
+    """Model built outside scope + compiled inside -> mixed warning."""
+    dist = dist_factory()
+    model = _build_model_outside_scope()
+
+    # All model weights should be local (SingleDeviceSharding)
+    for w in model.weights:
+        assert isinstance(
+            w.value.sharding, jax.sharding.SingleDeviceSharding
```
hertschuh: self.assertIsInstance

keras/src/backend/jax/trainer_sharding_fix_test.py
```diff
@@ -0,0 +1,147 @@
...
+from keras.src.backend import distribution_lib as backend_dlib
+from keras.src.distribution import distribution_lib
+
+if backend.backend() == "jax":
+    os.environ.setdefault("JAX_PLATFORMS", "cpu")
+    xla_flags = os.getenv("XLA_FLAGS") or ""
+    if "xla_force_host_platform_device_count" not in xla_flags:
+        os.environ["XLA_FLAGS"] = (
+            f"{xla_flags} --xla_force_host_platform_device_count=8"
+        )
```
hertschuh: The issue is that this will never be run as part of the current test suites because it's too late to create fake CPUs.

keras/src/backend/jax/trainer_sharding_fix_test.py
```diff
@@ -0,0 +1,147 @@
...
+
+_skip = pytest.mark.skipif(
+    backend.backend() != "jax" or len(jax.devices()) != 8,
+    reason="JAX backend with 8 simulated CPU devices required",
+)
```
hertschuh: Let not do that, it could not figure out when the tests are skipped when reading the code below.

keras/src/backend/jax/trainer_sharding_fix_test.py
```diff
@@ -0,0 +1,147 @@
...
+    return models.Model(inputs=inputs, outputs=y)
+
+
+def _build_model_inside_scope(dist):
+    """Build the same model entirely inside the distribution scope."""
+    with dist.scope():
+        inputs = layers.Input(shape=[16])
+        y = layers.Dense(units=32, activation="relu")(inputs)
+        y = layers.Dense(units=10, activation="softmax")(y)
+        return models.Model(inputs=inputs, outputs=y)
```
hertschuh: This actually hurts readability in the sense that the key parts of the code that trigger the result you expect are not in the test code directly. Unit tests are best when self contained and you don't have to jump up and down the file to understand what's going on.

On the other hand, you can minimize the code to not need helpers. For instance, a one-liner model:
```python
model = models.Sequential([layers.Dense(10)])
```

keras/src/backend/jax/trainer_sharding_fix_test.py
```diff
@@ -0,0 +1,147 @@
...
+    backend.backend() != "jax" or len(jax.devices()) != 8,
+    reason="JAX backend with 8 simulated CPU devices required",
+)
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+def _make_data_parallel():
+    return distribution_lib.DataParallel(devices=backend_dlib.list_devices())
```
hertschuh: You don't even need to pass the devices, it does it for you.

keras/src/backend/jax/trainer.py
amitsrivastava78: Done

keras/src/backend/jax/trainer_sharding_fix_test.py
amitsrivastava78: Done

keras/src/backend/jax/trainer_sharding_fix_test.py
amitsrivastava78: Done

keras/src/backend/jax/trainer_sharding_fix_test.py
amitsrivastava78: Done

keras/src/backend/jax/trainer_sharding_fix_test.py
amitsrivastava78: Done as part of above

keras/src/backend/jax/trainer_sharding_fix_test.py
amitsrivastava78: Done

amitsrivastava78: removed this simulation, and executing the test case in case there is 2 or more devices on jax backend

keras/src/backend/jax/trainer_sharding_fix_test.py
amitsrivastava78: Done

keras/src/backend/jax/trainer_sharding_fix_test.py
amitsrivastava78: Done

keras/src/backend/jax/trainer_sharding_fix_test.py
amitsrivastava78: Done

keras/src/backend/jax/trainer_test.py
```diff
@@ -0,0 +1,110 @@
...
+    # Mixed-sharding warning tests
+    # ----------------------------------------------------------------
+    @parameterized.named_parameters(
+        {"testcase_name": "DataParallel", "dist_type": "data_parallel"},
+        {"testcase_name": "ModelParallel", "dist_type": "model_parallel"},
+    )
+    def test_warns_when_model_built_outside_scope(self, dist_type):
+        """Model built outside distribution -> mixed warning on compile."""
+        self._skip_if_not_distributed()
+        import jax
```
hertschuh, line 44: Move to the top, this is a jax specific test, you can always import.

keras/src/backend/jax/trainer_test.py
```diff
@@ -0,0 +1,110 @@
...
+from keras.src import models
+from keras.src import testing
+from keras.src.backend import distribution_lib as backend_dlib
+from keras.src.distribution import distribution_lib
+
+
+class JAXTrainerTest(testing.TestCase, parameterized.TestCase):
+    def _skip_if_not_distributed(self):
+        if backend.backend() != "jax":
+            self.skipTest("Requires JAX backend")
```
hertschuh, line 17: nitpick: instead of this decorate the class with 
```python
@pytest.mark.skipif(backend.backend() != "jax", reason="JAX only)
```
for consistency and it's easier to search for jax only tests in the code like this.

But keep the rest here.

hertschuh (approved):

PR #22257, opened 2026-02-22
Add geomspace to keras.ops.numpy

MarcosAsh:
## Summary
Add numpy.geomspace to keras.ops, returning numbers spaced evenly on a log scale (geometric progression). This is the companion to logspace where endpoints are specified directly rather than as logarithms.
                                                                                                                                                
  Implemented across all backends (JAX, NumPy, TensorFlow, Torch, OpenVINO). TF/Torch/OV compose from existing logspace + log10 ops JAX and NumPy pass through to their native geomspace.

  Closes #20519

  ## Test plan

  - Dynamic shape tests
  - Static shape tests
  - Correctness tests (scalar and array inputs, with/without endpoint)
  - Dtype tests (parameterized across float dtypes and num values)

### DIFF
--- a/keras/src/ops/numpy.py
+++ b/keras/src/ops/numpy.py
@@ -3687,6 +3687,80 @@ def get_item(x, key):
     return x[key]
 
 
+class Geomspace(Operation):
+    def __init__(self, num=50, endpoint=True, dtype=None, axis=0, *, name=None):
+        super().__init__(name=name)
+        self.num = num
+        self.endpoint = endpoint
+        self.dtype = dtype
+        self.axis = axis
+
+    def call(self, start, stop):
+        return backend.numpy.geomspace(
+            start,
+            stop,
+            num=self.num,
+            endpoint=self.endpoint,
+            dtype=self.dtype,
+            axis=self.axis,
+        )
+
+    def compute_output_spec(self, start, stop):
+        start_shape = getattr(start, "shape", [])
+        stop_shape = getattr(stop, "shape", [])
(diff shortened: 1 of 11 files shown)


keras/src/ops/numpy.py
```diff
@@ -3687,6 +3687,82 @@ def get_item(x, key):
...
+            axis=self.axis,
+        )
+
+    def compute_output_spec(self, start, stop):
+        start_shape = getattr(start, "shape", [])
+        stop_shape = getattr(stop, "shape", [])
+        output_shape = broadcast_shapes(start_shape, stop_shape)
+        axis = canonicalize_axis(self.axis, len(output_shape) + 1)
+        output_shape = list(output_shape)
+        output_shape.insert(axis, self.num)
```
hertschuh, line 3714: This is more correct and shorter than what was done in `Logspace` and `Linspace`. Can you fix those (in a separate PR)?

Thanks!

hertschuh:
Can you run the code formatter?

MarcosAsh:
Hey just ran the code formatter again ill do another pr for this asap! Thanks for the review

hertschuh:
The code format check is still failing. Do you know why?

MarcosAsh:
I think the problem was ruff 0.9.2 wanted different line wrapping than my local ruff on a few lines in numpy.py and numpy_test.py.

MarcosAsh:
yeah it passed now so I think it was that ill do the same for the gru pr

PR #22265, opened 2026-02-23
Made layers.add_weight() args keyword-only with clear error handling

### LINKED ISSUE #22253: Layer.add_weight() misinterprets positional first argument as shape, raising "got multiple values for argument shape".
**Bug Description**

In Keras 3 with the TensorFlow backend, calling `Layer.add_weight()` with a positional first argument intended as `name `and also passing `shape=` causes a `TypeError`:
`Layer.add_weight() got multiple values for argument 'shape'`

This strongly suggests the Keras 3 `add_weight` signature treats the first positional argument as shape (not name), so a legacy call pattern like:
`self.add_weight("matrix", shape=(...), ...)`
collides with the explicit `shape=` kwarg.

This breaks older / common Keras code patterns (or at least produces a confusing error message). If positional `name `is no longer supported, Keras should raise a clearer error (e.g., “`name `must be passed as a keyword argument”) or document the breaking change prominently.

**Minimal reproduction script**

```
import os

import numpy as np
import tensorflow as tf
import keras


class MatrixLayer(keras.layers.Layer):
    def build(self, input_shape):
        num_fs = int(input_shape[-1])
        # BUG TRIGGER: positional first arg intended as name
        self.matrix = self.add_weight(
            name="matrix",
            shape=(num_fs, 4),
            initializer="glorot_uniform",
            trainable=True,
        )

    def call(self, inputs):
        return keras.ops.matmul(inputs, self.matrix)


inp = keras.Input(shape=(5, 9), name="inputs")
x = inp[:, :, :-1]
y = MatrixLayer()(x)
model = keras.Model(inp, y)

data = np.random.RandomState(2021).randn(2, 5, 9).astype("float32")
model(data, training=False)
```

**Triggering commands**
**GPU (single GPU):**
`CUDA_VISIBLE_DEVICES=0 KERAS_BACKEND=tensorflow python repro_add_weight_positional_name.py
`

**CPU-only:**
`CUDA_VISIBLE_DEVICES="" KERAS_BACKEND=tensorflow python repro_add_weight_positional_name.py`

**Actual behavior**
The script raises:
- `TypeError: Layer.add_weight() got multiple values for argument 'shape'`

**Expected behavior**

1. Backwards-compatible behavior: positional "matrix" is treated as name and the layer builds successfully; model call runs.
OR
2. If positional name is intentionally unsupported in Keras 3: raise a clear, explicit error, e.g.   “name must be passed as a keyword argument `(name="...")”` not a confusing Python signature collision error about `shape`.

**Traceback (full)**

Right now your GCFL harness prints only the exception message (not a stack trace), so you need to capture the real traceback by running the minimal repro above (no try/except).
`CUDA_VISIBLE_DEVICES=0 KERAS_BACKEND=tensorflow python repro_add_weight_positional_name.py
`

**Environment**

**Software**
```
- OS: Linux (add uname -a output in report)
- Python: 3.10.19
- Conda: 25.7.0
- Keras: 3.4.1
- TensorFlow: 2.17.0 (tensorflow[and-cuda]==2.17.0)
- TF build info (from tf.sysconfig.get_build_info()):
- CUDA runtime: 12.3
- cuDNN: 8
- CUDA compute caps include: sm_80 etc.
```

**Hardware**
```
- GPUs: 4× NVIDIA GeForce RTX 3090
- Driver: 550.78
- nvidia-smi: CUDA Version 12.4
```

sam-shubham:
Fixes #22253

Related #18467 

In Keras 3, `Layer.add_weight()` has `shape` as the first parameter. In Keras 2, it was `name`. Legacy code like:

```python
self.add_weight("matrix", shape=(num_fs, 4), initializer="glorot_uniform")
```

raises a confusing:

```
TypeError: Layer.add_weight() got multiple values for argument 'shape'
```

This happens because Python binds `"matrix"` to `shape` positionally, and then `shape=(num_fs, 4)` collides with it as a keyword argument.

Even a partial guard (`isinstance(shape, str)`), only caught the case where a string was the **only** positional argument. When both a positional string **and** a `shape=` keyword argument were passed, Python raised its own `TypeError` *before* the method body executed, so the guard was never reached.

---

## Solution

Changed the `add_weight` signature from:

```
add_weight(self, shape=None, initializer=None, dtype=None, ...)
```

to:

```
add_weight(self, *args, shape=None, initializer=None, dtype=None, ...)
```

This allows intercepting positional arguments early and providing actionable error messages.

New behavior:

- **String as first positional argument** →  
  `ValueError: name must be passed as a keyword argument. Received: add_weight('matrix', ...). Use: add_weight(shape=..., name='matrix').`

- **Non-string positional arguments** → accepted as `shape`, `initializer`, `dtype` (in order), preserving backward compatibility with existing Keras 3 usage like `add_weight((4, 5), "zeros", "float32")`

- **Positional + keyword conflict for the same parameter** → clear `ValueError` explaining the duplicate

- **More than 3 positional arguments** → `TypeError` with a clear message

---

## Backward Compatibility

This change is **fully backward compatible**:

- All existing internal callers already use keyword arguments
- `add_weight(shape=(3, 4), name="w")` → works as before
- `add_weight((3, 4))` → still works (positional shape accepted)
- `add_weight((4, 5), "zeros", "float32")` → works (positional shape, initializer, dtype)
- `add_weight()` → still works (defaults to scalar)

The only code this affects is code that was **already broken**. It replaces Python’s confusing `TypeError` with a clear, actionable `ValueError`.

### DIFF
--- a/keras/src/layers/layer.py
+++ b/keras/src/layers/layer.py
@@ -565,15 +565,18 @@ class Layer(BackendLayer, Operation):
         """
         self._check_super_called()
         if args:
+            # `args` is only kept to detect the legacy Keras 2 call style
+            # (`add_weight(shape, initializer, dtype, ...)`) and raise a clear
+            # error for positional `name`.
             if len(args) > 3:
                 raise TypeError(
-                    f"add_weight() takes at most 3 positional arguments "
+                    "add_weight() takes at most 3 positional arguments "
                     f"but {len(args)} were given."
                 )
             shape_arg = args[0]
             if isinstance(shape_arg, str):
                 raise ValueError(
-                    f"`name` must be passed as a keyword argument. "
+                    "`name` must be passed as a keyword argument. "
                     f"Received: add_weight('{shape_arg}', ...). "
                     f"Use: add_weight(shape=..., name='{shape_arg}')."
                 )


sam-shubham:
@fchollet

Would you mind taking a look? Open to any suggestions or improvements.

keras/src/layers/layer.py
```diff
@@ -563,6 +564,26 @@ def add_weight(
             name: String name of the variable. Useful for debugging purposes.
         """
         self._check_super_called()
+        if args:
+            if len(args) > 1:
+                raise TypeError(
+                    f"add_weight() takes at most 1 positional argument "
+                    f"but {len(args)} were given."
+                )
+            arg = args[0]
```
hertschuh: The issue is that it will break people who are using `add_weight` with positional arguments with Keras 3:
```python
self.add_weight((4, 5), "zeros", "float32")
```
Not breaking Keras 3 users is the top priority.

sam-shubham:
@hertschuh Thanks for review..

I've updated the implementation to accept up to 3 positional arguments (`shape`, `initializer`, `dtype`), matching the original signature's positional parameter order. So 
```python
self.add_weight((4, 5), "zeros", "float32")
```
continues to work as before.

The string-first detection now only fires on `args[0]`, and each positional arg has its own duplicate-check if the same value was also passed as a keyword argument.

Could you take another look?

keras/src/layers/layer.py
```diff
@@ -563,6 +564,39 @@ def add_weight(
             name: String name of the variable. Useful for debugging purposes.
         """
         self._check_super_called()
+        if args:
+            if len(args) > 3:
+                raise TypeError(
+                    f"add_weight() takes at most 3 positional arguments "
```
hertschuh: nitpick: remove `f` line 570, it is not needed.

keras/src/layers/layer.py
```diff
@@ -563,6 +564,39 @@ def add_weight(
...
+                    f"but {len(args)} were given."
+                )
+            shape_arg = args[0]
+            if isinstance(shape_arg, str):
+                raise ValueError(
+                    f"`name` must be passed as a keyword argument. "
```
hertschuh: nitpick: remove `f` line 576, it is not needed.

keras/src/layers/layer.py
```diff
@@ -563,6 +564,39 @@ def add_weight(
             name: String name of the variable. Useful for debugging purposes.
         """
         self._check_super_called()
+        if args:
```
hertschuh, line 567: This code would be confusing to someone reading it for the first time.

Can you add a comment explaining that `args` is added purely to detect the Keras 2 call style for `add_weight` and provide a clear error message.

sam-shubham:
@hertschuh removed those `f` and added comments as requested.. can you review

PR #22270, opened 2026-02-24
[OpenVINO backend] Implement unfold() layer for NumPy support

CityBoy-Claude:
This PR implements the unfold() layer for the OpenVINO backend and enables related tests.

Closes https://github.com/openvinotoolkit/openvino/issues/34275

### DIFF
--- a/keras/src/backend/openvino/nn.py
+++ b/keras/src/backend/openvino/nn.py
@@ -1,6 +1,7 @@
 import openvino.opset15 as ov_opset
 from openvino import Type
 
+import keras.src.backend.openvino.numpy as onp
 from keras.src import backend
 from keras.src.backend.openvino.core import OPENVINO_DTYPES
 from keras.src.backend.openvino.core import OpenVINOKerasTensor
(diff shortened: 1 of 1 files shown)


CityBoy-Claude:
Hi @hertschuh All tests pass. Would you please take a look when you have the time? Thanks!

keras/src/backend/openvino/nn.py
```diff
@@ -846,7 +846,50 @@ def dot_product_attention(
...
+    Args:
+        input: 4-D tensor, shape (N, C, H, W)  **required**.
+        kernel_size: int or (kH, kW)
+        dilation: int or (dH, dW), default 1
+        padding: int or (pH, pW), default 0
+        stride: int or (sH, sW), default 1
+
+    Returns:
+        3-D tensor, shape (N, C*kH*kW, L)
+    """
```
hertschuh: Remove docstring, only this docstring matter:
https://github.com/keras-team/keras/blob/master/keras/src/ops/nn.py#L3266

keras/src/backend/openvino/nn.py
```diff
@@ -846,7 +846,50 @@ def dot_product_attention(
...
+    import keras.src.backend.openvino.numpy as onp
```
hertschuh: Move this import `import keras.src.backend.openvino.numpy as onp` to the top.

CityBoy-Claude:
@hertschuh  Thanks for your review! Would you please take another look?

hertschuh (commented):
Thanks!

PR #22273, opened 2026-02-24
Add regression test for Sequential sublayer in custom `train_step`

### LINKED ISSUE #18459: Passing inputs to a sequential model in a custom training step called with fit to calculate loss fails - requires a layer redefinition
From source https://keras.io/examples/vision/shiftvit/

Training step incorporates a step with custom loss calculation. Relevant parts:

```
def train_step(self, inputs):
    with tf.GradientTape() as tape:
        total_loss, labels, logits = self._calculate_loss(
            data=inputs, training=True
        )
```

```
def _calculate_loss(self, data, training=False):
    (images, labels) = data

    # Augment the images
    augmented_images = self.data_augmentation(images, training=training)
...

self.data_augmentation = keras.Sequential(
        [
            layers.Resizing(config.input_shape[0] + 20, config.input_shape[0] + 20),
            layers.RandomCrop(config.image_size, config.image_size),
            layers.RandomFlip("horizontal"),
            layers.Rescaling(1 / 255.0),
        ]
    )
```
(sequential model is passed as a parameter to the function but that should make no difference)

When run with fit, it receives a KerasTensor despite it being passed a tensorflow tensor.
```
    augmented_images = self.data_augmentation(images, training=training)
E   ValueError: A KerasTensor cannot be used as input to a TensorFlow function. A KerasTensor is a symbolic placeholder for a shape and dtype, used when constructing Keras Functional models or Keras Functions. You can only use it as input to a Keras layer or a Keras operation (from the namespaces `keras_core.layers` and `keras_core.operations`). 
```
Workaround unclear.

DeborahOlaboye:
## Description                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                                                                                  
  Calling a Sequential model from a custom `train_step` during `model.fit()` previously caused:                                                        
                                                                                                                                                       
  ValueError: A KerasTensor cannot be used as input to a TensorFlow function.                                                                          

  This happened because `_symbolic_build` was tracing `train_step` with `KerasTensor` inputs.
  When `train_step` passed those tensors to a Sequential sublayer (e.g., a data augmentation
  pipeline), TF's internal conversion (`__tf_tensor__`) raised the error.

  ## Changes

  Adds a regression test to `trainer_test.py` to prevent this bug from silently returning
  in future refactors:

  - `SequentialSublayerInTrainStepModel` — a model that holds a `Sequential` sublayer
    (`Rescaling`) called **only** inside `train_step`, not in `call()`, mirroring the
    exact pattern from the issue.
  - `test_sequential_sublayer_in_custom_train_step` — runs `model.fit()` and asserts
    no KerasTensor-related error is raised.

  ## Testing

  ```bash
  KERAS_BACKEND=torch pytest keras/src/trainers/trainer_test.py \
    -k test_sequential_sublayer_in_custom_train_step -xvs

  Closes #18459

### DIFF
--- a/keras/src/backend/tensorflow/trainer.py
+++ b/keras/src/backend/tensorflow/trainer.py
@@ -715,6 +715,15 @@ class TensorFlowTrainer(base_trainer.Trainer):
             # When no distribution strategy is set, defer building
             # to when the train/test/predict function gets traced.
             # This maximizes backwards compatibility.
+            # Exception: if the model has unbuilt sublayers that won't be
+            # reached through `call()` (e.g. a Sequential sublayer only used
+            # in a custom `train_step`), we must pre-build them here.
+            # Otherwise `Sequential.build()` runs lazily inside `tf.function`
+            # tracing, where it creates a nested `FuncGraph` that is
+            # incompatible with the active `tf.function` context (gh-18459).
+            if all(layer.built for layer in self._flatten_layers()):
+                return
+            self._symbolic_build(iterator=iterator, data_batch=data_batch)
             return
 
         # Unlike jax/torch iterator, tf iterator returns an iterator instead
--- a/keras/src/trainers/trainer.py
+++ b/keras/src/trainers/trainer.py
@@ -1130,6 +1130,19 @@ class Trainer:
                     "Exception encountered:\n"
                     f"'{e}'"
                 )
+            # Pre-build any sublayers that were not reached through `call()`
+            # (e.g. a Sequential sublayer only used in a custom `train_step`).
+            # Without this, such layers would be built lazily during the first
+            # `tf.function` trace of `train_step`, where `Sequential.build()`
+            # creates a nested `FuncGraph` inside an active `tf.function`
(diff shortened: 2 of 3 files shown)


hertschuh (commented):
@DeborahOlaboye 

I'm confused about this, it doesn't include a bug fix.

To be clear, I don't think the bug is just the result of using `Sequential` in a sublayer with a custom `train_step`. The root cause of the bug still needs to be determined.

DeborahOlaboye:
> @DeborahOlaboye
> 
> I'm confused about this, it doesn't include a bug fix.
> 
> To be clear, I don't think the bug is just the result of using `Sequential` in a sublayer with a custom `train_step`. The root cause of the bug still needs to be determined.

@hertschuh, thanks for the review. You're right, the previous commit only added a regression test without a bug fix. I've now identified the root cause and added a fix.

hertschuh (changes requested):
The unit test already passes without the code changes to the trainer files and classes.

I don't understand what those code changes are trying to do, they contradict what the code does right before.

PR #22276, opened 2026-02-24
Fix clone_model recursive=True for multi-level nested models

### LINKED ISSUE #21120: clone_model(recursive=True) supports one level of nesting only
repro Colab (14 lines of code): https://colab.research.google.com/drive/1L7FbGwyNsKDHHAcKxrdAS0J6vepOyEBz?usp=sharing

If the model contains 2 or more levels of nested layers, they will not be entered by clone_model(recursive=True)

rstar327:
## Summary

Fixes `clone_model(recursive=True)` to properly recurse through all nesting levels.

Previously, the `_wrap_clone_function` helper called `clone_model()` for nested Sequential/Functional models but did not pass `recursive=True`, so only the first level of nesting was entered. Models with 2+ levels of nested layers would have their inner layers shared instead of cloned.

The fix adds `recursive=True` to both nested `clone_model()` call sites (Sequential and Functional).

Fixes #21120

### DIFF
--- a/keras/src/models/cloning_test.py
+++ b/keras/src/models/cloning_test.py
@@ -288,9 +288,7 @@ class CloneModelTest(testing.TestCase):
 
         # Every Dense in the deepest level should have been cloned
         # (i.e. clone_function was applied), not shared.
-        for l1, l2 in zip(
-            model._flatten_layers(), new_model._flatten_layers()
-        ):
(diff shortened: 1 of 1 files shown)


rstar327:
The existing `test_recursive` only has 1 level of nesting, so the recursive `clone_model()` call never encounters another nested model, the missing `recursive=True` doesn't matter when there's nothing further to recurse into. The bug only shows with 2+nesting levels.

Issue #21801, opened 2025-10-30
`plot_model` `rankdir="LR"` does not show U-shape plot for a UNet


khteh:
https://github.com/tensorflow/tensorflow/issues/102880
### Issue type

Bug

### Have you reproduced the bug with TensorFlow Nightly?

No

### Source

binary

### TensorFlow version

2.20.0

### Custom code

Yes

### OS platform and distribution

Ubuntu 25.04

### Python version

3.13.3

### Current behavior?

Plotting a UNet vertically works perfectly but the image is too long to fit in a canocical wide-screen monitor. Plotting it horizontally makes more sense and is easier to view the correspoding skip connections. However, it doesn't work at all. The plot (saved) is empty.

### Standalone code to reproduce the issue

```shell
plot_model(
    self._model,
    to_file="output/UNet.png",
    show_shapes=True,
    show_dtype=True,
    show_layer_names=True,
    rankdir="LR",
    expand_nested=True,
    show_layer_activations=True)
```

dhantule:
Hi @khteh, 

I tested `plot_model` function with `rankdir="LR" ` (Left to Right) in this colab [gist](https://colab.sandbox.google.com/gist/dhantule/642f8211cdd25b20f837b01ece1947f7/plot_model-rankdir-lr-show-empty-image-21801.ipynb#scrollTo=eG7Gmt2cXf9Q) and I'm able to get the plot horizontally, 
check if your model is built before plotting, provide a minimal reproducible colab if you're still facing issues. 

<img width="13486" height="517" alt="Image" src="https://github.com/user-attachments/assets/0c697c05-3252-4c69-a17c-60f79be52dc3" />

khteh:
Take a look at this plot. How easy is it to view it? And how "U" is it really UNet?
<img width="32767" height="302" alt="Image" src="https://github.com/user-attachments/assets/46063ac6-374a-4c4a-b4d4-7ae226849683" />

dhantule:
@khteh, try increasing the `dpi ` for more clarity

<img width="2824" height="1406" alt="Image" src="https://github.com/user-attachments/assets/d2defec6-d5b6-4b0b-8758-93255c3e03a7" />

khteh:
I can try that but how about the U-shape?

khteh:
I don't think the `dpi` works well. Enclosed is the plot with `dpi=1000`. How far should it go to have a decent plot?

[...]

PR #22277, opened 2026-02-24
Add splines argument to plot_model and model_to_dot

rstar327:
## Summary

Add a `splines` argument to `plot_model` and `model_to_dot`, allowing users to control edge routing style (e.g. `"ortho"`, `"curved"`, `"polyline"`, `"spline"`, `"line"`). Defaults to `"ortho"` (no behavior change).

### DIFF
--- a/keras/src/utils/model_visualization.py
+++ b/keras/src/utils/model_visualization.py
@@ -203,6 +203,7 @@ def model_to_dot(
     subgraph=False,
     show_layer_activations=False,
     show_trainable=False,
+    splines="ortho",
     **kwargs,
 ):
     """Convert a Keras model to dot format.
@@ -222,6 +223,9 @@ def model_to_dot(
         show_layer_activations: Display layer activations (only for layers that
             have an `activation` property).
         show_trainable: whether to display if a layer is trainable.
+        splines: Controls how edges are drawn. Defaults to `"ortho"`
+            (right-angle lines). Other options include `"curved"`,
+            `"polyline"`, `"spline"`, and `"line"`.
 
     Returns:
         A `pydot.Dot` instance representing the Keras model or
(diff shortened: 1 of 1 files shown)


hertschuh (commented):
Can you show screenshots of before and after? I tried it but I don't see any difference.

rstar327:
after_LR
before_LR

The main difference is the edge routing style: ortho draws straight right-angle lines, while curved draws smooth curves for the skip connections. On modern graphviz (14.x), both render correctly, the visual difference is subtle.
However, on older graphviz versions (e.g., 2.42 shipped with Ubuntu 22.04), ortho + LR can produce completely blank output or even crash for models with skip connections. The curved fallback ensures reliable rendering across graphviz versions.

hertschuh:
> The main difference is the edge routing style: ortho draws straight right-angle lines, while curved draws smooth curves for the skip connections. On modern graphviz (14.x), both render correctly, the visual difference is subtle. However, on older graphviz versions (e.g., 2.42 shipped with Ubuntu 22.04), ortho + LR can produce completely blank output or even crash for models with skip connections. The curved fallback ensures reliable rendering across graphviz versions.

To be honest, I prefer the straight lines, I find them easier to follow.

But most importantly, this doesn't address the bug. This change does not change the position of the boxes, which is what the bug is about. The reporter wants to see the U shape of the network, and wants to see a vertical layout. The `spline` option doesn't affect the layout of the boxes.

I am ok with adding a `spline` argument to `plot_model` and `model_to_dot`, so that users can change it, but I don't want to tie it to `rankdir` because it's completely unrelated.

Also, the bug should be removed from the description as it won't close the bug (and to be honest, I can't think of a reasonable way to fix the bug right now).

rstar327:
You're right, this doesn't address the U-shape layout, and tying splines to rankdir doesn't make sense. I'll rework the PR to expose splines as a configurable argument on plot_model and model_to_dot (default "ortho", no behavior change) and drop the Fixes #21801 reference.

PR #22279, opened 2026-02-24
Fix Torch backend LSTM cuDNN: remove blanket exception catch and forced cpu() transfer

### LINKED ISSUE #22274: Torch backend: LSTM fails to use cuDNN when criteria are satisfied
If you create a LSTM layer that satisfies requirements stated in [documentation](https://keras.io/api/layers/recurrent_layers/lstm/), you cannot do a forward pass on the layer created with `use_cudnn=True`.
Error:

```
ValueError: Exception encountered when calling LSTM.call().

use_cudnn=True was specified, but cuDNN is not supported for this layer configuration with this backend. Pass use_cudnn='auto' to fallback to a non-cuDNN implementation.

Arguments received by LSTM.call():
  • sequences=torch.Tensor(shape=torch.Size([1, 256, 128]), dtype=float32)
  • initial_state=None
  • mask=None
  • training=None
```

A reference to the cuDNN requirements in code (pretty much the same):
https://github.com/keras-team/keras/blob/0ddf96200f781f95dea0289aa87d85eb5eb3e733/keras/src/backend/torch/rnn.py#L548C1-L564C6

Colab to reproduce:
https://colab.research.google.com/drive/1OFLXgZpiUXDXTO1eebQYhXnq2HevZpoe

P.S. I explicitly defined the requirements even though on paper you should be good to go with the default values

rstar327:
## Summary

Fixes two issues in the Torch backend's cuDNN LSTM implementation that cause `use_cudnn=True` to fail even when all cuDNN criteria are satisfied:

1. **Narrowed exception catch from `Exception` to `RuntimeError`** — The blanket catch-all converted any exception into `NotImplementedError`, hiding real errors. Now only `RuntimeError` triggers the cuDNN fallback, preserving graceful degradation for legitimate config issues while letting real bugs propagate.

2. **Removed `.detach().clone().cpu()` on output tensors** — The `.cpu()` broke GPU computation, and `.detach().clone()` is unnecessary since the tensors don't need to be copied, matching the GRU implementation in #22115.

Fixes #22274

### DIFF
--- a/keras/src/layers/rnn/lstm_test.py
+++ b/keras/src/layers/rnn/lstm_test.py
@@ -67,6 +67,8 @@ class LSTMTest(testing.TestCase):
                 ]
             ),
             output,
+            atol=1e-5,
+            rtol=1e-5,
             tpu_atol=1e-3,
             tpu_rtol=1e-3,
         )
@@ -89,6 +91,8 @@ class LSTMTest(testing.TestCase):
                 ]
             ),
             output,
+            atol=1e-5,
+            rtol=1e-5,
             tpu_atol=1e-3,
             tpu_rtol=1e-3,
         )
@@ -111,6 +115,8 @@ class LSTMTest(testing.TestCase):
                 ]
             ),
             output,
+            atol=1e-5,
+            rtol=1e-5,
             tpu_atol=1e-3,
             tpu_rtol=1e-3,
         )
@@ -133,6 +139,8 @@ class LSTMTest(testing.TestCase):
                 ]
             ),
             output,
+            atol=1e-5,
+            rtol=1e-5,
             tpu_atol=1e-3,
             tpu_rtol=1e-3,
         )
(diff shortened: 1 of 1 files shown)


MarcosAsh:
Hey rstar327,

I think you're on the right track but you have some issues. I recently worked on the same thing on the GRU side in #22215 so I might be of help. For the try/except part you might want to narrow it rather than remove it entirely. The LSTM layer's call() method catches NotImplementedError to fall back to the non-cuDNN path, so removing the try/except means if cuDNN fails for a legitimate config reason it'll hard crash instead of falling back gracefully. Something like catching only RuntimeError from torch would preserve that behavior while still letting real bugs through.

Also I think Gemini's suggestion about .detach() is correct I fixed the same thing in #22115 if you want a reference. You can drop both .detach() and .clone() entirely, the tensors don't need to be copied.

Let me know if you want a hand with any of this!

rstar327:
Done

rstar327:
Done, added atol=1e-5, rtol=1e-5 to all assertAllClose calls in lstm_test.py, matching the same pattern from the GRU fix in #22115.

PR #22280, opened 2026-02-24
Use canonicalize_axis in Linspace and Logspace compute_output_spec

MarcosAsh:
## Summary                                                                                                                                       
                                                                                                                                              
  - Fix incorrect output shape computation for negative axis values in Linspace and Logspace compute_output_spec                                
  - Replace manual axis slicing with canonicalize_axis + list.insert, matching the pattern used in Geomspace (#22257)


hertschuh (approved):
Thanks for the improvement!

PR #22284, opened 2026-02-25
Add ignore_class parameter to SparseCategoricalCrossentropy 

cpuguy96:
Introduced ignore_class option to SparseCategoricalCrossentropy to allow ignoring specific classes during loss computation. Added unit tests to validate functionality of ignore_class with both standard and weighted inputs, as well as with logits.


keras/src/losses/losses.py
```diff
@@ -2344,6 +2344,8 @@ def sparse_categorical_crossentropy(
         y_true = ops.squeeze(y_true, axis=-1)
 
     if ignore_class is not None:
+        y_true = ops.convert_to_tensor(y_true)
+        y_pred = ops.convert_to_tensor(y_pred)
```
hertschuh: This should not be needed, please revert.

And if it is needed, this is an indication that there is a different bug.

hertschuh (commented):
Thanks for filling this gap!

keras/src/losses/losses.py
cpuguy96: This was added due to test failures in the new tests added only for torch.
```
FAILED keras/src/metrics/probabilistic_metrics_test.py::SparseCategoricalCrossentropyTest::test_ignore_class - TypeError: unsupported operand type(s) for *: 'numpy.ndarray' and 'Tensor'
FAILED keras/src/metrics/probabilistic_metrics_test.py::SparseCategoricalCrossentropyTest::test_ignore_class_from_logits - TypeError: unsupported operand type(s) for *: 'numpy.ndarray' and 'Tensor'
FAILED keras/src/metrics/probabilistic_metrics_test.py::SparseCategoricalCrossentropyTest::test_ignore_class_weighted - TypeError: unsupported operand type(s) for *: 'numpy.ndarray' and 'Tensor'
```
 I updated the function to use `ops.multiply` instead of `*` to remove the need of calling `ops.convert_to_tensor`.

hertschuh: Yes, that is the right fix.

hertschuh (approved):
Thanks!

PR #22288, opened 2026-02-25
fix: replace 32 bare except clauses with except Exception

haosenwang1018:
## What
Replace 32 bare `except:` clauses with `except Exception:`.

## Why
Bare `except:` catches `BaseException`, including `KeyboardInterrupt` and `SystemExit`, which can prevent clean process shutdown and mask critical errors. Using `except Exception:` catches all application-level errors while allowing system-level exceptions to propagate correctly.


haosenwang1018:
CI failures. Closing per policy.

hertschuh:
> CI failures. Closing per policy.

The tests are passing. The only thing that is missing is the CLA. Can you accept the CLA?

PR #22293, opened 2026-02-25
[BUG] Fixes random seed integer overflow bug in OpenVINO and Torch backend

### LINKED ISSUE #22291: [Bug][Keras 3 OpenVINO Backend]: Negative `seed` values are being generated(integer overflow)
Hi, devs.

I believe I have found a bug in the Keras 3 OpenVINO backend.

While working on issue https://github.com/openvinotoolkit/openvino/issues/34308 , I attempted to implement a Beta distribution random number generator. To achieve this, I called the Gamma random number generator twice.

```python
def beta(shape, alpha, beta, dtype=None, seed=None):
    from keras.src import backend

    dtype = dtype or floatx()

    seed1 = seed
    seed2 = seed
    # ...

    x = backend.random.gamma(shape, alpha, dtype=dtype, seed=seed1)
    y = backend.random.gamma(shape, beta, dtype=dtype, seed=seed2)

```

This resulted in the following example error(Negative seed values are being generated.):

Image

When calling the `gamma` function for the first time with `seed=None`, `seed1` is consistently generated as a positive value. However, when the `gamma` function is called multiple times, `seed1` occasionally results in a negative value.

I believe this issue occurs not only in `gamma` but also in other functions like `randint`.

Image

### My proposed solution is as follows:

(1) Add code within the `randint` and `gamma` functions to convert `seed1` to a positive value if it is negative.
This is a very intuitive approach and allows for a quick fix.

```python
    if seed1 < 0:
        seed1 = -1 * seed1
    if seed2 < 0:
        seed2 = -1 * seed2
```
However, we might need to apply this logic to each of the other functions individually.

(2) In search of a better approach, I conducted further debugging and I believe I have identified the root cause.
The execution enters `draw_seed(seed=None)` -> `global_seed_generator()`.

```python
def randint(shape, minval, maxval, dtype="int32", seed=None):
    dtype = dtype or "int32"
    ov_dtype = OPENVINO_DTYPES[dtype]
    seed_val = draw_seed(seed)
```

Image

Initially, it enters `SeedGenerator()` via the `if gen is None` condition.

Image

During the initialization process, it executes the `make_default_seed()` function, which consistently generates a positive random seed.

Image

However, starting from the second execution, it generates random numbers without calling `make_default_seed()`. During this process, negative random seeds can occur.

I believe (speculating that) the fundamental solution might be to ensure that `global_seed_generator()` does not generate negative seed values.

What do you think would be the best way to resolve this?

Please let me know if there is anything I have misunderstood or if you have a better solution in mind.

Thanks.

goyaladitya05:
Keras seed tensors carry 32-bit values, but both backends handled large seeds incorrectly.

**OpenVINO**
- `random_seed_dtype()` returned `"uint32"`, but Keras arithmetic promotes `uint32 * int32` → `int32`, making the declared dtype inconsistent with what the backend actually produced and causing `test_seed_generator_dtype` to fail.
- Seeds ≥ 2³¹ sign-flipped via PyBind11, violating OpenVINO C++'s `global_seed > 0` invariant and crashing graph compilation.
- `np.random.default_rng` rejected negative int32 seeds with `ValueError`.
- `np.array([large_value, 0], dtype="int32")` raised `OverflowError` on NumPy ≥ 1.24.

**Torch**
- Seeds ≥ 2³¹ silently wrapped to negative int32 values; two distinct seeds could collide and produce identical random output.

## Changes

**OpenVINO** (`core.py`, `random.py`)
- `random_seed_dtype()` now returns `"int32"` to match what the backend actually produces.
- `convert_to_tensor` list path: `np.asarray(x).astype(dtype)` instead of `np.array(x, dtype=dtype)` to avoid `OverflowError` on NumPy ≥ 1.24.
- `_rng_from_seed_data()`: reinterprets the int32 bit-pattern as uint32 before calling `np.random.default_rng`.
- `_random_uniform()`: masks seeds to 31 bits and clamps to ≥ 1 before forwarding to `ov_opset.random_uniform`. Replaces all bare call sites.

**Torch** (`core.py`)
- `random_seed_dtype()` returns `"int32"`; the bit-pattern is reinterpreted as uint32 at every call site via `& 0xFFFFFFFF` (matches PyTorch's `torch.Generator.manual_seed` behaviour).
- `convert_to_tensor` promotes Python `int` values outside `[−2³¹, 2³¹)` to `torch.int64` instead of silently truncating.

**Tests**
- `test_draw_seed_from_large_integer` moved from `random_test.py` to `seed_generator_test.py`; covers boundary values 2³¹ − 1, 2³¹, 2³² − 1.
-`SeedGeneratorTest::test_seed_generator_dtype` is now fixed and re-enabled.

Closes #22291
cc @rkazants
Thanks to @daehyun99 for reporting.

### DIFF
--- a/keras/src/backend/openvino/excluded_concrete_tests.txt
+++ b/keras/src/backend/openvino/excluded_concrete_tests.txt
@@ -29,7 +29,6 @@ AveragePoolingBasicTest::test_average_pooling2d6
 AveragePoolingBasicTest::test_average_pooling3d0
 AveragePoolingBasicTest::test_average_pooling3d1
 AveragePoolingBasicTest::test_average_pooling3d2
-BinaryCrossentropyTest::test_label_smoothing
 CircleTest::test_correctness
 CircleTest::test_correctness_weighted
 CircleTest::test_dtype_arg
@@ -267,7 +266,6 @@ EmbeddingTest::test_quantize_by_setting_dtype_policy_int4
 EmbeddingTest::test_quantize_int_int4
 EmbeddingTest::test_quantize_on_subclass_int4
 EmbeddingTest::test_quantize_when_already_quantized_int4
-ExtractSequencesOpTest::test_extract_sequences_call
 FalseNegativesTest::test_unweighted
 FalseNegativesTest::test_weighted
 FalsePositivesTest::test_unweighted
@@ -429,8 +427,6 @@ LSTMTest::test_pass_initial_state
 LSTMTest::test_statefulness
 MathOpsCorrectnessTest::test_erfinv_operation_basic
 MathOpsCorrectnessTest::test_erfinv_operation_dtype
-MathOpsCorrectnessTest::test_erfinv_operation_edge_cases
-MathOpsCorrectnessTest::test_extract_sequences
 MathOpsCorrectnessTest::test_logdet
 MaxPoolingBasicTest::test_max_pooling1d0
 MaxPoolingBasicTest::test_max_pooling1d1
@@ -470,7 +466,6 @@ NNOpsBehaviorTest::test_fold_no_padding
 NNOpsBehaviorTest::test_fold_non_square_output
 NNOpsBehaviorTest::test_fold_tuple_params
 NNOpsBehaviorTest::test_invalid_strategy_ctc_decode
-NNOpsBehaviorTest::test_logit_recovery_binary_crossentropy
 NNOpsCorrectnessTest::test_conv_transpose_
 NNOpsCorrectnessTest::test_ctc_decode
 NNOpsCorrectnessTest::test_fold
@@ -479,12 +474,7 @@ NNOpsCorrectnessTest::test_fold_no_padding
 NNOpsCorrectnessTest::test_fold_non_square_output
 NNOpsCorrectnessTest::test_fold_tuple_params
 NNOpsCorrectnessTest::test_glu
-NNOpsCorrectnessTest::test_log_softmax_correctness_with_axis_tuple
-NNOpsCorrectnessTest::test_moments
-NNOpsCorrectnessTest::test_normalize
 NNOpsCorrectnessTest::test_polar_corectness
-NNOpsCorrectnessTest::test_rms_normalization_10.0
-NNOpsCorrectnessTest::test_softmax_correctness_with_axis_tuple
 NNOpsCorrectnessTest::test_sparsemax
 NNOpsDtypeTest::test_ctc_decode
 NNOpsDtypeTest::test_glu_
@@ -653,14 +643,6 @@ StackedRNNTest::test_stacked_lstm_cell_mask
 StackedRNNTest::test_statefullness_single_state_stack
 StackedRNNTest::test_statefullness_two_states_stack
 TestNumericalUtils::test_build_pos_neg_masks
-TestTrainer::test_evaluate_flow_eager
-TestTrainer::test_evaluate_flow_graph_fn
-TestTrainer::test_evaluate_flow_jit
-TestTrainer::test_evaluate_return_list_respect_metrics_order
-TestTrainer::test_evaluate_with_different_batch_size_same_loss
-TestTrainer::test_on_batch_methods_without_training_eager
-TestTrainer::test_on_batch_methods_without_training_graph_fn
-TestTrainer::test_on_batch_methods_without_training_jit
 TestTrainer::test_predict_dropout
 TestTrainer::test_predict_flow_eager
 TestTrainer::test_predict_flow_graph_fn
@@ -674,7 +656,6 @@ TestTrainer::test_predict_preserve_order_1_non_jit
 TestTrainer::test_predict_preserve_order_50_eager
 TestTrainer::test_predict_preserve_order_50_jit
 TestTrainer::test_predict_preserve_order_50_non_jit
-TestTrainer::test_steps_per_execution_steps_count_without_training
 TestTrainer::test_symbolic_build
 TextDatasetFromDirectoryTest::test_sample_count_grain
 TextDatasetFromDirectoryTest::test_text_dataset_from_directory_binary_grain
--- a/keras/src/backend/torch/random.py
+++ b/keras/src/backend/torch/random.py
@@ -15,18 +15,15 @@ from keras.src.random.seed_generator import make_default_seed
 # see: https://github.com/pytorch/pytorch/issues/88576
 @dynamo.disable()
 def torch_seed_generator(seed):
-    first_seed, second_seed = draw_seed(seed)
     device = get_device()
     if device == "meta":
         # Generator is not supported by the meta device.
         return None
     generator = torch.Generator(device=get_device())
-    # Seeds are stored as int32 but are conceptually uint32 values.  Reinterpret
-    # the bit pattern via & 0xFFFFFFFF so manual_seed always receives a
-    # non-negative value, avoiding sign-flip for seeds >= 2**31.
-    generator.manual_seed(
-        (int(first_seed) & 0xFFFFFFFF) + (int(second_seed) & 0xFFFFFFFF)
-    )
+    first_seed, second_seed = draw_seed(seed)
+    # Re-interpret as uint32 and combine; the counter in second_seed ensures
+    # each SeedGenerator call produces a distinct seed.
+    generator.manual_seed(int(first_seed + second_seed) & 0xFFFFFFFF)
     return generator
 
 
@@ -210,9 +207,7 @@ def gamma(shape, alpha, dtype=None, seed=None):
     # Do not draw seed during symbolic execution
     if not get_device() == "meta":
         first_seed, second_seed = draw_seed(seed)
-        torch.manual_seed(
-            (int(first_seed) & 0xFFFFFFFF) + (int(second_seed) & 0xFFFFFFFF)
-        )
+        torch.manual_seed(int(first_seed + second_seed) & 0xFFFFFFFF)
     gamma_distribution = torch.distributions.gamma.Gamma(alpha, beta)
     sample = gamma_distribution.sample().type(dtype)
     torch.random.set_rng_state(prev_rng_state)
@@ -228,9 +223,7 @@ def binomial(shape, counts, probabilities, dtype=None, seed=None):
     # Do not draw seed during symbolic execution
     if not get_device() == "meta":
         first_seed, second_seed = draw_seed(seed)
-        torch.manual_seed(
-            (int(first_seed) & 0xFFFFFFFF) + (int(second_seed) & 0xFFFFFFFF)
-        )
+        torch.manual_seed(int(first_seed + second_seed) & 0xFFFFFFFF)
     binomial_distribution = torch.distributions.binomial.Binomial(
         total_count=counts, probs=probabilities
     )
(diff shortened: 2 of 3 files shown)


keras/src/random/random_test.py
```diff
@@ -182,6 +182,24 @@ def test_shuffle(self):
...
+        ),
+    )
+    def test_openvino_large_seed_boundary(self):
+        # 2**31 - 1 is the max for signed int32. 
+        # 2**31 is where the overflow/sign-flip occurs in C++.
+        # 2**32 - 1 is the max for uint32.
+        boundary_seeds = [2147483647, 2147483648, 4294967295]
+        
+        for s in boundary_seeds:
+            res = random.normal((2, 2), seed=s)
```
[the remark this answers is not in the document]
goyaladitya05: Good catch. Changed the func.

goyaladitya05:
Fixed a `RuntimeError` in the Torch backend caused by `int32` overflow when using random seeds $\ge 2^{31}$. 
Although PyTorch lacks a native `uint32` type, these changes ensure that seeds are correctly mapped to `int64` via the Keras "uint32" dtype. This allows the backend to support the full 32-bit unsigned range expected by Keras, aligning it with the JAX and TensorFlow backends and preventing crashes during random number generation.

keras/src/backend/openvino/random.py
```diff
@@ -13,10 +13,35 @@
...
+def _random_uniform(shape, minval, maxval, dtype, seed1, seed2):
+    # Pybind11 will sign-flip values >= 2**31, and int32 wrap-around can
+    # produce negative values.  Mask to 31 bits so seeds are in [0, 2**31-1],
+    # then clamp to at least 1: the OpenVINO C++ layer requires each seed to
+    # be strictly positive (> 0).
+    safe_seed1 = max(1, int(seed1) & 0x7FFFFFFF)
+    safe_seed2 = max(1, int(seed2) & 0x7FFFFFFF)
+    return ov_opset.random_uniform(
+        shape, minval, maxval, dtype, safe_seed1, safe_seed2
+    ).output(0)
```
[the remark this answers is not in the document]
goyaladitya05: added the docstring.

goyaladitya05:
Rebased this on upstream master.

Also,
- Normalized excluded_concrete_tests.txt to LF line endings with no blank lines (was CRLF with blank separators, fixes future rebase conflicts on this file). During my rebase, I discovered that it was inadvertently introduced by #22309.
- Verified` SeedGeneratorTest::test_seed_generator_dtype` passes with the OpenVINO backend and removed it from the exclusion list.

Hi @hertschuh,
I would request you to merge this branch before the others, as it also includes a fix for a line ending issue for `excluded_concrete_tests.txt` . Thanks !

keras/src/backend/torch/core.py
```diff
@@ -210,6 +210,10 @@ def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
         if isinstance(x, bool):
             return torch.as_tensor(x, dtype=torch.bool, device=get_device())
         elif isinstance(x, int):
+            if x < -(2**31) or x >= 2**31:
+                return torch.as_tensor(
+                    x, dtype=torch.int64, device=get_device()
+                )
```
hertschuh, line 216: It looks like we have a deeper problem with this. This applies to some of the other backends and won't work for python lists / tuple (the code is below).

(I'm not asking you to do anything about this right now).

keras/src/backend/torch/core.py
```diff
@@ -685,8 +689,9 @@ def unstack(x, num=None, axis=0):
 
 
 def random_seed_dtype():
-    # uint32 doesn't exist in torch, use int32 instead.
-    return "int32"
+    # uint32 doesn't exist in torch. Use int64 so that seed values up to
+    # 2**63-1 are representable without silent sign-flip or collision.
+    return "int64"
```
hertschuh: I'm not following how this solves the problem.

If it's signed, you still have the issue of sign-flip or collision.

I think this is relying on the fact that the values are between `0` and `2**32 - 1`, but that's not the intent, if you say `int64` it's supposed to use the whole range including negative values.

keras/src/backend/openvino/core.py
```diff
@@ -762,7 +762,10 @@ def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
                     for item in tree.flatten(x)
                 ]
             )
-        x = np.array(x, dtype=dtype)
+        # Use astype (unsafe casting) rather than np.array(..., dtype=dtype)
+        # to avoid OverflowError on NumPy >=1.24 when values exceed the target
+        # dtype range (e.g. a uint32 seed value stored in an int32 tensor).
+        x = np.asarray(x).astype(dtype)
```
hertschuh: But now this silently truncates instead of failing. I think this should be handled differently, but it's a bigger change outside of the scope of this PR.

keras/src/backend/torch/core.py
goyaladitya05: You're right. Now, I changed that func to return int32, and then enabled masking further in the code. 
(https://github.com/keras-team/keras/pull/22293/changes#diff-fd2d8efd3e760ef636a1678b4923c95391958632801011c9fd92bfc67fb6a49aR27)

keras/src/backend/openvino/core.py
goyaladitya05: In the latest commits, I fixed the main cause in `seed_generator.py`, when `random_seed_dtype()` is `int32`, the integer seed is now explicitly reinterpreted via n`p.array(seed & 0xFFFFFFFF, dtype=np.uint32).view(np.int32))` a 2's-complement bit-cast before `convert_to_tensor` is called, so the function never receives an out-of-range value.

keras/src/backend/torch/core.py
goyaladitya05: Agreed. We would have to implement a broader fix for lists/tuples, in a seperate PR.

keras/src/random/seed_generator.py
```diff
@@ -153,7 +153,19 @@ def draw_seed(seed):
...
+        # their seed dtype as a signed type (e.g. "int32"). np.array(x,
+        # dtype="int32") raises OverflowError on NumPy >= 1.24 for values
+        # >= 2**31. Perform an explicit 2's-complement bit-cast via uint32
+        # so that the integer passed to convert_to_tensor is always in the
+        # representable range of the declared dtype while preserving full
+        # 32-bit entropy.
+        if dtype == "int32":
+            seed = int(
+                np.array(seed & 0xFFFFFFFF, dtype=np.uint32).view(np.int32)
+            )
```
hertschuh: There is a simpler way to do this without requiring all the machinery of creating a Numpy array:
```python
# Re-interpret the bits of a uint32 as an int32
seed = (seed ^ 0x80000000) - 0x80000000
```

keras/src/backend/torch/random.py
```diff
@@ -21,7 +21,12 @@ def torch_seed_generator(seed):
         # Generator is not supported by the meta device.
         return None
     generator = torch.Generator(device=get_device())
-    generator.manual_seed(int(first_seed + second_seed))
+    # Seeds are stored as int32 but are conceptually uint32 values.  Reinterpret
+    # the bit pattern via & 0xFFFFFFFF so manual_seed always receives a
+    # non-negative value, avoiding sign-flip for seeds >= 2**31.
+    generator.manual_seed(
+        (int(first_seed) & 0xFFFFFFFF) + (int(second_seed) & 0xFFFFFFFF)
+    )
```
hertschuh: I'm not sure why it was extracting the 2 parts of the seed and summing them. I don't really think it makes sense actually.

I would do this actually. Line 18:
```python
# Re-interpret the int32 seed as a uint32
seed = int(draw_seed(seed)[0]) & 0xFFFFFFFF
```
Then simply:
```python
generator.manual_seed(seed)
```

Same in all places with the same comment `# Re-interpret the int32 seed as a uint32`

keras/src/random/seed_generator.py
goyaladitya05: I did it without the numpy array now.

> seed = (seed ^ 0x80000000) - 0x80000000

This would not be correct either, we'd still have to mask it. Updated it with masking.

keras/src/backend/torch/random.py
goyaladitya05: I was implemented some other logic earlier, this was a leftover bit I hadn't cleaned up. Corrected it now.

goyaladitya05:
Hi @hertschuh, I've updated the code with changes, and rebased on master. Please review.

keras/src/backend/torch/random.py
```diff
@@ -15,13 +15,14 @@
...
-    first_seed, second_seed = draw_seed(seed)
     device = get_device()
     if device == "meta":
         # Generator is not supported by the meta device.
         return None
     generator = torch.Generator(device=get_device())
-    generator.manual_seed(int(first_seed + second_seed))
+    # Re-interpret the int32 seed as a uint32.
+    seed = int(draw_seed(seed)[0]) & 0xFFFFFFFF
+    generator.manual_seed(seed)
```
hertschuh: Sorry, I guess I misunderstood how the whole seed thing works on Torch.

You'll have to go back to:
```python
first_seed, second_seed = draw_seed(seed)
```

And then:
```python
generator.manual_seed(int(first_seed + second_seed) & 0xFFFFFFFF)
```

keras/src/backend/openvino/excluded_concrete_tests.txt
```diff
@@ -29,6 +29,7 @@ AveragePoolingBasicTest::test_average_pooling2d6
 AveragePoolingBasicTest::test_average_pooling3d0
 AveragePoolingBasicTest::test_average_pooling3d1
 AveragePoolingBasicTest::test_average_pooling3d2
+BinaryCrossentropyTest::test_label_smoothing
```
hertschuh: A bunch of tests got re-added to the exclusion list. Is this a bad rebase?

keras/src/backend/torch/random.py
goyaladitya05: > seed = int(draw_seed(seed)[0]) & 0xFFFFFFFF

I tried this, but it would not work, because it always reads index [0] (the base seed),  ignoring index [1] (the counter that increments on each call). So a SeedGenerator would seed the generator identically on every call, producing the same random output.

keras/src/backend/torch/random.py
goyaladitya05: yeah, I'm just pushing the changes.

keras/src/backend/openvino/excluded_concrete_tests.txt
goyaladitya05: yeah. Those ops got merged in the morning, so I need to update that file as well. Pushing the changes soon.

goyaladitya05:
> I wish the Torch part had been a separate PR...
> 
> Hopefully this is the last change

Yeah. I initially thought this issue affected only OpenVINO, but later discovered that PyTorch also suffers from it, which is why the change ended up covering both.

goyaladitya05:
> Thanks for the complex fix!

@hertschuh Thanks a lot for your feedback on all these PRs. We're aiming for level parity of the OpenVINO backend, with the other backends, and that's the reason for a lot of OpenVINO PRs in the repo lately. I’m looking into Summer 2026/Jan 2027 roles and would appreciate the chance to chat briefly about the team's direction. Is there a better place than GitHub for this? I tried searching your email, but could not find it.

hertschuh:
I contacted you on linkedin.

PR #22295, opened 2026-02-26
fix(softmax): validate mask shape matches inputs shape

### LINKED ISSUE #19722: softmax sliently generate a wrong output when the mask has an  incompatible shape
As shown in this [colab](https://colab.research.google.com/drive/1B5c-9QEwn8GWvi0kOp_mPvSt5BGV34tM?usp=sharing), the mask can arbitrarily change the shape of output.
Although the document requires that the mask has the same shape with input, no check exists  actually.
In my project, the shape of input is hard to compute and this potential incompatibility can affect the result and is hard to debug since softmax now silently generates an wrong output.
Currently the output is:

You can identify the shape changed by mask(line 2 and line 3 of output), which is different from the correct shape computed by compute_output_shape() (line4).

rstar327:
## Summary
- Add shape validation in `Softmax.call()` when `mask` is provided
- Previously, an incompatible mask shape would silently broadcast via `numpy.where`, producing wrong output shape
- Now raises `ValueError` with a clear message when shapes don't match
- Adds test for mask shape mismatch

Fixes #19722

### DIFF
--- a/keras/src/layers/activations/softmax.py
+++ b/keras/src/layers/activations/softmax.py
@@ -36,8 +36,10 @@ class Softmax(Layer):
 
     Call arguments:
         inputs: The inputs (logits) to the softmax layer.
-        mask: A boolean mask of the same shape as `inputs`. The mask
-            specifies 1 to keep and 0 to mask. Defaults to `None`.
+        mask: A boolean mask that is broadcastable to `inputs`. The mask
+            specifies 1 to keep and 0 to mask. Each dimension of the mask
+            must either be 1 or match the corresponding dimension of
+            `inputs`; it must not be larger. Defaults to `None`.
 
     Returns:
         Softmaxed output with the same shape as `inputs`.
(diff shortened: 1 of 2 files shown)


keras/src/layers/activations/softmax.py
```diff
@@ -52,6 +52,12 @@ def __init__(self, axis=-1, **kwargs):
 
     def call(self, inputs, mask=None):
         if mask is not None:
+            if inputs.shape != mask.shape:
+                raise ValueError(
+                    "The `mask` must have the same shape as `inputs`. "
+                    f"Received: inputs.shape={inputs.shape}, "
+                    f"mask.shape={mask.shape}"
+                )
```
hertschuh: Please look at this thread: https://github.com/keras-team/keras/pull/19736#discussion_r1607342013

The mask doesn't need to be the same shape as the inputs, it just needs to be broadcastable to the input shape.

So:
- please update the docstring above to mention the broadcast vs. the exact shape match
- the bug mentioned in https://github.com/keras-team/keras/issues/19722 is that the broadcast happens both ways, if mask is "bigger" than the inputs, the inputs are broadcast. This is weird and we should disallow that. But the criteria is more complicated.

hertschuh (commented):
Thanks for working on this!

rstar327:
Done, updated validation to allow broadcastable masks while disallowing masks larger than inputs.
Updated docstring and added tests for broadcastable, fewer-dim, mismatch, and too-many-dims cases.

PR #22296, opened 2026-02-26
fix(normalization): sort axis to fix crash with unsorted contiguous axes

### LINKED ISSUE #22201: [Bug] Torch backend: RMSNormalization crashes with unsorted contiguous axes

rstar327:
## Summary
- Sort `axis` in `RMSNormalization.build()` and `LayerNormalization.build()` so weight shape matches dimension order
- Sort `axis` in `ops.nn._rms_normalization` and `ops.nn._layer_normalization` so `normalized_shape` is correct for PyTorch's `F.rms_norm`/`F.layer_norm`
- Unsorted but contiguous axes like `[-1, -2]` previously caused shape mismatch crashes on the Torch backend
- Adds tests for both layers verifying sorted and unsorted axes produce identical output

Fixes #22201

### DIFF
--- a/keras/src/layers/normalization/layer_normalization.py
+++ b/keras/src/layers/normalization/layer_normalization.py
@@ -151,8 +151,9 @@ class LayerNormalization(Layer):
         self.autocast = False
 
     def build(self, input_shape):
-        if isinstance(self.axis, list):
-            shape = tuple([input_shape[dim] for dim in self.axis])
+        if isinstance(self.axis, (list, tuple)):
+            self.axis = sorted(self.axis)
+            shape = tuple(input_shape[dim] for dim in self.axis)
         else:
             shape = (input_shape[self.axis],)
             self.axis = [self.axis]
--- a/keras/src/layers/normalization/layer_normalization_test.py
+++ b/keras/src/layers/normalization/layer_normalization_test.py
@@ -133,3 +133,13 @@ class LayerNormalizationTest(testing.TestCase):
         with backend.AutocastScope("float16"):
             layer.gamma.assign(large_value)
             self.assertAllClose(layer.gamma.value, large_value)
+
+    def test_unsorted_axis(self):
+        x = np.random.randn(2, 3, 4).astype("float32")
+        layer_sorted = layers.LayerNormalization(axis=[-2, -1])
+        layer_unsorted = layers.LayerNormalization(axis=[-1, -2])
+        out_sorted = layer_sorted(x)
+        out_unsorted = layer_unsorted(x)
+        self.assertEqual(out_sorted.shape, (2, 3, 4))
+        self.assertEqual(out_unsorted.shape, (2, 3, 4))
+        self.assertAllClose(out_sorted, out_unsorted)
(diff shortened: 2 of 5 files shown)


hertschuh (approved):
Thanks for the fix!

PR #22301, opened 2026-02-26
[OpenVINO backend] Implement image format transform operations for NumPy support

CityBoy-Claude:
This PR implements the image format transform operations (rgb_to_grayscale, rgb_to_hsv, hsv_to_rgb) for the OpenVINO backend. Update the test filters of Image operations, enable subset of the implemented operations.

Closes https://github.com/openvinotoolkit/openvino/issues/34377

### DIFF
--- a/keras/src/backend/openvino/image.py
+++ b/keras/src/backend/openvino/image.py
@@ -164,6 +164,15 @@ def hsv_to_rgb(images, data_format=None):
     )
 
     def hsv_planes_to_rgb_planes(hue, saturation, value):
+        def channel_value(channel_delta, one_minus_saturation):
+            return ov_opset.multiply(
+                value,
+                ov_opset.add(
+                    one_minus_saturation,
+                    ov_opset.multiply(saturation, channel_delta),
+                ),
+            )
+
         dh = ov_opset.multiply(
             ov_opset.mod(hue, ov_opset.constant(1.0, dtype=ov_type)),
             ov_opset.constant(6.0, dtype=ov_type),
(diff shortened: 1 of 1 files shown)


CityBoy-Claude:
Hi @hertschuh this PR is ready to review, would you like to take a look? Thanks!

keras/src/ops/image_test.py
```diff
@@ -1040,13 +1040,13 @@ def test_rgb_to_grayscale(self):
         out = kimage.rgb_to_grayscale(x)
         ref_out = tf.image.rgb_to_grayscale(x)
         self.assertEqual(tuple(out.shape), tuple(ref_out.shape))
-        self.assertAllClose(ref_out, out)
+        self.assertAllClose(ref_out.numpy(), out)
```
hertschuh, line 1043: Can you revert this file?

Under the hood `assertAllClose` already does the numpy conversion. If it's not working, it means something else is broken.

CityBoy-Claude: OpenVINO backend's `convert_to_numpy()` cannot convert tensorflow Tensor ([ref](https://github.com/keras-team/keras/blob/969345709d4b48324b54289aad7a892bd9a6b4de/keras/src/backend/openvino/core.py#L817)). I am not sure if I can update that 🤔 . Do you have any suggestions?

hertschuh: Oh, I didn't not realize that the reference was done with Tensorflow. In this case your change is correct.

hertschuh (approved):
Thanks!

PR #22302, opened 2026-02-27
Implement nanargmin function in keras.ops

shashaka:
Adds keras.ops.nanargmin, which computes the indices of the minimum values along a specified axis while ignoring NaN values in the input tensor.

The operation follows Jax semantics for axis and keepdims, returning integer indices corresponding to the minimum non-NaN values. For slices where all elements are NaN, it returns -1 per reduced slice


keras/src/ops/numpy.py
```diff
@@ -5319,6 +5319,67 @@ def moveaxis(x, source, destination):
...
+        )
+
+    def compute_output_spec(self, x):
+        if self.keepdims:
+            return KerasTensor(x.shape, dtype="int32")
+        if self.axis is None:
+            return KerasTensor([], dtype="int32")
+        return KerasTensor(
+            reduce_shape(x.shape, axis=[self.axis]), dtype="int32"
+        )
```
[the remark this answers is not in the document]
shashaka: Updated.

keras/src/backend/numpy/numpy.py
```diff
@@ -986,6 +986,21 @@ def moveaxis(x, source, destination):
...
+
+    nan_mask = np.isnan(x)
+
+    return np.where(
+        np.all(nan_mask, axis=axis, keepdims=keepdims),
+        -1,
+        np.nanargmin(
+            np.where(nan_mask, np.inf, x), axis=axis, keepdims=keepdims
+        ).astype("int32"),
+    )
```
hertschuh, line 1001: Wait, why can't you just do `return np.nanargmin(x, axis=axis, keepdims=keepdims)`?

shashaka: NumPy’s np.nanargmin raises a ValueError when it encounters an all-NaN slice:

```
a, mask = _replace_nan(a, np.inf)
if mask is not None and mask.size:
    mask = np.all(mask, axis=axis)
    if np.any(mask):
        raise ValueError("All-NaN slice encountered")
res = np.argmin(a, axis=axis, out=out, keepdims=keepdims)
return res
```

In contrast, JAX returns -1 for all-NaN slices instead of raising an error.
To align our behavior with JAX’s implementation, I added a custom NumPy-based implementation that returns -1 when an all-NaN slice is encountered, rather than raising an exception.

Issue #22162, opened 2026-02-12
Missing input validation for PReLU alpha_initializer and LeakyReLU alpha


amadhan882:
### Summary
`keras.layers.PReLU` and `keras.layers.LeakyReLU` lack sufficient input validation during instantiation and the `build()` phase. 

1. **PReLU:** Accepts `alpha_initializer=None`, leading to an inconsistent state that results in downstream failures during training or serialization.
2. **LeakyReLU:** Accepts `NaN` values for the `alpha` parameter, allowing the creation of a broken layer that produces non-numeric gradients.

This is a **usability and reliability bug** that violates the "Fail-Fast" principle of the Keras API.

### Reproduction Script
```python
import os
os.environ["KERAS_BACKEND"] = "tensorflow"
import keras
import numpy as np

print(f"Testing Keras version: {keras.__version__}")

# --- Case 1: PReLU ---
print("\n--- Testing PReLU ---")
try:
    layer_prelu = keras.layers.PReLU(alpha_initializer=None)
    layer_prelu.build((None, 10)) 
    print("[ISSUE] PReLU accepted None as alpha_initializer and built successfully!")
except Exception as e:
    print(f"[OK] PReLU caught error: {e}")

# --- Case 2: LeakyReLU ---
print("\n--- Testing LeakyReLU ---")
try:
    layer_leaky = keras.layers.LeakyReLU(alpha=float('nan'))
    print("[ISSUE] LeakyReLU accepted NaN as alpha!")
except Exception as e:
    print(f"[OK] LeakyReLU caught NaN.")
```

### Observed Result
```text
Testing Keras version: 3.14.0
--- Testing PReLU ---
[ISSUE] PReLU accepted None as alpha_initializer and built successfully!

--- Testing LeakyReLU ---
[ISSUE] LeakyReLU accepted NaN as alpha!
```

### Actual Behavior
The layers are instantiated and built without any validation errors. The `PReLU` layer remains in an inconsistent state, and `LeakyReLU` accepts mathematically undefined values (`NaN`), deferring crashes to the training phase.

### Expected Behavior
The constructor should raise a `ValueError` immediately if:
- `alpha_initializer` is `None` in `PReLU`.
- The `alpha` (or `negative_slope`) parameter in `LeakyReLU` is `NaN`.

### Root Cause
[...]

PR #22306, opened 2026-02-27
fix: correct negative-order detection in is_continuous_axis

shriramThakare3:


keras/src/utils/python_utils.py
```diff
@@ -6,21 +6,14 @@
 
 
 def is_continuous_axis(axis):
-    # Used to determine whether the dimensions in an axis are continuous
-    if isinstance(axis, int) or len(axis) == 1:
+    if len(axis) <= 1:
```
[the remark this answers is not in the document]
hertschuh: +1 please re-add the `isinstance(axis, int)`.

keras/src/utils/python_utils.py
```diff
@@ -6,21 +6,14 @@
 
 
 def is_continuous_axis(axis):
-    # Used to determine whether the dimensions in an axis are continuous
```
hertschuh, line 9: Please re-add the comment that was there, it's not entirely obvious what this function does.

hertschuh (commented):
Thank you for the improvement!

keras/src/utils/python_utils.py
shriramThakare3: Done — re-added the isinstance(axis, int) check and the comment.

keras/src/utils/python_utils.py
```diff
@@ -7,20 +7,14 @@
 
 def is_continuous_axis(axis):
     # Used to determine whether the dimensions in an axis are continuous
-    if isinstance(axis, int) or len(axis) == 1:
+    if isinstance(axis,int) or len(axis) <= 1:
```
hertschuh: You need the space between `axis,` and `int`. This is why the code format check is failing.

hertschuh (changes requested):

keras/src/utils/python_utils.py
shriramThakare3: added space

shriramThakare3:
### Summary

Fixes #22162

`PReLU` and `LeakyReLU` lacked sufficient input validation, allowing
broken layer configurations to be created silently and only failing
later during training or serialization.

AI tool used: Claude Sonnet 4.6

- [x] I have read the CONTRIBUTING document.
- [x] I have added the tests to cover my changes.
- [x] I have updated the documentation accordingly.

### Details and comments

**`keras/src/layers/activations/prelu.py`**
Added a `None` check for `alpha_initializer` in `__init__`. Passing
`None` previously caused a silent broken state that deferred errors to
`build()` or training. A `ValueError` is now raised immediately with a
clear message pointing to the correct fix.

**`keras/src/layers/activations/leaky_relu.py`**
Added `math.isnan()` to the existing validation condition. `NaN` values
were previously accepted and would produce non-numeric gradients silently
during training. The check is folded into the existing `if` guard
alongside the existing `None` and negative value checks, and the error
message is updated to mention `NaN`.

Before:
```python
if negative_slope is None or negative_slope < 0:
```
After:
```python
if negative_slope is None or math.isnan(negative_slope) or negative_slope < 0:
```

Both fixes follow the fail-fast principle already used elsewhere in
Keras — invalid configurations are caught at layer instantiation rather
than deferred to the training phase.

shriramThakare3:
@hertschuh removed the changes in prelu and leaky_relu

shriramThakare3:
@hertschuh  
changes are  made , but in prelu there is "+" showing but not in my code
apologizes i dont have idea about that

PR #22313, opened 2026-02-28
[OpenVINO backend] Support for LSTM 

andersendsa:
Details :
Implement LSTM operation for the OpenVINO backend.
Closes issue : https://github.com/openvinotoolkit/openvino/issues/34403

### DIFF
--- a/keras/src/layers/rnn/gru.py
+++ b/keras/src/layers/rnn/gru.py
@@ -583,7 +583,7 @@ class GRU(RNN):
                     # since cuDNN ops aren't XLA compatible.
                     if backend.backend() == "tensorflow":
                         self.supports_jit = False
-                     if getattr(self, "zero_output_for_mask", False) and mask is not None:
+                    if getattr(self, "zero_output_for_mask", False) and mask is not None:
                         last_output, outputs, new_states = out
                         mask_expanded = ops.expand_dims(
                             ops.cast(mask, "bool"), axis=-1


andersendsa:
Hi @hertschuh  this pr passes all the tests and it is ready for review

keras/src/backend/openvino/excluded_tests.txt
```diff
@@ -37,4 +37,5 @@ keras/src/saving
 keras/src/trainers/compile_utils_test.py
 keras/src/trainers/epoch_iterator_test.py
 keras/src/trainers/trainer_test.py
-keras/src/utils
\ No newline at end of file
+keras/src/utils
+keras/src/layers/rnn/gru_test.py
```
goyaladitya05: Agreed with gemini here. Adding gru_test.py when the parent directory is already excluded makes no sense. Please clean up the excluded_tests.txt diff.

keras/src/ops/__init__.py
goyaladitya05, line 1: Manually overriding __dir__ and __getattr__ to force-import the entire library seems unnecessary. We only make changes to the files involving the backend, for a backend specific implementation. Is there a reason for this?

keras/src/initializers/random_initializers.py
goyaladitya05, line 1: I'm confused by the inclusion of these changes in this PR. keras/src/initializers/ is a global utility and should remain backend-agnostic. Unless there is a specific reason these initializers need to be changed for all of Keras, I'd suggest reverting the changes to this file.

keras/src/backend/openvino/rnn.py
```diff
@@ -304,5 +476,18 @@ def numpy_scan(f, init, xs, reverse=False, mask=None):
...
 
-def cudnn_ok(*args, **kwargs):
-    return False
+def cudnn_ok(
+    activation,
+    recurrent_activation,
+    unroll,
+    use_bias,
+    reset_after=None,
+):
```
hertschuh: Just to clarify, in this context, `cudnn_ok` means "use the native OpenVino implementation" (we should rename it).

But is the criteria really the same as for CuDNN? That seems surprising.

keras/src/initializers/random_initializers.py
hertschuh: +1

Why was this change needed?

keras/src/ops/__init__.py
hertschuh: +1

Revert this.

keras/src/backend/openvino/rnn.py
```diff
@@ -2,6 +2,8 @@
 from openvino import Model
 from openvino import Type
 
+from keras.src import activations
+from keras.src import ops
```
hertschuh: You cannot import ops like this here, this is what's creating the circular imports.

But they're not even used.

keras/src/ops/__init__.py
hertschuh: Revert this whole file, it's failing the code format check.

keras/src/initializers/random_initializers.py
hertschuh: Revert this whole file, it's failing the code format check.

keras/src/backend/openvino/rnn.py
```diff
@@ -304,5 +476,5 @@ def numpy_scan(f, init, xs, reverse=False, mask=None):
     )
 
 
-def cudnn_ok(*args, **kwargs):
-    return False
+def cudnn_ok(*args , **kwargs):
+    return False
```
hertschuh: Revert this change file, it's failing the code format check.

keras/src/backend/openvino/rnn.py
```diff
@@ -284,8 +301,163 @@ def _slice_at_0(x):
...
+    ).output(0)
+    initial_state_c = ov_opset.unsqueeze(
+        initial_state_c, ov_opset.constant([1], Type.i32)
+    ).output(0)
+    direction = "forward"
+    if go_backwards:
+        direction = "reverse"
+
+    def get_activation_name(act):
+        from keras.src import activations
```
hertschuh: activations are already imported at the top, you don't need this import.

keras/src/backend/openvino/rnn.py
andersendsa: Hey @hertschuh just wanted to double-check—do you want me to revert the whole file back to the previous commit?

hertschuh: Oh, no, just the last 2 lines of this file.

goyaladitya05:
Hey @andersendsa, could you also enable the tests releted to LSTM from excluded_concrete_tests.txt ?

andersendsa:
hi @hertschuh  i have reverted the file and removed the test cases the pr is ready for review

goyaladitya05:
@andersendsa Is there an email on which I could reach out to you?

PR #22324, opened 2026-03-01
Use channels_last memory format for torch conv ops

MarcosAsh:
## Summary

- Ensures conv and conv_transpose tensors use PyTorch's `channels_last` memory format when `data_format="channels_last"`, allowing PyTorch to route to cuDNN's native NHWC kernels instead of internally copying tensors.
- After `_transpose_conv_kernel`, the permuted kernel has non-standard strides that prevent cuDNN's optimized NHWC path. `contiguous(memory_format=channels_last)` fixes this with one copy.
- No effect on conv1d (PyTorch has no channels_last for 3D tensors) or `channels_first` data format (already optimal).

Addresses #18457.

### DIFF
--- a/keras/src/backend/torch/nn.py
+++ b/keras/src/backend/torch/nn.py
@@ -322,6 +322,21 @@ def _transpose_conv_kernel(kernel):
     return kernel
 
 
+def _get_channels_last_memory_format(ndim):
+    if ndim == 4:
+        return torch.channels_last
+    elif ndim == 5:
+        return torch.channels_last_3d
+    return None
+
+
+def _maybe_convert_to_channels_last(tensor):
+    mem_fmt = _get_channels_last_memory_format(tensor.ndim)
+    if mem_fmt is not None and not tensor.is_contiguous(memory_format=mem_fmt):
+        return tensor.contiguous(memory_format=mem_fmt)
+    return tensor
+
+
 def max_pool(
     inputs,
     pool_size,
@@ -566,6 +581,10 @@ def conv(
 
     kernel = _transpose_conv_kernel(kernel)
 
+    if data_format == "channels_last":
+        inputs = _maybe_convert_to_channels_last(inputs)
+        kernel = _maybe_convert_to_channels_last(kernel)
+
     # calc. groups snippet
     in_channels = inputs.shape[1]
     kernel_in_channels = kernel.shape[1]
@@ -701,6 +720,11 @@ def conv_transpose(
         inputs = _transpose_spatial_inputs(inputs)
     # Transpose kernel from keras format to torch format.
     kernel = _transpose_conv_kernel(kernel)
+
+    if data_format == "channels_last":
+        inputs = _maybe_convert_to_channels_last(inputs)
+        kernel = _maybe_convert_to_channels_last(kernel)
+
     kernel_spatial_shape = kernel.shape[2:]
     if isinstance(dilation_rate, int):
         dilation_rate = [dilation_rate] * len(kernel_spatial_shape)


hertschuh (commented):
> Ensures conv and conv_transpose tensors use PyTorch's channels_last memory format when data_format="channels_last", allowing PyTorch to route to cuDNN's native NHWC kernels instead of internally copying tensors.

How do you know that it's now using cuDNN's native NHWC kernels?

It is faster? Where you able to benchmark before and after?

MarcosAsh:
How do you know that it's now using cuDNN's native NHWC kernels?  I ran this code to check:
                                                                                                                                                
  ```python                                                                                                                                     
  >>> x = torch.randn(4, 64, 56, 56, device="cuda").contiguous(memory_format=torch.channels_last)
  >>> w = torch.randn(128, 64, 3, 3, device="cuda").contiguous(memory_format=torch.channels_last)
  >>> out = F.conv2d(x, w, padding=1)
  >>> out.is_contiguous(memory_format=torch.channels_last)
 ```
Which gave an output of:
  True

  
Is it faster? Were you able to benchmark before and after? Here's what I used to test it on an A100 (on Google Colab) reproduces the same permute-to-NCHW flow Keras does, with and without the channels_last memory format conversion:
 ```python
  import time
  import torch
  import torch.nn.functional as F

  def bench_raw(batch, c_in, h, w, c_out, ks, channels_last_mem=False,
                warmup=20, iters=100):
      x_nhwc = torch.randn(batch, h, w, c_in, device="cuda")
      k_keras = torch.randn(ks, ks, c_in, c_out, device="cuda")

      x = x_nhwc.permute(0, 3, 1, 2)
      w = k_keras.permute(3, 2, 0, 1)

      if channels_last_mem:
          x = x.contiguous(memory_format=torch.channels_last)
          w = w.contiguous(memory_format=torch.channels_last)

      torch.cuda.synchronize()
      for _ in range(warmup):
          F.conv2d(x, w, padding=1).permute(0, 2, 3, 1)
      torch.cuda.synchronize()

      times = []
      for _ in range(iters):
          torch.cuda.synchronize()
          t0 = time.perf_counter()
          F.conv2d(x, w, padding=1).permute(0, 2, 3, 1)
          torch.cuda.synchronize()
          times.append(time.perf_counter() - t0)

      return sorted(times)[len(times) // 2] * 1000
 ```

  Results:

  config                           before      after  speedup
  3->64 @ 224x224                 0.410ms    0.388ms    1.06x
  64->128 @ 56x56                 0.239ms    0.223ms    1.08x
  128->256 @ 28x28                0.206ms    0.183ms    1.13x
  256->512 @ 14x14                0.214ms    0.195ms    1.09x
  3->64 @ 224x224 k=7             0.583ms    0.562ms    1.04x
  64->64 @ 56x56 bs=64            0.297ms    0.276ms    1.07x

 So a 4-13% improvement at the op level on A100. On T4 the gains are smaller (~1-2%).

One thing I noticed, the current approach permutes NHWC ->NCHW and then .contiguous(memory_format=channels_last) copies the data back into NHWC layout, so we're kind of doing a round trip. A potentially better approach would be to skip the permute entirely for conv ops and pass the NHWC tensor directly with channels_last memory format, which would avoid the extra copy. Would you prefer I rework it that way instead?

I understand this current approach is borderline not useful at all so no worries if you want to close the pr as to not waste time on this.

hertschuh:
> A potentially better approach would be to skip the permute entirely for conv ops and pass the NHWC tensor directly with channels_last memory format, which would avoid the extra copy.

I believe this what the bug intended to do.

I still submitted this as a first step. But would you be able to provide the deeper fix as a separate PR?

MarcosAsh:
Yes of course! On it

MarcosAsh:
Hey I have a few concerns. I dug into this a bit. tnn.pad already preserves channels_last in current PyTorch, so the format doesn't get broken by same-padding. The main overhead left is the kernel copy to channels_last on every forward pass. I looked at using torch.ops.aten.cudnn_convolution to skip dispatch overhead, but it's CUDA-only and the gains seem marginal. Did you have a specific approach in mind for the deeper fix?

PR #22327, opened 2026-03-01
fix: add oov_method parameter to IntegerLookup for uniform OOV bucketing

### LINKED ISSUE #22325: OOV bucketing uses `floormod` for integers in `IntegerLookup`, inconsistent with `Hashing`
## Context

When migrating from `Hashing` to `IntegerLookup` to avoid hash collisions on known-vocabulary categorical features, I noticed that OOV bucketing behaves differently across preprocessing layers for integer inputs. The inconsistency has real impact on OOV token distribution quality, as it uses a simple `floormod` (modulo function) instead of FarmHash/SipHash encodings used in Hashing. Therefore, there is a loss of consistency with the Hashing layer, and losing the uniform distribution properties of a hash function (avalanche effect).

---

## Current behavior

In `index_lookup.py`, `_lookup_dense` handles multi-bucket OOV assignment as follows:

```python
elif self.num_oov_indices > 1:
    if tf.as_dtype(self._key_dtype).is_integer:
        oov_indices = tf.math.floormod(inputs, self.num_oov_indices)  # ← floormod
    else:
        oov_indices = tf.strings.to_hash_bucket_fast(                 # ← FarmHash64
            inputs, num_buckets=self.num_oov_indices
        )
```

| Layer | Input dtype | OOV bucketing method |
|---|---|---|
| `IntegerLookup` | integer | `floormod(token, num_oov_indices)` |
| `StringLookup` | string | `tf.strings.to_hash_bucket_fast` (FarmHash64) |
| `Hashing` | any | FarmHash64 or SipHash64 if salt provided |

The inconsistency is **specific to `IntegerLookup`**. As for using FarmHash64, `StringLookup` already behaves consistently with `Hashing` : IndexLookup applies FarmHash only on strings.

`floormod` is problematic because distribution depends entirely on the arithmetic structure of the input. OOV IDs drawn from a single mod-class (e.g. all even numbers, stride-aligned user/product IDs) always map to the same bucket, creating severe imbalance:

```python
import tensorflow as tf

layer = tf.keras.layers.IntegerLookup(
    vocabulary=[10, 20, 30],
    num_oov_indices=4,
)

# All even OOV values → always bucket 0, never buckets 1, 2, or 3
oov_values = [100, 200, 300, 400, 1000, 2000]
print(layer(oov_values))  # All map to OOV index 0
```

A hash function like FarmHash64 distributes uniformly regardless of arithmetic structure.

---

## Proposed fix

Add an `oov_method` string parameter to `IntegerLookup` only, with choice between "floormod" (default) and "farmhash". If FarmHash is chosen, similarly to `Hashing`, values are converted to string before being passed to `tf.strings.to_hash_bucket_fast`.
 `StringLookup` already uses FarmHash64 and needs no change. `floormod` is kept as the default for full backwards compatibility.

- **Fully backwards-compatible** — default is `oov_method="floormod"`, existing behavior is unchanged
- **Mirrors `Hashing` layer API** — consistent hashing behavior across all three layers
- **No new dependencies** — change isolated to `IndexLookup._lookup_dense`, `IntegerLookup.__init__`

---

## Files affected

- `keras/src/layers/preprocessing/index_lookup.py` — core logic, `__init__`, `get_config` 
- `keras/src/layers/preprocessing/integer_lookup.py` — expose new param, update docstring

`string_lookup.py` requires no changes for this specific issue.

---

## Future work

We also notice that StringLookup only uses FarmHash64. To be completely consistent with Hashing, it would require to use SipHash64 as well, thus have a salt parameter. Same would go for IntegerLookup once this issue is done.
Thus, adding a `salt` parameter to enable SipHash (mirroring `Hashing(salt=...)`), and extending this to `StringLookup`, would be a natural follow-up. I'll track it in a separate issue.

---

## Contributing

I have a working implementation ready and am happy to open a PR. This is my first contribution to Keras, so any feedback on the design is very welcome.

tanguyguyot:
## Summary
- Adds an oov_method parameter ("floormod" | "farmhash") to IntegerLookup and the base IndexLookup class, allowing users to opt into FarmHash64-based OOV bucketing for integer inputs instead of the default floormod behavior, in the same way as Hashing layer does.
- Backwards compatibility : Default is oov_method="floormod", preserving existing behavior exactly.
- Tests files added

## Testing
Verified  pytest keras/src/layers/preprocessing ran locally

Fixes #22325

### DIFF
--- a/keras/src/layers/preprocessing/index_lookup.py
+++ b/keras/src/layers/preprocessing/index_lookup.py
@@ -87,8 +87,8 @@ class IndexLookup(Layer):
             If `True`, returns a `SparseTensor` instead of a dense `Tensor`.
             Defaults to `False`.
         oov_method: Only relevant when `num_oov_indices > 1` and the input
-            dtype is integer (i.e. for `IntegerLookup`). Controls how OOV
-            tokens are assigned to OOV buckets.
+            dtype is integer (i.e. for `IntegerLookup`). Controls how
+            Out-Of-Vocabulary (OOV) tokens are assigned to OOV buckets.
             - `"floormod"` (default): uses `token % num_oov_indices`.
               Preserves backwards compatibility but can produce severe bucket
               imbalance when input IDs share a common factor with


keras/src/layers/preprocessing/index_lookup.py
```diff
@@ -86,6 +86,17 @@ class IndexLookup(Layer):
             `"count"` and `"tf-idf"` output modes.
             If `True`, returns a `SparseTensor` instead of a dense `Tensor`.
             Defaults to `False`.
+        oov_method: Only relevant when `num_oov_indices > 1` and the input
+            dtype is integer (i.e. for `IntegerLookup`). Controls how OOV
```
jeffcarp: Style nit: maybe expand OOV to Out of Vocabulary the first time its used

jeffcarp:
Thanks! Lgtm with one style nit

keras/src/layers/preprocessing/index_lookup.py
tanguyguyot: done!

PR #22328, opened 2026-03-02
Fix RandomCrop doing nothing at validation/inference

### LINKED ISSUE #21868: RandomCrop does nothing at validation/inference

rstar327:
The `RandomCrop` layer's `transform_images` method was guarded by `if training:`, causing it to return the input image unchanged during inference (`training=False`). This meant the layer produced no cropping at all during validation/prediction, leading to shape mismatches when the model expects cropped dimensions.

The `get_random_transformation` method already correctly computes center crop coordinates for `training=False`, but those coordinates were never applied.

This fix removes the `training` guard from `transform_images` (and `transform_bounding_boxes`), so the crop is always applied — random during training, center during inference.

Fixes #21868

### DIFF
--- a/keras/src/layers/preprocessing/image_preprocessing/random_crop.py
+++ b/keras/src/layers/preprocessing/image_preprocessing/random_crop.py
@@ -129,60 +129,59 @@ class RandomCrop(BaseImagePreprocessingLayer):
         return h_start, w_start
 
     def transform_images(self, images, transformation, training=True):
-        if training:
-            images = self.backend.cast(images, self.compute_dtype)
-            crop_box_hstart, crop_box_wstart = transformation
-            crop_height = self.height
-            crop_width = self.width
+        images = self.backend.cast(images, self.compute_dtype)
+        crop_box_hstart, crop_box_wstart = transformation
+        crop_height = self.height
+        crop_width = self.width
 
-            if self.data_format == "channels_last":
-                if len(images.shape) == 4:
-                    images = images[
-                        :,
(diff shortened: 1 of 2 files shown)


hertschuh:
@rstar327 

Sorry, the conclusion from the thread in the bug is that it is working as intended. So we don't want to change this behavior.

PR #22329, opened 2026-03-02
Fix associative_scan infinite recursion when axis size is 0 or 1

### LINKED ISSUE #22050: keras.src.backend.tensorflow.core.associative_scan enters infinite recursion when scan axis size is 0 or 1
I found a potential bug in `keras.src.backend.tensorflow.core.associative_scan`. When the size of the dimension corresponding to the `axis` parameter is 0 or 1, the function enters infinite recursion.

Here is a reproducible example that demonstrates the bug:

```python
import os
import keras

sum_fn = lambda x, y: [x[0] + y[0], x[1] + y[1], x[2] + y[2]]
xs = [keras.ops.array([[1, 2]]) for _ in range(3)]
ys = keras.ops.associative_scan(sum_fn, xs, axis=0)
print(ys)
```

This results in a RecursionError: maximum recursion depth exceeded, while other backends correctly output [[[1, 2]], [[1, 2]], [[1, 2]]].

The root cause is in​ the recursion termination condition:
```python
at_base_case = tf.logical_or(
    tf.equal(elem_length, 2), tf.equal(elem_length, 3)
)
```
This condition fails to handle cases where `elem_length` is 0 or 1. In both scenarios, after the slicing operations and function `_combine`, `elem_length` becomes 0. This causes the function to enter the `_recursive_case` branch, which then calls `_scan` again with `elem_length == 0`, resulting in infinite recursion.

rstar327:
The TensorFlow backend's `associative_scan` implementation enters infinite recursion when the scan axis has size 0 or 1. The existing base case check only handles `elem_length == 2` and `elem_length == 3`, so smaller sizes fall through to the recursive case indefinitely.

This adds an early return for `elem_length <= 1` — no scanning is needed when there are zero or one elements along the axis.

Fixes #22050

### DIFF
--- a/keras/src/backend/tensorflow/core.py
+++ b/keras/src/backend/tensorflow/core.py
@@ -479,6 +479,11 @@ def associative_scan(f, elems, reverse=False, axis=0):
 
     def _scan(elems):
         elem_length = _get_dim(elems[0])
+        # Handle trivial cases: 0 or 1 elements need no scanning.
+        # Without this guard, the recursion never reaches the base
+        # cases (elem_length 2 or 3) and overflows the stack.
+        if isinstance(elem_length, int) and elem_length <= 1:
+            return elems
         a = [slice_along_axis(elem, 0, -1, step=2, axis=axis) for elem in elems]
         b = [
             slice_along_axis(elem, 1, None, step=2, axis=axis) for elem in elems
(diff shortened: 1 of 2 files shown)


hertschuh:
This was addressed by https://github.com/keras-team/keras/pull/22072

PR #22332, opened 2026-03-03
[Orbax Export Integration] Introduce Orbax Export dependency and create OrbaxExportArchive class

jerryxyj:
This is to introduce Orbax Export dependency to Keras, and create a new OrbaxExportArchive class for Keras, so it can use this to export TF model or Jax model by either SavedModelArchive or OrbaxExportArchive.
@hertschuh

### DIFF
--- a/keras/src/backend/jax/export.py
+++ b/keras/src/backend/jax/export.py
@@ -6,12 +6,11 @@ import warnings
 
 from keras.src import tree
 from keras.src.backend.common.stateless_scope import StatelessScope
-from keras.src.export.orbax_export_archive import OrbaxExportArchive
 from keras.src.export.saved_model_export_archive import SavedModelExportArchive
 from keras.src.utils.module_utils import tensorflow as tf
 
 
-class JaxExportArchive(SavedModelExportArchive, OrbaxExportArchive):
+class JaxExportArchive(SavedModelExportArchive):
     """JAX backend implementation of SavedModel export archive."""
 
     def _backend_init(self):
--- a/keras/src/backend/tensorflow/export.py
+++ b/keras/src/backend/tensorflow/export.py
@@ -1,10 +1,9 @@
 import tensorflow as tf
 
-from keras.src.export.orbax_export_archive import OrbaxExportArchive
 from keras.src.export.saved_model_export_archive import SavedModelExportArchive
 
 
-class TFExportArchive(SavedModelExportArchive, OrbaxExportArchive):
+class TFExportArchive(SavedModelExportArchive):
     """TensorFlow backend implementation of SavedModel export archive."""
 
     def _backend_track_layer(self, layer):
--- a/keras/src/utils/module_utils.py
+++ b/keras/src/utils/module_utils.py
@@ -58,25 +58,6 @@ class OrbaxLazyModule(LazyModule):
         return getattr(self.module, name)
 
 
-class OrbaxExportLazyModule(LazyModule):
-    def initialize(self):
-        try:
-            parent_module = importlib.import_module("orbax.export")
-            self.module = parent_module.v1
-            self.parent_module = parent_module
-        except ImportError:
-            raise ImportError(self.import_error_msg)
-
-    def __getattr__(self, name):
-        if name == "_api_checkpoint_path":
-            raise AttributeError
-        if self.module is None:
-            self.initialize()
-        if name == "multihost":
-            return self.parent_module.multihost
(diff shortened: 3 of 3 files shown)


keras/src/backend/jax/export.py
```diff
@@ -6,11 +6,12 @@
 
 from keras.src import tree
 from keras.src.backend.common.stateless_scope import StatelessScope
+from keras.src.export.orbax_export_archive import OrbaxExportArchive
 from keras.src.export.saved_model_export_archive import SavedModelExportArchive
 from keras.src.utils.module_utils import tensorflow as tf
 
 
-class JaxExportArchive(SavedModelExportArchive):
+class JaxExportArchive(SavedModelExportArchive, OrbaxExportArchive):
```
hertschuh: I don't think we should do that:
1. The `JaxExportArchive` should not change.
2. If `OrbaxExportArchive` was actually used, the unit tests would fail with `NotImplementedError`, so this proves that adding this inheritance doesn't do anything (that's because all the methods are implemented in `SavedModelExportArchive` and take precedence).

keras/src/backend/tensorflow/export.py
```diff
@@ -1,9 +1,10 @@
 import tensorflow as tf
 
+from keras.src.export.orbax_export_archive import OrbaxExportArchive
 from keras.src.export.saved_model_export_archive import SavedModelExportArchive
 
 
-class TFExportArchive(SavedModelExportArchive):
+class TFExportArchive(SavedModelExportArchive, OrbaxExportArchive):
```
hertschuh: I don't think we should do that:
1. The `TFExportArchive` should not change.
2. I though Orbax Export was not for pure TF use cases.
3. If `OrbaxExportArchive` was actually used, the unit tests would fail with `NotImplementedError`, so this proves that adding this inheritance doesn't do anything (that's because all the methods are implemented in `SavedModelExportArchive` and take precedence).

keras/src/utils/module_utils.py
```diff
@@ -88,3 +107,11 @@ def __getattr__(self, name):
...
     ),
 )
+oex = OrbaxExportLazyModule(
+    "orbax.export.v1",
+    pip_name="orbax-export",
+    import_error_msg=(
+        "OrbaxExport requires the 'orbax-export' package. "
+        "You can install it via pip install orbax-export"
+    ),
+)
```
hertschuh: Let's call it `orbax_export` instead of `oex`, it should be as close as the pip package name or module name as possible.

keras/src/utils/module_utils.py
```diff
@@ -58,6 +58,25 @@ def __getattr__(self, name):
...
+            raise ImportError(self.import_error_msg)
+
+    def __getattr__(self, name):
+        if name == "_api_checkpoint_path":
+            raise AttributeError
+        if self.module is None:
+            self.initialize()
+        if name == "multihost":
+            return self.parent_module.multihost
+        return getattr(self.module, name)
```
hertschuh: None of this logic that was done for Orbax Checkpoint makes sense for Orbax Export. This won't work.

You just need this (this would be instead of lines 110-117):
```
orbax_export = LazyModule("orbax.export", "orbax-export")
```

keras/src/export/orbax_export_archive.py
```diff
@@ -0,0 +1,24 @@
...
+        raise NotImplementedError()
+
+    def track_and_add_endpoint(self, name, resource, input_signature, **kwargs):
+        raise NotImplementedError()
+
+    def add_variable_collection(self, name, variables):
+        raise NotImplementedError()
+
+    def write_out(self, filepath, options=None, verbose=True):
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
jerryxyj: this class is no one use now, so it should be safe

keras/src/export/saved_model.py
```diff
@@ -179,9 +180,7 @@ def __new__(cls, format="saved_model", **kwargs):
             if export_model == "backend_saved_model":
                 return BackendSavedModelExportArchive()
             elif export_model == "orbax_export":
-                raise NotImplementedError(
-                    "Orbax ExportArchive is not supported in Keras 3 yet."
-                )
+                return OrbaxExportArchive()
```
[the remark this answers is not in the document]
jerryxyj: ditto

jerryxyj:
Addressed all the comments

PR #22342, opened 2026-03-04
Revert "[OpenVINO backend] Fix array(), maximum(), minimum(), reshape() and split() ops in the numpy suite"

goyaladitya05:
Reverts keras-team/keras#22319


goyaladitya05:
I'm not addressing the gemini review comments here, since it's a revert. I'll push the changes directly in the main PR.

Issue #22307, opened 2026-02-27
[Feature] Modernize benchmark infrastructure for multi-backend comparison


MarcosAsh:
## Context

The `benchmarks/` directory was originally built to compare Keras 3 against `tf.keras`, which made sense at the time. But now that JAX, PyTorch, and TF backends are all properly supported, the more useful question is how does the same Keras code actually perform across backends?

The [development roadmap (#19519)](https://github.com/keras-team/keras/issues/19519) lists "Official performance benchmarks" as a goal, so I wanted to gauge what that could look like and offer to help.

The main gap is that the benchmarks only compare Keras 3 vs `tf.keras`. `LayerBenchmark` hardcodes both a `keras.layers.X` and a `tf.keras.layers.X` instance and runs them side by side. There is no way to benchmark the same layer across JAX, Torch, and TF backends, which is probably what most users actually want to know.

A few other things I noticed while looking through it:

- Results just go to `print()` with no JSON or CSV, so there is no easy way to compare runs or track trends over time
- Benchmarks are not run in `actions.yml` or `nightly.yml`, so performance regressions only get caught when someone files a bug
- Only throughput is measured with no memory tracking, which actually matters a lot for RNN layers and loss functions (I ran into this myself while working on #22169)
- The warmup is just skipping batch 0, which is not enough to account for JIT compilation noise on JAX or PyTorch

## What I'd propose

Rather than one big PR, I'd tackle this incrementally:

1. **Multi-backend `LayerBenchmark`** - refactor the base class to accept a `--backend jax,torch,tensorflow` flag and produce a comparison table
2. **Structured output** - add a `--output_format json` flag so results include metadata (backend, hardware, shapes, timestamps)
3. **Memory profiling** - track peak memory alongside throughput using `torch.cuda.max_memory_allocated`, JAX device memory stats, etc.
4. **Better measurement** - configurable warmup, multiple runs, mean/std reporting
5. **CI integration** - a nightly workflow that runs benchmarks and stores results as artifacts; stretch goal would be a perf diff comment on PRs

## Where I'd start

The first PR would just be Phase 1 refactor `LayerBenchmark` to support multi-backend runs, update `conv_benchmark.py` as a reference implementation, and keep the existing Keras-vs-tf.keras path working so nothing breaks.

[...]

PR #22346, opened 2026-03-04
Add multi-backend benchmark comparison support

MarcosAsh:
Refactors LayerBenchmark to support running the same benchmark across jax, torch, and tensorflow backends via a --backend flag. Each backend runs in its own subprocess and results are displayed as a comparison table (or JSON with --output_format=json). The existing Keras 3 vs tf.keras default is preserved.

Phase 1 of #22307. Updated conv_benchmark.py as a reference implementation.

### DIFF
--- a/benchmarks/layer_benchmark/conv_benchmark.py
+++ b/benchmarks/layer_benchmark/conv_benchmark.py
@@ -323,7 +323,7 @@ def main(_):
     jit_compile = FLAGS.jit_compile
 
     if benchmark_name is None:
-        for name, benchmark_fn in BENCHMARK_NAMES:
+        for name, benchmark_fn in BENCHMARK_NAMES.items():
             benchmark_fn(num_samples, batch_size, jit_compile)
         return
 
--- a/benchmarks/layer_benchmark/base_benchmark.py
+++ b/benchmarks/layer_benchmark/base_benchmark.py
@@ -1,12 +1,28 @@
+import json
+import os
 import time
 
 import numpy as np
-import tensorflow as tf
 from absl import flags
 
 import keras
 
 FLAGS = flags.FLAGS
+
+# When set, benchmark_predict/benchmark_train emit a JSON line that the
+# multi-backend shell script can parse.
+_RESULT_JSON_ENV = "_BENCHMARK_RESULT_JSON"
+
+# Try importing tf.keras for legacy comparison.  When running under a
+# non-TF backend the import may fail, which is fine -- we just skip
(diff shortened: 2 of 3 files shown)


benchmarks/layer_benchmark/conv_benchmark.py
```diff
@@ -317,13 +339,20 @@ def benchmark_conv3D_transpose(
 
 
 def main(_):
+    if FLAGS.backend:
+        run_multi_backend_benchmark(
+            module_path="benchmarks.layer_benchmark.conv_benchmark",
+            benchmark_fn_map=BENCHMARK_NAMES,
+        )
+        return
```
jeffcarp: I'm wondering if a simpler approach is to call this script multiple times with different backends set via `KERAS_BACKEND`, instead of adding the multi-backend support in the script. There could be a separate bash script that invokes this script for each backend. That would also better isolate memory and prevent any cross-contamination while switching between backends in the same process.

jeffcarp (commented):
Thanks! Added an idea

benchmarks/layer_benchmark/conv_benchmark.py
MarcosAsh: Thanks for the review! Each backend runs in its own subprocess via subprocess.run() with KERAS_BACKEND set in the env (lines 428-445 in base_benchmark.py), so there's no shared state between backends.

The main reason I went with Python was to handle the result aggregation. Each subprocess outputs a JSON line that the parent collects and formats into the comparison table or JSON output. I didnt think of making it in bash but it would make it dependable on bash. Even tough it would be easier to run with a bash script.  

Happy to rethink the approach if you'd prefer something different though!

MarcosAsh:
Hey @jeffcarp, just checking in. Happy to rework this if you'd prefer a standalone shell script for the multi-backend orchestration. Let me know which way you'd like me to go and I'll update it.

MarcosAsh:
Reworked this based on your suggestion. Moved the multi-backend orchestration into a standalone shell script (`benchmarks/run_multi_backend.sh`) instead of baking it into the Python code. Each backend gets its own subprocess with `KERAS_BACKEND` set, and the script collects JSON output and prints a comparison table.

The Python changes are pretty small. Conditional TF import so benchmarks can run on jax/torch, and a `_emit_result()` helper that prints a JSON line when the script sets `_BENCHMARK_RESULT_JSON=1`. Also fixed a pre-existing `.items()` bug in `conv_benchmark.py` while I was in there.

PR #22353, opened 2026-03-05
Re-enable one step ahead `device_put` of data from datasets on JAX.

hertschuh:
The `_prefectch_numpy_iterator` feature in the JAX trainer had been disabled based on the idea that many dataset implementations already support prefectching. However, this came from a misundestanding of what `_prefetch_numpy_iterator` does, probably because of the misleading name.

`_prefetch_numpy_iterator` triggers `device_put` for all the arrays of a batch of data to transfer arrays from CPU to the accelerator (GPU or TPU). `device_put` is asynchronous. By triggering `device_put` one step ahead. we parallelize the transfer of the next batch of data with the running of the model with the current batch of data. Without this, the batch of data is immediately used after calling `device_put`, which causes a wait until `device_put` is complete.

Prefetching happens only on CPU and is independent.

- Removed the `builtin_prefetch` property, which now does not serve any purpose anymore.
- Renamed `_prefetch_numpy_iterator` to `_one_batch_ahead_iterator`, which more accurately describes its purpose.

### DIFF
--- a/keras/src/backend/jax/trainer.py
+++ b/keras/src/backend/jax/trainer.py
@@ -1,4 +1,3 @@
-import collections
 import itertools
 import warnings
 from functools import partial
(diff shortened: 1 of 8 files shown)


hertschuh:
It turns out this deadlocks with grain. Looking into it.

jeffcarp (approved):

PR #22356, opened 2026-03-05
Fix #19127 — Improve error message for named inputs mismatch in `Functional` model

### LINKED ISSUE #19127: Incorrect error message given for named inputs when unnamed input provided to fit()
While trying to author a subclass of PyDataset, I found that `__getitem__()` must return a dictionary with named values, otherwise an unrelated error is given.  In hind sight, this is obvious, but the error message led to hours of wasted debugging in the wrong places.

Attached is a small patch that demonstrates the problem on latest master:
[demonstrate-named-inputs-without-warning.patch.txt](https://github.com/keras-team/keras/files/14118247/demonstrate-named-inputs-without-warning.patch.txt)

The error looks like this:
```
  File "<path>/keras/integration_tests/./tf_distribute_training_test.py", line 90, in <module>
    test_model_fit()
  File "<path>/keras/integration_tests/./tf_distribute_training_test.py", line 74, in test_model_fit
    history = model.fit(dataset, epochs=epochs, callbacks=callbacks)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<path>/keras/env/lib/python3.11/site-packages/keras/src/utils/traceback_utils.py", line 123, in error_handler
  File "/tmp/__autograph_generated_fileczxvauzh.py", line 14, in tf__wrapped_fn
    retval_ = ag__.converted_call(ag__.ld(call_for_each_replica), (ag__.ld(strategy), ag__.ld(fn).python_function, ag__.ld(args), ag__.ld(kwargs)), None, fscope)
    ^^^^^
ValueError: in user code:

    File "<path>/keras/env/lib/python3.11/site-packages/keras/src/backend/tensorflow/trainer.py", line 105, in one_step_on_data  **
        return self.train_step(data)
    File "<path>/keras/env/lib/python3.11/site-packages/keras/src/backend/tensorflow/trainer.py", line 59, in train_step
        loss = self.compute_loss(
    File "<path>/keras/env/lib/python3.11/site-packages/keras/src/trainers/trainer.py", line 327, in compute_loss
        raise ValueError(

    ValueError: No loss to compute. Provide a `loss` argument in `compile()`.
```

`loss` is clearly provided in `compile()` which leads to a wild goose chase.

If you remove `name="inputs"` such as what you see in the commented out line, you get a related warning that is similar to what you would expect for this case:
```
<path>/keras/env/lib/python3.11/site-packages/keras/src/models/functional.py:104: UserWarning: When providing `inputs` as a dict, all keys in the dict must match the names of the corresponding tensors. Received key 'inputs' mapping to value <KerasTensor shape=(32, 100
), dtype=float32, sparse=None, name=keras_tensor> which has name 'keras_tensor'. Change the tensor name to 'inputs' (via `Input(..., name='inputs')`)
```

A warning like that could be given when the dataset provided to fit() contains unnamed inputs/outputs.

pctablet505:
**Fixes**: https://github.com/keras-team/keras/issues/19127  
This pull request improves the input validation logic for models with dictionary-based inputs in the Keras Functional API. The main change is to provide clearer error messages when users supply inputs in the wrong format, making it easier to diagnose and fix input-related issues. The corresponding test has also been updated to match the new error message.

Input validation improvements:

* Updated the `_assert_input_compatibility` method in `Model` (in `functional.py`) to check for cases where the model expects a dictionary of inputs but receives a non-dictionary (like a list, tuple, or array). If a list or tuple of matching length is provided, it falls back to positional matching; otherwise, it raises a clearer `ValueError` with guidance on the expected input format.

Testing updates:

* Modified the test `test_bad_input_spec` in `functional_test.py` to expect the new, more descriptive error message when a non-dictionary input is passed to a model expecting a dictionary.

#### Problem

When a `Functional` model is built with dict inputs (e.g. `keras.Input({"a": ..., "b": ...})`), passing a plain array or list to `model.fit()` / `model(...)` raised a confusing error: `"expects 2 input(s)"`. This message gave no hint that the model requires named dict inputs, making it very hard for users to diagnose the problem (see thread in #19127).

#### Root Cause

`Functional._assert_input_compatibility` unconditionally delegated to the parent `Model` implementation, which only checks the input count. It had no awareness that its `_inputs_struct` is a `dict`, so the error path never mentioned dict/named inputs at all.

#### Fix

Override `_assert_input_compatibility` in `Functional` to detect the mismatch early. If the model's input struct is a `dict` but the user passes a non-dict:
- A list/tuple of the correct length is still accepted (positional matching, same as before).
- Anything else raises a clear `ValueError` that names the expected keys and shows the correct calling syntax:
  ```
  Model "my_model" expects inputs as a `dict` with the following keys: ['a', 'b'].
  Instead received ndarray. Pass your data as `model.fit({'a': ..., 'b': ...}, ...)`.
  ```

#### Files Changed
- `keras/src/models/functional.py` — override `_assert_input_compatibility` with dict-aware check
- `keras/src/models/functional_test.py` — update test to match new error text

### DIFF
--- a/keras/src/models/functional.py
+++ b/keras/src/models/functional.py
@@ -213,8 +213,26 @@ class Functional(Function, Model):
             return output_shapes[0]
         return output_shapes
 
-    def _assert_input_compatibility(self, *args):
-        return super(Model, self)._assert_input_compatibility(*args)
+    def _assert_input_compatibility(self, inputs):
+        if isinstance(self._inputs_struct, dict) and not isinstance(
+            inputs, dict
+        ):
+            # Allow list/tuple with matching length (positional matching)
+            if isinstance(inputs, (list, tuple)):
+                if len(inputs) == len(self._inputs_struct):
+                    return super(
+                        Model, self
+                    )._assert_input_compatibility(inputs)
+            keys = list(self._inputs_struct.keys())
+            raise ValueError(
+                f'Model "{self.name}" expects inputs as a `dict` with '
+                f"the following keys: {keys}. Instead received "
+                f"{type(inputs).__name__}. Pass your data as "
+                "`model.fit({" + ", ".join(
+                    f"'{k}': ..." for k in keys
+                ) + "}, ...)`."
+            )
+        return super(Model, self)._assert_input_compatibility(inputs)
 
     def _maybe_warn_inputs_struct_mismatch(self, inputs, raise_exception=False):
         try:
--- a/keras/src/models/functional.py
+++ b/keras/src/models/functional.py
@@ -220,17 +220,17 @@ class Functional(Function, Model):
             # Allow list/tuple with matching length (positional matching)
             if isinstance(inputs, (list, tuple)):
                 if len(inputs) == len(self._inputs_struct):
-                    return super(
-                        Model, self
-                    )._assert_input_compatibility(inputs)
+                    return super(Model, self)._assert_input_compatibility(
+                        inputs
+                    )
             keys = list(self._inputs_struct.keys())
             raise ValueError(
                 f'Model "{self.name}" expects inputs as a `dict` with '
                 f"the following keys: {keys}. Instead received "
                 f"{type(inputs).__name__}. Pass your data as "
-                "`model.fit({" + ", ".join(
-                    f"'{k}': ..." for k in keys
-                ) + "}, ...)`."
+                "`model.fit({"
+                + ", ".join(f"'{k}': ..." for k in keys)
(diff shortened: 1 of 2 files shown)


hertschuh (changes requested):
First, this is already done right below in `_maybe_warn_inputs_struct_mismatch`. So maybe we just need to change the warning to a ValueError.

Second, this is hardcoded for dict inputs when `Functional` supports arbitrarily nested inputs.

PR #22357, opened 2026-03-05
Fix TensorBoard callback step counter never updating

### LINKED ISSUE #20143: TensorBoard Callback never updates internal step counter
Using the TensorBoard callback to generate time series data is currently not possible, since the internal `self._train_step` counter is never updated. Instead, looking at the time series tab only ever yields a single data point instead of a graph. This is caused by the fact that every call to `self.summary.scalar(...)` is given the same step value.

The simplest fix that comes to my mind is adding this line to `keras.callbacks.TensorBoard.on_train_batch_end()`, but it is likely that a better fix exists that I am not aware of:

```py3
def on_train_batch_end(...):
    ...
    self._train_step += 1
```

Below, you may find what this fix looks like, running multiple steps (despite what the screenshot says, multiple steps are logged).

Without the fix:
Screenshot from 2024-08-21 17-22-13

With the fix:
Screenshot from 2024-08-21 17-16-08

pctablet505:
### Summary
This PR fixes an issue where the `TensorBoard` callback's step counter was not updating for batch-level summaries, causing user-defined custom summaries (e.g. `tf.summary` calls inside a custom layer's `call` method) to overwrite each other at step 0, and preventing the `update_freq` gating mechanism from working correctly.

### Problem
The `TensorBoard` callback was only setting the default summary writer context once during training/testing start (`on_train_begin` and `on_test_begin`), passing the initial batch counter value of `0` to `writer.as_default(step)`.
Because `as_default(step)` in TensorFlow captures the step value statically upon entry, and `record_if(should_record)` captures the `should_record` closure:
1. Custom batch-level summaries logged inside a model (such as `tf.summary.histogram` in `Model.call()`) that do not supply an explicit `step` parameter defaulted to `0` for all batches. As a result, they overwrote each other and only showed as a single data point at step 0 in TensorBoard.
2. The `should_record` callback closure was created with `step=0`. Thus, `0 % update_freq == 0` always evaluated to `True`, causing batch-level metrics to be logged on every single batch regardless of the `update_freq` value configured in the callback.

### Solution
Instead of using a `backend.Variable` for the step counter (which introduced cross-backend device transfer and typing issues, especially when converting tensor outcomes to Python booleans in JAX and PyTorch), this PR implements a clean, backend-agnostic solution by re-entering the writer context at the start of each batch:
* In `on_train_batch_begin` and `on_test_batch_begin`, if `update_freq != "epoch"`, we pop the current writer and push it again with the updated batch counter (`self._global_train_batch` / `self._global_test_batch`).
* This ensures that the active summary writer context's default step is always up-to-date. As a result:
  - Custom summaries written inside the model call dynamically resolve to the correct batch number.
  - The recording gate (`should_record()`) correctly checks the current batch number against the `update_freq` parameter.

Fixes #20143

- [x]   I am a human, and not a bot.
- [x]   I will be responsible for responding to review comments in a timely manner.
- [x]   I will work with the maintainers to push this PR forward until submission.


keras/src/callbacks/tensorboard.py
```diff
@@ -310,16 +310,16 @@ def _configure_embeddings(self):
...
         if self.update_freq == "epoch":
             return
 
         def should_record():
-            return step % self.update_freq == 0
+            return step_var % self.update_freq == 0
 
         summary_context = (
-            writer.as_default(step),
+            writer.as_default(step_var),
```
hertschuh: These changes are a no-op

keras/src/callbacks/tensorboard.py
```diff
@@ -417,7 +420,10 @@ def on_train_end(self, logs=None):
         self._close_writers()
 
     def on_test_begin(self, logs=None):
-        self._push_writer(self._val_writer, self._global_test_batch)
+        import tensorflow as tf
+
+        self._test_step_var = tf.Variable(0, dtype=tf.int64, trainable=False)
```
hertschuh: We shouldn't use a `tf.Variable`. This is supposed to work with other backends too.

hertschuh (changes requested):

pctablet505:
@gemini-code-assist review

keras/src/callbacks/tensorboard.py
```diff
@@ -310,16 +310,16 @@ def _configure_embeddings(self):
...
         if self.update_freq == "epoch":
             return
 
         def should_record():
-            return step % self.update_freq == 0
+            return step_fn() % self.update_freq == 0
 
         summary_context = (
-            writer.as_default(step),
+            writer.as_default(),
```
hertschuh: But now the step is not passed at all anymore.

hertschuh (changes requested):

hertschuh (approved):
Thanks!

pctablet505:
There were some changes due to failed GPU tests.

hertschuh (changes requested):
I see you changed the `tf.Variable` to `backend.Variable`, but we shouldn't do that. I don't understand how the tests are passing.

The TensorBoard stuff is based on TF and the `step` you pass to `as_default` should be an int or a `tf.Variable`, not a torch Param for instance.

So revert back to the `tf.Variable`, but force it to be on CPU:
```python
with tf.device("CPU:0"):
    self._train_step_var = tf.Variable(0, ...)
```

keras/src/callbacks/tensorboard.py
```diff
@@ -432,6 +432,9 @@ def on_test_end(self, logs=None):
 
     def on_train_batch_begin(self, batch, logs=None):
         self._global_train_batch += 1
+        if self.update_freq != "epoch":
+            self._pop_writer()
+            self._push_writer(self._train_writer, self._global_train_batch)
```
hertschuh, line 437: While it looks like it produces the right result, it completely defeats the purpose of `_push_writer` and `_pop_writer` since:
- you never reuse a writer
- you create a writer in `on_train_begin` but then discard it without using it

If you're going to go this route, then, instead of this:
- remove `def _push_writer` and `def _pop_writer`  completely
- instead of `def should_record():` do an `if` in `on_train_batch_end`
- instead do this in `on_train_batch_end`:
```python
with self._train_writer.as_default():
   ...
```

This will be a lot simpler and less code.

hertschuh (changes requested):

pctablet505:
@hertschuh Thanks for the review! 

I tried the simplification you suggested (removing `_push_writer/_pop_writer` and using with `writer.as_default()` in `on_train_batch_end`), but it breaks `test_custom_summary` and any user code that calls `tf.summary` inside `model.call()` — the writer context needs to be active during the forward pass, which happens before on_train_batch_end.

The re-push approach keeps `_push_writer/_pop_writer` intact for that purpose, while fixing the step counter bug by re-pushing with the updated step on each batch begin. The should_record closure now correctly sees the fresh step value instead of the stale 0.

All 15 TensorBoard tests pass. Could you please re-review?

hertschuh (approved):
Thanks for the simplication!

PR #22360, opened 2026-03-05
Fix #20322 — Weight save/load fails when model uses a custom subclass of a built-in layer (e.g. LSTM)

### LINKED ISSUE #20322: Loading weights into custom LSTM layer fails: Layer 'lstm_cell' expected 3 variables, but received 0 variables during loading. Expected: ['kernel', 'recurrent_kernel', 'bias']
I'm using the official TF 2.17 container (**tensorflow/tensorflow:2.17.0-gpu-jupyter**) + **keras==3.5.0**.

The following code saves a model which contains a (dummy) custom LSTM layer, then inits a new copy of the model (with a vanilla LSTM) and tries to load the weights from the first model into the second.

Code:

```
import os

os.environ["CUDA_VISIBLE_DEVICES"] = "-1"

import keras
from keras import layers

# An extremely uninteresting custom layer
@keras.saving.register_keras_serializable()
class MyCustomLSTM(keras.layers.LSTM):
    def __init__(self, units, **kwargs):
        super().__init__(units, **kwargs)

def make_model(
    use_custom_lstm=True,
):
    inputs = layers.Input(shape=(None, 4), name="inputs")

    if use_custom_lstm:
        lstm = MyCustomLSTM
    else:
        lstm = layers.LSTM

    outputs = lstm(
        units=8,
        return_sequences=True,
        name="my_LSTM",
    )(inputs)

    model = keras.models.Model(inputs=inputs, outputs=outputs)

    return model


weights_file = "this_is_a_test.weights.h5"
if os.path.exists(weights_file):
    os.remove(weights_file)

model = make_model(use_custom_lstm=True)
model.compile()
model.save_weights(weights_file)

new_model = make_model(use_custom_lstm=False)
new_model.load_weights(weights_file)

```

Output:

```
  File "scratch_1.py", line 45, in <module>
    new_model.load_weights(weights_file)
  File "venv/lib/python3.11/site-packages/keras/src/utils/traceback_utils.py", line 122, in error_handler
  File "venv/lib/python3.11/site-packages/keras/src/saving/saving_lib.py", line 593, in _raise_loading_failure
    raise ValueError(msg)
ValueError: A total of 1 objects could not be loaded. Example error message for object <LSTMCell name=lstm_cell, built=True>:

Layer 'lstm_cell' expected 3 variables, but received 0 variables during loading. Expected: ['kernel', 'recurrent_kernel', 'bias']

List of objects that could not be loaded:
[<LSTMCell name=lstm_cell, built=True>]
```

Considering that the custom layer in this case is doing absolutely nothing of interest, I assume this is a bug.  If not, please let me know how one is meant to wrap a LSTM layer to avoid this issue.

Thanks!

pctablet505:
**Fixes**: https://github.com/keras-team/keras/issues/20322

**Problem**

When saving weights, Keras walks the layer tree and stores each sublayer keyed by `to_snake_case(cls.__name__)`. If the saved model uses `class CustomLSTM(LSTM)`, the key written is `custom_lstm`. When loading into a model that uses the base `LSTM` class (same weights, same architecture), the key expected is `lstm` — mismatch, and weights are silently skipped.

The same problem occurs in reverse: saving with the base class and loading into a subclass.

**Root Cause**

`_save_container_state` and `_load_container_state` use the concrete class name (`custom_lstm`) instead of the base class name (`lstm`), so custom subclasses and their base classes get different paths in the weights file even when they are weight-compatible.

**Fix**

Add `_get_saveable_class_name(saveable)`: walk up the MRO to find the first class defined in `keras.layers` (or `keras.src.layers`) and use that class's snake_case name as the serialization key. For `class CustomLSTM(LSTM)`, this returns `lstm` — the same key used by the base `LSTM` class. This makes weight transfer between subclasses and base classes possible without changing the file format or breaking backward compatibility.

**Files Changed**
- `keras/src/saving/saving_lib.py` — `_get_saveable_class_name()`, updated `_save_container_state` and `_load_container_state`
- `keras/src/saving/saving_lib_test.py` — regression test covering both directions (custom→base and base→custom)

## Contributor Agreement


hertschuh (changes requested):
Can you look into why tests are failing?

hertschuh:
Unfortunately, we can't do this.

This will break everybody who ever saved custom sub-classes, they won't be able to reload anymore.

I'll reply in the bug.

PR #22363, opened 2026-03-05
[OpenVINO backend] Support for numpy.eigh 

andersendsa:
### Details :
Implement numpy.eigh operation for the OpenVINO backend.

### Closes issue : https://github.com/openvinotoolkit/openvino/issues/34520

### DIFF
--- a/keras/src/backend/openvino/linalg.py
+++ b/keras/src/backend/openvino/linalg.py
@@ -1,11 +1,10 @@
-import numpy as np
 import openvino.opset15 as ov_opset
-from openvino import Type, Model
+from openvino import Model
+from openvino import Type
 
 from keras.src.backend import config
 from keras.src.backend import standardize_dtype
 from keras.src.backend.common import dtypes
-from keras.src.backend.openvino.core import OPENVINO_DTYPES
 from keras.src.backend.openvino.core import OpenVINOKerasTensor
 from keras.src.backend.openvino.core import cast
 from keras.src.backend.openvino.core import convert_to_tensor
@@ -48,7 +47,6 @@ def eigh(a):
     a_ov = get_ov_output(a)
     a_ov_type = a_ov.get_element_type()
     if not a_ov_type.is_real():
-        
         a_ov = ov_opset.convert(a_ov, Type.f32).output(0)
         out_ov_type = Type.f32
     else:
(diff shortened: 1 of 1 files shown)


andersendsa:
hi @hertschuh  this pr passes all the tests and is ready for review

keras/src/backend/openvino/linalg.py
```diff
@@ -42,7 +42,22 @@ def eig(a):
...
+        raise ValueError(
+            "eigh is only supported for static eager tensors "
+            "in the openvino backend. Received a dynamic or symbolic tensor."
+        ) from e
+    
+    w, v = np.linalg.eigh(a_np)
+    return (
+        ov_opset.constant(w).output(0),
+        ov_opset.constant(v).output(0),
+    )
```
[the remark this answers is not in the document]
goyaladitya05: Please address the gemini reviews given here.

keras/src/backend/openvino/excluded_concrete_tests.txt
```diff
@@ -348,7 +348,6 @@ LayerTest::test_stateless_call
 LinalgOpsCorrectnessTest::test_cholesky
 LinalgOpsCorrectnessTest::test_det
 LinalgOpsCorrectnessTest::test_eig
```
goyaladitya05, line 350: Could we implement eig as well in this PR?

keras/src/backend/openvino/linalg.py
```diff
@@ -42,7 +42,24 @@ def eig(a):
...
 
 def eigh(a):
-    raise NotImplementedError("`eigh` is not supported with openvino backend")
+    import numpy as np
+
+    from keras.src.backend.openvino.core import convert_to_numpy
+
+    a = convert_to_tensor(a)
+    try:
+        a_np = convert_to_numpy(a)
```
goyaladitya05: We do the implementations using OpenVINO opsets specifically

andersendsa: will change it asap

keras/src/backend/openvino/excluded_concrete_tests.txt
andersendsa: sure will do it

andersendsa:
hi @hertschuh  i have done the changes and the pr passes all the tests and is ready for review

keras/src/backend/openvino/linalg.py
```diff
@@ -42,7 +44,33 @@ def eig(a):
...
+
+    a_evaluated = OpenVINOKerasTensor(a_ov)
+    try:
+        a_np = convert_to_numpy(a_evaluated)
+    except Exception as e:
+        raise ValueError(
+            "eigh is only supported for static eager tensors "
+            "in the openvino backend. Received a dynamic or symbolic tensor."
+        ) from e
+
```
hertschuh: I don't understand why this would be needed. Also, we don't want numpy on the execution path, otherwise something that could run on GPU would run on CPU with NumPy instead of OV.

hertschuh (changes requested):

andersendsa:
Hi @hertschuh i have removed the numpy pathway and implemented jacobi algorithm instead   this pr passes all the tests and is ready for review

PR #22369, opened 2026-03-06
Add fmod to keras.ops.numpy

rstar327:
Add `keras.ops.fmod` — the C-style remainder of division (truncation toward zero).

Unlike `keras.ops.mod` which returns a remainder with the same sign as the divisor, `fmod` returns a remainder with the same sign as the dividend, matching the behavior of C's `fmod` and `numpy.fmod`.

Implements across all backends (numpy, jax, tensorflow, torch). OpenVINO raises `NotImplementedError`.

Includes dynamic shape, static shape, correctness, and dtype tests.

### DIFF
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -2174,9 +2174,14 @@ def fmod(x1, x2):
     dtype = dtypes.result_type(x1.dtype, x2.dtype)
     if dtype == "bool":
         dtype = "int32"
-    x1 = tf.cast(x1, dtype)
-    x2 = tf.cast(x2, dtype)
-    return tf.sign(x1) * tf.math.floormod(tf.abs(x1), tf.abs(x2))
+    # tf.math.floormod does not support uint8/uint16; compute in int32
+    compute_dtype = dtype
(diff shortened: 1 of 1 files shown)


keras/src/backend/tensorflow/numpy.py
```diff
@@ -2168,6 +2168,20 @@ def mod(x1, x2):
...
+    x2 = convert_to_tensor(x2)
+    dtype = dtypes.result_type(x1.dtype, x2.dtype)
+    if dtype == "bool":
+        dtype = "int32"
+    x1 = tf.cast(x1, dtype)
+    x2 = tf.cast(x2, dtype)
+    quotient = x1 / x2
+    truncated = tf.sign(quotient) * tf.math.floor(tf.math.abs(quotient))
+    truncated = tf.cast(truncated, dtype)
+    return x1 - truncated * x2
```
hertschuh: I think this would be a faster implementation:
```python
return tf.sign(x1) * tf.math.floormod(tf.abs(x1), tf.abs(x2))
```

rstar327:
Done

hertschuh (approved):
Thanks!

PR #22379, opened 2026-03-07
Fix Softmax OOM when using mask on torch backend

### LINKED ISSUE #22339: Memory leak with `torch` backend
I've a trained 3D segmentation model(s), inference worked as expected with `jax` and `tensorflow` backend. However, switching to `torch` backend gives OOM with keras version 3.14.0.dev2026030304. 

```python
OutOfMemoryError: Exception encountered when calling Softmax.call().

CUDA out of memory. Tried to allocate 1.53 GiB. GPU 0 has a total capacity of 14.56 GiB of which 155.81 MiB is free. Including non-PyTorch memory, this process has 14.41 GiB memory in use. Of the allocated memory 11.79 GiB is allocated by PyTorch, and 1.49 GiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)

Arguments received by Softmax.call():
  • inputs=torch.Tensor(shape=torch.Size([1, 8, 100, 512000]), dtype=float32)
  • mask=torch.Tensor(shape=torch.Size([1, 1, 100, 512000]), dtype=bool)
```

MarcosAsh:
Fixes #22339                                                                                                                                  
  
## Summary
  The `Softmax` layer hits OOM on the PyTorch backend when called with a mask on large tensors. The post-softmax masking step uses `multiply(outputs, cast(mask, dtype))` which creates an intermediate cast tensor, so three full-size tensors coexist in memory: the masked inputs, the softmax outputs, and the cast mask.

## Additions
  - Adds `del inputs` before post-softmax masking to free the pre-softmax tensor early
  - Replaces `multiply + cast` with `where(mask, outputs, 0.0)` to avoid the intermediate allocation

  Tested on an A100 with the reproduction from the issue (`(1, 20000, 20000)` float32). Peak memory drops from 8.97 GB to 7.37 GB this saved ~1.6GB (one full tensor). If you want to test it yourself I linked the script I used on colab.

  [Colab notebook](https://colab.research.google.com/drive/1K_tPXt4Irdqi0Ox7sZIvUH8EhhKerktK?usp=sharing)


innat:
@MarcosAsh 
Thanks for the PR. 

Just curious, during testing, did you observe the same issue when using pure PyTorch, or does it only occur with Keras (with PyTorch backend)?

MarcosAsh:
Hi! 
I only tested the Keras Softmax layer, both with and without the fix. I didn't test pure PyTorch. But the problem is only with Keras. torch.nn.functional.softmax doesn't take a mask, so there isn't a code path to compare it to. The extra space comes from Keras's post-softmax masking step, where multiply(outputs, cast(mask, dtype)) keeps the masked inputs, the softmax outputs, and a cast copy of the mask all alive at the same time. Pure PyTorch never goes down that path because it doesn't include masking in softmax.

innat:
@MarcosAsh 

Thanks for the heads up. 

From a user’s perspective, I did not encounter this OOM issue when using the tf or jax backends, only with the torch backend. This suggests that similar situations might arise from other keras.ops when running with torch backend.

Given that this type of issue can be quite subtle, would it make sense to have some way to benchmark or verify identical behavior across backends so that such discrepancies can be caught and addressed before they reach users?

I’m raising this because I’ve recently received several negative comments about using torch as a keras backend, particularly from users who primarily work with pytorch. Issues like this can be somewhat embarrassing from a user experience standpoint. I hope you understand the concern I’m trying to highlight.

MarcosAsh:
That's a good point. tf and jax don't hit this because they can combine operations and get rid of intermediates using graph compilation (XLA, tf.function). Because Torch runs quickly, every intermediate tensor stays in memory until it goes out of scope. The same Keras code can act very differently depending on the backend.

Would a tracking issue for torch backend memory problems be useful? That way they're easier to find and contribute to rather than scattered across individual bug reports. If it would help, I'd be happy to look at other operations for similar patterns. The multiply + cast idiom that caused this is easy to search for, and there may be other places where it would fit better. Please tell me if that's something worth looking into.

MarcosAsh:
> Given that this type of issue can be quite subtle, would it make sense to have some way to benchmark or verify identical behavior across backends so that such discrepancies can be caught and addressed before they reach users?

I already have an open proposal for this at #22307, the memory profiling part (phase 3) would catch exactly this kind of thing. I think it would streamline the developer workflow quite a lot and take some guessing out of it. Sorry I forgot to respond to that part.

keras/src/layers/activations/softmax.py
```diff
@@ -75,11 +75,13 @@ def call(self, inputs, mask=None):
...
         if mask is not None:
-            # Apply the mask to the softmax output to ensure that masked
-            # values are set to 0 in case the entire axis is masked.
-            outputs = backend.numpy.multiply(
-                outputs, backend.cast(mask, outputs.dtype)
-            )
+            # Free pre-softmax masked inputs to reduce peak GPU memory.
+            # Without this, the masked inputs, softmax outputs, and
+            # post-masked outputs all exist simultaneously.
+            del inputs
```
hertschuh: 2 nitpicks:

- move this block outside of the `if mask is not None:` block, mostly for the readability of the flow and make it clear it's not related to masking (but it won't actually change anything).
- remove mention of "GPU" in the comment, the comment is true for other computes too (TPU, CPU)

hertschuh (changes requested):
Thanks for the optimization!

MarcosAsh:
Hey just fixed the changes!

keras/src/layers/activations/softmax.py
MarcosAsh: I did the changes thanks for the review

PR #22388, opened 2026-03-09
[OpenVINO backend] Add slogdet support using LU decomposition with partial pivoting 

satheeshbhukya:
## Summary

This PR adds support for `slogdet` (sign and log of the absolute determinant) in the OpenVINO backend.

## Implementation

The operation is implemented using LU-style decomposition with partial pivoting, built entirely with OpenVINO ops.

The main steps are:

* Identify the pivot element in each column using `topk` on the absolute values.
* Perform row swaps when needed and track the sign changes of the determinant.
* Apply Gaussian elimination to transform the matrix into an upper triangular form.
* Accumulate `log(|det|)` from the diagonal (pivot) elements.
* Handle singular matrices by returning `sign = 0` and `logabsdet = -inf`, matching NumPy behavior.
* Support batched inputs and automatically promote integer or boolean inputs to floating point.

## Testing

* `NumpyOneInputOpsCorrectnessTest::test_slogdet` passes successfully.
* Full test suite results: **4933 passed, 896 skipped, 0 failed**. 

### **Closes:** [https://github.com/openvinotoolkit/openvino/issues/34559](https://github.com/openvinotoolkit/openvino/issues/34559)

cc @rkazants  for review.

### DIFF
--- a/keras/src/backend/openvino/excluded_concrete_tests.txt
+++ b/keras/src/backend/openvino/excluded_concrete_tests.txt
@@ -366,7 +366,7 @@ NumpyOneInputOpsCorrectnessTest::test_isreal
 NumpyOneInputOpsCorrectnessTest::test_nanargmax
 NumpyOneInputOpsCorrectnessTest::test_nanargmin
 NumpyOneInputOpsCorrectnessTest::test_real
-NumpyOneInputOpsCorrectnessTest::test_reshape 
+NumpyOneInputOpsCorrectnessTest::test_reshape
 NumpyOneInputOpsCorrectnessTest::test_sinc
 NumpyOneInputOpsCorrectnessTest::test_vectorize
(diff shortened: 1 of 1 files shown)


keras/src/backend/openvino/linalg.py
```diff
@@ -37,6 +39,202 @@ def det(a):
     raise NotImplementedError("`det` is not supported with openvino backend")
 
 
+def slogdet(a):
```
hertschuh: I don't know why `slogdet` was put in `numpy` and not `linalg`.

But please move this code to `numpy.py` for consistency.

satheeshbhukya: Sure, I will this code to numpy.py file

satheeshbhukya:
@hertschuh I have moved slogdet implementation to numpy.py, could you please review

PR #22389, opened 2026-03-09
[OpenVINO] Fix int8/uint8 bitwise ops broadcasting bug; re-enable int4 quantization tests

### LINKED ISSUE #22390: [Bug] [OpenVINO] `bitwise_and`, `bitwise_xor`, `bitwise_or` return wrong results for int8/uint8 tensors with last dimension ≥ 32
OpenVINO's `int8`/`uint8` bitwise kernels produce incorrect results when the last dimension of the input tensor is ≥ 32. Within the first 32 columns, only elements at indices that are multiples of 4 (columns 0, 4, 8, … 28) receive correct values, all other positions receive wrong values. Columns ≥ 32 are always correct.

This is likely an alignment bug in OV's vectorized (SIMD) execution of8-bit bitwise ops, where the kernel writes results back at a stride of 4 rather than 1. It occurs at graph execution time (i.e. when`convert_to_numpy` is called), not during graph construction.

Previously these operations were implemented in #22151.

### Reproducible Code

```python
import os

import numpy as np
from keras.src import ops, backend

# shape (1, 64): last dim >= 32 triggers the bug
x = np.arange(64, dtype=np.int8).reshape(1, 64)
xt = ops.convert_to_tensor(x)
mt = ops.array(np.int8(0x0F), dtype="int8")

result   = backend.convert_to_numpy(ops.bitwise_and(xt, mt))
expected = np.bitwise_and(x, np.int8(0x0F))

print(np.array_equal(result, expected))   # False
print("col 0 (mul of 4):", result[0, 0],  "expected:", expected[0, 0])   # correct
print("col 1 (not mul 4):", result[0, 1], "expected:", expected[0, 1])   # wrong
print("col 32 (>= 32):", result[0, 32],   "expected:", expected[0, 32])  # correct
```

The same bug affects bitwise_xor and bitwise_or for int8/uint8 inputs.

### Impact
Silently corrupts any computation using bitwise ops on int8/uint8 tensors with a wide last dimension. Mostly breaks int4 quantization unpacking (unpack_int4) for embedding layers with output_dim ≥ 64, causing ReversibleEmbedding, Embedding, Dense, and EinsumDense int4 quantization to produce wrong outputs.

goyaladitya05:
Fixes a broadcasting bug in OpenVINO where `bitwise_and`, `bitwise_xor`, and `bitwise_or` return incorrect values for int8/uint8 tensors when one operand is broadcast to match the other's shape. Same-shape operations are unaffected. The bug is in OpenVINO's runtime kernel and is reproducible at the raw openvino.opset15 API level without any Keras code.

- Added helper `_bitwise_op_i8u8` in `keras/src/backend/openvino/numpy.py` that casts int8/uint8 operands to int32/uint32 before the bitwise op and casts back, routing through the 32-bit kernel.

- Enabled 74 previously excluded tests.

Closes #22390


yuanyi-350:
I wonder if it would also make sense to add a targeted regression test for this case.

The fix here seems quite specific to `bitwise_and` / `bitwise_or` / `bitwise_xor` on `int8` / `uint8` when the last dimension is >= 32, while the newly enabled tests seem mostly quantization-related. A small test in `keras/src/ops/numpy_test.py` covering `int8`/`uint8` and a 31 vs 32 boundary case might help prevent future regressions.

goyaladitya05:
> I wonder if it would also make sense to add a targeted regression test for this case.

Yeah that sounds good to me. This still has some work to be done, after that I'll definitely add some specific test for this.

hertschuh (approved):
Thanks for the fix.

This is currently a draft. Is it ready to be submitted?

goyaladitya05:
> Thanks for the fix.
> 
> This is currently a draft. Is it ready to be submitted?

I was thinking of adding some specific test for this case. 
Could we wait another day before merging? I'm trying to find out if some other tests are also excluded due to similar issues

goyaladitya05:
Thank you for waiting. 

I would leave the decision of merging this to the maintainers. This PR introduces a working solution for the bug. While this works, there needs to be a proper fix done for this directly in OpenVINO. I'll be looking into it soon.

hertschuh:
> Thank you for waiting.
> 
> I would leave the decision of merging this to the maintainers. This PR introduces a working solution for the bug. While this works, there needs to be a proper fix done for this directly in OpenVINO. I'll be looking into it soon.

Is there a tracking issue in the OpenVino repository?

My inclination would be to:
- submit this for now
- remove this once the fix is in OpenVino (we just have to not forget to do this)

goyaladitya05:
After isolating the bug further, at the OpenVINO level, the actual issue was broadcasting. Specifically when one operand is a scalar or smaller tensor that needs to be broadcast to match the other's shape. Same-shape operations produce correct results regardless of how large the last dimension is. The fix is still the same, and works fine.

I've updated both the issue and PR description accordingly.

> Is there a tracking issue in the OpenVino repository?

Not yet, I'll be creating it soon. 

> My inclination would be to:
> submit this for now
> remove this once the fix is in OpenVino (we just have to not forget to do this)

@hertschuh Okay. I'll track the issue for this in OpenVINO, and will remove the helper when the issue is fixed (may be atleast 2-3 months, since OpenVINO has a longer release cycle).

goyaladitya05:
Tracking issue for the main bug: openvinotoolkit/openvino/issues/34638
A PR has been opened with the fixes, so this should be fixed in the next release of OpenVINO.

PR #22392, opened 2026-03-10
Replace cast+multiply masking with ops.where to reduce memory

### LINKED ISSUE #22386: Torch backend cast+multiply masking pattern causes unnecessary memory allocations
## Summary 
Following up from #22379, where Softmax hit OOM on the torch backend because of how masking was done. The problem was multiply(tensor, cast(mask, dtype)). On jax and tf this gets fused away during compilation, but torch runs eagerly so every intermediate sticks around in memory. The fix in #22379 was simple: use where instead, which skips the float copy of the mask entirely.

I went through the codebase and found the same pattern in a bunch of other places. Listing them here so they can be picked off over time.

  ### Layer forward passes (big tensors, most likely to cause OOM)

  - layers/attention/attention.py:181 -- casts padding_mask to float and multiplies with attention scores, which can get very arge
  - layers/attention/attention.py:240 -- same thing with q_mask on the attention output
  - layers/normalization/batch_normalization.py:442-464 -- this one is probably the worst after Softmax. Casts the mask, broadcasts it to input shape, and multiplies twice (once for mean, once for variance). Keeps 4+ full-size intermediates alive at the same time
  - layers/core/masking.py:65 -- inputs * backend.cast(boolean_mask, dtype=inputs.dtype)
  - layers/pooling/global_average_pooling1d.py:74-78 -- casts mask to float then multiplies with inputs
  - activations/activations.py:93-95 -- relu with custom threshold, casts a bool comparison to float and multiplies
  - constraints/constraints.py:127 -- NonNeg multiplies weights by ops.greater_equal(w, 0.0), implicit bool-to-float cast

  ### Losses and metrics (smaller tensors, less urgent)

  - losses/losses.py:2349-2353 -- sparse categorical crossentropy ignore_class, two cast+multiply calls
  - losses/loss.py:200 -- apply_mask casts the mask then multiplies with sample_weight
  - metrics/iou_metrics.py:131-135 -- casts the same valid_mask three separate times
  - metrics/confusion_metrics.py:670 -- cast+multiply on feasibility mask
  - metrics/metrics_utils.py:568-570 -- confusion matrix weighted update
  - metrics/regression_metrics.py:511,519 -- R2 metric, casts sample_weight twice

  ### The fix

  For most of these, it's just replacing:
  `tensor * ops.cast(bool_mask, tensor.dtype)`
  with:
  `ops.where(bool_mask, tensor, 0)`

  For places where a lot of intermediates pile up (like batch norm), you can also free things early with del.

Note: #22307 the memory profiling part (phase 3) would catch exactly this kind of thing so this doesnt happen in the future .

rstar327:
## Summary
- Replace `tensor * ops.cast(bool_mask, tensor.dtype)` with `ops.where(bool_mask, tensor, 0)` across 12 files
- On the torch backend these cast+multiply patterns run eagerly, so every intermediate sticks around in memory. `ops.where` skips the float copy of the mask entirely.
- Fixes the layer forward passes (attention, batch norm, masking, pooling, activations, constraints) and losses/metrics (sparse categorical crossentropy, loss masking, IoU, confusion metrics, R2)
- Added `del` on batch norm intermediates for earlier garbage collection

Fixes #22386

### DIFF
--- a/keras/src/layers/normalization/batch_normalization.py
+++ b/keras/src/layers/normalization/batch_normalization.py
@@ -455,6 +455,8 @@ class BatchNormalization(Layer):
             keepdims=True,
         )
         mean = masked_input_sum / (sum_of_weights + backend.epsilon())
+        # Explicit del to free large intermediates in eager mode (e.g. Torch).
+        # No effect under jit/tf.function, but harmless.
         del masked_inputs, masked_input_sum
 
         difference = ops.where(mask_broadcasted, inputs - mean, 0)
--- a/keras/src/layers/normalization/batch_normalization.py
+++ b/keras/src/layers/normalization/batch_normalization.py
@@ -440,9 +440,7 @@ class BatchNormalization(Layer):
             )
 
         mask_broadcasted = ops.expand_dims(mask, axis=-1)
-        mask_broadcasted = ops.broadcast_to(
-            mask_broadcasted, ops.shape(inputs)
-        )
+        mask_broadcasted = ops.broadcast_to(mask_broadcasted, ops.shape(inputs))
         mask_broadcasted = ops.cast(mask_broadcasted, "bool")
         masked_inputs = ops.where(mask_broadcasted, inputs, 0)
 
--- a/keras/src/layers/normalization/batch_normalization.py
+++ b/keras/src/layers/normalization/batch_normalization.py
@@ -443,6 +443,7 @@ class BatchNormalization(Layer):
         mask_broadcasted = ops.broadcast_to(
             mask_broadcasted, ops.shape(inputs)
         )
+        mask_broadcasted = ops.cast(mask_broadcasted, "bool")
         masked_inputs = ops.where(mask_broadcasted, inputs, 0)
 
         masked_input_sum = ops.sum(
--- a/keras/src/layers/normalization/batch_normalization.py
+++ b/keras/src/layers/normalization/batch_normalization.py
@@ -439,33 +439,36 @@ class BatchNormalization(Layer):
                 synchronized=self.synchronized,
             )
 
-        mask_weights = ops.cast(mask, inputs.dtype)
-        mask_weights_broadcasted = ops.expand_dims(mask_weights, axis=-1)
-        broadcasted_mask = ops.broadcast_to(
-            mask_weights_broadcasted, ops.shape(inputs)
+        mask_broadcasted = ops.expand_dims(mask, axis=-1)
+        mask_broadcasted = ops.broadcast_to(
+            mask_broadcasted, ops.shape(inputs)
         )
-        weighted_inputs = broadcasted_mask * inputs
+        masked_inputs = ops.where(mask_broadcasted, inputs, 0)
 
-        weighted_input_sum = ops.sum(
-            weighted_inputs,
+        masked_input_sum = ops.sum(
+            masked_inputs,
             self._reduction_axes,
             keepdims=True,
         )
         sum_of_weights = ops.sum(
-            broadcasted_mask,
+            ops.cast(mask_broadcasted, inputs.dtype),
             self._reduction_axes,
             keepdims=True,
         )
-        mean = weighted_input_sum / (sum_of_weights + backend.epsilon())
+        mean = masked_input_sum / (sum_of_weights + backend.epsilon())
+        del masked_inputs, masked_input_sum
 
-        difference = weighted_inputs - mean
+        difference = ops.where(mask_broadcasted, inputs - mean, 0)
         squared_difference = ops.square(difference)
+        del difference
         weighted_distsq = ops.sum(
-            broadcasted_mask * squared_difference,
+            squared_difference,
             self._reduction_axes,
             keepdims=True,
         )
+        del squared_difference
         variance = weighted_distsq / (sum_of_weights + backend.epsilon())
+        del weighted_distsq
 
         return ops.squeeze(mean), ops.squeeze(variance)
 
--- a/keras/src/layers/pooling/global_average_pooling1d.py
+++ b/keras/src/layers/pooling/global_average_pooling1d.py
@@ -71,14 +71,14 @@ class GlobalAveragePooling1D(BaseGlobalPooling):
     def call(self, inputs, mask=None):
         steps_axis = 1 if self.data_format == "channels_last" else 2
         if mask is not None:
-            mask = backend.cast(mask, inputs[0].dtype)
-            mask = ops.expand_dims(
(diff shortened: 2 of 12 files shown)


keras/src/layers/normalization/batch_normalization.py
```diff
@@ -439,33 +439,37 @@ def _moments(self, inputs, mask):
...
-        mask_weights = ops.cast(mask, inputs.dtype)
-        mask_weights_broadcasted = ops.expand_dims(mask_weights, axis=-1)
-        broadcasted_mask = ops.broadcast_to(
-            mask_weights_broadcasted, ops.shape(inputs)
+        mask_broadcasted = ops.expand_dims(mask, axis=-1)
+        mask_broadcasted = ops.broadcast_to(
+            mask_broadcasted, ops.shape(inputs)
         )
-        weighted_inputs = broadcasted_mask * inputs
+        mask_broadcasted = ops.cast(mask_broadcasted, "bool")
```
MarcosAsh, line 444: This is one of the OpenVINO failures. `mask` from the layer masking system can arrive as float, so `mask_broadcasted` inherits that dtype. OpenVINO's `Select` op (which backs `ops.where`) requires argument 0 to be strictly boolean.

Needs a `mask = ops.cast(mask, "bool")` guard at the top of this block, before the expand/broadcast.

keras/src/layers/pooling/global_average_pooling1d.py
```diff
@@ -71,14 +71,15 @@ def __init__(self, data_format=None, keepdims=False, **kwargs):
     def call(self, inputs, mask=None):
         steps_axis = 1 if self.data_format == "channels_last" else 2
         if mask is not None:
-            mask = backend.cast(mask, inputs[0].dtype)
-            mask = ops.expand_dims(
+            mask_expanded = ops.expand_dims(
                 mask, 2 if self.data_format == "channels_last" else 1
             )
-            inputs *= mask
+            mask_expanded = ops.cast(mask_expanded, "bool")
```
MarcosAsh, line 76: Same OpenVINO issue. `mask` comes in as `i32` here, not boolean. Needs `mask = ops.cast(mask, "bool")` before the expand_dims.

keras/src/losses/loss.py
```diff
@@ -210,17 +209,20 @@ def apply_mask(sample_weight, mask, dtype, reduction):
                 ops.prod(ops.convert_to_tensor(ops.shape(mask), dtype="int32")),
                 dtype,
             )
-            valid = ops.sum(mask)  # May be 0!
-            mask *= ops.divide_no_nan(total, valid)
+            valid = ops.sum(ops.cast(mask, dtype=dtype))  # May be 0!
+            mask_weight = ops.divide_no_nan(total, valid)
+            float_mask = ops.where(ops.cast(mask, "bool"), mask_weight, 0)
```
MarcosAsh, line 214: Same OpenVINO issue `mask` can be float after `squeeze_or_expand_to_same_rank`. Needs a bool cast before being used as the `ops.where` condition.

keras/src/losses/loss.py
```diff
@@ -210,17 +209,20 @@ def apply_mask(sample_weight, mask, dtype, reduction):
...
+        else:
+            float_mask = ops.cast(mask, dtype=dtype)
```
MarcosAsh, line 216: This `else` branch still does `ops.cast(mask, dtype=dtype)` and then `sample_weight *= float_mask` below that's the original cast+multiply pattern this PR is trying to remove. Should use `ops.where` here too.

keras/src/metrics/metrics_utils.py
```diff
@@ -567,7 +567,11 @@ def update_confusion_matrix_variables(
     def weighted_assign_add(label, pred, weights, var):
         label_and_pred = ops.cast(ops.logical_and(label, pred), dtype=var.dtype)
```
MarcosAsh: When `weights is not None`, this cast allocates a float tensor that gets immediately thrown away on the next line. And `ops.logical_and(label, pred)` is recomputed inside the `ops.where` below. Could keep the boolean and branch instead:

```python
label_and_pred = ops.logical_and(label, pred)
if weights is not None:
    result = ops.where(label_and_pred, ops.cast(weights, dtype=var.dtype), 0)
else:
    result = ops.cast(label_and_pred, dtype=var.dtype)
var.assign(var + ops.sum(result, 1))
```

MarcosAsh (commented):
Nice idea, a few things that need fixing before this can land. Thanks for the work on this pr!

### OpenVINO failures (all 10)

Every failure is the same:
```
Argument 0 must have boolean element type (element type: f32)
```
OpenVINO's `Select` op requires the condition to be strictly boolean. In several places the `mask` argument can arrive as float or int after round-tripping through backend masking logic see inline comments.

MarcosAsh:
Sorry just saw you fixed those changes the fixes I flagged on loss.py and metrics_utils.py still exist none the less.

MarcosAsh (commented):
Thanks for the PR.
                                                     
  A few things still open after the last round of fixes:                                             
   
  - metrics_utils.py:568 -- when weights is not None, the first line casts logical_and to float and then the next line throws it away. Also recomputes logical_and inside ops.where. See my earlier inline suggestion.
  - constraints.py -- ops.maximum(w, 0) would be simpler than ops.where(ops.greater_equal(...), ...) here.
  - attention.py:181 -- ops.where(padding_mask, scores - max_value, scores) still materializes a full-size scores - max_value temporary, so there's no memory improvement over the original. Not a regression either, but worth noting since it doesn't match the stated goal.
  - The del statements in batch norm only help in CPython eager mode (no effect under jit or tf.function). Fine given the Torch motivation, but maybe worth a short comment in the code so someone doesn't remove them later thinking they're dead code.

MarcosAsh (commented):
Looks good. Went through all 12 files: the where swaps are all equivalent to the originals, and the batch norm del additions make sense given how many full-size tensors that function holds at once.

Two minor notes, neither blocking:

  - The regression_metrics.py change is more of a "hoist the cast" cleanup than a where-vs-multiply fix. Not wrong, just slightly different from the rest of the PR.
  - constraints.py using ops.maximum(w, 0) instead of ops.where is a nice touch simpler than either the old or new pattern.

Good job !

rstar327:
Thanks for the thorough review\! Good point about `regression_metrics.py` — it is more of a cleanup than a direct memory fix, but consistent with the goal of simplifying the masking pattern. Glad the `ops.maximum` change landed well.

hertschuh (approved):
Thanks for the optimization!

PR #22393, opened 2026-03-10
Fix LayerNormalization rms_scaling to use mean of squares instead of variance

### LINKED ISSUE #21234: LayerNormalization with rms_scaling documentation is different from implementation
The documentation mentions that 
> rms_scaling: If True, center and scale are ignored, and the inputs are scaled by gamma and the inverse square root of the square of all inputs. This is an approximate and faster approach that avoids ever computing the mean of the input.

However, in the [implementation](https://github.com/keras-team/keras/blob/master/keras/src/layers/normalization/layer_normalization.py#L207), it actually does the following:
```
        if self.rms_scaling:
            # Calculate outputs with only variance and gamma if rms scaling
            # is enabled
            # Calculate the variance along self.axis (layer activations).
            variance = ops.var(inputs, axis=self.axis, keepdims=True)
            inv = ops.rsqrt(variance + self.epsilon)

            outputs = (
                inputs * inv * ops.cast(_broadcast(self.gamma), inputs.dtype)
            )
```
So the mean is indeed used, as variance is computed here rather than RMS norm.

There was also a discussion during the addition of RMS Normalization (https://github.com/keras-team/keras/pull/20911#issuecomment-2687658774) that confirms this behavior.

I think the docs could use an update to clarify this behavior. Right now, it sounds like the mean isn't used when `rms_scaling` is on, but the code suggests otherwise.

rstar327:
## Summary
- When `rms_scaling=True`, LayerNormalization was using `var(x)` (variance, which subtracts the mean) instead of `mean(x^2)` (mean of squares)
- True RMS normalization should compute `1/sqrt(mean(x^2) + epsilon)`, not `1/sqrt(var(x) + epsilon)`
- This is a correctness fix: RMS norm by definition does not center inputs around the mean

Fixes #21234

### DIFF
--- a/keras/src/layers/normalization/layer_normalization_test.py
+++ b/keras/src/layers/normalization/layer_normalization_test.py
@@ -123,7 +123,9 @@ class LayerNormalizationTest(testing.TestCase):
         )
         inputs = np.arange(5).astype("float32")[None, :]
         out = layer(inputs)
-        self.assertAllClose(out, [[0.0, 0.70693, 1.41386, 2.12079, 2.82772]])
+        self.assertAllClose(
+            out, [[0.0, 0.40821, 0.81643, 1.22464, 1.63286]], atol=1e-4
+        )
 
     def test_large_value_within_autocast_scope(self):
         layer = layers.LayerNormalization()
--- a/keras/src/ops/nn.py
+++ b/keras/src/ops/nn.py
@@ -3140,9 +3140,11 @@ def _layer_normalization(
         return v
 
     if rms_scaling:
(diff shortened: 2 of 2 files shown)


hertschuh:
Sorry for the time you spent on this but this issue should have been closed:

This is all working as designed after https://github.com/keras-team/keras/pull/21352

`RMSNormalization` is the supported way. `rms_scaling=True` is deprecated and hidden. We did NOT change the behavior of `rms_scaling=True` for backwards compatibility in case people rely on the numerics.

PR #22396, opened 2026-03-11
Remove Orbax dependency and update the OrbaxExport name to Neptune

jerryxyj:
OrbaxModel will be migrated to NeptuneModel coming soon, so we decide not to depend on old OrbaxModel repository and will wait for new repository, which is NeptuneModel released then depending on the NeptuneModel directly.

@hertschuh

### DIFF
--- a/keras/src/export/saved_model.py
+++ b/keras/src/export/saved_model.py
@@ -178,17 +178,9 @@ class ExportArchive:
 
     def __new__(cls, format="saved_model", **kwargs):
         if format == "saved_model":
-            export_model = kwargs.get("export_model", "backend_saved_model")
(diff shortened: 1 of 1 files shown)


keras/src/export/saved_model.py
```diff
@@ -180,7 +182,7 @@ def __new__(cls, format="saved_model", **kwargs):
             if export_model == "backend_saved_model":
                 return BackendSavedModelExportArchive()
             elif export_model == "orbax_export":
```
hertschuh: Please change the name line 184: `elif export_model == "orbax_export"`

jerryxyj: good catch! Done.

PR #22399, opened 2026-03-11
Add cuDNN LSTM for JAX backend

MarcosAsh:
## Summary                                                                                                                    
The JAX backend's lstm() was a NotImplementedError stub, so it always fell back to the generic lax.scan loop. This wires it up to jax.experimental.rnn.lstm, which goes through cuDNN directly on GPU.

The weight conversion is straightforward,  Keras and cuDNN use the same gate order [i, f, g, o], so the kernels just need to be transposed and flattened. Masking works by computing seq_lengths from the boolean mask, with the same right-padded constraint the torch backend has.                                            
                                                  
Falls back to lax.scan on CPU or when cuDNN isn't available, so nothing changes for non-GPU users.     

Benchmarks from the JAX repo put the cuDNN path at around 5x faster than lax.scan for single-layer LSTMs.

### DIFF
--- a/keras/src/backend/jax/rnn.py
+++ b/keras/src/backend/jax/rnn.py
@@ -211,12 +211,126 @@ def rnn(
     return last_output, outputs, new_states
 
 
-def cudnn_ok(*args, **kwargs):
-    return False
+def _is_gpu_available():
+    import jax
+
+    return jax.default_backend() == "gpu"
 
 
-def lstm(*args, **kwargs):
-    raise NotImplementedError
+def cudnn_ok(
+    activation,
+    recurrent_activation,
+    unroll,
+    use_bias=True,
+):
+    from keras.src import activations
+    from keras.src import ops
+
+    return (
+        activation in (activations.tanh, jnp.tanh, ops.tanh)
+        and recurrent_activation in (activations.sigmoid, ops.sigmoid)
+        and not unroll
+        and use_bias
+        and _is_gpu_available()
+    )
+
+
+def lstm(
+    inputs,
+    initial_state_h,
+    initial_state_c,
+    mask,
+    kernel,
+    recurrent_kernel,
+    bias,
+    activation,
+    recurrent_activation,
+    return_sequences=False,
+    go_backwards=False,
+    unroll=False,
+):
+    # Masking is not supported by the cuDNN path; fall back to the
+    # generic RNN loop which handles masking correctly.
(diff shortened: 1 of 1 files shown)


MarcosAsh:
I added test for this! hope it helps

MarcosAsh:
Rebased onto master and added tests that mock the cuDNN path so the weight conversion, mask handling, and output logic get coverage on CPU. Companion PR to #22401 (optimized GRU for JAX).

keras/src/backend/jax/rnn.py
```diff
@@ -211,12 +211,157 @@ def _step(states, current_input):
     return last_output, outputs, new_states
 
 
-def cudnn_ok(*args, **kwargs):
-    return False
+def _is_gpu_available():
+    import jax
 
+    return any(d.platform == "gpu" for d in jax.devices())
```
hertschuh:
```python
return jax.default_backend() == "gpu"
```

keras/src/backend/jax/rnn_test.py
```diff
@@ -0,0 +1,407 @@
...
+from keras.src import testing
+
+
+@pytest.mark.skipif(
+    backend.backend() != "jax",
+    reason="JAX-specific LSTM tests.",
+)
+class JaxLSTMTest(testing.TestCase):
+    def test_cudnn_ok_standard(self):
+        from jax import numpy as jnp
```
hertschuh: We shouldn't need any of these new tests. The ones in `lstm_test.py` should already cover these use cases.

hertschuh:
Please also look at the GPU test failures. It looks like masking is not handled correctly. Thanks!

keras/src/backend/jax/rnn_test.py
MarcosAsh: I removed the tests

keras/src/backend/jax/rnn.py
MarcosAsh: _is_gpu_available() now uses jax.default_backend() == "gpu"

MarcosAsh:
Hi thanks for the review! I fixed the masking issue the cuDNN path now raises NotImplementedError when a mask is present, falling back to the generic RNN loop (same approach as the torch backend). The root problem was that backend.lstm() doesn't receive zero_output_for_mask, so the cuDNN path couldn't produce correct results for that case.

PR #22401, opened 2026-03-11
Add optimized GRU for JAX backend

MarcosAsh:
## Summary 

The JAX backend's gru() was just a NotImplementedError stub, so every GRU call went through the generic rnn() loop. This       
  replaces it with a lax.scan implementation that precomputes all the input projections in one big matmul before entering the
  loop, cutting the per-step work down to a single recurrent matmul.                                                             
                                                                                                                               
  Same constraints as the torch backend: only works with reset_after=True, and masked sequences still go through the generic   
  path since the layer handles zero_output_for_mask separately.
                                                                                                                                 
  There's no cuDNN GRU primitive in JAX (jax-ml/jax#18867), so this is a manual scan. The precomputed projections plus XLA fusion should still help though, especially on longer sequences.

If needed I could make tests for this.


MarcosAsh:
I added tests for it. Hope that helps

MarcosAsh:
@hertschuh Rebased onto master. This is a companion to #22399 (cuDNN LSTM for JAX) both add optimized RNN implementations for the JAX backend. Would appreciate a review when you have time.

hertschuh:
Can you rebase on top of https://github.com/keras-team/keras/pull/22399 ?

Some of the code is shared, right (like `cudnn_ok`)?

keras/src/backend/jax/rnn_test.py
```diff
@@ -0,0 +1,432 @@
+import numpy as np
+import pytest
+
+from keras.src import backend
+from keras.src import testing
+
```
hertschuh: Same thing, we don't need these tests. The cross-backend ones should cover this.

keras/src/backend/jax/rnn.py
```diff
@@ -219,8 +219,70 @@ def lstm(*args, **kwargs):
...
+
+        z = recurrent_activation(x_z + h_z)
+        r = recurrent_activation(x_r + h_r)
+        hh = activation(x_h + r * h_h)
+
+        h_new = z * h + (1 - z) * hh
+        return h_new, h_new
+
+    h_last, outputs = lax.scan(step, h_0, x_all)
+
```
hertschuh, line 392: Is this faster than the existing keras ops based default implementation?

hertschuh (changes requested):
Questions about this one:

keras/src/backend/jax/rnn_test.py
MarcosAsh: Ok I removed them

keras/src/backend/jax/rnn.py
MarcosAsh: Hi, I ran a [benchmark on Colab](https://colab.research.google.com/drive/1vkdE-FlDqbDG7RyS_Dc6gm42v7UhGOQ1?usp=sharing) (T4 GPU,batch=32, input=64):                                                                                 
                              
  | seq_len | hidden | fwd generic | fwd optimized | fwd speedup | grad generic | grad optimized | grad speedup |
  |---------|--------|-------------|---------------|-------------|--------------|----------------|--------------|                   
  | 32 | 64 | 265.10ms | 189.58ms | **1.40x** | 2.11ms | 1.84ms | **1.15x** |
  | 128 | 64 | 270.67ms | 193.28ms | **1.40x** | 7.93ms | 6.81ms | **1.16x** |                          
  | 512 | 64 | 285.51ms | 203.56ms | **1.40x** | 27.82ms | 24.47ms | **1.14x** |
  | 1024 | 64 | 312.36ms | 213.21ms | **1.47x** | 50.27ms | 42.50ms | **1.18x** |                       
  | 32 | 256 | 254.74ms | 194.52ms | **1.31x** | 2.51ms | 2.04ms | **1.23x** |                          
  | 128 | 256 | 255.56ms | 197.30ms | **1.30x** | 9.60ms | 7.85ms | **1.22x** |                         
  | 512 | 256 | 269.65ms | 204.70ms | **1.32x** | 32.35ms | 27.53ms | **1.17x** |                       
  | 1024 | 256 | 288.48ms | 217.37ms | **1.33x** | 59.21ms | 48.89ms | **1.21x** |                      
                                                                                                        
  1.3-1.5x forward speedup and  around 1.2x gradient speedup on GPU across all configurations. The win comes from precomputing input projections in one matmul outside the scan rather than T separate ones inside it.

MarcosAsh:
Hi, thanks for the review! and sorry for the late review. I fixed all the issues and added some benchmark numbers to the one of the comments.

hertschuh (approved):
Thanks!

PR #22409, opened 2026-03-12
[OpenVINO] Implement cond operator; enable tests for Hinge loss funcs and CosineDecay

### LINKED ISSUE #34672: [Keras 3 OpenVINO backend]: Support cond operator and hinge losses
### Request Description

This issue tracks adding support for `cond` operator, and hinge losses (which required cond operator) for the Keras OpenVINO backend.
The operator is currently not implemented in keras/backend/openvino, limiting parity for this backend.

I will submit a PR implementing this and enable the corresponding tests.

Ref: https://github.com/openvinotoolkit/openvino/issues/34017

### Feature Use Case

_No response_

### Issue submission checklist

- [x] The feature request or improvement must be related to OpenVINO

goyaladitya05:
This PR implements `cond` operator for the OpenVINO backend using `ov_opset.select`, which evaluates both branches as graph nodes and selects the result based on pred. 

Enabled 16 previously excluded tests for cond, CosineDecay ,Hinge, and SquaredHinge, which now pass due to the implementation.

Closes: openvinotoolkit/openvino/issues/34672

### DIFF
--- a/keras/src/backend/openvino/core.py
+++ b/keras/src/backend/openvino/core.py
@@ -864,10 +864,9 @@ def cond(pred, true_fn, false_fn):
             pred_ov = ov_opset.convert(pred_ov, Type.boolean).output(0)
 
     def _select(t, f):
-        t_ov = get_ov_output(t)
-        f_ov = get_ov_output(f)
-        if t_ov.get_element_type() != f_ov.get_element_type():
-            f_ov = ov_opset.convert(f_ov, t_ov.get_element_type()).output(0)
+        t_ov, f_ov = align_operand_types(
+            get_ov_output(t), get_ov_output(f), "cond"
+        )
         return OpenVINOKerasTensor(
             ov_opset.select(pred_ov, t_ov, f_ov).output(0)
         )
--- a/keras/src/backend/openvino/core.py
+++ b/keras/src/backend/openvino/core.py
@@ -850,7 +850,31 @@ def cast(x, dtype):
 
 
 def cond(pred, true_fn, false_fn):
-    raise NotImplementedError("`cond` is not supported with openvino backend")
+    true_val = true_fn()
+    false_val = false_fn()
+
+    if true_val is None:
+        return None
+
+    if isinstance(pred, bool):
+        pred_ov = ov_opset.constant(pred, Type.boolean).output(0)
+    else:
+        pred_ov = get_ov_output(pred)
+        if pred_ov.get_element_type() != Type.boolean:
+            pred_ov = ov_opset.convert(pred_ov, Type.boolean).output(0)
+
+    def _select(t, f):
+        t_ov = get_ov_output(t)
+        f_ov = get_ov_output(f)
+        if t_ov.get_element_type() != f_ov.get_element_type():
+            f_ov = ov_opset.convert(f_ov, t_ov.get_element_type()).output(0)
+        return OpenVINOKerasTensor(
+            ov_opset.select(pred_ov, t_ov, f_ov).output(0)
+        )
+
+    if isinstance(true_val, (list, tuple)):
+        return type(true_val)(_select(t, f) for t, f in zip(true_val, false_val))
+    return _select(true_val, false_val)
 
 
 def vectorized_map(function, elements):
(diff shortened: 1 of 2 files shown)


keras/src/backend/openvino/core.py
```diff
@@ -850,7 +850,31 @@ def cast(x, dtype):
 
 
 def cond(pred, true_fn, false_fn):
-    raise NotImplementedError("`cond` is not supported with openvino backend")
+    true_val = true_fn()
+    false_val = false_fn()
```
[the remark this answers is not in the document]
goyaladitya05: OpenVINO builds a computation graph where all nodes are tensor operations with no side effects. Both branches are graph nodes that get compiled but only the selected path's result is used, semantically equivalent to ov_opset.if for this backend's use case, without the complexity of building sub-model graphs.

keras/src/backend/openvino/excluded_concrete_tests.txt
```diff
@@ -55,20 +55,12 @@ CoreOpsCallsTests::test_scan_basic_call
...
 CoreOpsCorrectnessTest::test_switch
 CoreOpsCorrectnessTest::test_unstack
 CoreOpsCorrectnessTest::test_vectorized_map
-CosineDecayRestartsTest::test_alpha
-CosineDecayRestartsTest::test_decay
-CosineDecayRestartsTest::test_float64
-CosineDecayRestartsTest::test_mmul
-CosineDecayRestartsTest::test_tmul
-CosineDecayTest::test_warmup
-CosineDecayTest::test_warmup_decay
```
goyaladitya05, line 71: These have no use case for OpenVINO backend (it being inference-only), these will never be used. But after the implementation of cond, the tests are passing. 

After this change, the tests will run in the CI (since they do not have `@pytest.mark.requires_trainable_backend` decorator) and pass, since they are mathematical tests.

Should I keep them excluded or enable them ? 
Or maybe we could have `@pytest.mark.requires_trainable_backend` decorators for LR schedule releted tests, since they will be used in training only.

keras/src/backend/openvino/core.py
```diff
@@ -850,7 +850,32 @@ def cast(x, dtype):
...
```
hertschuh, line 854: Hmm... so the whole point of `cond` is that you should never have to evaluate the value of both `true_fn` and `false_fn`.

But I suppose this is a limitation of OpenVino, right?

keras/src/backend/openvino/core.py
```diff
@@ -850,7 +850,32 @@ def cast(x, dtype):
...
+
+    if true_val is None:
+        return None
```
hertschuh, line 857: Is this really needed?

keras/src/backend/openvino/excluded_concrete_tests.txt
hertschuh: If they pass, let's run them.

The `@pytest.mark.requires_trainable_backend` was really a way to say "requires auto differentiation (gradients)" support. I think it's fine to still run tests that are training related but don't use gradients like LR schedules.

keras/src/backend/openvino/core.py
goyaladitya05: If I remove that check, then it would result in `get_ov_output(None)`, which throws a `ValueError`. A `ValueError` could be better in this case than returning `None` .

But there is a specific `None` test in `CoreOpsCorrectnessTest`(line:418)-

- With this implementation: The check is triggered, it returns None. Test passes.
- Without this: `get_ov_output(None)` is triggered, and `ValueError` would cause the test to fail.

keras/src/backend/openvino/core.py
goyaladitya05: yes, it's a limitation of OpenVINO. Both branches must be traced so their ops can be wired into an `ov_opset.select` node. At inference time, only one branch executes. Even `tf.cond` works similarly, but Tensorflow handles it internally.

keras/src/backend/openvino/core.py
hertschuh: Oh weird, I didn't realize we supported a None output. Leave it then.

hertschuh (approved):
Thanks for implementing this!

PR #22413, opened 2026-03-13
Warn when trainable state changes after compile()

MarcosAsh:
## Summary
Toggling `.trainable` after `compile()` silently doesn't take effect until `compile()` is called again. This flags the mismatch at runtime.

Snapshots `{id(v) for v in self.trainable_variables}` at `_symbolic_build` time and emits a `UserWarning` on later fit/evaluate calls if the set has drifted. `compile()` resets the snapshot.

No gradient-path change: the 4 backend trainers are untouched. This replaces the earlier enforcement approach in this PR, which would have changed observable training behavior for seeded code.

Known gap: TF without a distribute strategy lazy-builds during tracing and skips `_symbolic_build`, so the warning doesn't fire there. Test is skipped on TF for that reason.

### DIFF
--- a/keras/src/trainers/trainer.py
+++ b/keras/src/trainers/trainer.py
@@ -210,6 +210,8 @@ class Trainer:
         self.test_function = None
         self.predict_function = None
 
+        self._compiled_trainable_variable_ids = None
+
         self._compile_config = serialization_lib.SerializableDict(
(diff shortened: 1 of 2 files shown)


MarcosAsh:
Closing, no prior issue for this behavior change and a known gap on TF.

Issue #22414, opened 2026-03-13
Conv1D layer incorrectly infers non-empty static shape (0-length sequence) when kernel_size > input _length


Isa-Fay:
#### Issue Description
When using keras.layers.Conv1D with a kernel_size larger than the input sequence length, the static shape inference returns a symbolic shape with a 0-length sequence dimension (e.g., (None, 0, 256)) instead of raising an error/warning or marking the shape as invalid. This silent failure leads to broken models that only crash during runtime (eager execution), making debugging difficult for new users.
#### Reproducible Code (Colab)
```
import keras
import numpy as np

eager_input = np.random.randn(4, 10, 128)
symbolic_input = keras.Input(shape=(10,128))

layer = keras.layers.Conv1D(
    filters=256,
    kernel_size=11,
    strides=5
)

try:
  eager_shape = layer(eager_input).shape
  print(eager_shape)
except Exception as e:
  print(e)

try:
  symbolic_shape = layer(symbolic_input).shape
  print(symbolic_shape)
except Exception as e:
  print(e)
```
Colab link (placeholder):
https://colab.research.google.com/drive/16he4e5DmPezftVQJvZJMIkR3NaA1EAV1?usp=sharing

#### Expected Behavior
- Static shape inference (symbolic mode):
The Conv1D layer should either:
-- Raise an explicit ValueError during model construction when kernel_size > input sequence length (leading to 0-length output), or
-- Emit a clear warning that the output sequence length will be 0 (invalid for downstream layers).

#### Actual Behavior
- Static shape inference:
Returns (None, 0, 256) without any error/warning.
- Runtime (eager execution):
Throws an error.
```
Exception encountered when calling Conv1D.call().

The convolution operation resulted in an empty output. Output shape: (4, 0, 256). This can happen if the input is too small for the given kernel size, strides, dilation rate, and padding mode. Please check the input shape and convolution parameters.

Arguments received by Conv1D.call():
  • inputs=tf.Tensor(shape=(4, 10, 128), dtype=float32)
```

dhantule:
Hi @Isa-Fay, I have reproduced this issue with `Keras 3.13.2` in this attached [gist](https://colab.sandbox.google.com/gist/dhantule/2f567471a307c34936a3c001be02a73c/conv1d-layer-incorrectly-infers-non-empty-static-shape-0-length-sequence-when-kernel_size-input-_length-22414.ipynb).

hertschuh:
[...]

PR #22418, opened 2026-03-13
Fix Conv2D to raise error when kernel_size > input spatial dimensions (fixes #22416)

### LINKED ISSUE #22416: Conv2D layer incorrectly infers non-empty static shape (0-size spatial dimensions) when kernel_size > input spatial size (related to #22414)
#### Issue Description
When using keras.layers.Conv2D with a kernel_size larger than the input spatial dimensions (height/width), the static shape inference returns a symbolic shape with 0-size spatial dimensions (e.g., (None, 0, 0, 1)) instead of raising an error/warning or marking the shape as invalid.
This silent failure is a 2D counterpart of the identical issue in Conv1D (#22414 : Conv1D layer incorrectly infers non-empty static shape (0-length sequence) when kernel_size > input length). Both layers suffer from the same core problem—0-dimension output is treated as "valid" in static shape inference but causes runtime crashes—with Conv2D specific to 2D spatial dimensions (height/width) and additional parameters like groups.
#### Reproducible Code (Colab)
```
import keras
import numpy as np

filters = 1
kernel_size = [11, 11]

eager_input = np.random.randn(4, 10, 10, 128)
symbolic_input = keras.Input(shape=(10,10,128))

layer = keras.layers.Conv2D(
    filters=filters,
    kernel_size=kernel_size,
)

try:
  eager_shape = layer(eager_input).shape
  print(eager_shape)
except Exception as e:
  print(e)

try:
  symbolic_shape = layer(symbolic_input).shape
  print(symbolic_shape)
except Exception as e:
  print(e)
```
Colab link:
https://colab.research.google.com/drive/17hy83vSlA3m8Ons6f1N8KipEAhihFqHA?usp=sharing
####  Expected Behavior
- Static shape inference (symbolic mode):
Align with the expected fix for Conv1D (#22414 ):
-- Raise an explicit ValueError during model construction when kernel_size > input spatial dimensions (leading to 0-size output spatial dimensions), or
-- Emit a high-visibility warning that the output height/width will be 0 (invalid for downstream layers like Dense, MaxPool2D).

#### Actual Behavior
- Static shape inference:
Returns (None, 0, 0, 1) without any error/warning (0-size spatial dimensions are treated as "valid"—identical to Conv1D's static shape behavior).
- Runtime (eager execution):
Throws an error.
```
Exception encountered when calling Conv2D.call().

The convolution operation resulted in an empty output. Output shape: (4, 0, 0, 1). This can happen if the input is too small for the given kernel size, strides, dilation rate, and padding mode. Please check the input shape and convolution parameters.

Arguments received by Conv2D.call():
  • inputs=tf.Tensor(shape=(4, 10, 10, 128), dtype=float32)
```

attadevelops:
Fixes #22416

Added validation in `Conv2D.compute_output_shape` to raise a clear error when the kernel_size is larger than the input spatial dimensions with `padding="valid"`, which would result in invalid 0-size output dimensions. This matches the expected behavior described in the related Conv1D issue #22414. Now instead of silently returning `(None, 0, 0, filters)` during model construction, it raises a ValueError with a helpful message about what's wrong and how to fix it.

### DIFF
--- a/keras/src/layers/convolutional/conv2d.py
+++ b/keras/src/layers/convolutional/conv2d.py
@@ -1,3 +1,4 @@
+from keras.src import ops
 from keras.src.api_export import keras_export
 from keras.src.layers.convolutional.base_conv import BaseConv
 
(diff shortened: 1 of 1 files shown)


attadevelops:
Closing this PR — I took on more than I could follow through on. Apologies for the noise.

PR #22419, opened 2026-03-13
Add rad2deg to keras.ops.numpy

rstar327:
## Description

Add `keras.ops.rad2deg` — the counterpart of the existing `keras.ops.deg2rad`. Converts angles from radians to degrees using `deg = rad * (180 / π)`.

This completes the pair of angle conversion functions in the Keras ops API, matching NumPy's `numpy.rad2deg`.

## Changes

- Added `Rad2deg` operation class and `rad2deg` function to `keras/src/ops/numpy.py`
- Implemented backend functions for numpy, jax, tensorflow, torch, and openvino
- Added tests: dynamic shape, static shape, correctness, and dtype tests
- Updated API exports


hertschuh (approved):
Thanks for adding this!

PR #22425, opened 2026-03-14
[OpenVINO] Implement fmod op

### LINKED ISSUE #34701: [Keras 3 OpenVINO backend]: Implement support for fmod op
### Request Description

This issue tracks improving support for fmod operator for the Keras OpenVINO backend.

The operations is currently not implemented properly in keras/backend/openvino/numpy.py, and have some tests excluded, limiting Numpy parity for this backend.

I will submit a PR implementing this and enable the corresponding tests.

Parent Issue: https://github.com/openvinotoolkit/openvino/issues/34017

### Feature Use Case

_No response_

### Issue submission checklist

- [x] The feature request or improvement must be related to OpenVINO

goyaladitya05:
`fmod` was added to `keras.ops` in #22369.
This PR implements it for the OpenVINO Backend.

Closes: openvinotoolkit/openvino/issues/34701

### DIFF
--- a/keras/src/backend/openvino/excluded_concrete_tests.txt
+++ b/keras/src/backend/openvino/excluded_concrete_tests.txt
@@ -299,7 +299,6 @@ NNOpsDtypeTest::test_glu_
 NNOpsDtypeTest::test_polar_
 NNOpsDynamicShapeTest::test_glu
 NumpyDtypeTest::test_array
-NumpyDtypeTest::test_fmod
 NumpyDtypeTest::test_maximum_python_types
 NumpyDtypeTest::test_minimum_python_types
 NumpyDtypeTest::test_nanargmax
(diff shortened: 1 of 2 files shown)


keras/src/backend/openvino/numpy.py
```diff
@@ -2849,7 +2849,10 @@ def mod(x1, x2):
 
 
 def fmod(x1, x2):
-    raise NotImplementedError("fmod is not supported by openvino backend.")
+    x1 = get_ov_output(x1)
+    x2 = get_ov_output(x2)
+    x1, x2 = _align_operand_types(x1, x2, "fmod()")
+    return OpenVINOKerasTensor(ov_opset.mod(x1, x2).output(0))
```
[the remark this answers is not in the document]
goyaladitya05: Actually its the opposite for OpenVINO. the implementation of `mod`, uses `floor_mod`. Vice-versa here.

keras/src/backend/openvino/excluded_concrete_tests.txt
goyaladitya05, line 1: I have a question, unrelated to this PR. 

Scenario: A test has the decorator `@pytest.mark.requires_trainable_backend`, so it wouldn't run in CI for OpenVINO anyways. Do we need to explicitly keep it excluded in the file, or can we remove it? 
We'd obviously go for the file level exclusion instead of the current test specific exclusions.

There are multiple such tests.

hertschuh: > Do we need to explicitly keep it excluded in the file, or can we remove it?

You can remove it from the excluded file.

The eventual goal is to remove everything in the excluded file.

hertschuh (approved):
Thanks for implementing this!

PR #22426, opened 2026-03-14
Add fliplr and flipud numpy ops

rstar327:
## Summary

Add `keras.ops.fliplr` and `keras.ops.flipud` numpy operations with support across all backends (NumPy, JAX, TensorFlow, PyTorch, OpenVINO).

- `fliplr` reverses elements along axis 1 (left/right flip)
- `flipud` reverses elements along axis 0 (up/down flip)

These are standard NumPy operations commonly used in image preprocessing and data augmentation.

## Implementation

- Operation classes with `compute_output_spec` and dimension validation
- Backend implementations using native ops (`np.fliplr`, `jnp.fliplr`, `tf.reverse`, `torch.fliplr`, OpenVINO `flip`)
- Symbolic shape tests (dynamic and static) and correctness tests

### DIFF
--- a/keras/src/backend/jax/numpy.py
+++ b/keras/src/backend/jax/numpy.py
@@ -749,10 +749,12 @@ def flip(x, axis=None):
 
 
 def fliplr(x):
+    x = convert_to_tensor(x)
     return jnp.fliplr(x)
 
 
 def flipud(x):
+    x = convert_to_tensor(x)
     return jnp.flipud(x)
 
 


keras/src/backend/jax/numpy.py
```diff
@@ -748,6 +748,14 @@ def flip(x, axis=None):
     return jnp.flip(x, axis=axis)
 
 
+def fliplr(x):
+    return jnp.fliplr(x)
+
+
+def flipud(x):
+    return jnp.flipud(x)
```
hertschuh, line 758: Please add `x = convert_to_tensor(x)` to both.

rstar327:
Done

PR #22427, opened 2026-03-14
Add i0 (modified Bessel function) numpy op

rstar327:
## Summary

Add `keras.ops.i0` — the modified Bessel function of the first kind, order 0 — with support across NumPy, JAX, TensorFlow, and PyTorch backends.

- NumPy: `np.i0(x)`
- JAX: `jnp.i0(x)`
- TensorFlow: `tf.math.bessel_i0(x)`
- PyTorch: `torch.i0(x)`
- OpenVINO: not supported (raises `NotImplementedError`, added to excluded tests)

This is a standard NumPy operation used in signal processing (e.g., Kaiser window functions) and statistical applications.

## Implementation

- Operation class with `compute_output_spec` and proper dtype handling
- Backend implementations using native ops
- Symbolic shape tests (dynamic and static) and correctness tests


hertschuh (approved):
Thanks for adding this!

PR #22433, opened 2026-03-14
[Keras 3 OpenVINO Backend] Add GRU implementation for OpenVINO backend

satheeshbhukya:
## Summary
This PR adds support for `gru` (Gated Recurrent Unit) in the OpenVINO backend.

## Implementation
The operation is implemented using OpenVINO ops and integrates with 
the existing `rnn()` loop infrastructure.

The main steps are:
* Pre-compute the input projection (`inputs @ kernel + input_bias`) 
  once over the full sequence outside the time loop for efficiency.
* Split the projected input into three gate slices (z, r, h) using 
  `variadic_split` along the feature axis.
* For `reset_after=True` (cuDNN-compatible default), compute a single 
  recurrent matmul and split into gate projections, then add the 
  recurrent bias if present.
* For `reset_after=False`, split the recurrent kernel into per-gate 
  sub-kernels and apply the reset gate to `h_prev` before the 
  recurrent projection.
* Compute the update gate `z`, reset gate `r`, candidate hidden state 
  `hh`, and final hidden state `h = z * h_prev + (1 - z) * hh`.
* Pass recurrent kernel and bias as invariant inputs to the OpenVINO 
  `Loop` op so they are wired once and reused every iteration.
* Masking, `go_backwards`, and `time_major` are handled via the 
  existing `rnn()` infrastructure.

### Closes: [openvinotoolkit/openvino#34408](https://github.com/openvinotoolkit/openvino/issues/34408)

cc @hertschuh for review


goyaladitya05:
There's a shape bug in the current implementation of RNN, so until that is resolved, I don't think either of LSTM or GRU can be implemented.

satheeshbhukya:
> There's a shape bug in the current implementation of RNN, so until that is resolved, I don't think either of LSTM or GRU can be implemented.

I think you are implementing these

hertschuh:
Closing as this was done and fixed in https://github.com/keras-team/keras/pull/22436

PR #22434, opened 2026-03-14
Fix ReversibleEmbedding failing with .summary()

### LINKED ISSUE #22432: ReversibleEmbedding fails with ".summary()"
I am using keras version 3.13.2

The following code fails on `simple_model.summary()`, even though I think it should work?

```python
import keras

class SimpleModel(keras.models.Model):

    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.embed_and_unembed = keras.layers.ReversibleEmbedding(
            input_dim=5,     # vocab size
            output_dim=10,   # model dimension
            tie_weights=True
        )

    def call(self, idx):
        emb = self.embed_and_unembed(idx)
        logits = self.embed_and_unembed(emb, reverse=True)
        return logits


simple_model = SimpleModel()
sequence = keras.random.randint(shape=(2,20), minval=0, maxval=5)
print(simple_model(sequence).shape) # prints (2,20,5) as expected
simple_model.summary()
```
The error is: "ReversibleEmbedding.compute_output_shape() got an unexpected keyword argument 'inputs_shape'"

rstar327:
## Summary

Fix `ReversibleEmbedding.compute_output_shape()` failing when called by `model.summary()`.

The `build` method uses `inputs_shape` as its parameter name, so `_build_shapes_dict` contains `{"inputs_shape": ...}`. When `summary_utils` calls `compute_output_shape(**_build_shapes_dict)`, it passes `inputs_shape` as a keyword argument, but `compute_output_shape` expected `input_shape` (singular), causing a `TypeError`.

The fix renames the parameter from `input_shape` to `inputs_shape` to match the `build` signature.

Fixes #22432

### DIFF
--- a/keras/src/layers/core/reversible_embedding.py
+++ b/keras/src/layers/core/reversible_embedding.py
@@ -106,8 +106,8 @@ class ReversibleEmbedding(layers.Embedding):
         self.reverse_dtype = reverse_dtype
         self.logit_soft_cap = logit_soft_cap
 
-    def build(self, inputs_shape=None):
-        super().build(inputs_shape)
+    def build(self, input_shape=None):
+        super().build(input_shape)
         if not self.tie_weights and self.quantization_mode not in (
             "int8",
             "int4",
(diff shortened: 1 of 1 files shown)


themavik (commented):
Good fix for the summary display issue with ReversibleEmbedding. The approach correctly handles the case where the layer has a different output shape format. Clean and minimal change.

rstar327:
Done

PR #22355, opened 2026-03-05
[OpenVINO backend] Support for numpy.det using Faddeev-LeVerrier algorithm

andersendsa:
### Details :
Implement numpy.det operation for the OpenVINO backend.

### Closes issue : https://github.com/openvinotoolkit/openvino/issues/34510

PR #22437, opened 2026-03-15
[OpenVINO backend] Support for numpy.det using LU decomposition with partial pivoting

satheeshbhukya:
### Description

This PR adds the implementation of the `det` operation for the OpenVINO backend in Keras.

The determinant is computed using LU decomposition with partial pivoting. During the decomposition, pivot elements are accumulated to compute the determinant while tracking row swaps to correctly adjust the sign. This approach follows the standard mathematical relation:

det(A) = sign × product of pivot elements

The implementation supports batched matrices by flattening batch dimensions and performing the computation across the batch.

### Key points

* Supports inputs of shape `(..., n, n)`
* Handles batched matrices
* Casts integral and boolean inputs to floating point types
* Tracks row swaps to correctly compute determinant sign
* Reshapes outputs back to batch dimensions

### Closes: [openvinotoolkit/openvino#34510](https://github.com/openvinotoolkit/openvino/issues/34510)
cc @hertschuh for review

### DIFF
--- a/keras/src/backend/openvino/excluded_concrete_tests.txt
+++ b/keras/src/backend/openvino/excluded_concrete_tests.txt
@@ -201,7 +201,6 @@ LayerTest::test_end_to_end_masking
 LayerTest::test_quantized_layer_with_remat
 LayerTest::test_stateless_call
 LinalgOpsCorrectnessTest::test_cholesky
-LinalgOpsCorrectnessTest::test_det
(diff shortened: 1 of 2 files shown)


andersendsa:
Hi @satheeshbhukya   I have already implemented this issue with my pr #22355
i would urge u to close this pull request

hertschuh:
Already implemented in https://github.com/keras-team/keras/pull/22355

PR #22438, opened 2026-03-16
Fix Inconsistent tensor conversion

Abhinavexists:
- Convert L2.l2 to tensor in __init__ via ops.convert_to_tensor() to match L1 behavior, avoiding repeated implicit float-to-tensor conversion on every forward pass


Abhinavexists:
> Thanks!
> 
> Can you fix the unit tests? It's just a question of change some `assertEqual` to `assertAllClose`.

Sure

hertschuh (approved):
Thanks!

PR #22443, opened 2026-03-17
Use jax2onnx for direct JAX-to-ONNX export

MarcosAsh:
## Summary

- JAX ONNX export is broken on JAX >= 0.4.36 because the old pipeline required `enable_xla=False` and `native_serialization=False` in `jax2tf`, both of which were removed from JAX
- Replace the `jax2tf` + `tf2onnx` pipeline with [`jax2onnx`](https://github.com/enpasos/jax2onnx) for direct JAX-to-ONNX conversion -- no TensorFlow intermediary needed
- Also unblocks NumPy 2.0 compatibility for JAX-backend ONNX export (the `tf2onnx` path was incompatible)
- Bidirectional LSTMs (reverse scan) are not yet supported by `jax2onnx` and are skipped in tests until upstream adds support

### DIFF
--- a/keras/src/export/onnx_test.py
+++ b/keras/src/export/onnx_test.py
@@ -79,8 +79,8 @@ def get_model(type="sequential", input_shape=(10,), layer_list=None):
 )
 @pytest.mark.skipif(testing.uses_gpu(), reason="Fails on GPU")
 @pytest.mark.skipif(
-    np.version.version.startswith("2."),
-    reason="ONNX export is currently incompatible with NumPy 2.0",
+    np.version.version.startswith("2.") and backend.backend() != "jax",
+    reason="ONNX export via tf2onnx is incompatible with NumPy 2.0",
 )
 class ExportONNXTest(testing.TestCase):
     @parameterized.named_parameters(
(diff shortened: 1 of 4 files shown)


keras/src/export/onnx_test.py
```diff
@@ -79,8 +79,8 @@ def get_model(type="sequential", input_shape=(10,), layer_list=None):
 )
 @pytest.mark.skipif(testing.uses_gpu(), reason="Fails on GPU")
 @pytest.mark.skipif(
-    np.version.version.startswith("2."),
-    reason="ONNX export is currently incompatible with NumPy 2.0",
+    np.version.version.startswith("2.") and backend.backend() != "jax",
+    reason="ONNX export via tf2onnx is incompatible with NumPy 2.0",
```
hertschuh, line 83: Can `jax2onnx` work on GPU?

hertschuh:
Also, can you check and fix the code format?

keras/src/export/onnx_test.py
MarcosAsh: Jax2onnx works purely through JAX abstract tracingt there is  no actual on the device computation happens during export, so GPU vs CPU shouldn't be an issue. The GPU skip here is pre existing and applies to all backends though. I'll keep it as is to limit scope should I change this in a seperate PR?

MarcosAsh:
Hi, I fixed the issue thanks for the review! @hertschuh

MarcosAsh:
Hi, I dont know why the stat: awaiting response from contributor hasnt been removed

PR #22446, opened 2026-03-18
Upgrade OpenVINO to use opset16

goyaladitya05:
OpenVINO released opset16 in the 2026.0 version.

This PR upgrades opset15 to opset16.

opset16 upgraded `avg_pool` from opset14, adding a dilations parameter that the C++ runtime now requires to be explicitly set. Added `dilations=[1] * num_spatial_dims` to all `avg_pool` calls, mirroring the existing pattern already used for max_pool.

There were 6 new operators introduced in opset16.
To limit the scope of this PR, I will add the functional improvements in a follow up PR.


goyaladitya05:
Keeping this on hold for now, as there is a bug with opset16. All ops should've been ported from opset15 to opset16, but 10 ops that were previously introduced in opset15 are missing.

goyaladitya05:
> Clean opset16 upgrade with a well-documented workaround for the upstream OpenVINO bug. The _opset module centralizes the fallback nicely. One nit: in _opset.py, `del _opset15, _op` — the `_op` is the loop variable after iteration; consider `del _opset15` only to avoid confusion. The workaround can be removed once [openvinotoolkit/openvino#34780](https://github.com/openvinotoolkit/openvino/issues/34780) is fixed and released.

I specifically added that, to avoid `_op` leaking into the namespace.

goyaladitya05:
> How soon do you think an openvino release will happen with the fix?

The fix got merged yesterday, but its milestoned for 2026.2, and the latest release is 2026.0, so it would take 4-5 months.

> This is kind of ugly. And is there anything we need from `opset16` urgently?

It would make the implementations of `istft` and `segment_max` more efficient. 

- `istft`: 9 tensor memory round trips (reshape→transpose→pad pipeline) → 1 fused kernel; ~80 graph nodes → ~12.
- `segment_max`: ~28 graph nodes → ~7.

I think the switch to opset16 can be postponed as well, there is nothing that's being blocked.

hertschuh:
> It would make the implementations of `istft` and `segment_max` more efficient.
> 
> * `istft`: 9 tensor memory round trips (reshape→transpose→pad pipeline) → 1 fused kernel; ~80 graph nodes → ~12.
> * `segment_max`: ~28 graph nodes → ~7.

Alternatively, you can import `opset16` just for `istft` and `segment_max`, I think it would be cleaner.

goyaladitya05:
> Alternatively, you can import `opset16` just for `istft` and `segment_max`, I think it would be cleaner.

Okay, I'll look into it.

PR #22453, opened 2026-03-20
Fix `_parameterize_data` in OpenVino `Trainer` class.

hertschuh:
The `_parameterize_data` method:
- did not handle `OpenVINOKerasTensor` instances.
- has a syntax error for the `raise` case


hertschuh:
@goyaladitya05

goyaladitya05:
> @goyaladitya05

@hertschuh Both fixes look good to me. Was there something specific you needed from me here?

hertschuh:
> > @goyaladitya05
> 
> @hertschuh Both fixes look good to me. Was there something specific you needed from me here?

No. That's it, thanks!

PR #22456, opened 2026-03-20
[Fix] Bug: Pipeline.from_config() crashes with KeyError/TypeError on malformed config

ChiragSW:
Fixes the crash issue with wrong error for: #22450

### DIFF
--- a/keras/src/layers/preprocessing/pipeline.py
+++ b/keras/src/layers/preprocessing/pipeline.py
@@ -68,9 +68,15 @@ class Pipeline(Layer):
 
     @classmethod
     def from_config(cls, config):
+        layers = config.get("layers")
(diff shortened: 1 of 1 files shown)


ChiragSW:
please review the PR, the crash issue has been fixed. 
@hertschuh

hertschuh:
I can't access the issue, it seems to be gone.

But `from_config` is a purely internal deserialization utility that is not called by users. If you use Keras "normally", you would never to able to trigger this error because `get_config` returns the right structure. I don't know what the use case of the issue was, but it's not a standard user workflow.

All that to say that we cannot add code with friendly user messages in every place that the code assumes some data is present because that's basically everywhere in the code.

PR #22461, opened 2026-03-20
Move TensorFlow CUDA tests to GPU runners.

hertschuh:
We are migrating to GPU custom runners for GPU tests instead of Kokoro.

- Disable TF32 for better numerical accuracy
- Pin tensorflow to 2.20 as 2.21 doesn't seem to work with our L4 + CUDA13 setup


laxmareddyp (approved):
LGTM,Thanks

PR #22464, opened 2026-03-22
Fix custom_gradient for OpenVINO backend

goyaladitya05:
Custom_gradient in `openvino/core.py` was intended to be a class (like the numpy backend) but was missing the class keyword, leaving __init__ and __call__ as dead local functions and causing the decorator to return None. 

This made any `@ops.custom_gradient` decorated function uncallable (TypeError: 'NoneType' object is not callable), breaking all quantized forward passes in Dense, EinsumDense, and related quantizer ops.

### DIFF
--- a/keras/src/backend/openvino/core.py
+++ b/keras/src/backend/openvino/core.py
@@ -1650,7 +1650,7 @@ class custom_gradient:
 
     OpenVINO is an inference-only backend, so this acts as a pass-through:
     it runs the forward pass and discards the gradient function.
-    
+
     Arguments:
         fun: The forward pass function.
     """


keras/src/backend/openvino/excluded_concrete_tests.txt
```diff
@@ -114,9 +63,6 @@ FilterSafePathsTest::test_member_within_base_dir
 FilterSafePathsTest::test_symbolic_link_in_base_dir
 FilterSafePathsTest::test_symlink_within_base_dir
 GetFileTest::test_join_simple
-GPTQQuantizerTest::test_quantize_clipping_behavior_extremes
-GPTQQuantizerTest::test_zero_scale_guard_no_nans_for_finite_inputs
-GroupedQuantizationParametersTest::test_grouped_weight_shapes_divisible
 GroupedQuantizationParametersTest::test_grouped_weight_shapes_non_divisible
```
goyaladitya05, line 66: This still fails for a seperate reason, will fix it in a different PR.

keras/src/backend/openvino/excluded_concrete_tests.txt
```diff
@@ -278,10 +222,8 @@ OptimizerTest::test_gradient_accumulation_with_weigth_decay9
 OptimizerTest::test_overwrite_with_gradient_with_gradient_accumulation
 QrOpTest
 QuantizersTest::test_abs_max_quantizer
-QuantizersTest::test_compute_float8_amax_history
 QuantizersTest::test_compute_float8_scale
 QuantizersTest::test_grouped_quantize_with_padding
```
goyaladitya05, line 226: These still fail for a seperate reason, will fix it in a different PR.

Issue #22406, opened 2026-03-12
`TextVectorization` fails on G4 on Colab (all fine on H100, or A100 and earlier)


jchwenger:
Hi there,

Trying to use `TextVectorization` (see the [Colab experiment](https://colab.research.google.com/drive/1SLIjk06hC_PEYgPt2X8MPqmMZImmMl8g?usp=sharing)) on the mentioned hardware fails with this error:
```
InternalError: {{function_node __wrapped__Cast_device_/job:localhost/replica:0/task:0/device:GPU:0}} 'cuLaunchKernel(function, gridX, gridY, gridZ, blockX, blockY, blockZ, 0, reinterpret_cast<CUstream>(stream), params, nullptr)' failed with 'CUDA_ERROR_INVALID_HANDLE' [Op:Cast] name: 
```

I'm guessing one could hedge one's bet and always tokenize on CPU, but since older code worked without these caveats, I'm guessing you will want to keep it that way?

Thanks in advance!

dhantule:
Hi @jchwenger, thanks for reporting this.

I have reproduce this error in this [gist](https://colab.sandbox.google.com/gist/LakshmiKalaKadali/1bde204092041098838e6f650c25f9ac/textvectorization-error.ipynb), we'll look into this and update you.

jchwenger:
My pleasure @dhantule, thanks in advance!

ChiragSW:
@hertschuh I have fixed the issue in my PR, please review

hertschuh:
Fixed by https://github.com/keras-team/keras/pull/22465

Thanks @ChiragSW

PR #22465, opened 2026-03-22
[Fix] TextVectorization fails on G4 on Colab (all fine on H100, or A100 and earlier)

ChiragSW:
Fixes issue: #22406 

Root cause: TF string/hash/lookup ops always produce CPU-resident tensors. On newer GPUs (G4/L4), TF incorrectly dispatches the subsequent `Cast` op to the GPU, raising `CUDA_ERROR_INVALID_HANDLE`. Fix: wrap all such ops in `with tf.device("CPU:0")`.
 
## Files Changed
 
| File | Why |
|---|---|
| `text_vectorization.py` | String ops in `_preprocess` need CPU pinning. Fixed a bug that I saw, comment wrote about the return but `output_mode=None` never returned. |
| `index_lookup.py` | `StaticHashTable.lookup` is CPU-only, pinned the full call chain covering `StringLookup` and `IntegerLookup` too. |
| `hashing.py` | `tf.strings.to_hash_bucket_fast/strong` and `tf.as_string` are CPU-only ops. |
| `hashed_crossing.py` | `tf.sparse.cross_hashed` is a CPU-only op. |
| 5 Test files | GPU regression tests added, skipped automatically on CPU-only machines. |

### DIFF
--- a/keras/src/layers/preprocessing/hashed_crossing_test.py
+++ b/keras/src/layers/preprocessing/hashed_crossing_test.py
@@ -210,20 +210,3 @@ class HashedCrossingTest(testing.TestCase):
             tf.sparse.to_dense(cloned_outputs),
             tf.sparse.to_dense(original_outputs),
         )
-
-    @pytest.mark.skipif(
-        not tf.test.is_gpu_available(),
-        reason="GPU not available; skipping GPU-specific regression test",
-    )
-    def test_call_on_gpu_does_not_raise(self):
-        """Regression test for CUDA_ERROR_INVALID_HANDLE on G4/L4 GPUs.
-
-        tf.sparse.cross_hashed produces CPU-resident tensors. On newer GPU
-        architectures a Cast after a CPU hash tensor was incorrectly
-        dispatched to the GPU.
-        """
-        layer = layers.HashedCrossing(num_bins=5)
-        feat1 = np.array(["A", "B", "A", "B", "A"])
-        feat2 = np.array([101, 101, 101, 102, 102])
-        output = layer((feat1, feat2))
-        self.assertAllClose(output, np.array([1, 4, 1, 1, 3]))
--- a/keras/src/layers/preprocessing/hashing_test.py
+++ b/keras/src/layers/preprocessing/hashing_test.py
@@ -431,27 +431,6 @@ class HashingTest(testing.TestCase):
         ):
             layers.Hashing(num_bins=3, output_mode="int", sparse=True)
 
-    @pytest.mark.skipif(
-        not tf.test.is_gpu_available(),
-        reason="GPU not available; skipping GPU-specific regression test",
-    )
-    def test_call_on_gpu_does_not_raise(self):
-        """Regression test for CUDA_ERROR_INVALID_HANDLE on G4/L4 GPUs.
-
-        tf.strings.to_hash_bucket_fast/strong produce CPU-resident tensors.
-        On newer GPU architectures a Cast after a CPU string tensor was
-        incorrectly dispatched to the GPU.
-        """
-        layer = layers.Hashing(num_bins=5)
-        inp = [["A"], ["B"], ["C"], ["D"], ["E"]]
-        output = layer(inp)
-        self.assertAllClose(output, np.array([[1], [0], [1], [1], [2]]))
-
-        # Also test with salt (SipHash path)
-        layer_salt = layers.Hashing(num_bins=5, salt=[133, 137])
-        output_salt = layer_salt(inp)
-        self.assertEqual(output_salt.shape, (5, 1))
-
 
 # TODO: support tf.RaggedTensor.
 # def test_hash_ragged_string_input_farmhash(self):
--- a/keras/src/layers/preprocessing/integer_lookup_test.py
+++ b/keras/src/layers/preprocessing/integer_lookup_test.py
@@ -2,7 +2,6 @@ import os
 
 import numpy as np
 import pytest
-import tensorflow as tf
 from tensorflow import data as tf_data
 
 from keras.src import backend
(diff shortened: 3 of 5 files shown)


ChiragSW:
@hertschuh The issue specifically targets the Turing architecture (T4/G4 GPUs), which are the standard for Google Colab's free tier. The L4 (Ada Lovelace) uses a different memory management and kernel execution path where this specific CUDA_ERROR_INVALID_HANDLE doesn't trigger.

hertschuh:
> @hertschuh The issue specifically targets the Turing architecture (T4/G4 GPUs), which are the standard for Google Colab's free tier. The L4 (Ada Lovelace) uses a different memory management and kernel execution path where this specific CUDA_ERROR_INVALID_HANDLE doesn't trigger.

But the current unit tests run on T4, and they don't fail.

ChiragSW:
> > @hertschuh The issue specifically targets the Turing architecture (T4/G4 GPUs), which are the standard for Google Colab's free tier. The L4 (Ada Lovelace) uses a different memory management and kernel execution path where this specific CUDA_ERROR_INVALID_HANDLE doesn't trigger.
> 
> But the current unit tests run on T4, and they don't fail.

The issue mentions G4 specifically, can you please check if current unit tests run on G4?

ChiragSW:
@hertschuh any updates?

keras/src/layers/preprocessing/hashed_crossing_test.py
```diff
@@ -210,3 +210,20 @@ def test_tf_string(self):
...
+
+        tf.sparse.cross_hashed produces CPU-resident tensors. On newer GPU
+        architectures a Cast after a CPU hash tensor was incorrectly
+        dispatched to the GPU.
+        """
+        layer = layers.HashedCrossing(num_bins=5)
+        feat1 = np.array(["A", "B", "A", "B", "A"])
+        feat2 = np.array([101, 101, 101, 102, 102])
+        output = layer((feat1, feat2))
+        self.assertAllClose(output, np.array([1, 4, 1, 1, 3]))
```
hertschuh: There's nothing in this test that's not covered by the tests above. Please revert this file.

The fact that the issue was not caught has nothing to do with the existing tests, just the hardware used.

keras/src/layers/preprocessing/hashing_test.py
```diff
@@ -431,6 +431,27 @@ def test_hashing_sparse_with_int_mode(self):
...
+        layer = layers.Hashing(num_bins=5)
+        inp = [["A"], ["B"], ["C"], ["D"], ["E"]]
+        output = layer(inp)
+        self.assertAllClose(output, np.array([[1], [0], [1], [1], [2]]))
+
+        # Also test with salt (SipHash path)
+        layer_salt = layers.Hashing(num_bins=5, salt=[133, 137])
+        output_salt = layer_salt(inp)
+        self.assertEqual(output_salt.shape, (5, 1))
+
```
hertschuh: There's nothing in this test that's not covered by the tests above. Please revert this file.

The fact that the issue was not caught has nothing to do with the existing tests, just the hardware used.

keras/src/layers/preprocessing/integer_lookup_test.py
```diff
@@ -307,3 +308,16 @@ def test_oov_method_farmhash_output_is_correct(self):
...
+        reason="GPU not available; skipping GPU-specific regression test",
+    )
+    def test_call_on_gpu_does_not_raise(self):
+        """Regression test for CUDA_ERROR_INVALID_HANDLE on G4/L4 GPUs."""
+        layer = layers.IntegerLookup(
+            vocabulary=[1, 2, 3],
+            output_mode="int",
+        )
+        output = layer([1, 2, 99])
+        self.assertAllClose(output, [1, 2, 0])
```
hertschuh: There's nothing in this test that's not covered by the tests above. Please revert this file.

The fact that the issue was not caught has nothing to do with the existing tests, just the hardware used.

keras/src/layers/preprocessing/string_lookup_test.py
```diff
@@ -261,3 +262,20 @@ def test_one_hot_symbolic_output_shape_nested_input(self):
...
+
+        StaticHashTable lookup is CPU-resident; the Cast that follows must also
+        stay on CPU on newer GPU architectures.
+        """
+        layer = layers.StringLookup(
+            vocabulary=["a", "b", "c"],
+            output_mode="int",
+        )
+        output = layer(["a", "b", "d"])
+        self.assertAllClose(output, [1, 2, 0])
```
hertschuh: There's nothing in this test that's not covered by the tests above. Please revert this file.

The fact that the issue was not caught has nothing to do with the existing tests, just the hardware used.

keras/src/layers/preprocessing/text_vectorization_test.py
```diff
@@ -524,3 +524,23 @@ def test_output_sequence_length_with_non_int_mode(self):
...
+        dispatched to the GPU. Verify no error is raised when a GPU is present.
+        """
+        layer = layers.TextVectorization(
+            output_mode="int",
+            output_sequence_length=4,
+        )
+        layer.adapt(["foo bar", "bar baz", "baz bada boom"])
+        input_data = [["foo bar"], ["baz boom"]]
+        output = layer(input_data)
+        self.assertEqual(output.shape, (2, 4))
```
hertschuh: There's nothing in this test that's not covered by the tests above. Please revert this file.

The fact that the issue was not caught has nothing to do with the existing tests, just the hardware used.

hertschuh (changes requested):
Thanks for looking into this! The code changes themselves look good.

But please revert all the unit test files.

hertschuh:
The new test would actually fail:
https://github.com/keras-team/keras/actions/runs/23668108953/job/68955074278

ChiragSW:
> The new test would actually fail: https://github.com/keras-team/keras/actions/runs/23668108953/job/68955074278

Yes this test was removed in my latest commit. As reviewed by you it was not needed

ChiragSW:
@hertschuh any updates?

PR #22469, opened 2026-03-22
docs: clarify safe_mode limitations in deserialize_keras_object

### LINKED ISSUE #22466: [Docs] safe_mode=True does not isolate built-in class names  from get_custom_objects() registry
## Description
The documentation for `safe_mode` in `deserialize_keras_object` only states it protects against "unsafe lambda deserialization."

However, the parameter name `safe_mode` implies broader safety guarantees. Users are not warned that safe_mode=True does NOT isolate built-in class name resolution from the custom object registry. This creates a documentation gap that could mislead developers into assuming stronger protections than exist.

## Poc

```
import keras

class CustomDense(keras.layers.Layer):
    def __init__(self, **kwargs):
        super().__init__()
        print("CustomDense called instead of real Dense")
    def call(self, x):
        return x

keras.utils.get_custom_objects()["Dense"] = CustomDense

config = {
    "class_name": "Dense",
    "config": {"units": 5, "name": "test"},
    "registered_name": "Dense"
}


serialization_lib.deserialize_keras_object(config, safe_mode=True)

```

## Observed Result

```

CustomDense called instead of real Dense
<CustomDense name=custom_dense, built=False>

```

## Expected Behavior

safe_mode=True should resolve "Dense" to the built-in keras.layers.Dense, ignoring any custom registry overrides for built-in class names.

## Actual Behavior

safe_mode=True silently resolves "Dense" to the custom registry entry, bypassing the built-in class entirely.

## Suggested improvement

`safe_mode` parameter documentation to explicitly state that it only protects against Lambda layer deserialization and does not isolate built-in class names from overrides via `keras.utils.get_custom_objects()`.

Abineshabee:
Fixes #22466

## Problem
The `safe_mode` parameter documentation in `deserialize_keras_object` 
only stated it protects against unsafe Lambda deserialization. 
This was misleading as users might assume broader safety guarantees.

## Changes
Updated the `safe_mode` docstring in `keras/src/saving/serialization_lib.py` to clarify:
- It only protects against Lambda layer deserialization
- It does NOT isolate built-in class names from 
  `keras.utils.get_custom_objects()` or `keras.utils.CustomObjectScope()` overrides
- Added `.. warning::` block for visibility in Sphinx docs

### DIFF
--- a/keras/src/saving/serialization_lib.py
+++ b/keras/src/saving/serialization_lib.py
@@ -508,7 +508,7 @@ def deserialize_keras_object(
             serialized within the Keras model file being loaded. It does
             not provide isolation from the local Python environment and
             does not guard against modifications made outside of the
-             serialized file.
+            serialized file.
 
     Returns:
         The object described by the `config` dictionary.


keras/src/saving/serialization_lib.py
```diff
@@ -501,10 +501,23 @@ class ModifiedMeanSquaredError(keras.losses.MeanSquaredError):
...
+            built-in class name resolution from the custom object registry.
+            If a custom object has been registered under the same name as a
+            built-in class (e.g., `"Dense"`) via
+            `keras.utils.get_custom_objects()`, `safe_mode=True` will still
+            resolve to the custom object instead of the built-in class.
+            `safe_mode` only protects against Lambda layer deserialization
+            and does not provide broader registry isolation.
+
+            .. warning::
+                ``safe_mode=True`` only protects against Lambda layer
```
hertschuh: We don't support this format: `.. warning::` it won't render correctly on keras.io

amadhan882:
Hi @Abineshabee,

Following @hertschuh’s feedback, I’ve updated the docstring to keep the explanation general and removed specific examples and unsupported Sphinx tags.

Here is the revised version:

safe_mode: Boolean, defaults to `False`. If `True`, disables unsafe lambda deserialization.

Note that safe_mode is designed to protect against code serialized within the Keras model file being loaded. It does not provide isolation from the local Python environment and does not guard against modifications made outside of the serialized file.

hertschuh:
> Hi @Abineshabee,
> 
> Following @hertschuh’s feedback, I’ve updated the docstring to keep the explanation general and removed specific examples and unsupported Sphinx tags.
> 
> Here is the revised version:
> 
> safe_mode: Boolean, defaults to `False`. If `True`, disables unsafe lambda deserialization.
> 
> Note that safe_mode is designed to protect against code serialized within the Keras model file being loaded. It does not provide isolation from the local Python environment and does not guard against modifications made outside of the serialized file.

I don't see any update in the code. Did you forget to push?

amadhan882:
Hi @hertschuh, sorry for the confusion in my previous comment. I was suggesting the revised wording to help @Abineshabee, but they haven't pushed the changes to the PR yet.

@Abineshabee, please update the code with the suggested docstring. If you're unable to update it, let me know—I’m happy to open a new PR with these changes to get this resolved quickly.

Abineshabee:
Sorry for the delay, changes have been pushed. Please review @hertschuh!

PR #22470, opened 2026-03-22
Rewrite torch LSTM to use functional API with CPU fallback

### LINKED ISSUE #19177: [Feature Request] Add cuDNN-accelerated LSTM and GRU to PyTorch
Hi,

are there any plans to add cuDNN-accelerated versions of LSTM and GRU to the PyTorch backend? Without cuDNN acceleration, the LSTM and GRU are considerably (several times) slower, even when running on GPU; however, we still use RNNs heavily (for example, adding them after Transformer encoder still helps in some cases).

The `torch.nn.LSTM`/`torch.nn.GRU` offer cuDNN acceleration, and wrapping them to a `keras.layers.Layer` works, but the resulting model is not backend-agnostic (so the resulting model cannot be used cross-frameworks).

Thanks for consideration :pray:  and cheers!

PS: Relatedly, `torch.nn.LSTM/GRU` offers bidirectional computation by a single call (by passing `bidirectional=True`) -- I am not sure how much faster it is compared to two asynchronous unidirectional computations, but if it is faster, `keras.layers.Bidirectional` would probably have to be updated to handle `keras.layers.LSTM` and `keras.layers.GRU` specifically to support it.

### LINKED ISSUE #22275: Torch backend: LSTM model training is terribly slow
LSTM model created with `use_cudnn=True` is VERY slow comparing to pure-pytorch counterpart on Keras 3.11.3.
Could not test it on higher versions due to https://github.com/keras-team/keras/issues/22274
The keras model likely does not truly utilize cuDNN-optimized path as GPU utilization stays really low on training.

Colab to reproduce:
https://colab.research.google.com/drive/1OOjpO4OGNGJi6w0sfxbVv6P7S_-idhXv

MarcosAsh:
## Summary
Same treatment as the GRU in #22115 -switch from creating a `torch.nn.LSTM` module every forward pass to `torch._VF.lstm`, and add a pure-torch fallback for CPU / masking / non-standard activations.

The old code had a few problems: it called `detach().clone().cpu()` on the outputs (breaks gradients), allocated a new module + copied weights per call, and raised `NotImplementedError` on CPU so the layer fell back to the slow generic cell loop.

Fixes #19177, fixes #22275

### DIFF
--- a/keras/src/backend/torch/rnn.py
+++ b/keras/src/backend/torch/rnn.py
@@ -568,9 +568,12 @@ def lstm(
     if bias is not None:
         bias = convert_to_tensor(bias)
 
-    inputs = convert_to_tensor(inputs, dtype="float32")
-    initial_state_h = convert_to_tensor(initial_state_h, dtype="float32")
-    initial_state_c = convert_to_tensor(initial_state_c, dtype="float32")
+    # Cast inputs/states to the kernel's dtype so integer inputs are promoted
+    # to float and mixed-precision dtypes (e.g. float16) are respected.
+    compute_dtype = kernel.dtype
+    inputs = convert_to_tensor(inputs).to(compute_dtype)
+    initial_state_h = convert_to_tensor(initial_state_h).to(compute_dtype)
+    initial_state_c = convert_to_tensor(initial_state_c).to(compute_dtype)
 
     # Preprocess for go_backwards by flipping the sequence
     if go_backwards:


keras/src/backend/torch/rnn.py
```diff
@@ -646,95 +629,97 @@ def _cudnn_lstm(
...
+    initial_state_h,
+    initial_state_c,
+    kernel,
+    recurrent_kernel,
+    bias,
+    activation,
+    recurrent_activation,
+    return_sequences,
+    batch_first,
+):
```
hertschuh, line 684: Just curious, is this fallback better / faster than the non-torch specific fallback that already exists?

keras/src/backend/torch/rnn.py
```diff
@@ -621,23 +583,44 @@ def lstm(
...
-    except RuntimeError:
+    cudnn_supported = cudnn_ok(
+        activation,
+        recurrent_activation,
+        unroll,
+        use_bias=bias is not None,
+    )
+
+    if mask is not None:
         raise NotImplementedError
```
hertschuh: Move this to the top (line 557). All the computations you do between 557 and here are just wasted.

hertschuh:
> pure-torch fallback for CPU / masking / non-standard activations.

Oh, you also mentioned the fallback handles masking, but it doesn't. Is that something you intended to add?

keras/src/backend/torch/rnn.py
MarcosAsh: Good call, moved it to the very first thing in lstm() so  we skip all the tensor conversion and device transfer work.

keras/src/backend/torch/rnn.py
MarcosAsh: Yeah, a fair bit. The generic path calls step_function per timestep, which does a full matmul against kernel (input_dim x 4*units) at every step. This fallback pre-computes all input projections in one batched matmul across the sequence, so the per-step loop only hits the smaller recurrent projection (units x 4*units).

  I made a quick CPU benchmark:            
  
  | Config                 | Generic | Fallback | Speedup | 
  |------------------------|---------|----------|---------|
  | B=32 T=50 F=64 U=64    | 40.7ms  | 6.8ms    | 6.0x    |
  | B=32 T=100 F=128 U=128 | 152.2ms | 15.5ms   | 9.8x    | 
  | B=64 T=50 F=256 U=256  | 111.2ms | 30.6ms   | 3.6x    |
                                                              
 The gap grows with sequence length since the batched matmul amortizes better.

MarcosAsh:
Hi thanks for the review. Done I removed the xfail from test_basics and dropped torch_uses_gpu() from the test_masking xfail. You're right, the PR description was inaccurate. Neither the cuDNN path nor the fallback handles masking, masked inputs raise NotImplementedError and fall through to the generic RNN loop. I updated the description.

MarcosAsh:
I think the same issue as my other pr where the label of stat: awaiting response from reviewer persists

hertschuh:
@MarcosAsh 

Did you see the error on GPU? You probably need a cast before the matmul in the fallback implementation.

MarcosAsh:
@hertschuh 
Thanks for the review and sorry for the late response. I noticed the convert_to_tensor calls were forcing dtype="float32" on inputs and states, but kernel/recurrent_kernel kept their autocast dtype (float16 under mixed_float16). Dropped the dtype override so everything stays in the compute dtype the layer gives it.

Issue #22474, opened 2026-03-23
ConvLSTM1D allows invalid strides/dilation config when built with Keras Input


Isa-Fay:
### Summary
`keras.layers.ConvLSTM1D` correctly rejects `strides > 1` together with `dilation_rate > 1` in eager execution, but the same invalid configuration is silently accepted when the layer is built from a `keras.Input` placeholder.

### Reproduction
```python
import keras
import numpy as np
import tensorflow as tf
import keras
import tensorflow as tf
import numpy as np

def call_func(
    filters,
    kernel_size,
    strides=1,
    padding="valid",
    data_format=None,
    dilation_rate=1,
    activation="tanh",
    recurrent_activation="sigmoid",
    use_bias=True,
    kernel_initializer="glorot_uniform",
    recurrent_initializer="orthogonal",
    bias_initializer="zeros",
    unit_forget_bias=True,
    kernel_regularizer=None,
    recurrent_regularizer=None,
    bias_regularizer=None,
    activity_regularizer=None,
    kernel_constraint=None,
    recurrent_constraint=None,
    bias_constraint=None,
    dropout=0.0,
    recurrent_dropout=0.0,
    seed=None,
    return_sequences=False,
    return_state=False,
    go_backwards=False,
    stateful=False,
    unroll=False,
    inputs=None,
    initial_state=None,
    mask=None,
    training=False
):
    layer = keras.layers.ConvLSTM1D(
        filters=filters,
        kernel_size=kernel_size,
        strides=strides,
        padding=padding,
        data_format=data_format,
        dilation_rate=dilation_rate,
        activation=activation,
        recurrent_activation=recurrent_activation,
        use_bias=use_bias,
        kernel_initializer=kernel_initializer,
        recurrent_initializer=recurrent_initializer,
        bias_initializer=bias_initializer,
        unit_forget_bias=unit_forget_bias,
        kernel_regularizer=kernel_regularizer,
        recurrent_regularizer=recurrent_regularizer,
        bias_regularizer=bias_regularizer,
        activity_regularizer=activity_regularizer,
        kernel_constraint=kernel_constraint,
        recurrent_constraint=recurrent_constraint,
        bias_constraint=bias_constraint,
        dropout=dropout,
        recurrent_dropout=recurrent_dropout,
        seed=seed,
        return_sequences=return_sequences,
        return_state=return_state,
        go_backwards=go_backwards,
        stateful=stateful,
        unroll=unroll
    )
[...]

Issue #22476, opened 2026-03-23
ConvLSTM3D fails to validate incompatible strides and dilation_rate in symbolic mode


Isa-Fay:
## Summary
`keras.layers.ConvLSTM3D` correctly rejects `strides > 1` together with `dilation_rate > 1` when called on an eager tensor, but the same invalid configuration is accepted when the input is a `keras.Input` placeholder.

## Reproduction
```python
import keras
import tensorflow as tf

x1 = tf.random.normal((2, 5, 8, 8, 8, 3))
x2 = keras.Input(shape=(5, 8, 8, 8, 3))

layer = keras.layers.ConvLSTM3D(
    filters=1,
    kernel_size=1,
    strides=2,
    padding="same",
    dilation_rate=[1, 2, 2],
    return_state=True,
)

# Eager tensor: raises as expected
print(layer(x1))

# Symbolic tensor: should also raise, but succeeds
print(layer(x2))
```

## Expected behavior
Both calls should raise an error like:

> `strides > 1` not supported in conjunction with `dilation_rate > 1`

## Actual behavior
Testing with eager tensor:
Dynamic execution error: Exception encountered when calling ConvLSTMCell.call().

`strides > 1` not supported in conjunction with `dilation_rate > 1`. Received: strides=[2 2 2] and dilation_rate=[1 2 2]

Arguments received by ConvLSTMCell.call():
  • inputs=tf.Tensor(shape=(2, 8, 8, 8, 3), dtype=float32)
  • states=('tf.Tensor(shape=(2, 4, 4, 4, 1), dtype=float32)', 'tf.Tensor(shape=(2, 4, 4, 4, 1), dtype=float32)')
  • training=None

Testing with Keras.Input placeholder:
Static output shapes: [(None, 4, 4, 4, 1), (None, 4, 4, 4, 1), (None, 4, 4, 4, 1)]

LakshmiKalaKadali:
[...]

Issue #22475, opened 2026-03-23
ConvLSTM2D allows invalid strides+dilation_rate config when built with Keras.Input


Isa-Fay:
### Summary
`keras.layers.ConvLSTM2D` correctly rejects `strides > 1` together with `dilation_rate > 1` when called on eager tensors, but the same invalid configuration is silently accepted when the layer is built from a `Keras.Input` placeholder.

### Reproduction
```python
import keras
import tensorflow as tf

x1 = tf.random.normal((2, 5, 32, 32, 3))
x2 = keras.Input(shape=(5, 32, 32, 3))

layer = keras.layers.ConvLSTM2D(
    filters=1,
    kernel_size=(3, 3),
    strides=(2, 2),
    dilation_rate=(3, 3),
    padding="same",
    return_sequences=True,
)

# Eager tensor: raises an error
print(layer(x1))

# Symbolic input: succeeds unexpectedly
print(layer(x2))
```

### Expected behavior
Both execution paths should reject this configuration with an error like:
`strides > 1 not supported in conjunction with dilation_rate > 1`.

### Actual behavior
- Eager tensor call: raises the expected error.
- `Keras.Input` call: succeeds and returns output shape `(None, 5, 16, 16, 1)`.

### Impact
Invalid layer configurations can slip through model construction when using symbolic inputs, causing inconsistent behavior between build time and runtime.

LakshmiKalaKadali:
Hi @Isa-Fay , Thanks for reporting the issue. It has been resolved in PR https://github.com/keras-team/keras/pull/22477. Thanks for the PR @ChiragSW. I have verified the fix with `keras-nightly-3.14.0.dev2026032404`. Now, the layer `convLSTM2D` handles the error for strides > 1 together with dilation_rate > 1.  Please refer to the [gist](https://colab.sandbox.google.com/gist/LakshmiKalaKadali/741e91489030e3ed26cc3c80dac27ab6/keras_22475.ipynb#scrollTo=0jWbjNX34vAG). Thank You

whitakercaleb09-dev:
Okay

PR #22477, opened 2026-03-23
[Fix] ConvLSTM1D allows invalid strides/dilation config when built with Keras Input

ChiragSW:
Fixes issues: #22474 #22475 #22476

Root cause: ConvLSTMCell normalized strides and dilation_rate but didn’t validate their incompatible combination (> 1 together).
So, symbolic layers.Input builds didn’t raise errors

Solution: Added validation in ```ConvLSTMCell.__init__```to raise ValueError when both are > 1.
Also added to regression tests to ensure symbolic builds now correctly fail across ConvLSTM1D/2D/3D.

### DIFF
--- a/keras/src/layers/rnn/conv_lstm1d_test.py
+++ b/keras/src/layers/rnn/conv_lstm1d_test.py
@@ -78,15 +78,13 @@ class ConvLSTM1DTest(testing.TestCase):
             tpu_rtol=1e-3,
         )
 
-    @pytest.mark.requires_trainable_backend
     def test_symbolic_invalid_strides_dilation_rate(self):
-        inputs = layers.Input(shape=(10, 32, 3))
         with self.assertRaisesRegex(
             ValueError,
             r"Specifying `strides > 1` is not compatible with "
             r"`dilation_rate > 1`",
         ):
-            layer = layers.ConvLSTM1D(
+            layers.ConvLSTM1D(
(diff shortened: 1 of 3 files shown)


keras/src/layers/rnn/conv_lstm1d_test.py
```diff
@@ -77,3 +77,21 @@ def test_correctness(self):
             tpu_atol=1e-3,
             tpu_rtol=1e-3,
         )
+
+    @pytest.mark.requires_trainable_backend
```
hertschuh: Remove `@pytest.mark.requires_trainable_backend` in all 3 tests, it's unrelated.

keras/src/layers/rnn/conv_lstm1d_test.py
```diff
@@ -77,3 +77,21 @@ def test_correctness(self):
...
+        ):
+            layer = layers.ConvLSTM1D(
+                filters=1,
+                kernel_size=1,
+                strides=2,
+                padding="same",
+                dilation_rate=3,
+                return_sequences=True,
+            )
+            layer(inputs)
```
hertschuh: In all 3 tests:

- Remove `inputs = layers.Input(shape=(10, 32, 3))`
- Remove `layer(input)`

This is dead code because it fails even before you call the layer.

PR #22488, opened 2026-03-24
[jax_layer] Pass native_serialization_platforms to jax2tf

Caslyn:
* Add `native_serialization_platforms` argument to `JaxLayer.__init__`.
* Pass the new argument to `jax2tf.convert` inside `_jax2tf_convert`.
* Remove unused `dtype` doc string from `JaxLayer.__init__`

This allows users to specify exact hardware target platforms (e.g., 'cpu', 'cuda', 'tpu') during JAX-to-TF conversion.

### DIFF
--- a/keras/src/utils/jax_layer.py
+++ b/keras/src/utils/jax_layer.py
@@ -364,13 +364,11 @@ class JaxLayer(Layer):
         from jax.experimental import jax2tf
 
         jax2tf_kwargs = {"polymorphic_shapes": polymorphic_shapes}
-        # Backwards compatibility for users < JAX v0.4.24
+        # Backwards compatibility for users < JAX v0.4.32
         if self.jax2tf_native_serialization_platforms is not None:
             jax2tf_kwargs["native_serialization_platforms"] = (
                 self.jax2tf_native_serialization_platforms
             )
-            jax2tf_kwargs["native_serialization"] = True
-
         converted_fn = jax2tf.convert(fn, **jax2tf_kwargs)
         # Autograph won't work with the output of jax2tf.
         converted_fn = tf.autograph.experimental.do_not_convert(converted_fn)


keras/src/utils/jax_layer.py
```diff
@@ -226,8 +226,11 @@ def my_haiku_module_fn(inputs, training):
             argument, then `init_fn` is called at build time to initialize the
             non-trainable state of the model.
       seed: Seed for random number generator. Optional.
-      dtype: The dtype of the layer's computations and weights. Can also be a
-            `keras.DTypePolicy`. Optional. Defaults to the default policy.
+      native_serialization_platforms: Sequence of platforms ('cpu', 'cuda',
+            'rocm', 'tpu') to compile for when using `jax2tf.convert` with
+            native serialization. If `None`, the function is compiled for the
+            default backend. If multiple platforms are specified, the exported
+            module will be device-polymorphic.
```
hertschuh, line 234: You should call out the fact that this is only used when the Keras backend is tensorflow, most people use this with the jax backend, and this argument doesn't do anything in that case.

keras/src/utils/jax_layer.py
```diff
@@ -355,7 +362,15 @@ def get_single_jax2tf_shape(shape):
     def _jax2tf_convert(self, fn, polymorphic_shapes):
         from jax.experimental import jax2tf
 
-        converted_fn = jax2tf.convert(fn, polymorphic_shapes=polymorphic_shapes)
+        jax2tf_kwargs = {"polymorphic_shapes": polymorphic_shapes}
+        # Backwards compatibility for users < JAX v0.4.24
+        if self.jax2tf_native_serialization_platforms is not None:
```
hertschuh: > `# Backwards compatibility for users < JAX v0.4.24`

I don't understand this comment. Which part is for backwards compatibility? Is it the `jax2tf_kwargs["native_serialization"] = True` part?

keras/src/utils/jax_layer.py
Caslyn: Done ([09d1a3f](https://github.com/keras-team/keras/pull/22488/commits/09d1a3fab7aef0719934dcc415e7d042f416f083))

keras/src/utils/jax_layer.py
Caslyn: Thanks for flagging this. I got a little mixed up about when this feature was introduced and how to incorporate backwards compatibility.

I believe https://github.com/jax-ml/jax/commit/1a9f49963cc4343307ffcb5245100ad3a00ee0e4 is the original commit that introduced `native_serialization_platforms` to `jax2tf.convert`, and was included in Jax 0.4.32.

Per [the changelog](https://github.com/jax-ml/jax/blob/69629aa30eb02253276b1417a305b260e858a16a/CHANGELOG.md?plain=1#L287) the `native_serialization` boolean is deprecated anyway and isn't required to be set when passing `native_serialization_platforms`.

I removed `native_serialization` and corrected the BW date for `native_serialization_platforms`. The comment now only pertains to `native_serialization_platforms`. ([e0fa7cc](https://github.com/keras-team/keras/pull/22488/commits/e0fa7cc3d7c035edd0a3b1ddd088223908e9d130))

PR #22493, opened 2026-03-25
Fix JAX RNN Fusion Error on CPU

pctablet505:
**Problem**
When running Keras RNN tests using the JAX backend on CPU, the following error occurs:

`JaxRuntimeError: UNIMPLEMENTED: Fusion kind not implemented on CPU`

**Root Cause**
Tracing the issue into the JAX codebase (`jax/_src/numpy/lax_numpy.py` and `jax/_src/lax/lax.py`), `jnp.split` binds to a low-level `split_p` primitive. When this primitive is compiled inside a loop (`lax.scan`) on CPU, XLA's fusion pass fails because it lacks the implementation for this specific fusion combination on CPU.

**Solution**
Instead of modifying the common Keras layers (`lstm.py`, `gru.py`) which would complicate the code and potentially hurt performance on accelerators (GPU/TPU), this PR intercepts the split call at the JAX backend level (`keras/src/backend/jax/numpy.py`).

If the active JAX device is CPU and `indices_or_sections` is an integer (uniform split), the backend uses standard Python slicing to split the tensor. Standard slicing lowers to simpler HLO primitives (like `slice_in_dim`) that the XLA CPU compiler can digest without triggering unimplemented fusion errors.


hertschuh (commented):
> Problem
When running Keras RNN tests using the JAX backend on CPU, the following error occurs:

> JaxRuntimeError: UNIMPLEMENTED: Fusion kind not implemented on CPU

Can you provide more details? All the tests pass on CPU right now. Is this a different use case?

Also, shouldn't this be fixed on the JAX side?

pctablet505:
> > Problem
> > When running Keras RNN tests using the JAX backend on CPU, the following error occurs:
> 
> > JaxRuntimeError: UNIMPLEMENTED: Fusion kind not implemented on CPU
> 
> Can you provide more details? All the tests pass on CPU right now. Is this a different use case?
> 
> Also, shouldn't this be fixed on the JAX side?

@hertschuh 
This was a workaround, as I was getting this in some of TAP failures, currently not able to reproduce the issue. Maybe it might have been fixed by jax team

hertschuh:
Let's close it then

PR #22499, opened 2026-03-25
[OpenVINO] Fix cond side-effect isolation for optimizer tests

goyaladitya05:
OpenVINO's cond called both true_fn() and false_fn() directly, so both branches' var.assign() side effects fired unconditionally, corrupting optimizer state during gradient accumulation and EMA steps.

After this change, each branch now runs inside a _TrackingScope (subclass of StatelessScope) that records assignments without applying them. After both branches execute, `ov_opset.select(pred, true_val, false_val)` is used to build a symbolic node for each updated variable, and only that selected value is committed via `_direct_assign`.

Removed 44 tests from excluded_concrete_tests.txt that now pass.

### DIFF
--- a/.github/workflows/trigger_gpu_tpu_tests.yml
+++ b/.github/workflows/trigger_gpu_tpu_tests.yml
@@ -1,0 +1,26 @@
+name: Trigger GPU & TPU Tests
+
+on:
+  pull_request_target:
+    types: [labeled]
+
+permissions:
+  contents: read
+  issues: write
+  pull-requests: write
+
+jobs:
+  remove_label:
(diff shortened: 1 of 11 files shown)


hertschuh (approved):
Wow. I guess that solves the problem. The truth is that we don't really care about optimizers since we can't train anyway.

But I suppose the same issue could happen in layers too.

goyaladitya05:
>  The truth is that we don't really care about optimizers since we can't train anyway.
> 
> But I suppose the same issue could happen in layers too.

Yeah, my intent behind this was, that if the tests can run, then they should pass.

PR #22500, opened 2026-03-25
[OpenVINO] Implement QR decomposition

goyaladitya05:
This PR implements `qr` in using the Householder reflection algorithm. The loop applies successive reflectors to zero out subdiagonals of R while accumulating Q via right-multiplication. Handles batched inputs, both reduced and complete modes, and rank-2 through N-D inputs.

Implementation:

- All slice/concat operands use 1-D tensors (OpenVINO concat rejects scalars)
- Householder vector built via one-hot mask to preserve shape linkage through the loop body, which OpenVINO needs for matmul shape inference
- Skips the reflector (zeroes v_hat) when sub_m == 1, matching LAPACK's tau=0 convention for the trivial last step
- Reshapes loop output to concrete dimensions immediately after get_iter_value, since dynamic k in body slices causes the loop to lose static shape info
- Works in f32 throughout (workaround for an OpenVINO CPU Loop constant-folding bug with f64, same pattern as det())

Closes: openvinotoolkit/openvino/issues/34540

### DIFF
--- a/keras/src/backend/openvino/linalg.py
+++ b/keras/src/backend/openvino/linalg.py
@@ -624,7 +624,8 @@ def qr(x, mode="reduced"):
 
     # Work in f32:
     #   f64 — constant-folding bug in OpenVINO CPU Loop evaluate (same as det())
-    #   f16/bf16 — upcast to f32 for numerical stability in iterative Householder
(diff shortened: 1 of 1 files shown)


satheeshbhukya:
bro i have implemented this but you have created pr what should i do.. i want to create pr for it..

PR #22505, opened 2026-03-25
Fix KeyError in model.summary() when a layer returns a dict

### LINKED ISSUE #22445: KeyError: 0 in model.summary() when a nested layer returns a dictionary

MarcosAsh:
## Summary
 `model.summary()` crashes with `KeyError: 0` when a nested layer's `compute_output_shape` returns a dict (e.g. a multi-task head).

`format_layer_shape` assumed `output_shapes` was always a list or tuple. When `compute_output_shape` returns a dict, `tree.map_shape_structure` preserves it, so `len(output_shapes) == 1` passes for a single-key dict but `output_shapes[0]` fails because dicts are keyed by name, not index.

The fix checks for dict before indexing single-key dicts extract the value, multi-key dicts stringify normally.

 Fixes #22445

### DIFF
--- a/keras/src/utils/summary_utils.py
+++ b/keras/src/utils/summary_utils.py
@@ -111,8 +111,9 @@ def format_layer_shape(layer):
                 )
         except NotImplementedError:
             return "?"
-    if len(output_shapes) == 1:
-        return output_shapes[0]
+    flat_output_shapes = tree.flatten(output_shapes)
+    if len(flat_output_shapes) == 1:
+        return flat_output_shapes[0]
     out = str(output_shapes)
     out = out.replace("'", "")
     return out
--- a/keras/src/utils/summary_utils_test.py
+++ b/keras/src/utils/summary_utils_test.py
@@ -124,3 +124,38 @@ class SummaryUtilsTest(testing.TestCase):
         self.assertIn("Total params: 56", summary_content)
         self.assertIn("Trainable params: 56", summary_content)
         self.assertIn("Non-trainable params: 0", summary_content)
+
+    def test_print_model_summary_dict_output(self):
+        """model.summary() should not crash when a layer returns a dict."""
+
+        class DictHead(layers.Layer):
+            def __init__(self):
+                super().__init__()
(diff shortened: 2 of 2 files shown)


keras/src/utils/summary_utils.py
```diff
@@ -111,7 +111,10 @@ def format_shape(shape):
                 )
         except NotImplementedError:
             return "?"
-    if len(output_shapes) == 1:
+    if isinstance(output_shapes, dict):
+        if len(output_shapes) == 1:
+            return next(iter(output_shapes.values()))
+    elif len(output_shapes) == 1:
```
hertschuh: Let's extend this to support any nested structure to any level (you could in theory have nested lists, nested dictionaries etc.)

```python
flat_output_shapes = tree.flatten(output_shapes)
if len(flat_output_shapes) == 1:
    return flat_output_shapes[0]
```

hertschuh (changes requested):
Thanks for working on this!

MarcosAsh:
Thanks for the review! I didnt think of that I updated to use tree.flatten(output_shapes) so it handles any nested structure.

PR #22512, opened 2026-03-26
Fix predict_on_batch silently enabling XLA on GPU for uncompiled models

ssam18:
When calling predict_on_batch() on an uncompiled model on GPU (TF backend), make_predict_function() was hitting self.jit_compile (the property), which lazily auto-resolves to True on GPU for models that support JIT. This quietly baked XLA into the prediction path and caused ~1e-2 numerical differences compared to a direct @tf.function(jit_compile=False) call, while passing fine on CPU since the auto-resolution returns False there.

The fix is a single line change: switch to self._jit_compile (the raw stored attribute, None for uncompiled models) so XLA only kicks in when the user explicitly requests it via compile(jit_compile=True). Added two regression tests in TFTrainerCorrectnessTest to cover the uncompiled and explicitly-compiled cases.

### DIFF
--- a/keras/src/trainers/trainer_test.py
+++ b/keras/src/trainers/trainer_test.py
@@ -2884,19 +2884,25 @@ class TestTrainer(testing.TestCase):
         self.assertLessEqual(tracing_count[0], 2)
 
 
-class TFTrainerCorrectnessTest(test_case.TestCase):
+class TFTrainerCorrectnessTest(test_case.TestCase, parameterized.TestCase):
+    @parameterized.named_parameters(
+        ("uncompiled", None, None),
+        ("compiled_jit_false", {"jit_compile": False}, False),
+    )
     @pytest.mark.skipif(
         backend.backend() != "tensorflow",
         reason="TF-only test",
     )
-    def test_predict_on_batch_consistent_with_direct_call(self):
-        """predict_on_batch() should not auto-enable XLA on uncompiled models.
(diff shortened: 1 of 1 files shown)


ssam18:
I am not sure why api_gen is failing. Any input?

hertschuh:
> When calling predict_on_batch() on an uncompiled model on GPU (TF backend), make_predict_function() was hitting self.jit_compile (the property), which lazily auto-resolves to True on GPU for models that support JIT. This quietly baked XLA into the prediction path and caused ~1e-2 numerical differences compared to a direct @tf.function(jit_compile=False) call, while passing fine on CPU since the auto-resolution returns False there.
> 
> The fix is a single line change: switch to self._jit_compile (the raw stored attribute, None for uncompiled models) so XLA only kicks in when the user explicitly requests it via compile(jit_compile=True). Added two regression tests in TFTrainerCorrectnessTest to cover the uncompiled and explicitly-compiled cases.

All of this is intentional. Keras 3 compiles by default. This applies to `model.fit()` and other `Model` functions.

The lazy auto-resolution is also intentional. A value of `None` for `self._jit_compile` means "figure out if we can auto-enable".

The supported way of turning compilation off is:

```python
# for model.fit
model.compile(jit_compile=False)
model.fit(...)

# for other functions
model.jit_compile = False
model.predict(...)
```

See https://keras.io/guides/migrating_to_keras_3/#jitcompile-is-set-to-true-by-default-on-gpu

Issue #22543, opened 2026-03-26
keras.ops.sort(axis=None) fails on eager tensors instead of flattening and sorting


AGFACBNNR:
### Summary
`keras.ops.sort` with `axis=None` is documented to flatten the tensor before sorting, but on TensorFlow eager tensors it raises an error.

### Reproduction
```python
import keras
import tensorflow as tf

x = tf.constant([[[1, 4, 3], [9, 7, 6]]], dtype=tf.float32)
print(keras.ops.sort(x, axis=None))
```

### Expected result
The tensor should be flattened and returned in ascending order.

### Actual result
An exception is raised:

`Attempt to convert a value (None) with an unsupported type (<class 'NoneType'>) to a Tensor.`

### Notes
The same API call with a `keras.Input` placeholder does not raise this error, so the issue appears specific to eager execution.

whitakercaleb09-dev:
Thank you for reporting this issue with keras.ops.sort.

dhantule:
Hi @AGFACBNNR, thanks for reporting this. 

I have tested this with `keras-nightly`, it seems to be fixed. Please refer this attached [gist](https://colab.sandbox.google.com/gist/dhantule/63ecc689fb769ffed60eac3a9f92562f/keras-ops-sort-axis-none-fails-on-eager-tensors-instead-of-flattening-and-sorting-22543.ipynb).

divyashreepathihalli:
@keras-team /triage

divyashreepathihalli:
@keras-team /triage

Saumay:
Looks like this is already fixed on current master (e94cb07d5) by #22544. Running `keras.ops.sort(tf.constant([[1.,4.,3.]]), axis=None)` now returns a flattened sorted tensor of shape `(3,)` as expected. Probably safe to close.

pctablet505:
Fixed on master. The TensorFlow backend sort implementation reshapes to [-1] and sets axis=0 when axis=None. Verified in backend/tensorflow/numpy.py and ops/numpy_test.py.

PR #22544, opened 2026-03-26
[Fix] keras.ops.sort(axis=None) fails on eager tensors instead of flattening and sorting

ChiragSW:
Fixes issue: #22543


ChiragSW:
Please review, the issue is now fixed.

Issue #22538, opened 2026-03-26
keras.ops.blackman returns NaN for window length 1


AGFACBNNR:
## Summary
`keras.ops.blackman()` produces an invalid result for a valid scalar window length of `1`.

## Reproduction
```python
import keras

x = keras.ops.convert_to_tensor(1)
print(keras.ops.blackman(x))
```

## Expected behavior
A valid 1D Blackman window of length 1 should be returned (e.g. a single finite value), and symbolic inputs should infer a concrete length-1 output shape.

## Actual behavior
- Execution returns `array([nan], dtype=float32)`

divyashreepathihalli:
@keras-team /triage

Whatsonyourmind:
The Blackman window formula is:

```
w(n) = 0.42 - 0.5 * cos(2*pi*n/(N-1)) + 0.08 * cos(4*pi*n/(N-1))
```

When N = 1, the denominator N-1 = 0, producing 0/0 = NaN in the cosine arguments.

The mathematical convention is that a window of length 1 should return `[1.0]` -- a single sample captures the full signal with no tapering. This is consistent across all window functions: Hann, Hamming, Blackman, Kaiser, etc. NumPy, SciPy, and MATLAB all return `[1.0]` for length-1 windows.

The fix is a simple guard at the top:

```python
def blackman(M):
    if M == 1:
        return ops.ones(1)
    # ... existing implementation
```

This should probably be applied to all window functions in `keras.ops` (Hamming, Hann, Bartlett, etc.) as they all share the same N-1 denominator singularity.

hertschuh:
There are several issues with this op currently:

### This op should accept ints and not tensors

(much like `ops.numpy.ones` takes a shape as ints). Eager scalar tensors will typically work, but this op:
- 
- won't be compilable on JAX
- won't be allowed in a Functional graph

In practice, this means that:
   - the `Blackman` operation class should be removed
   - the example code should be changed to:
```python
    >>> keras.ops.blackman(5)
```

### The Nan result

The Nan result for a window of size 1 should be fixed on a per-backend basis.

Saumay:
Can't reproduce on current master (40d635f04). `keras.ops.blackman(1)` now returns `[1.0]`, which matches `numpy.blackman(1)`. No NaN. Probably closable.

hertschuh:
Fixed by https://github.com/keras-team/keras/pull/22748

PR #22545, opened 2026-03-26
[Fix] keras.ops.blackman returns NaN for window length 1

ChiragSW:
Fixes issue: #22538


ChiragSW:

ChiragSW:
@hertschuh is the fix correct or do I have any more changes to do?

PR #22546, opened 2026-03-26
Add unique op

pass-lin:
Provide an implementation of `keras.ops.numpy.unique`
Referencing:
https://docs.jax.dev/en/latest/_autosummary/jax.numpy.unique.html
https://pytorch.org/docs/stable/generated/torch.unique.html

Include backends for TensorFlow, JAX, PyTorch, and NumPy.
However, do not include the OpenVINO backend for now.

### Contributor Agreement

### DIFF
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -3779,11 +3779,8 @@ def unique(
     if is_flatten:
         x = tf.reshape(x, [-1])
         dim = 0
-        if return_counts:
-            y, inverse, counts = tf.unique_with_counts(x)
-        else:
-            y, inverse = tf.unique(x)
-            counts = None
+        y, inverse, counts = tf.unique_with_counts(x)
+
     else:
         ndim = x.shape.rank
         dim = axis + ndim if axis < 0 else axis
@@ -3791,8 +3788,6 @@ def unique(
         y, inverse, counts = tf.raw_ops.UniqueWithCountsV2(
             x=x, axis=axis_to_use, out_idx=tf.int32
         )
-        if not return_counts:
-            counts = None
 
     if sorted:
         num_unique = tf.shape(y)[dim]


keras/src/backend/tensorflow/numpy.py
```diff
@@ -3741,3 +3741,71 @@ def histogram(x, bins=10, range=None):
...
+            sort_indices = tf.range(num_unique, dtype=tf.int32)
+
+            # Stable sort from the last column to the first column
+            # to achieve lexicographical order.
+            for i in range(num_cols - 1, -1, -1):
+                col = tf.gather(y_2d[:, i], sort_indices)
+                # We use stable=True to preserve the order established by
+                # previous column sorts.
+                perm = tf.argsort(col, stable=True)
+                sort_indices = tf.gather(sort_indices, perm)
```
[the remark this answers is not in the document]
pass-lin: tf does not support lexsort, use tf.while_loop instead

pass-lin:
> @pass-lin
> 
> Thanks for working on this! I had actually started working on this a long time ago and didn't finish. I think you can reuse a lot of it though: [master...hertschuh:keras:unique_op](https://github.com/keras-team/keras/compare/master...hertschuh:keras:unique_op)
> 
> Some key comments:
> 
> * Please add the `return_index` argument.
> * Keep the `sorted` argument (mine didn't have it)
> * I'd like to use the JAX signature, i.e. add the `size` and `fill_value` arguments. This is to make it compilable with JAX. Making it compilable makes it way more useful in general. The downside is that it adds complexity for the other backend as you need to pad the result (I have a working implementation for the NumPy backend).
> 
> Also please use the current pattern for unit tests instead of having a separate class.

If we only consider the parameters of the interface,
we can observe that torch.unique ⊆ numpy.unique (version > 2.3) ⊆ jnp.unique.
Therefore, using torch.unique as the standard minimizes bugs, as we only need to adjust the parameter inputs and outputs appropriately.
In contrast, tf.unique is relatively primitive, requiring most functionality to be implemented manually. Consequently, a more complex interface is also more prone to bugs.

Therefore, I recommend using torch.unique as the standard API.

hertschuh:
> > @pass-lin
> > Thanks for working on this! I had actually started working on this a long time ago and didn't finish. I think you can reuse a lot of it though: [master...hertschuh:keras:unique_op](https://github.com/keras-team/keras/compare/master...hertschuh:keras:unique_op)
> > Some key comments:
> > 
> > * Please add the `return_index` argument.
> > * Keep the `sorted` argument (mine didn't have it)
> > * I'd like to use the JAX signature, i.e. add the `size` and `fill_value` arguments. This is to make it compilable with JAX. Making it compilable makes it way more useful in general. The downside is that it adds complexity for the other backend as you need to pad the result (I have a working implementation for the NumPy backend).
> > 
> > Also please use the current pattern for unit tests instead of having a separate class.
> 
> If we only consider the parameters of the interface, we can observe that torch.unique ⊆ numpy.unique (version > 2.3) ⊆ jnp.unique. Therefore, using torch.unique as the standard minimizes bugs, as we only need to adjust the parameter inputs and outputs appropriately. In contrast, tf.unique is relatively primitive, requiring most functionality to be implemented manually. Consequently, a more complex interface is also more prone to bugs.
> 
> Therefore, I recommend using torch.unique as the standard API.

Ok, let's say we remove the `return_index` argument.

I still want the `size` and `fill_value` arguments, otherwise, it makes `unique` unusable on JAX. The padding logic that need to be added would be very similar for Torch and Numpy and is already done in my branch.

pass-lin:
> > > @pass-lin
> > > Thanks for working on this! I had actually started working on this a long time ago and didn't finish. I think you can reuse a lot of it though: [master...hertschuh:keras:unique_op](https://github.com/keras-team/keras/compare/master...hertschuh:keras:unique_op)
> > > Some key comments:
> > > 
> > > * Please add the `return_index` argument.
> > > * Keep the `sorted` argument (mine didn't have it)
> > > * I'd like to use the JAX signature, i.e. add the `size` and `fill_value` arguments. This is to make it compilable with JAX. Making it compilable makes it way more useful in general. The downside is that it adds complexity for the other backend as you need to pad the result (I have a working implementation for the NumPy backend).
> > > 
> > > Also please use the current pattern for unit tests instead of having a separate class.
> > 
> > 
> > If we only consider the parameters of the interface, we can observe that torch.unique ⊆ numpy.unique (version > 2.3) ⊆ jnp.unique. Therefore, using torch.unique as the standard minimizes bugs, as we only need to adjust the parameter inputs and outputs appropriately. In contrast, tf.unique is relatively primitive, requiring most functionality to be implemented manually. Consequently, a more complex interface is also more prone to bugs.
> > Therefore, I recommend using torch.unique as the standard API.
> 
> Ok, let's say we remove the `return_index` argument.
> 
> I still want the `size` and `fill_value` arguments, otherwise, it makes `unique` unusable on JAX. The padding logic that need to be added would be very similar for Torch and Numpy and is already done in my branch.

You are right. I have modified the code and passed the tests. Please review it again.

keras/src/ops/numpy_test.py
```diff
@@ -2302,6 +2302,21 @@ def test_array_split(self):
         self.assertEqual(splits[0].shape, (None, 4))
         self.assertEqual(splits[1].shape, (None, 4))
 
+    @pytest.mark.skipif(
+        keras.config.backend() == "openvino",
+        reason="OpenVINO backend does not support unique op yet.",
+    )
```
hertschuh: Remove decorator, this test doesn't need a concrete implementation, just `compute_output_shape`.

keras/src/ops/numpy_test.py
```diff
@@ -3101,6 +3116,37 @@ def test_array_split(self):
         self.assertEqual(splits[1].shape, (3, 4))
         self.assertEqual(splits[2].shape, (2, 4))
 
+    @pytest.mark.skipif(
+        keras.config.backend() == "openvino",
+        reason="OpenVINO backend does not support unique op yet.",
+    )
```
hertschuh: Remove decorator, this test doesn't need a concrete implementation, just `compute_output_shape`.

keras/src/ops/numpy_test.py
```diff
@@ -3101,6 +3116,37 @@ def test_array_split(self):
...
+        v, inv = knp.unique(x, axis=0, return_inverse=True)
+        # The size of the unique axis is unknown at compile time
+        self.assertEqual(v.shape, (None, 4))
+        # Inverse indices for axis is always 1D with length of that axis
+        self.assertEqual(inv.shape, (2,))
+
+    @pytest.mark.skipif(
+        keras.config.backend() == "openvino",
+        reason="OpenVINO backend does not support unique op yet.",
+    )
```
hertschuh: Remove decorator, this test doesn't need a concrete implementation, just `compute_output_shape`.

keras/src/ops/numpy_test.py
```diff
@@ -6876,6 +6922,184 @@ def test_angle(self):
 
         self.assertAllClose(knp.Angle()(x), np.angle(x))
 
+    @pytest.mark.skipif(
+        keras.config.backend() == "openvino",
+        reason="OpenVINO backend does not support unique op yet.",
+    )
```
hertschuh: Instead of this, add an entry in https://github.com/keras-team/keras/blob/master/keras/src/backend/openvino/excluded_concrete_tests.txt#L88

That's how we keep track of unimplemented OpenVino features. Same for all the other ones below.

keras/src/ops/numpy_test.py
```diff
@@ -6876,6 +6922,184 @@ def test_angle(self):
...
+    def test_unique_basic(self):
+        x = np.array([3, 1, 2, 1, 4, 2])
+        expected_v = np.array([1, 2, 3, 4])
+        self.assertAllClose(knp.unique(x), expected_v)
+
+    @pytest.mark.skipif(
+        keras.config.backend() == "openvino",
+        reason="OpenVINO backend does not support unique op yet.",
+    )
+    def test_unique_full_outputs_1d(self):
```
hertschuh: I think you can just combine all the test methods into a single `test_unique` method, it's just easier to track.

You can have comments per section like `# 1D ouputs` etc.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -3761,3 +3761,114 @@ def histogram(x, bins=10, range=None):
...
+        axis_to_use = tf.constant([0], dtype=tf.int32)
+    else:
+        ndim = x.shape.rank
+        dim = axis + ndim if axis < 0 else axis
+        axis_to_use = tf.constant([dim], dtype=tf.int32)
+
+    # TF native op returns unique elements in appearance order
+    y, idx, count = tf.raw_ops.UniqueWithCountsV2(
+        x=x, axis=axis_to_use, out_idx=tf.int32
+    )
```
hertschuh: The TF implementation is quite complex.

I'm curious why you couldn't use `tf.unique_with_counts`?

pass-lin: > The TF implementation is quite complex.
> 
> I'm curious why you couldn't use ?`tf.unique_with_counts`

I used tf.raw_ops.UniqueWithCountsV2 because tf.unique_with_counts only supports 1-D tensors and cannot handle multi-dimensional inputs with the axis parameter.
I agree that the previous implementation was complex. I have refactored the TensorFlow backend based on your logic to simplify the code while preserving N-D support and ensuring Graph Mode compatibility.

PR #22547, opened 2026-03-26
Fix: resolve ops.padrank error for eager tensors 

### LINKED ISSUE #22540: keras.ops.pad fails on TensorFlow eager tensors when pad_width=0
### Summary
`keras.ops.pad()` works for a Keras symbolic input with `pad_width=0`, but raises a TensorFlow `PadV2` rank error for an eager tensor.

### Reproduction
```python
import keras
import tensorflow as tf

x = tf.constant([[[1., 2.], [3., 4.]]], dtype=tf.float64)

# Works
y1 = keras.ops.pad(keras.Input(shape=(2, 2), batch_size=1), 0, mode="constant", constant_values=0.5)
print(y1.shape)

# Fails
y2 = keras.ops.pad(x, 0, mode="constant", constant_values=0.5)
print(y2)
```

### Expected behavior
`pad_width=0` should be a no-op and return the input tensor unchanged for both eager tensors and Keras symbolic tensors.

### Actual behavior
The eager TensorFlow path fails with:

`The first dimension of paddings must be the rank of inputs [...] [Op:PadV2]`

while the symbolic Keras input returns the original shape correctly.

maitry63:
Fix: This PR addresses a ValueError that occurred when calling keras.ops.pad() with a scalar pad_width on eager tensors with a rank greater than 1.

The fix ensures that the pad_width is properly broadcasted to match the rank of the input tensor, which resolves the issue in eager execution.

Fixes: #22540

Contributor Agreement:

### DIFF
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -66,9 +66,11 @@ def einsum(subscripts, *operands, **kwargs):
     else:
         ov_compute_type = ov_result_type
     inputs = [
-        ov_opset.convert(inp, ov_compute_type).output(0)
-        if inp.get_element_type() != ov_compute_type
-        else inp
+        (
+            ov_opset.convert(inp, ov_compute_type).output(0)
+            if inp.get_element_type() != ov_compute_type
+            else inp
+        )
         for inp in inputs
     ]
     result = ov_opset.einsum(inputs, subscripts).output(0)
@@ -1196,9 +1198,11 @@ def concatenate(xs, axis=0):
         target_type = dtypes.result_type(*keras_types)
         ov_target_type = OPENVINO_DTYPES[target_type]
         elems = [
-            ov_opset.convert(x, ov_target_type).output(0)
-            if x.get_element_type() != ov_target_type
-            else x
+            (
+                ov_opset.convert(x, ov_target_type).output(0)
+                if x.get_element_type() != ov_target_type
+                else x
+            )
             for x in elems
         ]
     res = ov_opset.concat(elems, axis).output(0)
@@ -5354,8 +5358,17 @@ def slogdet(x):
 
 def argpartition(x, kth, axis=-1):
     x = get_ov_output(x)
+    if axis is None:
+        x = ov_opset.reshape(
+            x,
+            ov_opset.constant([-1], dtype="int64"),
+            False,
+        )
+        axis = 0
+
     x_shape = x.get_partial_shape()
     rank = x_shape.rank.get_length()
+
     axis = canonicalize_axis(axis, rank)
     axes = list(range(rank))
     axes[axis], axes[-1] = axes[-1], axes[axis]
@@ -5366,6 +5379,7 @@ def argpartition(x, kth, axis=-1):
         ov_opset.constant(-1),
         ov_opset.constant(0),
     )
+
     if isinstance(kth, int) and kth < 0:
         kth_tensor = ov_opset.add(
             n,
@@ -5373,8 +5387,10 @@ def argpartition(x, kth, axis=-1):
         )
     else:
         kth_tensor = ov_opset.constant(kth, n.get_element_type())
+
     one = ov_opset.constant(1, kth_tensor.get_element_type())
     k_val = ov_opset.add(kth_tensor, one)
+
     bottom_ind = ov_opset.topk(
         ov_opset.negative(x),
         k=k_val,
@@ -5408,13 +5424,16 @@ def argpartition(x, kth, axis=-1):
         sort="value",
     ).output(1)
     result = ov_opset.concat([bottom_ind, top_ind], axis=-1)
+
     inv_axes = [0] * rank
     for i, a in enumerate(axes):
         inv_axes[a] = i
+
     result = ov_opset.transpose(
         result,
         ov_opset.constant(inv_axes),
     ).output(0)
+
     return OpenVINOKerasTensor(result)
 
 
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -66,9 +66,11 @@ def einsum(subscripts, *operands, **kwargs):
     else:
         ov_compute_type = ov_result_type
     inputs = [
-        ov_opset.convert(inp, ov_compute_type).output(0)
-        if inp.get_element_type() != ov_compute_type
-        else inp
+        (
+            ov_opset.convert(inp, ov_compute_type).output(0)
+            if inp.get_element_type() != ov_compute_type
+            else inp
+        )
         for inp in inputs
     ]
     result = ov_opset.einsum(inputs, subscripts).output(0)
@@ -1196,9 +1198,11 @@ def concatenate(xs, axis=0):
         target_type = dtypes.result_type(*keras_types)
         ov_target_type = OPENVINO_DTYPES[target_type]
         elems = [
-            ov_opset.convert(x, ov_target_type).output(0)
-            if x.get_element_type() != ov_target_type
-            else x
+            (
+                ov_opset.convert(x, ov_target_type).output(0)
+                if x.get_element_type() != ov_target_type
+                else x
+            )
             for x in elems
         ]
     res = ov_opset.concat(elems, axis).output(0)
@@ -3704,7 +3708,7 @@ def pad(x, pad_width, mode="constant", constant_values=None):
             constant_values, x.get_element_type()
         ).output(0)
 
-    # Split pad_width into two tensors pads_begin and pads_end
+    # split pad_width into two tensors pads_begin and pads_end
     pads_begin = []
     pads_end = []
     for pads_pair in pad_width:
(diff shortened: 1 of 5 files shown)


keras/src/backend/openvino/numpy.py
```diff
@@ -66,9 +66,11 @@ def einsum(subscripts, *operands, **kwargs):
...
         ov_compute_type = ov_result_type
     inputs = [
-        ov_opset.convert(inp, ov_compute_type).output(0)
-        if inp.get_element_type() != ov_compute_type
-        else inp
+        (
+            ov_opset.convert(inp, ov_compute_type).output(0)
+            if inp.get_element_type() != ov_compute_type
+            else inp
+        )
```
hertschuh: Please revert this file, there is no logic change, mostly formatting.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -1587,10 +1587,18 @@ def expm1(x):
...
+    if not isinstance(axis, (list, tuple)):
+        axis = [axis]
+    return tf.reverse(x, axis)
+
+
+# def flip(x, axis=None):
+#     x = convert_to_tensor(x)
+#     if axis is None:
+#         return tf.reverse(x, tf.range(tf.rank(x)))
+#     return tf.reverse(x, [axis])
```
hertschuh: These changes are unrelated and incorrect, please revert.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -3040,9 +3052,11 @@ def take_along_axis(x, indices, axis=None):
...
         # than make some expensive dynamic checks here.
         broadcast_shape = [
-            tf.maximum(x_original_shape[i], indices_original_shape[i])
-            if dim is None
-            else dim
+            (
+                tf.maximum(x_original_shape[i], indices_original_shape[i])
+                if dim is None
+                else dim
+            )
```
hertschuh: Revert

keras/src/backend/torch/numpy.py
```diff
@@ -1598,14 +1598,20 @@ def pad(x, pad_width, mode="constant", constant_values=None):
...
-        if pad_width_sum == 0:  # early break when no padding in higher order
+    for p in pad_width_list:
+        pad_width_sum += p[0] + p[1]
+
+    for p in pad_width_list:
+        pad_sum += [p[0], p[1]]
+        pad_width_sum -= p[0] + p[1]
+        if (
+            pad_width_sum == 0
+        ):  # Early break when remaining higher dims have 0 padding
```
hertschuh: There is no actual logic change here, revert.

keras/src/ops/numpy.py
```diff
@@ -6474,16 +6474,27 @@ def _process_pad_width(self, pad_width):
...
-        if len(self.pad_width) > 1 and len(self.pad_width) != len(x.shape):
+        if builtins.all(
+            [before == 0 and after == 0 for before, after in self.pad_width]
+        ):
+            return x
+
+        pad_width = self.pad_width
+        x_rank = len(x.shape)
+        if len(pad_width) == 1:
+            if x_rank > 1:
```
hertschuh: `if len(pad_width) == 1 and x_rank > 1:`

hertschuh (changes requested):
Thanks for addressing this!

keras/src/backend/tensorflow/numpy.py
maitry63: Reverted

keras/src/backend/tensorflow/numpy.py
maitry63: Reverted

keras/src/backend/torch/numpy.py
maitry63: Reverted

keras/src/ops/numpy.py
maitry63: Corrected

keras/src/backend/openvino/numpy.py
maitry63: Reverted back

hertschuh (changes requested):
Can you revert all the formatting changes in openvino and tensorflow folders? I wasn't able to locate the actual changes.

hertschuh:
Oh and can you rebase too?

keras/src/backend/openvino/numpy.py
```diff
@@ -3760,25 +3704,20 @@ def pad(x, pad_width, mode="constant", constant_values=None):
...
-    # split pad_width into two tensors pads_begin and pads_end
+    # Split pad_width into two tensors pads_begin and pads_end
     pads_begin = []
     pads_end = []
     for pads_pair in pad_width:
         pads_begin.append(pads_pair[0])
         pads_end.append(pads_pair[1])
     pads_begin = ov_opset.constant(pads_begin, Type.i32).output(0)
     pads_end = ov_opset.constant(pads_end, Type.i32).output(0)
+
```
hertschuh: These are just formatting changes, undo this file.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -2573,16 +2551,11 @@ def pad(x, pad_width, mode="constant", constant_values=None):
                 f"Received: mode={mode}"
             )
         kwargs["constant_values"] = constant_values
+
     pad_width = convert_to_tensor(pad_width, "int32")
```
hertschuh: These are just formatting changes, undo this file.

keras/src/backend/torch/numpy.py
```diff
@@ -1613,14 +1598,20 @@ def pad(x, pad_width, mode="constant", constant_values=None):
...
```
hertschuh: These are mostly formatting changes and a rename. Correct me if I'm wrong, but this doesn't fix the bug, right?

keras/src/ops/numpy.py
```diff
@@ -6616,16 +6474,26 @@ def _process_pad_width(self, pad_width):
...
                 f"Received: pad_width={self.pad_width} "
-                f"(of length {len(self.pad_width)}) and x.shape={x.shape} "
-                f"(of length {len(x.shape)})"
+                f"and x.shape={x.shape}"
             )
+
         return backend.numpy.pad(
             x,
-            pad_width=self.pad_width,
+            pad_width=pad_width,
```
hertschuh, line 6520: This is only supposed to be run in the context of creating a Functional model.

However, this line below is incorrect:
```python
return Pad(pad_width, mode=mode)(x, constant_values=constant_values)
```

It should be:
```python
# pad_width validation here
if any_symbolic_tensors((x, constant_values)):
    return Pad(pad_width, mode=mode).symbolic_call(x, constant_values=constant_values)
return backend.numpy.pad(x, pad_width=self.pad_width, constant_values=constant_values)
```

But once you do that, your fix won't be called anymore. So it needs to move to each backend implementation that needs it (hopefully most backends don't need it).

Then, `_process_pad_width` needs to be extracted and called at the beginning of `def pad`.

hertschuh (changes requested):

keras/src/backend/openvino/numpy.py
maitry63: Reverted

keras/src/backend/tensorflow/numpy.py
maitry63: Reverted

keras/src/backend/torch/numpy.py
maitry63: Yes, corrected the formatting.

keras/src/ops/numpy.py
maitry63: Thank you so much for the feedback.
I understand the concern about separating symbolic and eager execution using any_symbolic_tensors.

In my change, I mainly ensured that `pad_width` (including cases like 0 and shortcuts) is normalized early so both symbolic and eager paths receive consistent inputs and avoid TensorFlow eager errors.

I agree that `_process_pad_width` should be moved to a frontend-level step so normalization happens before backend dispatch, keeping backend implementations simpler and consistent.

keras/src/ops/numpy.py
```diff
@@ -6688,90 +6580,6 @@ def pad(x, pad_width, mode="constant", constant_values=None):
     return Pad(pad_width, mode=mode)(x, constant_values=constant_values)
```
hertschuh, line 6580: This was implemented incorrectly. It should be:
```python
if any_symbolic_tensors((x, constant_values)):
    return Pad(pad_width, mode=mode).symbolic_call(x, constant_values)
return backend.numpy.pad(
      x,
      pad_width=pad_width,
      pad_width=pad_width, 
      mode=mode,
      constant_values=constant_values,
  )
```

However, once you make that fix, it will make it so that your fix doesn't work anymore.

- the part that validates `pad_width` should be here before `if any_symbolic_tensors`. But it should validate without modifying `pad_width`.
- if any backend needs `pad_width` to be modified like `_process_pad_width` does, it should be in the backend specific code. If the same logic is needed by many backends, we can have a utility method shared between them. I would call it `standardize_padding` or something like that rather than `_process_pad_width` and I would put it in `keras/src/ops/operation_utils.py`.
- and for skipping padding when the padding is zero, I would only do that for backends that need it.

hertschuh (changes requested):

keras/src/ops/numpy.py
hertschuh: But there is a bug that remains, which is that the `pad` op doesn't work in a functional model. This still has to be fixed. And when that's done, all your code has to be moved.

Issue #22539, opened 2026-03-26
keras.ops.flip fails for eager tensors when axis is a list of axes


AGFACBNNR:
`keras.ops.flip()` raises a TensorFlow `ReverseV2` error when used with an eager tensor and `axis` provided as a list of axes, even though the [API docs](https://keras.io/api/ops/numpy/#flip-function) say `axis` may be an axis or axes.

### Reproduction
```python
import keras
import tensorflow as tf

x = tf.ones((3, 4, 5))
print(keras.ops.flip(x, axis=[0, 1]))
```

### Expected behavior
The tensor should be flipped along axes 0 and 1, returning a tensor of the same shape.

### Actual behavior
```
Traceback (most recent call last):
  File "/mnt/data2/Users/sunshuo/2026/dltest_3.3/src/test.py", line 5, in <module>
    print(keras.ops.flip(x, axis=[1]))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/data2/Users/sunshuo/anaconda3/envs/dltest_to_raise_issue/lib/python3.12/site-packages/keras/src/ops/numpy.py", line 3338, in flip
    return backend.numpy.flip(x, axis=axis)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/data2/Users/sunshuo/anaconda3/envs/dltest_to_raise_issue/lib/python3.12/site-packages/keras/src/backend/tensorflow/numpy.py", line 1552, in flip
    return tf.reverse(x, [axis])
           ^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/data2/Users/sunshuo/anaconda3/envs/dltest_to_raise_issue/lib/python3.12/site-packages/tensorflow/python/ops/gen_array_ops.py", line 9263, in reverse_v2
    return reverse_v2_eager_fallback(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/data2/Users/sunshuo/anaconda3/envs/dltest_to_raise_issue/lib/python3.12/site-packages/tensorflow/python/ops/gen_array_ops.py", line 9309, in reverse_v2_eager_fallback
    _result = _execute.execute(b"ReverseV2", 1, inputs=_inputs_flat,
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #22548, opened 2026-03-26
[Fix] keras.ops.flip fails for eager tensors when axis is a list of axes

ChiragSW:
Fixes issue: #22539

### DIFF
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -1588,12 +1588,13 @@ def expm1(x):
 
 def flip(x, axis=None):
     x = convert_to_tensor(x)
+    rank = x.shape.rank
     if axis is None:
-        return tf.reverse(x, tf.range(tf.rank(x)))
-    axis = to_tuple_or_list(axis)
-    axis = tf.convert_to_tensor(axis, dtype=tf.int32)
-    axis = tf.where(axis < 0, axis + tf.rank(x), axis)
-    return tf.reverse(x, axis)
+        return tf.reverse(x, tf.range(rank))
+    else:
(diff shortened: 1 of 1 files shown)


ChiragSW:

keerthanakadiri:

ChiragSW:
## changes
Backend update: Refactored the flip call in the TF backend to pass the axis directly to tf.reverse.

Added new test cases in ```keras/ops/math_test.py``` that specifically test flipping with axis=(-1,) and axis=(0, 2) to prevent future regressions.

ChiragSW:

keras/src/backend/tensorflow/numpy.py
```diff
@@ -1590,7 +1590,10 @@ def flip(x, axis=None):
     x = convert_to_tensor(x)
     if axis is None:
         return tf.reverse(x, tf.range(tf.rank(x)))
-    return tf.reverse(x, [axis])
+    axis = to_tuple_or_list(axis)
+    axis = tf.convert_to_tensor(axis, dtype=tf.int32)
+    axis = tf.where(axis < 0, axis + tf.rank(x), axis)
```
hertschuh: Let's handle the axis statically as they're supposed to be ints, not tensors.
Also, `canonicalize_axis` does some validation:

```python
rank = x.shape.rank

if axis is None:
    return tf.reverse(x, tf.range(rank))
else:
    axis = to_tuple_or_list(axis)
    axis = [canonicalize_axis(a, rank) for a in axis]
    return tf.reverse(x, axis)
```

hertschuh (changes requested):
Thanks for looking into this issue!

Issue #22535, opened 2026-03-26
MelSpectrogram skips `fft_length >= sequence_length` validation for symbolic Keras inputs


AGFACBNNR:
## Summary
`keras.layers.MelSpectrogram` validates `fft_length >= sequence_length` for eager tensors, but the check is skipped when the input is a symbolic `keras.Input`. This leads to invalid configurations producing an output instead of an error.

## Reproduction
```python
import keras
import tensorflow as tf

x_eager = tf.random.uniform((2, 16000), dtype=tf.float32)
x_symbolic = keras.Input(shape=(16000,), batch_size=2, dtype=tf.float32)

layer = keras.layers.MelSpectrogram(
    fft_length=256,
    sequence_stride=256,
    sequence_length=512,
    num_mel_bins=80,
)

# Expected to fail
print(layer(x_eager))

# Incorrectly succeeds
print(layer(x_symbolic))
```

## Expected behavior
Both calls should raise an error because `fft_length` must be greater than or equal to `sequence_length`.

## Actual behavior
- Eager tensor: raises
  `fft_length must equal or larger than sequence_length`
- Keras Input: returns output with shape `(2, 80, 63)`

divyashreepathihalli:
@keras-team /triage

dhantule:
Hi @AGFACBNNR, 

PR #22549 was merged which addressed this issue, I have tested your code, it's working fine, please refer this attached [gist](https://colab.sandbox.google.com/gist/dhantule/0e790c823ba78186c9b17c1f895e2e5a/melspectrogram-skips-fft_length-sequence_length-validation-for-symbolic-keras-inputs-22535.ipynb).

pctablet505:
Fixed on master. Validation was moved to MelSpectrogram.init, catching both eager and symbolic inputs when fft_length is less than sequence_length.

PR #22549, opened 2026-03-26
[Fix] MelSpectrogram skips fft_length >= sequence_length validation for symbolic Keras inputs

ChiragSW:
Fixes issue: #22535

### DIFF
--- a/keras/src/layers/preprocessing/mel_spectrogram_test.py
+++ b/keras/src/layers/preprocessing/mel_spectrogram_test.py
@@ -1,4 +1,5 @@
 import numpy as np
+import pytest
 from absl.testing import parameterized
 from tensorflow import data as tf_data
 


ChiragSW:
@hertschuh The issue has been resolved, please review

keerthanakadiri:
> Fixes issue: #22535

ChiragSW:
## Root Cause
The root cause was a validation bypass where the MelSpectrogram layer only checked the fft_length >= sequence_length constraint for eager tensors, allowing invalid configurations to pass silently when using symbolic Keras inputs.

## The Fix
Ensures this logic is applied during the model-building phase by inspecting the shape attribute of symbolic tensors. It raises a ValueError if parameters are mathematically incompatible.

ChiragSW:

ChiragSW:
The tests are failing due to a merge conflict that I have not resolved. Will be working on it

ChiragSW:
The conflict is solved. Please review

hertschuh (approved):
Thanks for the fix!

Issue #22531, opened 2026-03-26
keras.ops.image.extract_patches returns incorrect output shape for channels_first


AGFACBNNR:
## Bug report

`keras.ops.image.extract_patches()` produces inconsistent and incorrect shapes when `data_format="channels_first"` is used.

### Reproduction
```python
import numpy as np
import keras
from keras.layers import Input
from keras.models import Model

x = np.random.random((2, 20, 20, 3)).astype("float32")

# Eager call
print(keras.ops.image.extract_patches(
    x, size=[3, 5], dilation_rate=[1, 1], padding="valid",
    data_format="channels_first"
).shape)

# Symbolic call
inp = Input(shape=(20, 20, 3))
out = keras.ops.image.extract_patches(
    inp, size=[3, 5], dilation_rate=[1, 1], padding="valid",
    data_format="channels_first"
)
print(Model(inp, out).output_shape)
```

### Expected behavior
`extract_patches()` should return a valid patches tensor with the same semantics in eager and symbolic mode, and the inferred static shape should match the runtime output.

### Actual behavior
- Eager output shape: `(2, 20, 20, 3)`
- Static output shape: `(None, 300, 6, 0)`

hertschuh:
What I get with this example is `ValueError: Computed output size would be zero or negative.`.

`(20, 20, 3)` is not a valid "channels_first" shape, `(3, 20, 20)` should be used.

However, this works as expected:
```python
x = np.random.random((2, 3, 20, 20)).astype("float32")

# Eager call
print(keras.ops.image.extract_patches(
    x, size=[3, 5], dilation_rate=[1, 1], padding="valid",
    data_format="channels_first"
).shape)

# Symbolic call
inp = keras.layers.Input(shape=(3, 20, 20))
out = keras.ops.image.extract_patches(
    inp, size=[3, 5], dilation_rate=[1, 1], padding="valid",
    data_format="channels_first"
)
print(keras.models.Model(inp, out).output_shape)
```

```
(2, 45, 6, 4)
(None, 45, 6, 4)
```

PR #22550, opened 2026-03-26
[Fix] keras.ops.image.extract_patches returns incorrect output shape for channels_first

ChiragSW:
Fixes issue: #22531

### DIFF
--- a/keras/src/ops/image.py
+++ b/keras/src/ops/image.py
@@ -801,6 +801,19 @@ def _extract_patches_2d(
     if len(images_shape) == 3:
         images_shape = [1] + images_shape
     out_dim = patch_h * patch_w * channels_in
+    if channels_in is not None:
+        try:
+            compute_conv_output_shape(
+                images_shape,
+                out_dim,
+                (patch_h, patch_w),
+                strides=strides,
+                padding=padding,
+                data_format=data_format,
+                dilation_rate=dilation_rate,
+            )
+        except TypeError:
+            pass
     kernel = backend.numpy.eye(out_dim, dtype=images.dtype)
     kernel = backend.numpy.reshape(
         kernel, (patch_h, patch_w, channels_in, out_dim)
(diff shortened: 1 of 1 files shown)


ChiragSW:

keerthanakadiri:

ChiragSW:
## Root cause

Root cause of the issue is that ```keras.ops.image.extract_patches``` produced inconsistent and incorrect output shapes when using the ```data_format="channels_first"``` configuration, leading to a mismatch between eager execution and symbolic static shapes.

## Fix 

Updated the logic of the ```extract_patches``` op to correctly handle the axis ordering and spatial dimension calculations specifically for the NCHW (channels_first) format, so that the inferred static shape matches the actual runtime output tensor across all execution modes.

ChiragSW:

keras/src/ops/image.py
```diff
@@ -845,14 +892,20 @@ def _extract_patches_3d(
...
+        volumes = backend.numpy.transpose(volumes, (0, 2, 3, 4, 1))
+        conv_data_format = "channels_last"
     patches = backend.nn.conv(
         inputs=volumes,
         kernel=kernel,
         strides=strides,
         padding=padding,
-        data_format=data_format,
+        data_format=conv_data_format,
         dilation_rate=dilation_rate,
```
hertschuh: What this points to is that the tensorflow implementation of `nn.conv` has an issue with the `data_format` (although at first sight the code looks fine). So this is where the bug fix should go, not here.

keras/src/ops/image.py
```diff
@@ -831,11 +860,29 @@ def _extract_patches_3d(
...
+        volumes_shape = [1] + volumes_shape
+    compute_conv_output_shape(
+        volumes_shape,
+        filters=1,
+        kernel_size=(patch_d, patch_h, patch_w),
+        strides=strides,
+        padding=padding,
+        data_format=data_format,
+        dilation_rate=dilation_rate,
+    )
```
hertschuh: You're not using the result of `compute_conv_output_shape`, just remove this.

ChiragSW:

keras/src/backend/tensorflow/nn.py
```diff
@@ -791,6 +791,47 @@ def conv(
...
+            data_format = "channels_last"
+            result = conv(
+                inputs,
+                kernel,
+                strides=strides,
+                padding=padding,
+                data_format=data_format,
+                dilation_rate=dilation_rate,
+            )
+            return tf.transpose(result, (0, 4, 1, 2, 3))
```
hertschuh, line 833: `data_format` is handled below.

What is this intended to fix?

keras/src/ops/image.py
```diff
@@ -795,14 +822,30 @@ def _extract_patches_2d(
...
+    except (ValueError, RuntimeError) as e:
+        message = str(e)
+        message_lower = message.lower()
+        if (
+            "empty output" in message_lower
+            or "computed output size would be zero or negative" in message_lower
+            or "kernel after dilation has size" in message_lower
+            or "kernel size can't be greater than actual input size"
+            in message_lower
+            or "calculated padded input size per channel" in message_lower
```
hertschuh, line 843: This is extremely brittle. I think I'd rather let the exception go through even if the message is a bit unclear in the context of `extract_patches`.

keras/src/ops/image.py
```diff
@@ -784,9 +784,36 @@ def _extract_patches_2d(
...
+        try:
+            compute_conv_output_shape(
+                images_shape,
+                out_dim,
+                (patch_h, patch_w),
+                strides=strides,
+                padding=padding,
+                data_format=data_format,
+                dilation_rate=dilation_rate,
+            )
```
hertschuh, line 814: Please add a comment saying this is just for validating the parameters.

keras/src/ops/image.py
```diff
@@ -784,9 +784,36 @@ def _extract_patches_2d(
...
+        except TypeError:
+            pass
```
hertschuh, line 816: Why ignore `TypeError`? When does it happen?

ChiragSW:
Actually the bug was fixed. This PR is good to close. Sorry for the wrong overhead

PR #22551, opened 2026-03-26
Use `pull_request_target` event for GPU and TPU tests.

hertschuh:
This change https://github.com/keras-team/keras/pull/22504 causes the `kokoro:force-run` label to be removed automatically. However, unlike removing the label manually, it does not trigger the GPU / TPU tests.

### DIFF
--- a/.github/workflows/gpu_tests.yml
+++ b/.github/workflows/gpu_tests.yml
@@ -3,7 +3,7 @@ name: Keras GPU Tests
 on:
   push:
     branches: [master]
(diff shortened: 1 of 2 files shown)


.github/workflows/gpu_tests.yml
sachinprasadhs: actions/checkout@v6.0.2, to be consistent with other actions.

hertschuh: Done

PR #22553, opened 2026-03-26
Configure concurrency on github test workflows.

hertschuh:
**Problem**: if the tests are triggered multiple times, for instance when pushing an update to the PR, the old tests are not cancelled. The runners are consumed for tests that are no longer useful.

**Solution**: configure the `concurrency` of the workflows to cancel the runs on PRs, but not on master and on releases.

Also log the ref in the title of the checkout action for debugging purposes.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -13,6 +13,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: ${{ github.event_name != 'release' && github.ref != 'refs/heads/master' }}
+
 jobs:
   build:
     strategy:
@@ -38,7 +42,9 @@ jobs:
       KERAS_HOME: .github/workflows/config/${{ matrix.backend }}
       JAX_NUM_CPU_DEVICES: ${{ matrix.multi_device && 4 || 1 }}
     steps:
-      - uses: actions/checkout@v6.0.2
+      - name: Checkout ${{ github.ref }}
+        uses: actions/checkout@v6.0.2
+
       - name: Check for changes in keras/src/applications
         uses: dorny/paths-filter@v3
         id: filter
@@ -136,7 +142,8 @@ jobs:
     name: Check the code format
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6.0.2
+      - name: Checkout ${{ github.ref }}
+        uses: actions/checkout@v6.0.2
       - name: Set up Python 3.11
         uses: actions/setup-python@v6
         with:
--- a/.github/workflows/gpu_tests.yml
+++ b/.github/workflows/gpu_tests.yml
@@ -11,6 +11,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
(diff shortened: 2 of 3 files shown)


.github/workflows/gpu_tests.yml
```diff
@@ -11,6 +11,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: ${{ !contains(github.ref, 'release/') && github.ref != 'master' }}
+
```
sachinprasadhs, line 17: This may still cancel the previous run for the push to master.
github.ref is usually a full ref not the branch name.

Any of these would considered safe for master: 
```
github.ref != 'refs/heads/master'
github.ref_name != 'master'
```
For release also, It should be `github.event_name != 'release'`

`cancel-in-progress: ${{ github.event_name != 'release' && github.ref_name != 'master' }}`

.github/workflows/actions.yml
```diff
@@ -13,6 +13,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: ${{ !contains(github.ref, 'release/') && github.ref != 'master' }}
+
```
sachinprasadhs, line 19: This may still cancel the previous run for the push to master.
github.ref is usually a full ref not the branch name.

Any of these would considered safe for master: 
```
github.ref != 'refs/heads/master'
github.ref_name != 'master'
```
For release also, It should be `github.event_name != 'release'`

`cancel-in-progress: ${{ github.event_name != 'release' && github.ref_name != 'master' }}`

.github/workflows/tpu_tests.yml
```diff
@@ -11,6 +11,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
```
sachinprasadhs, line 14: This may still cancel the previous run for the push to master.
github.ref is usually a full ref not the branch name.

Any of these would considered safe for master: 
```
github.ref != 'refs/heads/master'
github.ref_name != 'master'
```
For release also, It should be `github.event_name != 'release'`

`cancel-in-progress: ${{ github.event_name != 'release' && github.ref_name != 'master' }}`

hertschuh: I don't know what the `ref` value is, but based on what JAX does everywhere, I assumed it's the short branch name on a push (i.e. `master`).

https://github.com/jax-ml/jax/blob/b4bd9ef541fe154fe0ed82a238a194533d3d271e/.github/workflows/cloud-tpu-ci-presubmit.yml#L30

The doc is a bit unclear: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context

> For workflows triggered by push, this is the branch or tag ref that was pushed.

I agree with the `github.event_name != 'release'` part though.

.github/workflows/gpu_tests.yml
hertschuh: Done

.github/workflows/actions.yml
hertschuh: Done

.github/workflows/tpu_tests.yml
hertschuh: Done

sachinprasadhs (approved):

PR #22559, opened 2026-03-27
Use float32 for LoRA weights to avoid the risk of underflow and overflow.

james77777778:
## Description
As reported in https://github.com/keras-team/keras-hub/issues/2629

We should use high precision (float32) for LoRA weights to stabilize the finetuning.

References:
- https://huggingface.co/docs/peft/developer_guides/troubleshooting (`autocast_adapter_dtype`)
- https://github.com/huggingface/peft/issues/2421
- https://github.com/huggingface/peft/blob/c75485a2144a15a526e0290835a7439daefc3925/src/peft/tuners/lora/layer.py#L966-L968 (`huggingface/peft` impl)

## Contributor Agreement

### DIFF
--- a/keras/src/layers/convolutional/base_conv.py
+++ b/keras/src/layers/convolutional/base_conv.py
@@ -230,11 +230,17 @@ class BaseConv(Layer):
             raise AttributeError(
                 "You must build the layer before accessing `kernel`."
             )
+        kernel = self._kernel
         if self.lora_enabled:
-            return self._kernel + (
-                self.lora_alpha / self.lora_rank
-            ) * ops.matmul(self.lora_kernel_a, self.lora_kernel_b)
-        return self._kernel
+            kernel = ops.cast(
+                ops.add(
+                    kernel,
+                    (self.lora_alpha / self.lora_rank)
+                    * ops.matmul(self.lora_kernel_a, self.lora_kernel_b),
+                ),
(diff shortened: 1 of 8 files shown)


james77777778:
PR rebased. The openvino test failure should be unrelated to this PR.

keras/src/layers/convolutional/base_conv.py
```diff
@@ -296,16 +299,21 @@ def enable_lora(
                 "lora is already enabled. This can only be done once per layer."
             )
         self._tracker.unlock()
+
```
amitsrivastava78, line 302: The PR itself is fine; just worth adding a note that users should merge LoRA weights before deploying for inference.

amitsrivastava78:
Thanks for the PR, it's very well written, please check my minor comment about this.

keras/src/layers/convolutional/base_conv.py
james77777778: Sure. The comments have been updated:

```python
        # LoRA weights should be float32 to avoid the risk of underflow or
        # overflow during fine-tuning.
        # When deploying the model, these weights should be merged with the
        # original kernel while maintaining the original kernel's dtype.
        ...
```

amitsrivastava78 (approved):

PR #22560, opened 2026-03-27
Fix: keras.ops.argpartition failed when axis==None

### LINKED ISSUE #22537: keras.ops.argpartition failed when axis==None
According to the [document ](https://keras.io/api/ops/numpy/#argpartition-function), the parameter axis should act as:

```
axis: Axis along which to sort. The default is -1 (the last axis). If None, the flattened array is used.
```

However, in the following code:

```python
import keras
import numpy as np

x = np.random.normal((4, 4, 4))
print(keras.ops.argpartition(x, 0, axis=None))
```

The execution fails with the error message:

```
  File "/mnt/data2/Users/sunshuo/2026/dltest_3.3/src/test.py", line 5, in <module>
    print(keras.ops.argpartition(x, 0, axis=None))
  File "/mnt/data2/Users/sunshuo/anaconda3/envs/dltest_to_raise_issue/lib/python3.12/site-packages/keras/src/ops/numpy.py", line 7827, in argpartition
    return backend.numpy.argpartition(x, kth, axis)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/data2/Users/sunshuo/anaconda3/envs/dltest_to_raise_issue/lib/python3.12/site-packages/keras/src/backend/tensorflow/numpy.py", line 3306, in argpartition
    x = swapaxes(x, axis, -1)
        ^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/data2/Users/sunshuo/anaconda3/envs/dltest_to_raise_issue/lib/python3.12/site-packages/keras/src/backend/tensorflow/numpy.py", line 2565, in swapaxes
    axis1 = tf.where(axis1 < 0, tf.add(axis1, x_rank), axis1)
                     ^^^^^^^^^
TypeError: '<' not supported between instances of 'NoneType' and 'int'
```

By the way, it works correctly with KerasTensor:

```python
import keras
import numpy as np

x_sym = keras.layers.Input(shape=(4, 4))
print(keras.ops.argpartition(x_sym, 0, axis=None)) #<KerasTensor shape=(None, 4, 4), dtype=int32, sparse=False, ragged=False, name=keras_tensor_1>
```

maitry63:
This PR fixes `keras.ops.argpartition` fail with axis==None
Fixes: #22537

Contributor Agreement:

### DIFF
--- a/keras/src/backend/jax/numpy.py
+++ b/keras/src/backend/jax/numpy.py
@@ -1628,8 +1628,10 @@ def slogdet(x):
 
 def argpartition(x, kth, axis=-1):
     if axis is None:
-        axis = 0
-    return jnp.argpartition(x, kth, axis)
+        x = jnp.reshape(x, (-1,))
+        return jnp.argpartition(x, kth, axis=0)
+
+    return jnp.argpartition(x, kth, axis=axis)
 
 
 def histogram(x, bins=10, range=None):
--- a/keras/src/backend/jax/numpy.py
+++ b/keras/src/backend/jax/numpy.py
@@ -1627,6 +1627,8 @@ def slogdet(x):
 
 
 def argpartition(x, kth, axis=-1):
+    if axis is None:
+        axis = 0
     return jnp.argpartition(x, kth, axis)
 
 
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -66,11 +66,9 @@ def einsum(subscripts, *operands, **kwargs):
     else:
         ov_compute_type = ov_result_type
     inputs = [
-        (
-            ov_opset.convert(inp, ov_compute_type).output(0)
-            if inp.get_element_type() != ov_compute_type
-            else inp
-        )
+        ov_opset.convert(inp, ov_compute_type).output(0)
+        if inp.get_element_type() != ov_compute_type
+        else inp
         for inp in inputs
     ]
     result = ov_opset.einsum(inputs, subscripts).output(0)
--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -48,7 +48,9 @@ def add(x1, x2):
 def einsum(subscripts, *operands, **kwargs):
     inputs = [get_ov_output(operand) for operand in operands]
     keras_types = [ov_to_keras_type(inp.get_element_type()) for inp in inputs]
-    result_dtype = dtypes.result_type(*keras_types) if keras_types else config.floatx()
+    result_dtype = (
+        dtypes.result_type(*keras_types) if keras_types else config.floatx()
+    )
     if set(keras_types) == {"int8"}:
         result_dtype = "int32"
     ov_result_type = OPENVINO_DTYPES[result_dtype]
@@ -155,7 +157,9 @@ def _compute_extrema(x, operation, axis=None, keepdims=False, initial=None):
         reduction_op = ov_opset.reduce_max
         elementwise_op = ov_opset.maximum
     else:
-        raise ValueError(f"Operation must be 'min' or 'max', received {operation}")
+        raise ValueError(
+            f"Operation must be 'min' or 'max', received {operation}"
+        )
 
     x = get_ov_output(x)
     x_type = x.get_element_type()
@@ -236,7 +240,9 @@ def all(x, axis=None, keepdims=False):
     if axis is None:
         return OpenVINOKerasTensor(x)
     x = ov_opset.convert(x, Type.boolean).output(0)
-    return OpenVINOKerasTensor(ov_opset.reduce_logical_and(x, axis, keepdims).output(0))
+    return OpenVINOKerasTensor(
+        ov_opset.reduce_logical_and(x, axis, keepdims).output(0)
+    )
 
 
 def allclose(x1, x2, rtol=1e-05, atol=1e-08, equal_nan=False):
@@ -249,7 +255,9 @@ def allclose(x1, x2, rtol=1e-05, atol=1e-08, equal_nan=False):
         try:
             return OpenVINOKerasTensor(
                 ov_opset.constant(
-                    np.allclose(x1, x2, rtol=rtol, atol=atol, equal_nan=equal_nan),
+                    np.allclose(
+                        x1, x2, rtol=rtol, atol=atol, equal_nan=equal_nan
+                    ),
                     Type.boolean,
                 ).output(0)
             )
@@ -283,7 +291,9 @@ def any(x, axis=None, keepdims=False):
     if axis is None:
         return OpenVINOKerasTensor(x)
     x = ov_opset.convert(x, Type.boolean).output(0)
-    return OpenVINOKerasTensor(ov_opset.reduce_logical_or(x, axis, keepdims).output(0))
+    return OpenVINOKerasTensor(
+        ov_opset.reduce_logical_or(x, axis, keepdims).output(0)
+    )
 
 
 def amax(x, axis=None, keepdims=False):
@@ -488,7 +498,9 @@ def argmax(x, axis=None, keepdims=False):
     if rank == 0:
         return OpenVINOKerasTensor(ov_opset.constant([0], Type.i32).output(0))
     if axis is None:
-        flatten_shape = ov_opset.constant([-1] + [1] * (rank - 1), Type.i32).output(0)
+        flatten_shape = ov_opset.constant(
+            [-1] + [1] * (rank - 1), Type.i32
+        ).output(0)
         x = ov_opset.reshape(x, flatten_shape, False).output(0)
         axis = 0
         k = ov_opset.constant(1, Type.i32).output(0)
@@ -518,7 +530,9 @@ def argmin(x, axis=None, keepdims=False):
     if rank == 0:
         return OpenVINOKerasTensor(ov_opset.constant([0], Type.i32).output(0))
     if axis is None:
-        flatten_shape = ov_opset.constant([-1] + [1] * (rank - 1), Type.i32).output(0)
+        flatten_shape = ov_opset.constant(
+            [-1] + [1] * (rank - 1), Type.i32
+        ).output(0)
         x = ov_opset.reshape(x, flatten_shape, False).output(0)
         axis = 0
         k = ov_opset.constant(1, Type.i32).output(0)
@@ -654,7 +668,9 @@ def view(x, dtype=None):
 
     if old_itemsize == new_itemsize:
         # Same-width signed↔unsigned: convert preserves bit pattern
-        return OpenVINOKerasTensor(ov_opset.convert(x_ov, new_ov_type).output(0))
+        return OpenVINOKerasTensor(
+            ov_opset.convert(x_ov, new_ov_type).output(0)
+        )
     elif old_itemsize > new_itemsize:
         return _view_int_expand(x_ov, new_ov_type, old_itemsize, new_itemsize)
     else:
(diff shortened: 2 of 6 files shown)


keerthanakadiri:
> ## Code Review
> This pull request implements support for `axis=None` in the `argpartition` operation, which now correctly flattens the input tensor before partitioning. It also includes a fix in the TensorFlow backend to handle scalar updates in `tensor_scatter_nd_update` and improves the error message formatting for the `kth` argument. A review comment suggests moving a local `import math` to the top of the file to comply with PEP 8 standards.

Hi @maitry63, Can you please follow the suggestions.And provide more context about the problem that this PR solves.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -3044,9 +3044,11 @@ def take_along_axis(x, indices, axis=None):
         # we rely on the broacast itself to fail in the incorrect case rather
         # than make some expensive dynamic checks here.
         broadcast_shape = [
-            tf.maximum(x_original_shape[i], indices_original_shape[i])
-            if dim is None
-            else dim
+            (
+                tf.maximum(x_original_shape[i], indices_original_shape[i])
```
hertschuh: Spurious reformatting change. Revert.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -3707,7 +3709,11 @@ def argpartition(x, kth, axis=-1):
     indices = tf.where(mask)
     updates = tf.squeeze(tf.zeros(tf.shape(indices)[0], dtype=tf.int32))
 
-    final_mask = tf.tensor_scatter_nd_update(x, indices, updates)
+    final_mask = tf.tensor_scatter_nd_update(
+        x,
+        indices,
+        tf.reshape(updates, [-1]) if tf.rank(updates) == 0 else updates,
+    )
```
hertschuh: Is this change needed?

keras/src/ops/numpy.py
```diff
@@ -9207,8 +9221,16 @@ def argpartition(x, kth, axis=-1):
...
     """
+
     if any_symbolic_tensors((x,)):
         return Argpartition(kth, axis).symbolic_call(x)
+
+    x = backend.convert_to_tensor(x)
+
+    if axis is None:
+        x = backend.numpy.reshape(x, [-1])
+        return backend.numpy.argpartition(x, kth, axis=0)
```
hertschuh: This is not how this should be fixed. There shouldn't be any code here. Each backend has to support this correctly.

hertschuh (changes requested):
Thanks for looking into this!

keras/src/backend/tensorflow/numpy.py
maitry63: Reverted unnecessary reformatting in tensorflow backend.

keras/src/backend/tensorflow/numpy.py
maitry63: fixed to use `updates = tf.zeros(tf.shape(indices)[0])` to avoid rank-0 errors.

keras/src/ops/numpy.py
maitry63: Removed flattening from ops layer; backend handles flattening for eager tensors; symbolic tensors preserve shape and handled `axis=None` properly in backend functions, removed unnecessary ops-layer reshaping.

keras/src/backend/tensorflow/numpy.py
hertschuh: It's still here.

hertschuh (commented):
Can you look at the failing tests? Thanks!

keras/src/backend/jax/numpy.py
```diff
@@ -1627,6 +1627,8 @@ def slogdet(x):
 
 
 def argpartition(x, kth, axis=-1):
+    if axis is None:
+        axis = 0
```
hertschuh: This will not flatten the array.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -3044,9 +3043,11 @@ def take_along_axis(x, indices, axis=None):
...
         # than make some expensive dynamic checks here.
         broadcast_shape = [
-            tf.maximum(x_original_shape[i], indices_original_shape[i])
-            if dim is None
-            else dim
+            (
+                tf.maximum(x_original_shape[i], indices_original_shape[i])
+                if dim is None
+                else dim
+            )
```
hertschuh: Undo spurious formatting changes.

keras/src/ops/numpy_test.py
```diff
@@ -2237,6 +2237,19 @@ def test_argpartition(self):
...
+        self.assertEqual(out_dynamic.shape, (None, 4, 4))
+        self.assertEqual(out_dynamic.dtype, "int32")
+
+        x_static = KerasTensor((2, 3, 4))
+        out_static = knp.argpartition(x_static, 5, axis=None)
+        self.assertEqual(out_static.shape, (2, 3, 4))
+
+        x_eager = knp.ones((3, 3))
+        out_eager = knp.argpartition(x_eager, 0, axis=None)
+        self.assertEqual(out_eager.shape, (3, 3))
```
hertschuh: The expected output is a 1 dimensional tensor. This means compute_output_shape needs to be fixed too.

hertschuh (changes requested):
The behavior should be that the array is flattened, not that axis 0 is used.

keras/src/backend/jax/numpy.py
maitry63: Thank you for your review. 
Updated to handle `axis=Non`e by flattening before calling `jnp.argpartition`, keeping behavior consistent with NumPy semantics.

keras/src/backend/tensorflow/numpy.py
maitry63: Corrected  & updated to explicitly handle `axis=None` by flattening the tensor and delegating to axis-based logic.

keras/src/ops/numpy_test.py
maitry63: Updated to validate `argpartition` using `pivot-based` correctness checks instead of ordering assumptions, ensuring proper NumPy-aligned behavior.

keras/src/backend/tensorflow/numpy.py
```diff
@@ -3072,9 +3043,11 @@ def take_along_axis(x, indices, axis=None):
...
```
hertschuh: Formatting change, please undo

keras/src/backend/tensorflow/numpy.py
```diff
@@ -3725,23 +3696,24 @@ def slogdet(x):
 
 
 def argpartition(x, kth, axis=-1):
-    x = convert_to_tensor(x, tf.int32)
+    x = tf.convert_to_tensor(x)
+
```
hertschuh: Please use `x = convert_to_tensor(x)` like before, it does some useful things that `tf.convert_to_tensor` doesn't.

keras/src/backend/tensorflow/numpy.py
hertschuh, line 3721: Just curious, why did you also change the implementation?

keras/src/ops/numpy_test.py
```diff
@@ -2328,12 +2233,27 @@ def test_vsplit(self):
...
+
+        self.assertTrue(np.all(left <= kth_value))
+        self.assertTrue(np.all(right >= kth_value))
+
+        self.assertEqual(np.sum(flat_x[result[:kth]] <= kth_value), kth)
+
+        self.assertEqual(
+            np.sum(flat_x[result[kth + 1 :]] >= kth_value),
+            len(flat_x) - kth - 1,
+        )
```
hertschuh: This is not the test where you run actual values. This is the symbolic tensor based test. 

Revert and, in this test, add:

```python
self.assertEqual(knp.argpartition(x, 1, axis=None).shape, (None,))
```

In this test https://github.com/keras-team/keras/blob/master/keras/src/ops/numpy_test.py#L3143 add:

```python
self.assertEqual(knp.argpartition(x, 1, axis=None).shape, (6,))
```

In this test https://github.com/keras-team/keras/blob/master/keras/src/ops/numpy_test.py#L7202 add more with different axes, in particular None.

hertschuh (changes requested):

keras/src/backend/tensorflow/numpy.py
maitry63: corrected

keras/src/backend/tensorflow/numpy.py
maitry63: In this latest implementation, Keras handles computation and shape inference separately.

The backend handles the actual operation, while` compute_output_spec` defines only the output shape.
 
For `axis=None`, the tensor is treated as flattened, so the output shape becomes a single dimension containing all elements. 
Therefore, updating `compute_output_spec` is sufficient to fix the shape issue without changing backend computation.

keras/src/ops/numpy_test.py
maitry63: Updated

keras/src/backend/tensorflow/numpy.py
maitry63: corrected.

keras/src/backend/openvino/numpy.py
```diff
@@ -5468,6 +5354,13 @@ def slogdet(x):
 
 def argpartition(x, kth, axis=-1):
     x = get_ov_output(x)
+    if axis is None:
+        x = ov_opset.reshape(
+            x,
+            ov_opset.constant([-1], dtype="int64"),
+            False,
+        )
```
hertschuh: I think you need `get_ov_output` here:

```python
x = get_ov_output(ov_opset.reshape(...))
```

That hopefully fixes the OpenVino tests.

hertschuh (changes requested):

keras/src/backend/openvino/numpy.py
maitry63: Thank you for the feedback. Wrapped the output of `ov_opset` into `get_ov_output`, now the output is  reusing as Keras Tensor instead of default openvino node.

keras/src/backend/openvino/numpy.py
```diff
@@ -5468,6 +5354,13 @@ def slogdet(x):
 
 def argpartition(x, kth, axis=-1):
     x = get_ov_output(x)
+    if axis is None:
+        x = ov_opset.reshape(
+            x,
+            ov_opset.constant([-1], dtype="int64"),
+            False,
+        ).output(0)
+    axis = 0
```
hertschuh: `axis = 0` should be indented to be in the `if` above.

hertschuh (changes requested):

keras/src/backend/openvino/numpy.py
maitry63: corrected

keras/src/ops/numpy_test.py
hertschuh, line 2248: I guess I missed this before, but this test shouldn't be removed.

It's different from the one line 3014 in that it has dynamic dimensions (None). So please put it back.

hertschuh (changes requested):

keras/src/ops/numpy_test.py
maitry63: corrected

keras/src/ops/numpy_test.py
```diff
@@ -3138,9 +3012,9 @@ def test_dstack(self):
         self.assertEqual(knp.dstack([x, y]).shape, (2, 3, 9))
 
     def test_argpartition(self):
-        x = KerasTensor((2, 3))
-        self.assertEqual(knp.argpartition(x, 3).shape, (2, 3))
-        self.assertEqual(knp.argpartition(x, 1, axis=1).shape, (2, 3))
+        x = KerasTensor((None, 3))
+        self.assertEqual(knp.argpartition(x, 3).shape, (None, 3))
+        self.assertEqual(knp.argpartition(x, 1, axis=1).shape, (None, 3))
```
hertschuh: Undo this. There are 2 separate tests on purpose: one with dynamic dimensions and one with static dimensions.

keras/src/ops/numpy_test.py
hertschuh, line 2248: Revert this (put back the test).

hertschuh (changes requested):

keras/src/ops/numpy_test.py
maitry63: updated

keras/src/ops/numpy_test.py
maitry63: Thank you for the review. Updated as yes, correct 2 tests have separate purpose. 
- one covers dynamic dimensions (None)
- one covers static dimensions (2, 3)

hertschuh (approved):
Thanks!

Issue #22517, opened 2026-03-26
keras.ops.separable_conv rejects valid symbolic inputs with dilated kernels


ningxiudg:
## Summary
`keras.ops.separable_conv()` works with eager tensors, but fails when called with `keras.Input()` placeholders for the same valid shapes.

## Reproduction
```python
import keras
import numpy as np

x = keras.Input(shape=(5, 5, 3))
dw = keras.Input(shape=(3, 3, 3, 1))
pw = keras.Input(shape=(1, 1, 3, 4))

# Works with eager tensors
print(keras.ops.separable_conv(
    np.random.random((2, 5, 5, 3)).astype("float32"),
    np.random.random((3, 3, 3, 1)).astype("float32"),
    np.random.random((1, 1, 3, 4)).astype("float32"),
    dilation_rate=3,
).shape)

# Fails with symbolic tensors
keras.ops.separable_conv(x, dw, pw, dilation_rate=3)
```

## Expected behavior
The symbolic call should succeed and return a valid KerasTensor shape, just like the eager call.

## Actual behavior
The symbolic call raises:

`Kernel shape must have the same length as input, but received kernel of shape (None, 3, 3, 3, 3) and input of shape (None, 5, 5, 3).`

## Notes
- The same inputs and parameters succeed in eager mode.
- The failure appears specific to symbolic/placeholder execution.

dhantule:
Hi @ningxiudg, thanks for reporting this. 

I was able to reproduce this issue with keras 3.13.2 in this colab [gist](https://colab.sandbox.google.com/gist/dhantule/8d816fcf9fba614255b2f81c15d3ceb7/keras-ops-separable_conv-rejects-valid-symbolic-inputs-with-dilated-kernels-22517.ipynb), we'll look into this and update you.

ChiragSW:
Hi @dhantule, can I work on this issue?

dhantule:
> Hi [@dhantule](https://github.com/dhantule), can I work on this issue?

Sure, feel free to raise a PR with a fix.

hertschuh:
@keras-team /triage

hertschuh:
@ningxiudg 

`keras.Input` takes a shape that excludes the batch size. When you say `dw = keras.Input(shape=(3, 3, 3, 1))`, the real shape of the input is `(None, 3, 3, 3, 1)`, e.i. the batch size is added.

So you're not comparing the same shapes in your example. You have to do this to compare the same shapes (take out the first dimension):
[...]

PR #22566, opened 2026-03-27
[Fix] keras.ops.separable_conv rejects valid symbolic inputs with dilated kernels

ChiragSW:
Fixes issue: #22517

## Root cause

keras.Input(shape=...) adds a batch dim, so your kernels become rank-5 like (None, 3, 3, 3, 1) instead of rank-4.
Symbolic shape inference treated that leading None as part of the kernel shape, triggering the kernel-rank mismatch error.

## Fix

In keras/src/ops/nn.py, symbolic Conv/DepthwiseConv shape inference now ignores a leading (None|1) dim from kernels before computing output spec.
In keras/src/ops/operation_utils.py, compute_conv_output_shape now clamps negative/zero VALID outputs to 0 (instead of raising) for consistent symbolic behavior.

### DIFF
--- a/keras/src/ops/operation_utils.py
+++ b/keras/src/ops/operation_utils.py
@@ -255,9 +255,9 @@ def unexpand_kernel_shape(kernel_shape, expected_kernel_rank):
     kernels are not batched; this helper strips such a leading dimension of
     `None` or `1` when the kernel rank is one greater than expected.
     """
-    if (
-        len(kernel_shape) == expected_kernel_rank + 1
-        and kernel_shape[0] in (None, 1)
+    if len(kernel_shape) == expected_kernel_rank + 1 and kernel_shape[0] in (
+        None,
+        1,
     ):
         return kernel_shape[1:]
     return kernel_shape


ChiragSW:
@hertschuh please review the PR

keras/src/layers/convolutional/base_conv.py
```diff
@@ -273,6 +273,19 @@ def compute_output_shape(self, input_shape):
...
+            else:
+                spatial_dims = output_shape[2:]
+            if any(d == 0 for d in spatial_dims if d is not None):
+                raise ValueError(
+                    "Computed output size would be negative. Received: "
+                    f"`inputs.shape={input_shape}`, "
+                    f"`kernel_size={self.kernel_size}`, `dilation_rate="
+                    f"{self.dilation_rate}`, `strides={self.strides}`."
+                )
+        return output_shape
```
hertschuh: Please move this code inside `compute_conv_output_shape` so that all ops can benefit from this check.

keras/src/ops/operation_utils.py
```diff
@@ -216,16 +216,10 @@ def compute_conv_output_shape(
...
             if i not in none_dims and output_spatial_shape[i] <= 0:
-                raise ValueError(
-                    "Computed output size would be zero or negative. Received "
-                    f"`inputs shape={input_shape}`, "
-                    f"`kernel shape={kernel_shape}`, "
-                    f"`dilation_rate={dilation_rate}`, "
-                    f"`strides={strides}`, "
-                    f"`padding={padding}`."
-                )
+                output_spatial_shape[i] = 0
```
hertschuh: Replace with the more correct test that you added in `keras/src/layers/convolutional/base_conv.py`.

And don't return 0 of the dimension is computed as negative.

keras/src/ops/operation_utils.py
```diff
@@ -247,6 +241,13 @@ def compute_conv_output_shape(
     return output_shape
 
 
+def unexpand_kernel_shape(kernel_shape, input_rank):
+    """Removes a leading batch dimension from `kernel_shape` if present."""
+    if len(kernel_shape) == input_rank + 1 and kernel_shape[0] in (None, 1):
+        return kernel_shape[1:]
+    return kernel_shape
```
hertschuh: This part does not make sense to me. The kernel shape is unrelated to whether the batch dimension is present in the inputs or not. There is no "kernel batch dimension".

ChiragSW: In the new commit, I have added the description for this in docstring

ChiragSW:

hertschuh:
> ## Root cause
> keras.Input(shape=...) adds a batch dim, so your kernels become rank-5 like (None, 3, 3, 3, 1) instead of rank-4. Symbolic shape inference treated that leading None as part of the kernel shape, triggering the kernel-rank mismatch error.

You have it here, this is a user error, it's just that `keras.Input` was used incorrectly. We shouldn't make this code change.

PR #22568, opened 2026-03-27
[OpenVINO] Fix convert_to_numpy to handle external tensor types 

goyaladitya05:
`convert_to_numpy` raised `ValueError` for any type it didn't explicitly recognise, including `tf.EagerTensor`. This caused failures in tests that use `tf.data` pipelines, where dataset outputs are EagerTensors passed into `assertAllClose`.

Added a `np.array(x)` fallback before the ValueError. Objects like `tf.EagerTensor` support `__array__` , so this handles them without special-casing. The `ValueError` is still raised if np.array() itself fails.

Enabled 27 now passing tests.

### DIFF
--- a/keras/src/backend/openvino/core.py
+++ b/keras/src/backend/openvino/core.py
@@ -819,8 +819,6 @@ def convert_to_numpy(x):
         else:
             return x.value.data
     if not isinstance(x, OpenVINOKerasTensor):
-        # objects like tf.EagerTensor support the __array__
-        # protocol, so np.array() handles them without special-casing.
         try:
             return np.array(x)
         except Exception as e:


hertschuh:
> `convert_to_numpy` raised `ValueError` for any type it didn't explicitly recognise, including `tf.EagerTensor`. This caused failures in tests that use `tf.data` pipelines, where dataset outputs are EagerTensors passed into `assertAllClose`.
> 
> Added a `np.array(x)` fallback before the ValueError. Objects like `tf.EagerTensor` support `__array__` , so this handles them without special-casing. The `ValueError` is still raised if np.array() itself fails.

This should not happen. The data adapters should use `get_numpy_iterator` which in turn should return NumPy arrays.

`convert_to_numpy` is not expected to handle TF tensors. So something else should be fixed.

goyaladitya05:
> This should not happen. The data adapters should use `get_numpy_iterator` which in turn should return NumPy arrays.
> 
> `convert_to_numpy` is not expected to handle TF tensors. So something else should be fixed.

I was looking at it from the perspective of the tests.

These tests don't go through the trainer or get_numpy_iterator. They iterate tf.data.Dataset directly in the test body, so raw tf.EagerTensors arrive at assertAllClose -> test_case.py:56 -> backend.convert_to_numpy. There's no data adapter in the call path.

The tests were failing because OpenVINO's convert_to_numpy raised ValueError for any unrecognised type instead of falling back to np.array(x), which all other backends do unconditionally as their final line.

This was my analysis of the flow.

1) test_basics_tf
2) for batch in dataset:          # tf.data.Dataset -> raw EagerTensors
3) self.assertAllClose(targets, inputs[:, 0] * 2)
4) test_case.py:56: x1 = backend.convert_to_numpy(x1)  # EagerTensor
5) openvino/core.py: raise ValueError()         # no fallback, as compared to other backends

hertschuh:
Addressing the issue in https://github.com/keras-team/keras/pull/22594

goyaladitya05:
> Addressing the issue in #22594

Alright, thanks.

PR #22587, opened 2026-03-28
Fix cross output shape for symbolic inputs

### LINKED ISSUE #22525: keras.ops.cross returns incorrect symbolic output shape
`keras.ops.cross()` produces a different output shape for Keras symbolic tensors than for eager TensorFlow tensors when `axis=None`.

### Reproduction steps

```python
import keras
import tensorflow as tf


def call_func(inputs, axis=None):
    x1, x2 = inputs
    return keras.ops.cross(x1, x2, axis=axis)

x1 = tf.ones((3, 4, 3))

x2 = tf.ones((3, 4, 3))

print(call_func([x1, x2], axis=None).shape)
print(call_func([keras.Input(shape=(4, 3)), keras.Input(shape=(4, 3))], axis=None).shape)
```

### Expected result

Both eager and symbolic inputs should produce the same output shape, which is `(3, 4, 3)` for this case.

### Actual result

- Eager tensor output shape: `(3, 4, 3)`
- Keras Input output shape: `(None, 3, 4)`

rstar327:
## Summary

Fix `keras.ops.cross` returning incorrect symbolic output shape (e.g. `(3, None)` instead of `(None, 3)`).

Two bugs in `compute_output_spec`:
1. Used `self.axisa` instead of `self.axisb` to index `x2_shape` (line 2296)
2. Negative `axisc` (-1) caused incorrect insertion position via Python slicing — normalized to positive index before inserting

Fixes #22525

### DIFF
--- a/keras/src/ops/numpy.py
+++ b/keras/src/ops/numpy.py
@@ -2314,9 +2314,9 @@ class Cross(Operation):
         else:
             value_size = []
 
-        # Normalize axisc so negative indices insert correctly.
-        ndim = len(output_shape) + len(value_size)
-        axisc = self.axisc if self.axisc >= 0 else ndim + self.axisc
+        axisc = canonicalize_axis(
+            self.axisc, len(output_shape) + len(value_size)
+        )
         output_shape = output_shape[:axisc] + value_size + output_shape[axisc:]
 
         dtype = dtypes.result_type(x1.dtype, x2.dtype)


keerthanakadiri:
> ## Code Review
> This pull request fixes a bug in `compute_output_spec` where the wrong axis was used for the second input's shape and introduces logic to correctly handle negative indices for the output axis. The review feedback suggests using the `canonicalize_axis` utility to simplify the manual axis normalization, ensuring consistency with existing repository patterns.

Hi @rstar327, Can you check this once .Thanks !

PR #22589, opened 2026-03-28
[Fix] Mixed Precision (float16) numerical instability in GroupNormalization with small epsilon

ChiragSW:
## Root Cause

With autocast=True (it is true in default), inputs were cast to float16 before reaching call(). Values exceeding float16's max (65504) overflowed to inf, causing NaN propagation through normalization math. The existing internal float32 upcast couldn't recover already-lost values.

## Fix
```self.autocast = False``` keeps inputs in their original dtype (float32), preventing overflow
```autocast=False``` on gamma/beta weights stores weights in float32 for precision
```ops.cast(outputs, self.compute_dtype)``` returns proper float16 output for mixed precision

I also added regression tests:
1. ```test_large_value_within_autocast_scope``` : verifies weights aren't corrupted by autocast (same test in BatchNormalization and LayerNormalization)
2.  ```test_mixed_float16_large_inputs``` : catches actual NaN bug

### DIFF
--- a/keras/src/layers/normalization/group_normalization_test.py
+++ b/keras/src/layers/normalization/group_normalization_test.py
@@ -197,7 +197,7 @@ class GroupNormalizationTest(testing.TestCase):
                 axis=-1,
                 scale=False,
                 center=False,
-                epsilon=1e-3,
+                epsilon=1e-2,
                 dtype="mixed_bfloat16",
             )
             self.assertFalse(any("epsilon" in str(x.message) for x in w))

### CODE REVIEW

**Comment by ChiragSW** [comment]
@hertschuh please review

**Comment by amadhan882** [comment]
The fix in #22589 correctly addresses the numerical instability I reported in #22586. Disabling autocast to keep internal computations in float32 is the right approach and aligns with other normalization layers. Thanks for the quick fix @ChiragSW!

File: keras/src/layers/normalization/group_normalization.py [hunk 1]
```diff
@@ -81,6 +81,7 @@ def __init__(
     ):
         super().__init__(**kwargs)
         self.supports_masking = True
+        self.autocast = False
```
hertschuh: We should not hardcode `self.autocast = False`. The fix is indeed to do this:

```python
keras.layers.GroupNormalization(groups=8, epsilon=1e-12, autocast=False)
```

But this should be controlled by users, not hardcoded.

The contract of autocast is to accept lower precision to improve speed, and that option should remain open to people who want it.

Now, we could print a warning if the epsilon is lower than the precision, because this is not achievable.

ChiragSW:

keras/src/layers/normalization/group_normalization.py
```diff
@@ -77,9 +83,10 @@ def __init__(
         gamma_regularizer=None,
         beta_constraint=None,
         gamma_constraint=None,
+        autocast=True,
```
hertschuh: Please revert all the changes related to `autocast`. We don't want any behavior change for backwards compatibility. It should stay false by default and people should opt in of they want that behavior.

keras/src/layers/normalization/group_normalization.py
```diff
@@ -129,6 +137,7 @@ def build(self, input_shape):
                 initializer=self.gamma_initializer,
                 regularizer=self.gamma_regularizer,
                 constraint=self.gamma_constraint,
+                autocast=False,
```
hertschuh: Undo this

keras/src/layers/normalization/group_normalization.py
```diff
@@ -140,18 +149,44 @@ def build(self, input_shape):
                 initializer=self.beta_initializer,
                 regularizer=self.beta_regularizer,
                 constraint=self.beta_constraint,
+                autocast=False,
```
hertschuh: Undo this

keras/src/layers/normalization/group_normalization.py
```diff
@@ -140,18 +149,44 @@ def build(self, input_shape):
...
+            return
+        if self.epsilon != 0 and self.epsilon < float(min_pos):
+            warnings.warn(
+                "The configured `epsilon` is smaller than what can be "
+                f"represented in the layer compute dtype ({compute_dtype}); "
+                "it may be rounded to 0 under autocast. Consider increasing "
+                "`epsilon` or setting `autocast=False` for this layer.",
+                stacklevel=3,
+            )
+            self._epsilon_too_small_warning_issued = True
```
hertschuh: Move this into `__init__` and remove `self._epsilon_too_small_warning_issued`.

All of this will work after `super().init()` is called.

keras/src/layers/normalization/group_normalization.py
```diff
@@ -140,18 +149,44 @@ def build(self, input_shape):
...
+        return ops.cast(outputs, self.compute_dtype)
+
+    def _maybe_warn_about_epsilon_precision(self):
+        if self._epsilon_too_small_warning_issued:
+            return
+        if not self.autocast:
+            return
+        compute_dtype = backend.standardize_dtype(self.compute_dtype)
+        if compute_dtype not in ("float16", "bfloat16"):
+            return
```
hertschuh: Remove this. The test on `finfo` below should apply if you're using `float64` vs `float32` for instance.

keras/src/layers/normalization/group_normalization.py
```diff
@@ -140,18 +149,44 @@ def build(self, input_shape):
...
+        try:
+            finfo = ml_dtypes.finfo(compute_dtype)
+            min_pos = getattr(finfo, "smallest_subnormal", finfo.tiny)
```
hertschuh: `smallest_subnormal` and `tiny` are way too small, it would never catch the issue in the bug.

I think you should use `finfo.eps` instead.

keras/src/layers/normalization/group_normalization_test.py
```diff
@@ -175,3 +180,56 @@ def test_broadcasting_2d_channels_first(self):
...
+                groups=2,
+                axis=-1,
+                scale=False,
+                center=False,
+                epsilon=1e-12,
+                dtype="mixed_float16",
+                autocast=False,
+            )
+            _ = layer(x)
+            self.assertFalse(any("epsilon" in str(x.message) for x in w))
```
hertschuh, line 203: What's the difference between this test and the previous one line 212?

keras/src/layers/normalization/group_normalization_test.py
```diff
@@ -175,3 +180,56 @@ def test_broadcasting_2d_channels_first(self):
...
+        x = np.ones((1, 4, 4), dtype="float32")
+
+        with pytest.warns(UserWarning, match="epsilon"):
+            layer = layers.GroupNormalization(
+                groups=2,
+                axis=-1,
+                scale=False,
+                center=False,
+                epsilon=1e-12,
+                dtype="mixed_float16",
```
hertschuh: This combination should work after you use `finfo.eps`:

```
epsilon=1e-4,
dtype="mixed_bfloat16",
```

keras/src/layers/normalization/group_normalization_test.py
```diff
@@ -175,3 +180,56 @@ def test_broadcasting_2d_channels_first(self):
...
+            output = layer(x)
+        output = backend.convert_to_numpy(output)
+        self.assertFalse(np.any(np.isnan(output)))
+
+    def test_autocast_is_user_controllable(self):
+        layer_default = layers.GroupNormalization()
+        self.assertTrue(layer_default.autocast)
+
+        layer_no_autocast = layers.GroupNormalization(autocast=False)
+        self.assertFalse(layer_no_autocast.autocast)
```
hertschuh: Remove these, I don't think these tests exercise anything new.

ChiragSW:
I have applied the reviewed changes, once check @hertschuh please

PR #22591, opened 2026-03-29
[OpenVINO] Remove passing/skipped tests from exclusion list

goyaladitya05:
Removed 82 tests from the exclusion list, which are passing now due to a combination of the recent implementations.

### DIFF
--- a/keras/src/backend/openvino/excluded_concrete_tests.txt
+++ b/keras/src/backend/openvino/excluded_concrete_tests.txt
@@ -3,6 +3,7 @@ AdagradTest::test_correctness_with_golden
 AdamaxTest::test_correctness_with_golden
 AdamTest::test_correctness_with_golden
 AdamWTest::test_correctness_with_golden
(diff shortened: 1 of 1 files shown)


keras/src/backend/openvino/excluded_concrete_tests.txt
```diff
@@ -26,19 +21,7 @@ ConvLSTM2DTest::test_correctness
 ConvLSTMTest::test_correctness
 CTCTest::test_correctness
 CTCTest::test_dtype_arg
-EarlyStoppingTest::test_early_stopping
-EarlyStoppingTest::test_early_stopping_reuse
```
goyaladitya05, line 33: These are the skipped ones, due to the trainable backend decorater.

hertschuh: Let me merge https://github.com/keras-team/keras/pull/22557 first and then rebase this

PR #22593, opened 2026-03-29
Validate axis1 != axis2 in ops.diagonal

### LINKED ISSUE #22528: keras.ops.diagonal fails on eager TensorFlow tensors when axis1 == axis2
`keras.ops.diagonal()` raises a TensorFlow `transpose` error for an eager tensor when `axis1` and `axis2` are the same. The same call on a Keras symbolic input does not fail, but returns an unexpected shape.

### Reproduction

```python
import keras
import tensorflow as tf

x = tf.constant([[[1., 2., 3., 4.],
                  [5., 6., 7., 8.],
                  [9., 10., 11., 12.]],
                 [[13., 14., 15., 16.],
                  [17., 18., 19., 20.],
                  [21., 22., 23., 24.]]])
symbolic_input = keras.Input(shape=(3, 4))
print(keras.ops.diagonal(symbolic_input, offset=-2, axis1=0, axis2=0).shape)
print(keras.ops.diagonal(x, offset=-2, axis1=0, axis2=0))
```

### Output

(3, 4, None)

tensorflow.python.framework.errors_impl.InvalidArgumentError: {{function_node __wrapped__Transpose_device_/job:localhost/replica:0/task:0/device:CPU:0}} transpose expects a vector of size 3. But input(1) is a vector of size 4 [Op:Transpose]

rstar327:
## Summary

Add validation that `axis1` and `axis2` are not the same in `keras.ops.diagonal`, matching NumPy's behavior (`ValueError: axis1 and axis2 cannot be the same`).

Previously, passing `axis1 == axis2` on eager TensorFlow tensors caused a cryptic `InvalidArgumentError` from the transpose op. Now it raises a clear `ValueError` for both symbolic and eager inputs.

Fixes #22528

### DIFF
--- a/keras/src/ops/numpy.py
+++ b/keras/src/ops/numpy.py
@@ -2689,15 +2689,6 @@ class Diagonal(Operation):
                 f"`x` is of shape {x.shape}."
             )
 
-        ndim = len(x_shape)
-        ax1 = canonicalize_axis(self.axis1, ndim)
-        ax2 = canonicalize_axis(self.axis2, ndim)
-        if ax1 == ax2:
-            raise ValueError(
-                "`axis1` and `axis2` cannot be the same. "
-                f"Received: axis1={self.axis1}, axis2={self.axis2}"
-            )
-
         shape_2d = [x_shape[self.axis1], x_shape[self.axis2]]
         x_shape[self.axis1] = -1
         x_shape[self.axis2] = -1
@@ -2765,12 +2756,6 @@ def diagonal(x, offset=0, axis1=0, axis2=1):
     array([[0, 6],
            [1, 7]])
     """
-    if any_symbolic_tensors((x,)):
-        return Diagonal(
-            offset=offset,
-            axis1=axis1,
-            axis2=axis2,
-        ).symbolic_call(x)
     x_ndim = len(x.shape)
     ax1 = canonicalize_axis(axis1, x_ndim)
     ax2 = canonicalize_axis(axis2, x_ndim)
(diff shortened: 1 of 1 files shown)


keras/src/ops/numpy.py
```diff
@@ -2694,6 +2694,15 @@ def compute_output_spec(self, x):
                 f"`x` is of shape {x.shape}."
             )
 
+        ndim = len(x_shape)
+        ax1 = self.axis1 if self.axis1 >= 0 else self.axis1 + ndim
+        ax2 = self.axis2 if self.axis2 >= 0 else self.axis2 + ndim
```
hertschuh: please use `ax1 = canonicalize_axis(self.axis1, ndim)` which does some validation.

keras/src/ops/numpy.py
```diff
@@ -2767,6 +2776,14 @@ def diagonal(x, offset=0, axis1=0, axis2=1):
...
             axis2=axis2,
         ).symbolic_call(x)
+    x_ndim = len(x.shape)
+    ax1 = axis1 if axis1 >= 0 else axis1 + x_ndim
+    ax2 = axis2 if axis2 >= 0 else axis2 + x_ndim
+    if ax1 == ax2:
+        raise ValueError(
+            "`axis1` and `axis2` cannot be the same. "
+            f"Received: axis1={axis1}, axis2={axis2}"
+        )
```
hertschuh, line 2766: The current pattern used is to not have any code in these functions, they're supposed to delegate to the backend functions.

So please move to whichever backend doesn't throw a proper exception (tensorflow is sounds like).

hertschuh: Actually, never mind. Validation is fine. But:
- use `canonicalize_axis` here
- move it before the `if any_symbolic_tensor`
- remove the other one in `compute_output_shape`, it's not longer needed

keras/src/ops/numpy.py
rstar327: Done. Switched to canonicalize_axis in both paths.

keras/src/ops/numpy.py
rstar327: Good call. Moved the validation before the if any_symbolic_tensors check so it runs once for both paths, and removed the duplicate from compute_output_spec.

rstar327:
Done

Issue #22472, opened 2026-03-23
rgb_to_hsv does not validate channel count for Keras Input with channels_first


asfdf-sudo:
Description: The reproduction clearly shows inconsistent behavior for the same API call: eager execution correctly errors on a 4-channel tensor with `data_format='channels_first'`, while the Keras Input placeholder path accepts it and returns a symbolic output shape instead of validating the channel count. This strongly indicates a real shape/validation bug in the static path.

## Summary
`keras.ops.image.rgb_to_hsv()` behaves inconsistently for `data_format="channels_first"`: eager tensors are rejected when the input does not have 3 channels, but a `keras.Input` placeholder is accepted and produces a symbolic output shape.

## Reproduction
```python
import keras
import numpy as np
import tensorflow as tf
import numpy as np
import keras
from keras import ops

def call_func(inputs, data_format=None):
    return keras.ops.image.rgb_to_hsv(images=inputs, data_format=data_format)

# Create test input
test_input = np.random.random((4, 4, 3)).astype(np.float32)

# Test with eager tensor (dynamic)
print("Testing with eager tensor:")
try:
    dynamic_result = call_func(test_input, data_format="channels_first")
    print(f"Dynamic output shape: {dynamic_result.shape}")
except Exception as e:
    print(f"Dynamic execution error: {e}")

# Test with Keras.Input placeholder (static)
print("\nTesting with Keras.Input placeholder:")
try:
    placeholder_input = keras.Input(shape=(4, 3), dtype='float32')
    static_result = call_func(placeholder_input, data_format="channels_first")
    print(f"Static output shape: {static_result.shape}")
except Exception as e:
    print(f"Static execution error: {e}")

print(f"\nInput shape: {test_input.shape}")
print(f"Data format: channels_first")
```
**Colab link** : https://colab.research.google.com/drive/1lsdB_CDC5N0dyHylS7IHvK391-tXSKhe?usp=sharing
## Expected
Both eager and symbolic inputs should validate the channel dimension and raise an error unless the input has 3 channels for `channels_first`.

## Actual
- Eager tensor: `input must have 3 channels but input only has 4 channels`
- `keras.Input` placeholder: returns output shape `(None, 4, 3)` without raising an error
```
Testing with eager tensor:
Dynamic execution error: {{function_node __wrapped__RGBToHSV_device_/job:localhost/replica:0/task:0/device:CPU:0}} input must have 3 channels but input only has 4 channels. [Op:RGBToHSV] name: 

[...]

PR #22595, opened 2026-03-31
[Fix] issues mentioned in comments of #22472

ChiragSW:
Fixes issues mentioned in comments of #22472 
```
Actually, we have discovered that many APIs in the keras.ops module exhibit the issue described above—they lack validity checks when accepting Keras.Input as an input.
These include:

keras.ops.flip
keras.ops.log_softmax
keras.ops.sparse_categorical_crossentropy
keras.ops.roll
keras.ops.sparsemax
keras.ops.trace
keras.ops.image.pad_images
keras.ops.image.crop_images
```
## Root cause
keras.ops functions branch like:
if any_symbolic_tensors(inputs): call Operation(...).symbolic_call() which relies on compute_output_spec() for validation
else: run backend/eager code paths where the real argument checks happen
So any checks that only existed in the eager helper/backend path were skipped for keras.Input.

### keras.ops.image.pad_images / crop_images (```keras/src/ops/image.py```)

1. ```pad_images()```: added the missing “must specify exactly two of …” argument validation in the public wrapper so it runs for both eager and symbolic inputs.
2. ```crop_images()```: added the same “exactly two of …” validation in the wrapper.
3. ```CropImages.compute_output_spec()```: added validation by inferring missing crop amounts when input_height/input_width are known, and raising if the inferred values would be negative.

### keras.ops.flip, keras.ops.roll, keras.ops.trace (```keras/src/ops/numpy.py```)

1. ```flip()```: validate axis is None, int, or a sequence of ints.
2. ```roll()```: validate axis/shift compatibility.
3. ```trace()```: validate axis1 != axis2 during symbolic shape inference.

### keras.ops.log_softmax, keras.ops.sparsemax, keras.ops.sparse_categorical_crossentropy (```keras/src/ops/nn.py```)

1. ```log_softmax()``` and ```sparsemax()```: validate axis bounds against rank when rank is known (so axis=-3 for a rank-2 input raises for symbolic too)
2. ```sparse_categorical_crossentropy()```: enforce the existing constraint axis == -1 in the wrapper so symbolic inputs don’t bypass it.

### DIFF
--- a/keras/src/ops/image.py
+++ b/keras/src/ops/image.py
@@ -10,13 +10,13 @@ from keras.src.ops.operation_utils import compute_conv_output_shape
 def _validate_image_input_rank_for_keras_input(images):
     if (
         backend.is_keras_tensor(images)
-        and len(images.shape) == 3
         and hasattr(images, "_keras_history")
-        and getattr(images._keras_history.operation, "is_input", False)
+        and images._keras_history.operation.__class__.__name__ == "InputLayer"
+        and len(images.shape) - 1 not in (3, 4)
     ):
         raise ValueError(
-            "Invalid images rank: expected rank 4 for a Keras Input image "
-            "batch. Received input with shape: "
+            "Invalid images rank: expected rank 3 (single image) "
+            "or rank 4 (batch of images). Received input with shape: "
             f"images.shape={images.shape}"
         )
 
--- a/keras/src/ops/image.py
+++ b/keras/src/ops/image.py
@@ -7,6 +7,20 @@ from keras.src.ops.operation import Operation
 from keras.src.ops.operation_utils import compute_conv_output_shape
 
 
+def _validate_image_input_rank_for_keras_input(images):
+    if (
+        backend.is_keras_tensor(images)
+        and len(images.shape) == 3
+        and hasattr(images, "_keras_history")
+        and getattr(images._keras_history.operation, "is_input", False)
+    ):
+        raise ValueError(
+            "Invalid images rank: expected rank 4 for a Keras Input image "
+            "batch. Received input with shape: "
+            f"images.shape={images.shape}"
+        )
+
+
 class RGBToGrayscale(Operation):
     def __init__(self, data_format=None, *, name=None):
         super().__init__(name=name)
@@ -1138,6 +1152,7 @@ def pad_images(
             f"target_width={target_width}"
         )
 
+    _validate_image_input_rank_for_keras_input(images)
     if any_symbolic_tensors((images,)):
         return PadImages(
             top_padding,
@@ -1434,6 +1449,7 @@ def crop_images(
             f"target_width={target_width}"
         )
 
+    _validate_image_input_rank_for_keras_input(images)
     if any_symbolic_tensors((images,)):
         return CropImages(
             top_cropping,
--- a/keras/src/ops/image.py
+++ b/keras/src/ops/image.py
@@ -1115,6 +1115,23 @@ def pad_images(
     >>> padded_batch.shape
     (2, 20, 30, 3)"""
 
+    if [top_padding, bottom_padding, target_height].count(None) != 1:
+        raise ValueError(
+            "Must specify exactly two of "
+            "top_padding, bottom_padding, target_height. "
+            f"Received: top_padding={top_padding}, "
+            f"bottom_padding={bottom_padding}, "
+            f"target_height={target_height}"
+        )
+    if [left_padding, right_padding, target_width].count(None) != 1:
+        raise ValueError(
+            "Must specify exactly two of "
+            "left_padding, right_padding, target_width. "
+            f"Received: left_padding={left_padding}, "
+            f"right_padding={right_padding}, "
+            f"target_width={target_width}"
+        )
+
     if any_symbolic_tensors((images,)):
         return PadImages(
             top_padding,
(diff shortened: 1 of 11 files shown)


keras/src/ops/nn.py
```diff
@@ -1014,6 +1014,28 @@ def log_softmax(x, axis=-1):
     array([-2.40760596, -1.40760596, -0.40760596], shape=(3,), dtype=float64)
 
     """
+    ndim = len(getattr(x, "shape", []))
+    if isinstance(axis, int):
+        if axis < -ndim or axis >= ndim:
+            raise ValueError(
+                f"axis {axis} is out of bounds for array of dimension {ndim}"
+            )
+        axis = axis if axis >= 0 else axis + ndim
```
hertschuh: Use `canonicalize_axis` for this.

ChiragSW:
There are some crashes and bugs that need to be fixed. I will look into it soon

ChiragSW:
## Summary of new changes
- Updated many ops to stop using len(getattr(x, "shape", [])) for rank detection, since it breaks on Python lists (no .shape) and can crash when x.shape is None for unknown-rank symbolic tensors. 
- For ```log_softmax```, ```sparsemax```, ```flip```, ```roll```, and ```trace```,  now converted x to a tensor and only does canonicalization when the rank is actually known, so those failures donot happen. 
- Also fixed roll’s validation so a single integer shift is allowed when axis is a tuple or list. 
- ```Trace.compute_output_spec``` now handles unknown-rank symbolic inputs by raising an error and not crashing.

ChiragSW:
The changes resolve the issues. Please review @keerthanakadiri @hertschuh

hertschuh (changes requested):
I'm not following the context of this. It seems unrelated to the linked bug.

Are we missing some validation for the axis argument in some places?

If so, can you add tests with `self.assertRaises` to demonstrate the problem.

Also, there should never be any code in-between `if any_symbolic_tensor(...)` and the call to `symbolic_call`. Some validation can happen before the `if`. But some validation has to happen in the backend specific implementation after `convert_to_tensor`.

ChiragSW:
Problem is that Keras ops have two execution paths, symbolic and eager. Validation checks only lived in the eager path, so invalid inputs via keras.Input would silently pass through without any error.

Fix: 
- Run shared validation before the any_symbolic_tensors branch when rank is statically knowable
- Keep validation after convert_to_tensor on the eager path
- For purely symbolic cases, put validation inside compute_output_spec

Tests added
Across nn_test.py, numpy_test.py, image_test.py, and operation_utils_test.py, all asserting that bad axis or rank inputs on keras.Input now correctly raise ValueError.

ChiragSW:
The issue has been fixed. Please review @hertschuh

keras/src/ops/image.py
```diff
@@ -7,6 +7,20 @@
 from keras.src.ops.operation_utils import compute_conv_output_shape
 
 
+def _validate_image_input_rank_for_keras_input(images):
+    if (
+        backend.is_keras_tensor(images)
+        and hasattr(images, "_keras_history")
+        and images._keras_history.operation.__class__.__name__ == "InputLayer"
```
hertschuh, line 14: I don't understand why you are looking at the keras history, you should never have to do that. Also, you shouldn't even care if it's a keras tensor or a normal tensor.

keras/src/ops/image_test.py
```diff
@@ -7,6 +7,7 @@
 import tensorflow as tf
 from absl.testing import parameterized
 
+import keras
 from keras.src import backend
```
hertschuh, line 11: Do not `import keras` in unit test, please import the feature from `keras.src`

hertschuh (changes requested):
First, this PR is combining a few unrelated things into one. Please split this PR in at least 3 separate PRs:

- the image padding / cropping validation
- the axis verification / canonicalization
- the RNG seed generator changes (and I don't understand the context of these changes)

Then, I see a lot of code duplication like this:
```python
    if axis is not None:
        ndim = get_static_tensor_ndim(x)
        if isinstance(axis, int):
            if ndim is not None:
                canonicalize_axis(axis, ndim)
        elif isinstance(axis, (tuple, list)):
            for a in axis:
                if not isinstance(a, int):
                    raise TypeError(
                        "Argument `axis` must be an integer or a sequence "
                        f"of integers. Received: axis={axis}"
                    )
                if ndim is not None:
                    canonicalize_axis(a, ndim)
        else:
            raise TypeError(
                "Argument `axis` must be an integer, a sequence of "
                f"integers, or `None`. Received: axis={axis}"
            )
```

Please create a helper function called `canonicalize_axes` after `canonicalize_axis`. It will take an int or a list of ints and always return a tuple of ints after doing the validation.

Next, see this pattern a lot:

```python
# Some validation
if any_symbolic_tensors(...)
   ...
x = convert_to_tensor(x)
# The same validation
```

This is not a pattern that we should use:
- It creates a ton of code duplication
- We should let the backend implementation do the `x = convert_to_tensor(x)` part, sometimes the backend implementation needs to look at the type or value of x before converting it to a tensor

If you can't do the validation because you don't fully know the shape of the input, it means it shouldn't be here. It's ok to move the validation in the backend specific implementations (as long as it's factored as one-liner) and in `symbolic_call`.

ChiragSW:
For now I will close this PR and put up 3 separate PRs. I will add the details in what each PR fixes. Should I go ahead with this process @hertschuh ?
The 3 PRs will be:
1. image padding / cropping validation
2. axis verification / canonicalization
3. RNG seed generator changes

hertschuh:
> For now I will close this PR and put up 3 separate PRs. I will add the details in what each PR fixes. Should I go ahead with this process @hertschuh ? The 3 PRs will be:
> 
> 1. image padding / cropping validation
> 2. axis verification / canonicalization
> 3. RNG seed generator changes

Yes, thank you!

PR #22598, opened 2026-04-01
Remove backend-specific strides + dilation_rate restriction from DepthwiseConv and SeparableConv

ssam18:
The strides > 1 + dilation_rate > 1 combination is not universally unsupported. It only fails on the TensorFlow backend. The previous validation in BaseDepthwiseConv.__init__ and BaseSeparableConv.__init__ was raising a ValueError for all backends, which was too strict. This PR removes that blanket restriction and its associated docstring notes and tests, allowing JAX, PyTorch, and other backends to use the combination freely. Also adds compute_output_shape validation in BaseSeparableConv.build() so invalid output dimensions are still caught at build time regardless of backend.

## Contributor Agreement

### DIFF
--- a/keras/src/layers/convolutional/base_depthwise_conv.py
+++ b/keras/src/layers/convolutional/base_depthwise_conv.py
@@ -46,7 +46,6 @@ class BaseDepthwiseConv(Layer):
         strides: int or tuple/list of `rank` integers, specifying the stride
             length of the depthwise convolution. If only one int is specified,
             the same stride size will be used for all dimensions.
-
         padding: string, either `"valid"` or `"same"` (case-insensitive).
             `"valid"` means no padding. `"same"` results in padding evenly to
             the left/right or up/down of the input. When `padding="same"` and
--- a/keras/src/layers/convolutional/base_depthwise_conv.py
+++ b/keras/src/layers/convolutional/base_depthwise_conv.py
@@ -148,8 +148,6 @@ class BaseDepthwiseConv(Layer):
                 f"strides={self.strides}"
             )
 
-
-
     def build(self, input_shape):
         if self.data_format == "channels_last":
             channel_axis = -1
--- a/keras/src/layers/convolutional/base_depthwise_conv.py
+++ b/keras/src/layers/convolutional/base_depthwise_conv.py
@@ -46,7 +46,7 @@ class BaseDepthwiseConv(Layer):
         strides: int or tuple/list of `rank` integers, specifying the stride
             length of the depthwise convolution. If only one int is specified,
             the same stride size will be used for all dimensions.
-            `strides > 1` is incompatible with `dilation_rate > 1`.
+
         padding: string, either `"valid"` or `"same"` (case-insensitive).
             `"valid"` means no padding. `"same"` results in padding evenly to
             the left/right or up/down of the input. When `padding="same"` and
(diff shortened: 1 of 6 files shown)


hertschuh (commented):
What it looks like to me is that it was working on all backends except Tensorflow. So I would argue that we should do the reverse, i.e. remove the check on `BaseDepthwiseConv` and `BaseSeparableConv` and accept the fact that it doesn't work on Tensorflow. I think this check was inherited from Keras 2, which was TensorFlow only.

ssam18:
@hertschuh Good point! I have reversed the approach and removed the global validation from depthwise_conv and separable_conv. The TF-specific skip is restored in the correctness tests, and the shape tests are back to asserting the correct output shapes for backends that support the combination.

hertschuh:
@ssam18 

This PR is now empty.

SamareshSingh:
Forgot to push the changes. Will do that soon

ssam18:
I am seeing failure from OpenVino that is not related to the change in this PR.  
@hertschuh can you re-run the flaky test_fft2 test?

keras/src/layers/convolutional/base_depthwise_conv.py
```diff
@@ -46,7 +46,7 @@ class BaseDepthwiseConv(Layer):
         strides: int or tuple/list of `rank` integers, specifying the stride
             length of the depthwise convolution. If only one int is specified,
             the same stride size will be used for all dimensions.
-            `strides > 1` is incompatible with `dilation_rate > 1`.
+
```
hertschuh: Remove empty line here, it will break the rendering of the documentation.

keras/src/layers/convolutional/base_separable_conv.py
```diff
@@ -33,7 +33,7 @@ class BaseSeparableConv(Layer):
         strides: int or tuple/list of `rank` integers, specifying the stride
             length of the depthwise convolution. If only one int is specified,
             the same stride size will be used for all dimensions.
-            `stride value != 1` is incompatible with `dilation_rate != 1`.
+
```
hertschuh: Remove empty line here, it will break the rendering of the documentation.

hertschuh (commented):
A couple very small tweaks:

Also, can you rebase?

keras/src/layers/convolutional/base_separable_conv.py
ssam18: done

keras/src/layers/convolutional/separable_conv_test.py
```diff
@@ -233,7 +219,6 @@ def test_invalid_output_shape_raises(self):
         ):
             layer2(x2)
 
-
```
hertschuh, line 236: Undo (you removed an empty line that's required to follow the code format standard)

keras/src/layers/convolutional/separable_conv_test.py
```diff
@@ -189,20 +189,6 @@ def test_bad_init_args(self):
...
-            r"`dilation_rate > 1`. Received: strides=\(2, 2\) and "
-            r"dilation_rate=\(2, 1\)",
-        ):
-            layers.SeparableConv2D(
-                depth_multiplier=2,
-                filters=2,
-                kernel_size=(2, 2),
-                strides=2,
-                dilation_rate=(2, 1),
-            )
```
hertschuh: Remove on extra empty line here.

hertschuh (commented):
Can you update the PR description and title? It's no longer what this is.

A couple of things don't conform to the code format:

ssam18:
> Can you update the PR description and title? It's no longer what this is.
> 
> A couple of things don't conform to the code format:

It is done. Please check

hertschuh (approved):
Thanks!

PR #22631, opened 2026-04-03
Use jax.random.split for SeedGenerator state advancement

### LINKED ISSUE #18426: current rng setup is full of footguns in jax
right now unseeded calls to e.g. `keras.random.uniform` are going to acquire static seeds at trace time. this has a few undesirable consequences:

1) subsequent calls will have the same randomness each time (e.g. dropout will have a fixed mask instead of random each step)
2) the jax compiler cache will ~never hit, as the constant rng seed values will be different every time

to get around this, some kind of rng state management is necessary. flax does this with hierarchical management of rng's from the `Scope`. such an approach is fairly complex however, and there might be simpler options e.g. a single global `rng` state, which gets included with the training state in `model.fit`, unseeded rng calls would then do something along the lines of

```
state.seed, local_seed = jax.random.split(state.seed)
```

MarcosAsh:
SeedGenerator.next() previously used a custom counter to advance RNG state. This is a known footgun on JAX where proper key splitting via jax.random.split is needed for statistical independence between successive keys.

This adds a split_seed backend op and refactors SeedGenerator.next() to use it. On JAX it calls jax.random.split for proper PRNG key splitting. Other backends keep the existing counter-based approach since their stateless random ops handle entropy from any unique seed pair.

No public API changes. Non-JAX backends produce identical seed sequences as before.

Fixes #18426

## Contributor Agreement

### DIFF
--- a/keras/src/backend/jax/random.py
+++ b/keras/src/backend/jax/random.py
@@ -6,6 +6,11 @@ from keras.src.random.seed_generator import draw_seed
 from keras.src.random.seed_generator import make_default_seed
 
 
+def split_seed(seed, ordered=True):
+    keys = jax.random.split(seed)
+    return keys[0], keys[1]
+
+
 def jax_draw_seed(seed):
     if isinstance(seed, jax.Array):
         return seed
--- a/keras/src/backend/numpy/random.py
+++ b/keras/src/backend/numpy/random.py
@@ -7,6 +7,16 @@ from keras.src.random.seed_generator import draw_seed
 from keras.src.random.seed_generator import make_default_seed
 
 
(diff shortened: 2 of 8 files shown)


hertschuh:
Two comments about this:
- The first concern about this change is that it breaks backwards compatibility and cross-backend compatibility. I.e. the same code, even when seeded, will no longer produce the same results with the new version of Keras and no longer produce the same results across backend.
- The other comment is that it doesn't address the core of https://github.com/keras-team/keras/issues/18426 which is that it is very easy to have hardcoded seeds in JAX without realizing. But fixing this is a deeper change, and probably also a breaking change.

PR #22638, opened 2026-04-05
Adding Torch Support for Model Parallel & Data Parallel 

buildwithsuhana:
This PR introduces comprehensive support for distributed training in the PyTorch backend, enabling both Data Parallelism (DP) and Model Parallelism (MP). These changes align the PyTorch backend with the Keras 3 distribution API, allowing users to scale models across multiple devices and hosts with minimal code changes.

Key Implementations:

Data Parallelism (DP) via Native DDP
Leverages torch.nn.parallel.DistributedDataParallel (DDP) for optimized data-parallel training.
Introduces _KerasModuleWrapper to present Keras models as standard torch.nn.Module objects to DDP, enabling gradient bucketing and overlapping compute with communication.
Automatically injects torch.utils.data.distributed.DistributedSampler into data pipelines to ensure correct data sharding across ranks.
Model Parallelism (MP) via DTensor
Integrates PyTorch DTensor (torch.distributed.tensor) to support weight sharding.
Updates the Variable class to be "distribution-aware," allowing it to query the active ModelParallel strategy and initialize as a sharded DTensor instead of a local Parameter.

Colab notebook with comparison for model parallel and data parallel with jax and torch: [Colab link](https://colab.research.google.com/drive/1l9dLZFwqTzdTBZyKMM080guHAFLyf8r4?resourcekey=0-gmB6h0xmfV0p3MmmTyXtiQ&usp=sharing)

## Contributor Agreement


keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,222 @@
...
+    ]
+
+
+# Patch DTensor.unbind: PyTorch's DTensor lacks a registered sharding strategy
+# for unbind, which breaks tensor iteration in embedding layers.
+def _dtensor_unbind_patched(self, dim=0):
+    return unbind_dtensor(self, dim=dim)
+
+
+DTensor.unbind = _dtensor_unbind_patched
```
amitsrivastava78: This is a module-level side effect , it fires on import, affecting all code in the process. Even though unbind_dtensor() was extracted as a utility, this patch is still applied globally. The patch should be removed

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+def process_id():
+    """Get the rank of the current process."""
+    if torch.distributed.is_initialized():
+        return torch.distributed.get_rank()
+    return 0
+
+
+def _to_backend_device(device_name):
```
amitsrivastava78: The device_name parameter is accepted but never used

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,222 @@
...
+        return tensor.redistribute(
+            device_mesh=torch_mesh, placements=placements
+        )
+
+    return torch.distributed.tensor.distribute_tensor(
+        tensor, device_mesh=torch_mesh, placements=placements
+    )
+
+
+def distribute_variable(value, layout, trainable=True):
```
amitsrivastava78: When there's no ModelParallel distribution, the variable is moved to _to_backend_device(None) (always CUDA local rank). This will move the tensor away from the device it was already on, which can break CPU-only workflows or multi-GPU setups where the device is managed externally.

keras/src/backend/torch/core.py
```diff
@@ -205,47 +233,72 @@ def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
...
+        and isinstance(dist, dist_lib.ModelParallel)
+        and is_tensor(x)
+    ):
+        from torch.distributed.tensor import DTensor
+
+        if not isinstance(x, DTensor):
+            from keras.src.distribution import TensorLayout
+
+            layout = TensorLayout([None] * x.ndim, dist.device_mesh)
+            x = dist_lib.distribute_tensor(x, layout)
```
amitsrivastava78: convert_to_tensor is a hot path called on every tensor operation. Adding a dist_lib.distribution() call + isinstance check on every invocation introduces overhead even when no distribution is active. This should be guarded with a fast early-exit

keras/src/backend/torch/trainer.py
```diff
@@ -160,6 +289,22 @@ def make_predict_function(self, force=False):
                 f"Received: steps_per_execution={self.steps_per_execution}"
             )
 
+        from keras.src.distribution import distribution_lib as dist_lib
+
+        dist = dist_lib.distribution()
+        if dist is not None and isinstance(dist, dist_lib.DataParallel):
```
amitsrivastava78: This identical block appears 3 times. Extract it into a _setup_ddp() helper called once (e.g. from compile() or lazily via @property). Also object.__setattr__ bypasses Keras's attribute tracking which could cause subtle bugs when saving/restoring model state.

keras/src/backend/torch/trainer.py
```diff
@@ -103,12 +167,45 @@ def test_step(self, data):
...
+                for variable in metric.variables:
+                    val = variable.value
+                    if isinstance(val, DTensor):
+                        local_val = val.to_local()
+                        torch.distributed.all_reduce(
+                            local_val, op=torch.distributed.ReduceOp.SUM
+                        )
+                        variable.assign(local_val)
+                    else:
+                        torch.distributed.all_reduce(
```
amitsrivastava78: When val is a DTensor, you do all_reduce on the local shard and then assign back — but the variable's underlying storage is still a DTensor. The variable.assign(local_val) will try to assign a plain local tensor into a DTensor-backed variable, which may silently work through .copy_() but the semantics are incorrect — the variable will no longer be distributed. Also always using SUM is wrong for metrics like MeanAbsoluteError which track a sum and a count separately; the correct aggregate for "mean" is SUM of numerator and denominator individually, but custom metrics using MAX or MIN will produce wrong results.

keras/src/trainers/data_adapters/py_dataset_adapter.py
```diff
@@ -252,14 +265,22 @@ def _standardize_batch(self, batch):
 
     def _infinite_generator(self):
         for i in itertools.count():
-            yield self._standardize_batch(self.py_dataset[i])
+            if i % self._num_processes == self._process_id:
```
amitsrivastava78: This still calls self.py_dataset[i] for every index even when the index is skipped — the modulo check guards yield but not the __getitem__ call. For expensive datasets (e.g. those that load from disk), this means every worker reads every batch but discards most of them. The index filtering should happen before __getitem__.

keras/src/trainers/data_adapters/torch_data_loader_adapter.py
```diff
@@ -13,12 +13,65 @@ class TorchDataLoaderAdapter(DataAdapter):
...
+                        rank = process_id
+                    else:
+                        num_replicas = num_model_replicas
+                        processes_per_replica = (
+                            num_process // num_model_replicas
+                        )
+                        rank = process_id // processes_per_replica
+
+            if num_replicas is not None and rank is not None:
+                sampler = torch.utils.data.distributed.DistributedSampler(
```
amitsrivastava78: The original dataloader's shuffle setting is ignored — it's always replaced with shuffle=True. Should respect dataloader.dataset's shuffle intent or at least pass it as a parameter.

amitsrivastava78:
Test coverage is only 12% which is very less, add test cases for loading and restoration of checkpoint.

keras/src/backend/torch/core.py
```diff
@@ -102,19 +104,78 @@ def to_torch_dtype(dtype):
...
+        self._initialize_layout()
+
+        if self._layout is not None:
+            from keras.src.backend.torch.distribution_lib import (
+                distribute_tensor,
+            )
+
+            if isinstance(value, torch.nn.Parameter):
+                value = value.data
+            value = convert_to_tensor(value, dtype=self._dtype)
```
amitsrivastava78, line 127: This runs on every convert_to_tensor call in ModelParallel mode, meaning all intermediate activations (not just weights) get wrapped as replicated DTensors. This is very different from JAX's approach 
DTensor is designed so that sharding only needs to be declared on the leaf tensors (weights). The sharding of intermediate tensors propagates automatically through ops via DTensor's dispatch mechanism:

Weight is a DTensor with Shard(1) (e.g., column-sharded)
Input is a plain local torch.Tensor
output = input @ weight → DTensor dispatch handles this automatically, producing an output DTensor with the correct placement based on the op's sharding rule
This is exactly how JAX/XLA handles it too — you annotate the arrays with sharding, not every op output.

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,254 @@
...
+    return new_strategy
+
+
+def _register_unbind_strategy():
+    """Registers sharding propagation for `unbind` (used by MHA heads)."""
+    global _UNBIND_REGISTERED
+    if _UNBIND_REGISTERED:
+        return
+
+    from torch.distributed.tensor._op_schema import RuntimeSchemaInfo
```
amitsrivastava78: _register_unbind_strategy uses private PyTorch internals
PyTorch ≥ 2.4 handles unbind on DTensors natively, so skip registration entirely on modern versions

keras/src/trainers/data_adapters/py_dataset_adapter.py
```diff
@@ -203,14 +203,27 @@ def __init__(
...
         self.class_weight = class_weight
         self.enqueuer = None
         self.shuffle = shuffle
         self._output_signature = None
         self._within_epoch = False
 
+        dist = distribution or distribution_lib.distribution()
+        self._num_processes = 1
+        self._process_id = 0
+        if dist is not None and getattr(dist, "auto_shard_dataset", True):
```
amitsrivastava78: Default True means any distribution object that doesn't have auto_shard_dataset will trigger sharding. Should be False

keras/src/backend/torch/trainer.py
```diff
@@ -38,19 +57,70 @@ def _should_torch_compile(self):
 
         return self.jit_compile
 
+    def _setup_ddp(self):
```
amitsrivastava78, line 46: These methods call self._setup_ddp() on every batch. While the hasattr guard prevents re-wrapping, it's still an isinstance + attribute lookup per step. Move this entirely to make_train/test/predict_function.

keras/src/backend/torch/trainer.py
```diff
@@ -38,19 +57,70 @@ def _should_torch_compile(self):
...
+        from keras.src.distribution import distribution_lib as dist_lib
+
+        dist = dist_lib.distribution()
+        if dist is not None and isinstance(dist, dist_lib.DataParallel):
+            if not hasattr(self, "_ddp_model"):
+                ddp_model = torch.nn.parallel.DistributedDataParallel(
+                    _KerasModuleWrapper(self),
```
amitsrivastava78: _KerasModuleWrapper registers non-trainable weights as register_buffer, but DDP only syncs gradients of parameters. BatchNorm's running mean/var buffers need explicit all_reduce after each forward pass. Without this, running stats will diverge across ranks

keras/src/trainers/data_adapters/py_dataset_adapter.py
```diff
@@ -251,7 +266,9 @@ def _standardize_batch(self, batch):
         return batch
 
     def _infinite_generator(self):
-        for i in itertools.count():
+        for i in itertools.count(
+            start=self._process_id, step=self._num_processes
+        ):
             yield self._standardize_batch(self.py_dataset[i])
 
     def _finite_generator(self):
```
amitsrivastava78, line 257: If [len(indices)] is not divisible by num_processes, some ranks will have fewer batches. Any all_reduce in _sync_metrics will deadlock waiting for ranks that have already exhausted their data.

keras/src/backend/torch/trainer.py
```diff
@@ -17,12 +20,28 @@
...
+        return self._keras_model(*args, **kwargs)
+
+
 class TorchTrainer(base_trainer.Trainer):
     def __init__(self):
         super().__init__()
         self.train_function = None
         self.test_function = None
         self.predict_function = None
+        self._in_ddp_context = False
```
amitsrivastava78, line 29: Once DataParallel scope is entered and _setup_ddp() runs, _in_ddp_context = True permanently even if the distribution scope is exited. The trainer will keep calling self._ddp_model(x) even in a non-distributed context.

keras/src/trainers/data_adapters/array_data_adapter_test.py
```diff
@@ -300,3 +303,79 @@ def test_errors(self):
             array_data_adapter.ArrayDataAdapter(
                 x=x, y=nested_y, class_weight=cw
             )
+
+    def test_torch_data_loader_nested_input(self):
+        # This tests ArrayDataset.__len__ with nested input
```
amitsrivastava78: change is fragile
Accessing .array on the result of tree.flatten assumes the leaf is an internal array wrapper with a .array attribute. This will fail for plain numpy arrays or torch tensors.

keras/src/backend/torch/core.py
```diff
@@ -6,6 +6,7 @@
 import ml_dtypes
 import numpy as np
 import torch
+from torch.distributed.tensor import DTensor
```
amitsrivastava78, line 9: def _get_dtensor():
    from torch.distributed.tensor import DTensor
    return DTensor

DTensor unconditionally imported at module level in core.py — import fragility, do like above

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+def list_devices(device_type=None):
+    """Returns Keras device strings representing global indices."""
+    device_type = device_type or "gpu"
+    count = get_device_count(device_type)
+    return [f"{device_type.lower()}:{i}" for i in range(count)]
+
+
+def initialize(job_addresses=None, num_processes=None, process_id=None):
```
amitsrivastava78, line 84: Calling init_process_group twice raises RuntimeError. Earlier commits had if not torch.distributed.is_initialized():.
Fix:
def initialize(...):
    if not torch.distributed.is_initialized():
        local_rank = int(os.environ.get("LOCAL_RANK", 0))
        ...

keras/src/backend/torch/trainer.py
```diff
@@ -38,18 +58,137 @@ def _should_torch_compile(self):
 
         return self.jit_compile
 
-    def train_step(self, data):
+    def _setup_ddp(self):
```
amitsrivastava78, line 46: _setup_ddp has no cleanup path — stale DDP model

amitsrivastava78: When you exit a DataParallel scope, _in_ddp_context stays True and all subsequent calls route through the stale _ddp_model.
Fix: Add the cleanup branch (was present in an earlier commit):
def _setup_ddp(self):
    dist = dist_lib.distribution()
    if dist is not None and isinstance(dist, dist_lib.DataParallel):
        if not hasattr(self, "_ddp_model"):
            self._ddp_model = ...
            self._in_ddp_context = True
    else:
        if hasattr(self, "_ddp_model"):
            del self._ddp_model
        self._in_ddp_context = False

keras/src/trainers/data_adapters/array_data_adapter.py
```diff
@@ -289,7 +291,49 @@ def __iter__(self):
...
+            num_replicas = None
+            rank = None
+            if isinstance(dist, dist_lib.DataParallel):
+                num_replicas = torch.distributed.get_world_size()
+                rank = torch.distributed.get_rank()
+            elif isinstance(dist, dist_lib.ModelParallel):
+                mesh_batch_dim_index = dist.device_mesh.axis_names.index(
+                    dist.batch_dim_name
+                )
+                num_model_replicas = dist.device_mesh.shape[
```
amitsrivastava78: The ModelParallel rank computation:This identical block appears in array_data_adapter.py, torch_data_loader_adapter.py, and trainer.py.
Fix: Extract into distribution_lib.py:
def get_num_replicas_and_rank(dist):
    ...

keras/src/backend/torch/core.py
```diff
@@ -190,6 +251,9 @@ def __eq__(self, other):
 
 
 def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
```
amitsrivastava78: These lazy imports inside a hot path (called for every tensor) add overhead. Move to module-level or at least guard with if dist is not None.

keras/src/backend/torch/trainer.py
```diff
@@ -38,18 +58,137 @@ def _should_torch_compile(self):
...
 
-        # Compute predictions
+        if dist is not None and isinstance(dist, dist_lib.DataParallel):
+            return self._ddp_model(x, training=training)
         if self._call_has_training_arg:
-            y_pred = self(x, training=True)
+            return self(x, training=training)
+        return self(x)
+
+    def _sync_ddp_buffers(self, dist):
```
amitsrivastava78, line 149: _sync_ddp_buffers always uses SUM + divide regardless of buffer semantics
This averages all non-trainable buffers. For BatchNormalization, num_batches_tracked should be summed (not averaged), while running_mean/running_var should be averaged. Applying the same reduction to all buffers is incorrect.

Fix: Inspect the buffer name or use variable.aggregation metadata if available; at minimum skip num_batches_tracked-style integer counters.

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,222 @@
...
+    if layout is None:
+        return tensor
+
+    from keras.src.distribution import distribution_lib as dist_lib
+
+    dist = dist_lib.distribution()
+    if not isinstance(dist, dist_lib.ModelParallel):
+        return tensor
+
+    _setup_dtensor_ops()
```
amitsrivastava78: _setup_dtensor_ops() replaces three global symbols — DTensor.unbind, torch.unbind, and torch.nn.functional.dropout — for the entire process lifetime. This means:

Any other PyTorch code in the same process (user code, other libraries) silently gets the patched versions, which may cause subtle correctness bugs.
Both except: blocks swallow all exceptions including KeyboardInterrupt, SystemExit, and legitimate PyTorch errors — making debugging very difficult.
The patched unbind falls back to tensor.to_local().unbind(dim) on error, losing the distributed placement information and potentially silently producing wrong results on multi-device setups.
The correct approach is to register a sharding propagation strategy via PyTorch's official register_op_strategy API instead of patching.
Fix
Replace _setup_dtensor_ops() and all call sites with _register_unbind_strategy():

# ── REMOVE the old _setup_dtensor_ops entirely ──────────────────────────────

# ── ADD these instead ────────────────────────────────────────────────────────

_UNBIND_REGISTERED = False


def _unbind_op_strategy(op_schema):
    from torch.distributed.tensor._dtensor_spec import DTensorSpec
    from torch.distributed.tensor._op_schema import OpSpec, OpStrategy

    input_strategy = op_schema.args_schema[0]
    dim = op_schema.args_schema[1] if len(op_schema.args_schema) > 1 else 0
    dim = dim if dim >= 0 else dim + input_strategy.ndim

    mesh = input_strategy.mesh
    new_strategy = OpStrategy([])

    for arg_strategy in input_strategy.strategies:
        arg_spec = arg_strategy.output_spec
        is_sharded_on_dim = any(
            isinstance(p, Shard) and p.dim == dim
            for p in arg_spec.placements
        )
        if is_sharded_on_dim:
            rep_placements = tuple(Replicate() for _ in arg_spec.placements)
            rep_spec = DTensorSpec(
                mesh=mesh,
                placements=rep_placements,
                tensor_meta=arg_spec.tensor_meta,
            )
            out_spec = DTensorSpec(mesh=mesh, placements=rep_placements)
            new_strategy.strategies.append(
                OpSpec(
                    output_specs=(out_spec,) * input_strategy.shape[dim],
                    input_specs=(rep_spec,),
                )
            )
        else:
            out_placements = [
                Shard(p.dim - 1) if isinstance(p, Shard) and p.dim > dim else p
                for p in arg_spec.placements
            ]
            out_spec = DTensorSpec(
                mesh=mesh, placements=tuple(out_placements)
            )
            new_strategy.strategies.append(
                OpSpec(
                    output_specs=(out_spec,) * input_strategy.shape[dim],
                    input_specs=(arg_spec,),
                )
            )
    return new_strategy


def _register_unbind_strategy():
    """Register sharding propagation for `unbind` (used by MHA heads).
    No-ops if already registered or if the PyTorch internal API is unavailable.
    """
    global _UNBIND_REGISTERED
    if _UNBIND_REGISTERED:
        return
    try:
        from torch.distributed.tensor._op_schema import RuntimeSchemaInfo
        from torch.distributed.tensor._ops import register_op_strategy

        register_op_strategy(
            torch.ops.aten.unbind.int, schema_info=RuntimeSchemaInfo(1)
        )(_unbind_op_strategy)
        _UNBIND_REGISTERED = True
    except (ImportError, AttributeError):
        # PyTorch version does not expose these internal APIs yet;
        # unbind on sharded DTensors will fall back to PyTorch's default.
        pass

Then update the two call sites in distribute_tensor and distribute_data_input:
# Before:
_setup_dtensor_ops()

# After:
_register_unbind_strategy()

keras/src/trainers/data_adapters/py_dataset_adapter.py
```diff
@@ -255,13 +270,19 @@ def _infinite_generator(self):
...
+        num_batches_per_rank = (
+            num_batches + self._num_processes - 1
+        ) // self._num_processes
+        for i in range(num_batches_per_rank):
+            idx = i * self._num_processes + self._process_id
+            if idx >= num_batches:
+                idx = num_batches - 1
+            yield self._standardize_batch(self.py_dataset[indices[idx]])
 
     def _infinite_enqueuer_generator(self):
```
amitsrivastava78, line 266: _finite_generator correctly shards across ranks, but _infinite_generator ignores self._process_id entirely every rank processes every batch
Fix
def _infinite_generator(self):
    for i in itertools.count(
        start=self._process_id, step=self._num_processes
    ):
        yield self._standardize_batch(self.py_dataset[i])

keras/src/trainers/data_adapters/data_adapter_utils.py
```diff
@@ -355,6 +355,33 @@ def is_torch_tensor(value):
     return False
 
 
+def _add_distributed_sampler(dataloader, num_replicas, rank):
+    import torch
+
+    sampler = torch.utils.data.distributed.DistributedSampler(
```
amitsrivastava78: sampler = torch.utils.data.distributed.DistributedSampler(
    dataloader.dataset,   # ← crashes with TypeError if IterableDataset (no __len__)

Fix: add if isinstance(dataloader.dataset, torch.utils.data.IterableDataset): return dataloader

keras/src/distribution/distribution_lib.py
```diff
@@ -775,6 +775,42 @@ def distribute_dataset(self, dataset):
...
+    rank = None
+    if isinstance(dist, DataParallel):
+        num_replicas = distribution_lib.num_processes()
+        rank = distribution_lib.process_id()
+    elif isinstance(dist, ModelParallel):
+        mesh_batch_dim_index = dist.device_mesh.axis_names.index(
+            dist.batch_dim_name
+        )
+        num_model_replicas = dist.device_mesh.shape[mesh_batch_dim_index]
+        if num_model_replicas > 0:
```
amitsrivastava78: if num_model_replicas > 0:   # ← should be > 1; single-replica mesh triggers distributed path
Fix: change to if num_model_replicas > 1:.

keras/src/trainers/data_adapters/__init__.py
```diff
@@ -85,15 +89,19 @@ def get_data_adapter(
...
-            )
-        return PyDatasetAdapter(x, class_weight=class_weight, shuffle=shuffle)
-
+        return PyDatasetAdapter(
+            x,
+            class_weight=class_weight,
+            shuffle=shuffle,
+            distribution=distribution,
+        )
+        # TODO: should we warn or not?
```
amitsrivastava78: Commented-out dead code left in. Either restore the warning or delete it.

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,268 @@
+import os
+
+import torch
+from torch.distributed.device_mesh import init_device_mesh
```
amitsrivastava78: from torch.distributed.device_mesh import init_device_mesh  # ← module-level, fails on PyTorch < 2.0
Minor robustness issue — should be lazily imported inside _to_backend_mesh() where it's actually used (same pattern used elsewhere in the file for DTensor, Replicate, Shard).

keras/src/backend/torch/core.py
```diff
@@ -103,19 +111,81 @@ def to_torch_dtype(dtype):
...
+    def __init__(self, *args, layout=None, **kwargs):
+        self._layout = layout
+        super().__init__(*args, **kwargs)
+
+    def _initialize_layout(self):
+        distribution = global_state.get_global_attribute("distribution")
+        if self._layout is None and distribution is not None:
+            tensor_layout = distribution.get_variable_layout(self)
+
+            from keras.src.distribution import TensorLayout
```
amitsrivastava78, line 113: from keras.src.distribution import TensorLayout           # ← unconditional import on every
from keras.src.distribution import distribution_lib as dist_lib   # symbolic build call
 Inside convert_keras_tensor_to_torch, these run even when there's no active distribution. The same imports in convert_to_tensor were correctly guarded by if dist is not None:. Add dist = dist_lib.distribution() check first.

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,268 @@
...
+
+
+def _get_dtensor():
```
amitsrivastava78: Identical function copy-pasted. core.py already imports distribution_lib — it could just call torch_dist_lib._get_dtensor() instead of redefining it.

keras/src/backend/torch/trainer.py
```diff
@@ -38,18 +59,148 @@ def _should_torch_compile(self):
...
+                    "max": torch.distributed.ReduceOp.MAX,
+                    "min": torch.distributed.ReduceOp.MIN,
+                }
+                for name, buf in module.named_buffers():
+                    if buf is None:
+                        continue
+                    if not buf.requires_grad:
+                        if torch.distributed.is_initialized():
+                            aggregation = module._buffer_aggregations.get(name)
+                            if aggregation in (None, "none"):
```
amitsrivastava78, line 167: When aggregation is None, it falls through to "mean" and applies all_reduce + div_. For BatchNormalization.num_batches_tracked (an integer counter), this is wrong — SUM+divide corrupts the value. num_batches_tracked should use MAX or simply be broadcast from rank 0.

Fix
if aggregation in (None, "none"):
    # Unknown aggregation — broadcast from rank 0 (safe default).
    torch.distributed.broadcast(buf, src=0)
    continue
if aggregation not in aggregation_to_op:
    continue

keras/src/trainers/data_adapters/__init__.py
```diff
@@ -147,7 +146,6 @@ def get_data_adapter(
                 stacklevel=2,
             )
         return GrainDatasetAdapter(x)
-
     elif isinstance(x, types.GeneratorType):
```
amitsrivastava78, line 151: Every other adapter (tf.data.Dataset, torch DataLoader, grain, generator) warns when shuffle=True is passed but will be silently ignored. The PyDataset branch is the only one that drops this warning. For infinite PyDatasets (num_batches is None), shuffle=True is a no-op and the user gets no feedback. Please restore
Fix
elif isinstance(x, py_dataset_adapter.PyDataset):
    if y is not None:
        raise_unsupported_arg("y", "the targets", "PyDataset")
    if sample_weight is not None:
        raise_unsupported_arg(
            "sample_weights", "the sample weights", "PyDataset"
        )
    if getattr(x, "num_batches", None) is None and shuffle:
        warnings.warn(
            "`shuffle=True` was passed, but will be ignored since the "
            "data `x` was provided as an infinite PyDataset. The "
            "PyDataset is expected to already be shuffled.",
            stacklevel=2,
        )
    return PyDatasetAdapter(
        x,
        class_weight=class_weight,
        shuffle=shuffle,
        distribution=distribution,
    )

keras/src/backend/torch/numpy.py
```diff
@@ -328,8 +328,8 @@ def arange(start, stop=None, step=None, dtype=None):
         start, stop = 0, start
     if step is None:
         step = 1
-    return torch.arange(
-        start, stop, step=step, dtype=dtype, device=get_device()
+    return convert_to_tensor(
+        torch.arange(start, stop, step=step, dtype=dtype, device=get_device())
```
hertschuh, line 338: Why this change? This shouldn't make any difference since `torch.arange` returns a tensor.

keras/src/backend/torch/core.py
```diff
@@ -103,19 +104,81 @@ def to_torch_dtype(dtype):
 
 
 class Variable(KerasVariable):
+    def __init__(self, *args, layout=None, **kwargs):
+        self._layout = layout
+        super().__init__(*args, **kwargs)
```
hertschuh: This was needed when you first created this PR, but it is no longer needed (as of https://github.com/keras-team/keras/pull/22680/changes#diff-ef65411710c16fa09f65f31ca1befe257e41339f5aa5bfd77ead35b4df4d364d )

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,268 @@
+import os
+
+import torch
+
+
+def _get_dtensor():
```
hertschuh: Why do you need this? Why not import `torch.distributed.tensor.DTensor` where you need it?

keras/src/backend/torch/core.py
```diff
@@ -103,19 +104,81 @@ def to_torch_dtype(dtype):
...
 
     def _direct_assign(self, value):
-        with torch.no_grad():
-            self.value.copy_(value)
+        value = convert_to_tensor(value, dtype=self._dtype).detach()
+        if self._layout is not None:
+            value = torch_dist_lib.distribute_tensor(value, self._layout)
+            self._value = torch.nn.Parameter(
+                value, requires_grad=self.trainable
+            )
```
hertschuh: Why not `self._value.copy_` here?

I'm concerned can creating a new `Parameter` will have unintended side effects, like having many copies of the same variables. Also, I doubt this will be compiled correctly.

I think this is the reason you needed the wrapper to register parameters, it's because by doing this, you're losing the already registered parameters and creating new unregistered ones.

keras/src/backend/torch/core.py
```diff
@@ -206,47 +269,75 @@ def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
...
+                dtype = dtype or x.dtype
+            if dtype is None:
+                dtype = result_type(
+                    *[
+                        getattr(item, "dtype", type(item))
+                        for item in tree.flatten(x)
+                    ]
+                )
+            dtype = to_torch_dtype(dtype)
+            x = torch.as_tensor(x, dtype=dtype, device=get_device())
```
hertschuh: There's a PR up that I think addressed some of these issues: https://github.com/keras-team/keras/pull/22821

keras/src/backend/torch/core.py
```diff
@@ -206,47 +269,75 @@ def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
...
+        if DTensor is not None and isinstance(x, DTensor):
+            return x
+
+        is_parameter = isinstance(x, torch.nn.Parameter)
+        is_leaf_tensor = is_tensor(x) and getattr(x, "is_leaf", False)
+        if not (is_parameter or is_leaf_tensor):
+            return x
+
+        layout = TensorLayout([None] * x.ndim, dist.device_mesh)
+        x = torch_dist_lib.distribute_tensor(x, layout)
```
hertschuh: I don't think `convert_to_tensor` should blindly distribute like that. The JAX implementation doesn't. And if we need to add a `layout` argument to `convert_to_tensor`, that's one thing. But here it's assuming replication.

Why did you add this? Was it solving some issue?

keras/src/backend/torch/core.py
```diff
@@ -310,11 +401,21 @@ def convert_keras_tensor_to_torch(x, fill_value=None):
...
+
+            dist = global_state.get_global_attribute("distribution")
+            if dist is not None:
+                from keras.src.distribution import TensorLayout
+                from keras.src.distribution import distribution_lib as dist_lib
+
+                if isinstance(dist, dist_lib.ModelParallel):
+                    layout = TensorLayout([None] * len(shape), dist.device_mesh)
+                    t = torch_dist_lib.distribute_tensor(t, layout)
+            return t
```
hertschuh: Is this needed? The purpose of this is purely to determine the output shape and dtype, we shouldn't care about distribution.

Also, you're making assumptions about which layout to apply.

keras/src/backend/torch/trainer.py
```diff
@@ -17,12 +19,31 @@
...
+        object.__setattr__(self, "_keras_model", keras_model)
+        self._buffer_aggregations = {}
+        for i, v in enumerate(keras_model.trainable_weights):
+            self.register_parameter(f"p{i}", v.value)
+        for i, v in enumerate(keras_model.non_trainable_weights):
+            self.register_buffer(f"b{i}", v.value)
+            self._buffer_aggregations[f"b{i}"] = v.aggregation
+
+    def forward(self, *args, **kwargs):
+        return self._keras_model(*args, **kwargs)
```
hertschuh: This shouldn't be needed, it's covered by:
- https://github.com/keras-team/keras/blob/master/keras/src/backend/torch/layer.py#L21-L26
- https://github.com/keras-team/keras/blob/master/keras/src/backend/torch/layer.py#L60

If there's an issue with this system, we should fix it instead.

keras/src/backend/torch/trainer.py
```diff
@@ -38,18 +59,151 @@ def _should_torch_compile(self):
...
+            self._in_ddp_context = True
+        else:
+            if hasattr(self, "_ddp_model"):
+                delattr(self, "_ddp_model")
+            self._in_ddp_context = False
+
+    def _get_distribution(self):
+        from keras.src.distribution import distribution_lib as dist_lib
+
+        return dist_lib, dist_lib.distribution()
```
hertschuh: This is a weird API, just import at the top and use `dist_lib.distribution()` directly:

```python
```

keras/src/backend/torch/trainer.py
```diff
@@ -38,18 +59,151 @@ def _should_torch_compile(self):
...
+            return t
+
+        return tree.map_structure(_distribute_if_tensor, data)
+
+    def _apply_distribution(self, dist, x, y=None):
+        if dist is not None:
+            x = self._distribute_inputs(dist, x)
+            if y is not None:
+                y = self._distribute_inputs(dist, y)
+        return x, y
```
hertschuh: Follow the same pattern as JAX, which is simpler:

- call it `_distribute_data`
- distribute before calling `unpack_x_y_sample_weight` so that you don't need special handling of x, y and weight, just one `tree.map_structure`

keras/src/backend/torch/trainer.py
```diff
@@ -171,6 +371,27 @@ def one_step_on_data(data):
...
+
+        if dist is not None:
+            if data_batch is not None:
+                data_batch = self._distribute_inputs(
+                    dist, data_batch, replicate=True
+                )
+            super()._symbolic_build(data_batch=data_batch)
+            return
+
+        super()._symbolic_build(data_batch=data_batch, iterator=iterator)
```
hertschuh: Why did you need this? `_symbolic_build` is about determining the shape of variables, but the distribution of variables doesn't depend on the distribution of the data.

keras/src/distribution/distribution_lib.py
```diff
@@ -775,6 +775,42 @@ def distribute_dataset(self, dataset):
...
+            num_process = distribution_lib.num_processes()
+            process_id = distribution_lib.process_id()
+            if num_model_replicas >= num_process:
+                num_replicas = num_process
+                rank = process_id
+            else:
+                num_replicas = num_model_replicas
+                processes_per_replica = num_process // num_model_replicas
+                rank = process_id // processes_per_replica
+    return num_replicas, rank
```
hertschuh: These should be properties of the `Distribution` object.
- add `num_processes` to `Distribution` and change the `__init__` in the super class and the subclasses to populate it correctly
- add `num_model_replicas` property
- add another property for what you call `rank`. But let's not call it rank because it's super confusing. Right now in the `ModelParallel` code, it's called `data_shard_id`, which I think is a lot more descriptive

Now... based on my other comment about adding a `distribute_torch_dataloader` method in the Distribution classes, you may not need this at all.

keras/src/trainers/data_adapters/py_dataset_adapter.py
```diff
@@ -203,14 +203,27 @@ def __init__(
...
         self.shuffle = shuffle
         self._output_signature = None
         self._within_epoch = False
 
+        dist = distribution or distribution_lib.distribution()
+        self._num_processes = 1
+        self._process_id = 0
+        if dist is not None and getattr(dist, "auto_shard_dataset", False):
+            self._num_processes = backend_distribution_lib.num_processes()
+            self._process_id = backend_distribution_lib.process_id()
```
hertschuh: So this makes sense with Torch because there is one process per device, but it will do the wrong thing for other backends because it will skip one in N batches. So this needs to be conditional.

Now, I'm trying to think of a good way to do this.
- For the better or the worse, everything related to the distribution of `tf.data.Dataset` is handled in the `Distribution` subclasses in `distribute_dataset`
- Your approach is to sprinkle the code changes in the individual `DatasetAdapter` classes

I'd be tempted to try to follow the current pattern and see if we can make it work, i.e.:
- rename `distribute_dataset` to `distribute_tf_dataset`
- add a `distribute_torch_dataloader` in all the distribution classes, I'm hoping this way the code can be shared between multiple `DatasetAdapter`

keras/src/trainers/data_adapters/data_adapter_utils.py
```diff
@@ -355,6 +355,36 @@ def is_torch_tensor(value):
     return False
 
 
+def _add_distributed_sampler(dataloader, num_replicas, rank):
+    import torch
+
+    if isinstance(dataloader.dataset, torch.utils.data.IterableDataset):
+        return dataloader
```
hertschuh: But that's going to not sample in this case, right? That would be incorrect.

keras/src/trainers/data_adapters/data_adapter_utils.py
```diff
@@ -355,6 +355,36 @@ def is_torch_tensor(value):
     return False
 
 
+def _add_distributed_sampler(dataloader, num_replicas, rank):
+    import torch
```
hertschuh: The name of this util should make it clear it's for torch dataloaders only.

hertschuh (changes requested):
First pass.

The main design question is how to handle datasets, see my comment below.

keras/src/backend/torch/core.py
buildwithsuhana: In JAX, the jax.Array type is unified, and the JIT compiler handles implicit sharding promotion during operation alignment. However, in PyTorch, DTensor and standard torch.Tensor are distinct types that cannot be mixed in the same operation without triggering a RuntimeError.
When ModelParallel is active, model weights are promoted to DTensors. Any leaf tensor (like constants from numpy or literals) that interacts with these weights must also be a DTensor on the same DeviceMesh. We assume replication because it is safe, semantics-preserving default for global constants.

keras/src/backend/torch/numpy.py
buildwithsuhana: While torch.arange returns a tensor, it is a local one. Wrapping it in convert_to_tensor ensures it is automatically promoted to a DTensor when a ModelParallel strategy is active. This is required to avoid RuntimeError: Mixing DTensor and non-DTensor when these functional constants interact with sharded model weights.

hertschuh: But then, does it mean we have to do this for all the ops?

And what layout does it apply?

keras/src/backend/torch/core.py
hertschuh: The PR just got submitted, you'll need to rebase

keras/src/backend/torch/core.py
hertschuh: Oh thinking about this some more, can we use the `meta` device for this?

keras/src/backend/torch/core.py
```diff
@@ -103,19 +104,76 @@ def to_torch_dtype(dtype):
...
+                    ),
+                )
 
     def _direct_assign(self, value):
+        value = convert_to_tensor(value, dtype=self._dtype).detach()
+        if self._layout is not None:
+            value = torch_dist_lib.distribute_tensor(value, self._layout)
+        else:
+            from torch.distributed.tensor import DTensor
+
```
hertschuh: Why not import at the top?

keras/src/backend/torch/core.py
```diff
@@ -191,6 +252,8 @@ def __eq__(self, other):
 
 
 def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
+    if x is None:
+        return None
```
hertschuh: I think None should fail, we should not let it go through

keras/src/backend/torch/core.py
```diff
@@ -206,47 +269,78 @@ def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
...
+        layout = TensorLayout([None] * x.ndim, dist.device_mesh)
+        x = torch_dist_lib.distribute_tensor(x, layout)
+    return x
 
 
 def convert_to_numpy(x):
     def transform(x):
         if is_tensor(x):
+            from torch.distributed.tensor import DTensor
+
```
hertschuh: Move import to the top

keras/src/trainers/data_adapters/data_adapter_utils.py
```diff
@@ -355,6 +355,85 @@ def is_torch_tensor(value):
     return False
 
 
+def _add_torch_distributed_sampler(dataloader, num_replicas, rank):
+    import itertools
+
```
hertschuh: Move this import to the top

keras/src/backend/torch/core.py
buildwithsuhana: Done

keras/src/backend/torch/core.py
buildwithsuhana: yes, done

keras/src/backend/torch/numpy.py
buildwithsuhana: I've been testing model parallel with the OPT and GPT-2 models using ModelParallel, and in those specific cases, the arange change was the only one required. It was needed because the positional indices generated during the forward pass were hitting sharded embedding weights; without the wrap, PyTorch catches the mismatch between the local index tensor and the distributed weights.

Regarding the layout, convert_to_tensor applies a replicated layout by default. This makes the local tensor a DTensor that's consistent across all devices, so it can safely interact with sharded parameters without needing any actual network communication.

While arange was the immediate blocker for those models, we can hit the same issue whenever a model generates functional constants on the fly that interact with sharded activations or weights. So, should we extend this fix to other ops as well which are used on the fly to create some constants in some layers?

keras/src/trainers/data_adapters/py_dataset_adapter.py
```diff
@@ -251,17 +263,25 @@ def _standardize_batch(self, batch):
...
 
-        for i in indices:
-            yield self._standardize_batch(self.py_dataset[i])
+        num_batches_per_rank = (
+            num_batches + self._num_processes - 1
+        ) // self._num_processes
+        for i in range(num_batches_per_rank):
+            idx = i * self._num_processes + self._process_id
+            if idx >= num_batches:
+                idx = num_batches - 1
```
amitsrivastava78: idx = num_batches - 1  # silently repeats last batch
When indices don't divide evenly across ranks, this duplicates the last batch on some ranks instead of skipping. The correct options are: drop the remainder (drop_last=True semantics), or let the caller handle uneven counts. Silently feeding duplicate data corrupts training.

keras/src/distribution/distribution_lib_test.py
```diff
@@ -355,7 +359,7 @@ def test_distribute_dataset(self):
         distribution = distribution_lib.ModelParallel(
             layout_map=layout_map, batch_dim_name="data"
         )
-        distributed_dataset = distribution.distribute_dataset(dataset)
+        distributed_dataset = distribution.distribute_tf_dataset(dataset)
```
amitsrivastava78: renames the public method distribute_dataset() to distribute_tf_dataset() in the Distribution base class and both DataParallel / ModelParallel subclasses.Any user code written against the existing Keras API breaks:
# User code written today — works on current Keras, silently breaks after this PR merges
distributed_ds = distribution.distribute_dataset(my_tf_dataset)

keras/src/backend/torch/numpy.py
amitsrivastava78, line 338: The promotion logic should live at the dispatch layer, not per-op. The cleanest place is inside convert_to_tensor itself, but only for the "op output" case, not for user data.

keras/src/backend/torch/trainer.py
```diff
@@ -102,14 +242,52 @@ def test_step(self, data):
...
+            x,
+            _,
+            _,
+        ) = self._unpack_and_distribute_data(data)
+
+        y_pred = self._forward(dist, x, training=False)
+        self._sync_ddp_buffers(dist)
         return y_pred
 
+    def _sync_metrics(self):
```
amitsrivastava78, line 289: For metrics: mock torch.distributed.all_reduce and assert it's called with the right op and value. For gradient accumulation: set gradient_accumulation_steps=2 and verify no_sync is entered on step 1 and not step 2. For checkpointing: either use torch.multiprocessing.spawn with world_size=2, or mock distributed state to simulate rank 0 vs rank 1 save/load.

keras/src/backend/torch/trainer.py
```diff
@@ -38,18 +43,148 @@ def _should_torch_compile(self):
 
         return self.jit_compile
 
-    def train_step(self, data):
+    def _setup_ddp(self):
```
amitsrivastava78, line 46: _setup_ddp wraps self into DDP without first ensuring model parameters are on the target CUDA device. DDP's constructor requires all parameters to already be on the same device — if the user builds the model on CPU (the default) and runs on a GPU machine, this raises a confusing ValueError at model.fit() time with no indication of what went wrong. Either move the model to the correct device before wrapping, or raise a clear error/warning if parameters are on CPU when CUDA is available.
suggested fix 
def _setup_ddp(self):
    dist = dist_lib.distribution()
    if dist is not self._current_dist:
        self.train_function = None
        self.test_function = None
        self.predict_function = None
        self._current_dist = dist

    if dist is not None and isinstance(dist, dist_lib.DataParallel):
        if not hasattr(self, "_ddp_model"):
            if torch.cuda.is_available():
                device = torch.device(f"cuda:{torch.cuda.current_device()}")
                # Move model to the target device before DDP wrapping.
                # DDP requires all parameters to be on the same device.
                self.to(device)
                device_ids = [torch.cuda.current_device()]
            else:
                device_ids = None
            ddp_model = torch.nn.parallel.DistributedDataParallel(
                self,
                device_ids=device_ids,
            )
            object.__setattr__(self, "_ddp_model", ddp_model)
        self._in_ddp_context = True
    else:
        if hasattr(self, "_ddp_model"):
            object.__delattr__(self, "_ddp_model")
        self._in_ddp_context = False

keras/src/backend/torch/core.py
```diff
@@ -103,19 +105,73 @@ def to_torch_dtype(dtype):
...
-            self._value = value
-        else:
+        self._shape = self._validate_shape(value.shape)
+        self._initialize_layout()
+
+        if self._layout is not None:
+            from keras.src.backend.torch.distribution_lib import (
+                distribute_tensor,
+            )
+
```
hertschuh: Use the import you already have: `torch_dist_lib.distribute_tensor`

keras/src/backend/torch/core.py
```diff
@@ -103,19 +105,73 @@ def to_torch_dtype(dtype):
...
+            if isinstance(value, torch.nn.Parameter):
+                value = value.data
+            value = convert_to_tensor(value, dtype=self._dtype)
+            value = distribute_tensor(value, self._layout)
```
hertschuh: This is were it gets weird because it will get distributed twice:
- once by convert_to_tensor
- once by distribute_tensor

keras/src/backend/torch/core.py
```diff
@@ -103,19 +105,73 @@ def to_torch_dtype(dtype):
...
+                    param_value,
+                    requires_grad=self.trainable
+                    and (
+                        torch.is_floating_point(param_value)
+                        or torch.is_complex(param_value)
+                    ),
+                )
 
     def _direct_assign(self, value):
+        value = convert_to_tensor(value, dtype=self._dtype).detach()
```
hertschuh: Why do you need `detach()` here?

keras/src/backend/torch/core.py
```diff
@@ -313,10 +396,14 @@ def convert_keras_tensor_to_torch(x, fill_value=None):
                 for i, e in enumerate(shape):
                     if e is None:
                         shape[i] = fill_value
+            device = get_device()
+            if global_state.get_global_attribute("distribution") is not None:
+                device = "meta"
```
hertschuh: I think it should work if you *always* do "meta". Can you try?

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,264 @@
...
+import torch
+
+
+def get_device_count(device_type=None):
+    """Returns total device count across all hosts."""
+    if torch.distributed.is_initialized():
+        return torch.distributed.get_world_size()
+    if "WORLD_SIZE" in os.environ:
+        return int(os.environ["WORLD_SIZE"])
+    return torch.cuda.device_count() or 1
```
hertschuh: This is not taking into account the `device_type`. If should handle "gpu" vs. "cpu". And there should be an `else` that returns 0 to handle "tpu".

Also, I'm not sure what should be done when these 3 values disagree.

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+def list_devices(device_type=None):
+    """Returns Keras device strings representing global indices."""
+    device_type = device_type or "gpu"
+    count = get_device_count(device_type)
+    return [f"{device_type.lower()}:{i}" for i in range(count)]
```
hertschuh: This should handle `device_type` correctly.
- return 1 if it's "cpu"
- return the GPU if there are GPUs
- return an empty list otherwise (e.g. TPU)

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,264 @@
...
+
+def initialize(job_addresses=None, num_processes=None, process_id=None):
+    """Initialize the current process for distributed training."""
+    if not torch.distributed.is_initialized():
+        local_rank = int(os.environ.get("LOCAL_RANK", 0))
+        if torch.cuda.is_available():
+            torch.cuda.set_device(local_rank)
+        backend = "nccl" if torch.cuda.is_available() else "gloo"
+        torch.distributed.init_process_group(backend=backend)
+
```
hertschuh: What about the world size and rank? Why are they not passed?

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,264 @@
...
+        return torch.distributed.get_rank()
+    return 0
+
+
+def _to_backend_device(device_name):
+    """Returns the local device for the current process."""
+    if device_name is not None:
+        device_name_lower = device_name.lower()
+        if "cpu" in device_name_lower:
+            return torch.device("cpu")
```
hertschuh, line 133: What if you're running multiple processes on CPU. How would this work? Do we want to support it?

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,264 @@
...
+        return tensor
+
+    from keras.src.distribution import distribution_lib as dist_lib
+
+    dist = dist_lib.distribution()
+    if not isinstance(dist, dist_lib.ModelParallel):
+        return tensor
+
+    _register_distributed_strategies()
+
```
hertschuh: I think we should do this in `set_distribution`. But that means we have to add a hook in the current `set_distribution` to call a backend specific `set_distribution` which can be empty for JAX.

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,264 @@
...
+    return torch.distributed.tensor.distribute_tensor(
+        tensor, device_mesh=layout.device_mesh, placements=layout.placements
+    )
+
+
+def distribute_variable(value, layout, trainable=True):
+    """Wraps a distributed tensor as a Parameter."""
+    dtensor = distribute_tensor(value, layout)
+    return torch.nn.Parameter(dtensor, requires_grad=trainable)
```
hertschuh: Remove, this no longer exists and is not used.

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,264 @@
...
+    if not isinstance(tensor, torch.Tensor):
+        from keras.src.backend.common import global_state
+        from keras.src.backend.torch import core as torch_core
+
+        dist = global_state.get_global_attribute("distribution")
+        global_state.set_global_attribute("distribution", None)
+        try:
+            tensor = torch_core.convert_to_tensor(tensor)
+        finally:
+            global_state.set_global_attribute("distribution", dist)
```
hertschuh: We're going to have to find a cleaner way to do this...

To me, everything points to `convert_to_tensor` not blindly applying a layout by default.

keras/src/backend/torch/nn.py
```diff
@@ -1447,38 +1474,41 @@ def dot_product_attention(
...
+            q_l, k_l, v_l, m_l, is_causal
         )
     elif flash_attention is True:
-        # Use `raise_error=True` to provide more details if the inputs failed to
-        # use flash attention
+        q_l = query.to_local() if is_dtensor else query
+        k_l = key.to_local() if is_dtensor else key
+        v_l = value.to_local() if is_dtensor else value
+        m_l = mask.to_local() if is_dtensor and mask is not None else mask
         _can_use_flash_attention(
```
hertschuh: Let's say you have 16 GPUs. The way I read this, what's going to happen is that the 16 GPU are each going to get a copy of the full tensors and each do all of the math.

Is there no way to be more efficient?

keras/src/trainers/trainer_test.py
```diff
@@ -500,6 +501,10 @@ def test_compile_eager_vs_jit_torch(self):
             ("steps_per_epoch_jit", False, True, True),
         ]
     )
+    @pytest.mark.skipif(
+        backend.backend() == "torch" and bool(os.environ.get("CI")),
+        reason="Torch Inductor C++ compilation fails in CI (missing cc1plus)",
+    )
```
hertschuh: Wait, I fixed that, inductor is working in the CI, so undo these.

keras/src/ops/numpy.py
```diff
@@ -715,7 +715,9 @@ def arange(start, stop=None, step=None, dtype=None):
     dtype = None if dtype is None else backend.standardize_dtype(dtype)
     if any_symbolic_tensors((start, stop, step)):
         return Arange(dtype=dtype).symbolic_call(start, stop, step=step)
-    return backend.numpy.arange(start, stop, step=step, dtype=dtype)
+    return backend.convert_to_tensor(
+        backend.numpy.arange(start, stop, step=step, dtype=dtype)
+    )
```
hertschuh: This shouldn't be here. It should only be in the torch backend code for `arange`.

keras/src/backend/torch/core.py
```diff
@@ -218,38 +279,60 @@ def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
...
+        from keras.src.distribution import distribution_lib as dist_lib
+
+        if not isinstance(dist, dist_lib.ModelParallel):
+            return x
+        if isinstance(x, DTensor):
+            return x
+
+        if isinstance(x, torch.Tensor):
+            layout = TensorLayout([None] * x.ndim, dist.device_mesh)
+            x = torch_dist_lib.distribute_tensor(x, layout)
```
hertschuh: This is replication, right?

I don't think it should do that, it should stay on the current device and only the current device. Otherwise it defeats the purpose of having multiple devices if you force all operations to be performed on all devices.

hertschuh (changes requested):
Related to the sharding in `convert_to_tensor`.

What happens if you do 
- `x + y` where `x` is a `DTensor` and `y` a regular `Tensor`?
- `x + 1` where `x` is a `DTensor`?

buildwithsuhana:
> Related to the sharding in `convert_to_tensor`.
> 
> What happens if you do
> 
> * `x + y` where `x` is a `DTensor` and `y` a regular `Tensor`?
> * `x + 1` where `x` is a `DTensor`?

1. Mixed Tensor Types (x + y)
 If x is a DTensor and y is a regular torch.Tensor, the operation will fail with a TypeError (got mixed torch.Tensor and DTensor). Torch does not perform implicit promotion of local tensors to distributed tensors.

2. Scalar Operations (x + 1)
 * Python Scalars: If 1 is a Python int or float, the operation succeeds. DTensor supports standard scalar-tensor arithmetic.
 * Scalar Tensors: If 1 is a 0-dimensional torch.Tensor, it fails for the same reason as x + y above.

The general rule for PyTorch DTensor is that all tensor participants in an op must be `DTensor`s belonging to the same DeviceMesh.

keras/src/backend/torch/core.py
```diff
@@ -218,38 +279,64 @@ def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
...
+    dist = global_state.get_global_attribute("distribution")
+    if dist is not None:
+        from keras.src.distribution import distribution_lib as dist_lib
+
+        if not isinstance(dist, dist_lib.ModelParallel):
+            return x
+        if isinstance(x, DTensor):
+            return x
+
+        layout = dist.get_data_layout(x.shape)
```
hertschuh: Why use the data layout by default?

Intuitively, the default layout should be one:
- with the device mesh from the current distribution
- targeting a single device, which is the device the current process is for

Is that what it's doing?

Also, I see that sometimes you pass `layout=None`. What happens then? Does the tensor go on the first device of the mesh? In what cases do you need this?

JAX kind of has a trick whereby instead of passing a layout, you can pass a single device. It would be convenient if we could do that because `get_device()` is already used in many places. However, it's probably not possible since `get_device()` returns a string.

hertschuh:
> > Related to the sharding in `convert_to_tensor`.
> > What happens if you do
> > 
> > * `x + y` where `x` is a `DTensor` and `y` a regular `Tensor`?
> > * `x + 1` where `x` is a `DTensor`?
> 
> 1. Mixed Tensor Types (x + y)
>    If x is a DTensor and y is a regular torch.Tensor, the operation will fail with a TypeError (got mixed torch.Tensor and DTensor). Torch does not perform implicit promotion of local tensors to distributed tensors.
> 2. Scalar Operations (x + 1)
> 
> * Python Scalars: If 1 is a Python int or float, the operation succeeds. DTensor supports standard scalar-tensor arithmetic.
> * Scalar Tensors: If 1 is a 0-dimensional torch.Tensor, it fails for the same reason as x + y above.
> 
> The general rule for PyTorch DTensor is that all tensor participants in an op must be `DTensor`s belonging to the same DeviceMesh.

In many places, we do things like `ops.add(x, 1)`, which should work, except that usually, the first step in an op is to do `convert_to_tensor`. But it would probably be better to let Torch determine what to do with the scalar instead of applying some fixed layout in `convert_to_tensor`.

keras/src/backend/torch/core.py
buildwithsuhana: For layout=None, the function now returns a local tensor placed on the device for the current process rather than defaulting to the first device.

buildwithsuhana:
> > > Related to the sharding in `convert_to_tensor`.
> > > What happens if you do
> > > 
> > > * `x + y` where `x` is a `DTensor` and `y` a regular `Tensor`?
> > > * `x + 1` where `x` is a `DTensor`?
> > 
> > 
> > 
> > 1. Mixed Tensor Types (x + y)
> >    If x is a DTensor and y is a regular torch.Tensor, the operation will fail with a TypeError (got mixed torch.Tensor and DTensor). Torch does not perform implicit promotion of local tensors to distributed tensors.
> > 2. Scalar Operations (x + 1)
> > 
> > 
> > * Python Scalars: If 1 is a Python int or float, the operation succeeds. DTensor supports standard scalar-tensor arithmetic.
> > * Scalar Tensors: If 1 is a 0-dimensional torch.Tensor, it fails for the same reason as x + y above.
> > 
> > The general rule for PyTorch DTensor is that all tensor participants in an op must be `DTensor`s belonging to the same DeviceMesh.
> 
> In many places, we do things like `ops.add(x, 1)`, which should work, except that usually, the first step in an op is to do `convert_to_tensor`. But it would probably be better to let Torch determine what to do with the scalar instead of applying some fixed layout in `convert_to_tensor`.

I've updated convert_to_tensor. It now defaults to a replicated layout instead of a data layout, ensuring constants are available mesh-wide. Regarding scalars, while PyTorch natively handles raw Python values, Keras expects convert_to_tensor to return objects with tensor attributes . To satisfy this while avoiding the TypeError caused by local tensors, I'm converting them to 0-dimensional replicated DTensors. This keeps them compatible with the distributed graph while maintaining the 'tensor' interface Keras requires.

hertschuh:
> I've updated convert_to_tensor. It now defaults to a replicated layout instead of a data layout, ensuring constants are available mesh-wide. Regarding scalars, while PyTorch natively handles raw Python values, Keras expects convert_to_tensor to return objects with tensor attributes . To satisfy this while avoiding the TypeError caused by local tensors, I'm converting them to 0-dimensional replicated DTensors. This keeps them compatible with the distributed graph while maintaining the 'tensor' interface Keras requires.

In this case, do you still need the concepts of `layout="auto"` vs. `layout=None`?

buildwithsuhana:
> > I've updated convert_to_tensor. It now defaults to a replicated layout instead of a data layout, ensuring constants are available mesh-wide. Regarding scalars, while PyTorch natively handles raw Python values, Keras expects convert_to_tensor to return objects with tensor attributes . To satisfy this while avoiding the TypeError caused by local tensors, I'm converting them to 0-dimensional replicated DTensors. This keeps them compatible with the distributed graph while maintaining the 'tensor' interface Keras requires.
> 
> In this case, do you still need the concepts of `layout="auto"` vs. `layout=None`?

Both layout="auto" and layout=None are necessary to provide control over tensor placement. While layout="auto" ensures that tensors are automatically synchronized across the device mesh by default, layout=None is required to produce a standard local tensor. This is useful for internal workflows that must first prepare data locally before wrapping it into a distributed format, without the ability to opt out of automatic distribution, these internal calls would trigger infinite recursion.

hertschuh:
> Both layout="auto" and layout=None are necessary to provide control over tensor placement. While layout="auto" ensures that tensors are automatically synchronized across the device mesh by default, layout=None is required to produce a standard local tensor. This is useful for internal workflows that must first prepare data locally before wrapping it into a distributed format, without the ability to opt out of automatic distribution, these internal calls would trigger infinite recursion.

Sorry for the delay. I'll have to think about this some more...

Meanwhile, can you create a separate PR for the dataset aspect, that is:
- everything under `keras/src/trainers/data_adapters`
- `distribute_tf_dataset` / `distribute_torch_dataset` in `distribution_lib`

Thanks!

keras/src/backend/torch/trainer.py
```diff
@@ -38,18 +43,154 @@ def _should_torch_compile(self):
...
```
amitsrivastava78, line 46: Doesn't move Variable._value for variables that have already initialized as DTensors (DTensor.to() has surprising semantics)
Doesn't move the optimizer state (slot variables stay on CPU)
Silently breaks anyone who built their model on a specific device on purpose
This should at least check whether the model already has parameters on device before moving, and warn or refuse if there's mixed placement.

keras/src/backend/torch/nn.py
```diff
@@ -1430,6 +1430,33 @@ def _can_use_flash_attention(
...
+        query.contiguous(),
+        key.contiguous(),
+        value.contiguous(),
+        attn_mask=mask,
+        is_causal=is_causal,
+        scale=scale,
+    )
+
+
 def dot_product_attention(
```
amitsrivastava78, line 1433: If Q/K/V have Shard(head_dim), then to_local() returns this rank's shard of heads, SDPA computes attention only over that shard, and DTensor.from_local(..., query.placements) re-labels it as if it were the original sharded tensor. The result has the right shape but wrong values because attention across heads doesn't require cross-head communication, but attention across sequence when sequence is sharded does — and the current code doesn't distinguish. There's no test that verifies attention output matches a non-distributed reference.

keras/src/backend/torch/nn.py
```diff
@@ -1430,6 +1430,33 @@ def _can_use_flash_attention(
...
+    if mask is not None:
+        mask = mask.contiguous()
+    return torch.nn.functional.scaled_dot_product_attention(
+        query.contiguous(),
+        key.contiguous(),
+        value.contiguous(),
+        attn_mask=mask,
+        is_causal=is_causal,
+        scale=scale,
+    )
```
hertschuh: This was added in master, you'll need to rebase.

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,354 @@
+import os
+
+import torch
+
+
```
hertschuh, line 5: Please create a separate PR with the following from this file:

- get_device_count
- list_devices
- initialize
- num_processes
- process_id
- to_backend_device

And the corresponding unit tests.

Thanks!

hertschuh (changes requested):

keerthanakadiri:
Hi @buildwithsuhana , Can you kindly resolve the conflicts? Thank you!

PR #22641, opened 2026-04-06
fix: transpose 1D channels_first input in TF depthwise_conv to support dilations

### LINKED ISSUE #22534: DepthwiseConv1D produces incompatible shapes for channels_first inputs with dilation_rate
### Summary
`keras.layers.DepthwiseConv1D` fails at runtime for `data_format='channels_first'` with `dilation_rate > 1`, and the symbolic output shape is also incorrect.

### Reproduction
```python
import keras
import tensorflow as tf
import numpy as np

x = tf.constant(np.random.rand(4, 10, 12), dtype=tf.float32)

layer = keras.layers.DepthwiseConv1D(
    kernel_size=7,
    strides=1,
    padding='valid',
    depth_multiplier=2,
    data_format='channels_first',
    dilation_rate=2,
    activation='relu',
)

# Runtime failure
try:
    y = layer(x)
    print(y.shape)
except Exception as e:
    print(type(e).__name__, e)

# Symbolic shape inference
inp = keras.Input(shape=(10, 12))
out = layer(inp)
print(out.shape)
```

### Expected
- The layer should either:
  - compute a valid output tensor, or
  - raise a clear validation error if the configuration is unsupported.
- Static and dynamic shape behavior should match.

### Actual
- Eager execution raises:
  `Incompatible shapes: [4,10,12] vs. [1,20,1]`
- Symbolic output shape is reported as `(None, 20, 0)`

satishkc7:
## Summary

`tf.nn.depthwise_conv2d` does not support `channels_first` (NCHW) format with dilations on CPU. When `DepthwiseConv1D` is used with `data_format='channels_first'` and `dilation_rate > 1`, the TF backend's `depthwise_conv` function passed NCHW data directly to `tf.nn.depthwise_conv2d`, which fails at runtime with an `Incompatible shapes` error.

## Root Cause

In `keras/src/backend/tensorflow/nn.py`, the 1D depthwise conv path for `channels_first` set `spatial_start_dim = 2` and expanded the input as `(N, C, 1, W)` in NCHW format. Passing NCHW + dilations to `tf.nn.depthwise_conv2d` is not supported on CPU, causing the op to return incorrect results and ultimately an `Incompatible shapes` error when adding the bias.

## Fix

For the 1D `channels_first` case, use the existing `_transpose_spatial_inputs` and `_transpose_spatial_outputs` helpers to transpose to `channels_last` before the convolution and back after, consistent with the pattern used by `max_pool` and `average_pool` in the same file.

## Changes

- `keras/src/backend/tensorflow/nn.py`: use `_transpose_spatial_inputs/outputs` helpers for 1D channels_first depthwise conv, compute in NHWC format
- `keras/src/layers/convolutional/depthwise_conv_test.py`: add `channels_first` + `dilation_rate=(2,)` test cases to both the basic shape test and the correctness test

Fixes #22534

## Contributor Agreement

### DIFF
--- a/keras/src/backend/tensorflow/nn.py
+++ b/keras/src/backend/tensorflow/nn.py
@@ -863,12 +863,13 @@ def depthwise_conv(
         dilation_rate = (dilation_rate,) * num_spatial_dims
     if num_spatial_dims == 1:
         # 1D depthwise conv.
-        if data_format == "channels_last":
-            strides = (1,) + strides * 2 + (1,)
-            spatial_start_dim = 1
-        else:
-            strides = (1, 1) + strides * 2
-            spatial_start_dim = 2
+        # `tf.nn.depthwise_conv2d` does not support `channels_first` with
+        # dilations on CPU. Transpose to `channels_last`, compute, and
+        # transpose back to avoid the limitation.
+        if data_format == "channels_first":
+            inputs = tf.transpose(inputs, perm=[0, 2, 1])
+        strides = (1,) + strides * 2 + (1,)
+        spatial_start_dim = 1
         inputs = tf.expand_dims(inputs, spatial_start_dim)
         kernel = tf.expand_dims(kernel, axis=0)
 
@@ -879,10 +880,13 @@ def depthwise_conv(
             kernel,
             strides,
             padding,
-            data_format=tf_data_format,
+            data_format=_convert_data_format("channels_last", 4),
             dilations=dilation_rate,
         )
-        return tf.squeeze(outputs, [spatial_start_dim])
+        outputs = tf.squeeze(outputs, [spatial_start_dim])
+        if data_format == "channels_first":
+            outputs = tf.transpose(outputs, perm=[0, 2, 1])
+        return outputs
 
     if data_format == "channels_last":
         strides = (1,) + strides + (1,)
--- a/keras/src/backend/tensorflow/nn.py
+++ b/keras/src/backend/tensorflow/nn.py
@@ -867,8 +867,7 @@ def depthwise_conv(
         # dilations on CPU. Transpose to `channels_last`, compute, and
         # transpose back to avoid the limitation.
         need_transpose = data_format == "channels_first" and all(
-            d.device_type == "CPU"
-            for d in tf.config.list_logical_devices()
+            d.device_type == "CPU" for d in tf.config.list_logical_devices()
         )
         if need_transpose:
             inputs = _transpose_spatial_inputs(inputs)
--- a/keras/src/backend/tensorflow/nn.py
+++ b/keras/src/backend/tensorflow/nn.py
@@ -867,7 +867,7 @@ def depthwise_conv(
         # dilations on CPU. Transpose to `channels_last`, compute, and
         # transpose back to avoid the limitation.
         if data_format == "channels_first":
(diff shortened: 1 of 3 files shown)


keras/src/layers/convolutional/depthwise_conv_test.py
```diff
@@ -209,6 +220,11 @@ def test_depthwise_conv1d_basic(
         input_shape,
         output_shape,
     ):
+        if data_format == "channels_first" and backend.backend() == "openvino":
+            self.skipTest(
+                "OpenVINO backend does not support channels_first for "
+                "depthwise_conv."
+            )
```
hertschuh: Can you add the test name to this list instead?
https://github.com/keras-team/keras/blob/master/keras/src/backend/openvino/excluded_concrete_tests.txt

This is how we track feature gaps in OpenVino

keras/src/layers/convolutional/depthwise_conv_test.py
```diff
@@ -369,6 +393,11 @@ def test_depthwise_conv1d(
         data_format,
         dilation_rate,
     ):
+        if data_format == "channels_first" and backend.backend() == "openvino":
+            self.skipTest(
+                "OpenVINO backend does not support channels_first for "
+                "depthwise_conv."
+            )
```
hertschuh: Can you add the test name to this list instead?
https://github.com/keras-team/keras/blob/master/keras/src/backend/openvino/excluded_concrete_tests.txt

This is how we track feature gaps in OpenVino

keras/src/backend/tensorflow/nn.py
```diff
@@ -863,12 +863,13 @@ def depthwise_conv(
...
-            strides = (1,) + strides * 2 + (1,)
-            spatial_start_dim = 1
-        else:
-            strides = (1, 1) + strides * 2
-            spatial_start_dim = 2
+        # `tf.nn.depthwise_conv2d` does not support `channels_first` with
+        # dilations on CPU. Transpose to `channels_last`, compute, and
+        # transpose back to avoid the limitation.
+        if data_format == "channels_first":
+            inputs = _transpose_spatial_inputs(inputs)
```
hertschuh, line 873: Can we add a check to only do this on CPU? I would assume transposing is slower than the current GPU / TPU implementations that currently work with "channels_first".

hertschuh (commented):
Thanks for working on this fix!

keras/src/backend/tensorflow/nn.py
```diff
@@ -866,25 +866,36 @@ def depthwise_conv(
         # `tf.nn.depthwise_conv2d` does not support `channels_first` with
         # dilations on CPU. Transpose to `channels_last`, compute, and
         # transpose back to avoid the limitation.
-        if data_format == "channels_first":
+        need_transpose = data_format == "channels_first" and not bool(
+            tf.config.list_logical_devices("GPU")
```
hertschuh: I assume you also need TPU here. How about:

```python
all([d.device_type == "CPU" for d in tf.config.list_logical_devices()])
```

hertschuh (commented):
Thanks!

One last tweak:

hertschuh:
Please also rebase.

hertschuh (approved):
Thanks!

PR #22659, opened 2026-04-09
[OpenVINO] Fix dynamic tensor slicing

goyaladitya05:
## Description
`OpenVINOKerasTensor.__getitem__` crashed when slice bounds were dynamic tensors (e.g. RandomCrop computing crop_box_hstart : crop_box_hstart + crop_height). 

This PR does the following changes:

- Replace index == builtins.slice(None) with an explicit attribute check to avoid triggering OpenVINOKerasTensor.__eq__ during slice comparison.
- Handle dynamic OpenVINOKerasTensor start/stop values in slice bound construction, instead of assuming all bounds are static int.

## Contributor Agreement


hertschuh (approved):
Thanks for the fix!

PR #22661, opened 2026-04-09
Pass Python strings to TextVectorization standardize on non-TF backends

### LINKED ISSUE #22626: TextVectorization custom standardize callable receives EagerTensor instead of a Python string
Environment:

- Keras version: 3.14.0
- Backend: JAX
- Python: 3.12

When using TextVectorization with a custom standardize callable and a non-TensorFlow backend (e.g. JAX), calling .adapt() raises an AttributeError because the callable receives a tf.EagerTensor instead of a plain Python string.

```
import os

import re
import string
import keras

strip_chars = string.punctuation
def my_standardize(input_string):
    input_string = input_string.lower()          # AttributeError here
    return re.sub(f"[{re.escape(strip_chars)}]", "", input_string)

layer = keras.layers.TextVectorization(standardize=my_standardize)
layer.adapt(["Hello, world."])
```

`AttributeError: 'tensorflow.python.framework.ops.EagerTensor' object has no attribute 'lower'`

TextVectorization should be independent of TensorFlow.

MarcosAsh:
## Description
Fixes #22626.
 
When using `TextVectorization` with a custom `standardize` callable on the JAX or torch backend, the callable was receiving a `tf.EagerTensor` instead of a Python string. Idiomatic Python like `input_string.lower()` then crashed with `AttributeError: 'EagerTensor' object has no attribute 'lower'`, contradicting the layer's documented backend-agnostic behavior outside the compiled graph.                          
                                       
This PR routes callable standardize through a new `_apply_python_standardize` helper on non-TF backends: it decodes the input to plain Python strings, invokes the callable per element, and re-packs the result so the existing TF-backed split / ngrams pipeline can take it from there. Results that  come back as a `tf.Tensor` (i.e. tf-style callables run on a  non-TF backend) are coerced back to `str` so they survive the re-pack. TF backend behavior is unchanged.         

## Contributor Agreement

### DIFF
--- a/keras/src/layers/preprocessing/text_vectorization.py
+++ b/keras/src/layers/preprocessing/text_vectorization.py
@@ -410,7 +410,8 @@ class TextVectorization(Layer):
         Arguments:
             data: The data to train on. It can be passed either as a
                 batched `tf.data.Dataset`, as a list of strings,
-                or as a NumPy array.
+                as a NumPy array, or as any iterable of batches (e.g.
+                a generator yielding batches of strings).
             steps: Integer or `None`.
                 Total number of steps (batches of samples) to process.
                 If `data` is a `tf.data.Dataset`, and `steps` is `None`,
@@ -425,11 +426,18 @@ class TextVectorization(Layer):
                 data = data.take(steps)
             for batch in data:
                 self.update_state(batch)
+        elif hasattr(data, "__iter__") and not (
+            isinstance(data, (list, tuple, np.ndarray))
+            or backend.is_tensor(data)
+            or tf.is_tensor(data)
+        ):
+            for i, batch in enumerate(data):
+                if steps is not None and i >= steps:
+                    break
+                self.update_state(batch)
         else:
             data = tf_utils.ensure_tensor(data, dtype="string")
             if data.shape.rank == 1:
-                # A plain list of strings
-                # is treated as as many documents
                 data = tf.expand_dims(data, -1)
             self.update_state(data)
         self.finalize_state()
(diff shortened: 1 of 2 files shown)


hertschuh (changes requested):
I don't think the contract should be that we should convert everything to individual Python strings. Maybe `standardize` needs to have access to the array of strings as a whole to be able to process it.

I think the contract should be that `standardize` receives:
- TF tensors with TensorFlow
- Numpy string arrays on other backends. This means that people would use `np.chars` and `np.strings` operations.

So the code as written in the issue would'nt work.

MarcosAsh:
Hi thanks for the review ! I changed it so the Non-TF backends now hand the callable a single NumPy unicode array (dtype.kind == 'U')  instead of per-element strings.

keras/src/layers/preprocessing/text_vectorization.py
```diff
@@ -526,20 +532,69 @@ def set_vocabulary(self, vocabulary, idf_weights=None):
...
+                return value.decode(self._encoding)
+            return str(value)
+
+        if np_inputs.ndim == 0:
+            decoded = np.array(_decode(np_inputs.item()), dtype=np.str_)
+        else:
+            flat = np_inputs.reshape(-1)
+            decoded = np.array(
+                [_decode(item) for item in flat], dtype=np.str_
+            ).reshape(np_inputs.shape)
```
hertschuh: Why is this decoding needed?

keras/src/layers/preprocessing/text_vectorization.py
hertschuh, line 633: If you look at what happens here, you'll see that if converts inputs to numpy and then to TF, then `_apply_python_standardize` re-converts to numpy and then to TF again.

This should be cleaned up to remove the extra conversion.

Note that the other place where `_preprocess` is used is `update_state`. Probably `_preprocess` should take care of what currently happens here between lines 634 and 637

keras/src/layers/preprocessing/text_vectorization.py
```diff
@@ -526,20 +532,69 @@ def set_vocabulary(self, vocabulary, idf_weights=None):
         """
         self._lookup_layer.set_vocabulary(vocabulary, idf_weights=idf_weights)
 
+    def _apply_python_standardize(self, inputs):
```
hertschuh: Let's change the name to something like `_apply_standardize_with_np_arrays`.

hertschuh (commented):
Thanks for the update!

keras/src/layers/preprocessing/text_vectorization.py
MarcosAsh: Done, I renamed it.

keras/src/layers/preprocessing/text_vectorization.py
MarcosAsh: Dropped the per-element decode loop. Single dtype.kind check now, with vectorized np.char.decode only for byte/object arrays.

keras/src/layers/preprocessing/text_vectorization.py
MarcosAsh: Moved the input coercion from call() into _preprocess(), so update_state shares the path and the np -> tf -> np -> tf round-trip is gone.

MarcosAsh:
Thanks for the review! I made all the changes requested.

hertschuh (approved):
Thanks for your work on this!

PR #22662, opened 2026-04-09
Validate SeparableConv output shape in build()

### LINKED ISSUE #22496: SeparableConv1D returns incorrect symbolic output shape for valid strided inputs
### Bug report

`keras.layers.SeparableConv1D` computes a correct output shape for eager tensors, but the symbolic output shape from `keras.Input()` is wrong when using `padding='valid'` and `strides > 1`.

### Reproduction
```python
import keras
import tensorflow as tf
import numpy as np

x_eager = tf.constant(np.random.rand(4, 10, 12), dtype=tf.float32)
x_sym = keras.Input(shape=(10, 12))

layer = keras.layers.SeparableConv1D(64, 11, strides=5, padding='valid', activation='relu')

y_eager = layer(x_eager)
y_sym = layer(x_sym)

print(y_eager.shape)
print(y_sym.shape)
```

### Expected result
Both outputs should report the same spatial shape.

### Actual result
- Eager output shape: `(4, 10, 64)`
- Symbolic output shape: `(None, 0, 64)`

This indicates incorrect static shape inference for `SeparableConv1D`.

MarcosAsh:
## Description
Fixes #22496.

`SeparableConv1D` (and `SeparableConv2D`) silently ran the convolution on out-of-bounds memory when the kernel/stride combination would yield a non-positive spatial dimension. The user's reproducer (`SeparableConv1D(64, 11, strides=5, padding='valid')` on input length 10) returned shape `(4, 10, 64)` filled with garbage values like `1e30`, while `Conv1D` with the same parameters raised a clear `ValueError`.      
                                                       
Root cause: `BaseSeparableConv.build()` never called `self.compute_output_shape(input_shape)`, so the validation logic in `compute_conv_output_shape` (introduced by #22092) was never reached on the eager path. `BaseConv.build()` already calls it for exactly this reason. This PR mirrors that call in `BaseSeparableConv.build()`. 

## Contributor Agreement

### DIFF
--- a/keras/src/layers/convolutional/base_separable_conv.py
+++ b/keras/src/layers/convolutional/base_separable_conv.py
@@ -183,6 +183,10 @@ class BaseSeparableConv(Layer):
             self.filters,
         )
 
+        # compute_output_shape contains some validation logic for the input
+        # shape, and makes sure the output shape has all positive dimensions.
+        self.compute_output_shape(input_shape)
+
         self.depthwise_kernel = self.add_weight(
             name="depthwise_kernel",
             shape=depthwise_kernel_shape,
--- a/keras/src/layers/convolutional/separable_conv_test.py
+++ b/keras/src/layers/convolutional/separable_conv_test.py
@@ -204,6 +204,35 @@ class SeparableConvBasicTest(testing.TestCase):
                 dilation_rate=(2, 1),
             )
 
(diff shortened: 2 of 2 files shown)


hertschuh (approved):
Thanks for adding this!

PR #22666, opened 2026-04-11
Fix GroupNormalization NaN with mixed_float16 policy

### LINKED ISSUE #22586: [Bug] Mixed Precision (float16) numerical instability in GroupNormalization with small epsilon
## Description

## Poc

```
import os
import keras
import numpy as np

output = model(extreme_data)


```

## Observed Result

```
NaN detected: True

```

## Expected Behavior

The layer should either:

## Actual Behavior

## Notes

Grizouforever:
## Summary

- **Fixes #22586**: `GroupNormalization` produces NaN outputs when using `mixed_float16` policy with small epsilon or large input values
- Moves the float32 upcast from `_apply_normalization` to the beginning of `call()`, before `_reshape_into_groups`, so inputs are protected from float16 overflow before any computation occurs
- Matches the approach already used by `BatchNormalization` (which upcasts at the top of its `call()` method)

## Root Cause

With `mixed_float16` policy and `autocast=True` (default), Keras casts layer inputs to float16 **before** `call()` is invoked. This causes two failure modes:

1. **Input overflow**: Values > 65504 overflow to `inf` in float16, causing NaN propagation through mean/variance computation
2. **Epsilon underflow**: Epsilon values < ~6.1e-5 round to zero in float16, causing division by zero when variance is small

The existing float32 upcast inside `_apply_normalization` cannot recover values that are already `inf` from the float16 cast.

## Changes

- `group_normalization.py`: Upcast inputs to float32 at the start of `call()`, before reshaping, and cast output back to `compute_dtype` at the end
- `group_normalization_test.py`: Added regression test `test_mixed_float16_no_nan` that verifies no NaN output with large inputs and small epsilon under `mixed_float16`

## Test plan

- [ ] New test `test_mixed_float16_no_nan` validates the fix with values near float16 max and epsilon=1e-12
- [ ] Existing `GroupNormalizationTest` tests continue to pass (no behavioral change for float32 inputs)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/keras/src/layers/normalization/group_normalization.py
+++ b/keras/src/layers/normalization/group_normalization.py
@@ -167,12 +167,18 @@ class GroupNormalization(Layer):
         super().build(input_shape)
 
     def call(self, inputs):
+        input_dtype = self.compute_dtype
+        # GN is prone to overflow with float16/bfloat16 inputs, so we
+        # upcast to float32 for the subsequent computations.
+        compute_dtype = backend.result_type(inputs.dtype, "float32")
+        inputs = ops.cast(inputs, compute_dtype)
+
         reshaped_inputs = self._reshape_into_groups(inputs)
         normalized_inputs = self._apply_normalization(
             reshaped_inputs, inputs.shape
         )
         outputs = ops.reshape(normalized_inputs, ops.shape(inputs))
-        return ops.cast(outputs, self.compute_dtype)
+        return ops.cast(outputs, input_dtype)
 
     def _reshape_into_groups(self, inputs):
         input_shape = ops.shape(inputs)
--- a/keras/src/layers/normalization/group_normalization_test.py
+++ b/keras/src/layers/normalization/group_normalization_test.py
@@ -201,3 +201,22 @@ class GroupNormalizationTest(testing.TestCase):
                 dtype="mixed_bfloat16",
             )
             self.assertFalse(any("epsilon" in str(x.message) for x in w))
+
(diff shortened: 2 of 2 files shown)


keerthanakadiri:
Hi @Grizouforever, Can you please sign the CLA? Thank you!

Saumay:
Confirmed the NaN still reproduces on master (e94cb07d5) with the issue's repro.

The full fix probably needs `self.autocast = False` in `__init__` in addition to the upcast this PR already adds. Happy to pick this up as a follow-up PR if this one is stalled.

hertschuh:
> Confirmed the NaN still reproduces on master ([e94cb07](https://github.com/keras-team/keras/commit/e94cb07d59cfb0eaab76e3ec3076808ab71e342c)) with the issue's repro.
> 
> The full fix probably needs `self.autocast = False` in `__init__` in addition to the upcast this PR already adds. Happy to pick this up as a follow-up PR if this one is stalled.

To be clear, the fix is for the user to pass `autocast = False` in the constructor. We shouldn't force that decision on behalf of the user.

Note that https://github.com/keras-team/keras/pull/22589 already addresses the epsilon part of the issue.

So the only thing I would do on top of that is document clearly that `autocast = False` can be used in mixed precision scenarios to improve numerical stability (which is true for any layer).

Saumay:
Understood. Noted on #22589. I'll close this out on my side.

PR #22671, opened 2026-04-13
Remove duplication of qr in linalg.py and math.py for all backends

goyaladitya05:
## Description
keras.ops.qr is defined in linalg.py, and always delegates to linalg.py implementation of qr for all backends. 
This PR removes the duplication of qr in math.py, since there is no path that calls these implementations.

## Contributor Agreement


hertschuh (approved):
Good catch, thanks!

PR #22676, opened 2026-04-14
Validate DepthwiseConv output shape in build() 

MarcosAsh:
## Description
Follow-up to #22662, which added the same check to SeparableConv.

`BaseDepthwiseConv.build()` wasn't calling `self.compute_output_shape(input_shape)`, so `DepthwiseConv1D`/`DepthwiseConv2D` silently produced wrong-shaped outputs on eager inputs when the kernel/stride combo yielded a non-positive spatial dim. The symbolic path already raised `ValueError("Computed output size would be  zero or negative. ...")` via `compute_conv_output_shape`; this brings the eager path in line.

One-line fix in `base_depthwise_conv.py` and a regression test covering both 1D and 2D (with `data_format` pinned to `channels_last` so the spatial dim under test is the same regardless of the backend's default `image_data_format`).

## Contributor Agreement

### DIFF
--- a/keras/src/layers/convolutional/depthwise_conv_test.py
+++ b/keras/src/layers/convolutional/depthwise_conv_test.py
@@ -328,6 +328,34 @@ class DepthwiseConvBasicTest(testing.TestCase):
                 depth_multiplier=2, kernel_size=(2, 2), strides=(1, 0)
             )
 
+    def test_invalid_output_shape_raises(self):
+        # DepthwiseConv1D/2D used to silently produce a wrong-shaped output
+        # on eager inputs when the kernel/stride combination would yield a
+        # non-positive spatial dimension. It must raise the same `ValueError`
+        # that the symbolic path raises. Pin `data_format` so the spatial dim
+        # under test is the same regardless of the backend's default
+        # `image_data_format`.
+        x = np.random.rand(4, 10, 3).astype("float32")
+        layer = layers.DepthwiseConv1D(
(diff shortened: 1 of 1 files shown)


amadhan882 (commented):
@MarcosAsh ,

The fix correctly enforces output shape validation during build and aligns eager and symbolic behavior.

However, calling `compute_output_shape(input_shape)` unconditionally may introduce a regression when the channel dimension is `None` (e.g., during symbolic tracing). In such cases, `compute_output_shape` may attempt to perform arithmetic with `None`, leading to a `TypeError`.

It may be safer to guard this validation with a check ensuring the channel dimension is known, or to make `compute_output_shape` robust to `None` values.

Otherwise, this change could break valid symbolic use cases.

MarcosAsh:
@amadhan882 

I see why gemini made that comment but if the channel dim is None, build() already fails at line 172-180 (add_weight(shape=depthwise_shape, ...) before the new compute_output_shape(input_shape) call. You cant create a kernel weight with None in its shape. So that scenerio gemini and you flagged is impossible by the time my line runs.

hertschuh:
@MarcosAsh 

Can you rebase (which will retrigger the tests). Thanks!

MarcosAsh:
Hi, I just rebased thanks for the review! I noticed that the fix was added in #22686 so its just the tests. If this isn't needed anymore I can just close it.

hertschuh (approved):
Thanks for adding the test!

PR #22680, opened 2026-04-15
Allow Initializers to handle the distribution layout.

hertschuh:
### Problem

Some variables are so large that they don't fit on a single accelerator. Currently, the sharding is usually applied after the initializer has be called, which created the value on the first accelerator. Alternatively, we create a jitted function that is able to apply the sharding, but jitting has a cost (jitting all initializers is prohibitively slow) and some limitations.

### Solution

This change allows users to implemented their own initializer which handles the layout, meaning its able to return a sharded array. This allow things as simple as:
```python
class MyInitializer(keras.initializers.Initializer):
    def __call__(self, shape, dtype=None, layout=None):
        return jnp.ones(shape, dtype=dtype, out_sharding=layout)
```

- Also made `layout` a backend independent argument in `Variable.__init__` to allow support on other backends in the future.

### What's Next

- Upgrade all random ops and some numpy ops to optionally accept a `layout` argument like JAX.
- Plumb the layout argument in all the Keras built-in initializers.

## Contributor Agreement

### DIFF
--- a/keras/src/backend/common/variables.py
+++ b/keras/src/backend/common/variables.py
@@ -40,6 +40,9 @@ class Variable:
             to `"none"`.
         name: Optional. A unique name for the variable. Automatically generated
             if not set.
+        layout: Optional. Sharding layout for the variable. Can be a
+            `keras.distribution.TensorLayout` or a backend-specific layout.
+        kwargs: Additional backend-specific keyword arguments.
 
     Attributes:
         shape: The shape of the variable (tuple of integers).
@@ -51,7 +54,6 @@ class Variable:
         value: The current value of the variable (NumPy array or tensor).
         name: The name of the variable (string).
         path: The path of the variable within the Keras model or layer (string).
-        kwargs: Additional backend-specific keyword arguments.
 
     Examples:
 
@@ -99,9 +101,9 @@ class Variable:
         aggregation="none",
         synchronization="auto",
         name=None,
+        layout=None,
         **kwargs,
     ):
-        del kwargs
         name = name or auto_name(self.__class__.__name__)
         if not isinstance(name, str) or "/" in name:
             raise ValueError(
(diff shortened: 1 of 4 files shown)


keras/src/backend/jax/core.py
```diff
@@ -53,12 +48,18 @@ def _initialize(self, value):
...
-        shape = self._shape
-        if should_shard_at_init(layout, shape):
+
+        if self._layout is None:
+            super()._initialize_with_initializer(initializer)
+        elif "layout" in inspect.signature(initializer.__call__).parameters:
+            if jax.set_mesh(self._layout.mesh):
+                self._value = initializer(
+                    self._shape, dtype=self._dtype, layout=self._layout
+                )
```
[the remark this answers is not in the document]
hertschuh: Oops typo, fixed

keras/src/backend/jax/core.py
```diff
@@ -53,12 +48,18 @@ def _initialize(self, value):
 
     def _initialize_with_initializer(self, initializer):
         self._initialize_layout()
-        layout = self._layout
-        shape = self._shape
-        if should_shard_at_init(layout, shape):
+
+        if self._layout is None:
+            super()._initialize_with_initializer(initializer)
+        elif "layout" in inspect.signature(initializer.__call__).parameters:
```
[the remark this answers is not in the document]
hertschuh: Fixed

keras/src/backend/common/variables.py
```diff
@@ -99,9 +100,8 @@ def __init__(
         aggregation="none",
         synchronization="auto",
         name=None,
-        **kwargs,
```
jeffcarp, line 102: Do we want to keep this in case other backends need to use additional kwargs?

keras/src/initializers/initializer_test.py
```diff
@@ -0,0 +1,90 @@
...
+
+from keras.src import backend
+from keras.src import initializers
+from keras.src import testing
+from keras.src.distribution import distribution_lib
+
+
+class InitializerWithLayout(initializers.Initializer):
+    def __call__(self, shape, dtype=None, layout=None):
+        import jax
```
jeffcarp: Does this need to be imported here?

jeffcarp (approved):

keras/src/backend/common/variables.py
hertschuh: Re-added

keras/src/initializers/initializer_test.py
hertschuh: No, this was in expectation for when it will be changed it to the keras op. Moved up.

PR #22681, opened 2026-04-15
[OpenVino] Revert change to `_parameterize_data`.

hertschuh:
This is a follow-up to https://github.com/keras-team/keras/pull/22670
    
The change in `_parameterize_data` to make the batch size dynamic breaks `generate` in KerasHub. The dynamic batch size aspect is now opt-in.

## Contributor Agreement

### DIFF
--- a/keras/src/backend/openvino/trainer.py
+++ b/keras/src/backend/openvino/trainer.py
@@ -86,23 +86,23 @@ class OpenVINOTrainer(base_trainer.Trainer):
 
         self.test_function = test_step
 
-    def _parameterize_data(self, data):
+    def _parameterize_data(self, data, dynamic_batch_size=False):
         if isinstance(data, (list, tuple)):
             parametrize_data = []
             for elem in data:
-                param_elem = self._parameterize_data(elem)
+                param_elem = self._parameterize_data(elem, dynamic_batch_size)
                 parametrize_data.append(param_elem)
         elif isinstance(data, dict):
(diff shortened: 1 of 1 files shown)


divyashreepathihalli (approved):

sachinprasadhs (approved):

hertschuh:
@goyaladitya05

Analysis from @sachinprasadhs :

> The issue is caused by changes in trainer.py from the above commit, which forces the batch dimension to be dynamic (-1). KerasHub then attempts to create a cache of zeros using this dynamic shape, but the OpenVINO implementation of ops.zeros in numpy.py does not support shapes containing dynamic dimensions.

goyaladitya05:
> The issue is caused by changes in trainer.py from the above commit, which forces the batch dimension to be dynamic (-1). KerasHub then attempts to create a cache of zeros using this dynamic shape, but the OpenVINO implementation of ops.zeros in numpy.py does not support shapes containing dynamic dimensions.

LGTM, thanks!

hertschuh:
> > The issue is caused by changes in trainer.py from the above commit, which forces the batch dimension to be dynamic (-1). KerasHub then attempts to create a cache of zeros using this dynamic shape, but the OpenVINO implementation of ops.zeros in numpy.py does not support shapes containing dynamic dimensions.
> 
> LGTM, thanks!

Feel free to create a better fix if you can think of one.

PR #22683, opened 2026-04-16
Fix Blackman output shape inference for symbolic tensors

Jah-yee:
Good day,

## Problem
The  operation in  had an incorrect  implementation. It returned a KerasTensor with the same shape as the input (which is  for scalar inputs), instead of correctly specifying the 1D output shape .

This caused incorrect shape inference when  was called with symbolic inputs (KerasTensors).

## Solution
Updated  to correctly handle both 1D tensor inputs (preserving their shape) and scalar inputs (using the scalar value as the 1D dimension).

## Changes
- : Modified  to return a 1D output shape  instead of using the input's shape directly.

## Testing
All existing blackman tests pass (14 tests), including:
- : Basic functionality with random window lengths
- : M=1 case with concrete values
- : Symbolic tensor shape inference

Warmly,
RoomWithOutRoof

### DIFF
--- a/keras/src/ops/numpy.py
+++ b/keras/src/ops/numpy.py
@@ -1867,7 +1867,7 @@ class Blackman(Operation):
         # The output is a 1D tensor of shape (window_length,).
         # If x is a 1D tensor representing window lengths, preserve its shape.
         # If x is a scalar (0D), use x itself as the window length dimension.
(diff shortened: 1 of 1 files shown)


hertschuh:
https://github.com/keras-team/keras/issues/22538 has already been fixed.

PR #22686, opened 2026-04-16
Validate output shape in BaseDepthwiseConv.build

### LINKED ISSUE #22534: DepthwiseConv1D produces incompatible shapes for channels_first inputs with dilation_rate

rstar327:
## Description

Fixes #22534.

`DepthwiseConv1D` (and `DepthwiseConv2D`) silently produce tensors with zero-size spatial dimensions in eager mode when the kernel/dilation configuration is incompatible with the input size, while the symbolic path correctly raises a `ValueError`.

The root cause is that `BaseDepthwiseConv.build()` does not call `self.compute_output_shape(input_shape)` for validation, unlike `BaseConv.build()` which already does this.

This one-line fix adds `self.compute_output_shape(input_shape)` to `BaseDepthwiseConv.build()`, matching the pattern in `BaseConv`. This ensures both eager and symbolic paths raise a clear error for invalid configurations.

## Contributor Agreement


hertschuh (approved):
Thanks for the fix!

PR #22687, opened 2026-04-17
Validate input channels match kernel in depthwise/separable conv

### LINKED ISSUE #22516: keras.ops.separable_conv mishandles data_format shape validation and kernel shape inference
## Summary
`keras.ops.separable_conv` raises shape errors when used with TensorFlow backend, and the error reporting/shape handling appears inconsistent between eager tensors and `Keras.Input` placeholders.

## Reproduction
```python
import keras
import numpy as np

x = np.random.rand(2, 5, 5, 3).astype(np.float32)
dw = np.random.rand(3, 3, 3, 1).astype(np.float32)
pw = np.random.rand(1, 1, 3, 4).astype(np.float32)

# Eager
keras.ops.separable_conv(x, dw, pw, padding="same", data_format="channels_first", dilation_rate=2)

# Symbolic
x2 = keras.Input(shape=(5, 5, 3))
dw2 = keras.Input(shape=(3, 3, 3, 1))
pw2 = keras.Input(shape=(1, 1, 3, 4))
keras.ops.separable_conv(x2, dw2, pw2, padding="same", data_format="channels_first", dilation_rate=2)
```

## Expected
- The op should either:
  - work correctly for the declared `data_format`, or
  - raise a clear validation error before dispatching to the backend.
- Eager and symbolic execution should behave consistently.

## Actual
- Eager execution fails with:
  `input depth must be evenly divisible by filter depth: 5 vs 3`
- Symbolic execution fails with:
  `Kernel shape must have the same length as input, but received kernel of shape (None, 3, 3, 5, 3) and input of shape (None, 5, 5, 3)`

## Notes
The reproducer uses the same tensor shapes for both paths, but the backend behavior differs and the op does not appear to handle/validate the `data_format` and kernel shapes consistently.

rstar327:
## Description

Fixes #22516.

`keras.ops.separable_conv` and `keras.ops.depthwise_conv` previously gave confusing error messages when the input's channel dimension didn't match the kernel's expected input channels (e.g., passing channels-last shaped data with `data_format="channels_first"`):

- **Eager**: TF backend error: `input depth must be evenly divisible by filter depth: 5 vs 3`
- **Symbolic**: `Kernel shape must have the same length as input`

Neither message clearly identifies the root cause: a mismatch between the input's channel count and the kernel's expected input channels.

This PR adds explicit channel validation in both the symbolic path (`compute_output_spec`) and the eager path (before dispatching to the backend), producing a clear error:

```
ValueError: The number of input channels must match the kernel's input channels.
Received: input channels=5, kernel input channels=3, data_format='channels_first'.
```

Changes:
- Added `_check_input_channels_match_kernel` helper in `keras/src/ops/nn.py`
- Added channel validation in `DepthwiseConv.compute_output_spec` (symbolic path)
- Added channel validation in `depthwise_conv()` and `separable_conv()` (eager path)

## Contributor Agreement


keras/src/ops/nn.py
```diff
@@ -1619,6 +1651,7 @@ def depthwise_conv(
         return DepthwiseConv(
             strides, padding, data_format, dilation_rate
         ).symbolic_call(inputs, kernel)
+    _check_input_channels_match_kernel(inputs, kernel, data_format)
```
hertschuh: The reason why we don't do any validation involving the inputs in the backend-independent code is that there are scenarios where this won't work:
- inputs can be a nested Python array and won't have a shape
- inputs can be a tracer / placeholder tensor and have dynamic dimensions

I would check which backend don't handle this correctly, then add this check for the backends that need it after the `convert_to_tensor(inputs)`.

hertschuh (changes requested):
Thanks for looking into this!

rstar327:
Thanks for the review! I've pushed a refactor addressing the feedback:

- Removed the check from `keras/src/ops/nn.py` (both the standalone helper and the `DepthwiseConv.compute_output_spec` check), so the backend-independent path no longer touches `inputs.shape`.
- Added a single shared helper `check_depthwise_conv_input_channels` in `keras/src/backend/common/backend_utils.py`.
- Called it from each of the four backends (TF / JAX / Torch / NumPy) **after** `convert_to_tensor(inputs)`, so we only run it against a concrete tensor. The JAX / Torch / NumPy `separable_conv` already delegate to their own `depthwise_conv`, so the check fires once per call.

Verified all four backends now raise the same clear `ValueError` on the #22516 repro, and the existing `nn_test.py` + `separable_conv_test.py` + `depthwise_conv_test.py` suites pass on all four backends. Let me know if you'd like any adjustments.

keras/src/backend/common/backend_utils.py
```diff
@@ -307,6 +307,30 @@ def standardize_axis_for_numpy(axis):
...
+        inputs.shape[-1] if data_format == "channels_last" else inputs.shape[1]
+    )
+    kernel_input_channels = kernel.shape[-2]
+    if (
+        input_channels is None
+        or kernel_input_channels is None
+        or input_channels == kernel_input_channels
+    ):
+        return
+    raise ValueError(
```
hertschuh: Write it like this:

```python
if (
    isinstance(input_channels, int)
    and isinstance(kernel_input_channels, int)
    and input_channels != kernel_input_channels
):
    raise ValueError(...)
```

The reason is that dynamic dimensions can come in different forms, not just `None`, during tracing.

keras/src/backend/tensorflow/nn.py
hertschuh, line 860: Can you move the `check_depthwise_conv_input_channels` after the check on `num_spatial_dims`.

The reason is that if the number of dimensions is incorrect, the math in `check_depthwise_conv_input_channels` will be incorrect and meaningless and may raise a confusing error.

keras/src/backend/numpy/nn.py
```diff
@@ -683,6 +686,9 @@ def depthwise_conv(
     dilation_rate=1,
 ):
     data_format = backend.standardize_data_format(data_format)
+    inputs = convert_to_tensor(inputs)
+    kernel = convert_to_tensor(kernel)
+    check_depthwise_conv_input_channels(inputs, kernel, data_format)
```
hertschuh, line 691: Any reason why you didn't do `separable_conv`?

keras/src/backend/tensorflow/nn.py
hertschuh, line 942: Can you move the `check_depthwise_conv_input_channels` after the check on `num_spatial_dims`.

The reason is that if the number of dimensions is incorrect, the math in `check_depthwise_conv_input_channels` will be incorrect and meaningless and may raise a confusing error.

keras/src/backend/jax/nn.py
```diff
@@ -783,6 +786,9 @@ def depthwise_conv(
     dilation_rate=1,
 ):
     data_format = backend.standardize_data_format(data_format)
+    inputs = convert_to_tensor(inputs)
+    kernel = convert_to_tensor(kernel)
+    check_depthwise_conv_input_channels(inputs, kernel, data_format)
```
hertschuh, line 791: Any reason why you didn't do `separable_conv`?

keras/src/backend/torch/nn.py
```diff
@@ -656,7 +659,10 @@ def depthwise_conv(
     data_format=None,
     dilation_rate=1,
 ):
+    data_format = backend.standardize_data_format(data_format)
+    inputs = convert_to_tensor(inputs)
     kernel = convert_to_tensor(kernel)
+    check_depthwise_conv_input_channels(inputs, kernel, data_format)
```
hertschuh, line 665: Any reason why you didn't do `separable_conv`?

rstar327:
Thanks for the review! Pushed `2fae3a72d` addressing all four points:

1. **`check_depthwise_conv_input_channels` now uses `isinstance(..., int)`** instead of an explicit `None` check, so dynamic/tracing dims in other forms fall through to the no-op path instead of accidentally passing the validation.
2. **TF `depthwise_conv` / `separable_conv`**: moved the channel check after the `num_spatial_dims > 2` guard so we don't run wrong-axis math on a misranked input.
3. **JAX / Torch / NumPy `separable_conv`**: added the check explicitly. My original reasoning was that these three backends' `separable_conv` delegates to their own `depthwise_conv` (which already has the check), but you're right that making it explicit is cleaner and safer against future refactors.

Re-verified the #22516 repro raises the clear error on all four backends for both `depthwise_conv` and `separable_conv`, and the existing `nn_test.py` depthwise/separable tests (24 each) pass on TF / JAX / Torch / NumPy.

hertschuh:
> 3. **JAX / Torch / NumPy `separable_conv`**: added the check explicitly. My original reasoning was that these three backends' `separable_conv` delegates to their own `depthwise_conv` (which already has the check), but you're right that making it explicit is cleaner and safer against future refactors.

Oh, I did not realize that.

hertschuh (approved):
Thanks for your contribution!

PR #22689, opened 2026-04-17
Support negative axes in keras.ops.transpose

mgomes0:
## Description
keras.ops.transpose does not fully support negative axes in all backends. The support seems to be based on the support of the corresponding op in each backend. I believe this is contrary to the policy of the other ops.
This PR adds support for that by using `canonicalize_axis` and also adds tests.

## Contributor Agreement


miguelteixeiragomes:
Hey @gbaned , tagging you since you were originally assigned to this PR.
I created what I thought was the correct fix for handling negative axis indexes, but it fails several tests, but I don't understand the failure.
I thought the function I edited in src/ops/numpy.py was the entrypoint used by all backends when calling a keras.ops function.

hertschuh:
Can you rebase this change? (this will fix the JAX tests)
Can you look into the OpenVino tests are now failing?

miguelteixeiragomes:
Thanks for the comment @hertschuh ! I merged latest `master` commits.

Added the canonicalization of the axes to `keras.ops.transpose` and to `knp.Transpose`. In the latter, I didn't want to add a new member nor canonicalize `self.axes`, as I'm not very familiar with the whole architecture of the code.

hertschuh (changes requested):
Please undo the changes in `numpy.py`.

- `operation_utils.compute_transpose_output_shape` already handles it correctly.
- `x` may not have a shape in this context (nested Python arrays)

Instead, fix tensorflow and OpenVino (other backends are fine):

```diff
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -3484,6 +3484,8 @@ def transpose(x, axes=None):
         output = tf.sparse.transpose(x, perm=axes)
         output.set_shape(compute_transpose_output_shape(x.shape, axes))
         return output
+    if axes:
+        axes = tuple(canonicalize_axis(axis, len(x.shape)) for axis in axes)
     return tf.transpose(x, perm=axes)

--- a/keras/src/backend/openvino/numpy.py
+++ b/keras/src/backend/openvino/numpy.py
@@ -4915,8 +4915,8 @@ def transpose(x, axes=None):
             rank_minus_one, const_minus_one, const_minus_one, "i64"
         ).output(0)
     else:
-        if isinstance(axes, tuple):
-            axes = list(axes)
+        rank = x.get_partial_shape().rank.get_length()
+        axes = [canonicalize_axis(a, rank) for a in axes]
         axes = ov_opset.constant(axes, Type.i32).output(0)
     return OpenVINOKerasTensor(ov_opset.transpose(x, axes).output(0))
```

keras/src/ops/numpy.py
```diff
@@ -8750,6 +8756,8 @@ def transpose(x, axes=None):
     Returns:
         `x` with its axes permuted.
     """
+    if axes:
+        axes = tuple(canonicalize_axis(axis, len(x.shape)) for axis in axes)
```
hertschuh: Remove this, `x` may not have a shape at this stage and this is handled by each backend.

So you can revert the whole file.

hertschuh (changes requested):

keras/src/ops/numpy.py
mgomes0: Hmm, I didn't realize in that op it could also have no shape.
Done, thank you!

hertschuh (approved):
Thanks!

PR #22709, opened 2026-04-18
[Fix] Isolate RNG seed generator changes and clarify documentation

ChiragSW:
Fixes part 3 of #22595 

## Contributor Agreement

### DIFF
--- a/keras/src/random/seed_generator.py
+++ b/keras/src/random/seed_generator.py
@@ -144,11 +144,9 @@ def make_default_seed():
     """Return the default integer seed used by random initializers."""
     from keras.src.utils import rng_utils
 
-    if rng_utils.get_random_seed() is not None:
-        # Keep implicit initializer seeds deterministic after
-        # `keras.utils.set_random_seed()` calls.
-        return rng_utils.consume_default_initializer_seed()
-    return int(np.random.randint(1, int(1e9)))
+    # Keep implicit initializer seeds deterministic after
+    # `keras.utils.set_random_seed()` calls.
+    return rng_utils.consume_default_initializer_seed()
 
 
 def draw_seed(seed):
--- a/keras/src/random/seed_generator.py
+++ b/keras/src/random/seed_generator.py
@@ -141,9 +141,12 @@ def global_seed_generator():
 
 
 def make_default_seed():
+    """Return the default integer seed used by random initializers."""
     from keras.src.utils import rng_utils
 
     if rng_utils.get_random_seed() is not None:
+        # Keep implicit initializer seeds deterministic after
+        # `keras.utils.set_random_seed()` calls.
         return rng_utils.consume_default_initializer_seed()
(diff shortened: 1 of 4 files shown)


ChiragSW:
@hertschuh please review

hertschuh (commented):
What is the context of this? What is the issue this is addressing?

ChiragSW:
> What is the context of this? What is the issue this is addressing?

Context: this PR was split out from #22595 after review feedback asked that the mixed PR be separated into smaller PRs. This is the “RNG seed generator changes” slice.

It is not part of the original #22472 symbolic validation bug. The purpose is a separate reproducibility cleanup: after `keras.utils.set_random_seed(seed)`, implicit seeds assigned to random initializers should come from Keras global seed state rather than from Python’s global `random` stream. This makes the sequence of default initializer seeds replayable after resetting the Keras seed, and avoids coupling initializer seed allocation to unrelated RNG draws.

- `set_random_seed(seed)` now initializes a separate default-initializer seed counter.
- `seed_generator.make_default_seed()` consumes that counter.
- unseeded behavior still falls back to a random integer.
- tests cover replaying the default seed sequence after `set_random_seed()` and avoiding NumPy RNG advancement.

ChiragSW:
@hertschuh any reviews for the changes made?

ChiragSW:
@hertschuh is there any further review for this PR?

hertschuh (commented):
@ChiragSW 

Sorry for the delay on this one. I've been thinking about this for a while.

In theory, it's nice to have Keras fully in control of the seed generation. Before I go in the details of the implementation, I do have a high level comment and question.

My main concern with this change is that it changes how the seeds are picked, and some users have come to depend on the specifics of the seeds and random numbers that are created.

Would it be possible to have an implementation that returns the same sequence of numbers as Python's `random.randint`? This way, by default, things would stay the same for users who use `keras.utils.set_random_seed`.

ChiragSW:
No problem, it was a topic to think upon. I'll give default initializer seeds their own `random.Random(seed)` instance and generate values with `randint(1, int(1e9))`. This keeps the sequence same as `random.randint` after `keras.utils.set_random_seed(seed)`. Unseeded inits will continue using the global RNG. I can update the implementation for the same if it sounds good.

hertschuh:
> No problem, it was a topic to think upon. I'll give default initializer seeds their own `random.Random(seed)` instance and generate values with `randint(1, int(1e9))`. This keeps the sequence same as `random.randint` after `keras.utils.set_random_seed(seed)`. Unseeded inits will continue using the global RNG. I can update the implementation for the same if it sounds good.

Ok, I lost track a bit here. This sounds good to me, but does it still address the replayability issue this is supposed to address?

ChiragSW:
Yes it still addresses the replayability, the only intentional behavior change is unrelated Python RNG draws no longer shift subsequent initializer seeds

PR #22710, opened 2026-04-18
fix: raise ValueError in deserialize_keras_object when required Keras config keys are missing

### LINKED ISSUE #22704: [Bug] deserialize_keras_object returns dict when top-level structure is incomplete
## Description

## Reproduction Script

```python
import os

# Missing class_name
config = {
    "module": "keras.layers",
    "config": {"units": 32}
}

print("Value:", result)
````

## Observed behavior

* Input is returned unchanged  

## Expected behavior

## Notes

AlgorithmicDeer:
Fixes #22704

## Description
`deserialize_keras_object` silently returned a raw `dict` when the input was missing `class_name` but contained Keras-specific keys like `module` or `registered_name`. This made it impossible for callers to detect a failed deserialization.
**Root cause:** The block meant to handle plain Python dicts also accidentally caught malformed Keras configs.
**Fix:** Added a guard — if the incomplete dict contains `module` or `registered_name` (keys that only appear in Keras-serialized objects), raise a descriptive `ValueError` listing exactly which keys are missing. Plain dicts without those keys continue to pass through unchanged, preserving existing behavior. This is consistent with the same validation already present in the `module_objects` code path.
**Tests added** (`serialization_lib_test.py`):
- `test_deserialize_missing_class_name_raises` — reproduces the exact bug
- `test_deserialize_missing_class_name_with_registered_name_raises` — same bug via `registered_name`
- `test_deserialize_plain_dict_no_keras_keys_passthrough` — regression guard for plain dicts

## Contributor Agreement

### DIFF
--- a/keras/src/saving/serialization_lib.py
+++ b/keras/src/saving/serialization_lib.py
@@ -611,14 +611,10 @@ def deserialize_keras_object(
 
     if "class_name" not in config or "config" not in config:
         if "module" in config or "registered_name" in config:
-            missing = [
-                k
-                for k in ("class_name", "config")
-                if k not in config
-            ]
+            missing = [k for k in ("class_name", "config") if k not in config]
             raise ValueError(
-                "A Keras config dict is missing required keys "
-                f"{missing}. Received: config={config}"
+                f"A Keras config dict is missing required keys {missing}. "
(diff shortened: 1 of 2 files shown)


hertschuh:

Issue #22701, opened 2026-04-18
[Bug]  deserialize_keras_object raises AttributeError for invalid config types


amadhan882:
## Description
`deserialize_keras_object` may raise `AttributeError` when provided with invalid config types (e.g., list or string), indicating that input validation occurs after internal method calls.

## Reproduction Script

```python
import os
os.environ["KERAS_BACKEND"] = "tensorflow"

from keras.saving import deserialize_keras_object

config = {
    "class_name": "Dense",
    "module": "keras.layers",
    "config": [1, 2, 3]   # invalid type
}

try:
    deserialize_keras_object(config)
except Exception as e:
    print("Exception type:", type(e).__name__)
    print("Message:", e)
````

## Observed behavior

* Raises `AttributeError: 'list' object has no attribute 'get'`

## Expected behavior

* Should raise a clear `ValueError` or `TypeError` for invalid input structure

## Notes

This suggests input validation happens too late and exposes internal implementation details.

dhantule:
Hi @amadhan882, 

I have tested this issue with Keras 3.14.0 and I was able to reproduce it, attaching a [gist](https://colab.sandbox.google.com/gist/dhantule/8d3a8be8c111ec03e0bb33a73901bfe7/-bug-deserialize_keras_object-raises-attributeerror-for-invalid-config-types-22701.ipynb) for reference.

Darshanjajoria18:
Hi, I’d like to work on this issue.

From my understanding, the problem is that `deserialize_keras_object` does not validate the type of `config` early enough, which leads to an AttributeError instead of a more informative error like ValueError or TypeError.

I noticed that there are already some PRs addressing this. Is there still any part of this issue that needs work, or any edge cases that are not yet covered?

I’d be happy to help with additional validation or tests if needed.

Jah-yee:
## Update

Hi! I've addressed the reviewer feedback from PR #22749 and created a new PR with a context-aware fix:

🔗 **New PR:** https://github.com/keras-team/keras/pull/22856

The new PR validates the `config` field based on `class_name`:
- `None` → treated as `{}` ✅
- `class_name="function"` → string config allowed ✅
- `class_name="__typespec__"` → list/tuple config allowed ✅  
- Other classes → must be `dict` ✅

This ensures invalid types still raise a clear `TypeError` (fixing the original issue) while preserving existing serialization patterns.

Let me know if you'd like any adjustments! 🙏

Hbakri144:
[...]

PR #22717, opened 2026-04-19
Fix deserialize_keras_object error message for invalid config types

Jah-yee:
Good day

## Summary

When `deserialize_keras_object` is called with a config dict where the `config` field is not a dict (e.g., a list or string), the code would raise a confusing `AttributeError: 'list' object has no attribute 'get'`. This made debugging difficult as the root cause was unclear.

## Fix

Added validation in `deserialize_keras_object` to check that when `class_name` is a class (not `function`), the `config` field must be a dict. If not, a clear `TypeError` is raised:

```
TypeError: Expected 'config' to be a dict, but got list: [1, 2, 3]
```

The fix is applied only for class deserialization. The case of `class_name == 'function'` is excluded because `inner_config` is a string (function name) in that case, which is valid.

## Reproduction

```python

config = {
    'class_name': 'Dense',
    'module': 'keras.layers',
    'config': [1, 2, 3]  # invalid type
}

# Before fix: AttributeError: 'list' object has no attribute 'get'
# After fix: TypeError: Expected 'config' to be a dict, but got list: [1, 2, 3]
deserialize_keras_object(config)
```

## Testing

- All 19 existing tests in `serialization_lib_test.py` pass
- The fix was verified against the exact reproduction case from issue #22701

Warmly,
RoomWithOutRoof

### DIFF
--- a/keras/src/saving/serialization_lib.py
+++ b/keras/src/saving/serialization_lib.py
@@ -618,7 +618,19 @@ def deserialize_keras_object(
         }
 
     class_name = config["class_name"]
-    inner_config = config["config"] or {}
+    inner_config = config["config"]
+    # Validate that inner_config is a dict when class_name is a class
+    # (not "function"), since from_config() expects a dict.
+    if (
+        inner_config is not None
+        and not isinstance(inner_config, dict)
+        and class_name != "function"
+    ):
+        raise TypeError(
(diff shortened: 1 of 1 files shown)


keerthanakadiri:

Aanushka001:
placeholder

Issue #22699, opened 2026-04-18
[Bug]  deserialize_keras_object accepts non-boolean values for trainable field


amadhan882:
## Description

`deserialize_keras_object` accepts non-boolean values for the `trainable`field without validation.

## Reproduction Script 

```python
from keras.saving import deserialize_keras_object

config = {
    "class_name": "Dense",
    "module": "keras.layers",
    "config": {
        "units": 32,
        "trainable": "yes"
    }
}

layer = deserialize_keras_object(config)
print(type(layer.trainable), layer.trainable)
````

## Observed behavior

* `trainable` is set to `"yes"` (string)

## Expected behavior

* Input should be validated as `bool`
* Invalid types should raise an error

## Notes

This may lead to inconsistent behavior in code that relies on strict boolean checks.

PR #22722, opened 2026-04-19
Fix: validate trainable field type in Layer.__init__

Jah-yee:
Good day

## Issue
deserialize_keras_object accepts non-boolean values for the `trainable` field without validation ([#22699](https://github.com/keras-team/keras/issues/22699)).

## Reproduction
```python
config = {
    "class_name": "Dense",
    "module": "keras.layers",
    "config": {
        "units": 32,
        "trainable": "yes"  # string instead of bool
    }
}
layer = deserialize_keras_object(config)
print(type(layer.trainable), layer.trainable)  # str, "yes"
```

## Fix
Added an explicit type check in `Layer.__init__` that raises a clear `TypeError` when `trainable` is not a `bool`:

```python
if not isinstance(trainable, bool):
    raise TypeError(
        f"Expected `trainable` to be a bool, but got "
        f"{type(trainable).__name__}: {trainable!r}"
    )
```

This ensures validation happens at layer construction time (during deserialization), catching the bad type early with a descriptive error.

## Testing
Verified that valid `bool` values (`True`/`False`) work correctly, while invalid values like `"yes"`, `1`, `0`, etc. now raise `TypeError` as expected.

Warmly,
RoomWithOutRoof

### DIFF
--- a/keras/src/layers/layer.py
+++ b/keras/src/layers/layer.py
@@ -297,6 +297,12 @@ class Layer(BackendLayer, Operation):
                 f"passed to {self.__class__.__name__}: {kwargs}"
             )
 
+        if not isinstance(trainable, bool):
+            raise TypeError(
+                f"Expected `trainable` to be a bool, but got "
+                f"{type(trainable).__name__}: {trainable!r}"
+            )
+
         self._path = None  # Will be determined in `build_wrapper`
         self.built = False
         self.autocast = autocast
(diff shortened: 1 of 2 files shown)


keerthanakadiri:

hertschuh:
This issue was addressed by https://github.com/keras-team/keras/pull/22740

PR #22723, opened 2026-04-19
Dynamically handle future scikit-learn metadata routing deprecation

andersendsa:
## Description
This PR resolves a code health TODO by future-proofing the scikit-learn metadata routing check in keras/src/wrappers/fixes.py. It dynamically falls back to checking sklearn.__version__ if the config key is removed, ensuring metadata routing correctly defaults to enabled in future scikit-learn versions while preserving backward compatibility.

## Contributor Agreement

### DIFF
--- a/keras/src/wrappers/fixes.py
+++ b/keras/src/wrappers/fixes.py
@@ -1,4 +1,5 @@
 from packaging.version import parse as parse_version
+
 try:
     import sklearn
 except ImportError:
@@ -60,6 +61,7 @@ def _routing_enabled():
         return True
 
     return False
+
 
 def _raise_for_params(params, owner, method):
     """Raise an error if metadata routing is not enabled and params are passed.
--- a/keras/src/wrappers/fixes.py
+++ b/keras/src/wrappers/fixes.py
@@ -1,3 +1,4 @@
+from packaging.version import parse as parse_version
 try:
     import sklearn
 except ImportError:
@@ -55,16 +56,10 @@ def _routing_enabled():
     if "enable_metadata_routing" in config:
         return config.get("enable_metadata_routing", False)
 
-    try:
-        from packaging.version import parse as parse_version
-
-        if parse_version(sklearn.__version__) >= parse_version("1.3"):
-            return True
-    except ImportError:
-        pass
+    if parse_version(sklearn.__version__) >= parse_version("1.3"):
+        return True
 
     return False
-
 
 def _raise_for_params(params, owner, method):
     """Raise an error if metadata routing is not enabled and params are passed.
--- a/keras/src/wrappers/fixes.py
+++ b/keras/src/wrappers/fixes.py
@@ -47,10 +47,23 @@ def _routing_enabled():
         enabled : bool
             Whether metadata routing is enabled. If the config is not set, it
             defaults to False.
-
-    TODO: remove when the config key is no longer available in scikit-learn
     """
-    return sklearn.get_config().get("enable_metadata_routing", False)
+    if sklearn is None:
+        return False
+
+    config = sklearn.get_config()
+    if "enable_metadata_routing" in config:
+        return config.get("enable_metadata_routing", False)
+
+    try:
+        from packaging.version import parse as parse_version
+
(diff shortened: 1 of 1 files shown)


andersendsa:
Hi @hertschuh  this pr has passes all the tests and is ready for review

keras/src/wrappers/fixes.py
```diff
@@ -47,10 +47,23 @@ def _routing_enabled():
...
+
+    config = sklearn.get_config()
+    if "enable_metadata_routing" in config:
+        return config.get("enable_metadata_routing", False)
+
+    try:
+        from packaging.version import parse as parse_version
+
+        if parse_version(sklearn.__version__) >= parse_version("1.3"):
+            return True
```
hertschuh: Why would it be on all the time for sklearn >= 1.3 ?

hertschuh (commented):
Thanks for looking into this!

andersendsa:
Hi @hertschuh i believe that meta data routing wasnt available until version 1.3

hertschuh:
> Hi @hertschuh i believe that meta data routing wasnt available until version 1.3

If that's the case, shouldn't the logic be this?

```python
    if parse_version(sklearn.__version__) < parse_version("1.3"):
        return False

    config = sklearn.get_config()
    if "enable_metadata_routing" in config:
        return config.get("enable_metadata_routing", False)

    return False
```

andersendsa:
> > Hi @hertschuh i believe that meta data routing wasnt available until version 1.3
> 
> If that's the case, shouldn't the logic be this?
> 
> ```python
>     if parse_version(sklearn.__version__) < parse_version("1.3"):
>         return False
> 
>     config = sklearn.get_config()
>     if "enable_metadata_routing" in config:
>         return config.get("enable_metadata_routing", False)
> 
>     return False
> ```

If we use that logic and a future version of scikit-learn (e.g., version 2.0) completely removes the enable_metadata_routing config key, your suggested logic would fall through to the final return False.

andersendsa:
> > > Hi @hertschuh i believe that meta data routing wasnt available until version 1.3
> > 
> > If that's the case, shouldn't the logic be this?
> > 
> > ```python
> >     if parse_version(sklearn.__version__) < parse_version("1.3"):
> >         return False
> > 
> >     config = sklearn.get_config()
> >     if "enable_metadata_routing" in config:
> >         return config.get("enable_metadata_routing", False)
> > 
> >     return False
> > ```
> 
> @hertschuh  If we use that logic and a future version of scikit-learn (e.g., version 2.0) completely removes the enable_metadata_routing config key, your suggested logic would fall through to the final return False.

keras/src/wrappers/fixes.py
```diff
@@ -47,10 +47,23 @@ def _routing_enabled():
...
+    if sklearn is None:
+        return False
+
+    config = sklearn.get_config()
+    if "enable_metadata_routing" in config:
+        return config.get("enable_metadata_routing", False)
+
+    try:
+        from packaging.version import parse as parse_version
+
```
hertschuh: Move import to the top and remove try / except

`packaging` is a required dependency of the pypi wheel.

hertschuh (changes requested):

hertschuh:
> > > > Hi @hertschuh i believe that meta data routing wasnt available until version 1.3
> > > 
> > > 
> > > If that's the case, shouldn't the logic be this?
> > > ```python
> > >     if parse_version(sklearn.__version__) < parse_version("1.3"):
> > >         return False
> > > 
> > >     config = sklearn.get_config()
> > >     if "enable_metadata_routing" in config:
> > >         return config.get("enable_metadata_routing", False)
> > > 
> > >     return False
> > > ```
> > 
> > 
> > @hertschuh  If we use that logic and a future version of scikit-learn (e.g., version 2.0) completely removes the enable_metadata_routing config key, your suggested logic would fall through to the final return False.

Oh, I see. I misunderstood what the deprecation was about.

andersendsa:
Hi @hertschuh  I have done the changes this pr is ready for review

hertschuh (approved):
Thanks!

PR #22725, opened 2026-04-19
[Fix] deserialize_keras_object accepts arbitrary registered_name values without validation

### LINKED ISSUE #22703: [Bug] deserialize_keras_object accepts arbitrary registered_name values without validation
## Description
The `registered_name` field in `deserialize_keras_object` is accepted without validation. Arbitrary string values, including unrelated module paths, are allowed and do not raise any error.

## Reproduction Script

```python
import os

# Case 1: arbitrary registered_name
config_1 = {
    "class_name": "Dense",
    "module": "keras.layers",
    "registered_name": "builtins.eval",
    "config": {"units": 1}
}

obj_1 = deserialize_keras_object(config_1)
print("Case 1 type:", type(obj_1))


# Case 2: another arbitrary value
config_2 = {
    "class_name": "Dense",
    "module": "keras.layers",
    "registered_name": "subprocess.Popen",
    "config": {"units": 1}
}

obj_2 = deserialize_keras_object(config_2)
print("Case 2 type:", type(obj_2))


# Case 3: empty string vs None
config_3 = {
    "class_name": "Dense",
    "module": "keras.layers",
    "registered_name": "",
    "config": {"units": 1}
}

config_4 = {
    "class_name": "Dense",
    "module": "keras.layers",
    "registered_name": None,
    "config": {"units": 1}
}

obj_3 = deserialize_keras_object(config_3)
obj_4 = deserialize_keras_object(config_4)

print("Case 3 type (empty string):", type(obj_3))
print("Case 4 type (None):", type(obj_4))
print("Same behavior (empty vs None):", type(obj_3) == type(obj_4))
````

## Observed behavior

* Arbitrary values for `registered_name` are accepted without validation
* Values such as `"builtins.eval"` and `"subprocess.Popen"` do not raise errors
* Empty string (`""`) and `None` are treated equivalently

## Expected behavior

* `registered_name` should be validated or restricted to known/registered values
* Invalid or unexpected values should raise an error or be rejected
* Behavior should be consistent and clearly defined

## Notes

The `registered_name` field currently has no effect on deserialization and is not validated. This may lead to confusion or inconsistent handling of configuration data.

ChiragSW:
Fixes: #22703 

## Root cause 
```deserialize_keras_object()``` accepted any registered_name string for class configs, but if that name did not resolve to a real registered custom object, Keras silently ignored it and fell back to module + class_name. That made values like "builtins.eval" look accepted even though they were meaningless.

## Fix
Added validation so class deserialization now rejects invalid or empty registered_name values instead of ignoring them. The change still preserves compatibility for valid cases like None, real registered custom-object keys, and the legacy/internal fallback where registered_name matches the class name.

## Contributor Agreement

### DIFF
--- a/keras/src/saving/serialization_lib.py
+++ b/keras/src/saving/serialization_lib.py
@@ -17,6 +17,7 @@ from keras.src.utils import python_utils
 from keras.src.utils.module_utils import tensorflow as tf
 
 PLAIN_TYPES = (str, int, float, bool)
+MISSING = object()
 
 # List of Keras modules with built-in string representations for Keras defaults
 BUILTIN_MODULES = frozenset(
@@ -717,6 +718,7 @@ def deserialize_keras_object(
     # Below: classes and functions.
     module = config.get("module", None)
     registered_name = config.get("registered_name", class_name)
+    validate_registered_name = config.get("registered_name", None)
 
     if class_name == "function":
         fn_name = inner_config
@@ -727,6 +729,7 @@ def deserialize_keras_object(
             obj_type="function",
             full_config=config,
             custom_objects=custom_objects,
+            validate_registered_name=validate_registered_name,
         )
 
     # Below, handling of all classes.
@@ -743,6 +746,7 @@ def deserialize_keras_object(
         obj_type="class",
         full_config=config,
         custom_objects=custom_objects,
+        validate_registered_name=validate_registered_name,
     )
 
     if isinstance(cls, types.FunctionType):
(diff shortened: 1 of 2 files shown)


ChiragSW:
@hertschuh please review

Saumay:
Heads-up: I opened #22729 as a parallel approach to this same issue. Happy to close mine if this one is preferred. The two main differences: mine keeps `''` and `None` passing through (the issue reports them as equivalent, so I left them equivalent), and uses a simpler inline check rather than a helper. Defer to the maintainers.

hertschuh (changes requested):
This is not the approach. The whole point about registered names is that 3rd parties can use any name. Registered names are not used for keras classes and functions.

The only verification to do is that `registered_name` is `None` when the class or function that was resolved is a keras_* class or function.

I.e. verification should only kick in when the `registered_name` is ignored.

ChiragSW:
The previous approach validated ```registered_name``` too early and too broadly, so I changed it to validate only after a Keras API class/function is actually resolved and ```registered_name``` is therefore ignored.

hertschuh:
Fixed by https://github.com/keras-team/keras/pull/22729

PR #22726, opened 2026-04-19
Fix deserialize_keras_object raises ValueError for missing class_name with module/config

Jah-yee:
Good day

## Summary

This PR fixes  to raise a  when the config dict contains  and/or  keys but is missing the required  key.

## Problem (Issue #22704)

When deserializing a config like:

The function was silently returning the dict instead of raising an error. This is problematic because:
1. It looks like a malformed Keras object config (has  and  but no )
2. Users expect a clear error message for invalid inputs
3. Silent failures make debugging difficult

## Solution

Added validation to detect when  is missing but  or  is present, which indicates a malformed Keras object config. In this case, a clear  is raised with the received config.

The fix still allows normal dicts without Keras-specific keys (like ) to be recursively deserialized, preserving backward compatibility.

## Testing

- All 19 existing tests in  pass
- Added manual verification of the issue case

Warmly,
RoomWithOutRoof

### DIFF
--- a/keras/src/saving/serialization_lib.py
+++ b/keras/src/saving/serialization_lib.py
@@ -618,7 +618,19 @@ def deserialize_keras_object(
         }
 
     class_name = config["class_name"]
-    inner_config = config["config"] or {}
+    inner_config = config["config"]
+    # Validate that inner_config is a dict when class_name is a class
+    # (not "function"), since from_config() expects a dict.
+    if (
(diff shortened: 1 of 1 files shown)


keerthanakadiri:
Hi @Jah-yee, Can you kindly sign the CLA? Thank you!

hertschuh:

PR #22727, opened 2026-04-19
[Bug] deserialize_keras_object raises internal TypeError for malformed layer list instead of validating input (fixes #22720)

### LINKED ISSUE #22720: [Bug]  deserialize_keras_object raises internal TypeError for malformed layer list instead of validating input
## Description

## Reproduction Script

```python
import os

import keras

payload = {
    "class_name": "Sequential",
    "module": "keras",
    "config": {
        "layers": [
            {
                "class_name": "Dense",
                "module": "keras.layers",
                "config": {"units": 4}
            },
            {},  # malformed layer
            {
                "class_name": "Dense",
                "module": "keras.layers",
                "config": {"units": 2}
            }
        ]
    }
}

````

## Observed behavior

* Raises:

## Expected behavior

## Impact

## Notes

### PR DESCRIPTION
**Pull request by Jah-yee**
Good day,

## Summary
This PR fixes issue #22720 where `deserialize_keras_object` raises an internal Python `TypeError` when a Sequential model contains malformed layer entries (e.g., `{}`) in the layers list, instead of raising a clear validation error.

## Problem
When deserializing a Sequential model with a malformed layer entry, the error propagated was:
```
TypeError: string indices must be integers, not 'str'
```

This error originates from internal logic rather than input validation, exposing implementation details to users.

## Solution
Added validation in `Sequential.from_config()` to check that `config` is a list when it does not contain the 'name' key. If a non-list config is passed (e.g., an empty dict `{}`), a clear `TypeError` is raised before any internal processing occurs.

## Change
In `keras/src/models/sequential.py`, added a type check in `from_config()`:
```python
if not isinstance(config, list):
    raise TypeError(
        f"Sequential.from_config` expects a list of layer configs "
        f"or a dict with 'name' and 'layers' keys. Received: "
        f"type(config)={type(config).__name__}."
    )
```

## Verification
- Added validation catches malformed layer entries early
- Valid Sequential models continue to serialize/deserialize correctly
- All 19 existing Sequential tests pass

Warmly,
RoomWithOutRoof


keerthanakadiri:

PR #22728, opened 2026-04-19
more informative error-handling of missing keys during deserialization of sequential model

### LINKED ISSUE #22720: [Bug]  deserialize_keras_object raises internal TypeError for malformed layer list instead of validating input

ShaunakDas88:
## Description
This PR fixes https://github.com/keras-team/keras/issues/22720.

It does the following during deserialization of a sequential model:
- requires either `name` and `layer` keys simultaneously be provided under `config`, or `config` to be a list of layer dicts
- raises a `ValueError` upon missing `config` key (e.g. the case of a malformed/empty layer dict originally raised in the linked issue), with a more descriptive error message (I'm fine with making this a `KeyError` if folks deem that better)

## Contributor Agreement

- [X] I am a human, and not a bot.
- [X] I will be responsible for responding to review comments in a timely manner.
- [X] I will work with the maintainers to push this PR forward until submission.

### DIFF
--- a/keras/src/legacy/saving/saving_utils.py
+++ b/keras/src/legacy/saving/saving_utils.py
@@ -54,9 +54,11 @@ def model_from_config(config, custom_objects=None):
         MODULE_OBJECTS.ALL_OBJECTS["Model"] = models.Model
         MODULE_OBJECTS.ALL_OBJECTS["Sequential"] = models.Sequential
 
-    if "config" not in config:
+    if not isinstance(config, dict) or "config" not in config:
         raise ValueError(
-            f"Missing `config` in model config. Received: config={config}."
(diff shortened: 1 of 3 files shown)


jeffcarp (approved):
Thank you!

PR #22732, opened 2026-04-20
Validate registered_name/class_name consistency in deserialize_keras_object

### LINKED ISSUE #22706: [Bug] deserialize_keras_object does not validate consistency between registered_name and class_name/module
## Description

`deserialize_keras_object` prioritizes `registered_name` over `class_name` and `module` during object resolution. When `registered_name` resolves successfully, the other fields are ignored.

While this behavior is documented, there is no validation to ensure consistency between `registered_name`, `class_name`, and `module`. As a result, a configuration may indicateone class but instantiate a different one without any warning or error.

## Reproduction Script

```python
import os

import keras
from keras.saving import deserialize_keras_object, register_keras_serializable

@register_keras_serializable(package="test", name="Dense")
class EvilLayer(keras.layers.Layer):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        print("EVIL INIT")

    def call(self, inputs):
        print("EVIL CALL")
        return inputs

config = {
    "class_name": "Dense",
    "module": "keras.layers",
    "registered_name": "test>Dense",
    "config": {}
}

obj = deserialize_keras_object(config)
print("Resolved type:", type(obj))

import tensorflow as tf
x = tf.ones((1, 4))
obj(x)
````

## Observed behavior

* `registered_name` overrides `class_name` and `module`
* A different class (`EvilLayer`) is instantiated
* No validation or warning is provided for conflicting metadata

## Expected behavior

* Consistency between `registered_name`, `class_name`, and `module` should be validated
* Mismatches should raise an error or at least emit a warning

## Notes

The documentation states that `registered_name` takes precedence when it resolves. However, this behavior does not validate consistency with `class_name` or `module`.

This can lead to configurations that appear to describe one class while instantiating another, which may be confusing or lead to unintended behavior when configuration sources are not strictly controlled.

### PR DESCRIPTION
**Pull request by rstar327**
## Description

Fixes #22706.

`deserialize_keras_object` prioritizes `registered_name` over `class_name`
and `module` when resolving an object from a config dict, and does not
validate that the resolved class actually matches the `class_name` / `module`
that the config claims to describe. As reported in #22706, this can lead to
configurations that appear to describe one class while instantiating another:

```python
@register_keras_serializable(package="test", name="Dense")
class EvilLayer(keras.layers.Layer): ...

deserialize_keras_object({
    "class_name": "Dense",
    "module": "keras.layers",
    "registered_name": "test>Dense",
    "config": {},
})  # silently returns an EvilLayer
```

A correctly serialized config always has `class_name` equal to the resolved
class's `__name__` (see `serialize_with_public_class`), so this PR adds a
validation step in `_retrieve_class_or_fn`: when a class is resolved via
`registered_name`, we now check that `class_name` matches the resolved class's
`__name__` and raise a `ValueError` with the full config if not.

A regression test (`test_inconsistent_registered_name_raises`) is added. All
existing saving tests continue to pass.

## Contributor Agreement

### DIFF
--- a/keras/src/saving/serialization_lib.py
+++ b/keras/src/saving/serialization_lib.py
@@ -774,6 +774,25 @@ def _retrieve_class_or_fn(
             registered_name, custom_objects=custom_objects
         )
     if custom_obj is not None:
+        # Validate consistency between `registered_name` and `class_name`.
+        # A correctly serialized config always has `class_name` equal to the
+        # resolved class's `__name__` (see `serialize_with_public_class`).
+        # A mismatch indicates a tampered or malformed config where
+        # `registered_name` resolves to a different object than what
+        # `class_name` / `module` claim to describe.
+        if (
+            obj_type == "class"
+            and name is not None
+            and getattr(custom_obj, "__name__", None) is not None
+            and custom_obj.__name__ != name
+        ):
+            raise ValueError(
+                f"Inconsistent deserialization config: `registered_name` "
+                f"'{registered_name}' resolves to class "
+                f"'{custom_obj.__name__}', but the config specifies "
+                f"`class_name='{name}'`. These values must be consistent. "
+                f"Full object config: {full_config}"
+            )
         return custom_obj
 
     if module:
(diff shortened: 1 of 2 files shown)


hertschuh:
Unfortunately, we just cannot do this. This is a breaking change. Many existing saved models won't reload if we do this.

By design, the registered name can be different from the class name. Many teams rely on this feature, sometimes to give shorter / longer name, sometimes to maintain backwards compatibility after the class is renamed. Such a breaking change is just not possible.

PR #22737, opened 2026-04-21
Support dict/list inputs in torch ONNX export

MarcosAsh:
## Description

`export_onnx` on the torch backend previously raised when given a model whose `input_signature` was a dict (and the test harness skipped the dict and tuple cases outright). This flattens any nested `input_signature` into positional tensors and, when the structure is actually nested, wraps the model in a small `torch.nn.Module` that repacks the flat args back into the original structure before calling the underlying model. The wrapped model is what gets passed to `torch.onnx.export`, so the exported graph sees a flat tuple of inputs that `onnxruntime` can feed by name.

Input names are derived from the flattened specs (falling back to `input_{i}`), and dynamic-axis names stay unique per flat input. The `convert_spec_to_tensor` call is unchanged; only the post-processing of the resulting sample tensors differs when the structure is nested.

The `test_model_with_input_structure` parametrization no longer skips the `dict` and `array` variants on torch. The `tuple` variant is still skipped since tuple-of-tensors input handling is a separate concern outside this fix.

## Contributor Agreement

### DIFF
--- a/keras/src/export/onnx.py
+++ b/keras/src/export/onnx.py
@@ -160,7 +160,7 @@ def export_onnx(
 
                 def forward(self, *flat_args):
                     inputs = tree.pack_sequence_as(
-                        self._structure, list(flat_args)
+                        self._structure, flat_args
                     )
                     if len(inputs) == 1:
                         return self._wrapped(inputs[0])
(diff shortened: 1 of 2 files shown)


MarcosAsh:
I amended the commit with what gemini stated it seemed valid.

hertschuh (approved):
Thanks for adding this!

PR #22738, opened 2026-04-21
Fix stale return_attention_scores flag in Attention.compute_output_spec

MarcosAsh:
## Description
`Attention.compute_output_spec` reads `self._return_attention_scores or return_attention_scores`. The cached attribute is set from the first `call()` and never reset, so once a layer has been called with `return_attention_scores=True`, every later symbolic call reports a tuple output even when the caller passes `False`. The runtime `call()` only honors the per-call argument, so spec and output disagree.
This drops the cached-flag fallback in `compute_output_spec` so the spec matches `call()`. A regression test covers the True-then-False order. `compute_output_shape` is unchanged since it has no per-call signal.
## Contributor Agreement

### DIFF
--- a/keras/src/layers/attention/attention.py
+++ b/keras/src/layers/attention/attention.py
@@ -268,29 +268,23 @@ class Attention(Layer):
         training=None,
         use_causal_mask=False,
     ):
-        # Validate and unpack inputs
         self._validate_inputs(inputs, mask)
         query = inputs[0]
         value = inputs[1]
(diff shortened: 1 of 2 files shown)


MarcosAsh:
Addressed all the fixes from gemini

MarcosAsh:
I think the openvino failure is just a flaky test.

hertschuh (approved):
Thanks for the cleanup!

PR #22739, opened 2026-04-21
Fix stale current_shard_path in ShardedH5IOStore get

AdonaiVera:
## Description
While fine-tuning Gemma 4 Instruct 26B on a TPU v5litepod-8 (see https://github.com/keras-team/kinetic/pull/204), I found a silent weight loading issue. The model trained fine, but when loading weights back, some layers were empty. No error, no warning, just wrong values. After debugging, the issue was in Keras, not in the training setup.

What I found is that `ShardedH5IOStore.get()` switches H5 files when weights are in a different shard, but it does not update `current_shard_path`. After the first switch, `current_shard_path` becomes outdated. This affects large models that span multiple shards, like Gemma 4 26B.

## Contributor Agreement

- [X] I am a human, and not a bot.
- [X] I will be responsible for responding to review comments in a timely manner.
- [X] I will work with the maintainers to push this PR forward until submission.

### DIFF
--- a/keras/src/saving/saving_lib.py
+++ b/keras/src/saving/saving_lib.py
@@ -1320,7 +1320,8 @@ class ShardedH5IOStore(H5IOStore):
 
         if filename is not None and filename != self.current_shard_path.name:
             self.close()
-            self.h5_file = self._get_h5_file(self.path.with_name(filename))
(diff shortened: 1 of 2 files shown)


jeffcarp (approved):
Thank you for catching and fixing this!

PR #22740, opened 2026-04-21
Validate trainable arg as bool in Layer.__init__

### LINKED ISSUE #22699: [Bug]  deserialize_keras_object accepts non-boolean values for trainable field

bzantium:
## Description

Fixes #22699.

`deserialize_keras_object` (and direct construction) silently accepted non-boolean values for the `trainable` argument, leaving `layer.trainable` set to a non-bool and breaking downstream strict boolean checks. For example, a malformed config with `"trainable": "yes"` would make `layer.trainable == "yes"` (str).

This change raises `ValueError` at construction time when `trainable` is not a `bool`, catching malformed configs early at the input boundary (per the Keras style guide, which prescribes `ValueError` for user input validation).

## Contributor Agreement

### DIFF
--- a/keras/src/layers/layer.py
+++ b/keras/src/layers/layer.py
@@ -305,9 +305,9 @@ class Layer(BackendLayer, Operation):
         self.supports_jit = True
 
         if not isinstance(trainable, bool):
-            raise TypeError(
+            raise ValueError(
                 "Expected `trainable` to be a boolean. "
-                f"Received: trainable={trainable} (of type "
+                f"Received: trainable={trainable!r} (of type "
                 f"{type(trainable).__name__})"
             )
         self._trainable = trainable
--- a/keras/src/layers/layer.py
+++ b/keras/src/layers/layer.py
@@ -304,6 +304,12 @@ class Layer(BackendLayer, Operation):
         self._called = False
         self.supports_jit = True
 
+        if not isinstance(trainable, bool):
+            raise TypeError(
+                "Expected `trainable` to be a boolean. "
+                f"Received: trainable={trainable} (of type "
+                f"{type(trainable).__name__})"
+            )
         self._trainable = trainable
         self._losses = []
         self._loss_ids = set()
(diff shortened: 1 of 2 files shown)


keras/src/layers/layer.py
```diff
@@ -304,6 +304,12 @@ def __init__(
         self._called = False
         self.supports_jit = True
 
+        if not isinstance(trainable, bool):
+            raise ValueError(
+                "Expected `trainable` to be a boolean. "
+                f"Received: trainable={trainable!r} (of type "
+                f"{type(trainable).__name__})"
```
hertschuh: Small nitpick for consistency with the rest of the code:

just do:
- `{trainable}` instead of `{trainable!r}`
- `{type(trainable)}` instead of `{type(trainable).__name__}`

keras/src/layers/layer_test.py
```diff
@@ -1378,6 +1378,18 @@ def custom_change_dtype(self):
         self.assertEqual(layer.w.dtype, "int8")
         self.assertEqual(layer.w.trainable, False)
 
+    def test_trainable_init_arg_validation(self):
+        # Issue #22699: non-boolean `trainable` values (e.g. from a malformed
+        # config passed to `deserialize_keras_object`) were silently accepted,
+        # leaving `layer.trainable` as a non-bool and breaking downstream
+        # strict-boolean checks.
```
hertschuh: Remove this comment. The context of why this test was added can be found using Blame.

hertschuh (changes requested):
Thanks for working on this!

hertschuh (approved):
Thanks!

PR #22741, opened 2026-04-21
Validate inner config dict type in deserialize_keras_object

### LINKED ISSUE #22701: [Bug]  deserialize_keras_object raises AttributeError for invalid config types

bzantium:
## Description

Fixes #22701.

When the `config` field of an object dict was not a dict (e.g. a list passed by mistake or by a malformed config), `deserialize_keras_object` surfaced the failure as an internal `AttributeError: 'list' object has no attribute 'get'` raised from inside `cls.from_config(...)`.

This change validates `config["config"]` at the input boundary and raises a clear `ValueError` that includes the bad type, value, and full config, so malformed inputs are rejected at the edge rather than deep in class-specific code. The check inspects the raw value in `config["config"]` rather than the `inner_config = config["config"] or {}` local, so falsy non-dicts (e.g. `[]`, `0`) can't sneak through by being coerced into an empty dict.

## Contributor Agreement

### DIFF
--- a/keras/src/saving/serialization_lib.py
+++ b/keras/src/saving/serialization_lib.py
@@ -728,6 +728,16 @@ def deserialize_keras_object(
             f"Full object config: {config}"
         )
 
+    # Check the raw value: `inner_config = config["config"] or {}` would
+    # hide falsy non-dicts (e.g. `[]`) by coercing them to `{}`.
+    raw_inner_config = config["config"]
+    if raw_inner_config is not None and not isinstance(raw_inner_config, dict):
+        raise ValueError(
+            f"Cannot deserialize '{class_name}': expected the `config` field "
+            f"to be a dict, got {type(raw_inner_config)} "
+            f"(value: {raw_inner_config}). Full object config: {config}"
+        )
+
     # Instantiate the class from its config inside a custom object scope
     # so that we can catch any custom objects that the config refers to.
     custom_obj_scope = object_registration.CustomObjectScope(custom_objects)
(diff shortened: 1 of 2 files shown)


keras/src/saving/serialization_lib.py
```diff
@@ -728,6 +728,16 @@ class ModifiedMeanSquaredError(keras.losses.MeanSquaredError):
...
 
+    # Check the raw value: `inner_config = config["config"] or {}` would
+    # hide falsy non-dicts (e.g. `[]`) by coercing them to `{}`.
+    raw_inner_config = config["config"]
+    if raw_inner_config is not None and not isinstance(raw_inner_config, dict):
+        raise ValueError(
+            f"Cannot deserialize '{class_name}': expected the `config` field "
+            f"to be a dict, got {type(raw_inner_config)} "
+            f"(value: {raw_inner_config}). Full object config: {config}"
+        )
```
hertschuh, line 739: The issue with this is that it misses a lot of other cases between lines 624 and 689 where `inner_config` is used as a dict.

The only case when `inner_config` is allowed to be something else (a string) is line 695.

I would:
- move that block first
- then validate `inner_config` is a dict
- then do the other if / else blocks.

Also, I would remove the `or {}` part of `inner_config = config["config"] or {}`. I don't think that's ever allowed. Empty config should be an empty dict.

hertschuh (changes requested):

PR #22742, opened 2026-04-21
Reject malformed object configs in deserialize_keras_object

### LINKED ISSUE #22704: [Bug] deserialize_keras_object returns dict when top-level structure is incomplete

bzantium:
## Description

Fixes #22704.

If the input dict was missing either `class_name` or `config`, `deserialize_keras_object` silently returned the raw dict, so callers had to add defensive type checks to distinguish a real deserialized object from a pass-through dict. For example, `{"module": "keras.layers", "config": {"units": 32}}` was returned unchanged instead of being flagged as a malformed object config.

This change tightens the fallback branch: when the input dict carries any object-config-only key (`class_name`, `module`, or `registered_name`) but is still missing one of the required fields (`class_name`/`config`), raise `ValueError` listing the missing keys. Plain user dicts that don't contain any of those markers still pass through unchanged so existing behavior is preserved.

## Contributor Agreement

### DIFF
--- a/keras/src/saving/serialization_lib.py
+++ b/keras/src/saving/serialization_lib.py
@@ -610,6 +610,18 @@ def deserialize_keras_object(
         raise TypeError(f"Could not parse config: {config}")
 
     if "class_name" not in config or "config" not in config:
+        # If the dict looks like an object config (has any of these markers)
+        # but is missing a required field, treat it as malformed rather than
+        # a plain user dict.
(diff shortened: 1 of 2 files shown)


hertschuh:

PR #22743, opened 2026-04-21
Validate registered_name field in deserialize_keras_object

### LINKED ISSUE #22703: [Bug] deserialize_keras_object accepts arbitrary registered_name values without validation

bzantium:
## Description

Fixes #22703.

The `registered_name` field of a serialized object config was accepted without validation: arbitrary strings like `"builtins.eval"` or `"subprocess.Popen"` passed through silently, non-string types were accepted, and empty string vs. `None` were treated equivalently without being documented as such. The values were not actually dangerous (the existing `_retrieve_class_or_fn` lookup only resolves entries from the Keras registry / `custom_objects`, so unknown names fall through), but the field gave no feedback at all when it was meaningless.

This change adds a small validation step when reading the field:

- reject non-string, non-`None` values with `ValueError` (per the Keras style guide, user input validation uses `ValueError`);
- normalize empty string to `None` so the two behave the same;
- warn when an explicitly provided non-empty value neither resolves via `custom_objects` / `GLOBAL_CUSTOM_OBJECTS` / the active custom object scope nor equals `class_name` (the legacy default). In that case the field has no effect on deserialization, so the warning tells the user their value is being ignored and points at the expected `package>name` format.

The documented patterns (None for built-ins, `package>name` for registered classes, `registered_name == class_name` for legacy configs) continue to load without warnings, so existing saved models are unaffected.

## Contributor Agreement

### DIFF
--- a/keras/src/saving/serialization_lib.py
+++ b/keras/src/saving/serialization_lib.py
@@ -690,7 +690,9 @@ def deserialize_keras_object(
 
     # Below: classes and functions.
     module = config.get("module", None)
-    registered_name = config.get("registered_name", class_name)
+    registered_name = _get_validated_registered_name(
+        config, class_name, custom_objects
+    )
 
     if class_name == "function":
         fn_name = inner_config
@@ -758,6 +760,38 @@ def deserialize_keras_object(
             instance, config["shared_object_id"]
         )
     return instance
+
+
+def _get_validated_registered_name(config, class_name, custom_objects):
+    """Return the `registered_name` field after type/value validation."""
+    if "registered_name" not in config:
+        return class_name
+    registered_name = config["registered_name"]
+    if registered_name is not None and not isinstance(registered_name, str):
+        raise ValueError(
+            "`registered_name` must be None or a string, got "
+            f"{type(registered_name)} "
+            f"(value: {registered_name}). Full config: {config}"
+        )
+    if registered_name == "":
+        registered_name = None
+    if (
+        registered_name is not None
+        and registered_name != class_name
+        and object_registration.get_registered_object(
+            registered_name, custom_objects=custom_objects
+        )
+        is None
+    ):
+        warnings.warn(
+            f"Ignoring `registered_name={registered_name}`: it is not "
+            "registered via `keras.saving.register_keras_serializable` and "
+            "was not found in `custom_objects`. The object will be "
+            "resolved from `class_name` and `module` instead. Valid "
+            "registered names have the format 'package>name'.",
+            stacklevel=3,
+        )
+    return registered_name
 
 
 def _retrieve_class_or_fn(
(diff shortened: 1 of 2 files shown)


bzantium:
Re: the suggestion to also validate `registered_name` in the `module_objects` branch (lines 545-605):

I left that out deliberately — that path is entered internally by `keras.activations.get` and friends, and before recursing it repackages the config via `serialize_with_public_class` / `serialize_with_public_fn`, which produces a fresh `registered_name` (either `None` or the object's canonical registry name). The user's original `registered_name` is therefore discarded rather than acted on, so the silent-ignore concern from #22703 doesn't apply there in the same way.

Adding the same warning to that branch risks false-positive warnings for the internal callers (activations, etc.) and expanded scope beyond the reported issue. Happy to add it if a maintainer prefers, but I'd rather keep this PR narrow.

hertschuh (changes requested):
This is not the approach. The whole point about registered names is that 3rd parties can use any name, and it doesn't have to match the class name, this is on purpose. Registered names are not used for keras classes and functions.

The only verification to do is that `registered_name` is `None` when the class or function that was resolved is a keras_* class or function.

I.e. verification should only kick in when the `registered_name` is ignored.

hertschuh:
Fixed by https://github.com/keras-team/keras/pull/22729

PR #22744, opened 2026-04-21
Validate layer entries in Sequential.from_config

### LINKED ISSUE #22720: [Bug]  deserialize_keras_object raises internal TypeError for malformed layer list instead of validating input

bzantium:
## Description

Fixes #22720.

A malformed entry inside the `layers` list (for example, an empty dict) used to surface as a cryptic internal `TypeError: string indices must be integers, not 'str'` raised from deep inside the legacy loader (`saving_utils.model_from_config`). The malformed entry was never rejected at the input boundary, so the failure leaked implementation details to users loading external model configs.

Validate each entry at the start of the loop before dispatching to either the legacy or the modern deserialization path: the entry must be a dict and must contain both `class_name` and `config` (both paths require them to reconstruct a layer). Otherwise raise `ValueError` identifying the offending index, value, and type, so callers can pinpoint and fix bad inputs.

## Contributor Agreement

### DIFF
--- a/keras/src/models/sequential.py
+++ b/keras/src/models/sequential.py
@@ -378,7 +378,18 @@ class Sequential(Model):
             name = None
             layer_configs = config
         model = cls(name=name)
-        for layer_config in layer_configs:
+        for idx, layer_config in enumerate(layer_configs):
+            if (
+                not isinstance(layer_config, dict)
+                or "class_name" not in layer_config
+                or "config" not in layer_config
+            ):
+                raise ValueError(
+                    f"Cannot deserialize Sequential model: entry at index "
+                    f"{idx} of `layers` is not a valid layer config. Expected "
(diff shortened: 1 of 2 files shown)


jeffcarp:
Already fixed by #22728 - thanks for the contribution!

PR #22747, opened 2026-04-21
[OpenVINO] Fix zeros/ones/empty/full for dynamic shapes

goyaladitya05:
## Description
ops.zeros, ops.ones, ops.empty, and ops.full previously built the output shape by calling ov_opset.constant(shape, Type.i32) directly. This crashes when any element of the shape is an OpenVINOKerasTensor scalar, returned by ops.shape(x) on a tensor with dynamic dimensions, such as one produced during graph construction with a dynamic batch axis.

Fix: All four ops now call shape_to_ov_output() (introduced in #22692), which handles mixed static/dynamic shape tuples by building per-dim constants and concatenating them into a single i32 shape tensor. A guard is also added to shape_to_ov_output to raise a clear error if None appears in the shape tuple, directing callers to use ops.shape(x) rather than x.shape to obtain a runtime-resolved shape.

Note: This PR does not revert #22681. The _parameterize_data dynamic batch workaround in the trainer is left in place; fixing that properly will be tracked separately.

## Contributor Agreement

### DIFF
--- a/keras/src/backend/openvino/excluded_concrete_tests.txt
+++ b/keras/src/backend/openvino/excluded_concrete_tests.txt
@@ -32,5 +32,4 @@ NumpyOneInputOpsCorrectnessTest::test_real
 RandAugmentTest::test_random_augment_randomness
 RandomSaturationTest::test_random_saturation_full_saturation
 RandomZoomTest::test_connect_with_flatten
-RandomZoomTest::test_dynamic_shape
 UpSampling2dTest::test_upsampling_2d_lanczos_interpolation_methods
--- a/keras/src/backend/openvino/core.py
+++ b/keras/src/backend/openvino/core.py
@@ -156,9 +156,15 @@ def shape_to_ov_output(shape):
 
     Unlike get_ov_output, handles mixed shapes where some dims are
     OpenVINOKerasTensor scalars (from ops.shape() on dynamic tensors).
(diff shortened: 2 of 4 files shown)


hertschuh (dismissed):
Thanks for the fix!

hertschuh (commented):
Wait, but did you test this with KerasHub?

I don't think this will work. What KerasHub was doing is calling `ops.ones` with the shape of the result of `_parameterize_data`. The -1 in there still won't work.

goyaladitya05:
> Wait, but did you test this with KerasHub?
> 
> I don't think this will work. What KerasHub was doing is calling `ops.ones` with the shape of the result of `_parameterize_data`. The -1 in there still won't work.

Addressing this further in keras-team/keras-hub/pull/2700.

keras/src/backend/openvino/excluded_concrete_tests.txt
```diff
@@ -32,5 +32,4 @@ NumpyOneInputOpsCorrectnessTest::test_real
 RandAugmentTest::test_random_augment_randomness
 RandomSaturationTest::test_random_saturation_full_saturation
 RandomZoomTest::test_connect_with_flatten
-RandomZoomTest::test_dynamic_shape
```
goyaladitya05, line 32: this was already passing.

hertschuh (approved):
Thanks!

PR #22748, opened 2026-04-22
Fix: Issue #22538 Remove symbolic tensor support for blackman operation

gaga1313:
## Description
Removed support for symbolic tensors, as the Blackman function depends on the tensor value. This causes runtime errors in JAX and Keras functional graphs.

Restricted the blackman function input to Python integers.

## Contributor Agreement

### DIFF
--- a/keras/src/ops/numpy.py
+++ b/keras/src/ops/numpy.py
@@ -1859,33 +1859,27 @@ def right_shift(x, y):
     return backend.numpy.right_shift(x, y)
 
 
-class Blackman(Operation):
-    def call(self, x):
-        return backend.numpy.blackman(x)
-
-    def compute_output_spec(self, x):
-        return KerasTensor(x.shape, dtype=backend.floatx())
-
-
 @keras_export(["keras.ops.blackman", "keras.ops.numpy.blackman"])
 def blackman(x):
     """Blackman window function.
     The Blackman window is a taper formed by using a weighted cosine.
 
     Args:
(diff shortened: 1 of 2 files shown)


gaga1313:
Also updated the doc-string with correct output values.

keras/src/ops/numpy.py
```diff
@@ -1859,33 +1859,27 @@ def right_shift(x, y):
...
-    array([-1.3877788e-17,  3.4000000e-01,  1.0000000e+00,  3.4000000e-01,
-           -1.3877788e-17], dtype=float32)
+    array([-1.4901161e-08,  3.4000003e-01,  9.9999994e-01,  3.3999997e-01,
+       -1.4901161e-08], dtype=float32)
     """
     if any_symbolic_tensors((x,)):
-        return Blackman().symbolic_call(x)
+        raise TypeError(
+            f"""Blackman operation only supports scalar inputs and
+            non-eager tensors. Received input x = {x} of type {type(x)}"""
```
hertschuh: It's quite difficult to explain what's supported (scalar eager tensors are always supported, non-eager tensors might be supported on TF).

Anyway, let's just say what the code does: `Blackman operation does not support symbolic tensors. Received ...`.

One more details, we don't use triple double-quotes in this context, just multiple lines:
```python
raise TypeError(
  "..."
  f"... {x}"
)
```

hertschuh (changes requested):

hertschuh (approved):
Thanks for the fix!

PR #22749, opened 2026-04-22
Fix: validate config is dict in deserialize_keras_object (good first issue)

Jah-yee:
Good day

## Summary

This PR fixes issue #22701 by adding validation to ensure the  field in  is a dictionary, rather than letting invalid types propagate internal AttributeErrors.

## Fix

**Before**: Passing a non-dict  (e.g., a list) resulted in:

**After**: A clear, informative error is raised:

## Changes

- Added type check for  field in 
- Raises  with a descriptive message when  is not a dict
- Fix is minimal and targeted — only affects the error path that was previously leaking internal implementation details

## Testing

The fix correctly handles the reproduction case from the issue:

---

Warmly,
Jah-yee

### DIFF
--- a/keras/src/saving/serialization_lib.py
+++ b/keras/src/saving/serialization_lib.py
@@ -618,7 +618,14 @@ def deserialize_keras_object(
         }
 
     class_name = config["class_name"]
-    inner_config = config["config"] or {}
+    inner_config = config["config"]
+    if not isinstance(inner_config, dict):
+        raise TypeError(
+            f"Expected 'config' to be a dict, got {type(inner_config).__name__}. "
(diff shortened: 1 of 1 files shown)


Jah-yee:
## New PR

Hi! I've addressed the reviewer feedback and created a new PR:

🔗 **https://github.com/keras-team/keras/pull/22856**

The new PR implements context-aware validation that:
1. Allows `None` config (becomes `{}`)
2. Allows string config for `class_name="function"`
3. Allows list/tuple config for `class_name="__typespec__"`
4. Raises `TypeError` with clear message for unexpected types

Would you mind taking a look? 🙏

PR #22753, opened 2026-04-22
Fix deserialize_keras_object raises ValueError for incomplete top-level structure (fixes #22704)

### LINKED ISSUE #22704: [Bug] deserialize_keras_object returns dict when top-level structure is incomplete

Jah-yee:
## Contributor Agreement

---

Good day

Fixes #22704.

## Description

The previous code raised a ValueError when BOTH 'class_name' and 'config' were missing. However, this broke recursive dictionary deserialization where a dict may contain multiple Keras configs (e.g., {'layer1': {...}, 'layer2': {...}}).

This fix:
- Raises ValueError when 'class_name' is missing (covers #22704)
- Preserves recursive dict deserialization when 'config' is missing but 'class_name' is present
- Removes the inner_config type validation that was breaking tf.TypeSpec

## Changes

- Separate the combined condition: raise ValueError only when 'class_name' is missing
- When 'config' is missing but 'class_name' is present, preserve recursive dict deserialization
- Removes unnecessary inner_config type check

## Before/After

Before:  returns the dict unchanged
After: Same input raises ValueError

## Addresses Review Comments

Addresses feedback from @gemini-code-assist[bot]:
- Critical: Preserves recursive dict deserialization (lists/dicts with Keras configs inside)
- High: Removes inner_config type check that failed on tf.TypeSpec (non-dict configs)

## Testing

Verified the fix with the reproduction script from #22704.

Thank you for your attention.

Warmly,
Jah-yee

### DIFF
--- a/keras/src/saving/serialization_lib.py
+++ b/keras/src/saving/serialization_lib.py
@@ -609,7 +609,12 @@ def deserialize_keras_object(
     if not isinstance(config, dict):
         raise TypeError(f"Could not parse config: {config}")
 
-    if "class_name" not in config or "config" not in config:
+    if "class_name" not in config:
+        raise ValueError(
+            f"Missing `class_name` in the config. Received: {config}"
+        )
+
+    if "config" not in config:
         return {
             key: deserialize_keras_object(
                 value, custom_objects=custom_objects, safe_mode=safe_mode


keras/src/saving/serialization_lib.py
```diff
@@ -609,7 +609,12 @@ class ModifiedMeanSquaredError(keras.losses.MeanSquaredError):
...
-    if "class_name" not in config or "config" not in config:
+    if "class_name" not in config:
+        raise ValueError(
+            f"Missing `class_name` in the config. Received: {config}"
+        )
+
+    if "config" not in config:
         return {
             key: deserialize_keras_object(
                 value, custom_objects=custom_objects, safe_mode=safe_mode
```
Jah-yee, line 615: Hi @gemini-code-assist[bot] — thank you for the thorough review comments on v1.

v2 (commit 7fa5303fc) has addressed both issues:

1. **Critical (recursive dict deserialization)**: Fixed by splitting the combined  condition. Now:
   - Raises  when 'class_name' is missing (covers the original bug #22704)
   - When 'config' is missing but 'class_name' is present, recursively deserializes dict values (preserving the original behavior for nested configs)

2. **High (tf.TypeSpec / inner_config type check)**: The unnecessary  type validation has been removed. The function now relies on the existing  check at the top.

The new behavior:
-  → raises  ✅
-  → recursively deserializes ✅
-  and other non-dict  cases → handled by existing code path ✅

Would you mind taking another look? Thank you!

keerthanakadiri:
Hi @Jah-yee , Can you please sign the CLA? Thank you!

hertschuh:

PR #22763, opened 2026-04-23
Add keras.ops.pinv

MarcosAsh:
## Description

Adds `keras.ops.pinv` (and `keras.ops.linalg.pinv`), the Moore-Penrose pseudoinverse. Fills the gap in `keras.ops.linalg` alongside `inv`, `solve`, and `lstsq`.

Each trainable backend delegates to its native implementation (`tf.linalg.pinv`, `jnp.linalg.pinv`, `torch.linalg.pinv`, `np.linalg.pinv`). `rcond` maps to torch's `rtol` (same semantics: relative threshold vs. the largest singular value). OpenVINO raises `NotImplementedError` and the correctness tests are added to its exclusion list, matching the existing `lstsq` pattern.

Tests cover tall / wide / square / batched matrices, symbolic shape inference, and the Moore-Penrose identity `A @ pinv(A) @ A == A`. All 7 new tests pass locally on numpy, jax, torch, and tensorflow backends.

## Contributor Agreement


keras/src/backend/numpy/linalg.py
```diff
@@ -98,5 +98,12 @@ def lstsq(a, b, rcond=None):
     return np.linalg.lstsq(a, b, rcond=rcond)[0]
 
 
+def pinv(x, rcond=None):
+    x = convert_to_tensor(x)
+    if rcond is None:
+        return np.linalg.pinv(x)
+    return np.linalg.pinv(x, rcond=rcond)
```
hertschuh: Why would you need this `if`, the default for `rcond` is None?

keras/src/backend/tensorflow/linalg.py
```diff
@@ -246,6 +246,13 @@ def lstsq(a, b, rcond=None):
     return x
 
 
+def pinv(x, rcond=None):
+    x = convert_to_tensor(x)
+    if rcond is None:
+        return tf.linalg.pinv(x)
+    return tf.linalg.pinv(x, rcond=rcond)
```
hertschuh: Why would you need this `if`, the default for `rcond` is None?

keras/src/backend/torch/linalg.py
```diff
@@ -82,5 +82,16 @@ def lstsq(a, b, rcond=None):
...
 
+def pinv(x, rcond=None):
+    x = convert_to_tensor(x)
+    if rcond is None:
+        return torch.linalg.pinv(x)
+    # `torch.linalg.pinv` expresses the threshold as `rtol`
+    # (relative tolerance), which has the same meaning as numpy's `rcond`:
+    # a singular value is treated as zero when it's smaller than
+    # `rtol * largest_singular_value`.
+    return torch.linalg.pinv(x, rtol=rcond)
```
hertschuh: Would would you need the `if`, the default for `rtol` is None?

keras/src/ops/linalg.py
```diff
@@ -691,6 +691,58 @@ def lstsq(a, b, rcond=None):
...
+    """
+    if any_symbolic_tensors((x,)):
+        return Pinv(rcond=rcond).symbolic_call(x)
+    return _pinv(x, rcond=rcond)
+
+
+def _pinv(x, rcond=None):
+    x = backend.convert_to_tensor(x)
+    _assert_2d(x)
+    return backend.linalg.pinv(x, rcond=rcond)
```
hertschuh: We don't do this, just in general. Just call `backend.linalg.pinv` directly.

If one of the backend doesn't check for 2D inputs, fix it in that backend.

hertschuh (changes requested):
Thanks for adding this!

keras/src/backend/tensorflow/linalg.py
MarcosAsh: I dropped it, I didn't notice it was redundant.

keras/src/backend/numpy/linalg.py
MarcosAsh: I dropped it, I didn't notice it was redundant.

MarcosAsh:
Hi, thanks for the review! I made the changes.

keras/src/backend/torch/linalg.py
MarcosAsh: I dropped this too, I did keep a one-line comment that rtol here maps to numpy's rcond so the parameter mapping  isn't lost.

keras/src/ops/linalg.py
MarcosAsh: Done. Removed the _pinv helper. Both Pinv.call and the eager path now go straight to backend.linalg.pinv(x, rcond=rcond).

hertschuh (approved):
Thanks!

hertschuh:
Can you add `tpu_atol` and `tpu_rtol` in the new tests' `assertAllClose`? The TPU test are failing because the precision is lower. Just round up the reported absolute / relative errors.

MarcosAsh:
Hi, many thanks for the review! I made all the changes.

hertschuh:
Can you rebase? This is conflicting with https://github.com/keras-team/keras/pull/22772

Thanks!

MarcosAsh:
Hi, just rebased! Many thanks.

MarcosAsh:
Hi, sorry I properly resolved it this time

hertschuh:
Sorry, there was a breakage in the master branch, please rebase.

MarcosAsh:
Hi, just rebased! Many thanks.

hertschuh (approved):

PR #22768, opened 2026-04-24
[Bug] Fix deserialize_keras_object returning dict instead of error for missing class_name

### LINKED ISSUE #22704: [Bug] deserialize_keras_object returns dict when top-level structure is incomplete

Jah-yee:
Bug Fix: deserialize_keras_object error handling for missing class_name

Issue: Fixes #22704

Problem: deserialize_keras_object returns a raw dict when the input config is missing the required class_name key, instead of raising an error.

Changes: Added validation in deserialize_keras_object to raise a clear TypeError when:
1. A non-empty config dict lacks class_name (malformed Keras object config)
2. The config field is not a dict (e.g., is a list) - prevents confusing AttributeError

Closes #22704

### DIFF
--- a/keras/src/layers/core/embedding.py
+++ b/keras/src/layers/core/embedding.py
@@ -369,6 +369,31 @@ class Embedding(Layer):
     @classmethod
     def from_config(cls, config):
         config = config.copy()
+        # Validate input_dim and output_dim are integers
+        if "input_dim" in config:
+            input_dim = config["input_dim"]
+            if (
+                not isinstance(input_dim, int)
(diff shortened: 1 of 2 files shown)


hertschuh:
Issue https://github.com/keras-team/keras/issues/22704 is closed.

PR #22770, opened 2026-04-24
Optimize common cases for `tree.flatten` and `tree.map_structure`.

MarcosAsh:
## Summary

Reduces `Layer.__call__` overhead on the torch backend by skipping `tree` traversal in the positional-argument validation and mask population paths when the layer is invoked with a single tensor (the common case).

## Benchmark

CPU-only torch, 4x64 float32 input, 20000 iterations x 7 trials, median ns per `Layer.__call__`. Reproducible via a small script using only `keras` and `torch`.

| layer | master | this PR |
|---|---|---|
| Identity | 51.0us | 33.1us |
| Dense(8) | 100.1us | 81.5us |
| ReLU | 62.0us | 42.9us |

Keras `3.15.0`, torch `2.10.0+cu128`, Python 3.11, CPU.

## Contributor Agreement

### DIFF
--- a/keras/src/layers/layer.py
+++ b/keras/src/layers/layer.py
@@ -904,20 +904,14 @@ class Layer(BackendLayer, Operation):
         ##########################################################
         # 2. Enforce that only tensors can be passed positionally.
         if not self._allow_non_tensor_positional_args:
-            # Fast path: single tensor arg (the common case), no tree.flatten.
-            if len(args) == 1 and is_backend_tensor_or_symbolic(
-                args[0], allow_none=True
-            ):
-                pass
-            else:
-                for arg in tree.flatten(args):
-                    if not is_backend_tensor_or_symbolic(arg, allow_none=True):
-                        raise ValueError(
-                            "Only input tensors may be passed as "
-                            "positional arguments. The following argument "
-                            f"value should be passed as a keyword argument: "
-                            f"{arg} (of type {type(arg)})"
-                        )
+            for arg in tree.flatten(args):
+                if not is_backend_tensor_or_symbolic(arg, allow_none=True):
+                    raise ValueError(
+                        "Only input tensors may be passed as "
+                        "positional arguments. The following argument value "
+                        f"should be passed as a keyword argument: {arg} "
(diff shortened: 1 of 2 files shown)


hertschuh (commented):
I'm thinking there might be a shorter and broader approach.

How about adding this here:
https://github.com/keras-team/keras/blob/master/keras/src/tree/tree_api.py#L138

```python
    if not tree_impl.is_nested(structure):
        return [structure]
    if isinstance(structure, (list, tuple)) and len(structure) == 1 and not tree_impl.is_nested(structure[0]):
        return [structure[0]]
    return tree_impl.flatten(structure)
```

Can you benchmark and check if you get the same benefits?

We could do something similar for non-nested inputs to `map_structure`

https://github.com/keras-team/keras/blob/master/keras/src/tree/tree_api.py#L200
```python
    if len(structures) == 1 and not tree_impl.is_nested(structures[0]):
        if structures[0] is None and not none_is_leaf:
            return None
        return func(structures[0])
    return tree_impl.map_structure(func, *structures, none_is_leaf=none_is_leaf)
```

MarcosAsh:
Hi, thanks for the review!

Moved the fast paths into \`tree.flatten\` and \`tree.map_structure\` as suggested. Layer code is back to plain \`tree.flatten\` / \`tree.map_structure\` calls.

Bench on torch CPU, 4x64 input, 20k iters × 7 trials, median per call:

| layer | master | layer.py inline | tree_api |
|---|---|---|---|
| Identity | 51.0us | 39.2us | 33.1us |
| Dense(8) | 100.1us | 88.4us | 81.5us |
| ReLU | 62.0us | 45.5us | 42.9us |

The tree-level path is actually a bit faster than the inline one: \`optree.tree_is_leaf\` is a C call, while \`is_backend_tensor_or_symbolic\` chains several Python isinstance checks. And every caller benefits, not just \`Layer.__call__\`.

hertschuh:
This is great, thanks for the work on the benchmarking!

hertschuh (approved):
Thanks!

PR #22771, opened 2026-04-24
Add validation for invalid Sequential layer configs

szepszilo:
## Summary
Add validation in `Sequential.from_config()` for invalid layer configs.

## Problem
Deserializing a `Sequential` model with an invalid layer config such as `{}` raised a low-level internal error instead of a clear validation error.

## Solution
Validate each layer config in `Sequential.from_config()` before deserializing it, and raise a clearer `TypeError` when the layer config is invalid.

## Testing
- Added a regression test in `keras/src/models/sequential_test.py`
- Ran:
  - `pytest keras/src/models/sequential_test.py`

## Contributor Agreement

### DIFF
--- a/keras/src/models/sequential.py
+++ b/keras/src/models/sequential.py
@@ -370,18 +370,31 @@ class Sequential(Model):
 
     @classmethod
     def from_config(cls, config, custom_objects=None):
-        if "name" in config:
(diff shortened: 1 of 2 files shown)


hertschuh (commented):
Can you check the failing tests?

PR #22772, opened 2026-04-24
Add keras.ops.matrix_rank

MarcosAsh:
## Description

Adds `keras.ops.matrix_rank` (and `keras.ops.linalg.matrix_rank`), which returns the rank of one or more matrices via SVD. Fills the gap in `keras.ops.linalg` alongside `inv`, `solve`, `lstsq`, and (in-flight) `pinv`.

Each trainable backend delegates to its native implementation (`tf.linalg.matrix_rank`, `jnp.linalg.matrix_rank`, `torch.linalg.matrix_rank`, `np.linalg.matrix_rank`). The `tol` argument maps to the native threshold semantics on each backend (absolute singular-value cutoff; when `None`, the backend's default is used). OpenVINO raises `NotImplementedError` and the correctness test is added to its exclusion list, matching the existing `lstsq` / `pinv` pattern.

Tests cover square / rectangular / batched inputs, symbolic shape inference, and a rank-deficient correctness case. All 3 new tests pass locally on numpy, jax, torch, and tensorflow backends.

## Contributor Agreement

### DIFF
--- a/keras/src/ops/linalg.py
+++ b/keras/src/ops/linalg.py
@@ -691,6 +691,53 @@ def lstsq(a, b, rcond=None):
     return backend.linalg.lstsq(a, b, rcond=rcond)
 
 
+class MatrixRank(Operation):
+    def __init__(self, tol=None, *, name=None):
+        super().__init__(name=name)
+        self.tol = tol
+
+    def call(self, x):
+        return _matrix_rank(x, tol=self.tol)
+
+    def compute_output_spec(self, x):
+        _assert_2d(x)
+        return KerasTensor(x.shape[:-2], dtype="int32")
+
+
+@keras_export(["keras.ops.matrix_rank", "keras.ops.linalg.matrix_rank"])
+def matrix_rank(x, tol=None):
+    """Return the matrix rank of one or more matrices using SVD.
+
+    The rank is the number of singular values that exceed `tol`. If `tol`
+    is `None`, each backend uses its own default threshold derived from
+    the largest singular value and the matrix dimensions.
+
+    Args:
+        x: Input tensor of shape `(..., M, N)`.
+        tol: Absolute threshold below which singular values are treated
+            as zero. If `None` (default), the backend's default is used.
+
+    Returns:
+        An integer tensor of shape `(...,)` with the rank of each matrix
+        in the batch.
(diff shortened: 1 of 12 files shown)


keras/src/ops/linalg.py
```diff
@@ -691,6 +691,53 @@ def lstsq(a, b, rcond=None):
...
+        An integer tensor of shape `(...,)` with the rank of each matrix
+        in the batch.
+
+    Example:
+
+    >>> a = keras.ops.convert_to_tensor([[1., 2.], [2., 4.]])
+    >>> keras.ops.matrix_rank(a)
+    1
+    """
+    if any_symbolic_tensors((x,)):
```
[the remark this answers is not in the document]
MarcosAsh: Supporting a symbolic tol would need tol to be a call-time argument  rather than __init__-stored state, which is outside this PR's scope.

keras/src/ops/linalg.py
```diff
@@ -691,6 +691,53 @@ def lstsq(a, b, rcond=None):
...
+        return MatrixRank(tol=tol).symbolic_call(x)
+    return _matrix_rank(x, tol=tol)
+
+
+def _matrix_rank(x, tol=None):
+    x = backend.convert_to_tensor(x)
+    _assert_2d(x)
+    return backend.linalg.matrix_rank(x, tol=tol)
```
hertschuh: We don't do this, just in general. Call `backend.linalg.matrix_rank` directly.

If some backend doesn't raise when it's not 2D, then add the check just for that backend.

hertschuh (changes requested):
Thanks for adding this!

keras/src/ops/linalg.py
MarcosAsh: Done. Removed the _matrix_rank helper. Both MatrixRank.call and the eager path call backend.linalg.matrix_rank(x, tol=tol) directly. Also dropped a redundant if tol is None branch in the torch backend (torch.linalg.matrix_rank accepts atol=None directly).

MarcosAsh:
Hi, thanks for the review! I made all the changes requested.

hertschuh (approved):
Thanks!

PR #22774, opened 2026-04-24
Fix deserialize_keras_object to handle tuple and missing fields

### LINKED ISSUE #22704: [Bug] deserialize_keras_object returns dict when top-level structure is incomplete

### LINKED ISSUE #22720: [Bug]  deserialize_keras_object raises internal TypeError for malformed layer list instead of validating input

Jah-yee:
Good day

Thank you for maintaining Keras!

## Summary

This PR improves `deserialize_keras_object` in two ways:

1. **Tuple return**: When deserializing a tuple config, return a tuple (not a list) to preserve the original type
2. **Clear validation**: Add early validation for missing required fields (`class_name`) and invalid config types to raise clear error messages instead of internal TypeErrors
3. **Layer list validation**: Validate each entry in layer lists, raising a clear ValueError for malformed entries (e.g. empty dict `{}`)

## Changes

- `keras/src/saving/serialization_lib.py`:
  - Return `tuple(result)` when config was tuple (line ~540)
  - Validate `class_name` presence before processing (line ~600)
  - Add `isinstance(inner_config, dict)` guard before accessing inner config
  - Validate each layer entry in list/tuple configs

## Issue
- Fixes #22720 (malformed layer entries)
- Fixes #22704 (missing class_name)

## Verification
- Python syntax check passes
- Minimal diff, self-contained

## Contributor Agreement

Thank you for your work on this project. Please let me know if there is anything to adjust.

Warmly,
RoomWithOutRoof

### DIFF
--- a/keras/src/saving/serialization_lib.py
+++ b/keras/src/saving/serialization_lib.py
@@ -618,7 +618,19 @@ def deserialize_keras_object(
         }
 
     class_name = config["class_name"]
-    inner_config = config["config"] or {}
+    inner_config = config["config"]
+    # Validate that inner_config is a dict when class_name is a class
+    # (not "function"), since from_config() expects a dict.
+    if (
+        inner_config is not None
+        and not isinstance(inner_config, dict)
+        and class_name != "function"
+    ):
+        raise TypeError(
+            f"Expected `config` to be a dict, but got {type(inner_config).__name__}: "
+            f"{inner_config}"
+        )
+    inner_config = inner_config or {}
     custom_objects = custom_objects or {}
 
     # Special cases:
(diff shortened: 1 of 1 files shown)


hertschuh:
Bath https://github.com/keras-team/keras/issues/22720 and https://github.com/keras-team/keras/issues/22704 are closed, this is working as designed.

PR #22782, opened 2026-04-28
[OpenVINO] Improve efficiency by using native ops

goyaladitya05:
## Description
This PR improves the efficiency of several OpenVINO backend operations by replacing decompositions with native OpenVINO ops wherever possible.

#### take_along_axis
- Before: Broadcast shape computation grew with tensor rank. O(rank) ops (e.g. 13 ops for rank=4: 3×rank + 1 concat).
- After: Fixed 5 ops regardless of rank.

#### hard_silu / hard swish
- Before: hard_sigmoid (constants + hard_sigmoid node) + multiply -> 4 OV nodes.
- After: native hswish op.

#### L2 norm 
- Before: multiply -> reduce_sum -> sqrt,  3 ops.
- After: native reduce_l2 op.

#### L1 norm 
- Before: abs -> reduce_sum,  2 ops.
- After: native reduce_l1 op.

#### fold
- Before: (range + one_hot + reshape) + convolution_backprop_data + conditional slice for padding,  ~15 ops.
- After:  native col2im op.

#### log_softmax
- Before: reduce_max -> subtract -> exp -> reduce_sum -> log -> subtract, 6 ops (single-axis path).
- After:  native log_softmax op, multi-axis tuple path unchanged.

## Contributor Agreement
**Please review our

### DIFF
--- a/keras/src/backend/openvino/linalg.py
+++ b/keras/src/backend/openvino/linalg.py
@@ -1096,15 +1096,11 @@ def norm(x, ord=None, axis=None, keepdims=False):
     # Ref: jax.numpy.linalg.norm
     if num_axes == 1:
         if ord is None or ord == 2:
-            # L2 norm: sqrt(sum(x * conj(x)))
-            x_conj = x_ov
-            x_sq = ov_opset.multiply(x_conj, x_conj).output(0)
             axis_for_const = list(axis) if isinstance(axis, tuple) else axis
             axis_const = ov_opset.constant(axis_for_const, Type.i32).output(0)
-            norm_result = ov_opset.reduce_sum(
-                x_sq, axis_const, keepdims
-            ).output(0)
-            norm_result = ov_opset.sqrt(norm_result).output(0)
+            norm_result = ov_opset.reduce_l2(x_ov, axis_const, keepdims).output(
+                0
+            )
         elif ord == float("inf"):
             axis_for_const = list(axis) if isinstance(axis, tuple) else axis
             axis_const = ov_opset.constant(axis_for_const, Type.i32).output(0)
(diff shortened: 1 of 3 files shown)


goyaladitya05:
I have a question @hertschuh, unrelated to this PR. These particular tests fail and are excluded because OpenVINO does not support complex dtypes, and as far as I know it is not planned anytime soon. Instead of excluded_concrete_tests.txt, could we add exclusions for these in the test files themselves ? 

```
LayerTest::test_complex_dtype_support
NNOpsCorrectnessTest::test_polar_corectness
NNOpsDtypeTest::test_polar_
NumpyOneInputOpsCorrectnessTest::test_conj
NumpyOneInputOpsCorrectnessTest::test_imag
NumpyOneInputOpsCorrectnessTest::test_isreal
NumpyOneInputOpsCorrectnessTest::test_real
```

hertschuh:
> I have a question @hertschuh, unrelated to this PR. These particular tests fail and are excluded because OpenVINO does not support complex dtypes, and as far as I know it is not planned anytime soon. Instead of excluded_concrete_tests.txt, could we add exclusions for these in the test files themselves ?
> 
> ```
> LayerTest::test_complex_dtype_support
> NNOpsCorrectnessTest::test_polar_corectness
> NNOpsDtypeTest::test_polar_
> NumpyOneInputOpsCorrectnessTest::test_conj
> NumpyOneInputOpsCorrectnessTest::test_imag
> NumpyOneInputOpsCorrectnessTest::test_isreal
> NumpyOneInputOpsCorrectnessTest::test_real
> ```

I actually had a different idea, which I think will make things easier in the long run:

Let's add a new "flag" like these, it would be called `SUPPORTS_COMPLEX_DTYPES`.

https://github.com/keras-team/keras/blob/master/keras/src/backend/openvino/core.py#L23-L25

It would be used to skip the tests and clean up things like this:

https://github.com/keras-team/keras/blob/master/keras/src/backend/common/dtypes_test.py#L38-L39

Is that something you can work on?

hertschuh (approved):
Thanks for the optimizations!

goyaladitya05:
> I actually had a different idea, which I think will make things easier in the long run:
> 
> Let's add a new "flag" like these, it would be called `SUPPORTS_COMPLEX_DTYPES`.
> 
> https://github.com/keras-team/keras/blob/master/keras/src/backend/openvino/core.py#L23-L25
> 
> It would be used to skip the tests and clean up things like this:
> 
> https://github.com/keras-team/keras/blob/master/keras/src/backend/common/dtypes_test.py#L38-L39
> 
> Is that something you can work on?

Sure, I'll work on this soon.

PR #22789, opened 2026-04-28
Fix: compute_fans logic for kernel with rank greater than 2

gaga1313:
## Description
This PR fixes the compute_fans logic for EinsumDense layers when dealing with non-standard weight shapes (such as those in Attention layers).

Previously, compute_fans assumed any tensor with rank greater than 2 was a convolution kernel, leading to incorrect variance scaling for rank-3 tensors in attention projections.

Key Changes
- **Extended compute_fans API:** Updated the function signature to accept two new optional arguments: input_axes and output_axes. This allows explicit calculation of fans based on specified axes instead of relying on shape heuristics.
- **Updated VarianceScaling Initializer:** Modified the VarianceScaling class constructor to accept and store input_axes and output_axes, passing them to compute_fans during execution.
    - **Subclass Compatibility:** Made subsequent changes to VarianceScaling subclasses (like GlorotUniform and HeNormal) to accept these new arguments and pass them to the parent class.
- **Intelligent Auto-detection in EinsumDense:** Updated the EinsumDense.build method to parse the einsum equation, automatically deduce the correct input and output axes for the kernel, and update the initializer accordingly.
- **Comprehensive Tests:** Added supporting tests in random_initializers_test.py to verify:
    - Correct functionality of compute_fans with explicit axes.
    - Correct automatic update of initializer axes in EinsumDense.
    - No regressions in standard Dense layers.

## Contributor Agreement
**Please review our


keras/src/layers/core/einsum_dense.py
```diff
@@ -161,6 +162,43 @@ def __init__(
...
+                    input_axes = [
+                        i
+                        for i, char in enumerate(kernel_spec)
+                        if char in input_spec and char not in output_spec
+                    ]
+                    output_axes = [
+                        i
+                        for i, char in enumerate(kernel_spec)
+                        if char in output_spec
+                    ]
```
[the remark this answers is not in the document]
gaga1313: But the kernel wouldn't have any grouping and broadcasting dimension right? Can you suggest any usecase for this?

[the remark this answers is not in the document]
gaga1313: The dimension in the kernel should be either present in fan_in or fan_out calculation. In the above edits, I do ensure that fan_in has all the input_axes which are present in kernel dims and input dims but not in output dims.

And for fan_out we want all the axes that are present in kernel dims and output dims.

So in this example GI, GIO -> GO

input_axes = [1]
output_axes = [0,2]

keras/src/layers/core/einsum_dense.py
```diff
@@ -161,6 +162,43 @@ def __init__(
         self.gptq_unpacked_column_size = gptq_unpacked_column_size
         self.quantization_config = quantization_config
 
+    def update_kernel_initializer(self):
+        """
```
hertschuh: Rename `_update_kernel_initializer`, this is internal only.

keras/src/layers/core/einsum_dense.py
```diff
@@ -161,6 +162,43 @@ def __init__(
...
+                        i
+                        for i, char in enumerate(kernel_spec)
+                        if char in input_spec and char not in output_spec
+                    ]
+                    output_axes = [
+                        i
+                        for i, char in enumerate(kernel_spec)
+                        if char in output_spec
+                    ]
+
```
hertschuh: Can't we just reuse the result of `_analyze_einsum_string`, which is right above in the `build`.

It seems to have a more complete way to analyze the string. You could pass the info to `_update_kernel_initializer` as parameters.

keras/src/layers/core/einsum_dense.py
```diff
@@ -161,6 +162,43 @@ def __init__(
...
         self.quantization_config = quantization_config
 
+    def update_kernel_initializer(self):
+        """
+        Updates the kernel initializer with the correct input and output axes.
+
+        If the initializer is a VarianceScaling instance, it will be updated
+        with the correct input and output axes based on the einsum equation.
+        """
+        if isinstance(self.kernel_initializer, VarianceScaling):
```
hertschuh: I think you should check if the current `VarianceScaling` initializer already has `input_axes` and `output_axes` and skip.

keras/src/layers/core/einsum_dense.py
```diff
@@ -171,6 +209,9 @@ def build(self, input_shape):
         kernel_shape, bias_shape, full_output_shape = shape_data
         self.full_output_shape = tuple(full_output_shape)
         self.input_spec = InputSpec(ndim=len(input_shape))
+
+        self.update_kernel_initializer()
```
hertschuh: I'd be tempted to make this local, i.e., not modify what was passed by the user, but do:

```python
kernel_initializer = self._update_kernel_initializer(self.kernel_initializer, kernel_shape, full_output_shape)
...
self._kernel = self.add_weight(
                name="kernel",
                shape=tuple(kernel_shape),
                initializer=kernel_initializer,
...
```

hertschuh (commented):
Thanks for the fix!

Is there a unit test that can be added for the changes in EinsumDense? You might need to mock something in `VarianceScaling`.

keras/src/layers/core/einsum_dense.py
gaga1313: @hertschuh If we infer from shape, there might be some ambiguity when different dimension have same size. We might land on the solution most of the time, but more chances of having edge case. 

for example:
input_shape = (batch, 12, 12) (Sequence length 12, Dim 12)
kernel_shape = (12, 12, 12) (Input dim 12, 12 heads, head dim 12)
full_output_shape = (batch, 12, 12, 12)
If we just look at the shapes, every dimension is 12! We cannot know which axis of the kernel belongs to the input and which belongs to the output by just looking at the numbers.

input_axes = [
                        i
                        for i, char in enumerate(kernel_spec)
                        if char in input_spec and char not in output_spec
                    ]
                    output_axes = [
                        i
                        for i, char in enumerate(kernel_spec)
                        if char in output_spec
                    ]
                    
if we use the above logic then input_axes will be [] in the given example.  
and output_axes will be [0, 1, 2]

gaga1313: One thing I can do is perform the calculation in _analyze_einsum_string. Not sure if it will save us calculations, but it will avoid creating a new private function.

keras/src/layers/core/einsum_dense.py
gaga1313: @hertschuh the kernel_initializer is also used in get_config method in EinsumDense. In the get_config method it serializes class attribute self.kernel_initializer. If we don't update the class attribute, we might have to change function signature and calls of get_config method.

keras/src/layers/core/einsum_dense.py
hertschuh: Yes, I would prefer that, please augment `_analyze_einsum_string` to also return the input / output axes.

keras/src/layers/core/einsum_dense.py
hertschuh: Right, so when we serialize and then de-serialize the `EinsumDense`, we end up in the same situation as initially (no `input_axes` and `output_axes` in the initializer), and `self.update_kernel_initializer` fixes it again, which is fine.

I like this better for a few reasons:
- idempotency: if you deserialize a model and re-serialize it, you should get the same thing. If you do change the initializer, the re-serialized model will have `input_axes` and `output_axes`, which the original model didn't have. Model serialization is supposed to be idempotent always.
- Just in general, what you serialize should be what the user gave you, not how you interpreted it. For instance, if the user gave you a negative axis, it's nicer to keep it negative in the serialized structure, even if the code resolves the axis as a positive integer
- Let's say we find a bug in `_update_kernel_initializer` (it resolves the wrong `input_axes` / `output_axes`), we can fix it. If the axes get saved, the bug fix will be ineffective, or will have to bypass the saved `input_axes` / `output_axes`, which might actually be user provided.

hertschuh:
Oh and also please rebase.

hertschuh (approved):
Thanks for the fix!

PR #22791, opened 2026-04-28
Fuse JAX Bidirectional LSTM into a single cuDNN call

MarcosAsh:
## Description

`Bidirectional(LSTM(...))` on the JAX backend issues two separate cuDNN LSTM calls today, one for the forward layer and one for the backward layer. cuDNN already supports a single bidirectional pass that handles both directions in one kernel launch, exposed via `jax.experimental.rnn.lstm(..., bidirectional=True)`.

I added a backend helper that invokes that fused kernel and a dispatch hook on `Bidirectional.call()` that routes to it when both children are plain LSTMs with no mask, no dropout, no stateful, matching units / activations / dtype, and `use_bias=True`. Anything outside that window continues through the existing two-call path.

GRU is unchanged because `jax.experimental.rnn` does not expose a cuDNN GRU binding.

### Benchmark

Inference on a Colab A100, JAX backend, `use_cudnn="auto"`, `return_sequences=True`:

| shape (b x s x in x h) | unfused  | fused    | speedup |
| ---------------------- | -------- | -------- | ------- |
| 32 x 128 x 128 x 128   | 3.95 ms  | 3.54 ms  | 1.12x   |
| 64 x 256 x 256 x 256   | 12.16 ms | 10.68 ms | 1.14x   |
| 64 x 512 x 256 x 256   | 22.04 ms | 21.06 ms | 1.05x   |
| 32 x 1024 x 512 x 512  | 53.08 ms | 48.12 ms | 1.10x   |

Modest but consistent. The dispatch is conservative; non-eligible configs cost nothing.

## Contributor Agreement

### DIFF
--- a/keras/src/layers/rnn/bidirectional.py
+++ b/keras/src/layers/rnn/bidirectional.py
@@ -5,6 +5,7 @@ from keras.src import ops
 from keras.src import utils
 from keras.src.api_export import keras_export
 from keras.src.layers.layer import Layer
+from keras.src.layers.rnn.lstm import LSTM
 from keras.src.saving import serialization_lib
 
 
@@ -201,7 +202,7 @@ class Bidirectional(Layer):
         mask=None,
         training=None,
     ):
-        if self._can_attempt_fused_lstm(mask):
+        if self._can_attempt_fused_lstm(mask, initial_state):
             try:
                 return self._call_fused_lstm(sequences, initial_state)
             except NotImplementedError:
@@ -265,7 +266,7 @@ class Bidirectional(Layer):
             return (output,) + states
         return output
 
-    def _can_attempt_fused_lstm(self, mask):
+    def _can_attempt_fused_lstm(self, mask, initial_state=None):
         # Layer-level preconditions for dispatching to
         # `backend.bidirectional_lstm`. Backends that don't support a
         # fused path (or don't have the runtime resources for one, e.g.
@@ -273,8 +274,8 @@ class Bidirectional(Layer):
         # the caller falls back to the two-layer path.
         if mask is not None or self.stateful:
             return False
-
-        from keras.src.layers.rnn.lstm import LSTM
+        if initial_state is not None and len(initial_state) != 4:
+            return False
 
         fwd, bwd = self.forward_layer, self.backward_layer
         if not isinstance(fwd, LSTM) or not isinstance(bwd, LSTM):
@@ -311,8 +312,6 @@ class Bidirectional(Layer):
             zeros = ops.zeros((batch_size, units), dtype=sequences.dtype)
             fwd_h0, fwd_c0, bwd_h0, bwd_c0 = zeros, zeros, zeros, zeros
         else:
-            if len(initial_state) != 4:
-                raise NotImplementedError
             fwd_h0, fwd_c0, bwd_h0, bwd_c0 = initial_state
 
         fwd_out, bwd_out = backend.bidirectional_lstm(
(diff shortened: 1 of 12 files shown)


hertschuh:
Sorry for the breakage in master, please rebase.

MarcosAsh:
Hi, just rebased! Many thanks.

keras/src/layers/rnn/bidirectional.py
```diff
@@ -200,6 +201,12 @@ def call(
         mask=None,
         training=None,
     ):
+        if self._can_use_jax_cudnn_fused(mask):
+            try:
+                return self._call_jax_cudnn_fused(sequences, initial_state)
+            except NotImplementedError:
+                pass
```
hertschuh: Can we make this backend independent by following the same pattern as the other operations (gru and lstm)? That is, each backend can implement `bidirectional_lstm` or throw `NotImplementedError`.

This will allow the addition of the same optimization for other backends if we ever wanted to.

hertschuh (changes requested):
Thanks for your work on this!

MarcosAsh:
Hi, refactored to dispatch via `backend.bidirectional_lstm` matching the `lstm`/`gru` pattern. Thanks!

keras/src/layers/rnn/bidirectional.py
```diff
@@ -258,6 +265,112 @@ def call(
...
+
+        sequences = ops.convert_to_tensor(sequences)
+        batch_size = ops.shape(sequences)[0]
+
+        if initial_state is None:
+            zeros = ops.zeros((batch_size, units), dtype=sequences.dtype)
+            fwd_h0, fwd_c0, bwd_h0, bwd_c0 = zeros, zeros, zeros, zeros
+        else:
+            if len(initial_state) != 4:
+                raise NotImplementedError
```
hertschuh: Can you move this test to `_can_attempt_fused_lstm`? You'll have to pass the `initial_state` to it.

keras/src/layers/rnn/bidirectional.py
```diff
@@ -258,6 +265,112 @@ def call(
...
+        # Layer-level preconditions for dispatching to
+        # `backend.bidirectional_lstm`. Backends that don't support a
+        # fused path (or don't have the runtime resources for one, e.g.
+        # no GPU) raise `NotImplementedError` from the call itself, and
+        # the caller falls back to the two-layer path.
+        if mask is not None or self.stateful:
+            return False
+
+        from keras.src.layers.rnn.lstm import LSTM
+
```
hertschuh: Move this import to the top (unless you can't)

hertschuh (changes requested):
Thanks for adding this!

MarcosAsh:
Hi, hoisted the `initial_state` arity check into `_can_attempt_fused_lstm` and moved the `LSTM` import up. Thanks!

hertschuh (dismissed):
Thanks!

hertschuh (changes requested):
Please look at the tests, the numerics don't match on GPU (i.e. when this is actually being used)

MarcosAsh:
Hi, thanks for the check! The flat weight blob was packed as [fwd_W, fwd_b, bwd_W, bwd_b], but jax.experimental.rnn expects all W's across both directions first, then all b's. Wrong layout made backward weights and forward biases decode as garbage. Pushed a fix.

hertschuh (approved):
Thanks!

PR #22799, opened 2026-04-29
Add CTC beam search decoder for torch backend

MarcosAsh:
## Description

`ctc_decode(strategy="beam_search")` raised `NotImplementedError` on the torch backend, which was the only backend that lacked it. Users training ASR/OCR/HTR models on the torch backend had no way to do beam-search inference without switching backends.

This PR ports JAX's `_ctc_beam_search_decode` to torch. Output is bit-equivalent to JAX (verified across multiple shapes); semantics match the existing reference, including the tie-breaking flip and emit/blank score tracks.

The 3 `skipTest("torch doesn't support 'beam_search' strategy")` markers in `nn_test.py` are removed; the existing test suite now exercises the new path on torch.

### Implementation note

A direct port using `torch.unique(dim=0)` for per-step path dedup is correct but slow because the row-wise sort dominates the inner loop. `_unique_padded` instead hashes each path to int64 (Knuth's multiplicative constant), sorts the 1D hash array, and verifies adjacent same-hash entries with row equality so collisions stay correct. This gave a 2-3x speedup over the trivial port.

### Benchmark

CPU, single-threaded, JAX backend run without an explicit `jax.jit` wrapper to keep the comparison apples-to-apples eager:

| shape (b x s x c) | beam | torch    | jax      |
| ----------------- | ---- | -------- | -------- |
| 4 x 20 x 12       | 10   | 33 ms    | 36 ms    |
| 4 x 40 x 16       | 15   | 97 ms    | 104 ms   |
| 8 x 60 x 20       | 25   | 429 ms   | 430 ms   |

## Contributor Agreement

### DIFF
--- a/keras/src/backend/torch/nn.py
+++ b/keras/src/backend/torch/nn.py
@@ -1070,6 +1070,12 @@ def _ctc_greedy_decode(
     return indices, scores
 
 
+# Knuth's multiplicative hash constant. Used as the polynomial base in
+# `_unique_padded`; int64 overflow during exponentiation is intentional
+# since we only need a deterministic path -> hash mapping.
+_KNUTH_HASH_CONSTANT = 2654435769
+
+
 def _unique_padded(paths, size, pad):
     """Hash-based row-dedup, padded to a fixed leading size with `pad` rows.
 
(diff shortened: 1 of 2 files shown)


hertschuh (approved):
Thanks for implementing this!

PR #22806, opened 2026-04-30
Fix rel_center_xywh bounding box conversion using wrong height/width multipliers

### LINKED ISSUE #22805: Bounding box conversion from rel_center_xywh produces wrong coordinates for non-square images
# Bug: `_rel_center_xywh_to_xyxy` swaps height and width multipliers

## Description

The `_rel_center_xywh_to_xyxy` method in `BoundingBox` incorrectly multiplies x-coordinates by `height` and y-coordinates by `width`, instead of the correct `width` for x and `height` for y.

## Location

`keras/src/layers/preprocessing/image_preprocessing/bounding_boxes/bounding_box.py`
Method: `_rel_center_xywh_to_xyxy`

## Current (buggy) code

```python
def _rel_center_xywh_to_xyxy(self, boxes, height=None, width=None):
    ops = self.backend
    rel_cx, rel_cy, rel_w, rel_h = ops.numpy.split(boxes, 4, axis=-1)
    half_rel_w = rel_w / 2.0
    half_rel_h = rel_h / 2.0
    x1 = (rel_cx - half_rel_w) * height   # BUG: should be * width
    y1 = (rel_cy - half_rel_h) * width    # BUG: should be * height
    x2 = (rel_cx + half_rel_w) * height   # BUG: should be * width
    y2 = (rel_cy + half_rel_h) * width    # BUG: should be * height
    return self.backend.numpy.concatenate([x1, y1, x2, y2], axis=-1)
```

## Impact

This affects all conversions from `rel_center_xywh` to any format that routes through `xyxy`, producing incorrect bounding box coordinates when `height != width`. This is particularly severe for non-square images (the most common case in practice).

## Minimal reproduction

```python
import numpy as np
import os
os.environ["KERAS_BACKEND"] = "numpy"

from keras.src.layers.preprocessing.image_preprocessing.bounding_boxes.bounding_box import BoundingBox

bbox = BoundingBox()

# rel_center_xywh: center=(0.5,0.5), width=0.4, height=0.2
# On a 200x100 image, expected absolute xyxy = [60, 40, 140, 60]
boxes = np.array([[[0.5, 0.5, 0.4, 0.2]]], dtype="float32")

result = bbox._rel_center_xywh_to_xyxy(boxes, height=100, width=200)
expected = np.array([[[60.0, 40.0, 140.0, 60.0]]], dtype="float32")

print("Expected:", expected)
print("Actual:  ", result)
print("Match:   ", np.allclose(result, expected))
```

### Actual output (buggy)

```
Expected: [[[ 60.  40. 140.  60.]]]
Actual:   [[[ 30.  80.  70. 120.]]]
Match:    False
```

x-coordinates are half what they should be (multiplied by `height=100` instead of `width=200`), and y-coordinates are double (multiplied by `width=200` instead of `height=100`).

pctablet505:
The `_rel_center_xywh_to_xyxy` method in `BoundingBox` incorrectly multiplied x-coordinates by `height` and y-coordinates by `width`, producing wrong bounding box coordinates for non-square images.

This 4-line swap fixes the issue.

Fixes #22805

## Contributor Agreement

### DIFF
--- a/keras/src/layers/preprocessing/image_preprocessing/bounding_boxes/bounding_box.py
+++ b/keras/src/layers/preprocessing/image_preprocessing/bounding_boxes/bounding_box.py
@@ -318,10 +318,10 @@ class BoundingBox:
         rel_cx, rel_cy, rel_w, rel_h = ops.numpy.split(boxes, 4, axis=-1)
         half_rel_w = rel_w / 2.0
         half_rel_h = rel_h / 2.0
-        x1 = (rel_cx - half_rel_w) * height
-        y1 = (rel_cy - half_rel_h) * width
-        x2 = (rel_cx + half_rel_w) * height
-        y2 = (rel_cy + half_rel_h) * width
+        x1 = (rel_cx - half_rel_w) * width
+        y1 = (rel_cy - half_rel_h) * height
+        x2 = (rel_cx + half_rel_w) * width
+        y2 = (rel_cy + half_rel_h) * height
         return self.backend.numpy.concatenate([x1, y1, x2, y2], axis=-1)
 
     def _xyxy_to_yxyx(self, boxes, height=None, width=None):


hertschuh (approved):
Thanks for the fix!

PR #22816, opened 2026-05-02
Fix Keras LiteRT export for Keras 3 + TF 2.20 + Python 3.13

pctablet505:
## Fix Keras LiteRT Export for Keras 3 + TF 2.20 + Python 3.13

**Upstream bug:** https://github.com/tensorflow/tensorflow/issues/117453

### The Problem: Circular Dependency + Python 3.13 Incompatibility

The old `litert.py` called `tf.lite.TFLiteConverter.from_keras_model(model)`. This created a **circular dependency** that crashed on Keras 3 + TF 2.20 + Python 3.13:

```
Keras litert.py
    ↓ calls TF
TensorFlow TFLiteConverter.from_keras_model()
    ↓ sees Keras 3 model, imports
Keras ExportArchive(model)
    ↓ calls TF
TensorFlow saved_model.save(_tf_trackable)
    ↓ walks trackables with
TF SavedModelTrackableView.descendants()
    ↓ calls children() → is_tf_type() on
Keras _DictWrapper (wrapt.ObjectProxy)
    ↓ Python 3.13 inspect.getattr_static crashes on
ObjectProxy.__dict__ descriptor
    ↓
💥 TypeError: 'NoneType' object is not callable
```

There are **two independent crash points** in this loop:

1. **`_DictWrapper` + `wrapt.ObjectProxy` + Python 3.13**: TF 2.20's `tensor_util.is_tf_type()` calls `typing.Protocol.__instancecheck__()`, which uses `inspect.getattr_static()` to access `__dict__`. On Python 3.13, `wrapt.ObjectProxy`'s `__dict__` descriptor is incompatible with `getattr_static`, producing `TypeError`. This was fixed in TF main (commits `e1fab9f0bba`, `80c1d06aa8f`) but not backported to 2.20.

2. **`keras_deps` bridge unregistered**: If the SavedModel path fails, TF's converter falls back to `_freeze_keras_model()` → `_trace_model_call()` → `keras_deps.get_call_context_function()()`. Keras 3 never registered this function, so it returns `None` → `TypeError: 'NoneType' object is not callable`.

### The Solution: Eliminate the Circular Dependency

Instead of handing the model to TF's `from_keras_model()` (which then re-imports Keras `ExportArchive`), we create `ExportArchive` **directly in Keras** and pass the resulting SavedModel to TF's `from_saved_model()`. This is TensorFlow's official Keras 3 conversion path.

```mermaid
flowchart TB
    subgraph OLD["❌ BROKEN: from_keras_model path"]
        K1["Keras litert.py"]
        T1["TF from_keras_model()"]
        K2["Keras ExportArchive"]
        T2["TF saved_model.save()"]
        T3["TrackableView.descendants()"]
        E1["💥 TypeError"]
        T4["_trace_model_call() fallback"]
        E2["💥 TypeError"]

        K1 --> T1
        T1 --> K2
        K2 --> T2
        T2 --> T3
        T3 --> E1
        T2 -.-> T4
        T4 --> E2
    end

    subgraph NEW["✅ FIXED: direct ExportArchive path"]
        K3["Keras litert.py"]
        K4["Keras ExportArchive"]
        T5["TF saved_model.save()"]
        T6["TF from_saved_model()"]
        E3["✅ .tflite"]

        K3 --> K4
        K4 --> T5
        T5 --> T6
        T6 --> E3
    end

    style E1 fill:#ffcdd2,stroke:#c62828
    style E2 fill:#ffcdd2,stroke:#c62828
    style E3 fill:#c8e6c9,stroke:#2e7d32
```

**Why this fixes both crashes:**
- **Crash 1 eliminated**: By controlling `ExportArchive` from Keras directly, we avoid TF's `from_keras_model()` code path entirely. The `_DictWrapper` issue in `SavedModelTrackableView` is also guarded with `try/except TypeError` (see change 2 below).
- **Crash 2 eliminated**: The `_freeze_keras_model()` fallback only exists in `from_keras_model()`. We never call it.

#### 1. `keras/src/export/litert.py` — Complete rewrite of conversion path

| Before | After |
|--------|-------|
| `tf.lite.TFLiteConverter.from_keras_model(self.model)` | `ExportArchive` → `write_out()` → `tf.lite.TFLiteConverter.from_saved_model(saved_model_dir)` |
| 3-case adapter branching (single / flat list / nested) | Single path for all model types |
| `_create_nested_inputs_adapter()` wrapper model | Removed — dict inputs handled natively by SavedModel signatures |
| `try/except` around conversion with vague error message | Direct conversion, clear error propagation |

**Key code:**
```python
with tempfile.TemporaryDirectory() as saved_model_dir:
    archive = keras_export.ExportArchive()
    archive.track(self.model)
    archive.add_endpoint(
        "serve",
        functools.partial(self.model.__call__, training=False),
        input_signature=input_signature,
    )
    archive.write_out(saved_model_dir, verbose=False)

    converter = tf.lite.TFLiteConverter.from_saved_model(saved_model_dir)
    converter.target_spec.supported_ops = [
        tf.lite.OpsSet.TFLITE_BUILTINS,
        tf.lite.OpsSet.SELECT_TF_OPS,
    ]
    converter.experimental_enable_resource_variables = True
    return converter.convert()
```

#### 2. `keras/src/export/saved_model_export_archive.py` — Guard + context manager

Master already has `_patch_tf_is_tf_type_for_object_proxy()` (a context-manager backport of TF main's fix). This PR adds a defensive `try/except TypeError` around `SavedModelTrackableView(root).descendants()` so that even if the `is_tf_type` patch misses a case, the export continues (skipping lookup-table tracking for that root instead of crashing).

```python
if hasattr(self, "_tracked"):
    for root in self._tracked:
        try:
            descendants = SavedModelTrackableView(root).descendants()
        except TypeError:
            # Python 3.13 + wrapt.ObjectProxy incompatibility
            continue
        for trackable in descendants:
            ...
```

#### 3. `keras/src/export/litert_test.py` — Test updates

- **Removed `tf.lite.Interpreter` fallback** — tests now require `ai_edge_litert` interpreter exclusively
- Dict-input tests feed dicts natively to interpreter (no adapter flattening)
- Custom signature tests include `name=` in `InputSpec` to preserve TFLite input names

## Contributor Agreement
**Please review our

### DIFF
--- a/keras/src/export/litert.py
+++ b/keras/src/export/litert.py
@@ -1,6 +1,6 @@
-from keras.src import layers
-from keras.src import models
-from keras.src import tree
+import functools
+import tempfile
+
 from keras.src.export.export_utils import get_input_signature
 from keras.src.utils import io_utils
 from keras.src.utils.module_utils import tensorflow as tf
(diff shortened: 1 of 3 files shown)


hertschuh (approved):
Thanks for the fix!

PR #22821, opened 2026-05-03
Fix torch convert_to_tensor: honor explicit dtype for Python scalars

### LINKED ISSUE #22815: Torch backend: convert_to_tensor creates float64 constants in torch.export graphs when dtype='float32' is requested
## Bug Description

In the PyTorch backend, `keras.ops.convert_to_tensor(scalar_float, dtype="float32")` incorrectly produces `float64` constants in `torch.export` graphs. This happens because the function internally converts Python scalar floats to `np.array(float)` (which defaults to `float64`) before passing to `torch.as_tensor`. During `torch.export`, the original `float64` numpy array gets lifted as a constant *before* the cast to `float32` occurs in the graph.

This causes downstream MLIR verification errors in strict conversion pipelines (e.g. `litert_torch.convert`), which reject dtype mismatches between operation operands.

## Standalone code to reproduce

```python
import os

import torch
import keras

class ModuleWithScalarConvert(torch.nn.Module):
    def forward(self, x):
        # This is what happens inside ops like maximum() when a Python
        # scalar (e.g. backend.epsilon()) is used alongside a float32 tensor.
        scalar = keras.ops.convert_to_tensor(1e-7, dtype="float32")
        return torch.maximum(x, scalar)

model = ModuleWithScalarConvert()
exported = torch.export.export(model, (torch.ones(5, dtype=torch.float32),))

print("Exported constants:")
for name, val in exported.constants.items():
    if isinstance(val, torch.Tensor):
        print(f"  {name}: dtype={val.dtype}, shape={val.shape}")
```

**Actual output:**
```
  lifted_tensor_0: dtype=torch.float64, shape=torch.Size([])
```

**Expected output:**
```
  lifted_tensor_0: dtype=torch.float32, shape=torch.Size([])
```

## Impact

Any layer that mixes Python scalar floats with float32 tensors via backend ops will silently inject `float64` constants into exported graphs. This was first observed with `layers.Normalization` + `litert_torch.convert`, but the root cause is in the core torch backend and could affect any `torch.export` workflow.

## Proposed Fix

In `keras/src/backend/torch/core.py`, skip `np.array()` for scalar Python values (`bool`, `int`, `float`) so they go directly to `torch.as_tensor()` with the requested dtype:

```diff
-    if not isinstance(x, (list, tuple)):
+    if isinstance(x, (list, tuple)):
+        if len(x) > 0 and any(isinstance(x1, torch.Tensor) for x1 in x):
+            return torch.stack([convert_to_tensor(x1) for x1 in x])
+    elif not isinstance(x, (bool, int, float)):
         x = np.array(x)
-    elif len(x) > 0 and any(isinstance(x1, torch.Tensor) for x1 in x):
-        return torch.stack([convert_to_tensor(x1) for x1 in x])
```

## System information

- Keras version: master (commit ~4caa4a087)
- Backend: torch
- PyTorch version: 2.11.0
- Python version: 3.13.7
- OS: Linux

MarcosAsh:
Fixes #22815.

## Summary

On the torch backend, `convert_to_tensor(scalar, dtype=...)` for a Python `bool`/`int`/`float` was routed through `np.array(x)` before the cast. `np.array(1.5)` is `float64`, so under `torch.export` the float64 constant gets lifted into the traced graph *before* the `torch.as_tensor(..., dtype=float32)` cast runs. Strict downstream pipelines (e.g. `litert_torch.convert`) then reject the dtype mismatch.  The scalar fast paths in `convert_to_tensor` were gated on `dtype is None`, so when an explicit dtype was passed they were skipped and the value fell through to `np.array(x)`. Lift that gate: handle Python scalars in both cases by passing the requested dtype straight to `torch.as_tensor`. No numpy round-trip, no float64 intermediate. I also added the necessary tests. 

Repro from the issue:

```python
import os
import torch, keras

class M(torch.nn.Module):
    def forward(self, x):
        return torch.maximum(x, keras.ops.convert_to_tensor(1e-7, dtype="float32"))

exported = torch.export.export(M(), (torch.ones(5, dtype=torch.float32),))
for name, val in exported.constants.items():
    print(name, val.dtype)
```

Before: `lifted_tensor_0 torch.float64`
After:  `lifted_tensor_0 torch.float32`

## Contributor Agreement

### DIFF
--- a/keras/src/backend/torch/core.py
+++ b/keras/src/backend/torch/core.py
@@ -207,19 +207,18 @@ def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
         if dtype is not None:
             x = x.to(to_torch_dtype(dtype))
         return x
-    if dtype is None:
-        if isinstance(x, bool):
-            return torch.as_tensor(x, dtype=torch.bool, device=get_device())
+    if isinstance(x, (bool, int, float, complex)):
+        if dtype is not None:
+            dt = to_torch_dtype(dtype)
+        elif isinstance(x, bool):
+            dt = torch.bool
         elif isinstance(x, int):
-            if x < -(2**31) or x >= 2**31:
-                return torch.as_tensor(
-                    x, dtype=torch.int64, device=get_device()
-                )
-            return torch.as_tensor(x, dtype=torch.int32, device=get_device())
+            dt = torch.int64 if x < -(2**31) or x >= 2**31 else torch.int32
         elif isinstance(x, float):
-            return torch.as_tensor(
-                x, dtype=to_torch_dtype(floatx()), device=get_device()
-            )
+            dt = to_torch_dtype(floatx())
+        else:
+            dt = torch.complex64
+        return torch.as_tensor(x, dtype=dt, device=get_device())
 
     # Convert to np in case of any array-like that is not list or tuple.
     if not isinstance(x, (list, tuple)):
--- a/keras/src/ops/core_test.py
+++ b/keras/src/ops/core_test.py
@@ -1805,6 +1805,50 @@ class CoreOpsBehaviorTests(testing.TestCase):
         with self.assertRaises(ValueError):
             ops.convert_to_numpy(KerasTensor((2,)))
 
+    def test_convert_to_tensor_python_float_with_dtype(self):
+        # Regression: a Python float passed with an explicit dtype must
+        # not go through a float64 numpy intermediate. On the torch
+        # backend, torch.export would otherwise lift the float64
+        # constant into the graph before the cast runs.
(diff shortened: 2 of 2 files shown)


MarcosAsh:
I applied geminis suggestions they where valid.

hertschuh (approved):
Thanks for the fix!

hertschuh (changes requested):
The export test is failing on GPU:

https://github.com/keras-team/keras/actions/runs/25339807840/job/74294083897?pr=22821

I think the model has to be fully on CPU for the export to work. You'll have to force it to do that.

MarcosAsh:
Hi thanks! I just fixed it

hertschuh (approved):
Thanks!

Issue #22705, opened 2026-04-18
[Bug] deserialize_keras_object accepts callable objects in config without validation


amadhan882:
## Description

`deserialize_keras_object` accepts callable objects ( functions or lambdas) in configuration fields without validation. These callables are stored and later invoked during layer execution (e.g., during `build()`), which can lead to unexpected runtime behavior.

## Reproduction Script 

```python
import os
os.environ["KERAS_BACKEND"] = "tensorflow"

from keras.saving import deserialize_keras_object

def test_fn(*args, **kwargs):
    print("CALLABLE EXECUTED")
    return 1.0

config = {
    "class_name": "Dense",
    "module": "keras.layers",
    "config": {
        "units": 4,
        "kernel_initializer": test_fn
    }
}

layer = deserialize_keras_object(config)

print("Deserialized type:", type(layer))

# Trigger execution
layer.build((None, 10))
````

## Observed behavior

* Callable object is accepted without validation
* Callable is invoked during layer execution (`build`)
* No error is raised during deserialization

## Expected behavior

* Configuration should be restricted to serializable and validated types
* Callable objects should be rejected or explicitly handled
* Deserialization should not allow arbitrary callables in config

## Notes

Keras emits a warning when serializing lambda functions, but during deserialization callable objects are still accepted and executed.

This may lead to unexpected or unsafe behavior if configuration input is not fully trusted.

**Comment by Saumay** [comment]
Took a shot at this in #22736, updated with wider coverage after a self-review pass.

The fix walks the config recursively and rejects raw Python callables: `types.FunctionType`, `BuiltinFunctionType`, `MethodType`, `BuiltinMethodType`, `functools.partial`, `functools.partialmethod`. The walk runs before `cls.from_config`, `build_from_config`, and `compile_from_config` see the config, so smuggled callables in `config["config"]`, `config["build_config"]`, `config["compile_config"]`, or any other outer-level key are all caught.

Quick examples (default `safe_mode=True`):

```python
# Rejected
deserialize_keras_object({
    "class_name": "Dense",
    "module": "keras.layers",
    "config": {"units": 4, "kernel_initializer": my_fn},  # raw callable
})

# Allowed: roundtrip of a real config, where the activation function is
# stored as a serialized dict (`{"class_name": "function", ...}`) rather
# than a raw callable.
layer = keras.layers.Dense(4, activation="relu")
deserialize_keras_object(keras.saving.serialize_keras_object(layer))
```

[...]

PR #22822, opened 2026-05-03
Fix namespace hijacking and callable injection in deserialize_keras_object

### LINKED ISSUE #22779: [Bug] Namespace Hijacking via Priority Inversion in `deserialize_keras_object`
## Description
The `registered_name` field in Keras serialization utilities allows for **Namespace Shadowing**. The framework prioritizes resolving an object via the `registered_name` registry over the standard `class_name` and `module` resolution. Currently, there is no validation to prevent custom objects from being registered with names that conflict with core Keras components (e.g., `Dense`). This enables a configuration to appear as a standard Keras layer while silently executing logic from a different, potentially malicious, package registry.

## Reproduction Script

```python
import os
import keras
import numpy as np

# 1. THE ATTACKER'S SIDE
# Attacker registers a malicious class with the EXACT same name as a built-in layer
# but under a different package name.
@keras.saving.register_keras_serializable(package='MaliciousPackage', name='Dense')
class HiddenLogicDense(keras.layers.Dense):
    def call(self, inputs):
        print("Malicious Logic Executed!")
        print("Exfiltrating Data Shape:", inputs.shape)
        return super().call(inputs)

# 2. THE VICTIM'S SIDE (Normal User)
# The config looks normal (class_name: Dense), but 'registered_name' triggers the hijack.
hijacked_config = {
    "module": "keras.layers",
    "class_name": "Dense",
    "config": {
        "units": 5,
        "activation": "relu",
        "name": "dense_layer"
    },
    "registered_name": "MaliciousPackage>Dense" 
}


try:
    # Deserialization prioritizes 'registered_name' over the built-in 'class_name'
    hijacked_layer = keras.saving.deserialize_keras_object(hijacked_config)
    
    # Verification: Creating dummy input to trigger the 'call' method
    test_input = np.random.random((1, 10)).astype("float32")
    output = hijacked_layer(test_input)
    
    print("\nResult:")
    if isinstance(hijacked_layer, HiddenLogicDense):
        print("STATUS: Hijack Successful. User is running Attacker's code!")
    else:
        print("STATUS: Safe. Running original Keras code.")

except Exception as e:
    print(f"Error during execution: {e}")
```

## Observed Behavior 

```text

Malicious Logic Executed!
Exfiltrating Data Shape: (1, 10)

Result:
STATUS: Hijack Successful. User is running Attacker's code!

```

## Expected Behavior
* **Reserved Namespaces:** The framework should maintain a list of reserved names (core layers, optimizers, etc.) that cannot be used in custom registrations.
* **Integrity Check:** During deserialization, if a `registered_name` is used, the framework should verify that the resulting object matches the class structure expected by the `class_name` field.
* **Shadowing Warnings:** Any attempt to override a built-in component should require an explicit opt-in or raise a high-severity warning.

## Notes
This behavior creates a significant security risk in model supply chains. Users loading model configurations from third-party sources may unintentionally execute untrusted code while under the impression they are utilizing standard, built-in Keras layers.

pctablet505:
Fixes #22779 and #22705.

**Issue #22779 — Namespace Hijacking via Priority Inversion**
- Prevents globally registered custom objects from shadowing built-in Keras classes when the config claims a keras.* module.
- Resolves the built-in class first and rejects mismatched registered_name via _assert_no_registered_name_for_builtin.

**Issue #22705 — Callable Objects in Config**
- Adds _validate_config helper that recursively rejects raw callable objects in inner_config before passing to cls.from_config().
- Handles circular references to avoid RecursionError on malicious input.

### DIFF
--- a/keras/src/saving/serialization_lib.py
+++ b/keras/src/saving/serialization_lib.py
@@ -732,6 +732,7 @@ def deserialize_keras_object(
     # so that we can catch any custom objects that the config refers to.
     custom_obj_scope = object_registration.CustomObjectScope(custom_objects)
     safe_mode_scope = SafeModeScope(safe_mode)
+    _validate_config(inner_config)
     with custom_obj_scope, safe_mode_scope:
         try:
             instance = cls.from_config(inner_config)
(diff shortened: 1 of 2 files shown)


divyashreepathihalli:
not a valid bug

PR #22825, opened 2026-05-04
Auto-enable tf2onnx large_model for models >1.5GB

### LINKED ISSUE #2340: keras_hub model export not wroking.
**Describe the bug**

**To Reproduce**

Provide a link to a [Colab Notebook](https://colab.research.google.com/gist/pctablet505/323fd77bccd1830eec536eea3242fe3d/keras-hub-model-export-bug.ipynb), which reproduces the bug.

**Expected behavior**

It shall generate a .onnx file without issues.

**Additional context**

probably it is due to external dependency on [`tf2onnx`](https://github.com/onnx/tensorflow-onnx) which is not managed properly.

**Would you like to help us fix it?**
yes, it would be great if it is fixed.

pctablet505:
TensorFlow GraphDef has a 2GB protobuf limit. When exporting large models (e.g. Llama 1B) to ONNX via tf2onnx, freezing all variables into the graph causes `import_graph_def` to fail with:

```
RuntimeError: size too big: 18446744072413376424 details: string length exceeds max size
```

This change estimates total weight size and passes `large_model=True` to tf2onnx when the model exceeds 1.5GB, triggering external weight storage and avoiding the protobuf limit.

Fixes keras-team/keras-hub#2340

**Tested:**
- Small model export continues to work (existing CI).
- Large model export (Llama 3.2 1B) now succeeds with `large_model=True`.

**Caveats:**
- Threshold is fixed at 1.5GB. Models with massive graph constants (not weights) could still hit the limit.
- External weight files are written next to the ONNX file when threshold is exceeded.


.github/workflows/tpu_tests.yml
```diff
@@ -27,15 +27,22 @@ jobs:
...
+            name: Run tests on multi-TPU
 
     container:
       image: python:3.11-slim
       options: --privileged --network host
 
-    name: ${{ format('Run tests on {0}TPU', matrix.multi_device && 'multi-' || '') }}
-    runs-on: ${{ matrix.multi_device && 'linux-x86-ct5lp-112-4tpu' || 'linux-x86-ct6e-44-1tpu' }}
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.runner }}
```
hertschuh: Undo, this is part of another PR

keras/src/export/onnx.py
```diff
@@ -99,13 +100,26 @@ def export_onnx(
...
+        # path stores weights externally to avoid this.
+        total_bytes = 0
+        for w in model.weights:
+            shape = w.shape
+            if shape is None or None in shape:
+                continue
+            total_bytes += math.prod(shape) * w.value.dtype.size
+
+        large_model = total_bytes > (3 * 1024**3 // 2)
+
```
hertschuh, line 128: `3 * 1024**3 // 2` is not super readable.

I would be willing to just change it to 1GB.

Also, is there any downside to always using the `large_model` option?

hertschuh (changes requested):

keras/src/export/onnx.py
hertschuh: Also, is this still a draft?

pctablet505: yeah, I am trying to find the root cause for it, maybe the right fix could be in ONNX repo instead of keras.

hertschuh (approved):
Thanks!

PR #22832, opened 2026-05-05
[Fix] Dense layer GPU OOM with rank-3 input due to BatchMatMulV2 gradient materialization

### LINKED ISSUE #22823: Dense layer GPU OOM with rank-3 input due to BatchMatMulV2 gradient materialization
## Description

The `Dense` layer in Keras 3 causes GPU out-of-memory errors when processing rank-3 inputs with large feature dimensions, due to the switch from `tf.tensordot` (Keras 2) to `ops.matmul` (Keras 3) in the layer's `call()` method.

## Root cause

In Keras 2, `tf.keras.layers.Dense` used `tf.tensordot(inputs, kernel, [[rank-1], [0]])` for rank > 2 inputs. This internally reshapes `[B, T, K]` → `[B*T, K]`, performs a 2D matmul, then reshapes back. The gradient is computed on the 2D tensors — memory-efficient.

In Keras 3, `Dense.call()` uses `ops.matmul(inputs, kernel)`, which on the TF backend maps to `tf.matmul` → `BatchMatMulV2`. When the kernel is rank-2 `[K, N]` and the input is rank-3 `[B, T, K]`, the gradient computation (in `tensorflow/python/ops/math_grad.py`) materializes the full broadcasted intermediate `[B, T, K, N]` before reducing to `[K, N]`.

For realistic model sizes (batch=8192, seq=10, features=1411, units=1411), this intermediate is 607 GiB — immediate OOM on any GPU.

## Reproduction

```python
import keras
import numpy as np

# Simulates a model processing sequences (e.g., DCN for ranking)
model = keras.Sequential([
    keras.layers.Dense(1411, name="cross_layer_dense"),
])

x = np.random.randn(8192, 10, 1411).astype("float32")
y = np.random.randn(8192, 10, 1411).astype("float32")

model.compile(optimizer="adam", loss="mse")
model.fit(x, y, batch_size=8192, epochs=1)  # OOM on GPU
```

## Expected behavior

The gradient should be computed in O(B * T * K + K*N) memory (as Keras 2 did via tensordot), not O(B * T * K * N).

## Suggested fix

In the TF backend's `matmul` implementation (`keras/src/backend/tensorflow/numpy.py`), detect the case where one input is rank-2 and use `tf.tensordot` instead of `tf.matmul`:

```python
def matmul(x1, x2):
    if x2.shape.rank == 2 and x1.shape.rank > 2:
        return tf.tensordot(x1, x2, axes=[[-1], [0]])
    # ... existing implementation
```

This restores the efficient gradient path that Keras 2 used, while maintaining identical forward-pass semantics.

## Environment

- Keras 3.13.2
- TensorFlow 2.19
- GPU: NVIDIA A10G (24 GB)

## Related

- PyTorch equivalent issue: https://github.com/pytorch/pytorch/issues/154128
- TF `BatchMatMulV2` gradient uses `reduce_sum` after full broadcast: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/math_grad.py

## Workaround

Manually reshape rank-3 inputs to 2D before Dense layers, then restore the original shape after.

ChiragSW:
## Description
Fixes: #22823 
The fix changes the TensorFlow backend matmul dense path so this case uses:

```tf.tensordot(x1, x2, axes=[[-1], [0]])```

instead of:

```tf.matmul(x1, x2)```

## Contributor Agreement
**Please review our

### DIFF
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -622,6 +622,13 @@ def matmul(x1, x2):
             output = tf.tensordot(x1, x2, axes=1)
         elif x1_shape.rank == 1:
             output = tf.tensordot(x1, x2, axes=[[0], [-2]])
+        elif (
+            x2_shape.rank == 2
+            and x1_shape.rank is not None
+            and x1_shape.rank > 2
+            and output_type is None
+        ):
+            output = tf.tensordot(x1, x2, axes=[[-1], [0]])
         else:
             output = tf.matmul(x1, x2, output_type=output_type)
         return tf.cast(output, result_dtype)


hertschuh (approved):
Thanks for the improvement!

PR #22834, opened 2026-05-05
Use AbstractMesh-backed sharding inside JAX tracing scope

### LINKED ISSUE #21027: Use JAX's AbstractMesh in distribution lib

MarcosAsh:
## Description

`distribute_tensor` passes a concrete-`Mesh` `NamedSharding` into `jax.lax.with_sharding_constraint` inside the tracing scope, so the JIT cache key includes the concrete mesh and any device swap busts the cache. Swapping in `mesh.abstract_mesh` keeps the constraint logically identical but keys the cache only on axis names/sizes.

I added a `_to_abstract_sharding` helper that downgrades `NamedSharding(mesh, spec)` to `NamedSharding(mesh.abstract_mesh, spec)` when the running JAX exposes `abstract_mesh`, and is a no-op otherwise. Only the in-tracing branch uses it; `device_put` and other real-placement paths still see the concrete mesh.

Closes #21027. Picks up the design from #22100 (closed after going stale).

## Contributor Agreement

### DIFF
--- a/keras/src/backend/jax/distribution_lib.py
+++ b/keras/src/backend/jax/distribution_lib.py
@@ -203,6 +203,29 @@ def _to_backend_mesh(device_mesh):
     return jax.sharding.Mesh(devices, device_mesh.axis_names)
 
 
+def _to_abstract_sharding(sharding):
+    """Convert a sharding to use AbstractMesh for JIT-stable cache keys.
+
+    For a `NamedSharding(mesh, spec)`, returns one with `mesh.abstract_mesh`
+    when the running JAX version exposes it; otherwise the sharding is
+    returned unchanged. Other sharding types are passed through.
+
+    Args:
+        sharding: A single `jax.sharding.Sharding`, or `None`.
+
+    Returns:
+        A sharding with `AbstractMesh` where available, else the original.
+    """
+    if sharding is None:
+        return None
+    if not isinstance(sharding, jax.sharding.NamedSharding):
+        return sharding
+    abstract = getattr(sharding.mesh, "abstract_mesh", None)
+    if abstract is None:
+        return sharding
+    return jax.sharding.NamedSharding(abstract, sharding.spec)
+
+
 def _to_backend_layout(tensor_layout):
     """Convert the TensorLayout to JAX backend specific Sharding.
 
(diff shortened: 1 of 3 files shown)


hertschuh (commented):
This code:

```python
    if jax_utils.is_in_jax_tracing_scope(tensor):
        return jax.lax.with_sharding_constraint(tensor, layout)
```
is actually dead code, it's never used by `model.fit` or `model.evaluate` or `model.predict`. This was done "just in case" with no actual concrete scenario in mind.

The other thing is that everywhere we use `jax.jit` we use concrete meshes.

> JIT cache key includes the concrete mesh and any device swap busts the cache.

So there is no workflow using Keras today where you would prevent a cache miss. `model.fit` etc. don't support swapping devices.

All that to say is that this change is very theoretical and will probably remain dead code for a long time.

I'd rather we address this end to end with a specific concrete scenario in mind that we implement and test end to end.

MarcosAsh:
Hi, thanks for the review. You're right that the `with_sharding_constraint` branch is dead in fit/evaluate/predict. Rescoped to `trainer.py:265-279` (and the test/predict equivalents), the spot you pointed @bitloi at in #21027. `state_shardings` now goes through `_to_abstract_sharding` before `jit(out_shardings=...)`. The original `distribute_tensor` change is reverted. Existing e2e distribution tests pass. Let me know if you'd like a dedicated cache-reuse test.

MarcosAsh:
Ran a Colab CPU repro with 8 fake devices, `device_put` inputs on two meshes with identical axes but different device slices, and the keras-trainer-shaped `jit(out_shardings=...)` rebuild scenario this helper targets:

```
MASTER (concrete out_shardings):   M1 first = 79.4 ms,  M2 first = 76.8 ms
WITH FIX (abstract out_shardings): M1 first = 71.6 ms,  M2 first = 73.4 ms
```

The M2 rebuild delta is 3.4 ms, within noise. The hoped-for "cache hit on device swap" doesn't materialize because JAX already content-addresses the executable below the PjitFunction layer, so changing the mesh in `out_shardings` doesn't actually trigger a recompile in the single-process trainer path. Sharding-identity helper passed in isolation (concrete shardings unequal, abstract equal), but that precondition isn't enough to deliver a user-visible win.

Closing. If a workflow turns up where the persistent disk cache (`jax_compilation_cache_dir`) across processes does show a recompile here, I'll reopen with a real benchmark.

PR #22853, opened 2026-05-08
Validate input channels in conv/depthwise_conv/separable_conv symbolic path

### LINKED ISSUE #22852: [Bug] keras.ops.conv / depthwise_conv / separable_conv skip input-channel validation in symbolic path
### Summary

`keras.ops.conv`, `keras.ops.depthwise_conv`, and `keras.ops.separable_conv` validate that the input tensor's channel dimension matches the kernel's input channels in the **eager** path, but the **symbolic** path silently returns an output shape based on the (mismatched) kernel — masking the bug until eager execution.

This matches the pattern of #22515 (separable_conv shape) and #22472 (rgb_to_hsv channel validation), where eager and symbolic paths disagree on validation.

### Reproduction

```python
import os
import keras
import keras.ops as ops
import numpy as np

# 1) keras.ops.conv
x = np.random.normal(size=(1, 10, 3)).astype("float32")        # 3 input channels
kernel = np.random.normal(size=(3, 5, 4)).astype("float32")     # kernel expects 5 input channels (mismatch)

try:
    print("eager   :", ops.conv(x, kernel, padding="valid").shape)
except Exception as e:
    print("eager   : raises", type(e).__name__)

print("symbolic:", ops.conv(keras.Input(shape=(10, 3)), kernel, padding="valid").shape)

# 2) keras.ops.depthwise_conv
dw = np.random.normal(size=(3, 5, 1)).astype("float32")        # depthwise kernel expects 5 input channels
try:
    print("eager   :", ops.depthwise_conv(x, dw, padding="valid").shape)
except Exception as e:
    print("eager   : raises", type(e).__name__)

print("symbolic:", ops.depthwise_conv(keras.Input(shape=(10, 3)), dw, padding="valid").shape)

# 3) keras.ops.separable_conv
pw = np.random.normal(size=(1, 5, 4)).astype("float32")
try:
    print("eager   :", ops.separable_conv(x, dw, pw, padding="valid").shape)
except Exception as e:
    print("eager   : raises", type(e).__name__)

print("symbolic:", ops.separable_conv(keras.Input(shape=(10, 3)), dw, pw, padding="valid").shape)
```

### Observed

```
# conv
eager   : raises ValueError
symbolic: (None, 8, 4)

# depthwise_conv
eager   : raises ValueError ("input channels=3, kernel input channels=5")
symbolic: (None, 8, 5)

# separable_conv
eager   : raises ValueError
symbolic: (None, 8, 4)
```

### Expected

Both paths should raise the same `ValueError` when input channels don't match the kernel's input channels. The symbolic path currently fabricates an output shape that the model would never actually be able to compute.

### Root cause

In `keras/src/ops/nn.py`:

- `Conv.compute_output_spec` (~line 1510) computes output shape from `kernel.shape[-1]` and the spatial dims, but does not check `inputs.shape[channel_dim] == kernel.shape[-2]`.
- `DepthwiseConv.compute_output_spec` (~line 1553) has the same gap.
- `SeparableConv.compute_output_spec` (~line 1659) inherits the gap by delegating to `depthwise_conv`.

The eager path validates via `keras/src/backend/common/backend_utils.check_depthwise_conv_input_channels` (added in #22687), but that helper isn't reused in the symbolic path.

### Impact

Models built from a `keras.Input` placeholder with wrong-channel kernels are accepted at construction time and only fail at eager execution, when the user has already finished wiring up the model. A clear early `ValueError` at shape inference is consistent with how other ops validate.

rstar327:
## Description

Fixes #22852.

`keras.ops.conv`, `keras.ops.depthwise_conv`, and `keras.ops.separable_conv` validate that the input tensor's channel dimension matches the kernel's input channels in their **eager** path, but the **symbolic** path (`compute_output_spec`) silently fabricates an output shape from the mismatched kernel. The mismatch only surfaces at eager execution time, after the model has already been wired up.

This change reuses the existing channel-validation helper (added in #22687 for the eager path) and calls it from the symbolic path for all three ops.

### Before

```python

x_sym = keras.Input(shape=(10, 3))                              # 3 channels
kernel = np.random.normal(size=(3, 5, 4)).astype("float32")      # in_channels=5

ops.conv(x_sym, kernel, padding="valid").shape  # (None, 8, 4)  silent
```

### After

```
ValueError: The number of input channels must match the kernel's input
channels. Received: input channels=3, kernel input channels=5,
data_format='channels_last'.
```

Same fix applied to `depthwise_conv` and `separable_conv`. Dynamic channel dimensions (`None`) still pass through, matching the eager path's behavior.

### Implementation

- Renamed `check_depthwise_conv_input_channels` → `check_conv_input_channels` since the helper is identical for all three convs (they share the `kernel.shape[-2]` input-channel convention). Updated the four backend call sites accordingly.
- Added the call in `Conv.compute_output_spec`, `DepthwiseConv.compute_output_spec`, and `SeparableConv.compute_output_spec` in `keras/src/ops/nn.py`.
- Added `test_conv_input_channel_validation` in `NNOpsStaticShapeTest` covering all three ops, plus a dynamic-channel negative case.

## Contributor Agreement

### DIFF
--- a/keras/src/backend/common/backend_utils.py
+++ b/keras/src/backend/common/backend_utils.py
@@ -307,12 +307,13 @@ def standardize_axis_for_numpy(axis):
     return tuple(axis) if isinstance(axis, list) else axis
 
 
-def check_depthwise_conv_input_channels(inputs, kernel, data_format):
-    """Validate a depthwise/separable conv input against its kernel shape.
+def check_conv_input_channels(inputs, kernel, data_format):
+    """Validate a conv input against its kernel shape.
 
-    Called from backend `depthwise_conv` / `separable_conv` after `inputs` has
-    been converted to a concrete tensor. Produces a clear error message before
-    the backend op raises its own implementation-specific one.
+    Used by `conv`, `depthwise_conv`, and `separable_conv` — all share the
+    convention that the kernel's input-channel dimension is `kernel.shape[-2]`.
+    Produces a clear error message before the backend op raises its own
+    implementation-specific one.
     """
     input_channels = (
         inputs.shape[-1] if data_format == "channels_last" else inputs.shape[1]
--- a/keras/src/backend/jax/nn.py
+++ b/keras/src/backend/jax/nn.py
@@ -16,9 +16,7 @@ from jax.experimental.pallas.ops.tpu.splash_attention import (
 )
 
 from keras.src import backend
-from keras.src.backend.common.backend_utils import (
-    check_depthwise_conv_input_channels,
-)
+from keras.src.backend.common.backend_utils import check_conv_input_channels
 from keras.src.backend.common.backend_utils import (
     compute_adaptive_pooling_window_sizes,
 )
@@ -788,7 +786,7 @@ def depthwise_conv(
     data_format = backend.standardize_data_format(data_format)
     inputs = convert_to_tensor(inputs)
     kernel = convert_to_tensor(kernel)
-    check_depthwise_conv_input_channels(inputs, kernel, data_format)
+    check_conv_input_channels(inputs, kernel, data_format)
     num_spatial_dims = inputs.ndim - 2
     dimension_numbers = _convert_to_lax_conv_dimension_numbers(
         num_spatial_dims,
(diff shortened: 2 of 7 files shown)


hertschuh (approved):
Thanks for the improvement!

PR #22861, opened 2026-05-11
Validate input channels in conv_transpose for symbolic and eager paths

### LINKED ISSUE #22857: [Bug] keras.ops.conv_transpose skips input-channel validation in symbolic path; eager raises a low-level backend error
### Summary

`keras.ops.conv_transpose` does not validate that the input tensor's channel dimension matches the kernel's input-channel dimension. The symbolic path silently returns a misleading output shape; the eager path eventually fails with an opaque, backend-specific error rather than a clear Keras-level `ValueError`.

This is the same family of bug as #22852 (which I filed for `conv` / `depthwise_conv` / `separable_conv`), but `conv_transpose` is missing because its kernel layout is different — input channels live at `kernel.shape[-1]`, not `kernel.shape[-2]`.

### Reproduction

```python
import os
import keras
import keras.ops as ops
import numpy as np

x = np.random.normal(size=(1, 10, 3)).astype("float32")        # 3 input channels
# kernel shape = (kernel_size, out_channels=4, in_channels=5) — mismatched
kernel = np.random.normal(size=(3, 4, 5)).astype("float32")

try:
    print("eager:", ops.conv_transpose(x, kernel, strides=1, padding="valid").shape)
except Exception as e:
    print("eager raises:", type(e).__name__, str(e)[:100])

print("symbolic:", ops.conv_transpose(keras.Input(shape=(10, 3)), kernel, strides=1,
                                      padding="valid").shape)
```

### Observed

```
eager raises: InvalidArgumentError {{function_node __wrapped__Conv2DBackpropInput_...}}: Conv2DCustomBackpropInput: filter and out_backprop must have the same out_depth
symbolic: (None, 12, 4)
```

### Expected

Both paths should raise the same clear `ValueError`:

```
ValueError: The number of input channels must match the kernel's input channels.
Received: input channels=3, kernel input channels=5, data_format='channels_last'.
```

### Root cause

`ConvTranspose.compute_output_spec` in `keras/src/ops/nn.py` reads `kernel.shape[-2]` (out_channels) and `kernel.shape[:-2]` (spatial) but never checks `inputs.shape[channel_dim]` against `kernel.shape[-1]` (input channels). The eager path relies on the backend op's own check, which surfaces as a confusing low-level error.

The fix mirrors the helper from #22687 (`check_depthwise_conv_input_channels`) but reads the kernel's input-channel index from `[-1]` to account for the transposed layout.

### Impact

Models using `keras.layers.Conv*Transpose` or `keras.ops.conv_transpose` with mismatched kernel/input channels are accepted at construction time and fail at execution with an unintelligible message. A clear early error is consistent with how non-transposed convs are validated.

rstar327:
## Description

Fixes #22857.

`keras.ops.conv_transpose` does not validate that the input tensor's channel dimension matches the kernel's input-channel dimension. The symbolic path silently fabricates a misleading output shape; the eager path eventually fails with an opaque backend-specific error (e.g. on TF: `Conv2DCustomBackpropInput: filter and out_backprop must have the same out_depth`).

This is the same family of bug as #22852, which I fixed for `conv` / `depthwise_conv` / `separable_conv` in #22853. `conv_transpose` was missing because its kernel layout differs — input channels live at `kernel.shape[-1]` instead of `kernel.shape[-2]`.

### Before

```python

x = np.random.normal(size=(1, 10, 3)).astype("float32")        # 3 in_channels
kernel = np.random.normal(size=(3, 4, 5)).astype("float32")    # in_channels=5

ops.conv_transpose(keras.Input(shape=(10, 3)), kernel, strides=1, padding="valid").shape
# -> (None, 12, 4)  silent

ops.conv_transpose(x, kernel, strides=1, padding="valid")
# -> InvalidArgumentError: filter and out_backprop must have the same out_depth
```

### After

```
ValueError: The number of input channels must match the kernel's input
channels. Received: input channels=3, kernel input channels=5,
data_format='channels_last'.
```

Raised by both the symbolic and eager paths, on all four backends (TF / JAX / Torch / NumPy). Dynamic channel dimensions (`None`) still pass through, matching the rest of the conv family.

### Implementation

- Added `check_conv_transpose_input_channels` helper in `keras/src/backend/common/backend_utils.py`. Mirrors `check_depthwise_conv_input_channels` from #22687, but reads input channels from `kernel.shape[-1]` to account for the transposed layout.
- Called the helper in `ConvTranspose.compute_output_spec` (`keras/src/ops/nn.py`) and in `conv_transpose` of all four backends after `convert_to_tensor`, after `data_format` is standardized.
- Added `test_conv_transpose_input_channel_validation` in `NNOpsStaticShapeTest` covering both the mismatch case and a dynamic-channel negative case.

## Contributor Agreement

### DIFF
--- a/keras/src/backend/common/backend_utils.py
+++ b/keras/src/backend/common/backend_utils.py
@@ -340,6 +340,30 @@ def check_depthwise_conv_input_channels(inputs, kernel, data_format):
         )
 
 
+def check_conv_transpose_input_channels(inputs, kernel, data_format):
+    """Validate a conv_transpose input against its kernel shape.
+
+    `conv_transpose` kernels use the layout
+    `(spatial..., out_channels, in_channels)`, so the input-channel dimension
+    is at `kernel.shape[-1]` (vs. `kernel.shape[-2]` for regular conv).
+    """
+    input_channels = (
+        inputs.shape[-1] if data_format == "channels_last" else inputs.shape[1]
+    )
+    kernel_input_channels = kernel.shape[-1]
+    if (
+        isinstance(input_channels, int)
+        and isinstance(kernel_input_channels, int)
+        and input_channels != kernel_input_channels
+    ):
+        raise ValueError(
+            "The number of input channels must match the kernel's input "
+            f"channels. Received: input channels={input_channels}, kernel "
+            f"input channels={kernel_input_channels}, "
+            f"data_format='{data_format}'."
+        )
+
+
 def to_tuple_or_list(value):
     """Convert the non-`None` value to either a tuple or a list."""
     if value is None:
--- a/keras/src/backend/jax/nn.py
+++ b/keras/src/backend/jax/nn.py
@@ -21,9 +21,6 @@ from keras.src.backend.common.backend_utils import (
     check_conv_transpose_input_channels,
 )
 from keras.src.backend.common.backend_utils import (
-    check_depthwise_conv_input_channels,
-)
-from keras.src.backend.common.backend_utils import (
     compute_adaptive_pooling_window_sizes,
 )
 from keras.src.backend.common.backend_utils import (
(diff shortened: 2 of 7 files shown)


hertschuh (approved):
Thanks for the improvement!

hertschuh (approved):

PR #22868, opened 2026-05-11
Validate newshape in reshape for consistent backend behavior

### LINKED ISSUE #22867: [Bug] keras.ops.reshape and keras.layers.Reshape accept invalid negative shape values (e.g. -2), with backend-dependent behavior
### Summary

`keras.ops.reshape` and `keras.layers.Reshape` document that only `-1` may be used as a wildcard dimension in the target shape, but neither validates this. The result depends on the backend:

| Backend     | `ops.reshape(x, (-2, 5))` on a 20-element tensor |
|-------------|--------------------------------------------------|
| NumPy       | succeeds, returns shape `(4, 5)` (treats `-2` the same as `-1`) |
| TensorFlow  | `InvalidArgumentError: Size 0 must be non-negative, not -2` |
| JAX         | `TypeError: cannot reshape array ... into shape (-2, 5) (size -10)` |
| Torch       | `RuntimeError: invalid shape dimension -2 at index 0 of shape [-2, 5]` |

Keras 3 is designed to behave consistently across backends. Three backends reject the input with three different error messages, while the fourth silently succeeds with a different result than the user likely intended.

### Reproduction

```python
import os
os.environ["KERAS_BACKEND"] = "numpy"  # try also "tensorflow", "jax", "torch"
import keras.ops as ops
import numpy as np

x = np.zeros(20, dtype="float32")
print(ops.reshape(x, (-2, 5)))   # backend-dependent
```

Same issue with the layer:

```python
import keras
keras.layers.Reshape(target_shape=(-2, 5))   # silently accepted at __init__
# call: produces "total size of the tensor must be unchanged" — the actual cause
# is that -2 is not a valid value for target_shape.
```

### Expected

All four backends should raise the same clear `ValueError` at the Keras level, e.g.:

```
ValueError: Each dimension in `newshape` must be a positive integer, or
`-1` for a single unknown dimension. Received: newshape=(-2, 5).
```

### Suggested fix

Validate `newshape` in `keras/src/ops/operation_utils.compute_reshape_output_shape` (already validates the \"more than one `-1`\" case) — reject any value `< -1`. The same helper is used by both `keras.ops.reshape` and `keras.layers.Reshape.compute_output_shape`, so a single change fixes both surfaces and ensures consistent behavior across all four backends.

### Impact

- Code ported from NumPy that accidentally uses `-2`/`-3` etc. silently \"works\" with the NumPy backend and breaks with cryptic backend-specific errors elsewhere.
- The `Reshape` layer's call-time error message obscures the real cause (invalid value in `target_shape`).

rstar327:
## Description

Fixes #22867.

`keras.ops.reshape` and `keras.layers.Reshape` document that only `-1` may be used as a wildcard dimension in the target shape, but neither validates this. The result depends on the backend:

| Backend | `ops.reshape(np.zeros(20), (-2, 5))` |
|---------|---------------------------------------|
| NumPy | silently returns shape `(4, 5)` |
| TensorFlow | `InvalidArgumentError: Size 0 must be non-negative, not -2` |
| JAX | `TypeError: cannot reshape array ... (size -10)` |
| Torch | `RuntimeError: invalid shape dimension -2` |

Keras 3 promises a consistent API across backends; three backends reject with three different error types while NumPy silently produces a different result than the documented contract.

### After this PR

All four backends raise the same clear `ValueError` at the Keras level:

```
ValueError: Each dimension in `newshape` must be a non-negative integer,
or `-1` for a single unknown dimension. Received: newshape=(-2, 5).
```

The same validation runs at `keras.layers.Reshape.__init__` so invalid `target_shape` is caught at construction time rather than at call time with a confusing \"total size of the tensor must be unchanged\" message.

### Implementation

- Extracted a new `validate_reshape_shape` helper in `keras/src/ops/operation_utils.py`. It rejects any concrete-int dimension `< -1` and requires at most one `-1`. Dynamic dimensions (backend tensor scalars resolved at runtime, used by `Reshape.call`) pass through unchecked.
- `compute_reshape_output_shape` now delegates its existing `-1`-count check to the helper, so the same rule is applied everywhere.
- `keras.ops.reshape` calls the helper before dispatching, so the eager path gets the same validation as the symbolic path.
- `keras.layers.Reshape.__init__` calls the helper, surfacing invalid `target_shape` at construction time instead of at call time.
- Added regression tests in `keras/src/ops/numpy_test.py` and `keras/src/layers/reshaping/reshape_test.py`.

## Contributor Agreement

### DIFF
--- a/keras/src/layers/reshaping/reshape.py
+++ b/keras/src/layers/reshaping/reshape.py
@@ -3,6 +3,7 @@ from keras.src.api_export import keras_export
 from keras.src.backend.common.keras_tensor import KerasTensor
 from keras.src.layers.layer import Layer
 from keras.src.ops import operation_utils
+from keras.src.ops.operation_utils import validate_reshape_shape
 
 
 @keras_export("keras.layers.Reshape")
@@ -37,12 +38,7 @@ class Reshape(Layer):
     def __init__(self, target_shape, **kwargs):
         super().__init__(**kwargs)
         target_shape = tuple(target_shape)
-        # test validity of target_shape
-        if target_shape.count(-1) > 1:
-            raise ValueError(
-                "The `target_shape` argument must not contain more than one "
-                f"`-1` value. Received: target_shape={target_shape}"
-            )
+        validate_reshape_shape(target_shape, newshape_arg_name="target_shape")
         self.target_shape = target_shape
         self.built = True
 
--- a/keras/src/layers/reshaping/reshape_test.py
+++ b/keras/src/layers/reshaping/reshape_test.py
@@ -95,6 +95,20 @@ class ReshapeTest(testing.TestCase):
         reshaped = layers.Reshape((8,))(input_layer)
         self.assertEqual(reshaped.shape, (None, 8))
 
+    def test_reshape_rejects_invalid_target_shape(self):
+        # Negative values other than -1 are invalid.
+        with self.assertRaisesRegex(
+            ValueError, "non-negative integer.*-1.*target_shape=\\(-2, 5\\)"
+        ):
+            layers.Reshape(target_shape=(-2, 5))
+
+        # Existing constraint: at most one -1.
+        with self.assertRaisesRegex(
+            ValueError,
+            "at most one unknown dimension.*target_shape=\\(-1, -1, 5\\)",
+        ):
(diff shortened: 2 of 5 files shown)


hertschuh (approved):
Thanks for the improvement!

PR #22873, opened 2026-05-11
MultiOptimizer support for model training #21741

gaga1313:
## Description

Added multi-optimizer support for training subnetworks with different optimizers.

Key Changes:

Implemented MultiOptimizer and OptimizerMap classes:
- OptimizerMap independently handles mapping between a variable identifier and an optimizer.
- MultiOptimizer distributes trainable and optimizer variables to their respective optimizers and performs build, apply, and stateless_apply operations.

Added unit tests for each component within MultiOptimizer:
- Rigorously tested variable mapping in OptimizerMap.
- Tested callback updates on learning rate and gradient updates performed via apply and stateless_apply.

## Contributor Agreement

### DIFF
--- a/keras/src/optimizers/multi_optimizer.py
+++ b/keras/src/optimizers/multi_optimizer.py
@@ -16,8 +16,7 @@ class OptimizerMap:
         optimizer: A list of optimizer instances.
         variable_identifier: A list of variable identifiers matching the
             optimizer list. Identifiers can be string regex patterns, Keras
-            Variables, list/tuple of Keras Variables, or callables returning a
-            boolean.
+            Variables, or list/tuple of Keras Variables.
     """
 
     def __init__(self, optimizer, variable_identifier):
@@ -53,9 +52,6 @@ class OptimizerMap:
         return None
 
     def _match(self, variable, identifier):
-        # If callable, evaluate it
-        if callable(identifier) and not isinstance(identifier, str):
-            return bool(identifier(variable))
         # If string, treat it as a regex pattern
         if isinstance(identifier, str):
             path = getattr(variable, "path", "") or ""
@@ -84,12 +80,11 @@ class MultiOptimizer(optimizer.Optimizer):
 
     # 2. Create mapping using different identifier types
     opt_map = keras.optimizers.OptimizerMap(
-        optimizer=[opt_adam, opt_sgd, opt_adam, opt_sgd, opt_adam],
+        optimizer=[opt_adam, opt_sgd, opt_adam, opt_sgd],
         variable_identifier=[
             "kernel",                            # variable name or path
             "^dense_1/.*",                       # regex pattern
             keras_var,                           # Keras Variable instance
-            lambda var: "bias" in var.name,      # custom callable function
             [var1, var2]                         # list/tuple
         ]
     )
@@ -130,7 +125,6 @@ class MultiOptimizer(optimizer.Optimizer):
         self.obj_map = obj_map
         self.default_optimizer = default_optimizer
 
-        # Disable loss scaling for all inner optimizers to avoid double scaling
         self.default_optimizer.loss_scale_factor = None
         for opt in self.obj_map.optimizer:
             opt.loss_scale_factor = None
@@ -225,6 +219,11 @@ class MultiOptimizer(optimizer.Optimizer):
             self.build(trainable_variables)
             self.built = True
 
+        # Statefully unscale gradients if loss_scale_factor is set
+        if self.loss_scale_factor is not None:
+            scale = self.loss_scale_factor
+            grads = [g if g is None else g / scale for g in grads]
+
         # Group gradients and variables by sub-optimizer
         optimizer_to_grads_and_vars = {
             opt: [] for opt in self._inner_optimizers
@@ -248,6 +247,11 @@ class MultiOptimizer(optimizer.Optimizer):
 
         if trainable_variables is None:
             trainable_variables = self._trainable_variables
+
+        # Statelessly unscale gradients if loss_scale_factor is set
+        if self.loss_scale_factor is not None:
+            scale = self.loss_scale_factor
+            grads = [g if g is None else g / scale for g in grads]
 
         own_var_count = len(self._variables)
         own_variables = optimizer_variables[:own_var_count]
@@ -366,10 +370,6 @@ class MultiOptimizer(optimizer.Optimizer):
                     serialized_identifiers.append(f"^({or_pattern})$")
                 else:
                     serialized_identifiers.append(identifier)
-            elif callable(identifier):
-                serialized_identifiers.append(
-                    serialization_lib.serialize_keras_object(identifier)
-                )
             else:
                 serialized_identifiers.append(identifier)
 
@@ -410,12 +410,7 @@ class MultiOptimizer(optimizer.Optimizer):
             for opt_config in serialized_opts
         ]
 
-        variable_identifiers = [
-            serialization_lib.deserialize_keras_object(
-                ident, custom_objects=custom_objects
-            )
-            for ident in obj_map_config["config"]["variable_identifier"]
-        ]
+        variable_identifiers = obj_map_config["config"]["variable_identifier"]
 
         map_cls = cls._get_map_class(
             obj_map_config["class_name"], custom_objects
--- a/keras/src/optimizers/multi_optimizer.py
+++ b/keras/src/optimizers/multi_optimizer.py
@@ -76,8 +76,7 @@ class MultiOptimizer(optimizer.Optimizer):
         default_optimizer: Default Keras Optimizer for any unmapped variables.
         name: String. The name of this optimizer.
     """
-
-    @tracking.no_automatic_dependency_tracking
+    
     def __init__(self, obj_map, default_optimizer, name=None, **kwargs):
         if not isinstance(obj_map, OptimizerMap):
             raise ValueError(
@@ -352,8 +351,6 @@ class MultiOptimizer(optimizer.Optimizer):
     def _get_map_class(cls, class_name, custom_objects=None):
         if custom_objects and class_name in custom_objects:
             return custom_objects[class_name]
-        if class_name == "OptimizerMap":
-            return OptimizerMap
         globals_dict = globals()
         if class_name in globals_dict:
             return globals_dict[class_name]
--- a/keras/src/optimizers/multi_optimizer.py
+++ b/keras/src/optimizers/multi_optimizer.py
@@ -2,6 +2,7 @@ import re
 
 from keras.src import ops
 from keras.src.api_export import keras_export
+from keras.src.backend.common.variables import Variable
 from keras.src.optimizers import optimizer
 from keras.src.saving import serialization_lib
 from keras.src.utils import tracking
@@ -85,7 +86,7 @@ class MultiOptimizer(optimizer.Optimizer):
     opt_map = keras.optimizers.OptimizerMap(
         optimizer=[opt_adam, opt_sgd, opt_adam, opt_sgd, opt_adam],
         variable_identifier=[
-            "kernel",                            # name substring
+            "kernel",                            # variable name or path
             "^dense_1/.*",                       # regex pattern
             keras_var,                           # Keras Variable instance
             lambda var: "bias" in var.name,      # custom callable function
@@ -210,9 +211,8 @@ class MultiOptimizer(optimizer.Optimizer):
 
     @property
     def iterations(self):
-        if self._inner_optimizers:
-            return self._inner_optimizers[0].iterations
-        return super().iterations
+        # master and inner optimizer iterations will be in sync.
+        return self._iterations
 
     def apply(self, grads, trainable_variables=None):
         if len(grads) == 0:
@@ -330,10 +330,12 @@ class MultiOptimizer(optimizer.Optimizer):
                 "optimizer weights before calling `set_weights()`."
             )
 
-        # Distribute weights back to wrapper and sub-optimizers
-        self._iterations.assign(weights[0])
+        # Distribute own variables dynamically (iterations, learning_rate, etc.)
+        own_var_count = len(self._variables)
+        for i in range(own_var_count):
+            self._variables[i].assign(weights[i])
 
-        idx = 1
+        idx = own_var_count
         for opt in self._inner_optimizers:
             num_opt_vars = len(opt.variables)
             if num_opt_vars > 0:
@@ -343,6 +345,34 @@ class MultiOptimizer(optimizer.Optimizer):
 
     def get_config(self):
         config = super().get_config()
+
+        serialized_identifiers = []
+        for identifier in self.obj_map.variable_identifier:
+            if isinstance(identifier, Variable):
+                serialized_identifiers.append(
+                    f"^{re.escape(identifier.path or identifier.name)}$"
+                )
+            elif isinstance(identifier, (list, tuple, set)):
+                if any(isinstance(item, Variable) for item in identifier):
+                    escaped_paths = []
+                    for item in identifier:
+                        if isinstance(item, Variable):
+                            escaped_paths.append(
+                                re.escape(item.path or item.name)
+                            )
+                        else:
+                            escaped_paths.append(re.escape(str(item)))
+                    or_pattern = "|".join(escaped_paths)
+                    serialized_identifiers.append(f"^({or_pattern})$")
+                else:
+                    serialized_identifiers.append(identifier)
+            elif callable(identifier):
+                serialized_identifiers.append(
+                    serialization_lib.serialize_keras_object(identifier)
+                )
+            else:
+                serialized_identifiers.append(identifier)
+
         serialized_map = {
             "class_name": self.obj_map.__class__.__name__,
             "config": {
@@ -350,7 +380,7 @@ class MultiOptimizer(optimizer.Optimizer):
                     serialization_lib.serialize_keras_object(opt)
                     for opt in self.obj_map.optimizer
                 ],
-                "variable_identifier": self.obj_map.variable_identifier,
+                "variable_identifier": serialized_identifiers,
             },
         }
         config.update(
@@ -379,7 +409,13 @@ class MultiOptimizer(optimizer.Optimizer):
             )
             for opt_config in serialized_opts
         ]
-        variable_identifiers = obj_map_config["config"]["variable_identifier"]
+
+        variable_identifiers = [
+            serialization_lib.deserialize_keras_object(
+                ident, custom_objects=custom_objects
+            )
+            for ident in obj_map_config["config"]["variable_identifier"]
+        ]
 
         map_cls = cls._get_map_class(
             obj_map_config["class_name"], custom_objects
--- a/keras/src/optimizers/multi_optimizer.py
+++ b/keras/src/optimizers/multi_optimizer.py
@@ -71,10 +71,41 @@ class OptimizerMap:
 class MultiOptimizer(optimizer.Optimizer):
     """An optimizer wrapper that delegates variables to different optimizers.
 
+    Example:
+
+    ```python
+    # 1. Define your sub-optimizers
+    opt_adam = keras.optimizers.Adam(learning_rate=1e-3)
+    opt_sgd = keras.optimizers.SGD(learning_rate=1e-2)
+    default_opt = keras.optimizers.RMSprop(learning_rate=1e-4)
+
+    # 2. Create mapping using different identifier types
+    opt_map = keras.optimizers.OptimizerMap(
+        optimizer=[opt_adam, opt_sgd, opt_adam, opt_sgd, opt_adam],
+        variable_identifier=[
+            "kernel",                            # 1. Match variable name substring
+            "^dense_1/.*",                       # 2. Match variable path regex pattern
+            keras_var,                           # 3. Match exact Keras Variable instance
+            lambda var: "bias" in var.name,      # 4. Match via custom callable function
+            [var1, var2]                         # 5. Match if variable is in list/tuple
+        ]
+    )
+
+    # 3. Wrap and compile
+    multi_opt = keras.optimizers.MultiOptimizer(
+        obj_map=opt_map,
+        default_optimizer=default_opt
+    )
+    model.compile(optimizer=multi_opt, loss="mse")
+    ```
+
+    Note: You can subclass `OptimizerMap` to define custom routing logic by
+    overriding its `__call__(self, variable)` method.
+
     Args:
-        obj_map: An OptimizerMap instance.
-        default_optimizer: Default Keras Optimizer for any unmapped variables.
-        name: String. The name of this optimizer.
+        obj_map: An `OptimizerMap` instance containing variable-to-optimizer rules.
+        default_optimizer: Default Keras `Optimizer` for any unmapped variables.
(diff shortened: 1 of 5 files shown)


keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,403 @@
...
+        if self._inner_optimizers:
+            self._inner_optimizers[0].learning_rate = value
+        else:
+            super().learning_rate = value
+
+    @property
+    def iterations(self):
+        if self._inner_optimizers:
+            return self._inner_optimizers[0].iterations
+        return super().iterations
```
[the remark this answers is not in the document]
gaga1313: Its a nitpick: they both should be in sync. But I have updated the code for reducing number of lines.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,403 @@
...
+    def get_config(self):
+        config = super().get_config()
+        serialized_map = {
+            "class_name": self.obj_map.__class__.__name__,
+            "config": {
+                "optimizer": [
+                    serialization_lib.serialize_keras_object(opt)
+                    for opt in self.obj_map.optimizer
+                ],
+                "variable_identifier": self.obj_map.variable_identifier,
```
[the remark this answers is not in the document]
gaga1313: resolved, update the serialization logic for keras variables and callable identifiers.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,403 @@
...
+        # Distribute weights back to wrapper and sub-optimizers
+        self._iterations.assign(weights[0])
+
+        idx = 1
+        for opt in self._inner_optimizers:
+            num_opt_vars = len(opt.variables)
+            if num_opt_vars > 0:
+                opt_weights = weights[idx : idx + num_opt_vars]
+                opt.set_weights(opt_weights)
+                idx += num_opt_vars
```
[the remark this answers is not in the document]
gaga1313: Updated! the idx will be based on len(self._variables)

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,403 @@
...
+        if self._inner_optimizers:
+            return self._inner_optimizers[0].learning_rate
+        return super().learning_rate
+
+    @learning_rate.setter
+    def learning_rate(self, value):
+        if self._inner_optimizers:
+            self._inner_optimizers[0].learning_rate = value
+        else:
+            super().learning_rate = value
```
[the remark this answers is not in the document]
gaga1313: The underlying implementation is abstract. The setter/getter updates the first optimizer learning rate parameter. For user it is not ambiguous.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,403 @@
...
+    def _match(self, variable, identifier):
+        # If callable, evaluate it
+        if callable(identifier) and not isinstance(identifier, str):
+            return bool(identifier(variable))
+        # If string, treat it as a regex pattern
+        if isinstance(identifier, str):
+            path = getattr(variable, "path", "") or ""
+            name = getattr(variable, "name", "") or ""
+            return bool(
+                re.match(identifier, path) or re.match(identifier, name)
```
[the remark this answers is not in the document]
gaga1313: Update the doc-string, previously it mentioned substring works. But we only support the exact name or path.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,403 @@
...
+            optimizer_to_grads_and_vars[opt].append((grad, var))
+
+        # Dispatch apply calls
+        for opt, grads_and_vars in optimizer_to_grads_and_vars.items():
+            if grads_and_vars:
+                sub_grads, sub_vars = zip(*grads_and_vars)
+                opt.apply(list(sub_grads), list(sub_vars))
+
+        # Keep master iterations in sync
+        self._iterations.assign_add(1)
```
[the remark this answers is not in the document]
gaga1313: Fix by scaling gradient in apply and stateless_apply method

gaga1313 (commented):
Answered comments

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+        return None
+
+    def _match(self, variable, identifier):
+        # If string, treat it as a regex pattern
+        if isinstance(identifier, str):
+            path = getattr(variable, "path", "") or ""
+            name = getattr(variable, "name", "") or ""
+            return bool(
+                re.match(identifier, path) or re.match(identifier, name)
+            )
```
hertschuh, line 61: For consistency with `LayoutMap` and `DTypePolicyMap` let's:
- only use the `path`
- do `re.fullmatch`

Without these, it's too easy to make mistakes

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+        matched_optimizers = []
+        for opt, identifier in zip(self.optimizer, self.variable_identifier):
+            if self._match(variable, identifier):
+                if opt not in matched_optimizers:
+                    matched_optimizers.append(opt)
+
+        if len(matched_optimizers) > 1:
+            raise ValueError(
+                f"Multiple optimizers assigned to variable "
+                f"{variable.path or variable.name}: {matched_optimizers}"
```
hertschuh, line 48: Only show the path.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+
+@keras_export("keras.optimizers.OptimizerMap")
+class OptimizerMap:
+    """A class mapping variables to optimizers.
+
+    Args:
+        optimizer: A list of optimizer instances.
+        variable_identifier: A list of variable identifiers matching the
+            length of the optimizer list. Identifiers can be string regex
+            patterns, Keras Variables, or list/tuple of Keras Variables.
```
hertschuh, line 19: Ignore based on the overall review comment

----

These should be `optimizers` and `variable_identifiers` plural to make it clear they are lists.

I'm also not sure about:
- this approach vs. pairs
- the order of `optimizers` and `variable_identifiers` vs, `variable_identifiers` and `optimizers` because a map normally goes **from** the key **to** the value.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+    """
+
+    def __init__(self, optimizer, variable_identifier):
+        self.optimizer = optimizer
+        self.variable_identifier = variable_identifier
```
hertschuh, line 24: Verify they have the same length

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+
+    def __call__(self, variable):
+        """Evaluates a single variable and returns its assigned optimizer.
+
```
hertschuh, line 28: Also mention this can be overridden to customize the behavior.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+    # 1. Define your sub-optimizers
+    opt_adam = keras.optimizers.Adam(learning_rate=1e-3)
+    opt_sgd = keras.optimizers.SGD(learning_rate=1e-2)
+    default_opt = keras.optimizers.RMSprop(learning_rate=1e-4)
+
+    # 2. Create mapping using different identifier types
+    opt_map = keras.optimizers.OptimizerMap(
+        optimizer=[opt_adam, opt_sgd, opt_adam, opt_sgd],
+        variable_identifier=[
+            "kernel",                            # variable name or path
```
hertschuh, line 86: You need something more, like `".*/kernel"` since we do a full path match.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+            "^dense_1/.*",                       # regex pattern
```
hertschuh, line 87: Remove the `^` since we do a full path match.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+            )
+
+        super().__init__(learning_rate=0.0, name=name)
+
+        self.obj_map = obj_map
+        self.default_optimizer = default_optimizer
+
+        self.default_optimizer.loss_scale_factor = None
+        for opt in self.obj_map.optimizer:
+            opt.loss_scale_factor = None
```
hertschuh, line 131: Why did you need to do this?

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+        for opt in self.obj_map.optimizer:
+            if opt not in unique_opts:
+                unique_opts.append(opt)
+        if self.default_optimizer not in unique_opts:
+            unique_opts.append(self.default_optimizer)
+        else:
+            unique_opts.remove(self.default_optimizer)
+            unique_opts.append(self.default_optimizer)
+
+        self._inner_optimizers = unique_opts
```
hertschuh, line 144: Let's not do it this way. If somebody subclasses `OptimizerMap` to do something completely different, it might not work.

Instead, collect that list within `build`.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+
+    def _var_key(self, variable):
+        return id(variable)
```
hertschuh, line 147: Remove. This will break on Tensorflow I believe.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+
+    @tracking.no_automatic_dependency_tracking
+    def build(self, var_list):
+        if self.built:
+            return
+
+        self._var_to_optimizer_idx = {}
+        self._optimizer_to_vars = {opt: [] for opt in self._inner_optimizers}
```
hertschuh, line 155: Neither optimizers nor variables are guaranteed to be hashable, let's use ids instead of actual instances as keys to dicts.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+
+        for var in var_list:
+            opt = self.obj_map(var)
+            if opt is None:
+                opt = self.default_optimizer
+
+            idx = self._inner_optimizers.index(opt)
```
hertschuh, line 162: Build the list of optimizers as you go and discover them from `obj_map.__call__`. The benefits are:
- the only contract between `MultiOptimizer` and `OptimizerMap` is `__call__`, people are free to do whatever they want in their subclasses, in fact we could allow plain `callable`s in addition to `OptimizerMap`
- you will collect only used optimizers and not unused ones

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+
+    def _get_optimizer_for_variable(self, variable):
+        idx = self._var_to_optimizer_idx.get(self._var_key(variable), None)
+        if idx is None:
+            raise ValueError(f"Variable {variable} not found in any optimizer.")
+        return self._inner_optimizers[idx]
+
+    def get_optimizer(self, index):
+        return self._inner_optimizers[index]
+
```
hertschuh, line 181: Remove. I don't see a need for this and we would have to clearly defined the contract (which order, are unused optimizers skipped, etc.)

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+    @property
+    def num_optimizers(self):
+        return len(self._inner_optimizers)
+
```
hertschuh, line 185: Remove I don't see a need for this.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+        vars = self._variables[:]
+        for opt in self._inner_optimizers:
+            vars.extend(opt.variables)
+        return vars
+
+    @property
+    def learning_rate(self):
+        if self._inner_optimizers:
+            return self._inner_optimizers[0].learning_rate
+        return super().learning_rate
```
hertschuh, line 198: I think the `default_optimizer` is a better choice than the 0th optimizer.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+
+    @learning_rate.setter
+    def learning_rate(self, value):
+        if self._inner_optimizers:
+            self._inner_optimizers[0].learning_rate = value
+        else:
+            super().learning_rate = value
```
hertschuh, line 205: I think the `default_optimizer` is a better choice than the 0th optimizer.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+
+    @property
+    def iterations(self):
+        # master and inner optimizer iterations will be in sync.
+        return self._iterations
```
hertschuh, line 210: Remove, the super class does this (more) correctly.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+
+        # Statefully unscale gradients if loss_scale_factor is set
+        if self.loss_scale_factor is not None:
+            scale = self.loss_scale_factor
+            grads = [g if g is None else g / scale for g in grads]
+
+        # Group gradients and variables by sub-optimizer
+        optimizer_to_grads_and_vars = {
+            opt: [] for opt in self._inner_optimizers
+        }
```
hertschuh, line 231: Since we don't want to key by optimizer and what you store is the optimizer index, you might as well have this map be from optimizer index to list of vars.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+        own_variables = optimizer_variables[:own_var_count]
+        remaining_opt_vars = optimizer_variables[own_var_count:]
+
+        inner_opt_variables = []
+        offset = 0
+        for opt in self._inner_optimizers:
+            opt_var_count = len(opt.variables)
+            opt_vars = remaining_opt_vars[offset : offset + opt_var_count]
+            inner_opt_variables.append(opt_vars)
+            offset += opt_var_count
```
hertschuh, line 267: Why did you use a completely different approach `stateless_apply` and `apply` (lines 229-234)?

I don't mind either way, but stay consistent. I kind of like this approach better because then you can remove `self._var_to_optimizer_idx` and `_get_optimizer_for_variable` completely.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+
+        optimizer_grads = [[] for _ in self._inner_optimizers]
+        optimizer_train_vars = [[] for _ in self._inner_optimizers]
+        optimizer_train_var_indices = [[] for _ in self._inner_optimizers]
+
+        for i, (grad, var) in enumerate(zip(grads, trainable_variables)):
+            # Map the incoming stateless tracer array
+            # to the static Keras Variable via index alignment
+            keras_var = self._trainable_variables[i]
```
hertschuh, line 276: `keras_var = self._trainable_variables[i]`

This is `var`, remove this line.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+            opt = self._get_optimizer_for_variable(keras_var)
+            opt_idx = self._inner_optimizers.index(opt)
```
hertschuh, line 278: This is weird, you're doing too much here. Instead have a way to get the optimizer index for each variable.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+                new_inner_opt_variables.extend(updated_opt_vars)
+            else:
+                new_inner_opt_variables.extend(opt_vars)
+
+        # Statelessly increment the master iterations counter
+        new_own_variables = list(own_variables)
+        if new_own_variables:
+            new_own_variables[0] = ops.cast(
+                new_own_variables[0] + 1, new_own_variables[0].dtype
+            )
```
hertschuh, line 309: This is making the assumption that variable 0 is the iteration. But there could be more variables (in a subclass for instance).

Instead use this:
```python
        with backend.StatelessScope(state_mapping=list(zip(self._variables, own_variables))) as scope:
            self._iterations.assign_add(1)
```

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+        for i, (grad, var) in enumerate(zip(grads, trainable_variables)):
+            # Map the incoming stateless tracer array
+            # to the static Keras Variable via index alignment
+            keras_var = self._trainable_variables[i]
+            opt = self._get_optimizer_for_variable(keras_var)
+            opt_idx = self._inner_optimizers.index(opt)
+
+            optimizer_grads[opt_idx].append(grad)
+            optimizer_train_vars[opt_idx].append(var)
+            optimizer_train_var_indices[opt_idx].append(i)
```
hertschuh, line 282: It seems like this could be simplified. You're sort of computing 2 different mappings.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+        # Dispatch apply calls
+        for opt, grads_and_vars in optimizer_to_grads_and_vars.items():
+            if grads_and_vars:
+                sub_grads, sub_vars = zip(*grads_and_vars)
+                opt.apply(list(sub_grads), list(sub_vars))
+
+        # Keep master iterations in sync
+        self._iterations.assign_add(1)
+
+    def stateless_apply(self, optimizer_variables, grads, trainable_variables):
```
hertschuh, line 245: Do you even need this? Doesn't the default implementation of `stateless_apply` which calls `apply` just work?

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+        if self.loss_scale_factor is not None:
+            return loss * self.loss_scale_factor
+        return loss
+
+    def finalize_variable_values(self, var_list):
+        optimizer_to_vars = {opt: [] for opt in self._inner_optimizers}
+        for var in var_list:
+            opt = self._get_optimizer_for_variable(var)
+            optimizer_to_vars[opt].append(var)
+
```
hertschuh, line 324: Now that I'm seeing this repeating multiple times, I think it's clear that the mapping you need is from  variable to optimizer index.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+        serialized_map = {
+            "class_name": self.obj_map.__class__.__name__,
+            "config": {
+                "optimizer": [
+                    serialization_lib.serialize_keras_object(opt)
+                    for opt in self.obj_map.optimizer
+                ],
+                "variable_identifier": serialized_identifiers,
+            },
+        }
```
hertschuh, line 386: Instead of this, simply do:
```python
serialization_lib.serialize_keras_object(self.obj_map)
```
Then implement `get_config` and `from_config` in `OptimizerMap`.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+                            escaped_paths.append(re.escape(str(item)))
+                    or_pattern = "|".join(escaped_paths)
+                    serialized_identifiers.append(f"^({or_pattern})$")
+                else:
+                    serialized_identifiers.append(identifier)
+            else:
+                serialized_identifiers.append(identifier)
+
+        serialized_map = {
+            "class_name": self.obj_map.__class__.__name__,
```
hertschuh, line 378: You should never have to do `"class_name": self.obj_map.__class__.__name__,` yourself, this should always be handled by `serialization_lib.serialize_keras_object`.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+    def _get_map_class(cls, class_name, custom_objects=None):
+        if custom_objects and class_name in custom_objects:
+            return custom_objects[class_name]
+        globals_dict = globals()
+        if class_name in globals_dict:
+            return globals_dict[class_name]
+        raise ValueError(
+            f"Could not find class {class_name} for "
+            "OptimizerMap deserialization."
+        )
```
hertschuh, line 435: You should never have to do this, this should come from `serialization_lib.deserialize_keras_object`.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+                    f"^{re.escape(identifier.path or identifier.name)}$"
+                )
+            elif isinstance(identifier, (list, tuple, set)):
+                if any(isinstance(item, Variable) for item in identifier):
+                    escaped_paths = []
+                    for item in identifier:
+                        if isinstance(item, Variable):
+                            escaped_paths.append(
+                                re.escape(item.path or item.name)
+                            )
```
hertschuh, line 367: Hmm, but this is cheating and not idempotent. What you get back is not the original. You had a Variable, but when you deserialize you get a string.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,435 @@
...
+                        else:
+                            escaped_paths.append(re.escape(str(item)))
+                    or_pattern = "|".join(escaped_paths)
+                    serialized_identifiers.append(f"^({or_pattern})$")
```
hertschuh, line 371: This is also not idempotent. You had a list of regexes coming in, but after deserialization you get a single regex.

hertschuh (commented):
After going through the whole thing and seeing the serialization aspect, I think we should simplify and align more with the way it's done in `LayoutMap` and `DTypePolicyMap`, i.e. make it a map from string (the regex) to the optimizer instance.

It's simpler, more consistent, and solves the serialization / deserialization problem.

Also, you don't need lists, you can have the same optimizer as value multiple times, or combine the regexes in a single one.

Rough draft:

```python
class OptimizerMap(MutableMapping):

    def __getitem__(self, key):
        if key in self._optimizer_map:
            return self._optimizer_map[key]

        matching_keys = [
            pattern
            for pattern in self._optimizer_map
            if re.fullmatch(pattern, key)
        ]
        if len(matching_keys) > 1:
            raise ValueError(...)
        elif len(matching_keys) == 1:
            return self._optimizer_map[matching_keys[0]]
        return None

    def __setitem__(self, key, optimizer):
        # Validation here...
        self._optimizer_map[key] = optimizer
```

Now, to support `DistributedEmbedding` and other custom use cases, I think we should also make it callable. The Map aspect would work on variable paths (like `LayoutMap` and `DTypePolicyMap`) but the `__call__` method could take the raw variable, like:

```python
    def __call__(self, variable):
         return self[variable.path]
```

Then, the `MultiOptimizer` could take any `callable` instead of `OptimizerMap`.
But let's table that for now (adding `__call__` and supporting `callable`).

The other question is whether the default optimizer should be part of the `OptimizerMap` or the `MultiOptimizer`. I'm not fully sure...

PR #22874, opened 2026-05-11
Fix torch BiLSTM dispatch and engage cuDNN

MarcosAsh:
## Description

Fixes a silent cuDNN dispatch failure in the torch LSTM backend (which I introduced) and adds a fused bidirectional cuDNN path mirroring #22791 (JAX equivalent).

### The bug

`torch._VF.lstm` returns a flat `(output, h_n, c_n)` tuple. In #22470 (March 2026, "Rewrite torch LSTM to use functional API with CPU fallback") I unpacked it as the nested `outputs, (h_n, c_n)` form, which raises `ValueError: too many values to unpack` at runtime. The exception is caught by the `try: ... except Exception: pass` in the surrounding `lstm()`, silently falling back to `_fallback_lstm` (the pure-Python per-timestep loop). CPU tests pass because `_fallback_lstm` is functionally correct; CPU never exercises the cuDNN path; no GPU CI checks LSTM perf, so the regression went unnoticed.

Net effect: every keras-on-torch user with `LSTM(use_cudnn='auto')` on GPU has been running through the slow fallback since #22470 merged. This PR restores the cuDNN path and also adds the bidirectional fused path that was previously a `NotImplementedError` stub.

### Changes

- `_cudnn_lstm` (unidirectional, `rnn.py:651`): fix unpacking to `outputs, h_n, c_n = torch._VF.lstm(...)`.
- `bidirectional_lstm`: implement the fused path using `torch._VF.lstm(..., bidirectional=True)`, mirroring #22791's JAX design.

### Benchmark (NVIDIA A100-SXM4-40GB, torch 2.10.0+cu128, B=32, L=128, H=256, 2 BiLSTM layers, `train_on_batch`)

| | master | this PR | reduction |
|---|---|---|---|
| BiLSTM warm step | 391 ms | 119.6 ms | **3.27x faster** |
| cudaLaunchKernel / step | 16,459 | 2,078 | **8x fewer launches** |

The kernel-launch drop is mechanical evidence that cuDNN engages: master was unrolling the per-timestep Python loop in `_fallback_lstm`; this PR routes through cuDNN's fused kernel. Confirmed cuDNN engagement on GPU via direct invocation of `Bidirectional._call_fused_lstm` and inspection of the installed source.

## Contributor Agreement

### DIFF
--- a/keras/src/backend/torch/rnn.py
+++ b/keras/src/backend/torch/rnn.py
@@ -897,6 +897,39 @@ def bidirectional_lstm(
     ``torch._VF.lstm(..., bidirectional=True)`` call instead of dispatching
     two unidirectional LSTM calls. Backward outputs are returned in original
     time order, ready for the caller's ``merge_mode`` to consume directly.
+
+    Args:
+        inputs: Input tensor of shape ``(batch, time, features)``.
+        fwd_initial_state_h: Initial hidden state for the forward direction,
+            shape ``(batch, hidden)``.
+        fwd_initial_state_c: Initial cell state for the forward direction,
+            shape ``(batch, hidden)``.
+        bwd_initial_state_h: Initial hidden state for the backward direction,
+            shape ``(batch, hidden)``.
+        bwd_initial_state_c: Initial cell state for the backward direction,
+            shape ``(batch, hidden)``.
+        mask: Sequence mask. Only ``None`` is supported; otherwise
+            ``NotImplementedError`` is raised so the caller can fall back to
+            the two-pass path.
+        fwd_kernel: Forward input kernel, shape ``(features, 4 * hidden)``.
+        fwd_recurrent_kernel: Forward recurrent kernel, shape
+            ``(hidden, 4 * hidden)``.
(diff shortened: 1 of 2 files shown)


hertschuh:
Please look at the error on torch.

MarcosAsh:
Hi, thanks for the review! I fixed the error with the torch test.

hertschuh:
So one test is failing on Torch:
```
FAILED keras/src/export/onnx_test.py::ExportONNXTest::test_standard_model_export_lstm - AssertionError: 
Not equal to tolerance rtol=1e-06, atol=1e-06
...
Max absolute difference among violations: 2.771616e-06
Max relative difference among violations: 0.00093703
```
This test is using BidiLSTM.

But feel free to adjust the atol and rtol on that assert.

Ignore the argpartition failure, that's fixed in master, so you can rebase to get the fix.

MarcosAsh:
HI thanks for the help! I just updated the tolerances and rebased.

hertschuh (approved):
Thanks for adding this!

PR #22362, opened 2026-03-05
Fix #20598 — Custom models/layers with sublayers in nested lists fail to save/load weights

pctablet505:
**Fixes**: https://github.com/keras-team/keras/issues/20598  

This pull request enhances the Keras model saving and loading functionality to better support models with nested containers (lists, tuples, and dicts) of layers. It also introduces comprehensive tests to ensure that models with such nested structures are correctly serialized and deserialized, preserving all weights and structure.

### Improvements to nested container support

* Updated the core saving and loading logic in `saving_lib.py` to recursively handle nested containers of type `list`, `dict`, and `tuple`, ensuring that all contained layers are processed correctly. Previously, `set` was also included, but this has been removed for consistency and correctness. [[1]](diffhunk://#diff-17ef13a0a19d21f71b9406af463432db195ec61272ba7042c98ef1df933855ddL754-R754) [[2]](diffhunk://#diff-17ef13a0a19d21f71b9406af463432db195ec61272ba7042c98ef1df933855ddL842-R842)
* Enhanced `_save_container_state` and `_load_container_state` to handle nested containers, assigning unique names to each container to avoid naming collisions during the recursive process. [[1]](diffhunk://#diff-17ef13a0a19d21f71b9406af463432db195ec61272ba7042c98ef1df933855ddR896-R909) [[2]](diffhunk://#diff-17ef13a0a19d21f71b9406af463432db195ec61272ba7042c98ef1df933855ddR946-R962)

### Expanded test coverage

* Added new tests in `saving_lib_test.py` to verify saving and loading of models with nested lists, tuples, and dicts of layers. These tests ensure that the structure and weights of all nested layers are preserved after serialization and deserialization.

These changes improve the robustness and reliability of Keras model serialization for complex model architectures containing deeply nested containers of layers.

## Contributor Agreement
**Please check all boxes below before submitting your PR for review:**

- [x] I am a human, and not a bot.
- [x] I will be responsible for responding to review comments in a timely manner.
- [x] I will work with the maintainers to push this PR forward until submission.

> **Note:** Failing to adhere to this agreement may result in your future PRs no longer being reviewed.

PR #22891, opened 2026-05-13
Fixes backward-compatiblity loading of legacy  files after #22362

JyotinderSingh:
## Description

Fixes the regression reported after #22362 where existing `.keras` files saved by older Keras versions fail to load with:

```text
ValueError: A total of N objects could not be loaded.
Layer 'dense_proj_0' expected 1 variables, but received 0 variables during
loading. Expected: ['kernel']
```

## Root cause

#22362 changed `_save_container_state` / `_load_container_state` in `keras/src/saving/saving_lib.py` to recursively descend into nested containers (e.g. `self.blocks = [(Dense, Dense), ...]`): an `elif isinstance(saveable, (list, dict, tuple))` branch was added inside both functions so sublayers inside nested containers now get their own `container/container_1/...` H5 path namespace.

The regression hits a common pattern, especially in KerasHub-style backbones: a Model rebuilds a list of sublayers inside `__init__` from hyperparameters, and stores those sublayers in a nested container (tuple-in-list, dict-in-list, list-in-list, etc.). With the pre-PR saver, the inner containers were silently skipped, but the sublayers were still reachable via the Model's flat `.layers` walk, so they got saved at `/layers/dense_X` and reloaded from there. With the post-PR loader, the same model now recurses into the nested containers first, looks for `/blocks/container/dense`, doesn't find it in the legacy file, and raises before the `/layers/...` fallback path has a chance to load them. The Dense layers end up in `failed_saveables`, and `load_model` aggregates them into the reported `ValueError`.

The fix added in PR #22362 itself is correct (recursing into nested containers is the right thing to do for newly-saved files). The regression is purely about loading old files written by code that did not produce the new `container*` HDF5 groups.

## Fix

Added a presence check before recursing into nested containers on load. Each I/O store (`H5IOStore`, `ShardedH5IOStore`, `NpzIOStore`, `DiskIOStore`) now exposes a `has_path(path) -> bool` method that answers "do you contain anything under this path?" using its own native lookup (H5 group membership, sharded weight-map, npz key set, on-disk path existence). A small helper `_container_path_present(weights_store, assets_store, path)` consults both stores polymorphically.

In the `elif isinstance(saveable, (list, dict, tuple))` branch of `_load_container_state`, if the constructed nested path is absent in both stores we emit a `UserWarning` and `continue` instead of recursing. That precisely matches the pre-PR semantics (the inner container was a no-op) for legacy files, while leaving the recursion intact for files actually saved by post-PR code.

## Testing

`keras/src/saving/saving_lib_test.py`:

- `test_legacy_load_skips_missing_nested_containers`: loads a real pre-PR `.keras` fixture (`test_data/pre_pr22362_nested_container.keras`, generated from PR #22362's parent commit `d5a88bdb1`). Asserts `load_model` succeeds, emits the new `"Skipping nested container"` warning, and restores the per-Dense kernel values the fixture was saved with.

## Checked-in binary fixture: provenance and integrity

The regression test loads a real `.keras` file checked in at `keras/src/saving/test_data/pre_pr22362_nested_container.keras`.

**SHA-256**

```text
e32c92a345632504f3d21acae9a4bff0363f5e787c76f244e1c455a55c08e27c
```

**Verification**

```bash
shasum -a 256 keras/src/saving/test_data/pre_pr22362_nested_container.keras
```

**Regen script.** Pins to PR #22362's parent commit (`d5a88bdb1`) so
the on-disk layout is exactly the pre-PR shape:

```bash
git worktree add --detach /tmp/keras-pre-pr d5a88bdb1
PYTHONPATH=/tmp/keras-pre-pr python <<'PY'
import keras, numpy as np

@keras.saving.register_keras_serializable(package="legacy_nested_container")
class NestedDenseModel(keras.Model):
    def __init__(self, num=3, **kwargs):
        super().__init__(**kwargs)
        self.num = num
        self.blocks = [
            (keras.layers.Dense(4, use_bias=False, name=f"d_{i}"),)
            for i in range(num)
        ]
        self.out = keras.layers.Dense(2)
    def call(self, x):
        for (d,) in self.blocks:
            x = d(x)
        return self.out(x)
    def get_config(self):
        cfg = super().get_config(); cfg["num"] = self.num; return cfg

m = NestedDenseModel(num=3)
m(np.zeros((2, 4), dtype="float32"))
for i, (d,) in enumerate(m.blocks):
    d.kernel.assign(np.full_like(d.kernel, 0.1 * (i + 1)))
m.out.kernel.assign(np.full_like(m.out.kernel, 0.5))
m.out.bias.assign(np.full_like(m.out.bias, 0.0))
m.save("keras/src/saving/test_data/pre_pr22362_nested_container.keras")
PY
git worktree remove /tmp/keras-pre-pr
shasum -a 256 keras/src/saving/test_data/pre_pr22362_nested_container.keras
# Update the SHA-256 above and re-run the test suite.
```

## Contributor Agreement


keras/src/saving/saving_lib.py
```diff
@@ -1023,6 +1053,14 @@ def get(self, path):
             return path
         return None
 
+    def has_path(self, path):
+        """Return True if `path` exists on disk under the store's root."""
+        if not path:
+            return file_utils.exists(self.working_dir)
```
hertschuh: Remove, I don't think this should be supported.

keras/src/saving/saving_lib.py
```diff
@@ -1221,6 +1259,12 @@ def get(self, path):
 
         return self
 
+    def has_path(self, path):
+        """Return True if a group exists at `path` in the H5 file."""
+        if not path:
+            return "vars" in self.h5_file
```
hertschuh: Remove, I don't think this should be supported.

keras/src/saving/saving_lib.py
```diff
@@ -1410,6 +1454,17 @@ def get(self, path):
...
 
+    def has_path(self, path):
+        """Return True if any shard has a group under `path`.
+
+        Resolves via the shard weight-map rather than the currently open
+        shard, so the answer is independent of which shard is active.
+        """
+        weight_map = self.sharding_config.get("weight_map", {})
+        leading = f"/{path}" if path else "/vars"
+        prefix = f"{leading}/"
```
hertschuh: Remove support for None / empty path, I don't think this should be supported.

keras/src/saving/saving_lib.py
```diff
@@ -1638,6 +1693,12 @@ def get(self, path):
             return self.contents[path].tolist()
         return {}
 
+    def has_path(self, path):
+        """Return True if `path` exists as a key in the npz contents."""
+        if not path:
+            return "__root__" in self.contents
```
hertschuh: Remove, I don't think this should be supported.

hertschuh (changes requested):
Thanks for fixing this!

One nitpick, I don't see any use in supporting None / empty paths. If the root is not available, we have a bigger problem.

JyotinderSingh:
Makes sense @hertschuh, I've removed the empty path handling.

hertschuh (approved):
Thanks!

PR #22892, opened 2026-05-13
 Fix bin boundaries calculation in Discretization layer

ul611:
### Problem
This PR fixes an issue where `Discretization` layer produced uneven bin boundaries, resulting in the last bin containing significantly more elements than others. After investigation, the root cause was found in `compress_summary` function: `cum_weight_percents` was normalized by the second-to-last value (`cum_weights[-1]`), not the last, as the last value is excluded in `summarize` function (variable `boundaries`). 

### Solution
The issue is fixed by including the last value in the `summarize` function. This is achieved by shifting the boundaries one position to the left: `boundaries = values[int(start) - 1 :: int(step)]`. To ensure that `int(start) - 1` is never negative, start is set to `start = max(increment, 1)`.

### Visualization

```Python
import tensorflow  as tf

n = 10000
viz_step = 1  # increase to speed up

# array
x = tf.random.uniform((n,))

for num_bins in range(1, 101, viz_step):

    discretization = tf.keras.layers.Discretization(num_bins=num_bins)
    discretization.adapt(x)
    y = discretization(x)

    fig, axis = plt.subplots(1, 1, figsize=(4, 4))
    counts,bins, _ = axis.hist(y, bins=num_bins, rwidth=0.5)
    plt.title(num_bins)
    plt.show()

# batched dataset
batch1 = np.random.uniform(0, 10, n).astype(np.float32)
batch2 = np.random.uniform(5, 15, n).astype(np.float32)
batch3 = np.random.uniform(150, 250, n).astype(np.float32)

x = tf.data.Dataset.from_tensor_slices([batch1, batch2, batch3])
all_data = np.concatenate([batch1, batch2, batch3])

for num_bins in range(1, 101, viz_step):
    discretization = tf.keras.layers.Discretization(num_bins=num_bins)
    discretization.adapt(x)
    y = discretization(all_data).numpy()

    fig, axis = plt.subplots(1, 1, figsize=(4, 4))
    counts,bins, _ = axis.hist(y, bins=num_bins, rwidth=0.5)
    plt.title(num_bins)
    plt.show()
```

## Contributor Agreement

### DIFF
--- a/keras/src/layers/preprocessing/discretization.py
+++ b/keras/src/layers/preprocessing/discretization.py
@@ -388,7 +388,7 @@ def compress_summary(summary, epsilon):
 
     percents = epsilon + np.arange(0.0, 1.0, epsilon)
     cum_weights = summary[1].cumsum()
-    cum_weight_percents = cum_weights / cum_weights[-1]
+    cum_weight_percents = cum_weights / (cum_weights[-1] + cum_weights[0])
     new_bins = np.interp(percents, cum_weight_percents, summary[0])
     cum_weights = np.interp(percents, cum_weight_percents, cum_weights)
     new_weights = cum_weights - np.concatenate(
--- a/keras/src/layers/preprocessing/discretization.py
+++ b/keras/src/layers/preprocessing/discretization.py
@@ -333,9 +333,9 @@ def summarize(values, epsilon):
     elements = np.size(values)
     num_buckets = 1.0 / epsilon
     increment = elements / num_buckets
-    start = max(increment, 1)
-    step = start
-    boundaries = values[int(start) - 1 :: int(step)]
+    step = max(increment, 1)
+    start = step - 1
+    boundaries = values[int(start) :: int(step)]
     weights = np.ones_like(boundaries)
     weights = weights * step
     return np.stack([boundaries, weights])
--- a/keras/src/layers/preprocessing/discretization.py
+++ b/keras/src/layers/preprocessing/discretization.py
@@ -333,9 +333,9 @@ def summarize(values, epsilon):
     elements = np.size(values)
     num_buckets = 1.0 / epsilon
     increment = elements / num_buckets
-    start = increment
+    start = max(increment, 1)
     step = max(increment, 1)
-    boundaries = values[int(start) :: int(step)]
+    boundaries = values[int(start) - 1 :: int(step)]
     weights = np.ones_like(boundaries)
     weights = weights * step
     return np.stack([boundaries, weights])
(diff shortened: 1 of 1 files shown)


ul611:
@gemini-code-assist review

ul611:
@gemini-code-assist review

keras/src/layers/preprocessing/discretization.py
```diff
@@ -333,9 +333,9 @@ def summarize(values, epsilon):
     elements = np.size(values)
     num_buckets = 1.0 / epsilon
     increment = elements / num_buckets
-    start = increment
-    step = max(increment, 1)
-    boundaries = values[int(start) :: int(step)]
+    start = max(increment, 1)
+    step = start
+    boundaries = values[int(start) - 1 :: int(step)]
```
hertschuh: Just to make it a little less confusing, can you rewrite it like this:

```python
    step = max(increment, 1)  
    start = step - 1
    boundaries = values[int(start) :: int(step)]
```

hertschuh (changes requested):
I _think_ the math is correct. But can you re-run your test but with a non-uniform distribution, like `np.random.normal`?

keras/src/layers/preprocessing/discretization.py
ul611: Done

ul611:
Thanks for the suggestion. I tested on normal distribution: the original layer produces tail boundary shifts up to `-0.25 ` (or even lower for a single batch) compared to theoretical, while the fixed version stays within `±0.03`. Code snippet and screenshots attached. The difference is negligible for small `num_bins` but significant for `num_bins ~ 1/epsilon`.

image

image

```Python
import numpy as np
import tensorflow as tf
from scipy.stats import norm

epsilon = 0.01
n = 40000
np.random.seed(42)

for num_bins in range(5, 100, 5):
    # Data (normal distribution)
    batch1 = np.random.normal(0, 1, n).astype(np.float32)
    batch2 = np.random.normal(0, 1, n).astype(np.float32)
    batch3 = np.random.normal(0, 1, n).astype(np.float32)
    all_data = np.concatenate([batch1, batch2, batch3])
    x_ds = tf.data.Dataset.from_tensor_slices([batch1, batch2, batch3])

    # Normal distribution parameters calculated from all data
    mu = np.mean(all_data)
    std = np.std(all_data, ddof=1)

    # Theoretical quantiles (ideal boundaries)
    quantiles = np.linspace(0, 1, num_bins + 1)[1:-1]
    theoretical = norm.ppf(quantiles, loc=mu, scale=std)

    # Layers
    original_layer = tf.keras.layers.Discretization(num_bins=num_bins, epsilon=epsilon)
    original_layer.adapt(x_ds)
    orig_bounds = np.array(original_layer.bin_boundaries)
    dev_orig = orig_bounds - theoretical

    # Plot deviations
    plt.figure(figsize=(12, 5))
    x = np.arange(1, len(theoretical)+1)
    plt.plot(x, dev_orig, 'o-', label='Original', markersize=3, alpha=0.7)
    plt.axhline(0, color='k', linewidth=0.8)
    plt.xlabel('Boundary index')
    plt.ylabel('Deviation from theoretical quantile')
    plt.title(f'Normal distribution, num_bins={num_bins}')
    plt.legend()
    plt.grid(True)
    plt.show()
    ```

hertschuh (approved):
Thanks for investigating this!

PR #22893, opened 2026-05-13
Fix LiteRT export input shape collapse and torch backend SymInt/SymFloat support

pctablet505:
## Summary

This PR fixes two critical issues discovered while exporting `google/gemma-3-270m` from KerasHub to LiteRT (TFLite) via both TensorFlow and torch backends:

1. **Torch backend crashes during `torch.export.export`** because Keras ops don't handle `torch.SymInt` / `torch.SymFloat` symbolic values from dynamic shape tracing.
2. **LiteRT export produces `[1, 1]` input shapes** even after the model is called with larger concrete inputs (e.g. `[1, 32]`). This causes TFLite models to be unusable for sequence lengths > 1 because the converter traces with collapsed sample tensors.

Both issues block real-world export of pretrained causal LM models.

---

### 1. Handle `torch.SymInt` / `torch.SymFloat` in torch backend

When `torch.export.export()` traces a model, shape indices become symbolic (`torch.SymInt`) rather than plain Python `int`. Keras' `convert_to_tensor` and `slice` fast paths rejected these, crashing export.

**`keras/src/backend/torch/core.py`**
- `convert_to_tensor`: detects scalar or list/tuple containing `SymInt`/`SymFloat` and routes them through `torch.as_tensor` directly (bypassing numpy which cannot handle symbolic dtypes).
- `slice`: accepts `torch.SymInt` / `torch.SymFloat` in `start_indices` and `shape` for the fast-path slice construction.

### 2. Merge `_build_shapes_dict` into `get_input_signature` for Functional models

`model.export(format="litert")` calls `get_input_signature(model)` to derive sample tensors. For Functional models, this was returning all-`None` shapes that collapsed to `[1, 1]`, ignoring shapes from recent `model()` calls.

**`keras/src/export/export_utils.py`**
- After building the signature from `_inputs_struct`, merges concrete shapes from `_build_shapes_dict` into the `InputSpec` objects.
- `_update_input_spec_shape` is a hoisted, recursive module-level helper that preserves explicitly defined non-None dimensions from the original `InputSpec` (e.g. a fixed sequence length set via `Input(shape=(None, 128))`) while filling in concrete values for dynamic dims from the most recent call.
- Handles nested dict and list/tuple input structures correctly.
- Uses `build_shapes.get("inputs_shape")` directly for Functional models.

### 3. Update `_build_shapes_dict` after `_maybe_build` for subclassed models

Subclassed models that don't override `build()` need their `_build_shapes_dict` refreshed after each call so `get_input_signature` stays current.

**`keras/src/layers/layer.py`**
- In `_maybe_build()`, updates `_build_shapes_dict` with the latest call shapes when `build` is the default implementation.
- Gated by `utils.is_default(self.build)` to avoid breaking models with custom `build()` that rely on `_build_shapes_dict` for deserialization.
- **Guarded against symbolic shapes**: skips the update when shapes contain non-serializable symbolic values (e.g. from `torch.export` tracing) to prevent `model.save()` from failing. `_is_concrete_shapes_dict` validates that shapes only contain JSON-safe types (`int`, `None`, `tuple`, `list`).
- **Scope guards**: skips the update in stateless or symbolic scopes to avoid interfering with `torch.export` or JAX tracing.
- Only updates when shapes have actually changed, avoiding unnecessary state mutations.
- Renamed internal variables (`_Model` → `_ModelClass`, `shapes_dict` → `build_shapes`, `standardize_shape_or_none` hoisted) for clarity.

### 4. Tests

**`keras/src/backend/torch/core_test.py`** (6 new tests)
- Scalar `SymInt` and `SymFloat` conversion
- List/tuple of symbolic values
- Explicit dtype override with symbolic values
- `slice` fast path with `SymInt`

**`keras/src/layers/layer_test.py`** (8 new tests)
- `_is_concrete_shapes_dict` validation (None, int, tuple, list, dict, invalid types)
- Scope guards (`stateless_scope`, `symbolic_scope`) skip `_build_shapes_dict` update
- Model `save()` / `load_model()` round-trip after `torch.export`
- Custom `build()` preservation
- Regular layer isolation (non-Models unaffected)
- Functional model explicit-`None` dim preservation

**`keras/src/export/litert_test.py`** (3 new tests)
- Subclass model input signature reflects recent call shape
- Custom build model preserves original shape
- Functional model dynamic dims

**`keras/src/export/litert_torch_test.py`** (2 new tests)
- Subclass model shape update for torch backend LiteRT export

**`keras/src/export/saved_model_test.py`** (1 new test)
- SavedModel export uses recent call shape for subclassed models

---

## Why these changes are needed

Without them, exporting a real Gemma 3 270M model produces a TFLite file with `[1, 1]` inputs that cannot generate more than 1 token. The torch backend export path is completely blocked by the `SymInt` crash. These are not edge cases — they affect every autoregressive model export via KerasHub.

---

## Related issues

- Fixes input shape collapse for LiteRT export of CausalLM models
- Unblocks torch-backend LiteRT export for pretrained models

## Verification

End-to-end verified on `google/gemma-3-270m-it`:

| Backend | Export | Input shape | On-device inference |
|---|---|---|---|
| TensorFlow | `gemma3_270m_it_tf.tflite` (1024 MB) | `[1, 128]` | ✅ Pixel 9 (Android 16) |
| Torch | `gemma3_270m_it_torch.tflite` (1024 MB) | `[1, 128]` | ✅ Pixel 9 (Android 16) |

All new tests pass:
- `keras/src/backend/torch/core_test.py` — 6 passed
- `keras/src/layers/layer_test.py` — 8 passed (66 total, 2 skipped)
- `keras/src/export/litert_test.py` — 3 passed
- `keras/src/export/litert_torch_test.py` — 2 passed
- `keras/src/export/saved_model_test.py` — 1 passed

## Contributor Agreement

### DIFF
--- a/keras/src/backend/torch/core.py
+++ b/keras/src/backend/torch/core.py
@@ -219,6 +219,10 @@ def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
         else:
             dt = torch.complex64
         return torch.as_tensor(x, dtype=dt, device=get_device())
+    if isinstance(x, (torch.SymInt, torch.SymFloat)):
+        # Scalar symbolic values from torch.export can't go through numpy.
+        dt = to_torch_dtype(dtype) if dtype is not None else None
+        return torch.as_tensor(x, dtype=dt, device=get_device())
 
     # Convert to np in case of any array-like that is not list or tuple.
     # Skip scalar Python values to avoid np.array(float) -> float64, which
@@ -228,6 +232,13 @@ def convert_to_tensor(x, dtype=None, sparse=None, ragged=None):
         if len(x) > 0 and any(isinstance(x1, torch.Tensor) for x1 in x):
             # Handle list or tuple of torch tensors
             return torch.stack([convert_to_tensor(x1) for x1 in x])
+        if len(x) > 0 and any(
+            isinstance(x1, (torch.SymInt, torch.SymFloat)) for x1 in x
+        ):
+            # Symbolic shape values from torch.export can't go through numpy
+            # and don't have a .dtype attribute. Use torch.as_tensor directly.
+            dt = to_torch_dtype(dtype) if dtype is not None else None
+            return torch.as_tensor(x, dtype=dt, device=get_device())
     elif not isinstance(x, (bool, int, float)):
         x = np.array(x)
     if isinstance(x, np.ndarray):
@@ -630,9 +641,9 @@ def slice(inputs, start_indices, shape):
     if isinstance(start_indices, (list, tuple)) and isinstance(
         shape, (list, tuple)
     ):
-        if all(isinstance(s, int) for s in start_indices) and all(
-            isinstance(s, int) for s in shape
-        ):
+        if all(
+            isinstance(s, (int, torch.SymInt)) for s in start_indices
+        ) and all(isinstance(s, (int, torch.SymInt)) for s in shape):
             slices = [
                 builtins.slice(start_index, start_index + length)
                 for start_index, length in zip(start_indices, shape)
--- a/keras/src/export/export_utils.py
+++ b/keras/src/export/export_utils.py
@@ -43,8 +43,7 @@ def _update_input_spec_shape(spec, shape):
         if len(shape) != len(spec):
             return spec
         result = [
-            _update_input_spec_shape(v, shape[i])
-            for i, v in enumerate(spec)
+            _update_input_spec_shape(v, shape[i]) for i, v in enumerate(spec)
         ]
         return tuple(result) if isinstance(spec, tuple) else result
     return spec
@@ -96,7 +95,6 @@ def get_input_signature(model):
             if isinstance(latest_input_shapes, dict) and set(
                 latest_input_shapes.keys()
             ) == set(input_signature[0].keys()):
-
                 input_signature[0] = {
                     k: _update_input_spec_shape(v, latest_input_shapes[k])
                     for k, v in input_signature[0].items()
--- a/keras/src/export/export_utils.py
+++ b/keras/src/export/export_utils.py
@@ -35,6 +35,79 @@ def get_input_signature(model):
         input_signature = [
             tree.map_structure(make_input_spec, model._inputs_struct)
         ]
+        # For models that were called with concrete shapes, merge those
+        # shapes into the signature so that exporters (especially torch
+        # backend LiteRT) trace with representative dimensions rather
+        # than all-None shapes that collapse to (1, 1) sample tensors.
+        shapes_dict = getattr(model, "_build_shapes_dict", None)
+        if (
+            shapes_dict
+            and input_signature
+            and isinstance(input_signature[0], dict)
+        ):
+            # For Functional models the call argument is always named
+            # ``inputs``, so the recorded shapes dict contains the key
+            # ``inputs_shape``.
+            actual_shapes = shapes_dict.get("inputs_shape")
+            if isinstance(actual_shapes, dict) and set(
+                actual_shapes.keys()
+            ) == set(input_signature[0].keys()):
+
+                def _update_spec(spec, shape):
+                    if isinstance(spec, layers.InputSpec):
+                        original_shape = spec.shape
+                        shape_tuple = tuple(shape)
+                        # Preserve explicit non-None dimensions from the
+                        # original InputSpec (e.g. a fixed sequence length
+                        # set via Input(shape=(None, 128))) while filling
+                        # in concrete values for dynamic dims from the
+                        # most recent call.
+                        if original_shape is not None:
+                            new_shape = tuple(
+                                original_shape[i]
+                                if i < len(original_shape)
+                                and original_shape[i] is not None
+                                else dim
+                                for i, dim in enumerate(shape_tuple)
+                            )
+                        else:
+                            new_shape = shape_tuple
+                        return layers.InputSpec(
+                            shape=new_shape,
+                            dtype=spec.dtype,
+                            name=getattr(spec, "name", None),
+                        )
+                    if isinstance(spec, dict):
+                        if not isinstance(shape, dict):
+                            return spec
+                        return {
+                            k: _update_spec(v, shape[k])
+                            for k, v in spec.items()
+                            if k in shape
+                        }
+                    if isinstance(spec, (list, tuple)):
+                        if not isinstance(shape, (list, tuple)):
+                            return spec
+                        if len(shape) != len(spec):
+                            return spec
+                        result = [
+                            _update_spec(v, shape[i])
+                            for i, v in enumerate(spec)
+                        ]
+                        return (
+                            tuple(result) if isinstance(spec, tuple) else result
+                        )
+                    return spec
+
+                # Use a plain dict comprehension rather than
+                # tree.map_structure because InputSpec may be registered
+                # as a pytree node in some backends (e.g.
+                # TensorFlow/optree) and would not match the tuple
+                # structure of actual_shapes.
+                input_signature[0] = {
+                    k: _update_spec(v, actual_shapes[k])
+                    for k, v in input_signature[0].items()
+                }
     elif isinstance(model, models.Sequential):
         input_signature = tree.map_structure(make_input_spec, model.inputs)
     else:
--- a/keras/src/export/export_utils.py
+++ b/keras/src/export/export_utils.py
@@ -6,6 +6,50 @@ from keras.src import tree
 from keras.src.utils.module_utils import tensorflow as tf
 
 
+def _update_input_spec_shape(spec, shape):
+    """Recursively update an InputSpec with concrete shape values.
+
+    Preserves explicitly defined non-None dimensions from the original
+    InputSpec while filling in dynamic dims from the most recent call.
+    """
+    if isinstance(spec, layers.InputSpec):
+        original_shape = spec.shape
+        shape_tuple = tuple(shape)
+        if original_shape is not None:
+            new_shape = tuple(
+                original_shape[i]
+                if i < len(original_shape) and original_shape[i] is not None
+                else dim
+                for i, dim in enumerate(shape_tuple)
+            )
+        else:
+            new_shape = shape_tuple
+        return layers.InputSpec(
+            shape=new_shape,
+            dtype=spec.dtype,
+            name=getattr(spec, "name", None),
+        )
+    if isinstance(spec, dict):
+        if not isinstance(shape, dict):
+            return spec
+        return {
+            k: _update_input_spec_shape(v, shape[k])
+            for k, v in spec.items()
+            if k in shape
+        }
+    if isinstance(spec, (list, tuple)):
+        if not isinstance(shape, (list, tuple)):
+            return spec
+        if len(shape) != len(spec):
+            return spec
+        result = [
+            _update_input_spec_shape(v, shape[i])
+            for i, v in enumerate(spec)
+        ]
+        return tuple(result) if isinstance(spec, tuple) else result
+    return spec
+
+
 def get_input_signature(model):
     """Get input signature for model export.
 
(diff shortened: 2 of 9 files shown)


keras/src/export/export_utils.py
```diff
@@ -35,6 +35,40 @@ def get_input_signature(model):
...
+        if shapes_dict and input_signature and isinstance(
+            input_signature[0], dict
+        ):
+            actual_shapes = None
+            for value in shapes_dict.values():
+                if isinstance(value, dict) and set(value.keys()) == set(
+                    input_signature[0].keys()
+                ):
+                    actual_shapes = value
+                    break
```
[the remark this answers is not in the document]
pctablet505: can be wrong for inputs having same shape.

pctablet505:
@gemini-code-assist review

keras/src/backend/torch/core.py
```diff
@@ -630,9 +641,9 @@ def slice(inputs, start_indices, shape):
     if isinstance(start_indices, (list, tuple)) and isinstance(
         shape, (list, tuple)
     ):
-        if all(isinstance(s, int) for s in start_indices) and all(
-            isinstance(s, int) for s in shape
-        ):
+        if all(
+            isinstance(s, (int, torch.SymInt)) for s in start_indices
+        ) and all(isinstance(s, (int, torch.SymInt)) for s in shape):
```
amitsrivastava78, line 646: Torch backend changes have effectively zero coverage

The SymInt/SymFloat fixes in keras/src/backend/torch/core.py are the primary motivation for the PR title ("Fix tflite export issues with torch"), but Codecov reports only 14% patch coverage on that file. Specifically, the PR currently has:

No direct test for convert_to_tensor with a SymInt scalar.

No direct test for convert_to_tensor with a list/tuple containing SymInt/SymFloat.

No direct test for slice with SymInt in start_indices or shape.

The closest test is test_subclass_model_input_signature_reflects_recent_call in litert_torch_test.py, but that goes through the full export path. If anything between torch.export and convert_to_tensor changes in the future, the integration test could pass while the actual SymInt branches never execute, leading to silent regressions.

Request: Please add focused unit tests in keras/src/backend/torch/core_test.py that pass a torch.SymInt directly to convert_to_tensor and slice.

keras/src/layers/layer.py
```diff
@@ -1987,6 +2031,27 @@ def standardize_shape_or_none(x):
     return shapes_dict
 
 
+def _is_concrete_shapes_dict(shapes_dict):
```
amitsrivastava78, line 2037: _is_concrete_shapes_dict needs direct unit tests

This new module-level helper acts as the safety net that prevents model.save() from breaking after symbolic tracing. Currently, it has zero direct coverage and is only exercised indirectly via the integration tests.

Since this utility is critical for serialization, please add direct unit tests for it (likely in export_utils_test.py). Ensure the tests cover the following edge cases:

None shapes (e.g., {"x_shape": None})

Nested dictionary shapes

Mixed concrete and symbolic shapes

Shapes containing np.integer values

An empty shapes_dict

Relying purely on the export integration tests leaves these branches vulnerable if the upstream shape inference logic changes in the future.

keras/src/export/litert_torch_test.py
```diff
@@ -121,6 +121,50 @@ def call(self, x):
         x = np.random.normal(size=(1, 10)).astype("float32")
         self._verify_litert_export(model, x)
 
+    def test_subclass_model_input_signature_reflects_recent_call(self):
```
amitsrivastava78, line 152: The dict-only gate is never tested with non-dict subclass inputs

test_subclass_model_input_signature_reflects_recent_call currently uses a dictionary input ({"token_ids": ...}). However, there is no test verifying this behavior for a subclass model with a standard single-tensor input (e.g., def call(self, x)).
Please add a parallel test for a subclass model with a single-tensor input being re-called at a new shape and successfully exported.

keras/src/layers/layer.py
```diff
@@ -1540,6 +1542,48 @@ def count_params(self):
...
+                # Avoid calling get_shapes_dict during torch tracing
+                # (ONNX export, torch.jit.trace) because tensor shapes
+                # become traced scalars and standardize_shape fails on
+                # them.
+                if backend.backend() == "torch":
+                    import torch
+
+                    if torch.jit.is_tracing():
+                        return
+                try:
```
amitsrivastava78: Untested try/except path and scope guards
 wrap get_shapes_dict(call_spec) in a try/except (TypeError, ValueError) and silently return. There are currently no tests exercising this exception path. If this block is too broad, swallows the wrong exception types, or if upstream logic changes to throw a different error, we will have no automated way to catch it.
Similarly, the in_stateless_scope() and in_symbolic_scope() guards are not covered. Since Keras relies on these scopes for functional purity during JAX/Torch tracing, we need to ensure these guards actually prevent state mutation.

1. Add a test that intentionally triggers the try/except block to verify it catches the expected errors and fails gracefully.
2. Add a test that calls a subclass model inside a model.stateless_call(...) (or a symbolic trace) and asserts that _build_shapes_dict remains strictly unchanged.

keras/src/layers/layer.py
```diff
@@ -1540,6 +1542,48 @@ def count_params(self):
 
     def _maybe_build(self, call_spec):
         if self.built:
+            # For models without a custom build(), update _build_shapes_dict
```
amitsrivastava78: Missing end-to-end test for model.save() after tracing
The PR's stated goal for introducing the symbolic-shape guards (like _is_concrete_shapes_dict) is to prevent model.save() from failing due to symbolic shapes polluting the model state during torch.export. However, there is no test verifying this actually works.
If this guard is fundamentally broken, or if symbolic shapes leak into another attribute not covered by this PR, the current test suite will still pass.
Please add a round-trip serialization test. The test should:
1. Initialize a subclass model.
2. Trigger a torch.export trace (which injects the SymInt/SymFloat types).
3. Call model.save("test.keras").
4. Successfully load the model back via keras.saving.load_model.
This is the only way to guarantee that the core bug this PR claims to fix is actually resolved and won't regress.

keras/src/export/export_utils.py
```diff
@@ -35,6 +35,56 @@ def get_input_signature(model):
...
         ]
+        # For models that were called with concrete shapes, merge those
+        # shapes into the signature so that exporters (especially torch
+        # backend LiteRT) trace with representative dimensions rather
+        # than all-None shapes that collapse to (1, 1) sample tensors.
+        shapes_dict = getattr(model, "_build_shapes_dict", None)
+        if (
+            shapes_dict
+            and input_signature
+            and isinstance(input_signature[0], dict)
```
amitsrivastava78, line 89: Dict check breaks explicit dynamic shapes
Gating the shape update behind isinstance(input_signature[0], dict) conflates data structures with model types.
If a user builds a Functional model with dictionary inputs explicitly defined as Input(shape=(None,)), this logic will incorrectly overwrite their None with the concrete shape of the last call, silently breaking dynamic exports.
Please respect explicitly defined None dimensions universally across all input types (dict, tuple, single tensor). The "last call shape" fallback should strictly apply to unspecified subclass inputs, not trigger just because the input is a dict.

keras/src/export/export_utils.py
```diff
@@ -35,6 +35,56 @@ def get_input_signature(model):
         input_signature = [
             tree.map_structure(make_input_spec, model._inputs_struct)
         ]
+        # For models that were called with concrete shapes, merge those
+        # shapes into the signature so that exporters (especially torch
+        # backend LiteRT) trace with representative dimensions rather
+        # than all-None shapes that collapse to (1, 1) sample tensors.
+        shapes_dict = getattr(model, "_build_shapes_dict", None)
```
amitsrivastava78: Sequential models are completely unhandled
The PR currently fixes Subclass models (via _maybe_build) and Functional models (via the dict-input check). However, Sequential models—which also use _build_shapes_dict and suffer from the exact same stale-shape symptom—are left out.
If a Sequential model is built at [1, 1] and later called at [1, 32], it will still incorrectly export at [1, 1].
Ensure the shape-update logic covers Sequential models as well. Please add a test verifying that a single-tensor Sequential model updates its export signature correctly after being called with a new shape.

keras/src/export/export_utils.py
```diff
@@ -35,6 +35,56 @@ def get_input_signature(model):
...
+                    if isinstance(spec, layers.InputSpec):
+                        new_shape = (None,) + tuple(shape)[1:]
+                        return layers.InputSpec(
+                            shape=new_shape,
+                            dtype=spec.dtype,
+                            name=getattr(spec, "name", None),
+                        )
+                    if isinstance(spec, dict):
+                        return {
+                            k: _update_spec(v, shape[k])
```
amitsrivastava78: Unsafe recursion in _update_spec

If spec is a list/tuple but shape is None or has a mismatched length, shape[i] will crash with an IndexError or TypeError mid-export. The top-level check only guards dictionary keys, leaving nested structural mismatches completely unprotected.

keras/src/layers/layer.py
```diff
@@ -1540,6 +1542,48 @@ def count_params(self):
...
+            # so that export tools (get_input_signature, saved-model serving,
```
amitsrivastava78: torch.jit.is_tracing() is the wrong probe for torch.export
torch.jit.is_tracing() only returns True for legacy TorchScript tracing. Since this PR explicitly targets torch.export (which uses Dynamo/FX), this guard will evaluate to False and effectively do nothing.
The fallback try/except handles the error after the fact, making this check dead weight and misleading to future readers.

keras/src/layers/layer.py
```diff
@@ -1540,6 +1542,48 @@ def count_params(self):
...
+            # etc.) use the most recent call shapes rather than the stale
```
amitsrivastava78: Significant hot-path regression in _maybe_build
_maybe_build runs on every __call__ of a layer or model, making it part of the hot path for training and inference loops. Master exits instantly with if self.built: return.
This PR modifies it to execute a non-trivial workload every single time a built model is called:
3+ deferred imports (cached in sys.modules but still incurs lookup overhead)
Dynamic scope and type checks (isinstance, scope guards)
A full get_shapes_dict() traversal and a recursive _is_concrete_shapes_dict() walk on every forward pass.
This introduces a measurable Python overhead regression for high-frequency loops.

keras/src/export/litert_test.py
```diff
@@ -1252,6 +1266,117 @@ def test_dict_input_preserves_variable_sharing(self):
...
+        )
+        model.export(temp_filepath, format="litert")
+        self.assertTrue(os.path.exists(temp_filepath))
+
+        interpreter = LiteRTInterpreter(model_path=temp_filepath)
+        interpreter.allocate_tensors()
+
+        input_details = interpreter.get_input_details()
+        self.assertEqual(len(input_details), 1)
+        # The single input should have shape [1, 32] (batch is dynamic None)
```
amitsrivastava78: This comment is technically incorrect and contradicts your other test. LiteRT/TFLite's get_input_details()["shape"] collapses dynamic dimensions (None) to 1 in its output array. Therefore, seeing 1 in the shape array does not mean the batch is dynamic—it's just TFLite's fallback representation.
Your functional test comment correctly states "TFLite uses [1, 1] as the default for dynamic dims", but this comment tells the exact opposite story for the same underlying framework behavior.

keras/src/export/litert_test.py
```diff
@@ -1252,6 +1266,117 @@ def test_dict_input_preserves_variable_sharing(self):
...
+
+            def call(self, x):
+                return self.dense(x)
+
+        model = CustomBuildModel()
+        model.build((None, 10))  # explicit build at feature dim 10
+        original_shapes = dict(model._build_shapes_dict)
+
+        # Subsequent call with different batch but same feature dim
+        model(ops.zeros((2, 10), dtype="float32"))
```
amitsrivastava78: This test checks if _build_shapes_dict changes after a second call, but both the build and the forward call use the exact same feature dimension (10). Because the feature dimensions match, this test would pass even on the buggy code this PR is trying to fix.
To actually exercise the guard, the second call needs to use a different feature shape (e.g., (2, 20)). However, doing that on this specific model will just trigger a shape mismatch crash because the internal Dense layer is already locked into a dimension of 10.
Update the test model design so that the feature dimension doesn't constrain subsequent calls (for example, by using a model with an initial Embedding layer or global pooling), and then call it with a completely different shape to prove the guard works.

keras/src/layers/layer.py
```diff
@@ -1540,6 +1551,46 @@ def count_params(self):
...
+                # During torch.export.export() or other tracing contexts,
+                # get_shapes_dict may fail on symbolic/traced shapes.
+                return
+            # Only update with concrete (non-symbolic) shapes.
+            # Symbolic values from torch.export/jax tracing are not
+            # JSON-serializable and would break model.save().
+            if _is_concrete_shapes_dict(shapes_dict):
+                existing = getattr(self, "_build_shapes_dict", None)
+                if shapes_dict != existing:
+                    self._build_shapes_dict = shapes_dict
```
hertschuh, line 1593: This is super weird, seems brittle, and will probably slow down running the model eagerly, in particular with Torch.

It's weird because if the model is built, you should not do anything. It's brittle, the of the exclusion with `if`s and the `try` / `except` blocks are a sign of that.

Does liteRT not support dynamic dimensions like dynamic batch sizes? Support for export with dynamic dimensions was added, how come this is needed?

> LiteRT export produces [1, 1] input shapes even after the model is called with larger concrete inputs (e.g. [1, 32]). This causes TFLite models to be unusable for sequence lengths > 1 because the converter traces with collapsed sample tensors.

Where do the `1`s come from? Does something replace all `None`s with `1`s?

If you run it with `[1, 32]`, and then export it, does it support other sequence lengths besides exactly 32?

pctablet505:
https://github.com/keras-team/keras/pull/22998 new PR for symInt fix.
Dynamic shape is WONT fix for now, and documented it in https://github.com/keras-team/keras/pull/22997

PR #22894, opened 2026-05-13
Fix validate_reshape_shape crash under torch.compile with dynamic shapes

MarcosAsh:
## Description
 `validate_reshape_shape` (added in #22868) uses a list comprehension + `.count()` pattern:
  
  ```python
  if [d for d in newshape if isinstance(d, int)].count(-1) > 1:
  ```
 Under torch.compile, dynamo traces this through sympy. When newshape contains torch.SymInt values mixed with a Python -1 (which happens any time ops.shape(x) flows into a downstream ops.reshape), the comparison fails with:

 InternalTorchDynamoError: TypeError: Can only compare inequalities with Expr
  
GroupNormalization._reshape_into_groups hits this on every train step on torch + jit_compile=True. Affects diffusion-style models that pair GroupNormalization with attention blocks using ops.shape.

### Fix

Replace the list-comp + .count() with a single counter loop. Semantically identical for static shapes. Dynamic dims now pass through cleanly under dynamo. I also added a regression test in operation_utils_test.py for the mixed dynamic + -1 path. Existing #22868 tests still pass.

Made a Colab notebook to reproduce the error.

[Colab notebook](https://colab.research.google.com/drive/1LP4xZO1YgarcdhYOCMOpDb4DTZnfwHJD?usp=sharing)

Verified on A100 and H100, torch  2.10.0+cu128.
## Contributor Agreement

### DIFF
--- a/keras/src/ops/operation_utils.py
+++ b/keras/src/ops/operation_utils.py
@@ -287,16 +287,22 @@ def validate_reshape_shape(newshape, newshape_arg_name="newshape"):
 
     Each dimension that is a concrete Python int must be either non-negative
     or `-1` (with at most one `-1`). Dynamic dimensions (e.g. backend tensor
-    scalars resolved at runtime) are not validated here.
+    scalars resolved at runtime, such as `torch.SymInt` under `torch.compile`)
+    are not validated here.
     """
+    neg_one_count = 0
     for dim in newshape:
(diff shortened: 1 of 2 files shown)


MarcosAsh:
I fixed the issues Gemini said

hertschuh (approved):
Thanks for the fix!

PR #22896, opened 2026-05-14
Reject symlinks in assets/ when loading directory-form models

LinZiyuu:
## Summary

`keras.saving.load_model(filepath)` accepts a directory as well as a `.keras` zip. When the path is a directory, `_load_model_from_dir` hands the model's `assets/` subtree to `DiskIOStore` in read-from-dir mode, which performs no extraction, no zipinfo / tarinfo filter, and no symlink check. Layer-side `load_assets(dir_path)` implementations then `open(...)` per-layer asset files (e.g. `assets/layers/<name>/_lookup_layer/vocabulary.txt`) and Python's `open` follows filesystem symlinks like any other file open.

A malicious model directory whose `vocabulary.txt` is a symlink to any file the victim process can read therefore causes Keras to load the target file's bytes into the model's vocabulary table. From there the attacker recovers them via `model.layers[i].get_vocabulary()`, by calling `model.save("exfil.keras")` and downloading the exported file, or by probing the model and reading back the integer token IDs that map to the leaked string.

This is **distinct** from CVE-2026-1669 (the HDF5 `h5py.ExternalLink` / `h5py.SoftLink` bug):

- CVE-2026-1669 lives in the HDF5 link layer; `safe_get_h5_group` / `safe_get_h5_dataset` are the defenses there.
- This bug lives at the POSIX filesystem layer — the assets are loose files, not HDF5 datasets — so the HDF5 link checks do not apply. The archive-extraction symlink filters (`filter_safe_zipinfos` / `filter_safe_tarinfos`) are also bypassed because the directory load path never extracts anything.

`safe_mode` does **not** gate this code path. Asset loading runs even when `safe_mode=True` (the default), so trusted-by-default users are still exposed.

## Fix

Reject symlinks under the asset directory at load time, and any path whose canonical target escapes the asset root, in two places:

- `_load_model_from_dir` performs a one-shot scan via `_validate_asset_dir_no_symlink_escape` **before** any layer-side `open(...)` so a malicious directory is refused up front rather than partway through state restoration.
- `DiskIOStore.get` adds defense-in-depth via `_reject_symlink_escape` for any other caller of the read-from-dir store.

Both helpers skip the check on remote URIs (e.g. `gs://`, `hdfs://`) since those backends have no concept of symbolic links.

Keras itself never writes symlinks into the asset tree, so any symlink found inside `assets/` — even one pointing to a sibling file inside the same tree — is treated as tamper evidence and rejected.

## Test plan

- [x] New `test_load_model_from_dir_rejects_symlinked_asset` — saves a `TextVectorization` model, unpacks it to a directory, swaps `vocabulary.txt` for a symlink to an out-of-tree file, and asserts `saving_lib.load_model(<dir>)` raises `ValueError`.
- [x] New `test_load_model_from_dir_rejects_symlink_to_inside_asset_root` — the same scenario but with the symlink target inside the asset directory; still rejected.
- [x] Existing directory-load tests (e.g. `test_save_unzipped`) still pass — non-symlink assets are unaffected.

## Disclosure

Reported by Ziyu Lin via huntr: https://huntr.com/bounties/0090e3ec-d53e-4595-860c-b22bc990b905

## Contributor agreement

### DIFF
--- a/keras/src/saving/saving_lib.py
+++ b/keras/src/saving/saving_lib.py
@@ -387,9 +387,9 @@ def _load_model_from_dir(dirpath, custom_objects, compile, safe_mode):
                 f"Expected a {_VARS_FNAME_H5} or {_VARS_FNAME_NPZ} file."
             )
         if len(all_filenames) > 3:
-            asset_store = DiskIOStore(
-                file_utils.join(dirpath, _ASSETS_DIRNAME), mode="r"
-            )
+            assets_dirpath = file_utils.join(dirpath, _ASSETS_DIRNAME)
+            _validate_asset_dir_no_symlink_escape(assets_dirpath)
+            asset_store = DiskIOStore(assets_dirpath, mode="r")
 
         else:
             asset_store = None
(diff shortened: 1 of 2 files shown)


LinZiyuu:
@googlebot I signed it!

hertschuh:
> A malicious model directory whose vocabulary.txt is a symlink to any file the victim process can read therefore causes Keras to load the target file's bytes into the model's vocabulary table.

But how would that "malicious model directory" end up on your hard drive? If the attacker has access to your file system, it has nothing to do with Keras.

Users should be allowed to put their files wherever they want, they are the ones creating the symlinks.

LinZiyuu:
> > A malicious model directory whose vocabulary.txt is a symlink to any file the victim process can read therefore causes Keras to load the target file's bytes into the model's vocabulary table.
> 
> But how would that "malicious model directory" end up on your hard drive? If the attacker has access to your file system, it has nothing to do with Keras.
> 
> Users should be allowed to put their files wherever they want, they are the ones creating the symlinks.

The malicious directory doesn't require attacker FS access. Keras itself ships several first-class distribution paths that materialize symlinks onto a victim's disk on the victim's behalf:

  1. hf://... — keras.saving.load_model("hf://attacker/model") calls huggingface_hub.snapshot_download(repo_id=...) (saving_lib.py:339-346), then hands the resulting cache directory straight to  _load_model_from_dir. HF Hub's git/LFS storage preserves symlinks, so an attacker who publishes a Keras model on HF — no special permissions needed — can ship assets/.../vocabulary.txt -> /etc/passwd. The user never wrote any file; Keras did, on their behalf.
  2. Tarballs and zips that have been extracted to a directory — tar -xzf model.tar.gz (and zip with the right flags, unzip -X, rsync -a, container image layers, NFS / shared research clusters, model-marketplace mirrors that resync rather than re-zip) all preserve symlinks. The user calls the documented load_model(<dir>) and gets disclosure. The .keras archive path is protected by filter_safe_zipinfos/filter_safe_tarinfos; the directory path is the one this PR is closing.

Prior art for this exact bug class being accepted in ML loaders: CVE-2024-3568 (transformers tar load), CVE-2024-0152 (TorchServe symlink in tar), and the symlink-safe path additions in torch.hub.download_url_to_file (2022). It's been a recognized vector across the ecosystem.

Keras's own save_model never emits symlinks, but if a user manually places one to share data across models, the security property isn't actually weakened — only escape-the-root symlinks let an attacker disclose /etc/....

  I'll push a follow-up commit that:
  - Keeps the realpath-escape check (rejects symlinks whose target falls outside assets/) and the up-front scan in _load_model_from_dir.
  - Drops the "any in-tree symlink is tampering" rule and the corresponding test_load_model_from_dir_rejects_symlink_to_inside_asset_root test.

That preserves the security boundary against the realistic threat (public HF Hub / shared model archives) while honoring the legitimate use case you raised. Does that sound acceptable?

PR #22897, opened 2026-05-14
MultiOptimizer support for training sub-networks.

gaga1313:
## Description

Added MultiOptimizer wrapper to enables subnetwork training using multiple optimizers.

Key Changes

**OptimizerMap:**
- Implements a highly flexible mapping container that routes specific variable identifiers (support for exact variable matches, regex path patterns, and standard substrings) to their targeted sub-optimizers independently.

**MultiOptimizer Core Wrapper:**
- Automatically distributes model trainable variables and private optimizer variables to their respective designated sub-optimizers.
- Orchestrates fully synchronized graph updates across all backend runtimes by implementing unified .build(), .apply(), and .stateless_apply() execution logic.
- Supports transparent model configuration serialization and deserialization.

**Verification & Testing**
- Mapping: Rigorously verified variable mapping functionality and conflict handling inside OptimizerMap.
- Backward Compatibility: Validated stateful/stateless gradient updates and callbacks updates, and integration with .fit() function.

## Contributor Agreement

### DIFF
--- a/keras/src/optimizers/multi_optimizer.py
+++ b/keras/src/optimizers/multi_optimizer.py
@@ -186,7 +186,7 @@ class MultiOptimizer(Optimizer):
         )
 
     To access the attributes of the sub-optimizers, iterate over the
-    optimizers using `get_optimizer(i)`:
+    optimizers using `.optimizers`:
 
     For example:
 
@@ -273,13 +273,6 @@ class MultiOptimizer(Optimizer):
                 opt.build(variables)
 
         super().build(var_list)
-
-    def get_optimizer(self, index):
-        return self._inner_optimizers[index]
-
-    @property
-    def num_optimizers(self):
-        return len(self._inner_optimizers)
 
     @property
     def optimizers(self):
--- a/keras/src/optimizers/multi_optimizer.py
+++ b/keras/src/optimizers/multi_optimizer.py
@@ -16,8 +16,7 @@ class OptimizerMap:
         optimizer: A list of optimizer instances.
         variable_identifier: A list of variable identifiers matching the
             optimizer list. Identifiers can be string regex patterns, Keras
-            Variables, list/tuple of Keras Variables, or callables returning a
-            boolean.
+            Variables, or list/tuple of Keras Variables.
     """
 
     def __init__(self, optimizer, variable_identifier):
@@ -53,9 +52,6 @@ class OptimizerMap:
         return None
 
     def _match(self, variable, identifier):
-        # If callable, evaluate it
-        if callable(identifier) and not isinstance(identifier, str):
-            return bool(identifier(variable))
         # If string, treat it as a regex pattern
         if isinstance(identifier, str):
             path = getattr(variable, "path", "") or ""
@@ -84,12 +80,11 @@ class MultiOptimizer(optimizer.Optimizer):
 
     # 2. Create mapping using different identifier types
     opt_map = keras.optimizers.OptimizerMap(
-        optimizer=[opt_adam, opt_sgd, opt_adam, opt_sgd, opt_adam],
+        optimizer=[opt_adam, opt_sgd, opt_adam, opt_sgd],
         variable_identifier=[
             "kernel",                            # variable name or path
             "^dense_1/.*",                       # regex pattern
             keras_var,                           # Keras Variable instance
-            lambda var: "bias" in var.name,      # custom callable function
             [var1, var2]                         # list/tuple
         ]
     )
@@ -130,7 +125,6 @@ class MultiOptimizer(optimizer.Optimizer):
         self.obj_map = obj_map
         self.default_optimizer = default_optimizer
 
-        # Disable loss scaling for all inner optimizers to avoid double scaling
         self.default_optimizer.loss_scale_factor = None
         for opt in self.obj_map.optimizer:
             opt.loss_scale_factor = None
@@ -225,6 +219,11 @@ class MultiOptimizer(optimizer.Optimizer):
             self.build(trainable_variables)
             self.built = True
 
+        # Statefully unscale gradients if loss_scale_factor is set
+        if self.loss_scale_factor is not None:
+            scale = self.loss_scale_factor
+            grads = [g if g is None else g / scale for g in grads]
+
         # Group gradients and variables by sub-optimizer
         optimizer_to_grads_and_vars = {
             opt: [] for opt in self._inner_optimizers
@@ -248,6 +247,11 @@ class MultiOptimizer(optimizer.Optimizer):
 
         if trainable_variables is None:
             trainable_variables = self._trainable_variables
+
+        # Statelessly unscale gradients if loss_scale_factor is set
+        if self.loss_scale_factor is not None:
+            scale = self.loss_scale_factor
+            grads = [g if g is None else g / scale for g in grads]
 
         own_var_count = len(self._variables)
         own_variables = optimizer_variables[:own_var_count]
@@ -366,10 +370,6 @@ class MultiOptimizer(optimizer.Optimizer):
                     serialized_identifiers.append(f"^({or_pattern})$")
                 else:
                     serialized_identifiers.append(identifier)
-            elif callable(identifier):
-                serialized_identifiers.append(
-                    serialization_lib.serialize_keras_object(identifier)
-                )
             else:
                 serialized_identifiers.append(identifier)
 
@@ -410,12 +410,7 @@ class MultiOptimizer(optimizer.Optimizer):
             for opt_config in serialized_opts
         ]
 
-        variable_identifiers = [
-            serialization_lib.deserialize_keras_object(
-                ident, custom_objects=custom_objects
-            )
-            for ident in obj_map_config["config"]["variable_identifier"]
-        ]
+        variable_identifiers = obj_map_config["config"]["variable_identifier"]
 
         map_cls = cls._get_map_class(
             obj_map_config["class_name"], custom_objects
--- a/keras/src/optimizers/multi_optimizer.py
+++ b/keras/src/optimizers/multi_optimizer.py
@@ -76,8 +76,7 @@ class MultiOptimizer(optimizer.Optimizer):
         default_optimizer: Default Keras Optimizer for any unmapped variables.
         name: String. The name of this optimizer.
     """
-
-    @tracking.no_automatic_dependency_tracking
+    
     def __init__(self, obj_map, default_optimizer, name=None, **kwargs):
         if not isinstance(obj_map, OptimizerMap):
             raise ValueError(
@@ -352,8 +351,6 @@ class MultiOptimizer(optimizer.Optimizer):
     def _get_map_class(cls, class_name, custom_objects=None):
         if custom_objects and class_name in custom_objects:
             return custom_objects[class_name]
-        if class_name == "OptimizerMap":
-            return OptimizerMap
         globals_dict = globals()
         if class_name in globals_dict:
             return globals_dict[class_name]
--- a/keras/src/optimizers/multi_optimizer.py
+++ b/keras/src/optimizers/multi_optimizer.py
@@ -1,11 +1,9 @@
 import re
 from collections.abc import MutableMapping
 
-from keras.src import ops
 from keras.src.api_export import keras_export
-from keras.src.optimizers import optimizer
+from keras.src.optimizers.optimizer import Optimizer
 from keras.src.saving import serialization_lib
-from keras.src.utils import tracking
 
 
 @keras_export("keras.optimizers.OptimizerMap")
@@ -17,14 +15,18 @@ class OptimizerMap(MutableMapping):
             for any unmapped variables.
     """
 
-    def __init__(self, default_optimizer):
-        if not isinstance(default_optimizer, optimizer.Optimizer):
+    def __init__(self, default_optimizer, optimizer_map=None):
+        if not isinstance(default_optimizer, Optimizer):
             raise TypeError(
                 "default_optimizer must be a Keras Optimizer instance. "
                 f"Received: {default_optimizer}"
             )
+        if optimizer_map is not None and not isinstance(optimizer_map, dict):
+            raise TypeError(
+                f"optimizer_map must be a dictionary. Received: {optimizer_map}"
+            )
         self._default_optimizer = default_optimizer
-        self._optimizer_map = dict()
+        self._optimizer_map = optimizer_map or dict()
 
     @property
     def default_optimizer(self):
@@ -95,30 +97,31 @@ class OptimizerMap(MutableMapping):
             return self._optimizer_map[matching_keys[0]]
         return self.default_optimizer
 
-    def __setitem__(self, key, optim):
+    def __setitem__(self, key, optimizer):
         """Set optimizer for a given variable.
 
         Args:
             key: string representing the variable path or regex pattern
-            optim: Keras Optimizer instance
+            optimizer: Keras Optimizer instance
         """
         if key in self._optimizer_map:
             raise ValueError(
-                f"{key} already exist in the OptimizerMap with "
+                f"'{key}' already exists in the OptimizerMap with "
                 f"value {self._optimizer_map[key]}. Please make sure to "
                 "not use duplicated keys."
             )
         if not isinstance(key, str):
             raise ValueError(f"key must be a string. Received: {key}")
-        if not isinstance(optim, optimizer.Optimizer):
+        if not isinstance(optimizer, Optimizer):
             raise ValueError(
-                f"optim must be a Keras Optimizer instance. Received: {optim}"
+                f"optimizer must be a Keras Optimizer instance. "
+                f"Received: {optimizer}"
             )
-        if isinstance(optim, MultiOptimizer):
+        if isinstance(optimizer, MultiOptimizer):
             raise ValueError(
-                "MultiOptimizer can not be nested inside an OptimizerMap."
+                "MultiOptimizer cannot be nested inside an OptimizerMap."
             )
-        self._optimizer_map[key] = optim
+        self._optimizer_map[key] = optimizer
 
     def __delitem__(self, key):
         del self._optimizer_map[key]
@@ -137,10 +140,9 @@ class OptimizerMap(MutableMapping):
             "default_optimizer": serialization_lib.serialize_keras_object(
                 self._default_optimizer
             ),
-            "optimizer_map": {
-                k: serialization_lib.serialize_keras_object(v)
-                for k, v in self._optimizer_map.items()
-            },
+            "optimizer_map": serialization_lib.serialize_keras_object(
+                self._optimizer_map
+            ),
         }
 
     @classmethod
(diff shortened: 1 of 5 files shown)


keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+        elif len(matching_keys) == 1:
+            return self._optimizer_map[matching_keys[0]]
+        return self.default_optimizer
+
+    def __setitem__(self, key, optim):
+        """Set optimizer for a given variable.
+
+        Args:
+            key: string representing the variable path or regex pattern
+            optim: Keras Optimizer instance
```
hertschuh: Let's call it `optimizer` instead of `optim`, we typically don't abbreviate.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+    def from_config(cls, config, custom_objects=None):
+        default_optimizer = serialization_lib.deserialize_keras_object(
+            config["default_optimizer"], custom_objects=custom_objects
+        )
+        obj_map = cls(default_optimizer)
+        for k, v in config["optimizer_map"].items():
+            obj_map[k] = serialization_lib.deserialize_keras_object(
+                v, custom_objects=custom_objects
+            )
+        return obj_map
```
hertschuh, line 157: Hmm, this is a bit unusual and dissymetrical.

Let's do the same as DTypePolicyMap, i.e., let's add the `optimizer_map` as an optional argument in the constructor.

https://github.com/keras-team/keras/blob/master/keras/src/dtype_policies/dtype_policy_map.py#L88

gaga1313 (commented):
Resolved comments

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+class MultiOptimizer(optimizer.Optimizer):
+    """An optimizer wrapper that delegates variables to different optimizers.
+
+    Initialize the object with an OptimizerMap instance or a callable
+    function that returns an optimizer for a given variable.
+
+    Example:
+        model.compile(
+            optimizer=MultiOptimizer(
+                OptimizerMap(default_optimizer=optimizers.Adam())
```
hertschuh: Just for the sake of correctness, and because in the other comment I asked for allowing the map as a second argument, let's do something like:
```python
OptimizerMap(default_optimizer=optimizers.SGD(), {"encoder/.*": optimizers.Adam()})
```

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+            ),
+            loss="binary_crossentropy",
+        )
+
+        # Or using a callable
+        def optimizer_fn(variable):
+            if "encoder" in variable.path:
+                return optimizers.Adam()
+            else:
+                return optimizers.SGD()
```
hertschuh: The name `optimizer_fn` can be a bit misleading, how about `optimizer_selector` or `map_optimizer` or `get_optimizer`

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+        default_optimizer=optimizers.Adam()
+    ))
+    optimizer['.encoder'] = optimizers.SGD()
+
+    for i in range(optimizer.num_optimizers):
+        optim = optimizer.get_optimizer(i)
+        print(optim.learning_rate)
+        print(optim.iterations)
+        print(optim.loss_scale_factor)
+        ...
```
hertschuh, line 202: We typically don't do that, it forces you to do a `for` loop on an index, which is not the most Pythonic.

Let's provide the optimizers as a list instead.
```
for optimizer in optimizer.optimizers:
```
The property would be `optimizers`.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+
+    The MultiOptimizer class instances will return the `learning_rate`
+    and `iterations` attributes of the first optimizer in the
+    OptimizerMap. In the case of callable function it will return the
```
hertschuh: Let's make it the default for the `OptimizerMap` (see my comment below for why)

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+    attributes of the first optimizer returned by the callable function.
+
+    Note: Optimizer-specific callbacks are only supported for the
+    first optimizer, irrespective of initialization method.
+    For example: ReduceLROnPlateau will only update the learning rate
+    of the first optimizer.
```
hertschuh: Hmm... but that's not user controlled. As a user, you have no idea which variable will be queried first, therefore you have no idea which optimizer will be "the first". You don't even know if it's going to be the same or a different one each time you run the same code.

For `OptimizerMap`, the clear winner is the default optimizer, which, by the way, can also appear one or multiple times in the map. So this let's you control which optimizer you target for the callbacks, even if the fallback logic is never actually used.

For a function, there is no obvious winner. Maybe we need to add this weird contract whereby if you call the callable with `None`, you get the "default" optimizer. It's a little bit weird. The alternative is to check for the `default_optimizer` attribute, which you already do. That means that your callable cannot really be a function but an object implementing `__call__`.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+
+    """
+
+    def __init__(
+        self, optim_map, name="multi_optimizer", loss_scale_factor=None
+    ):
```
hertschuh: - Rename  `optim_map` -> `optimizer_map`.
- Do `name=None`, there is magic to automatically resolve to `"multi_optimizer"` based on the class name. This way, when people subclass this, they'll get their class name automatically for free.
- Reorder `loss_scale_factor` before `name` for consistency with the super class

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+            optim_map: An OptimizerMap instance or a callable function that
+                returns an optimizer for a given variable.
+            name: The name of the optimizer.
+            loss_scale_factor: It overrides the loss_scale_factor passed
+            to the sub-optimizers.
+        """
+        super().__init__(
+            learning_rate=0.0, name=name, loss_scale_factor=loss_scale_factor
+        )
+        self._optim_map = optim_map
```
hertschuh: Rename `self._optim_map` -> `self._optimizer_map`.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+                    self._optimizer_to_vars[opt] = []
+
+        default_opt = getattr(self._optim_map, "default_optimizer", None)
+        if (
+            default_opt is not None
+            and default_opt not in self._inner_optimizers
+        ):
+            default_opt.loss_scale_factor = loss_scale_factor
+            self._inner_optimizers.append(default_opt)
+            self._optimizer_to_vars[default_opt] = []
```
hertschuh: Let's move this up so that the default optimizer is first.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+            )
+
+        new_optimizer_variables = new_own_variables + new_inner_opt_variables
+        return new_trainable_variables, new_optimizer_variables
+
+    def scale_loss(self, loss):
+        if self._inner_optimizers:
+            return self._inner_optimizers[0].scale_loss(loss)
+        else:
+            raise ValueError("MultiOptimizer has no inner optimizers.")
```
hertschuh: I don't think you should do this, the default if you don't override is correct. That's because the `MultiOptimizer` itself has the correct `loss_scale_factor` from the constructor.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+            num_opt_vars = len(opt.variables)
+            if num_opt_vars > 0:
+                opt_weights = weights[idx : idx + num_opt_vars]
+                opt.set_weights(opt_weights)
+                idx += num_opt_vars
+
+    def get_config(self):
+        return {
+            "optim_map": serialization_lib.serialize_keras_object(
+                self._optim_map
```
hertschuh: `"optimizer_map"`

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+            "loss_scale_factor": self.loss_scale_factor,
+        }
+
+    @classmethod
+    def from_config(cls, config, custom_objects=None):
+        return cls(
+            optim_map=serialization_lib.deserialize_keras_object(
+                config.pop("optim_map"), custom_objects=custom_objects
+            ),
+            name=config.pop("name"),
```
hertschuh: You're missing `loss_scale_factor`.

In general, it's safer to pass everything so that the code continues to work as new attributes are added:
```
config = config.copy()
config["optimizer_map"] = serialization_lib.serialize_keras_object(config["optimizer_map"])
return cls(**config)
```

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+            default_opt is not None
+            and default_opt not in self._inner_optimizers
+        ):
+            default_opt.loss_scale_factor = loss_scale_factor
+            self._inner_optimizers.append(default_opt)
+            self._optimizer_to_vars[default_opt] = []
+
+    def _var_key(self, variable):
+        return id(variable)
+
```
hertschuh: Remove, this will do the wrong thing on Tensorflow

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+    @tracking.no_automatic_dependency_tracking
+    def build(self, var_list):
```
hertschuh: Instead of having `self._optimizer_to_vars`, make it a local variable within `built`: `optimizer_to_vars`.

You don't need to initialize it in `__init__`. You can just do this here:
```python
optimizer_to_vars = {opt: [] for opt in self._inner_optimizers}
```

I believe this will allow you to remove ` @tracking.no_automatic_dependency_tracking`

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+            if opt not in self._inner_optimizers:
+                opt.loss_scale_factor = self.loss_scale_factor
+                self._inner_optimizers.append(opt)
+                self._optimizer_to_vars[opt] = []
+            idx = self._inner_optimizers.index(opt)
+            self._var_to_optimizer_idx[self._var_key(var)] = idx
+            self._optimizer_to_vars[opt].append(var)
+
+        # Build all sub-optimizers
+        for opt, variables in self._optimizer_to_vars.items():
```
hertschuh: Actually, per my comment in the previous version of this PR, I'm not a huge fan of using random objects as keys in maps, I'm surprised it works here because I didn't think `Optimizer`s were hashable.

But in this case, you don't need a map, you just need to create lists of variables by index of the optimizer.

So this will become:

```python
optimizer_vars = [[]] * len(self._inner_optimizers)
    ....
    optimizer_vars[idx].append(var)
...
for optimizer, variables in zip(self._inner_optimizers, optimizer_vars):
    ...
```

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+        if idx is None:
+            raise ValueError(f"Variable {variable} not found in any optimizer.")
+        return self._inner_optimizers[idx]
+
+    def get_optimizer(self, index):
+        return self._inner_optimizers[index]
+
+    @property
+    def num_optimizers(self):
+        return len(self._inner_optimizers)
```
hertschuh:
```python
@property
def optimizers(self):
```

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+            self._inner_optimizers[0].learning_rate = value
+        else:
+            super()._learning_rate.assign(value)
+
+    @property
+    def iterations(self):
+        if self._inner_optimizers:
+            return self._inner_optimizers[0].iterations
+        else:
+            return self._iterations
```
hertschuh: Remove, the `MultiOptimizer`'s iteration count is correct.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+            if variables:
+                opt.build(variables)
+
+        super().build(var_list)
+
+    def _get_optimizer_for_variable(self, variable):
+        idx = self._var_to_optimizer_idx.get(self._var_key(variable), None)
+        if idx is None:
+            raise ValueError(f"Variable {variable} not found in any optimizer.")
+        return self._inner_optimizers[idx]
```
hertschuh: Remove, what you really need is the index, see my other comments

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+            self.build(trainable_variables)
+            self.built = True
+
+        # Group gradients and variables by sub-optimizer
+        optimizer_to_grads_and_vars = {
+            opt: [] for opt in self._inner_optimizers
+        }
+        for grad, var in zip(grads, trainable_variables):
+            opt = self._get_optimizer_for_variable(var)
+            optimizer_to_grads_and_vars[opt].append((grad, var))
```
hertschuh: Let's not key by Optimizer.

```python
grads_and_vars = [[]] * len(self._inner_optimizers)
for grad, var in zip(grads, trainable_variables):
    idx = self._var_to_optimizer_idx[self._var_key(variable)]
    grads_and_vars[idx].append((grad, var))
```

Then do a zip

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+        for opt, grads_and_vars in optimizer_to_grads_and_vars.items():
+            if grads_and_vars:
+                sub_grads, sub_vars = zip(*grads_and_vars)
+                opt.apply(list(sub_grads), list(sub_vars))
+
+        # Keep master iterations in sync
+        self._iterations.assign_add(1)
+
+    def stateless_apply(self, optimizer_variables, grads, trainable_variables):
+        if len(grads) == 0:
```
hertschuh: Oh you didn't answer my question for the earlier review:

Do we need this at all? The default implementation of `stateless_apply`, which delegates to `apply` should in theory just work. That would remove a lot of code.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,471 @@
...
+        if self._inner_optimizers:
+            return self._inner_optimizers[0].scale_loss(loss)
+        else:
+            raise ValueError("MultiOptimizer has no inner optimizers.")
+
+    def finalize_variable_values(self, var_list):
+        optimizer_to_vars = {opt: [] for opt in self._inner_optimizers}
+        for var in var_list:
+            opt = self._get_optimizer_for_variable(var)
+            optimizer_to_vars[opt].append(var)
```
hertschuh: Same thing, don't key by Optimizer, build a list of lists and zip

hertschuh (changes requested):

keras/src/optimizers/multi_optimizer.py
gaga1313: Default optimizer can appear zero times in the map too, incase when all variables are mapped, it might not get used.

Yeah its tricky for callable function, if we don't simply return first optimizer attributes.

keras/src/optimizers/multi_optimizer.py
gaga1313: Will wait for the previous comment to be resolved.

keras/src/optimizers/multi_optimizer.py
gaga1313: We will need it for distributing the optimizer_variables, grads, and trainable_variables to each optimizer. Later we anyways delegate to `opt.stateless_apply`

gaga1313: I tried reducing the lines of code, but couldn't do much.

keras/src/optimizers/multi_optimizer.py
gaga1313: Good catch. Thanks!

gaga1313 (commented):
answered comments

keras/src/optimizers/multi_optimizer.py
hertschuh: But can you try just removing it and see if the tests pass.

It should just work, there is some stateless magic to make `apply` and the delegation the the sub-`applies` work.

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,399 @@
...
+            else:
+                return optimizers.SGD()
+
+        model.compile(
+            optimizer=MultiOptimizer(optimizer_selector),
+            loss="binary_crossentropy",
+        )
+
+    To access the attributes of the sub-optimizers, iterate over the
+    optimizers using `get_optimizer(i)`:
```
hertschuh: Update to say "using .optimizers"

keras/src/optimizers/multi_optimizer.py
```diff
@@ -0,0 +1,399 @@
...
+                opt.build(variables)
+
+        super().build(var_list)
+
+    def get_optimizer(self, index):
+        return self._inner_optimizers[index]
+
+    @property
+    def num_optimizers(self):
+        return len(self._inner_optimizers)
```
hertschuh: Remove these, they are no longer needed, and this is not a pattern we ever use.

keras/src/optimizers/multi_optimizer_test.py
```diff
@@ -0,0 +1,285 @@
...
+
+        # w1: updated by opt_sgd_1 (lr=0.1) -> 2.0 - 0.1 * 1.0 = 1.9
+        self.assertAllClose(w1.numpy(), [[1.9, 1.9]])
+        # w2: updated by opt_sgd_2 (lr=1.0) -> 3.0 - 1.0 * 1.0 = 2.0
+        self.assertAllClose(w2.numpy(), [[2.0, 2.0]])
+        # w_fallback: updated by default_opt (lr=0.0) -> Remains 4.0
+        self.assertAllClose(w_fallback.numpy(), [[4.0, 4.0]])
+
+        # Verify iteration increment
+        self.assertEqual(int(multi_opt.iterations), 1)
```
hertschuh, line 134: Can you verify the iterations of the other optimizers?

It's relevant when a learning rate schedule is used.

keras/src/optimizers/multi_optimizer_test.py
```diff
@@ -0,0 +1,285 @@
...
+        self.assertEqual(reconstructed.num_optimizers, 3)
+        self.assertEqual(
+            reconstructed.get_optimizer(0).__class__.__name__, "Adam"
+        )
+        self.assertEqual(
+            reconstructed.get_optimizer(1).__class__.__name__, "SGD"
+        )
+        self.assertEqual(
+            reconstructed.get_optimizer(2).__class__.__name__, "RMSprop"
+        )
```
hertschuh: You'll have to use `.optimizers`.

keras/src/optimizers/multi_optimizer_test.py
```diff
@@ -0,0 +1,285 @@
+from keras.src import backend
+from keras.src import optimizers
+from keras.src import testing
+
+
+class MultiOptimizerTest(testing.TestCase):
```
hertschuh, line 11: Can you add an end-to-end test with `model.fit`?

It can look similar to this:
https://github.com/keras-team/keras/blob/master/keras/src/optimizers/optimizer_test.py#L125

You can verify the values of the model variable at the end with low precision. This will exercise jit compilation and other things.

hertschuh (commented):
Thanks!

keras/src/optimizers/multi_optimizer_test.py
gaga1313: added a new test `test_multi_optimizer_with_model_fit`.

keras/src/optimizers/multi_optimizer_test.py
gaga1313: added for loop validating all optimizer iterations.

gaga1313 (commented):
resolved comments

hertschuh (approved):
Thanks!

PR #22905, opened 2026-05-15
Add keras.config.enable_tf32 toggle for cross-backend matmul precision

MarcosAsh:
## Description

Keras has no portable knob for matmul precision. Today, controlling TF32 requires reaching into each backend's API: `torch.set_float32_matmul_precision` + `torch.backends.cudnn.allow_tf32`, `jax.config.update("jax_default_matmul_precision", ...)`, or `tf.config.experimental.enable_tensor_float_32_execution(...)`. A multi-backend Keras program needs three different calls to get consistent behavior.

This PR adds `keras.config.enable_tf32` / `disable_tf32` / `is_tf32_enabled`, mirroring the `flash_attention` helper trio. It's opt-in: Keras does not touch any backend's default unless the user explicitly calls one of these. Calling `enable_tf32()` once opts the active backend into TF32 (Torch: `"high"` + `cudnn.allow_tf32 = True`; JAX: `"default"` matmul precision; TensorFlow: `enable_tensor_float_32_execution(True)`).

### Why this matters

On a small MoE training step on NVIDIA A100-SXM4-40GB (torch 2.10.0+cu128, 8 experts top-2, B=16, L=512, D=512, `train_on_batch`):

| | TF32 off | TF32 on | speedup |
|---|---|---|---|
| MoE warm step | 99.9 ms | 24.9 ms | **4.01x** |
| Self CUDA total / 10 steps | 1.003 s | 248 ms | 4.04x |

TF32-on lands at parity with the JAX backend (~24.1 ms) and the TensorFlow backend (~25.0 ms) on the same model. Profiler evidence is unambiguous: with TF32 off the top GEMM kernels are `ampere_sgemm_*` / `cutlass_80_simt_sgemm_*` (CUDA cores, fp32); with TF32 on they become `cutlass_80_tensorop_s1688gemm_*` (tensor cores, TF32). Reproduction: [Colab](https://colab.research.google.com/drive/15nUbtaEicw9SoLQ-n-07lKGZ7448O8MM?usp=sharing).

## Contributor Agreement

### DIFF
--- a/keras/src/backend/config.py
+++ b/keras/src/backend/config.py
@@ -288,6 +288,11 @@ def is_tf32_enabled():
 
     Returns:
         `False` if disabled. Otherwise, it indicates that it is enabled.
+
+    Example:
+
+    >>> keras.config.is_tf32_enabled()
+    True
     """
     from keras.src.backend.common import global_state
 
--- a/keras/src/backend/config.py
+++ b/keras/src/backend/config.py
@@ -248,6 +248,10 @@ def enable_tf32():
 
     Note that this only affects `float32` compute on supported GPUs. It has
     no effect on CPU, on TPU, or for other dtypes.
+
+    Example:
+
+    >>> keras.config.enable_tf32()
     """
     from keras.src.backend.common import global_state
 
--- a/keras/src/backend/config.py
+++ b/keras/src/backend/config.py
@@ -233,6 +233,80 @@ def is_flash_attention_enabled():
     return global_state.get_global_attribute("flash_attention", default=None)
 
 
+@keras_export("keras.config.enable_tf32")
+def enable_tf32():
+    """Enable TensorFloat-32 (TF32) for float32 matrix multiplications.
+
+    TF32 is a reduced-precision mode available on NVIDIA Ampere and later
+    GPUs that runs `float32` matmuls and convolutions on the tensor cores,
+    typically several times faster with a small loss of mantissa precision.
+
+    This is enabled by default. The JAX and TensorFlow backends already use
+    TF32 by default. The Torch backend does not, so without this Keras would
+    silently run `float32` matmuls slower on Torch than on the other
+    backends for the same model code.
+
+    Note that this only affects `float32` compute on supported GPUs. It has
+    no effect on CPU, on TPU, or for other dtypes.
+    """
+    from keras.src.backend.common import global_state
+
+    global_state.set_global_attribute("tf32", None)
+    _refresh_backend_tf32()
+
+
+@keras_export("keras.config.disable_tf32")
+def disable_tf32():
+    """Disable TensorFloat-32 (TF32) for float32 matrix multiplications.
+
+    Once disabled, `float32` matmuls run at full `float32` precision on all
+    backends. Use this when you need bit-for-bit `float32` accuracy and are
+    willing to give up the tensor-core speedup.
+    """
+    from keras.src.backend.common import global_state
+
+    global_state.set_global_attribute("tf32", False)
+    _refresh_backend_tf32()
+
+
+@keras_export("keras.config.is_tf32_enabled")
+def is_tf32_enabled():
+    """Checks whether TF32 is globally enabled in Keras.
+
+    TF32 lets `float32` matrix multiplications run on GPU tensor cores at
+    reduced mantissa precision for a large speedup. This function checks the
+    global Keras configuration. See `keras.config.enable_tf32` and
+    `keras.config.disable_tf32`.
+
+    Returns:
+        `False` if disabled. Otherwise, it indicates that it is enabled.
+    """
+    from keras.src.backend.common import global_state
+
+    return global_state.get_global_attribute("tf32", default=None)
+
+
+def _refresh_backend_tf32():
+    """Re-apply the current TF32 setting to the active backend.
+
+    The backend applies TF32 once at import time; this lets a later
+    `enable_tf32()` / `disable_tf32()` call take effect immediately.
+    """
+    name = backend()
+    try:
+        if name == "torch":
+            from keras.src.backend.torch.core import _apply_tf32
+        elif name == "jax":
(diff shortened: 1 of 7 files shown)


keras/src/backend/torch/core.py
```diff
@@ -820,3 +820,23 @@ def backward(ctx, grad_output):
...
+    `torch.set_float32_matmul_precision` keeps the Torch backend consistent
+    with the others. Applied once at import and re-applied by
+    `keras.config.enable_tf32` / `disable_tf32`.
+    """
+    from keras.src.backend.config import is_tf32_enabled
+
+    if is_tf32_enabled() is False:
+        torch.set_float32_matmul_precision("highest")
+    else:
+        torch.set_float32_matmul_precision("high")
```
amitsrivastava78, line 852: This silently changes the default numerics for every existing Torch-backend Keras user on upgrade. Torch's own default is "highest" (full FP32); after this PR they get TF32 without opting in. Numeric regression tests, golden-value tests, reproducibility scripts, and downstream model evals will all see drift.
Two options:
1. Preferred: make the default backend-dependent — keep each backend's existing default (JAX/TF on, Torch off) and let keras.config.enable_tf32() be the explicit cross-backend opt-in for users who want consistent fast-matmul behavior.
2. If the team decides to flip Torch's default to match JAX/TF, this needs:

              a) A "Backwards-incompatible change" section in the PR description
              b) Release-notes call-out
              c) A one-time RuntimeWarning on first matmul under Torch the first time the default flipped, so silent numerical drift doesn't go unnoticed
Default-on with no warning is the riskier path.

keras/src/backend/config.py
```diff
@@ -233,6 +233,93 @@ def is_flash_attention_enabled():
...
+        `False` if disabled. Otherwise, it indicates that it is enabled.
+
+    Example:
+
+    >>> keras.config.is_tf32_enabled()
+    True
+    """
+    from keras.src.backend.common import global_state
+
+    return global_state.get_global_attribute("tf32", default=None)
```
amitsrivastava78: This returns False, None, or True depending on state. The docstring even hedges: "False if disabled. Otherwise, it indicates that it is enabled." Tests reflect the issue — they use self.assertNotEqual(..., False) instead of assertTrue(...) because the latter would fail on None.
The other is_* accessors in keras.config (is_flash_attention_enabled, is_nnx_enabled) return strict bool. Suggest the same here:
val = global_state.get_global_attribute("tf32", default=True)
return val is not False
Easier to fix now than after the API is public.

keras/src/backend/config.py
```diff
@@ -233,6 +233,93 @@ def is_flash_attention_enabled():
...
+    Note that this only affects `float32` compute on supported GPUs. It has
+    no effect on CPU, on TPU, or for other dtypes.
+
+    Example:
+
+    >>> keras.config.enable_tf32()
+    """
+    from keras.src.backend.common import global_state
+
+    global_state.set_global_attribute("tf32", None)
```
amitsrivastava78: This is the source of the tristate problem flagged on is_tf32_enabled — the "enabled" state is stored as None, "disabled" as False, and the implicit "default" is also None. The user-visible accessor can't distinguish "default-on" from "explicitly-enabled" because both are None.
Set this to True here and False in disable_tf32, and have is_tf32_enabled return the stored bool with True as the default. The None sentinel is needlessly clever and leaks into the public return type.

keras/src/backend/torch/core.py
```diff
@@ -820,3 +820,23 @@ def backward(ctx, grad_output):
...
+
+    Torch defaults to full `float32` matmuls (`"highest"`), while JAX and
+    TF use TF32 by default. Mapping the Keras setting onto
+    `torch.set_float32_matmul_precision` keeps the Torch backend consistent
+    with the others. Applied once at import and re-applied by
+    `keras.config.enable_tf32` / `disable_tf32`.
+    """
+    from keras.src.backend.config import is_tf32_enabled
+
+    if is_tf32_enabled() is False:
```
amitsrivastava78: torch.set_float32_matmul_precision("high") only controls matmul TF32. Convolutions are gated by a separate switch: torch.backends.cudnn.allow_tf32, which defaults to False on Torch. So after this PR, Torch's float32 matmuls run on tensor cores but float32 convolutions still run full-precision on CUDA cores. JAX and TF apply TF32 to both by default — so this PR closes the matmul gap but leaves the conv gap open.
Either:
1. Apply both flags:
enabled = is_tf32_enabled() is not False
torch.set_float32_matmul_precision("high" if enabled else "highest")
torch.backends.cudnn.allow_tf32 = enabled

2. Or narrow the docstring claim "TF32 ... runs float32 matmuls and convolutions on the tensor cores" to matmuls only on Torch, and note convs are unchanged.

keras/src/backend/torch/core.py
MarcosAsh: Reverted to opt-in. enable_tf32 is now an explicit cross-backend switch and Keras does not touch any backend's default. Doing nothing leaves each backend at its native default.

keras/src/backend/config.py
MarcosAsh: Now a strict bool. Returns False by default and only True after enable_tf32 has been called explicitly.

keras/src/backend/config.py
MarcosAsh: Resolved by the same change. enable_tf32 stores True, disable_tf32 stores False, and the unset state is treated as off in the public return type.

keras/src/backend/torch/core.py
MarcosAsh: Wired both. enable_tf32 sets torch.backends.cudnn.allow_tf32 = True alongside set_float32_matmul_precision, and disable_tf32 sets both back to off.

MarcosAsh:
Hi, thanks for the review! Shipped this as opt-in instead of flipping the Torch default, and applied the other fixes inline.

PR #22910, opened 2026-05-16
Engage flash SDPA in MHA/GQA causal-only path

MarcosAsh:
## Description

`MultiHeadAttention(use_causal_mask=True)` today builds an explicit `[1, T, S]` causal mask and passes it to `dot_product_attention(..., mask=mask, is_causal=False)`. On torch this disqualifies the SDPA flash kernel (flash does not accept arbitrary masks) and dispatch falls to mem-efficient cutlass. Passing `is_causal=True` with no mask engages flash directly.

This PR routes through the `is_causal=True` path when causal masking is the only mask source. Anything with extra masks (key/query/value/attention mask, sliding window) keeps the existing path. Same change applied to `GroupedQueryAttention`.

### Benchmark (A100, torch 2.10.0+cu128, B=4, H=12, D=64, fp16, MHA forward)

| T | master | this PR | speedup |
|---:|---:|---:|---:|
| 1024 | 1.92 ms | 1.75 ms | 1.10x |
| 2048 | 1.92 ms | 1.74 ms | 1.11x |
| 4096 | 4.62 ms | **1.76 ms** | **2.62x** |

Profiler: master runs `fmha_cutlassF_f16_aligned_64x64_rf_sm80`, this PR runs `pytorch_flash::flash_fwd_kernel`. Outputs match to fp16 noise (max |d| ≤ 1e-3). Reproduction: [Colab](https://colab.research.google.com/drive/1nUJpesIOfPBXi1kDTOq9II8BFklCO9HS?usp=sharing).

## Contributor Agreement

### DIFF
--- a/keras/src/layers/attention/grouped_query_attention_test.py
+++ b/keras/src/layers/attention/grouped_query_attention_test.py
@@ -400,6 +400,35 @@ class GroupedQueryAttentionTest(testing.TestCase):
                 else:
                     self.assertEqual(scores[i, j], 0.0)
 
+    def test_causal_only_fast_path(self):
+        # When `use_causal_mask=True` is the only mask source, the layer
+        # should route through `dot_product_attention(is_causal=True)` and
+        # not materialize a [T, S] mask. The output must match the explicit
+        # mask path.
+        layer = layers.GroupedQueryAttention(
+            head_dim=4, num_query_heads=4, num_key_value_heads=2
+        )
+        x = np.random.RandomState(0).randn(2, 5, 16).astype("float32")
+        _ = layer(x, x, use_causal_mask=True)
+        fast = layer(x, x, use_causal_mask=True)
+
+        t = x.shape[1]
+        causal = np.tril(np.ones((1, t, t), dtype="bool"))
+        slow = layer(x, x, attention_mask=causal)
+
+        self.assertAllClose(fast, slow, atol=1e-5)
+
+    def test_causal_only_with_other_mask_falls_back(self):
+        # If any other mask is also present, the is_causal=True shortcut
+        # must not be taken since SDPA rejects mask + is_causal together.
+        layer = layers.GroupedQueryAttention(
+            head_dim=4, num_query_heads=4, num_key_value_heads=2
+        )
+        x = np.random.RandomState(0).randn(2, 5, 16).astype("float32")
+        extra_mask = np.ones((2, 5, 5), dtype="bool")
+        out = layer(x, x, attention_mask=extra_mask, use_causal_mask=True)
+        self.assertEqual(tuple(out.shape), (2, 5, 16))
+
     def test_sliding_window_validation(self):
         with self.assertRaisesRegex(ValueError, "sliding_window"):
             layers.GroupedQueryAttention(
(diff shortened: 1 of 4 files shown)


keras/src/layers/attention/grouped_query_attention_test.py
```diff
@@ -400,6 +400,54 @@ def test_sliding_window(self):
...
+        # If any other mask is also present, the is_causal=True shortcut
+        # must not be taken since SDPA rejects mask + is_causal together.
+        layer = layers.GroupedQueryAttention(
+            head_dim=4, num_query_heads=4, num_key_value_heads=2
+        )
+        x = np.random.RandomState(0).randn(2, 5, 16).astype("float32")
+        extra_mask = np.ones((2, 5, 5), dtype="bool")
+        out = layer(x, x, attention_mask=extra_mask, use_causal_mask=True)
+        self.assertEqual(tuple(out.shape), (2, 5, 16))
+
```
hertschuh: Hmm... but this doesn't verify that the `extra_mask` is not completely ignored.

Maybe do an assertAllClose with:
```python
out1 = layer(x, x, attention_mask=extra_mask, use_causal_mask=True)
```
and
```python
out2 = ayer(x, x, attention_mask=extra_mask & causal, use_causal_mask=False)
```

keras/src/layers/attention/multi_head_attention_test.py
```diff
@@ -523,6 +523,68 @@ def test_sliding_window(self):
...
+        # If any other mask source is present, the layer must NOT take the
+        # is_causal=True shortcut, because torch's SDPA rejects passing both
+        # an attn_mask and is_causal=True.
+        layer = layers.MultiHeadAttention(num_heads=2, key_dim=4)
+        x = np.random.RandomState(0).randn(2, 5, 8).astype("float32")
+        # Build with the slow path so the test exercises the merged path.
+        extra_mask = np.ones((2, 5, 5), dtype="bool")
+        out = layer(x, x, attention_mask=extra_mask, use_causal_mask=True)
+        self.assertEqual(tuple(out.shape), (2, 5, 8))
+
```
hertschuh: Hmm... but this doesn't verify that the `extra_mask` is not completely ignored.

Maybe do an assertAllClose with:
```python
out1 = layer(x, x, attention_mask=extra_mask, use_causal_mask=True)
```
and
```python
out2 = ayer(x, x, attention_mask=extra_mask & causal, use_causal_mask=False)
```

hertschuh (changes requested):
Thanks for adding this optimization!

keras/src/layers/attention/grouped_query_attention_test.py
MarcosAsh: I didn't think of this, thanks. Replaced the all-ones extra_mask with a non-trivial one and added an assertAllClose against the explicit merged-mask path so a regression that silently engages the causal-only shortcut while a mask is present would fail the test.

keras/src/layers/attention/multi_head_attention_test.py
MarcosAsh: Same fix applied here.

MarcosAsh:
Hi, thanks for the review! I made all the changes necessary.

hertschuh (approved):
Thanks!

PR #22911, opened 2026-05-16
Add keras.config.enable_jit_cache for JAX persistent compilation cache

MarcosAsh:
## Description

JAX recompiles every jit-decorated function on first call, which for a typical Keras model is 8-30 seconds of wall time before the first `fit` step begins. JAX has a persistent disk cache for compiled HLO, but enabling it requires `jax.config.update("jax_compilation_cache_dir", path)` plus two threshold flags. Most Keras users do not know about it.

This PR adds `keras.config.enable_jit_cache(path)` / `disable_jit_cache()` / `is_jit_cache_enabled()` mirroring the `flash_attention` helper trio. The path defaults to `~/.cache/keras/jit_cache`. Only the JAX backend is affected. Other backends are no-ops.

### Benchmark (Colab A100, JAX 0.7.2, small Transformer with MHA + LayerNorm + Dense + GAP)

Self-contained script invoked twice from python, identical model code, only the cache setting differs. First `fit` step wall time:

| | run 1 | run 2 |
|---|---:|---:|
| cache off (control) | 8.09 s | 8.10 s |
| cache on            | 8.43 s | **1.54 s** |

Run 2 with the cache enabled drops 5.5x because the compiled HLO is read from disk instead of re-traced. The control row confirms the speedup is the disk cache and not Python warmup. Cache directory after run 1 is populated with the expected jit-keyed entries. Reproduction: [Colab](https://colab.research.google.com/drive/1HA-9mrCNdJ-V9SbIkZCXPfUmP0R1r1dW?usp=sharing).

## Contributor Agreement


keras/src/backend/config.py
```diff
@@ -233,6 +233,109 @@ def is_flash_attention_enabled():
...
+    >>> keras.config.enable_jit_cache("/tmp/my_jit_cache")
+    >>> keras.config.is_jit_cache_enabled()
+    '/tmp/my_jit_cache'
+    """
+    from keras.src.backend.common import global_state
+
+    val = global_state.get_global_attribute("jit_cache_dir", default=None)
+    if val is False or val is None:
+        return False
+    return val
```
amitsrivastava78: is_jit_cache_enabled() returns either False or the cache path string. The rest of keras.config's is_* accessors (is_flash_attention_enabled, is_nnx_enabled) return bool. Mixed-type return breaks == True comparisons and is surprising by name. Suggest splitting into:

is_jit_cache_enabled() -> bool
get_jit_cache_dir() -> str | None
Easier to change now than after the API ships.

keras/src/backend/config.py
```diff
@@ -233,6 +233,109 @@ def is_flash_attention_enabled():
...
+    and is re-applied the next time `enable_jit_cache` is called.
+    """
+    if backend() != "jax":
+        return
+    try:
+        import jax
+    except ImportError:
+        return
+    if path is None:
+        jax.config.update("jax_compilation_cache_dir", "")
```
amitsrivastava78: enable_jit_cache sets jax_persistent_cache_min_compile_time_secs=0 and jax_persistent_cache_min_entry_size_bytes=0 (lines 335-336). disable_jit_cache only clears jax_compilation_cache_dir; the thresholds stay at 0 for the lifetime of the process. If a user later sets jax_compilation_cache_dir directly (outside Keras), they unexpectedly get aggressive caching they didn't ask for. Either reset both thresholds here when path is None, or document the residual state.

keras/src/backend/config.py
```diff
@@ -233,6 +233,109 @@ def is_flash_attention_enabled():
...
+        return
+    jax.config.update("jax_compilation_cache_dir", path)
+    jax.config.update("jax_persistent_cache_min_compile_time_secs", 0)
```
amitsrivastava78: Setting both thresholds to 0 means every JIT compile — even trivial 1-op functions — is persisted. Over time this bloats the cache directory with low-value entries. Two options:

Leave the thresholds at JAX defaults and only set the directory.
Expose them as kwargs on enable_jit_cache(path=None, min_compile_time_secs=0, min_entry_size_bytes=0) so users can dial back without dropping to raw jax.config.update(...).
Option 2 preserves current behavior as the default but gives users an escape hatch.

keras/src/backend/config_test.py
```diff
@@ -0,0 +1,53 @@
...
+        self.assertFalse(is_jit_cache_enabled())
+
+    def test_enable_default_path(self):
+        enable_jit_cache()
+        path = is_jit_cache_enabled()
+        self.assertTrue(path)
+        self.assertEqual(path, os.path.join(keras_home(), "jit_cache"))
+
+    def test_enable_custom_path(self):
+        enable_jit_cache("/tmp/keras_jit_cache_test")
```
amitsrivastava78: Hardcoded /tmp/keras_jit_cache_test is used here and in test_disable_after_enable and test_jax_runtime_picks_up_path. Issues:
1. Fails on Windows runners
2. Can collide across parallel test runs on the same machine.
3. Leaves disk state behind on failure (no cleanup).
Use tempfile.mkdtemp() in setUp and register shutil.rmtree(..., ignore_errors=True) via self.addCleanup.

keras/src/backend/config.py
MarcosAsh: is_jit_cache_enabled now returns a strict bool and get_jit_cache_dir returns the path or None.

keras/src/backend/config.py
MarcosAsh: Dropped the threshold overrides entirely so there is nothing residual to reset. Sticking with JAX's defaults keeps the API a one-liner and lets users tune via jax.config.update if they need to.

keras/src/backend/config.py
MarcosAsh: Removed the threshold overrides. JAX's defaults already cache the expensive compiles and skip the trivial ones, which is the use case here.

keras/src/backend/config_test.py
MarcosAsh: Switched to tempfile.mkdtemp with addCleanup so each test gets its own directory and cleans up on failure.

MarcosAsh:
Hi, thanks for the review! I applied all the changes.

hertschuh (commented):
```python
keras.config.enable_jit_cache(path)
```
is just an abstraction for
```python
jax.config.update("jax_compilation_cache_dir", path)
```

As such, I currently don't see a good reason to add it as a Keras API at this point:
- it doesn't make it easier for users (if anything, it's a layer of obfuscation)
- it only works on JAX, and there is an expectation that most things work across backends, especially as we claim to support jit compilation for JAX, TF and Torch

That being said, it would be nice to have users be aware that this JAX feature exists and that it works seamlessly with Keras.

This could be one of or both of:
- a guide on [keras.io](https://keras.io/guides/)
- a note on one of the APIs, maybe within the docstring of `compile`, in the section about `jit_compile`.

MarcosAsh:
Hi, thanks for the review! Agreed, it's a thin wrapper and JAX-only. I dropped the API and added a note under jit_compile in compile() instead, with the one-liner to enable the cache. Can do a keras.io guide too if useful.

hertschuh (approved):
Thanks!

I hope this renders correctly on keras.io.

PR #22913, opened 2026-05-17
Add fused bidirectional GRU for torch and harden gru() dispatch

MarcosAsh:
## Description

Adds a fused `bidirectional_gru` to the torch backend so `Bidirectional(GRU)` runs a single `torch._VF.gru(bidirectional=True)` call instead of two unidirectional cuDNN dispatches. Closes the asymmetry with BiLSTM, which has had a fused path since #22791 / #22874. Other backends raise `NotImplementedError` (matches `bidirectional_lstm`).

Also hardens the unidirectional `gru()` dispatch with the same fixes #22874 applied to `lstm()`: it now keys off `inputs.device` instead of the global default, skips cuDNN under TorchScript / Dynamo trace, and falls through to a new `_fallback_gru` rather than raising `NotImplementedError`. Unlike the LSTM case, the cuDNN path for unidirectional GRU was already engaged on master, so this is hardening rather than a regression fix.

### Benchmark (Colab A100, torch 2.10.0+cu128)

Two-layer `Bidirectional(GRU(256))`, batch 64, seq 256, features 128, steady-state `model.fit` epoch.

|        | master  | this PR | speedup |
|--------|---------|---------|---------|
| epoch  | 1.591 s | 1.478 s | 1.08x   |

Dispatch trace confirms `Bidirectional(GRU)` makes one fused `_VF.gru` call on the branch vs two on master.

### Tests

`bidirectional_test.py` gains `test_fused_gru_eligibility` and `test_fused_gru_matches_unfused` mirroring the BiLSTM coverage. Existing rnn tests still pass on torch, jax, tf, numpy.

## Contributor Agreement

### DIFF
--- a/keras/src/backend/torch/rnn.py
+++ b/keras/src/backend/torch/rnn.py
@@ -2,7 +2,6 @@ import torch
 
 from keras.src import tree
 from keras.src.backend.torch.core import convert_to_tensor
-from keras.src.backend.torch.core import get_device
 
 
 def rnn(
@@ -747,18 +746,11 @@ def gru(
     unroll=False,
     reset_after=True,
 ):
-    cudnn_supported = cudnn_ok(
-        activation,
-        recurrent_activation,
-        unroll,
-        use_bias=bias is not None,
-    )
-
-    if not cudnn_supported or not reset_after or mask is not None:
+    # Neither path below supports masking. The fallback also only knows the
+    # `reset_after=True` formulation. For anything else, bounce back to the
+    # generic RNN loop.
+    if not reset_after or mask is not None:
         raise NotImplementedError
-
-    # Get device from inputs
-    device = get_device()
 
     # Convert to torch tensors (convert_to_tensor unwraps Variables)
     kernel = convert_to_tensor(kernel)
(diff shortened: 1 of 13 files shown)


keras/src/backend/torch/rnn.py
```diff
@@ -878,6 +900,61 @@ def _cudnn_gru(
...
+def _fallback_gru(
+    inputs,
+    initial_state,
+    kernel,
+    recurrent_kernel,
+    bias,
+    activation,
+    recurrent_activation,
+    return_sequences,
+):
```
hertschuh, line 918: Why this?

Is this implementation noticeably faster than the keras ops based fallback?

keras/src/backend/torch/rnn.py
```diff
@@ -1033,3 +1110,121 @@ def bidirectional_lstm(
...
+    device = inputs.device
+    if (
+        device.type != "cuda"
+        or torch.jit.is_tracing()
+        or (
+            hasattr(torch.compiler, "is_compiling")
+            and torch.compiler.is_compiling()
+        )
+    ):
+        raise NotImplementedError
```
hertschuh: Move this up before doing all the "convert_to_tensor" call, which may be discarded.

hertschuh (changes requested):
Thanks for adding this.

I have a question about the unit tests, the `try` / `except` blocks,  and the fallback mechanism.

The issue is that if this CUDA implementation becomes broken (which has been the case in the past for GRU and LSTM), it will silently fall back to the fallback implementation and unit tests will succeed.

Any way to detect that?

keras/src/backend/torch/rnn.py
MarcosAsh: Yes, measured on A100 with torch 2.12 it is about 3x faster than the generic ops loop across the shapes I tried, roughly 20 ms vs 60 ms at B=32 T=64 U=256 and 77 ms vs 243 ms at B=16 T=256 U=128. Keeping it for that reason.

keras/src/backend/torch/rnn.py
MarcosAsh: Done, moved the device and cudnn_ok eligibility check above the convert_to_tensor calls in both gru and bidirectional_gru.

MarcosAsh:
Hi, thanks for the review! That is exactly what bit the LSTM dispatch before 22874 and went unnoticed for weeks. Added a test that spies on torch._VF.gru and asserts it actually fires on CUDA so a future cuDNN regression breaks loudly instead of silently routing through the fallback.

hertschuh (approved):
Great, thanks!

PR #22915, opened 2026-05-17
[OpenVINO]  Fix dynamic batch shape propagation through ops.slice

goyaladitya05:
## Description
`ops.slice` did `if length is not None and length >= 0`. On a OpenVINOKerasTensor, length >= 0 produces a boolean OpenVINOKerasTensor, and __bool__ attempts numerical evaluation at graph-build time, crashing because the tensor has no concrete value.

This PR skips the static >= 0 check for symbolic lengths and let OV resolve them at runtime. This PR also reverts #22681, since that workaround is no longer needed.
Part of the fix for dynamic batch shape propagation, alongside #22747.

## Contributor Agreement

### DIFF
--- a/keras/src/backend/openvino/core.py
+++ b/keras/src/backend/openvino/core.py
@@ -1390,7 +1390,9 @@ def slice(inputs, start_indices, shape):
         return val
 
     for idx, length in enumerate(shape):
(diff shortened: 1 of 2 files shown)


hertschuh (approved):
Thanks for the fix!

PR #22919, opened 2026-05-18
Raise on zero-size pooling outputs when pool_size > input spatial dim

### LINKED ISSUE #22916: [Bug] Pooling layers/ops produce zero-size outputs silently when pool_size > input spatial dims (analog of #22414/#22416 for Conv)
### Summary

`keras.layers.MaxPooling*`, `keras.layers.AveragePooling*`, and the underlying `compute_pooling_output_shape` helper silently produce an output with a `0`-sized spatial dimension when `pool_size` is larger than the input spatial size. The downstream model then either crashes with a confusing error or runs on an empty tensor.

This is the same family of bug as #22414 (Conv1D) and #22416 (Conv2D), which were fixed in #22418 by changing the validation check in `compute_conv_output_shape` from `< 0` to `<= 0` so that zero-size outputs also raise. The same fix was not applied to `compute_pooling_output_shape`.

### Reproduction

```python
import os
import keras
import numpy as np

# 2D pooling
layer = keras.layers.MaxPooling2D(pool_size=(20, 20))
x_sym = keras.Input(shape=(5, 5, 3))
print(layer(x_sym).shape)                        # (None, 0, 0, 3)  - silent
print(layer(np.zeros((1, 5, 5, 3), "float32")).shape)  # (1, 0, 0, 3)  - silent

# 1D pooling
layer = keras.layers.AveragePooling1D(pool_size=20)
print(layer(keras.Input(shape=(5, 3))).shape)   # (None, 0, 3)  - silent

# 3D pooling
layer = keras.layers.MaxPooling3D(pool_size=(5, 5, 5))
print(layer(keras.Input(shape=(3, 3, 3, 4))).shape)  # (None, 0, 0, 0, 4)  - silent
```

### Expected

A clear `ValueError` (mirroring the Conv message after #22418):

```
ValueError: Computed output size would be zero or negative. Received:
inputs.shape=(None, 5, 5, 3), pool_size=(20, 20), strides=(20, 20),
padding=valid.
```

### Root cause

`keras/src/ops/operation_utils.compute_pooling_output_shape` at the `padding == \"valid\"` branch:

```python
if i not in none_dims and output_spatial_shape[i] < 0:
    raise ValueError("Computed output size would be negative...")
```

This guard only fires for *strictly negative* outputs, but with `pool_size=20`, `input=5`, `strides=20`:

```
floor((5 - 20) / 20) + 1 = floor(-0.75) + 1 = -1 + 1 = 0
```

Zero passes the `< 0` check. The Conv equivalent at `compute_conv_output_shape:220` already uses `<= 0` and the corresponding error string \"zero or negative\". Aligning the pooling helper restores parity.

### Impact

- Models built with `pool_size > input` are accepted at construction time, produce empty tensors, and either crash downstream or silently train on garbage.
- The Conv family was already fixed; the matching pooling case was missed.

rstar327:
## Description

Fixes #22916.

`keras.layers.MaxPooling*` and `keras.layers.AveragePooling*` silently produce an output with a `0`-sized spatial dimension when `pool_size` exceeds the input spatial size — both in the symbolic and the eager paths. Downstream layers either crash with a confusing error or operate on an empty tensor.

This is the same family of bug as #22414 / #22416 for Conv, fixed in #22418 by changing the guard in `compute_conv_output_shape` from `< 0` to `<= 0`. The matching pooling case was missed at the time.

### Before

```python
import keras

keras.layers.MaxPooling2D(pool_size=(20, 20))(keras.Input(shape=(5, 5, 3))).shape
# -> (None, 0, 0, 3)  — silent
```

### After

```
ValueError: Computed output size would be zero or negative. Received:
`inputs.shape=[None 5 5 3]`, `pool_size=[20 20]`, `strides=(20, 20)`,
`padding=valid`.
```

Same fix applies to `MaxPooling1D`, `MaxPooling3D`, `AveragePooling1D`, `AveragePooling3D`. Dynamic spatial dimensions (`None`) still pass through unvalidated, matching the existing Conv behavior.

### Implementation

- `keras/src/ops/operation_utils.compute_pooling_output_shape`: change `output_spatial_shape[i] < 0` → `<= 0`, and update the error message to match the wording introduced for `compute_conv_output_shape` in #22418 (\"zero or negative\", with `strides` and `padding` included for context).
- `keras/src/ops/operation_utils_test.py`: add `test_compute_pooling_output_shape_rejects_zero_output` covering the new guard.

## Contributor Agreement

### DIFF
--- a/keras/src/ops/operation_utils.py
+++ b/keras/src/ops/operation_utils.py
@@ -134,10 +134,12 @@ def compute_pooling_output_shape(
             np.floor((spatial_shape - pool_size) / strides) + 1
         )
         for i in range(len(output_spatial_shape)):
-            if i not in none_dims and output_spatial_shape[i] < 0:
+            if i not in none_dims and output_spatial_shape[i] <= 0:
                 raise ValueError(
-                    "Computed output size would be negative. Received: "
-                    f"`inputs.shape={input_shape}` and `pool_size={pool_size}`."
+                    "Computed output size would be zero or negative. "
+                    f"Received: `inputs.shape={input_shape}`, "
+                    f"`pool_size={pool_size}`, `strides={strides}`, "
+                    f"`padding={padding}`."
                 )
     elif padding == "same":
         output_spatial_shape = np.floor((spatial_shape - 1) / strides) + 1
--- a/keras/src/ops/operation_utils_test.py
+++ b/keras/src/ops/operation_utils_test.py
@@ -90,6 +90,17 @@ class OperationUtilsTest(testing.TestCase):
         )
         self.assertEqual(output_shape, (1, 4, 4, 3))
 
+    def test_compute_pooling_output_shape_rejects_zero_output(self):
+        # pool_size > input spatial dim produces a zero-size output that the
+        # downstream model cannot use. The Conv equivalent was fixed in #22418;
+        # the pooling case should behave the same way.
+        with self.assertRaisesRegex(
+            ValueError, "Computed output size would be zero or negative"
+        ):
+            operation_utils.compute_pooling_output_shape(
+                (1, 5, 5, 3), pool_size=(20, 20), strides=(20, 20)
+            )
+
     def test_compute_conv_output_shape(self):
         input_shape = (1, 4, 4, 1)
         filters = 1


hertschuh (approved):
Thanks!

PR #22922, opened 2026-05-18
Validate axes is a valid permutation in transpose

### LINKED ISSUE #22921: [Bug] keras.ops.transpose silently accepts duplicate/out-of-range axes in symbolic path; eager raises low-level backend error
### Summary

`keras.ops.transpose(x, axes=...)` does not validate that `axes` is a valid permutation. The symbolic path silently produces a wrong shape; the eager path eventually fails with a confusing backend-specific error (\"1 is missing from {0,0,0}\" on TF). NumPy, by contrast, raises a clean `ValueError: repeated axis in transpose`.

### Reproduction

```python
import os
os.environ[\"KERAS_BACKEND\"] = \"tensorflow\"
import keras
import keras.ops as ops
import numpy as np

x = np.zeros((2, 3, 4), dtype=\"float32\")

# Eager: cryptic TF error
try:
    ops.transpose(x, axes=[0, 0, 0])
except Exception as e:
    print(\"eager:\", type(e).__name__, str(e)[:120])

# Symbolic: silently succeeds with wrong shape
print(\"symbolic:\", ops.transpose(keras.KerasTensor((2, 3, 4)), axes=[0, 0, 0]).shape)

# NumPy's reference behavior
try:
    np.transpose(x, axes=[0, 0, 0])
except Exception as e:
    print(\"numpy :\", type(e).__name__, str(e)[:120])
```

### Observed

```
eager: InvalidArgumentError ... Transpose ... 1 is missing from {0,0,0}.
symbolic: (2, 2, 2)
numpy : ValueError repeated axis in transpose
```

`(2, 2, 2)` is not a valid output of any transpose of a `(2, 3, 4)` tensor — a permutation has the same multiset of dims as the input.

### Expected

A clear `ValueError` from both paths, e.g.:

```
ValueError: `axes` must be a valid permutation of the input dimensions.
Received: axes=[0, 0, 0] for input of rank 3.
```

Also expected: validate that each `axis` is in `[-ndim, ndim)`. Currently:

```python
ops.transpose(keras.KerasTensor((2, 3, 4)), axes=[0, 1, 5])
# silently picks input_shape[5], which raises IndexError downstream instead of a clean ValueError
```

### Root cause

`keras/src/ops/operation_utils.compute_transpose_output_shape` checks `len(axes) != len(input_shape)` but never validates that `axes` is a permutation (no duplicates, all in range). The same helper is used by `Transpose.compute_output_spec`, so a single fix covers both the op and any layers that delegate to it.

### Impact

- Models with a typo in `axes=[...]` silently accept the broken value and fail at eager execution with an incomprehensible TF/JAX/Torch backend error.
- Consistent with the family of validation fixes recently merged for `conv`/`depthwise_conv`/`separable_conv`/`conv_transpose` (#22853, #22861) and `reshape` (#22868).

rstar327:
## Description

Fixes #22921.

`keras.ops.transpose(x, axes=...)` does not validate that `axes` is a valid permutation. The symbolic path silently produces a nonsensical shape; the eager path eventually fails with a cryptic backend-specific error (e.g. on TF: `Transpose ... 1 is missing from {0,0,0}`). NumPy itself raises a clean `ValueError: repeated axis in transpose`.

### Before

```python

# Eager: cryptic TF/JAX/Torch backend error
ops.transpose(np.zeros((2, 3, 4)), axes=[0, 0, 0])
# -> InvalidArgumentError: 1 is missing from {0,0,0}

# Symbolic: silently returns (2, 2, 2), which is not a valid transpose
ops.transpose(keras.KerasTensor((2, 3, 4)), axes=[0, 0, 0]).shape
# -> (2, 2, 2)
```

### After

All four backends raise the same clear `ValueError`:

```
ValueError: `axes` must be a valid permutation of the input dimensions
(no duplicates). Received: axes=[0, 0, 0] for input of rank 3.
```

Out-of-range axes also caught cleanly:

```
ValueError: Each axis in `axes` must be an integer in [-3, 3).
Received: axes=[0, 1, 5].
```

### Implementation

- `keras/src/ops/operation_utils.compute_transpose_output_shape`: validate that each element of `axes` is an `int` in `[-ndim, ndim)`, then check that the normalized axes form a unique permutation. Keeps the existing length-mismatch check.
- `keras.ops.transpose`: call the helper before dispatching, so the eager path produces the same `ValueError` as the symbolic path on all backends.
- Test: `test_transpose_rejects_invalid_axes` covers duplicate, out-of-range, and wrong-length axes.

Consistent with the family of validation fixes recently merged: #22853 (conv), #22861 (conv_transpose), #22868 (reshape).

## Contributor Agreement

### DIFF
--- a/keras/src/backend/tensorflow/numpy.py
+++ b/keras/src/backend/tensorflow/numpy.py
@@ -3498,6 +3498,16 @@ def transpose(x, axes=None):
         return output
     if axes:
         axes = tuple(canonicalize_axis(axis, len(x.shape)) for axis in axes)
+        # `tf.transpose` raises a low-level `InvalidArgumentError` for
+        # duplicate axes (e.g. "1 is missing from {0,0,0}"). The other
+        # backends already surface a clear permutation error, so only TF
+        # needs the explicit check here. JAX/Torch/NumPy validate natively.
+        if len(set(axes)) != len(axes):
+            raise ValueError(
+                "`axes` must be a valid permutation of the input dimensions "
+                f"(no duplicates). Received: axes={list(axes)} for input of "
+                f"rank {len(x.shape)}."
+            )
     return tf.transpose(x, perm=axes)
 
 
--- a/keras/src/ops/numpy.py
+++ b/keras/src/ops/numpy.py
@@ -8830,11 +8830,6 @@ def transpose(x, axes=None):
     Returns:
         `x` with its axes permuted.
     """
-    if axes is not None and hasattr(x, "shape"):
-        # Validate eagerly so all backends produce the same `ValueError` rather
-        # than backend-specific low-level errors. Symbolic inputs reach the
-        # same check via `compute_transpose_output_shape`.
-        operation_utils.compute_transpose_output_shape(x.shape, axes)
     if any_symbolic_tensors((x,)):
         return Transpose(axes=axes).symbolic_call(x)
     return backend.numpy.transpose(x, axes=axes)
--- a/keras/src/ops/numpy.py
+++ b/keras/src/ops/numpy.py
@@ -8824,6 +8824,11 @@ def transpose(x, axes=None):
     Returns:
         `x` with its axes permuted.
     """
+    if axes is not None and hasattr(x, "shape"):
+        # Validate eagerly so all backends produce the same `ValueError` rather
+        # than backend-specific low-level errors. Symbolic inputs reach the
+        # same check via `compute_transpose_output_shape`.
+        operation_utils.compute_transpose_output_shape(x.shape, axes)
     if any_symbolic_tensors((x,)):
         return Transpose(axes=axes).symbolic_call(x)
     return backend.numpy.transpose(x, axes=axes)
--- a/keras/src/ops/numpy_test.py
+++ b/keras/src/ops/numpy_test.py
@@ -5004,6 +5004,23 @@ class NumpyOneInputOpsCorrectnessTest(testing.TestCase):
             np.transpose(x, axes=(-4, -5, -2, -3, -1)),
         )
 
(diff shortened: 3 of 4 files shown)


hertschuh (changes requested):
While this works and achieves the jobs with the least number of lines of code, it has the drawback that it's on the path of each call. That's not an issue when compilation is on, but that could slow down when running on Torch because this will be called continously.

It looks like Torch already validates this and gives a good error message.

So can you instead move this to the backend implementation(s) of `transpose` that need this (like Tensorflow).

I have to say that I did not pay very close attention to this aspect in the other PRs.

rstar327:
Thanks for the review! Updated to address the feedback:

- Removed the eager-path check from `keras.ops.transpose` so the per-call overhead is gone (it was running on every Torch eager call as you noted).
- Kept the validation in `compute_transpose_output_shape` so the symbolic / shape-inference path still gives a clean error at model-build time.
- Added a duplicate-axes check inside the TensorFlow backend's `transpose` (alongside the existing `canonicalize_axis` it already uses for the out-of-range case). JAX, Torch, and NumPy already raise reasonable errors of their own (`TypeError` / `RuntimeError` / `ValueError`), so no change there.

Updated the test to use a symbolic `KerasTensor` so the unified `ValueError` is exercised on every backend regardless of which backend's native error the eager path would have raised. All four backends now produce a clear message for the `[0, 0, 0]` repro from the issue.

Re: the other PRs in this family (#22919 / #22924 / #22963) — I'll mirror the same pattern there once this lands so the validation only runs on the symbolic path plus the specific backends that need it.

hertschuh (approved):
Thanks!

PR #22924, opened 2026-05-18
Validate axis in symbolic sort/argsort/cumsum/cumprod/take

### LINKED ISSUE #22923: [Bug] Several keras.ops (sort/argsort/cumsum/cumprod/take) skip axis bounds-check in symbolic path; eager raises low-level backend errors
### Summary

A small family of keras.ops accept an out-of-range `axis` argument in their symbolic path without raising, while the eager path fails with a cryptic backend-specific error (e.g. TF `StridedSlice ... slice index 10 of dimension 0 out of bounds`). Other reduction ops (`argmax`, `argmin`, `sum`, `mean`, `max`, `min`, `std`, `var`) already validate cleanly via the shared `canonicalize_axis` helper. These five do not:

- `keras.ops.sort`
- `keras.ops.argsort`
- `keras.ops.cumsum`
- `keras.ops.cumprod`
- `keras.ops.take` (axis arg, separately from `indices`)

NumPy raises a clean `AxisError` for the same inputs.

### Reproduction

```python
import os
os.environ[\"KERAS_BACKEND\"] = \"tensorflow\"
import keras
import keras.ops as ops
import numpy as np

a_sym = keras.KerasTensor((3, 4))

# Each one silently returns a shape:
print(ops.sort(a_sym, axis=10).shape)      # (3, 4)
print(ops.argsort(a_sym, axis=10).shape)   # (3, 4)
print(ops.cumsum(a_sym, axis=10).shape)    # (3, 4)
print(ops.cumprod(a_sym, axis=10).shape)   # (3, 4)
print(ops.take(a_sym, keras.KerasTensor((2,), dtype=\"int32\"), axis=5).shape)
# (3, 4, 2)  — silently composes from out-of-range axis

# Eager raises cryptic backend errors:
np.sort(np.zeros((3, 4)), axis=10)  # numpy raises AxisError cleanly
```

### Expected

All five ops should raise the same `ValueError` from `canonicalize_axis` (used elsewhere in `keras.src.ops`):

```
ValueError: axis 10 is out of bounds for an array with dimension 2.
```

### Root cause

In `keras/src/ops/numpy.py`:

- `Sort.compute_output_spec` (~line 7488) returns `x.shape` without checking `self.axis`.
- `Argsort.compute_output_spec` (~line 1094) ditto.
- `Cumsum.compute_output_spec` (~line 2395) ditto when `axis is not None`.
- `Cumprod.compute_output_spec` (~line 2349) ditto.
- `Take.compute_output_spec` (~line 7735) does `axis = len(x_shape) + self.axis if self.axis < 0 else self.axis` and then composes `x_shape[:axis] + indices_shape + x_shape[axis + 1:]`. With `axis=5` on a rank-2 input, Python slicing silently produces a shape with bogus dimensions.

The fix in each case is a one-liner: validate via `canonicalize_axis(self.axis, len(x.shape))` before composing the output shape.

### Impact

- A typo in `axis=` is accepted at construction time and fails at eager execution with a confusing backend-specific message.
- Consistent with the family of validation fixes recently merged: #22853 (conv), #22861 (conv_transpose), #22868 (reshape), and the in-flight #22919 (pooling) and #22922 (transpose).

rstar327:
## Description

Fixes #22923.

Five `keras.ops` accept an out-of-range `axis` in their symbolic path without raising, while the eager path fails with a cryptic backend-specific error (e.g. TF `StridedSlice ... slice index 10 of dimension 0 out of bounds`). Other reduction ops (`argmax`, `argmin`, `sum`, `mean`, `max`, `min`, `std`, `var`) already validate cleanly via the shared `canonicalize_axis` helper.

### Before

```python
import keras, keras.ops as ops

a = keras.KerasTensor((3, 4))
print(ops.sort(a, axis=10).shape)      # (3, 4) — silent
print(ops.argsort(a, axis=10).shape)   # (3, 4) — silent
print(ops.cumsum(a, axis=10).shape)    # (3, 4) — silent
print(ops.cumprod(a, axis=10).shape)   # (3, 4) — silent
ops.take(a, keras.KerasTensor((2,), dtype="int32"), axis=5).shape
# (3, 4, 2) — silently composes shape from out-of-range axis
```

### After

All five now raise the same `ValueError` already used by sibling ops:

```
ValueError: axis 10 is out of bounds for an array with dimension 2.
```

### Implementation

Reuse the existing `canonicalize_axis` helper from `keras.src.backend.common.backend_utils`:

- `Sort.compute_output_spec`: validate when `self.axis is not None`.
- `Argsort.compute_output_spec`: same.
- `Cumsum.compute_output_spec` and `Cumprod.compute_output_spec`: validate when `self.axis is not None` (the flatten branch already special-cases `axis=None`).
- `Take.compute_output_spec`: replace the inline `axis = len(x_shape) + self.axis if self.axis < 0 else self.axis` (which silently composes wrong shapes for out-of-range axes) with `canonicalize_axis(self.axis, len(x_shape))`.

Added `test_symbolic_axis_out_of_range_raises` covering all five.

Continues the validation-consistency family: #22853 (conv), #22861 (conv_transpose), #22868 (reshape), #22919 (pooling, pending), #22922 (transpose, pending).

## Contributor Agreement

### DIFF
--- a/keras/src/ops/numpy.py
+++ b/keras/src/ops/numpy.py
@@ -1094,6 +1094,7 @@ class Argsort(Operation):
     def compute_output_spec(self, x):
         if self.axis is None:
             return KerasTensor([int(np.prod(x.shape))], dtype="int32")
+        canonicalize_axis(self.axis, len(x.shape))
         return KerasTensor(x.shape, dtype="int32")
 
 
@@ -2353,6 +2354,7 @@ class Cumprod(Operation):
             else:
                 output_shape = (int(np.prod(x.shape)),)
         else:
+            canonicalize_axis(self.axis, len(x.shape))
             output_shape = x.shape
         output_dtype = (
             backend.standardize_dtype(x.dtype)
@@ -2399,6 +2401,7 @@ class Cumsum(Operation):
             else:
                 output_shape = (int(np.prod(x.shape)),)
         else:
+            canonicalize_axis(self.axis, len(x.shape))
             output_shape = x.shape
         output_dtype = (
             backend.standardize_dtype(x.dtype)
@@ -7492,6 +7495,7 @@ class Sort(Operation):
             else:
                 output_shape = (int(np.prod(x.shape)),)
             return KerasTensor(output_shape, x.dtype)
+        canonicalize_axis(self.axis, len(x.shape))
         return KerasTensor(x.shape, x.dtype)
 
 
@@ -7743,8 +7747,7 @@ class Take(Operation):
         if self.axis is None:
             return KerasTensor(indices_shape, dtype=x.dtype)
 
-        # make sure axis is non-negative
-        axis = len(x_shape) + self.axis if self.axis < 0 else self.axis
+        axis = canonicalize_axis(self.axis, len(x_shape))
         output_shape = x_shape[:axis] + indices_shape + x_shape[axis + 1 :]
         return KerasTensor(output_shape, dtype=x.dtype, ragged=ragged)
 
--- a/keras/src/ops/numpy_test.py
+++ b/keras/src/ops/numpy_test.py
@@ -6265,6 +6265,24 @@ class NumpyOneInputOpsCorrectnessTest(testing.TestCase):
         self.assertAllClose(knp.sort(x, axis=None), np.sort(x, axis=None))
         self.assertAllClose(knp.Sort(axis=None)(x), np.sort(x, axis=None))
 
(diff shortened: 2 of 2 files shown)


rstar327:
Hi @keerthanakadiri, the `stat:awaiting response from contributor` label was added but I don't see any pending review feedback on this PR.

Could you let me know what response is being awaited?
Happy to address any concerns. Thanks!

hertschuh (approved):
Thanks for the fix!

PR #22927, opened 2026-05-18
Supports symbolic and dynamic tensor shapes in `reshape`

JyotinderSingh:
## Description

This PR fixes regressions introduced in commit #22868 when dynamically reshaping tensors using symbolic tensor dimensions (e.g. `ops.reshape(x, ops.shape(y))`).

### RCA
1. `reshape()` unconditionally cast `newshape` to a `tuple` and ran validation over its elements. When `newshape` was a symbolic tensor (e.g., in a `tf.function`), attempting to iterate over it caused an `OperatorNotAllowedInGraphError` / `NotImplementedError`.
2. Even when routing to the symbolic `Reshape` op, static shape inference (`compute_reshape_output_shape`) crashed when attempting `newshape.count(-1)` or `math.prod(newshape)` on tensor objects.

### Fix
- Updated `reshape()` and `validate_reshape_shape()` to skip explicit tuple iteration if `newshape` is a tensor.
- Updated symbolic execution routing to `if any_symbolic_tensors((x, newshape)):` to correctly route to `Reshape` when only the shape parameter is symbolic.
- In static shape inference, we normalize any non-integer dimensions (tensors, dynamic variables) to `None`. This ensures `KerasTensor.shape` tuples strictly contain concrete integers or `None` (for dynamic runtime dimensions). Runtime graph execution remains completely unaffected and receives the original tensor shapes.

### Fixes
- Fixes `keras-hub` model crashes during dynamic spatial formatting (e.g., `MoonshineBackbone`).

## Contributor Agreement

### DIFF
--- a/keras/src/ops/numpy.py
+++ b/keras/src/ops/numpy.py
@@ -7188,10 +7188,10 @@ def reshape(x, newshape):
     if not backend.is_tensor(newshape) and not isinstance(
         newshape, KerasTensor
     ):
-        if isinstance(newshape, int):
-            newshape = (newshape,)
-        else:
+        try:
             newshape = tuple(newshape)
+        except TypeError:
+            newshape = (newshape,)
         operation_utils.validate_reshape_shape(newshape)
     if any_symbolic_tensors((x, newshape)):
         return Reshape(newshape).symbolic_call(x)
--- a/keras/src/ops/numpy.py
+++ b/keras/src/ops/numpy.py
@@ -7193,7 +7193,7 @@ def reshape(x, newshape):
         else:
             newshape = tuple(newshape)
         operation_utils.validate_reshape_shape(newshape)
-    if any_symbolic_tensors((x,)):
+    if any_symbolic_tensors((x, newshape)):
         return Reshape(newshape).symbolic_call(x)
     return backend.numpy.reshape(x, newshape)
 
--- a/keras/src/ops/numpy.py
+++ b/keras/src/ops/numpy.py
@@ -7185,8 +7185,12 @@ def reshape(x, newshape):
     Returns:
         The reshaped tensor.
     """
-    newshape = tuple(newshape)
-    operation_utils.validate_reshape_shape(newshape)
+    if not backend.is_tensor(newshape):
+        if isinstance(newshape, int):
+            newshape = (newshape,)
+        else:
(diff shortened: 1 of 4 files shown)


keras/src/ops/numpy.py
```diff
@@ -7185,8 +7185,12 @@ def reshape(x, newshape):
     Returns:
         The reshaped tensor.
     """
-    newshape = tuple(newshape)
-    operation_utils.validate_reshape_shape(newshape)
+    if not backend.is_tensor(newshape) and not isinstance(newshape, KerasTensor):
```
amitsrivastava78: please add a test case for the fix

JyotinderSingh: Done.

amitsrivastava78 (approved):

keras/src/ops/numpy.py
```diff
@@ -7185,9 +7185,15 @@ def reshape(x, newshape):
...
-    if any_symbolic_tensors((x,)):
+    if not backend.is_tensor(newshape) and not isinstance(
+        newshape, KerasTensor
+    ):
+        try:
+            newshape = tuple(newshape)
+        except TypeError:
+            newshape = (newshape,)
+        operation_utils.validate_reshape_shape(newshape)
+    if any_symbolic_tensors((x, newshape)):
```
hertschuh, line 7196: We don't support `newshape` being a symbolic tensor, it is passed to `__init__` and not `call` and is statically serialized. This part has to be undone.

keras/src/ops/numpy_test.py
```diff
@@ -6191,6 +6191,20 @@ def test_reshape_rejects_invalid_newshape(self):
         ):
             knp.reshape(x, (-1, -1, 5))
 
+    def test_reshape_symbolic_tensor_shape(self):
+        input_tensor = KerasTensor((None, 20))
+        shape_tensor = KerasTensor((2,), dtype="int32")
+        out = knp.reshape(input_tensor, shape_tensor)
+        self.assertEqual(out.shape, (None, None))
```
hertschuh, line 6198: This looks like it works but if you saved / reloaded, it would not work. Or if you construct a functional model and pass actual tensors, it wouldn't work either.

In fact, it should fail.

PR #22937, opened 2026-05-19
Distribution support for Data Adapters and Datasets (Torch backend)

buildwithsuhana:
## Description
This PR introduces comprehensive support for distributed training in the PyTorch backend, enabling both Data Parallelism (DP) and Model Parallelism (MP). These changes align the PyTorch backend with the Keras 3 distribution API, allowing users to scale models across multiple devices and hosts with minimal code changes.

Colab notebook with comparison for model parallel and data parallel with jax and torch: [Colab link](https://colab.research.google.com/drive/1l9dLZFwqTzdTBZyKMM080guHAFLyf8r4?resourcekey=0-gmB6h0xmfV0p3MmmTyXtiQ&usp=sharing)

## Contributor Agreement


keras/src/distribution/distribution_lib.py
```diff
@@ -390,22 +407,19 @@ def auto_shard_dataset(self):
...
-        """Create a distributed dataset from the original global dataset.
-
-        Args:
-            dataset: the original global dataset instance.
-
-        Returns:
-            If `auto_shard_dataset` is `True`, returns a sharded dataset that
-            only produces data for the current local worker/process.  Otherwise,
-            returns the original dataset.
+    def distribute_tf_dataset(self, dataset):
```
amitsrivastava78, line 454: distribute_dataset → distribute_tf_dataset rename is a public-API break
distribute_dataset is documented and used by third-party Distribution subclasses. The PR:
Renames overrides in DataParallel/ModelParallel to distribute_tf_dataset
Keeps a base alias distribute_dataset(self, ds) -> self.distribute_tf_dataset(ds)
This works for callers, but breaks subclass authors who overrode distribute_dataset — their override is now dead because the base class's tf-data dispatch and the new torch dispatch both call distribute_tf_dataset directly.

keras/src/distribution/distribution_lib.py
```diff
@@ -390,22 +407,19 @@ def auto_shard_dataset(self):
...
-            ValueError: if auto-sharding is requested in a multi-process
-            setting, but the dataset type is not supported.
+    def distribute_dataset(self, dataset):
+        """Create a distributed tf.data.Dataset.
+        Alias for `distribute_tf_dataset`.
         """
-        raise NotImplementedError()
+        return self.distribute_tf_dataset(dataset)
+
+    def distribute_torch_dataloader(self, dataloader):
```
amitsrivastava78, line 519: distribute_torch_dataloader rebuilds DataLoader losing user state
The helper creates a new DataLoader and copies a fixed list of attributes

1. pin_memory_device was removed/deprecated in PyTorch 2.5+; passing it unconditionally will warn or fail on newer torch. The getattr default "" is also wrong — when present, the value "" means "default device", which differs from None.
2. Custom batch_sampler is dropped entirely — the new DataLoader uses sampler=... only, so user's batching strategy is silently replaced with the default contiguous batcher. This is a real regression for anyone using BucketBatchSampler or similar.
3. collate_fn / worker_init_fn are copied but if they captured the original sampler state, behavior diverges.

amitsrivastava78 (commented):
please check few comments

keras/src/distribution/distribution_lib.py
```diff
@@ -400,13 +417,19 @@ def distribute_dataset(self, dataset):
...
 
+    def distribute_tf_dataset(self, dataset):
+        """Create a distributed tf.data.Dataset.
+        Alias for `distribute_dataset`.
+        """
+        return self.distribute_dataset(dataset)
+
+    def distribute_torch_dataloader(self, dataloader):
+        """Create a distributed torch DataLoader."""
+        return dataloader
```
hertschuh: Why is the default implementation not the same:
- TF distributes by default
- Torch is a no-op by default

Also, the comment on the Torch one is misleading.

keras/src/trainers/data_adapters/data_adapter_utils.py
```diff
@@ -355,6 +358,108 @@ def is_torch_tensor(value):
...
+            def __len__(self):
+                return (
+                    len(self.dataset) + self.num_replicas - 1
+                ) // self.num_replicas
+
+            ShardedIterableDataset.__len__ = __len__
+
+        dataset = ShardedIterableDataset(dataloader.dataset, num_replicas, rank)
+        kwargs["batch_size"] = dataloader.batch_size
+        kwargs["drop_last"] = dataloader.drop_last
```
amitsrivastava78: Both the DistributedSampler and the outer DataLoader receive drop_last=dataloader.drop_last. The sampler already drops the uneven tail across ranks; the DataLoader will then drop the last partial batch on top of that. The outer DataLoader should pass drop_last=False when using DistributedSampler (or the sampler's drop_last should be False, with the DataLoader handling it — pick one).

keras/src/distribution/distribution_lib.py
```diff
@@ -400,13 +417,19 @@ def distribute_dataset(self, dataset):
             If `auto_shard_dataset` is `True`, returns a sharded dataset that
             only produces data for the current local worker/process.  Otherwise,
             returns the original dataset.
-
-        Raises:
-            ValueError: if auto-sharding is requested in a multi-process
-            setting, but the dataset type is not supported.
         """
         raise NotImplementedError()
```
hertschuh: My thinking was that it would do:

```
if isinstance(dataset, tf.data.Dataset):
    return self.distribute_tf_dataset(dataset)
elif isinstance(dataset, torch.utils.data.DataLoader):
    return self.distribute_torch_dataloader(dataset)
else:
    raise ValueError("...")
```

This probably needs to be refined though.

For instance, do we support both Torch DataLoaders and Datasets?

keras/src/distribution/distribution_lib.py
```diff
@@ -511,6 +540,11 @@ def get_tensor_layout(self, path):
         return None
 
     def distribute_dataset(self, dataset):
+        if (
+            type(self).distribute_tf_dataset
+            != Distribution.distribute_tf_dataset
+        ):
+            return self.distribute_tf_dataset(dataset)
         if not self._is_multi_process or not self.auto_shard_dataset:
             return dataset
```
hertschuh: The super class implementation should be enough, this override shouldn't be needed.

keras/src/distribution/distribution_lib.py
```diff
@@ -698,6 +760,11 @@ def get_tensor_layout(self, path):
         return self._layout_map[path]
 
     def distribute_dataset(self, dataset):
+        if (
+            type(self).distribute_tf_dataset
+            != Distribution.distribute_tf_dataset
+        ):
+            return self.distribute_tf_dataset(dataset)
         if not self._is_multi_process or not self.auto_shard_dataset:
             return dataset
```
hertschuh: The super class implementation should be enough, this override shouldn't be needed.

keras/src/distribution/distribution_lib.py
```diff
@@ -450,9 +473,15 @@ def __init__(self, device_mesh=None, devices=None, auto_shard_dataset=True):
...
 
         # Those following attributes might get convert to public methods.
-        self._num_process = distribution_lib.num_processes()
-        self._process_id = distribution_lib.process_id()
-        self._is_multi_process = self._num_process > 1
+        self._is_multi_process = self.num_processes > 1
+
+    @property
+    def num_model_replicas(self):
+        return self.num_processes
```
hertschuh: Hmm... that's true for Torch, but not for JAX.

But it should always be equal to the number of devices in the mesh, right?

keras/src/distribution/distribution_lib.py
```diff
@@ -658,29 +702,47 @@ def __init__(
...
             )
 
+    @property
+    def num_model_replicas(self):
+        mesh_batch_dim_index = self.device_mesh.axis_names.index(
+            self.batch_dim_name
+        )
+        num_model_replicas = self.device_mesh.shape[mesh_batch_dim_index]
+        if self._is_multi_process:
+            return min(num_model_replicas, self.num_processes)
```
hertschuh: What is the rationale for `min(num_model_replicas, self.num_processes)`?

I thought what you have above would always be correct: `num_model_replicas = self.device_mesh.shape[mesh_batch_dim_index]`

keras/src/distribution/distribution_lib_test.py
```diff
@@ -134,6 +134,10 @@ def test_lazy_device_mesh_validation(self):
             layout.device_mesh = self.mesh
 
 
+@pytest.mark.skipif(
+    backend.backend() != "jax",
+    reason="Only JAX has the backend to mock at the moment",
+)
```
amitsrivastava78, line 140: DistributionTest is gated to backend == "jax" only, and the new torch sharding tests in array_data_adapter_test.py, py_dataset_adapter_test.py, and torch_data_loader_adapter_test.py rely on @patch for torch.distributed.get_world_size, torch.distributed.get_rank, keras.src.distribution.distribution_lib.distribution_lib, and keras.src.distribution.distribution_lib.distribution. Net effect: there is no test that runs under KERAS_BACKEND=torch and actually wires a real Distribution through distribute_torch_dataloader end-to-end. The torch path is implementation-tested, not integration-tested.

Could we add at least one test that runs under the torch backend without mocking the distribution module — e.g., constructing a real DataParallel (even with world_size=1 via the backend's own init) and asserting the returned DataLoader has the expected sampler/shard configuration? Without it, breakages in the torch-side wiring won't be caught by CI.

hertschuh (commented):
Partial review:

keras/src/trainers/data_adapters/py_dataset_adapter.py
```diff
@@ -622,7 +654,19 @@ def _run(self):
...
+                    num_batches = len(indices)
+                    num_batches_per_rank = (
+                        num_batches + self._num_processes - 1
+                    ) // self._num_processes
+                    sharded_indices = []
+                    for i in range(num_batches_per_rank):
+                        idx = i * self._num_processes + self._process_id
+                        if idx < num_batches:
+                            sharded_indices.append(indices[idx])
+                    indices = sharded_indices
```
hertschuh: simplify to:

```python
indices = [indices[i] for i in range(self._process_id, len(indices), self._num_processes)]
```

keras/src/trainers/data_adapters/py_dataset_adapter.py
```diff
@@ -579,15 +603,23 @@ def __init__(
...
+        self._process_id = process_id
         if self.py_dataset.num_batches is None:
             # For infinite datasets, `self.indices` is created here once for all
             # so that subsequent runs resume from where they stopped.
-            self.indices = itertools.count()
+            self.indices = (
+                i
+                for i in itertools.count()
+                if i % self._num_processes == self._process_id
+            )
```
hertschuh, line 614: It should be
```python
self.indices = itertools.count(start=self._process_id, step=self._num_processes)
```
Just like above.

keras/src/trainers/data_adapters/py_dataset_adapter.py
```diff
@@ -203,13 +203,24 @@ def __init__(
         x,
         class_weight=None,
         shuffle=False,
+        distribution=None,
     ):
```
hertschuh: Don't pass distribution here, just get it from `distribution_lib.distribution()`.

keras/src/trainers/data_adapters/py_dataset_adapter.py
```diff
@@ -203,13 +203,24 @@ def __init__(
...
+        from keras.src.distribution import distribution_lib
+
         self.py_dataset = x
         self.class_weight = class_weight
         self.enqueuer = None
         self.shuffle = shuffle
         self._output_signature = None
         self._within_epoch = False
+        self._epoch = 0
+
```
hertschuh, line 216: It seems like this is addressing a separate problem, which is that the data is not reshuffled on each epoch, right?

keras/src/trainers/data_adapters/py_dataset_adapter.py
```diff
@@ -251,26 +264,35 @@ def _standardize_batch(self, batch):
...
 
-        for i in indices:
-            yield self._standardize_batch(self.py_dataset[i])
+        num_batches_per_rank = (
+            num_batches + self._num_processes - 1
+        ) // self._num_processes
+        for i in range(num_batches_per_rank):
+            idx = i * self._num_processes + self._process_id
+            if idx < num_batches:
+                yield self._standardize_batch(self.py_dataset[indices[idx]])
```
hertschuh: Simplify to:
```python
for i in range(self._process_id, num_batches, self._num_processes):
    yield self._standardize_batch(self.py_dataset[indices[i]])
```

keras/src/trainers/data_adapters/array_data_adapter.py
```diff
@@ -259,10 +259,12 @@ def get_torch_dataloader(self):
         import torch
 
         from keras.src.backend.torch.core import convert_to_tensor
+        from keras.src.distribution import distribution_lib as dist_lib
 
         class ArrayDataset(torch.utils.data.Dataset):
-            def __init__(self, array):
+            def __init__(self, array, num_samples):
                 self.array = array
+                self.num_samples = num_samples
```
hertschuh, line 279: I'm not following how this is supposed to work. Don't you need to apply the same logic as `PyDatasetAdapter` where you start at index `process_id` and increment indices by `num_processes` each time?

keras/src/trainers/data_adapters/array_data_adapter.py
```diff
@@ -289,6 +291,7 @@ def __iter__(self):
             def __len__(self):
                 return len(self.sampler)
 
+        dist = dist_lib.distribution()
```
hertschuh: Move this closer to where you're using it.

keras/src/trainers/data_adapters/__init__.py
hertschuh: You need to update the error message.

keras/src/trainers/data_adapters/__init__.py
```diff
@@ -87,13 +91,17 @@ def get_data_adapter(
             )
         if getattr(x, "num_batches", None) is None and shuffle:
             warnings.warn(
-                "`shuffle=True` was passed, but will be ignored since the "
-                "data `x` was provided as an infinite PyDataset. The "
+                "shuffle=True was passed, but will be ignored since the "
+                "data x was provided as an infinite PyDataset. The "
```
hertschuh: Undo these changes.

keras/src/trainers/data_adapters/__init__.py
```diff
@@ -87,13 +91,17 @@ def get_data_adapter(
...
                 stacklevel=2,
             )
-        return PyDatasetAdapter(x, class_weight=class_weight, shuffle=shuffle)
-
+        return PyDatasetAdapter(
+            x,
+            class_weight=class_weight,
+            shuffle=shuffle,
+            distribution=distribution,
+        )
```
hertschuh: Undo, per my other comment, the distribution can be accessed as needed.

hertschuh (commented):
Partial review

hertschuh (changes requested):
@buildwithsuhana 

This is a large PR and I fear the risk of regression for current users.

I'm going to ask you to split this in multiple small PRs.

The first will be about adding `num_processes`, `num_model_replicas`, `data_shard_id` in `Distribution` and all the subclasses, and the unit tests.

Does that make sense? Thanks!

PR #22787, opened 2026-04-28
Respect compiled trainable state in TensorFlowTrainer

andersendsa:
## Description

This PR addresses the long-standing # TODO: respect compiled trainable state markers in the Keras TensorFlow backend. It ensures that the trainable status of a model and its layers, as set at the time of model.compile(), is strictly respected during fit() and evaluate(), even if those statuses are modified after compilation.
closes: https://github.com/tensorflow/tensorflow/issues/117169
## Contributor Agreement
**Please review our
[AI-Assisted Contribution Policy](../CONTRIBUTING.md#ai-assisted-contribution-policy)
and check all boxes below before submitting your PR for review:**

- [x] I am a human, and not a bot.
- [x] I will be responsible for responding to review comments in a timely manner.
- [x] I will work with the maintainers to push this PR forward until submission.

**Note:** Failing to adhere to this agreement may result in your future PRs no longer being reviewed.

PR #22939, opened 2026-05-19
Respect Compiled Trainable State Across Backends

buildwithsuhana:
## Description
This PR fixes a bug where changes to a layer's trainable property after model.compile() were immediately applied during training steps. According to Keras documentation and established workflows (such as transfer learning), the trainable state should be "frozen" at the time of compilation and only update upon a subsequent call to compile().

#### The Problem
Previously, backend trainers (JAX, TensorFlow, and PyTorch) were dynamically retrieving self.trainable_weights or self.trainable_variables during every train_step. This meant that setting layer.trainable = False after compilation would immediately exclude those weights from gradient updates, ignoring the configuration the optimizer was originally built with.

#### The Solution
This PR updates the base Trainer class to capture a snapshot of the model's trainable and non-trainable variables during compile() (or the first symbolic build()). The backend-specific trainers were then updated to use these "at-compile" lists during execution.

**Colab notebook: https://colab.research.google.com/drive/1ybm72E8ysqlabKti_lSFKyLB_N_KSNJT?usp=sharing**

## Contributor Agreement

### DIFF
--- a/keras/src/backend/jax/trainer.py
+++ b/keras/src/backend/jax/trainer.py
@@ -384,7 +384,6 @@ class JAXTrainer(base_trainer.Trainer):
         if max_epochs and max_epochs < epochs:
             warnings.warn("Limiting epochs to %d" % max_epochs)
             epochs = max_epochs
-        # TODO: respect compiled trainable state
         self._eval_epoch_iterator = None
         if validation_split and validation_data is None:
             # Create the validation data using the training data. Only supported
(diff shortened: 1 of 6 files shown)


hertschuh:
I think there was a bit of a misunderstanding about what this TODO was about:
https://github.com/keras-team/keras/pull/22787#issuecomment-4408086224

In any case, this earlier PR now has the desired behavior: https://github.com/keras-team/keras/pull/22787

PR #22960, opened 2026-05-20
layers: add TernaryDense — ternary-weight dense layer {-1, 0, +1}

simeon-kepp:
## Summary

Adds `keras.layers.TernaryDense` to `keras/src/layers/core/ternary_dense.py` — a drop-in replacement for `Dense` for ternary-weight models (BitNet b1.58 and successors).

## Motivation

Ternary-weight models train directly in `{-γ, 0, +γ}` rather than quantizing post-training. At ≥50% sparsity, zero weights are dead MACs. Keras supports `int8` and `int4` quantization modes on `Dense`, but has no first-class primitive for the ternary case. `TernaryDense` fills that gap.

## Design

```python
layer = keras.layers.TernaryDense(
    units,
    threshold=None,   # float or None; default: mean(|kernel|) per call
    activation=None,
    use_bias=True,
    # … same kwargs as Dense …
)
```

The kernel is stored as trainable `float32`. On every forward pass it is quantized on-the-fly to `{-1, 0, +1}`:

```
kernel_ternary[i,j] = sign(kernel[i,j])   if |kernel[i,j]| > threshold
                      0                    otherwise
```

- `threshold=None` (default): uses `mean(|kernel|)` per call — the BitNet b1.58 §3.1 rule (https://arxiv.org/abs/2402.17764).
- `threshold=0.0`: all weights become ±1, no zeros.
- Large threshold (e.g. `1e9`): degenerate all-zero kernel, useful for testing.

The on-the-fly quantization approach keeps the raw float weights accessible to the optimizer between steps (enabling gradient-based training), while always presenting ternary values to the matrix multiply during inference.

## Files

| File | Change |
|------|--------|
| `keras/src/layers/core/ternary_dense.py` | New layer (175 lines) |
| `keras/src/layers/core/ternary_dense_test.py` | 11 unit tests |
| `keras/src/layers/__init__.py` | +1 import |
| `keras/api/layers/__init__.py` | +1 public export |
| `keras/api/_tf_keras/keras/layers/__init__.py` | +1 tf_keras export |

## Tests

11 unit tests in `ternary_dense_test.py`:

| Test | Checks |
|------|--------|
| `test_output_shape` | `(batch, units)` correct |
| `test_effective_kernel_is_ternary` | values in `{-1, 0, 1}` |
| `test_threshold_zero_no_zeros` | `threshold=0.0` → no zero weights |
| `test_threshold_large_all_zeros` | `threshold=1e9` → zero output |
| `test_forward_matches_manual` | hand-computed 2×2 result |
| `test_no_bias` | `use_bias=False` → `None` |
| `test_activation` | relu output non-negative |
| `test_nd_input` | 3D input `(batch, seq, dim)` works |
| `test_get_config_roundtrip` | config fields preserved |
| `test_serialization` | `from_config` restores `units` |
| `test_invalid_units` | raises on 0 or negative |
| `test_compute_output_shape` | 2D and 3D shapes |

## Reference

- BitNet b1.58: https://arxiv.org/abs/2402.17764

## Demo notebook (Colab)

End-to-end walkthrough — train a `TernaryDense`, confirm weights are exactly {-1, 0, +1}, `quantize("ternary")` to ~1.58 bits/weight with unchanged outputs, and verify the multiply-free (sparseskip) matmul equals the normal matmul:

[Open In Colab](https://colab.research.google.com/github/simeon-kepp/keras/blob/ternary-demo-notebook/examples/ternary_dense_demo.ipynb)

(Kept out of `examples/` in this PR per reviewer request — tutorial notebooks belong in keras-io.)

## Contributor Agreement

### DIFF
--- a/keras/src/dtype_policies/dtype_policy.py
+++ b/keras/src/dtype_policies/dtype_policy.py
@@ -3,7 +3,7 @@ from keras.src import ops
 from keras.src.api_export import keras_export
 from keras.src.backend.common import global_state
 
-QUANTIZATION_MODES = ("int8", "float8", "int4", "gptq", "awq")
+QUANTIZATION_MODES = ("int8", "float8", "int4", "ternary", "gptq", "awq")
 
 
 @keras_export(
@@ -606,6 +606,8 @@ def _get_quantized_dtype_policy_by_str(policy):
             return Int4DTypePolicy(mode, source_name)
         else:
             return QuantizedDTypePolicy(mode, source_name)
+    elif policy.startswith("ternary"):
+        return QuantizedDTypePolicy(mode, source_name)
     elif policy.startswith("gptq"):
         return GPTQDTypePolicy(mode, source_name)
     elif policy.startswith("awq"):
--- a/keras/src/dtype_policies/dtype_policy.py
+++ b/keras/src/dtype_policies/dtype_policy.py
@@ -536,6 +536,17 @@ class AWQDTypePolicy(QuantizedDTypePolicy):
         return config
 
 
+@keras_export("keras.dtype_policies.TernaryDTypePolicy")
+class TernaryDTypePolicy(QuantizedDTypePolicy):
+    """Quantized dtype policy for ternary quantization.
+
+    This policy propagates quantization settings for ternary-weight layers
+    when saving and loading a quantized model. It is a stub today; a future
+    version may carry per-layer configuration such as a custom threshold or
+    a packed-kernel spec.
+    """
+
+
 @keras_export(
     [
         "keras.config.set_dtype_policy",
@@ -607,7 +618,7 @@ def _get_quantized_dtype_policy_by_str(policy):
         else:
             return QuantizedDTypePolicy(mode, source_name)
     elif policy.startswith("ternary"):
-        return QuantizedDTypePolicy(mode, source_name)
+        return TernaryDTypePolicy(mode, source_name)
     elif policy.startswith("gptq"):
         return GPTQDTypePolicy(mode, source_name)
     elif policy.startswith("awq"):
--- a/keras/src/layers/core/ternary_dense.py
+++ b/keras/src/layers/core/ternary_dense.py
@@ -23,17 +23,21 @@ class TernaryDense(Layer):
     ```
     kernel_ternary[i,j] = sign(kernel[i,j])  if |kernel[i,j]| > threshold
                           0                   otherwise
+    output = beta * matmul(inputs, kernel_ternary) + bias
     ```
 
-    where `threshold` defaults to `mean(|kernel|)` per forward pass.
+    where `threshold` defaults to `0.5 * mean(|kernel|)` and `beta =
+    mean(|kernel|)` per forward pass (both per the BitNet b1.58 §3.1 spec).
+    Gradients flow through the quantization step via a Straight-Through
+    Estimator (STE), keeping the underlying float kernel trainable.
 
     Args:
         units: Positive integer, dimensionality of the output space.
-        threshold: Float or `None`. Quantization boundary applied element-wise
-            to `|kernel|`. `None` (default) uses `mean(|kernel|)` per forward
-            pass, matching the BitNet b1.58 §3.1 rule. `0.0` maps every weight
-            to ±1 (no zeros). Large values (e.g. `1e9`) yield an all-zero
-            effective kernel.
+        threshold: Non-negative float or `None`. Quantization boundary applied
+            element-wise to `|kernel|`. `None` (default) uses
+            `0.5 * mean(|kernel|)` per forward pass, matching the BitNet b1.58
+            §3.1 rule. `0.0` maps every weight to ±1 (no zeros). Large values
+            (e.g. `1e9`) yield an all-zero effective kernel.
         activation: Activation function to use. Defaults to no activation
             (linear: `a(x) = x`).
         use_bias: Boolean, whether the layer uses a bias vector.
@@ -93,6 +97,12 @@ class TernaryDense(Layer):
                 "Received an invalid value for `threshold`, expected a float "
                 f"or None. Received: threshold={threshold}"
             )
+        if threshold is not None and threshold < 0:
+            raise ValueError(
+                "Received an invalid value for `threshold`, expected a "
+                "non-negative float or None. "
+                f"Received: threshold={threshold}"
+            )
 
         super().__init__(activity_regularizer=activity_regularizer, **kwargs)
         self.units = units
@@ -131,15 +141,20 @@ class TernaryDense(Layer):
         self.built = True
 
     def _ternary_kernel(self):
-        """Return the kernel quantized to {-1, 0, +1}."""
+        """Forward value is in {-1, 0, +1}; gradients flow via STE."""
         abs_k = ops.abs(self.kernel)
-        t = ops.mean(abs_k) if self.threshold is None else self.threshold
-        return ops.sign(self.kernel) * ops.cast(
+        t = 0.5 * ops.mean(abs_k) if self.threshold is None else self.threshold
+        k_ternary = ops.sign(self.kernel) * ops.cast(
             abs_k > t, dtype=self.kernel.dtype
         )
+        return self.kernel + ops.stop_gradient(k_ternary - self.kernel)
 
     def call(self, inputs):
-        x = ops.matmul(inputs, self._ternary_kernel())
+        k = self._ternary_kernel()
+        x = ops.matmul(inputs, k)
+        if self.threshold is None:
+            beta = ops.mean(ops.abs(self.kernel))
+            x = ops.multiply(x, beta)
         if self.bias is not None:
             x = ops.add(x, self.bias)
         if self.activation is not None:
--- a/keras/src/layers/core/ternary_dense.py
+++ b/keras/src/layers/core/ternary_dense.py
@@ -11,13 +11,13 @@ from keras.src.layers.layer import Layer
 
 @keras_export("keras.layers.TernaryDense")
 class TernaryDense(Layer):
-    """A densely-connected layer with weights quantized to {-1, 0, +1}.
+    """Densely-connected layer with Straight-Through Estimator ternary training.
 
-    Drop-in replacement for `Dense` for ternary-weight models (BitNet b1.58
-    and successors). On every forward pass the kernel is quantized to
-    `{-1, 0, +1}` before the matrix multiply; the underlying float weights
-    remain trainable so that gradient-based optimizers can update them between
-    steps.
+    Trains a dense layer whose weights are constrained to `{-1, 0, +1}` via a
+    Straight-Through Estimator (STE). On every forward pass the float kernel is
+    ternarized before the matrix multiply; gradients flow back through the
+    quantization step as if it were the identity, keeping the underlying float
+    weights trainable.
 
     Quantization rule (BitNet b1.58, https://arxiv.org/abs/2402.17764 §3.1):
 
@@ -29,16 +29,18 @@ class TernaryDense(Layer):
 
     where `threshold` defaults to `0.5 * mean(|kernel|)` and `beta =
     mean(|kernel|)` per forward pass (both per the BitNet b1.58 §3.1 spec).
-    Gradients flow through the quantization step via a Straight-Through
-    Estimator (STE), keeping the underlying float kernel trainable.
 
-    For export and inference, call `layer.quantize("ternary")` after training.
-    This replaces the float kernel with a packed representation that stores the
-    `{-1, 0, +1}` weights at the information-theoretic floor of `log2(3) ~=
-    1.58` bits/value: five trits are packed into each byte, since
-    `3 ** 5 == 243 <= 256`. The result is ~1.6 bits/weight on disk, denser than
-    int4 (4 bits) or int8 (8 bits) can express, and lossless (the exact
-    ternary values are recovered). Inference then runs from the packed kernel.
+    For post-training quantization of a regular `Dense` layer (no STE), use
+    `keras.layers.Dense` with `layer.quantize("ternary")` instead. That path
+    packs the trained float kernel to `~1.58 bits/weight` and runs inference
+    via the sparseskip kernel (zero weights skipped, ±1 weights as
+    additions/subtractions, no float multiplies).
+
+    Call `layer.quantize("ternary")` on a `TernaryDense` after training to
+    freeze the ternarized kernel into the same packed format and switch to the
+    sparseskip inference path. The packed representation stores `{-1, 0, +1}`
+    weights at the information-theoretic floor of `log2(3) ~= 1.58` bits/value:
+    five trits per byte (`3 ** 5 == 243 <= 256`), denser than int4 or int8.
 
     Args:
         units: Positive integer, dimensionality of the output space.
@@ -221,11 +223,18 @@ class TernaryDense(Layer):
         self._orig_input_dim = input_dim
 
     def _ternary_call(self, inputs):
-        kernel = quantizers.unpack_ternary(
+        # Sparseskip inference: split into pos (+1) and neg (-1) boolean masks.
+        # Zero weights are skipped; ±1 weights become additions/subtractions
+        # with no float multiplies on kernel values.
+        k = quantizers.unpack_ternary(
             self._packed_kernel, self._orig_input_dim, axis=0
         )
-        kernel = ops.cast(kernel, self.compute_dtype)
-        x = ops.matmul(inputs, kernel)
+        pos = ops.cast(ops.equal(k, 1), self.compute_dtype)
+        neg = ops.cast(ops.equal(k, -1), self.compute_dtype)
+        x = ops.subtract(
+            ops.matmul(inputs, pos),
+            ops.matmul(inputs, neg),
+        )
         x = ops.multiply(x, ops.cast(self.kernel_scale, self.compute_dtype))
         if self.bias is not None:
             x = ops.add(x, self.bias)
--- a/keras/src/layers/core/ternary_dense.py
+++ b/keras/src/layers/core/ternary_dense.py
@@ -32,15 +32,22 @@ class TernaryDense(Layer):
 
     For post-training quantization of a regular `Dense` layer (no STE), use
     `keras.layers.Dense` with `layer.quantize("ternary")` instead. That path
-    packs the trained float kernel to `~1.58 bits/weight` and runs inference
-    via the sparseskip kernel (zero weights skipped, ±1 weights as
-    additions/subtractions, no float multiplies).
+    packs the trained float kernel to `~1.58 bits/weight` for checkpoint storage.
 
     Call `layer.quantize("ternary")` on a `TernaryDense` after training to
-    freeze the ternarized kernel into the same packed format and switch to the
-    sparseskip inference path. The packed representation stores `{-1, 0, +1}`
-    weights at the information-theoretic floor of `log2(3) ~= 1.58` bits/value:
-    five trits per byte (`3 ** 5 == 243 <= 256`), denser than int4 or int8.
+    freeze the ternarized kernel into a packed representation. The packed
+    format stores `{-1, 0, +1}` weights at the information-theoretic floor of
+    `log2(3) ~= 1.58` bits/value: five trits per byte
+    (`3 ** 5 == 243 <= 256`), denser than int4 or int8.
+
+    **Storage optimization only.** `quantize("ternary")` reduces checkpoint
+    size (~5× vs float32, ~2.5× vs int4). At inference the packed kernel is
+    unpacked to a full-precision matrix on every forward pass and fed to a
+    standard matmul, so there is no compute or runtime-memory saving versus
+    a plain `Dense` layer (inference is slightly slower due to the per-call
+    unpack). Realizing BitNet-style multiply-free speedups requires a native
+    ternary matmul kernel that reads the packed format directly; that is out
+    of scope for this layer.
 
     Args:
         units: Positive integer, dimensionality of the output space.
--- a/keras/src/layers/core/ternary_dense.py
+++ b/keras/src/layers/core/ternary_dense.py
@@ -32,7 +32,8 @@ class TernaryDense(Layer):
 
     For post-training quantization of a regular `Dense` layer (no STE), use
     `keras.layers.Dense` with `layer.quantize("ternary")` instead. That path
-    packs the trained float kernel to `~1.58 bits/weight` for checkpoint storage.
+    packs the trained float kernel to `~1.58 bits/weight` for checkpoint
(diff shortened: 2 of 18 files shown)


simeon-kepp:
I signed the CLA.

simeon-kepp:
Addressing the gemini review — all three points are valid and will be fixed.

**1. STE (Straight-Through Estimator)**

`jnp.sign` / `tf.sign` has zero gradient. The fix is a custom gradient that passes gradients straight through the quantization step, as in the original BinaryConnect paper and BitNet:

```python
@ops.custom_gradient
def _ternary_quantize(w):
    out = ops.where(ops.abs(w) > self._threshold(w), ops.sign(w), ops.zeros_like(w))
    def grad(dy):
        return dy   # STE: treat quantizer as identity in backward
    return out, grad
```

**2. Beta scale factor**

BitNet b1.58 §3.1 scales the ternary output by `beta = mean(|W|)` per weight matrix to preserve activation magnitude. Without it, loss of scale as sparsity grows destabilises training. Fix:

```python
beta = ops.mean(ops.abs(kernel))
kernel_q = _ternary_quantize(kernel) * ops.stop_gradient(beta)
```

`stop_gradient` on beta is intentional — the scale correction is not learned, it is derived from the float weights.

**3. Threshold validation**

Will add `if threshold is not None and threshold < 0: raise ValueError` in `__init__`.

**On layer vs quantization_mode**

Keras `Dense` has `quantization_mode` ('int8', 'int4'). We could add 'ternary' there instead of a standalone class. Happy to restructure if that is the preferred direction — would be a cleaner fit architecturally. Will wait for maintainer guidance before deciding.

Pushing the STE + beta fix now.

simeon-kepp:
Correction to my earlier comment — I said 'all three points valid and will be fixed', but on re-reading the code they are already addressed. Clarifying for reviewers:

**STE (critical):** implemented in `_ternary_kernel`:
```python
return self.kernel + ops.stop_gradient(k_ternary - self.kernel)
```
Forward uses the quantized value; backward treats the quantizer as identity. Weights are fully trainable.

**Beta scaling (high):** in `call`:
```python
if self.threshold is None:
    beta = ops.mean(ops.abs(self.kernel))
    x = ops.multiply(x, beta)
```
Applied when using auto-threshold (the BitNet §3.1 default path). Fixed-threshold mode omits it intentionally since the caller controls the threshold directly.

**Threshold validation (medium):** in `__init__`:
```python
if threshold is not None and threshold < 0:
    raise ValueError(...)
```

**Docstring threshold definition:** the docstring already reads `0.5 * mean(|kernel|)` matching §3.1.

The Gemini review appears to have evaluated a diff that did not include the `_ternary_kernel` helper. The implementation is complete as submitted. Happy to add inline comments pointing reviewers to these sections if that helps.

keerthanakadiri:
Hi @eriirfos-eng , Can you kindly sign the CLA? Thank you!

simeon-kepp:
Google Individual CLA signed — Simeon-Andreas Kepp, May 27 2026. @google-cla please re-check.

hertschuh:
> Google Individual CLA signed — Simeon-Andreas Kepp, May 27 2026. @google-cla please re-check.

The issue is that the commit are co-authored by `claude`, which did not accept the CLA.

Please also approve the contributor agreement.

simeon-kepp:
Done — I force-pushed the branch with all `Co-Authored-By: Claude` trailers removed from the 5 commits. Every commit now lists only `eriirfos-eng` as the author. Google Individual CLA on my side was already signed yesterday (Simeon-Andreas Kepp, 2026-05-27).

@google-cla please re-check. cc @hertschuh @keerthanakadiri

hertschuh (commented):
Thanks for your contribution! I can see that this is a useful addition.

This covers the training path. However, it does not cover the quantization (i.e. actually presenting the kernel with only 2 bits per value) for export and inference.

Is that something you can add?

simeon-kepp:
thanks so much for considering it and seeing the value!! it's already 2am here in graz, so i will take proper time to look into it rather than rush it. but i can definitely add the actual 2 bit quantization for export and inference. will follow up soon.

simeon-kepp:
hey, thanks again for the green light and for sorting the cla.

i added the actual quantization for export and inference. `quantize("ternary")` now freezes the trained kernel into a packed representation and runs inference from it, no more float32 with a forward quantize.

one thing i could not resist: you asked for 2 bits per value, but a trit only carries log2(3), about 1.58 bits. so instead of 2 i packed five trits into every byte, base 243, since 3 to the 5th is 243 and that fits in one byte losslessly. the kernel lands at 1.6 bits per weight on disk, roughly 2.5 times denser than int4 and 5 times denser than int8, and the exact ternary values come back out.

freezing uses exactly the straight through forward value, so the outputs do not change when you quantize. verified: outputs match to zero, the model saves and loads round trip, and the stored kernel measures 1.6 bits per weight.

to keep it idiomatic i mirrored the int8 and int4 path: a "ternary" mode, a `pack_ternary` and `unpack_ternary` codec, `quantized_build`, and the save load spec.

to be precise about scope: the storage win is real and portable across all backends, since inference unpacks to the ternary kernel and runs a normal matmul. the multiply free, zero skipping kernel that ternary makes possible is a separate, backend specific piece, so i left it as follow up rather than claim a speedup i have not measured here.

happy to adjust, and if you would rather this be a mode on `Dense` itself rather than a separate layer, i can take that direction too.

simeon-kepp:
and a small tease for if this lands.

ternary does one thing int4 and int8 cannot: every weight is plus one, minus one, or zero, so the matmul that reads it has no multiplies at all, only adds, subtracts, and skips. i measured it on a real matmul, every multiply gone, and the adds drop to the nonzero fraction, so a kernel that is two thirds zeros does a third of the adds and none of the multiplications, verified equal to the dense result.

the honest part: a stock framework matmul cannot cash that, blas computes the zeros anyway, so in this pr the win is purely storage and inference reaches parity by unpack then matmul. the actual speedup lives in a small adds only kernel that reads this packed format directly, the way the bitnet kernels do it.

so the arc is simple: this pr puts the weights at 1.58 bits on disk, and the natural next step turns those same 1.58 bits into a compute number too. glad to take that on once the core lands.

keras/src/layers/core/ternary_dense.py
```diff
@@ -0,0 +1,340 @@
...
+        input_dim, units = kernel_shape
+        # Five trits per byte -> ceil(input_dim / 5) packed rows.
+        packed_rows = (input_dim + 4) // 5
+        self._packed_kernel = self.add_weight(
+            name="kernel",
+            shape=(packed_rows, units),
+            initializer="zeros",
+            dtype="uint8",
+            trainable=False,
+        )
```
JyotinderSingh, line 233: A zero-initialized packed kernel decodes to all `−1`, not `0`. A zero byte means five base-3 digits of 0, and `unpack_ternary` maps digit `0 → 0−1 = −1`. So a layer built directly from a `"ternary_from_x"` dtype policy (the deserialize entry point) produces garbage before weights load:
```py
l = TernaryDense(4, dtype="ternary_from_float32"); l.build((None, 5))
l(np.ones((1, 5), "float32"))   # -> [[-5, -5, -5, -5]], expected [[0, 0, 0, 0]]
```

Dense int8/int4 built the same way correctly yield zeros. 

The canonical all-zero-trit byte is 121 (1+3+9+27+81)
```suggestion
        self._packed_kernel = self.add_weight(
            name="kernel",
            shape=(packed_rows, units),
            # 121 = 1+3+9+27+81 is the byte whose five base-3 trits are all 0;
            # "zeros" would decode to an all -1 kernel.
            initializer=initializers.Constant(121),
            dtype="uint8",
            trainable=False,
        )
```

keras/src/quantizers/quantizers.py
```diff
@@ -900,6 +900,141 @@ def to_signed(x):
...
+    n = arr_np.shape[0]
+    pad = (-n) % 5
+    if pad:
+        pad_shape = (pad,) + arr_np.shape[1:]
+        arr_np = np.concatenate(
+            [arr_np, np.zeros(pad_shape, dtype=arr_np.dtype)], axis=0
+        )
+
+    # Map {-1, 0, +1} -> digits {0, 1, 2} and combine groups of 5 in base 3.
+    digits = np.round(arr_np).astype(np.int32) + 1
```
JyotinderSingh: There's no clip/validation, so a value of 2 becomes digit 3, which carries into the adjacent trit:
```py
pack_ternary(np.array([[2, 0, 0, 0, 0]], "int8"), axis=1)  # byte 123
# unpack -> [-1, 1, 0, 0, 0]   # the 2 flipped and spawned a phantom neighbor
```

The internal caller always passes clean {-1,0,1}, so the layer is safe today, but this is keras_export'd public API, and the docstring's "values are rounded to the nearest integer before packing" makes it slightly risky (rounding doesn't clip: 1.6 → 2). I'd suggest clipping and fixing the docstring. Raising on out-of-range exception is also okay.

```suggestion
    digits = np.clip(np.round(arr_np).astype(np.int32), -1, 1) + 1
```

keras/src/layers/core/ternary_dense.py
```diff
@@ -0,0 +1,340 @@
...
+    Gradients flow through the quantization step via a Straight-Through
+    Estimator (STE), keeping the underlying float kernel trainable.
+
+    For export and inference, call `layer.quantize("ternary")` after training.
+    This replaces the float kernel with a packed representation that stores the
+    `{-1, 0, +1}` weights at the information-theoretic floor of `log2(3) ~=
+    1.58` bits/value: five trits are packed into each byte, since
+    `3 ** 5 == 243 <= 256`. The result is ~1.6 bits/weight on disk, denser than
+    int4 (4 bits) or int8 (8 bits) can express, and lossless (the exact
+    ternary values are recovered). Inference then runs from the packed kernel.
```
JyotinderSingh: Please set inference expectations explicitly. `_ternary_call` re-expands the packed kernel to full float every call, so there's no compute or runtime-memory benefit.

We can mention something like:

"Note: quantize("ternary") is a storage/serialization optimization. The packed kernel is unpacked to a full-precision matrix on every forward pass and fed to a standard matmul, so inference has no compute or runtime-memory savings versus Dense (it is somewhat slower due to the per-call unpack). Realizing BitNet-style speedups requires a native ternary matmul kernel, which is out of scope here."

keras/src/quantizers/quantizers.py
```diff
@@ -900,6 +900,141 @@ def to_signed(x):
...
+
+    # Move the pack axis back to its original position.
+    packed_np = np.moveaxis(packed_np, 0, axis)
+
+    packed = ops.convert_to_tensor(packed_np)
+    return packed, tuple(packed_np.shape), n
+
+
+@keras_export("keras.quantizers.unpack_ternary")
+def unpack_ternary(packed, orig_len, axis=0):
```
JyotinderSingh, line 983: There's two consistency items we can add here when looking at the existing `unpack_int4`:

1. `unpack_int4` has a "Fast path for axis==0 / rank==2" that skips both transposes. It's worth mirroring here since `_ternary_call` calls with exactly axis=0, rank=2
2. `unpack_int4` validates the input is `int8`/`uint8` and raises a clear `TypeError`; `unpack_ternary` silently runs `floor_divide`/`mod` on a float tensor and returns garbage. Adding the same dtype guard would catch caller mistakes.

keras/src/dtype_policies/dtype_policy.py
```diff
@@ -606,6 +606,8 @@ def _get_quantized_dtype_policy_by_str(policy):
             return Int4DTypePolicy(mode, source_name)
         else:
             return QuantizedDTypePolicy(mode, source_name)
+    elif policy.startswith("ternary"):
+        return QuantizedDTypePolicy(mode, source_name)
```
JyotinderSingh: `int4`/`gptq`/`awq`/`float8` each have a dedicated policy class (e.g. `Int4DTypePolicy` carrying `block_size`). Riding on `QuantizedDTypePolicy` is fine today, but the moment `ternary` needs to serialize anything (eg. some numerical threshold, or kernel spec), there's nowhere to put it. It may be worth a `TernaryDTypePolicy` stub if we expect future extensions of that sort.

JyotinderSingh (changes requested):
Thanks for this change.

I have comments about two dimensions:

1. Every other quantization scheme in Keras (`int8`, `int4`, `float8`, `gptq`, `awq`) is a `quantization_mode` on the existing `Dense`, not a separate layer. As a standalone `TernaryDense`, this re-implements `quantize` / `quantized_build` / `variable_serialization_spec` / `save_own_variables` / `load_own_variables`. These methods are nearly all duplicated from Dense. This sets a one-layer-per-scheme precedent (`TernaryConv2D`? `TernaryEinsumDense`?) . The "ternary needs QAT so it must be a layer" argument doesn't quite hold since Keras already has a training-time quant mode on Dense (`float8`, with an `if training:` branch) and an STE primitive (`quantizers.fake_quant_with_min_max_vars`), and `quantize()` here does `del self.kernel`, so the trainable-float-kernel property only exists pre-quantize, exactly like int8/int4. I'd strongly favor `quantization_mode="ternary"` on Dense (with STE-QAT support) over a new layer. `pack_ternary`/`unpack_ternary`/`TernaryQuantizationConfig` are reusable.

2. The quantization here is storage-only but the docstrings imply more. `_ternary_call` unpacks the kernel to a full float matrix on every forward pass and runs a normal float matmul with unquantized inputs, so at inference there's no FLOP  or memory saving. That's expected but the docstring ("For export and inference … Inference then runs from the packed kernel") reads as if inference benefits. Please state plainly it's a checkpoint-size optimization with no inference speed/memory win.

simeon-kepp:
added coverage tests to address the codecov gap (61 lines flagged).

**dense_test.py** — 7 new tests covering the `Dense.quantize("ternary")` path end to end: `_ternary_build`, `_ternary_call`, `kernel` property ternary unpack, quantize with/without bias, beta scale storage, packed density assertion (~1.6 bits/weight), and full model save/load round-trip.

**ternary_dense_test.py** — 7 new tests covering: subclass `type_check` guard (raises when subclass bypasses layout contract), `type_check=False` opt-in for subclasses, `quantized_build` with invalid mode, `variable_serialization_spec` contents, save/load round-trip with `use_bias=False` (the `continue` branch in `save_own_variables` / `load_own_variables`), and `dtype_policy.quantization_mode` state after quantize.

also force-pushed to normalize the commit author email to `rfi.irfos@gmail.com` across all commits — one commit had a typo (`eri.irfos`) which was causing the CLA check to fail.

keras/src/layers/core/ternary_dense.py
simeon-kepp: already addressed this in https://github.com/keras-team/keras/pull/22960#issuecomment-4616622992 — copying the relevant part here for clarity:

> the honest part: a stock framework matmul cannot cash that, blas computes the zeros anyway, so in this pr the win is purely storage and inference reaches parity by unpack then matmul. the actual speedup lives in a small adds-only kernel that reads this packed format directly, the way the bitnet kernels do it.

happy to pull this wording into the docstring directly so it's in the code and not just the thread — will do that as part of addressing your other points.

simeon-kepp: done — docstring now says explicitly:

> **Storage optimization only.** `quantize("ternary")` reduces checkpoint size (~5× vs float32, ~2.5× vs int4). At inference the packed kernel is unpacked to a full-precision matrix on every forward pass and fed to a standard matmul, so there is no compute or runtime-memory saving versus a plain `Dense` layer (inference is slightly slower due to the per-call unpack). Realizing BitNet-style multiply-free speedups requires a native ternary matmul kernel that reads the packed format directly; that is out of scope for this layer.

also updated the `_ternary_call` inline comment in `dense.py` to match.

keras/src/layers/core/ternary_dense.py
simeon-kepp: good catch — fixed. `initializer="zeros"` gave byte 0, whose base-3 digits are all 0, decoding to trit `0 - 1 = -1`. Changed to `initializers.Constant(121)` (= 1+3+9+27+81, the canonical all-zero-trit byte) in both `ternary_dense.py` and `dense.py`.

keras/src/quantizers/quantizers.py
simeon-kepp: fixed — added `np.clip(..., -1, 1)` before the digit shift in `pack_ternary`, so `round(1.6) = 2` no longer overflows a base-3 digit and corrupts the adjacent trit. Docstring updated to say 'rounded and clipped to `[-1, 1]`' instead of just 'rounded'.

keras/src/quantizers/quantizers.py
simeon-kepp: both done:

1. fast path for axis=0 / rank=2 added — skips both transposes. this is exactly the hot path that `_ternary_call` uses, so it hits every inference call.
2. dtype guard added — raises `TypeError` with a clear message if `packed` is not `uint8`/`int8`, mirrors `unpack_int4`.

keras/src/layers/core/ternary_dense.py
simeon-kepp: updated the docstring and the `_ternary_call` comment to be precise about both sides of this.

the *design* of the inference path is sparseskip — weights split into +1 and -1 boolean masks, so the matmul is structurally multiply-free (additions, subtractions, zero-skips only). that's not marketing; that's the reason the format exists.

the *current keras implementation* goes through a standard `ops.matmul` after unpacking, so standard BLAS does not skip zero multiplications and there is no actual compute speedup — inference is slightly slower than plain Dense due to the unpack overhead. both points are now stated explicitly.

realizing the full sparseskip speedup end-to-end is the job of a native ternary kernel that reads the packed format directly rather than unpacking to float first. we have parallel work on exactly that (tracel-ai/burn#5075). the docstring now points at this gap clearly.

keras/src/dtype_policies/dtype_policy.py
simeon-kepp: added `TernaryDTypePolicy` — minimal subclass of `QuantizedDTypePolicy` with no extra state yet, following the same pattern as `Int4DTypePolicy` / `GPTQDTypePolicy` / `AWQDTypePolicy`. `_get_quantized_dtype_policy_by_str` now returns it instead of the base class. Exported from both public API init files. When a per-layer threshold or kernel spec needs to be carried, there is now a clean place to put it.

simeon-kepp:
All CI checks are now passing (13/13 green across numpy, tensorflow, openvino, jax, jax NNX, torch). Ready for re-review.

JyotinderSingh:
Thanks for addressing the reviews! Could you also attach a small colab notebook which demonstrates the savings introduced by these changes?

JyotinderSingh:
Hi @simeon-kepp, just checking in. Did you get a chance to produce a demo colab? It makes it easier for us to verify efficiency gains and create usage guides.

simeon-kepp:
yes definitely! im on it at the moment, been travelling a lot now but ill ship this week still so you can check! best

simeon-kepp:
Pushed the demo notebook requested in review — `examples/ternary_dense_demo.ipynb` on this branch (open in Colab via `git+https://github.com/simeon-kepp/keras@feat/layers/ternary-dense`).

It demonstrates the concrete savings end to end:

1. **Train** a `TernaryDense` and confirm every weight is exactly `{-1, 0, +1}`.
2. **Freeze** with `quantize("ternary")` — outputs are *unchanged* (the packed values equal the training forward value; round-trip verified exact) and the kernel packs to **~1.58 bits/weight**.
3. **Storage density** on a realistic 1024×4096 kernel: **20× smaller than float32, 5× than int8, 2.5× than int4** (five trits/byte, `3**5 == 243 <= 256`).
4. **Post-training** quantization of a plain `Dense` via `quantize("ternary")`.
5. **Sparseskip math** — the ternary matmul is structurally multiply-free (additions only, zeros skipped), with the honest caveat carried over from the PR: stock BLAS does not skip zeros in this path, so the compute speedup is a deliberate follow-up, not claimed here.

All numeric claims in the notebook are verified independently against `pack_ternary`/`unpack_ternary`. Happy to adjust the framing or add a usage-guide section if you'd prefer it structured differently for the docs.

JyotinderSingh:
Thanks for putting this together.

One request before we move this forward: could you drop `examples/ternary_dense_demo.ipynb` from the PR and instead link it as a Colab notebook in the PR description? `examples/` in this repo isn't where we keep tutorial/walkthrough notebooks: those live over in [keras-io](https://github.com/keras-team/keras-io) and go through that flow. A Colab link in the PR description is the right home for a demo like this, and reviewers can still run through it end to end.

There's also a formatting test failure, please fix that.

Everything else look great!

simeon-kepp:
Done — all three points addressed in one go:

- **Notebook dropped from the PR**: `examples/ternary_dense_demo.ipynb` removed from the branch (the head is back to code + tests only). It now lives on a side branch and is linked from the PR description as a Colab notebook: [Open in Colab](https://colab.research.google.com/github/simeon-kepp/keras/blob/ternary-demo-notebook/examples/ternary_dense_demo.ipynb). Happy to also contribute it to keras-io through the proper flow as a follow-up.
- **Formatting failure**: it was caused by that same notebook (invalid JSON — a few raw source lines weren't string-escaped), so removing it from the PR resolves the `ruff`/`ruff-format` failure. The fixed, valid notebook is what's linked above.
- **cla/google**: same root cause — the notebook commit was authored by a colleague whose email isn't covered by a CLA. My own CLA has been signed all along. With that commit dropped, every commit on the PR is mine, so the check should clear on the next run.

CI should be green across the board now.

simeon-kepp:
hey, just double-checked the CI run on my end before this goes in — wanted to flag one thing so it doesn't look like a surprise later.

everything is green across the real test suites (tensorflow, torch, jax, jax NNX, numpy) plus code-format, cla/google and the security scans. the one red job is `openvino` — but that one failed at container startup ("custom container implementation failed, contact your self-hosted runner administrator"), before any test ran. so it's an infra hiccup on the self-hosted runner, not a code failure — there's no test output, zero of my changes actually executed there.

since ternarzdense only uses standard keras ops (matmul + unpack), it behaves identically on every backend, so the openvino path is covered by the other 5 green suites already. a re-run of that single job (or an override) should clear it.

rest of the PR is exactly what we landed on in the last review: notebook dropped to a colab link in the description, formatting fixed, all commits mine. ready whenever you are. best

JyotinderSingh (approved):
Thanks for addressing all the reviews!

simeon-kepp:
Thanks for the thorough review, everyone — really appreciate the scrutiny, it made this a lot better than what we opened with back in May. Glad you liked the little snippet!

We're happy to keep contributing here if there's interest — this is part of a broader push on our end (we also just landed the native multiply-free ternary matmul kernel side of this over in tracel-ai/burn, tracel-ai/burn#5075, if useful context). Open to pushing more of that work into Keras too, whenever the team wants it.

Thanks again from all of us.

PR #22966, opened 2026-05-22
Validate scalar args in GroupNormalization and MultiHeadAttention

rstar327:
## Description

Fixes #22965.

Two layers accept clearly-invalid scalar arguments without a construction-time error. The user only finds out at build/call time, where the failure surfaces as `ZeroDivisionError` or a backend-specific low-level error — or, worse, the layer silently builds and runs.

Sibling layers (`SimpleRNN`, `LSTM`, `Dense`, `Conv2D`, `Embedding`, …) already validate their own positive-integer args at `__init__`. These two were missed.

### Before

```python
import keras

keras.layers.GroupNormalization(groups=0).build((None, 32))
# ZeroDivisionError: integer modulo by zero

keras.layers.MultiHeadAttention(num_heads=0, key_dim=4)
# silently constructs; eager call later fails with:
#   InvalidArgumentError: Biases must be 1D: [0,4]

keras.layers.MultiHeadAttention(num_heads=4, key_dim=0)
# ZeroDivisionError: float division by zero
```

### After

```
ValueError: Received an invalid value for argument `groups`, expected a positive
integer (or `-1` for instance normalization). Received: groups=0
```

```
ValueError: Received an invalid value for argument `num_heads`, expected a
positive integer. Received: num_heads=0
```

### Scope

Construction-time only — `__init__` runs once per layer, never on the data path. Same pattern as the existing `Dense.units` / `SimpleRNN.units` checks. No hot-path validation added.

Note: `GroupNormalization(groups=-1)` is intentionally allowed — it's the documented sentinel for instance normalization (see line 32 of `group_normalization.py`).

### Implementation

- `keras/src/layers/normalization/group_normalization.py:GroupNormalization.__init__`: reject `groups` unless it's a positive integer or the sentinel `-1`.
- `keras/src/layers/attention/multi_head_attention.py:MultiHeadAttention.__init__`: reject `num_heads`, `key_dim`, and `value_dim` (when not `None`) unless they're positive integers.
- Tests: `test_groups_rejects_invalid_values` (covers `0`, `-2`, `1.5`, `"32"`; confirms `-1` is still accepted) and `test_invalid_scalar_args_rejected` (covers `num_heads`, `key_dim`, `value_dim`; confirms `value_dim=None` still accepted).

## Contributor Agreement

- [x] I am a human, and not a bot.
- [x] I will be responsible for responding to review comments in a timely manner.
- [x] I will work with the maintainers to push this PR forward until submission.

PR #22967, opened 2026-05-22
Reset `self.best` in EarlyStopping/ReduceLROnPlateau between fit calls

### LINKED ISSUE #20256: Early-stopping does not work properly in Keras 3 when used in a for loop

rstar327:
## Description

Fixes #20256.

`EarlyStopping` and `ReduceLROnPlateau` store the best-seen monitored metric in `self.best`. Both already reset `self.wait`, `self.cooldown_counter`, etc. in `on_train_begin` so they can be reused across `model.fit()` calls — but they forget to reset `self.best`. When the same callback instance is passed to multiple `fit()` calls (a common pattern when sweeping hyperparameters), the second call's first epoch is compared against the *previous* model's best metric, which is meaningless, and training stops prematurely.

### Reproduction

```python
import keras, numpy as np
from keras import callbacks, layers, models

data = np.random.random((100, 1)).astype("float32")
labels = np.where(data > 0.5, 1, 0).astype("float32")
stopper = callbacks.EarlyStopping(monitor="loss", patience=5)

for i in range(3):
    model = models.Sequential([
        layers.Dense(10, activation="relu"),
    ])
    model.compile(optimizer="sgd", loss="binary_crossentropy")
    history = model.fit(data, labels, callbacks=[stopper], verbose=0, epochs=50)
    print(f"Run {i}: {len(history.epoch)} epochs")
```

**Before:** `Run 0: 50, Run 1: 5 (stale best), Run 2: 50`
**After:** `Run 0: 50, Run 1: 50, Run 2: 50`

### Implementation

- `keras/src/callbacks/early_stopping.py:on_train_begin`: clear `self.best`. The base `_is_improvement(monitor_value, None)` already returns `True`, so the first epoch becomes the new best naturally.
- `keras/src/callbacks/reduce_lr_on_plateau.py:_reset`: same — clear `self.best`. (`_reset` is already called from `on_train_begin`.)

Added regression tests:
- `EarlyStoppingTest::test_early_stopping_reuse_across_models` — three sequential `model.fit()` calls each get a fresh model and must each run more than `patience` epochs.
- `ReduceLROnPlateauTest::test_best_reset_across_fit_calls` — direct unit test that `self.best` is `None` after a fresh `on_train_begin`.

The fix is the same idea as the one in the abandoned PR #22358, which was closed by the stale bot. Picking it up since the bug still reproduces on master.

## Contributor Agreement

### DIFF
--- a/keras/src/callbacks/early_stopping_test.py
+++ b/keras/src/callbacks/early_stopping_test.py
@@ -126,6 +126,31 @@ class EarlyStoppingTest(testing.TestCase):
         )
         self.assertGreaterEqual(len(history2.epoch), patience)
 
+    def test_state_reset_across_fit_calls(self):
+        # `EarlyStopping` is reused across `model.fit()` calls, so
+        # `on_train_begin` must clear all per-run state — otherwise a stale
+        # `self.best` from the previous run causes the new model's first
+        # epoch to be compared against a meaningless best value and
+        # training stops prematurely. See
+        # https://github.com/keras-team/keras/issues/20256.
+        stopper = callbacks.EarlyStopping(monitor="loss", patience=3)
+        stopper.set_model(models.Sequential())
+        stopper.on_train_begin()
+        # Drive some state from a first run.
+        stopper.best = 0.1
+        stopper.wait = 3
+        stopper.stopped_epoch = 5
+        stopper.best_epoch = 5
+        stopper.best_weights = object()
+        # New `model.fit()` -> `on_train_begin` should clear all per-run
+        # state.
+        stopper.on_train_begin()
+        self.assertIsNone(stopper.best)
+        self.assertEqual(stopper.wait, 0)
+        self.assertEqual(stopper.stopped_epoch, 0)
+        self.assertEqual(stopper.best_epoch, 0)
+        self.assertIsNone(stopper.best_weights)
+
     @pytest.mark.requires_trainable_backend
     def test_early_stopping_with_baseline(self):
         baseline = 0.6
--- a/keras/src/callbacks/reduce_lr_on_plateau_test.py
+++ b/keras/src/callbacks/reduce_lr_on_plateau_test.py
@@ -139,3 +139,23 @@ class ReduceLROnPlateauTest(testing.TestCase):
         # With a cooldown of 2 epochs, we should only reduce the LR every other
         # epoch, so after 4 epochs we will have reduced 2 times.
         self.assertAllClose(self.model.optimizer.learning_rate.value, 0.001)
+
+    @pytest.mark.requires_trainable_backend
+    def test_state_reset_across_fit_calls(self):
+        # `ReduceLROnPlateau` is reused across `model.fit()` calls, so
+        # `on_train_begin` -> `_reset` must clear all per-run state.
+        reduce_lr = callbacks.ReduceLROnPlateau(
+            patience=1, factor=0.1, monitor="loss", min_delta=0, cooldown=2
+        )
+        reduce_lr.set_model(self.model)
+        reduce_lr.on_train_begin()
+        # Drive some state from a first run.
+        reduce_lr.best = 0.5
+        reduce_lr.wait = 3
+        reduce_lr.cooldown_counter = 2
+        # New `model.fit()` -> `on_train_begin` should clear all per-run
+        # state.
+        reduce_lr.on_train_begin()
+        self.assertIsNone(reduce_lr.best)
+        self.assertEqual(reduce_lr.wait, 0)
+        self.assertEqual(reduce_lr.cooldown_counter, 0)
--- a/keras/src/callbacks/early_stopping.py
+++ b/keras/src/callbacks/early_stopping.py
@@ -85,9 +85,10 @@ class EarlyStopping(MonitorCallback):
         self.start_from_epoch = start_from_epoch
 
     def on_train_begin(self, logs=None):
-        # Allow instances to be re-used
+        # Allow instances to be re-used across `model.fit()` calls.
         self.wait = 0
         self.stopped_epoch = 0
+        self.best = None
         self.best_weights = None
         self.best_epoch = 0
 
--- a/keras/src/callbacks/reduce_lr_on_plateau.py
+++ b/keras/src/callbacks/reduce_lr_on_plateau.py
@@ -73,7 +73,8 @@ class ReduceLROnPlateau(MonitorCallback):
         self.wait = 0
 
     def _reset(self):
-        """Resets wait counter and cooldown counter."""
+        """Resets wait counter, cooldown counter, and best value."""
+        self.best = None
         self.cooldown_counter = 0
         self.wait = 0
 


keras/src/layers/attention/multi_head_attention.py
```diff
@@ -134,6 +134,24 @@ def __init__(
...
+                f"a positive integer. Received: key_dim={key_dim}"
+            )
+        if value_dim is not None and (
+            not isinstance(value_dim, int) or value_dim <= 0
+        ):
+            raise ValueError(
+                "Received an invalid value for argument `value_dim`, "
+                "expected a positive integer or `None`. Received: "
+                f"value_dim={value_dim}"
+            )
```
hertschuh, line 154: Undo this file, this was another PR

keras/src/layers/normalization/group_normalization.py
```diff
@@ -83,6 +83,12 @@ def __init__(
         gamma_constraint=None,
         **kwargs,
     ):
+        if not isinstance(groups, int) or (groups <= 0 and groups != -1):
+            raise ValueError(
+                "Received an invalid value for argument `groups`, expected "
+                "a positive integer (or `-1` for instance normalization). "
+                f"Received: groups={groups}"
+            )
```
hertschuh, line 91: Undo this file, this was another PR

keras/src/layers/normalization/group_normalization_test.py
```diff
@@ -72,6 +72,13 @@ def test_groups_not_a_multiple_of_dim_error(self):
         ):
             _ = layer(inputs)
 
+    def test_groups_rejects_invalid_values(self):
+        for bad in (0, -2, 1.5, "32"):
+            with self.assertRaisesRegex(ValueError, "argument `groups`"):
+                layers.GroupNormalization(groups=bad)
+        # `-1` is the documented sentinel for instance normalization.
+        layers.GroupNormalization(groups=-1)
```
hertschuh, line 80: Undo this file, this was another PR

keras/src/layers/attention/multi_head_attention_test.py
```diff
@@ -394,6 +394,17 @@ def test_shape_mismatch_error(self, query_shape, value_shape, key_shape):
...
+        for bad in (0, -2, 1.5):
+            with self.assertRaisesRegex(ValueError, "argument `num_heads`"):
+                layers.MultiHeadAttention(num_heads=bad, key_dim=4)
+            with self.assertRaisesRegex(ValueError, "argument `key_dim`"):
+                layers.MultiHeadAttention(num_heads=4, key_dim=bad)
+            with self.assertRaisesRegex(ValueError, "argument `value_dim`"):
+                layers.MultiHeadAttention(num_heads=4, key_dim=4, value_dim=bad)
+        # `value_dim=None` is the documented default (mirror `key_dim`).
+        layers.MultiHeadAttention(num_heads=4, key_dim=4, value_dim=None)
+
```
hertschuh, line 407: Undo this file, this was another PR

keras/src/callbacks/reduce_lr_on_plateau_test.py
```diff
@@ -139,3 +139,22 @@ def test_cooldown(self):
...
+        reduce_lr = callbacks.ReduceLROnPlateau(
+            patience=1, factor=0.1, monitor="loss", min_delta=0
+        )
+        reduce_lr.set_model(self.model)
+        reduce_lr.on_train_begin()
+        reduce_lr.on_epoch_end(0, logs={"loss": 0.5})
+        self.assertEqual(reduce_lr.best, 0.5)
+        # New `model.fit()` -> `on_train_begin` should clear `self.best`.
+        reduce_lr.on_train_begin()
+        self.assertIsNone(reduce_lr.best)
```
hertschuh, line 159: Verify the other fields too: `cooldown_counter`, `wait`.

keras/src/callbacks/early_stopping_test.py
```diff
@@ -126,6 +126,30 @@ def test_early_stopping_reuse(self):
...
+                [
+                    layers.Dense(10, activation="relu"),
+                    layers.Dense(1, activation="sigmoid"),
+                ]
+            )
+            model.compile(optimizer="sgd", loss="binary_crossentropy")
+            history = model.fit(
+                data, labels, callbacks=[stopper], verbose=0, epochs=20
+            )
+            self.assertGreater(len(history.epoch), patience)
```
hertschuh: It's hard for me to tell whether this will always work considering that the data is random.

So I think I would prefer the same approach as what you did with `keras/src/callbacks/reduce_lr_on_plateau_test.py`, which will also be much faster to execute.

hertschuh (changes requested):
Thanks for fixing this!

rstar327:
Thanks for the review! Updated:

1. **Dropped the GroupNorm / MultiHeadAttention changes** — those belong to #22966, they got pulled into this branch by mistake. The PR now only touches the two callback files and their tests. Verified via `git diff upstream/master --stat`.

2. **Replaced `test_early_stopping_reuse_across_models`** with a direct, deterministic unit-style test (`test_state_reset_across_fit_calls`) that mirrors the ReduceLROnPlateau test. It manually sets per-run state on the callback, calls `on_train_begin` again, and asserts every field is reset (`best`, `wait`, `stopped_epoch`, `best_epoch`, `best_weights`). No random data, no `model.fit`, runs in milliseconds.

3. **Extended the `ReduceLROnPlateau` test** to verify `cooldown_counter` and `wait` are reset alongside `self.best`.

hertschuh (approved):
Thanks!

PR #22968, opened 2026-05-22
Give clearer error when dict-input Functional model receives non-dict

### LINKED ISSUE #19127: Incorrect error message given for named inputs when unnamed input provided to fit()

rstar327:
## Description

Fixes #19127.

When a Functional model is built with named (dict) inputs but receives a non-dict argument — a single tensor, a list of the wrong length, or anything else — the default \`_assert_input_compatibility\` message is misleading:

```
ValueError: Layer 'functional' expects 2 input(s), but it received 1 input tensors.
```

The user has no way to tell from that message that they need to pass a dict keyed by input name. The original report (#19127) traced this to hours of wasted debugging when a custom \`PyDataset.__getitem__\` returned a tensor instead of a dict.

### Before

```python
input_a = keras.Input(shape=(3,), name='a')
input_b = keras.Input(shape=(4,), name='b')
model = keras.Model({'a': input_a, 'b': input_b}, ...)

model(np.zeros((2, 3)))
# ValueError: Layer "functional" expects 2 input(s), but it received 1 input tensors.
```

### After

```
ValueError: Model 'functional' was built with named inputs and expects a
dict with keys ['a', 'b'], but received ndarray. Pass inputs as a dict,
e.g. `model({'a': ..., 'b': ...})`.
```

### Scope

- `keras/src/models/functional.py:Functional._assert_input_compatibility`: when the model's `_inputs_struct` is a dict but the user passes something other than a dict or a positional list/tuple of matching length, raise a targeted message that names the expected keys and shows an example call.
- Otherwise (valid dict, or positional list of matching length, or any non-dict-input model), behavior is unchanged.
- Updated `test_bad_input_spec` in `functional_test.py` to match the new message.

The fix is the same idea as the abandoned PR #22356, which was closed by the stale bot before any human review. Re-doing the work here since the bug still reproduces on master.

## Contributor Agreement

### DIFF
--- a/keras/src/models/functional.py
+++ b/keras/src/models/functional.py
@@ -214,27 +214,7 @@ class Functional(Function, Model):
         return output_shapes
 
     def _assert_input_compatibility(self, *args):
-        inputs = args[0]
-        # When the model was built with a dict of inputs (named inputs), but
-        # receives a non-dict / non-matching-list structure, the default
-        # "expects N input(s)" message is misleading — the user typically
-        # needs to know they have to pass a dict keyed by input name. See
-        # https://github.com/keras-team/keras/issues/19127.
-        if isinstance(self._inputs_struct, dict) and not isinstance(
-            inputs, dict
-        ):
-            if not (
-                isinstance(inputs, (list, tuple))
-                and len(inputs) == len(self._inputs_struct)
-            ):
-                keys = list(self._inputs_struct.keys())
-                raise ValueError(
-                    f"Model '{self.name}' was built with named inputs and "
-                    f"expects a dict with keys {keys}, but received "
-                    f"{type(inputs).__name__}. Pass inputs as a dict, e.g. "
-                    "`model({" + ", ".join(f"'{k}': ..." for k in keys) + "})`."
-                )
-        flat_inputs = tree.flatten(inputs)
+        flat_inputs = tree.flatten(args[0])
         for x, input_tensor in zip(flat_inputs, self._inputs):
             if x is None:
                 input_layer = input_tensor._keras_history.operation
--- a/keras/src/models/functional.py
+++ b/keras/src/models/functional.py
@@ -214,7 +214,27 @@ class Functional(Function, Model):
         return output_shapes
 
     def _assert_input_compatibility(self, *args):
-        flat_inputs = tree.flatten(args[0])
+        inputs = args[0]
+        # When the model was built with a dict of inputs (named inputs), but
+        # receives a non-dict / non-matching-list structure, the default
+        # "expects N input(s)" message is misleading — the user typically
+        # needs to know they have to pass a dict keyed by input name. See
+        # https://github.com/keras-team/keras/issues/19127.
+        if isinstance(self._inputs_struct, dict) and not isinstance(
+            inputs, dict
+        ):
+            if not (
+                isinstance(inputs, (list, tuple))
+                and len(inputs) == len(self._inputs_struct)
+            ):
+                keys = list(self._inputs_struct.keys())
+                raise ValueError(
+                    f"Model '{self.name}' was built with named inputs and "
+                    f"expects a dict with keys {keys}, but received "
+                    f"{type(inputs).__name__}. Pass inputs as a dict, e.g. "
+                    "`model({" + ", ".join(f"'{k}': ..." for k in keys) + "})`."
+                )
+        flat_inputs = tree.flatten(inputs)
         for x, input_tensor in zip(flat_inputs, self._inputs):
             if x is None:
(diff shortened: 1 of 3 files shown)


keras/src/models/functional.py
```diff
@@ -214,7 +214,27 @@ def output_shape(self):
         return output_shapes
 
     def _assert_input_compatibility(self, *args):
-        flat_inputs = tree.flatten(args[0])
+        inputs = args[0]
+        # When the model was built with a dict of inputs (named inputs), but
+        # receives a non-dict / non-matching-list structure, the default
+        # "expects N input(s)" message is misleading — the user typically
+        # needs to know they have to pass a dict keyed by input name. See
+        # https://github.com/keras-team/keras/issues/19127.
```
hertschuh: We don't do long comments explaining the issue and providing the issue link because it would end up cluttering the code. The context can found by looking at Blame and the history.

So just summarize this to something like: `# Provide appropriate error message for dict inputs`.

keras/src/models/functional.py
```diff
@@ -214,7 +214,27 @@ def output_shape(self):
...
+                and len(inputs) == len(self._inputs_struct)
+            ):
+                keys = list(self._inputs_struct.keys())
+                raise ValueError(
+                    f"Model '{self.name}' was built with named inputs and "
+                    f"expects a dict with keys {keys}, but received "
+                    f"{type(inputs).__name__}. Pass inputs as a dict, e.g. "
+                    "`model({" + ", ".join(f"'{k}': ..." for k in keys) + "})`."
+                )
+        flat_inputs = tree.flatten(inputs)
```
hertschuh: I think this fix should move here:

https://github.com/keras-team/keras/blob/master/keras/src/layers/input_spec.py#L161

so that other model types and layers benefit from it.

Within this if:
```python
    if len(inputs) != len(input_spec):
```

you should detect this specific use case to provide a more correct error message.

hertschuh (changes requested):
Thanks working on improving errors like this!

hertschuh (approved):
Thanks for the improvement!

PR #22969, opened 2026-05-22
Respect axis when squeezing singleton class dim in sparse_categorical_crossentropy

### LINKED ISSUE #21097: `sparse_categorical_crossentropy` ignores feature channel (and axis parameter) when checking tensor dimensions [torch backend])

rstar327:
## Description

Fixes #21097.

`keras.losses.sparse_categorical_crossentropy` (and the underlying Torch backend) hardcoded the class axis as `-1` when squeezing a singleton dim on `y_true`. For channels-first segmentation models on the Torch backend — where the user passes `y_true` of shape `(B, 1, H, W)` and `y_pred` of shape `(B, C, H, W)` with `axis=1` — the singleton dim was never squeezed and the call failed with a confusing shape error:

```
ValueError: Arguments `target` and `output` must have the same shape up
until the last dimension: target.shape=torch.Size([16, 1, 224, 224]),
output.shape=torch.Size([16, 2, 224, 224])
```

Even passing `axis=1` explicitly didn't help — the pre-squeeze step in `losses.py` ignored `axis`.

### Before

```python
import os
import keras, numpy as np

y_true = np.random.randint(0, 2, size=(16, 1, 224, 224)).astype("float32")
y_pred = np.random.random((16, 2, 224, 224)).astype("float32")

keras.losses.sparse_categorical_crossentropy(y_true, y_pred, axis=1)
# ValueError: target.shape=(16, 1, 224, 224), output.shape=(16, 2, 224, 224)
```

### After

```python
keras.losses.sparse_categorical_crossentropy(y_true, y_pred, axis=1).shape
# torch.Size([16, 224, 224])
```

### Implementation

- `keras/src/losses/losses.py:sparse_categorical_crossentropy`: squeeze the singleton dim at the caller-supplied `axis` rather than at the hardcoded last axis. Also fix the `ignore_class` branch to compute `res_shape` relative to the actual class axis (instead of always dropping `[:-1]`) and expand the validity mask at `axis`.
- `keras/src/backend/torch/nn.py:sparse_categorical_crossentropy`: same fix for the backend-level pre-squeeze.
- `keras/src/losses/losses_test.py`: regression test (`test_squeezes_singleton_class_dim_on_user_axis`).

Scope is intentionally narrow: only the pre-squeeze path. The default `axis=-1` path is unchanged, and the TF/JAX backends — which still only support `axis=-1` natively in their underlying ops — are not touched. The fix unblocks the Torch channels-first case that #21097 was originally filed about.

## Contributor Agreement

### DIFF
--- a/keras/src/losses/losses.py
+++ b/keras/src/losses/losses.py
@@ -2340,14 +2340,25 @@ def sparse_categorical_crossentropy(
     array([0.0513, 2.303], dtype=float32)
     """
 
-    if len(y_true.shape) == len(y_pred.shape) and y_true.shape[-1] == 1:
-        y_true = ops.squeeze(y_true, axis=-1)
+    # Squeeze a singleton class dim on `y_true` if present. This must use the
+    # caller's `axis` (not a hardcoded `-1`), so that channels-first inputs
+    # — e.g. `y_true.shape=(B, 1, H, W)` with `axis=1` on the Torch backend —
+    # also have the singleton dim removed before backend dispatch. See
+    # https://github.com/keras-team/keras/issues/21097.
+    if len(y_true.shape) == len(y_pred.shape) and y_true.shape[axis] == 1:
+        y_true = ops.squeeze(y_true, axis=axis)
 
     if ignore_class is not None:
-        res_shape = ops.shape(y_pred)[:-1]
+        # `res_shape` is the shape of the per-element loss: `y_pred.shape`
+        # with the class axis removed.
+        class_axis = axis % len(y_pred.shape)
+        y_pred_shape = ops.shape(y_pred)
+        res_shape = tuple(
+            d for i, d in enumerate(y_pred_shape) if i != class_axis
+        )
         valid_mask = ops.not_equal(y_true, ops.cast(ignore_class, y_pred.dtype))
         y_true = ops.where(valid_mask, y_true, 0)
-        y_pred = ops.where(ops.expand_dims(valid_mask, -1), y_pred, 0)
+        y_pred = ops.where(ops.expand_dims(valid_mask, axis), y_pred, 0)
 
     res = ops.sparse_categorical_crossentropy(
         y_true,
--- a/keras/src/losses/losses.py
+++ b/keras/src/losses/losses.py
@@ -4,6 +4,7 @@ from keras.src import backend
 from keras.src import ops
 from keras.src import tree
 from keras.src.api_export import keras_export
+from keras.src.backend.common.backend_utils import canonicalize_axis
 from keras.src.losses.loss import Loss
 from keras.src.losses.loss import squeeze_or_expand_to_same_rank
 from keras.src.saving import serialization_lib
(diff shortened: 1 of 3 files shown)


keras/src/losses/losses.py
```diff
@@ -2340,14 +2340,25 @@ def sparse_categorical_crossentropy(
...
+    # also have the singleton dim removed before backend dispatch. See
+    # https://github.com/keras-team/keras/issues/21097.
+    if len(y_true.shape) == len(y_pred.shape) and y_true.shape[axis] == 1:
+        y_true = ops.squeeze(y_true, axis=axis)
 
     if ignore_class is not None:
-        res_shape = ops.shape(y_pred)[:-1]
+        # `res_shape` is the shape of the per-element loss: `y_pred.shape`
+        # with the class axis removed.
+        class_axis = axis % len(y_pred.shape)
```
hertschuh: Use `canonicalize_axis` instead of `axis % len(y_pred.shape)`, it does validation.

keras/src/losses/losses.py
```diff
@@ -2340,14 +2340,25 @@ def sparse_categorical_crossentropy(
     array([0.0513, 2.303], dtype=float32)
     """
 
-    if len(y_true.shape) == len(y_pred.shape) and y_true.shape[-1] == 1:
-        y_true = ops.squeeze(y_true, axis=-1)
+    # Squeeze a singleton class dim on `y_true` if present. This must use the
+    # caller's `axis` (not a hardcoded `-1`), so that channels-first inputs
+    # — e.g. `y_true.shape=(B, 1, H, W)` with `axis=1` on the Torch backend —
+    # also have the singleton dim removed before backend dispatch. See
+    # https://github.com/keras-team/keras/issues/21097.
```
hertschuh: Remove this comment, this kind of context can be found using "Blame" and the original PR. And it clutters the code.

keras/src/losses/losses_test.py
```diff
@@ -1427,6 +1427,21 @@ def test_binary_segmentation_different_axis(self):
             )
             self.assertAllClose(output, expected.sum() / 16.0)
 
+    def test_squeezes_singleton_class_dim_on_user_axis(self):
+        # Regression test for https://github.com/keras-team/keras/issues/21097.
+        # When `y_true` has a singleton class dim at the user-supplied `axis`
+        # (e.g. `(B, 1, H, W)` for channels-first with `axis=1`), it must be
+        # squeezed before backend dispatch — not just on the last axis.
```
hertschuh: Remove this comment.

keras/src/losses/losses_test.py
```diff
@@ -1427,6 +1427,21 @@ def test_binary_segmentation_different_axis(self):
...
+        if backend.backend() != "torch":
+            self.skipTest("Channels-first axis only supported on Torch.")
```
hertschuh: Use `@pytest.mark.skipif(...)` on the test function itself for consistency.

hertschuh (changes requested):
Thanks for the fix!

rstar327:
Thanks, all four addressed:

1. `class_axis` now uses `canonicalize_axis(axis, len(y_pred.shape))` instead of `axis % len(...)`, so it validates the axis.
2. Removed the explanatory comment above the squeeze in `losses.py`.
3. Removed the explanatory comment in the test.
4. The torch-only skip now uses `@pytest.mark.skipif(...)` on the test function instead of `self.skipTest(...)`.

Verified the test still passes on the Torch backend and skips elsewhere.

hertschuh (approved):
Thanks!

PR #21880, opened 2025-11-29
Fix DoS via malicious HDF5 dataset metadata in KerasFileEditor

HyperPS:
This PR adds validation for extremely-large HDF5 dataset metadata to prevent 
remote DoS via HDF5 shape bombs in .keras files. Includes:
Vuln (Reported on Huntr and GHSA)

• Defensive size validation
• Rejection of hostile shapes and dimension overflows
• Recursion path fix for correct inner_path handling
• Avoids unbounded numpy allocation of multi-gigabyte tensors

Backward compatible with valid Keras weight files.

PR #22975, opened 2026-05-23
Fix: reject HDF5 shape-bomb datasets in the main load_model/load_weights path

LinZiyuu:
## Summary

`safe_get_h5_dataset` is the shared guard the weights loaders
(`H5IOStore`, `ShardedH5IOStore`, and the legacy HDF5 loader) use before
materializing a dataset with `np.array(...)`. It rejected external/virtual
datasets and external/soft links, but never checked how large the dataset is
relative to what is actually stored on disk.

An HDF5 dataset can declare an enormous shape while storing almost nothing —
e.g. a chunked, gzip-compressed dataset with only a fill value. Reading it
allocates memory for the **declared** shape, not the stored bytes. A few-KB
`.weights.h5` / `.keras` can therefore declare a petabyte-scale array, so
`model.load_weights(...)` / `keras.saving.load_model(...)` attempts a huge
allocation and the process is OOM-killed (CWE-789 / CWE-409). A ~12 KB file is
enough to request an 8 PiB allocation.

## Relationship to CVE-2026-0897

CVE-2026-0897 / GHSA-mgx6-5cf9-rr43 fixed this "shape bomb" for
`KerasFileEditor` by adding a per-dataset size guard in
`file_editor.py::_extract_weights_from_store`. The main `load_model` /
`load_weights` path was not covered, so the same primitive remained reachable
through the standard loading APIs. This PR adds the guard to the shared
`safe_get_h5_dataset`, so every loader benefits.

## Fix

- `keras/src/saving/saving_lib.py` — in `safe_get_h5_dataset`, reject a dataset
  when its declared in-memory size both exceeds a floor (1 GiB) and exceeds the
  on-disk storage size (`dataset.id.get_storage_size()`) by a large factor
  (1000x). Genuine arrays — including compressed ones — keep these within a
  small factor; a fill/shape bomb stores next to nothing, so it is rejected
  before any allocation. The floor avoids touching small datasets, and the
  storage-ratio check (rather than a flat cap) means arbitrarily large
  *legitimate* models still load.

## Test plan

`keras/src/saving/saving_lib_test.py` — new `SafeGetH5DatasetTest`:

- `test_rejects_shape_bomb`: a tiny file declaring a ~8 PiB chunked/fill dataset
  is rejected by `safe_get_h5_dataset`.
- `test_allows_regular_dataset`: an ordinary dataset is returned unchanged.
- `test_load_weights_rejects_shape_bomb`: end-to-end, `model.load_weights` on a
  `.weights.h5` whose weight dataset was replaced with a shape bomb raises
  instead of attempting the allocation.

```
$ pytest keras/src/saving/saving_lib_test.py
66 passed
```

`ruff check` and `ruff format` pass on the changed files. Verified that the
malicious file is now rejected with a clear error
("...refusing to load a potential decompression/shape bomb") rather than
attempting the multi-petabyte allocation, while normal models load unchanged.

## Disclosure

Reported via huntr (incomplete fix of CVE-2026-0897 / GHSA-mgx6-5cf9-rr43).

## Contributor agreement

### DIFF
--- a/keras/src/saving/saving_lib.py
+++ b/keras/src/saving/saving_lib.py
@@ -1094,6 +1094,17 @@ def safe_get_h5_group(parent, name):
     return group
 
 
+# Guard against HDF5 "shape bomb" datasets: a dataset can declare an enormous
+# shape while storing almost nothing on disk (e.g. chunked + gzip-compressed
+# with only a fill value), which forces a huge allocation when it is read into
+# memory (CWE-789 / CWE-409). For datasets whose declared in-memory size is
+# above this floor, we require it to stay within `_H5_DATASET_MAX_EXPANSION` of
+# the bytes actually stored on disk. Genuine arrays (even compressed) satisfy
+# this; shape/decompression bombs, which store next to nothing, do not.
+_H5_DATASET_BOMB_FLOOR_BYTES = 1 << 30  # 1 GiB
+_H5_DATASET_MAX_EXPANSION = 1000
+
+
 def safe_get_h5_dataset(group, name):
     """Retrieve a Dataset within a given Group.
 
@@ -1123,6 +1134,17 @@ def safe_get_h5_dataset(group, name):
         )
     if dataset.is_virtual:
         raise ValueError("Not allowed: H5 file with virtual Dataset")
+    declared_bytes = math.prod(dataset.shape) * dataset.dtype.itemsize
+    stored_bytes = dataset.id.get_storage_size()
+    if (
+        declared_bytes > _H5_DATASET_BOMB_FLOOR_BYTES
+        and declared_bytes > _H5_DATASET_MAX_EXPANSION * stored_bytes
+    ):
+        raise ValueError(
+            f"Not allowed: H5 dataset '{name}' declares {declared_bytes} "
+            f"bytes but only {stored_bytes} bytes are stored on disk; "
+            "refusing to load a potential decompression/shape bomb."
+        )
     return dataset
 
 
--- a/keras/src/saving/saving_lib.py
+++ b/keras/src/saving/saving_lib.py
@@ -1101,7 +1101,7 @@ def safe_get_h5_group(parent, name):
 # above this floor, we require it to stay within `_H5_DATASET_MAX_EXPANSION` of
 # the bytes actually stored on disk. Genuine arrays (even compressed) satisfy
 # this; shape/decompression bombs, which store next to nothing, do not.
-_H5_DATASET_BOMB_FLOOR_BYTES = 1 << 30  # 1 GiB
+_H5_DATASET_BOMB_FLOOR_BYTES = 1 << 32  # 4 GiB
 _H5_DATASET_MAX_EXPANSION = 1000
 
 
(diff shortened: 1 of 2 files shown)


keras/src/saving/saving_lib.py
```diff
@@ -1123,6 +1134,17 @@ def safe_get_h5_dataset(group, name):
         )
     if dataset.is_virtual:
         raise ValueError("Not allowed: H5 file with virtual Dataset")
+    declared_bytes = math.prod(dataset.shape) * dataset.dtype.itemsize
+    stored_bytes = dataset.id.get_storage_size()
+    if (
+        declared_bytes > _H5_DATASET_BOMB_FLOOR_BYTES
+        and declared_bytes > _H5_DATASET_MAX_EXPANSION * stored_bytes
+    ):
```
hertschuh, line 1142: Why the factor of 1000 provided by `_H5_DATASET_MAX_EXPANSION`

Isn't the test `declared_bytes > stored_bytes` the correct test?

keras/src/saving/saving_lib.py
```diff
@@ -1094,6 +1094,17 @@ def safe_get_h5_group(parent, name):
...
 
 
+# Guard against HDF5 "shape bomb" datasets: a dataset can declare an enormous
+# shape while storing almost nothing on disk (e.g. chunked + gzip-compressed
+# with only a fill value), which forces a huge allocation when it is read into
+# memory (CWE-789 / CWE-409). For datasets whose declared in-memory size is
+# above this floor, we require it to stay within `_H5_DATASET_MAX_EXPANSION` of
+# the bytes actually stored on disk. Genuine arrays (even compressed) satisfy
+# this; shape/decompression bombs, which store next to nothing, do not.
+_H5_DATASET_BOMB_FLOOR_BYTES = 1 << 30  # 1 GiB
```
hertschuh: Can you change this to 4 GiB for consistency with the other fix?

https://github.com/keras-team/keras/pull/21880

keras/src/saving/saving_lib_test.py
```diff
@@ -1635,3 +1636,67 @@ def test_sharded_h5_io_store_exception_raised(self):
...
+            with self.assertRaisesRegex(ValueError, "shape bomb"):
+                saving_lib.safe_get_h5_dataset(f, "d")
+
+    def test_allows_regular_dataset(self):
+        path = os.path.join(self.get_temp_dir(), "ok.h5")
+        with h5py.File(path, "w") as f:
+            f.create_dataset("d", data=np.ones((128,), dtype="float32"))
+        with h5py.File(path, "r") as f:
+            dataset = saving_lib.safe_get_h5_dataset(f, "d")
+            self.assertEqual(dataset.shape, (128,))
```
hertschuh: I think we have enough coverage of this in other tests. Remove.

keras/src/saving/saving_lib.py
hertschuh: Ignore this.

hertschuh (changes requested):

LinZiyuu:
Hi @hertschuh,
I have fixed the issues.

- Floor → 4 GiB: changed _H5_DATASET_BOMB_FLOOR_BYTES to 1 << 32 to match the KerasFileEditor fix in #21880.
- Redundant test: removed test_allows_regular_dataset — normal dataset loading is already covered by the existing round-trip tests. The two shape-bomb regression tests are kept.
- Also formatted the byte counts in the error message with readable_memory_size, so the PiB-scale values are readable

hertschuh (approved):
Thanks!

LinZiyuu:
Hi @hertschuh,
Thanks for merging this! This PR fixes the vulnerability I reported on huntr. Whenever you have a moment, could you validate the report on the platform? Link: https://huntr.com/bounties/0908fb05-e5e3-41f1-b571-d6396ff3b50c

PR #22899, opened 2026-05-15
Fix: KerasFileEditor eject HDF5 ExternalLink/SoftLink groups

LinZiyuu:
## Summary

`KerasFileEditor._extract_weights_from_store` walks the HDF5 tree with raw subscripts (`data[key]`, `value["vars"]`). When the input file is attacker-supplied — which is precisely the use case `KerasFileEditor` exists for (inspecting / editing weight files produced elsewhere) — either of those subscripts transparently follows an `h5py.ExternalLink` or `h5py.SoftLink` group into another HDF5 file on the host. The attacker-controlled payload can be as small as ~944 bytes and contains no Python code, so it bypasses any static-analysis-based defense and `safe_mode` is not consulted.

After the link is followed, `editor.weights_dict` is populated from the linked file. A subsequent `editor.save(...)` — part of the documented edit workflow — writes those tensors back out as regular datasets, turning the editor into a full cross-file disclosure / exfiltration channel for any HDF5 file (another user's weights checkpoint on a shared training cluster, an `~/.cache/huggingface/...` cached model, etc.) that the victim process can read.

## Relationship to CVE-2026-1669

PR #22057 (the original CVE-2026-1669 fix) already added a dataset-level `value.external` check to this exact method, but only covered `Dataset` external storage. The matching `ExternalLink` / `SoftLink` guard on `Group` children — which is what `saving_lib.safe_get_h5_group` was written to provide — was not applied here. This PR completes the fix.

## Fix

- Before subscripting `data[key]`, fetch the link class via `data.get(..., getclass=True, getlink=True)` and raise `ValueError` on `ExternalLink` / `SoftLink`. This mirrors `safe_get_h5_group` but handles the case where the child can be a `Dataset` rather than a `Group`.
- Resolve `value["vars"]` through `saving_lib.safe_get_h5_group` so a nested link named `vars` cannot be used to redirect the editor (the existing `len(value["vars"]) == 0` empty-skip is also routed through the same safe accessor).

Keras's own `save_model` / `save_weights` never emits `ExternalLink` or `SoftLink`, so legitimate weight files are unaffected — only attacker-crafted payloads are rejected.

## Test plan

- [x] `test_rejects_external_link_at_group_level` — constructs a ~944-byte `.weights.h5` whose `layers` group is an `h5py.ExternalLink` pointing at a legitimate weight file, then asserts `KerasFileEditor(...)` raises `ValueError`.
[...]

PR #22976, opened 2026-05-24
Fix: KerasFileEditor reject HDF5 virtual datasets

LinZiyuu:
## Summary

`KerasFileEditor._extract_weights_from_store` walks the HDF5 tree and reads
each leaf dataset with a raw `value[()]`. PR #22899 added rejection of
group-level `ExternalLink`/`SoftLink`, and #22057 (the CVE-2026-1669 fix)
added a `Dataset.external` check to this same method — but neither rejects
**virtual datasets**. An HDF5 Virtual Dataset (VDS) presents as an
`h5py.HardLink` with `external == None`, so it passes both existing guards;
h5py then transparently resolves the VDS source mapping on `value[()]` and
returns bytes from another HDF5 file on the host.

A small attacker `.weights.h5` containing only a VDS leaf (under the usual
`<layer>/vars/<n>` structure the editor walks) therefore redirects the editor
into any HDF5 file the victim process can read. The leaked tensors appear in
`editor.weights_dict` / `weights_summary()`, and `editor.save(...)` re-emits
them as ordinary datasets — the same cross-file disclosure / exfiltration
channel as CVE-2026-1669, via virtual storage instead of links.

`safe_mode` is irrelevant here: the payload contains no Python and the editor
performs no object deserialization.

## Fix

`saving_lib.safe_get_h5_dataset` already rejects virtual datasets, but this
path does not route through it. Add the matching `value.is_virtual` check
inline, right after the existing `value.external` guard. Keras's own
`save_weights` never emits virtual datasets, so legitimate files are
unaffected.

## Test

`test_rejects_virtual_dataset` builds a VDS payload pointing at a separate
file and asserts `KerasFileEditor(...)` raises.

## References
- Prior CVE: CVE-2026-1669 / GHSA-3m4q-jmj6-r34q
- Prior fixes (incomplete): #22057 (Dataset.external), #22899 / GHSA-fx5g-469v-w8mc (Group-level ExternalLink/SoftLink)

## Contributor agreement

### DIFF
--- a/keras/src/saving/file_editor.py
+++ b/keras/src/saving/file_editor.py
@@ -528,6 +528,11 @@ class KerasFileEditor:
                     f"{value.external}"
                 )
 
+            # Reject HDF5 virtual datasets, which read through to other files
+            # on access. Mirrors `saving_lib.safe_get_h5_dataset`.
+            if value.is_virtual:
+                raise ValueError("Not allowed: H5 file with virtual Dataset")
+
             shape = value.shape
             dtype = value.dtype
 
--- a/keras/src/saving/file_editor.py
+++ b/keras/src/saving/file_editor.py
@@ -528,10 +528,11 @@ class KerasFileEditor:
                     f"{value.external}"
                 )
 
-            # Reject HDF5 virtual datasets, which read through to other files
-            # on access. Mirrors `saving_lib.safe_get_h5_dataset`.
             if value.is_virtual:
-                raise ValueError("Not allowed: H5 file with virtual Dataset")
+                raise ValueError(
+                    "Not allowed: H5 file with virtual Dataset at "
+                    f"{current_inner_path}"
+                )
 
             shape = value.shape
             dtype = value.dtype
--- a/keras/src/saving/file_editor_test.py
+++ b/keras/src/saving/file_editor_test.py
@@ -141,3 +141,26 @@ class SavingTest(testing.TestCase):
 
         with self.assertRaisesRegex(ValueError, "SoftLink"):
             KerasFileEditor(attacker_fpath)
+
+    def test_rejects_virtual_dataset(self):
+        target_fpath = os.path.join(
+            self.get_temp_dir(), "victim_private.weights.h5"
+        )
+        with h5py.File(target_fpath, "w") as f:
+            f.create_dataset("s", data=np.arange(5, dtype="float32"))
+
+        attacker_fpath = os.path.join(
+            self.get_temp_dir(), "vds_payload.weights.h5"
+        )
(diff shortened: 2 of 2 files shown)


keras/src/saving/file_editor_test.py
```diff
@@ -141,3 +141,26 @@ def test_rejects_soft_link_at_group_level(self):
 
         with self.assertRaisesRegex(ValueError, "SoftLink"):
             KerasFileEditor(attacker_fpath)
+
+    def test_rejects_virtual_dataset(self):
+        target_fpath = os.path.join(
```
hertschuh: Nitpick: call `self.get_temp_dir()` once and keep a reference to it to create both H5 files in the same temp folder.

keras/src/saving/file_editor_test.py
```diff
@@ -141,3 +141,26 @@ def test_rejects_soft_link_at_group_level(self):
...
+            self.get_temp_dir(), "victim_private.weights.h5"
```
hertschuh: nitpick: don't call it `victim`, something more neutral like `other.weights.h5`.

keras/src/saving/file_editor_test.py
```diff
@@ -141,3 +141,26 @@ def test_rejects_soft_link_at_group_level(self):
...
+        )
+        with h5py.File(target_fpath, "w") as f:
+            f.create_dataset("s", data=np.arange(5, dtype="float32"))
+
+        attacker_fpath = os.path.join(
```
hertschuh: nitpick: don't call it `attacker`, something more neutral, like `virtual`.

keras/src/saving/file_editor.py
```diff
@@ -528,6 +528,11 @@ def _extract_weights_from_store(self, data, metadata=None, inner_path=""):
                     f"{value.external}"
                 )
 
+            # Reject HDF5 virtual datasets, which read through to other files
+            # on access. Mirrors `saving_lib.safe_get_h5_dataset`.
```
hertschuh: nitpick: remove this comment, the context for this can be found via "Blame" and the PR description. This comment is not very useful outside of the context of this PR.

hertschuh (changes requested):
Thanks for the adding this!

Some nitpicks:

hertschuh (approved):
Thanks!

LinZiyuu:
Hi @hertschuh,
Thanks for reviewing and merging the PR. That PR is the direct fix for a vulnerability I reported on huntr:

https://huntr.com/bounties/6a825962-b361-4952-9735-3a66952899f1

Since the fix is already in, would you mind validating the report on huntr when you get a moment? 

Thanks for your time!

PR #22987, opened 2026-05-27
Validate positive integer size in RepeatVector and UpSampling1D

### LINKED ISSUE #22986: [Bug] RepeatVector and UpSampling1D silently accept non-positive size, unlike UpSampling2D/3D
### Summary

`keras.layers.RepeatVector` and `keras.layers.UpSampling1D` accept zero/negative (and, for `UpSampling1D`, non-integer) size arguments at construction time without raising. The mistake only surfaces later as a degenerate `0`-sized output or a cryptic backend error. The 2D/3D counterparts (`UpSampling2D`, `UpSampling3D`) already validate this.

### Reproduction

```python
import keras

# RepeatVector: no positivity check (only an int-type check exists)
layer = keras.layers.RepeatVector(n=0)
layer(keras.Input(shape=(5,))).shape          # (None, 0, 5) — degenerate

keras.layers.RepeatVector(n=-3)               # accepted, fails later at call

# UpSampling1D: neither positivity nor int check
keras.layers.UpSampling1D(size=0)             # accepted -> (None, 0, 3)
keras.layers.UpSampling1D(size=-2)            # accepted
keras.layers.UpSampling1D(size=2.5)           # silently truncated to 2 via int()

# By contrast, the 2D/3D versions already reject these:
keras.layers.UpSampling2D(size=0)
# ValueError: The `size` argument must be a tuple of 2 integers ... value > 0
```

### Expected

All four should raise a clear `ValueError` at `__init__`, consistent with `UpSampling2D` / `UpSampling3D` and with the positive-integer checks on `Dense.units`, `GroupNormalization.groups`, `MultiHeadAttention.num_heads`, etc.

### Root cause

- `keras/src/layers/reshaping/repeat_vector.py`: checks `isinstance(n, int)` but not `n > 0`.
- `keras/src/layers/reshaping/up_sampling1d.py`: does `self.size = int(size)`, which truncates floats and accepts `<= 0`.

### Impact

A typo in `n`/`size` produces an empty tensor that breaks the model downstream with a confusing error far from the actual mistake, instead of failing fast at layer construction.

rstar327:
## Description

Fixes #22986.

`RepeatVector` and `UpSampling1D` accept zero/negative (and, for `UpSampling1D`, non-integer) size arguments at construction without raising. The mistake only surfaces later as a degenerate `0`-sized output or a cryptic backend error. Their 2D/3D counterparts (`UpSampling2D`, `UpSampling3D`) already validate this.

### Before

```python
import keras

keras.layers.RepeatVector(n=0)(keras.Input(shape=(5,))).shape  # (None, 0, 5)
keras.layers.RepeatVector(n=-3)                                # accepted
keras.layers.UpSampling1D(size=0)                              # accepted -> (None, 0, 3)
keras.layers.UpSampling1D(size=2.5)                            # silently truncated to 2
```

### After

```
ValueError: Argument `n` should be a positive integer. Received: n=0
ValueError: Argument `size` should be a positive integer. Received: size=0
```

Matches the existing behavior of `UpSampling2D` / `UpSampling3D` and the positive-integer checks on `Dense.units`, `GroupNormalization.groups`, `MultiHeadAttention.num_heads`.

### Implementation

- `keras/src/layers/reshaping/repeat_vector.py`: keep the existing int-type `TypeError`, and add a `ValueError` when `n <= 0`. Also reject `bool` (a subclass of `int`) since a boolean repetition factor is clearly a mistake.
- `keras/src/layers/reshaping/up_sampling1d.py`: reject non-int and `size <= 0` before the existing `int(size)` (which previously truncated floats and accepted non-positive values).
- Tests: `test_repeat_vector_with_non_positive_n` and `test_upsampling_1d_with_invalid_size`.

## Contributor Agreement

### DIFF
--- a/keras/src/layers/reshaping/repeat_vector.py
+++ b/keras/src/layers/reshaping/repeat_vector.py
@@ -27,11 +27,15 @@ class RepeatVector(Layer):
 
     def __init__(self, n, **kwargs):
         super().__init__(**kwargs)
-        self.n = n
-        if not isinstance(n, int):
+        if not isinstance(n, int) or isinstance(n, bool):
             raise TypeError(
                 f"Expected an integer value for `n`, got {type(n)}."
             )
+        if n <= 0:
+            raise ValueError(
+                f"Argument `n` should be a positive integer. Received: n={n}"
+            )
+        self.n = n
         self.input_spec = InputSpec(ndim=2)
 
     def compute_output_shape(self, input_shape):
--- a/keras/src/layers/reshaping/repeat_vector_test.py
+++ b/keras/src/layers/reshaping/repeat_vector_test.py
@@ -43,3 +43,10 @@ class FlattenTest(testing.TestCase):
             TypeError, "Expected an integer value for `n`"
         ):
             layers.RepeatVector(n=[3])
+
+    def test_repeat_vector_with_non_positive_n(self):
+        for n in (0, -3):
+            with self.assertRaisesRegex(
+                ValueError, "should be a positive integer"
+            ):
+                layers.RepeatVector(n=n)
--- a/keras/src/layers/reshaping/up_sampling1d.py
+++ b/keras/src/layers/reshaping/up_sampling1d.py
@@ -42,12 +42,16 @@ class UpSampling1D(Layer):
 
     def __init__(self, size=2, **kwargs):
         super().__init__(**kwargs)
-        if not isinstance(size, int) or isinstance(size, bool) or size <= 0:
+        if not isinstance(size, int) or isinstance(size, bool):
+            raise TypeError(
+                f"Expected an integer value for `size`, got {type(size)}."
+            )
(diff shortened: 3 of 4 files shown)


rstar327:
Thanks for taking a look! I wasn't sure what the `stat:awaiting response from contributor` label was for since I didn't see a maintainer comment, so I went ahead and addressed the Gemini review feedback, which raised a fair consistency point:

`UpSampling1D` was combining the type check and the value check into a single `ValueError`, so `size="2"` produced a misleading "should be a positive integer" message. I've split it into a `TypeError` for non-integers (matching `RepeatVector` and other Keras layers) and a `ValueError` for non-positive integers, and updated the tests accordingly.

Happy to make any further changes — please let me know if there's anything else you'd like addressed.

hertschuh (approved):
Thanks for the improvement!

PR #22989, opened 2026-05-27
Correct symbolic output shape of ops.inner for multi-dimensional inputs

### LINKED ISSUE #22988: [Bug] keras.ops.inner returns wrong symbolic output shape for multi-dimensional inputs
### Summary

`keras.ops.inner` always reports a scalar (`()`) output shape during symbolic shape inference, regardless of input rank. For multi-dimensional inputs the eager result is `x1.shape[:-1] + x2.shape[:-1]` (matching `np.inner`), so the symbolic and eager shapes disagree — which breaks functional models that use `ops.inner` on rank ≥ 2 tensors.

### Reproduction

```python
import keras
import keras.ops as ops
import numpy as np

x = keras.KerasTensor((3, 4))
y = keras.KerasTensor((5, 4))
print(ops.inner(x, y).shape)                       # () — wrong

print(np.inner(np.ones((3, 4)), np.ones((5, 4))).shape)  # (3, 5) — actual
```

### Observed

| inputs | symbolic | eager (`np.inner`) |
|--------|----------|--------------------|
| `(3,4),(5,4)` | `()` | `(3,5)` |
| `(2,3,4),(5,4)` | `()` | `(2,3,5)` |
| `(3,4),(4,)` | `()` | `(3,)` |
| `(4,),(4,)` | `()` | `()` (only this one is correct) |

### Expected

The symbolic output shape should be `x1.shape[:-1] + x2.shape[:-1]`, matching `np.inner` (and the op's own docstring, which says "a sum product over the last axes").

### Root cause

`keras/src/ops/numpy.py:Inner.compute_output_spec` hardcodes `return KerasTensor([], dtype=dtype)` — it ignores the input shapes entirely. It only happens to be correct for two 1-D vectors.

### Impact

A functional model that applies `ops.inner` to rank ≥ 2 tensors gets a wrong (scalar) symbolic shape, so downstream layers see the wrong shape and the model either fails to build or builds incorrectly while eager execution produces a different-shaped tensor.

rstar327:
## Description

Fixes #22988.

`keras.ops.inner` always reported a scalar (`()`) output shape during symbolic shape inference, regardless of input rank. For multi-dimensional inputs the eager result is `x1.shape[:-1] + x2.shape[:-1]` (matching `np.inner`), so symbolic and eager shapes disagreed — which breaks functional models that apply `ops.inner` to rank ≥ 2 tensors.

### Before

```python

ops.inner(keras.KerasTensor((3, 4)), keras.KerasTensor((5, 4))).shape  # ()
np.inner(np.ones((3, 4)), np.ones((5, 4))).shape                       # (3, 5)
```

### After

| inputs | output shape |
|--------|--------------|
| `(3,4),(5,4)` | `(3,5)` |
| `(2,3,4),(5,4)` | `(2,3,5)` |
| `(3,4),(4,)` | `(3,)` |
| `(4,),(4,)` | `()` |

All match `np.inner`.

### Implementation

- `keras/src/ops/numpy.py:Inner.compute_output_spec`: compute `x1.shape[:-1] + x2.shape[:-1]` instead of returning a hardcoded scalar. Also validate that the last dimensions match (when both are concrete ints), raising a clear `ValueError` rather than letting the eager backend fail later.
- Updated the static-shape test (which had encoded the buggy `()` result for `(2,3),(2,3)`) to the correct `(2,2)`, and added eager + symbolic coverage for the multi-dimensional case.

## Contributor Agreement

### DIFF
--- a/keras/src/ops/numpy.py
+++ b/keras/src/ops/numpy.py
@@ -8270,7 +8270,22 @@ class Inner(Operation):
             getattr(x1, "dtype", type(x1)),
             getattr(x2, "dtype", type(x2)),
         )
-        return KerasTensor([], dtype=dtype)
+        x1_shape = getattr(x1, "shape", ())
+        x2_shape = getattr(x2, "shape", ())
+        # `inner` sums over the last axis of each input, so the last
+        # dimensions must match. The output shape is the concatenation of
+        # the leading dimensions of both inputs.
+        if x1_shape and x2_shape:
+            d1 = x1_shape[-1]
+            d2 = x2_shape[-1]
+            if isinstance(d1, int) and isinstance(d2, int) and d1 != d2:
+                raise ValueError(
+                    "`inner` requires the last dimension of `x1` and `x2` "
+                    f"to match. Received: x1.shape={x1_shape}, "
+                    f"x2.shape={x2_shape}."
+                )
+        output_shape = tuple(x1_shape[:-1]) + tuple(x2_shape[:-1])
+        return KerasTensor(output_shape, dtype=dtype)
 
 
 @keras_export(["keras.ops.inner", "keras.ops.numpy.inner"])
--- a/keras/src/ops/numpy.py
+++ b/keras/src/ops/numpy.py
@@ -8273,21 +8273,25 @@ class Inner(Operation):
             getattr(x1, "dtype", type(x1)),
             getattr(x2, "dtype", type(x2)),
         )
-        x1_shape = getattr(x1, "shape", ())
-        x2_shape = getattr(x2, "shape", ())
-        # `inner` sums over the last axis of each input, so the last
-        # dimensions must match. The output shape is the concatenation of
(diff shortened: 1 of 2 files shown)


rstar327:
Addressed the coverage gap — and the new test for it actually surfaced a real edge case I'd missed: a **scalar operand**. `np.inner` treats a 0-d input as scalar multiplication, so e.g. `inner((), (3,))` -> `(3,)` and `inner((2,3), ())` -> `(2,3)`, not `()`. The previous version computed `x1.shape[:-1] + x2.shape[:-1]` which gave the wrong shape for scalar inputs.

Updated `Inner.compute_output_spec` to handle scalar operands explicitly, and added symbolic test cases covering the scalar path, the dynamic-last-dim path (skips the static match check), and the existing mismatch-raise path. All branches in the new code are now exercised. 48 `inner` tests pass.

hertschuh (approved):
Thanks for the fix!

PR #23000, opened 2026-05-27
Make EinsumDense.compute_output_shape work before build

### LINKED ISSUE #22999: [Bug] EinsumDense.compute_output_shape raises AttributeError before the layer is built
### Summary

`keras.layers.EinsumDense.compute_output_shape` raises `AttributeError: 'EinsumDense' object has no attribute 'full_output_shape'` when called before the layer is built. `compute_output_shape` is a public API and is expected to work on an unbuilt layer (e.g. for shape inference, in `Sequential.compute_output_spec`, or when a user calls it directly).

### Reproduction

```python
import keras
from keras import layers

layer = layers.EinsumDense("ab,bc->ac", output_shape=(7,))
layer.compute_output_shape((None, 5))
# AttributeError: 'EinsumDense' object has no attribute 'full_output_shape'
```

(It works only if you first build/call the layer.)

### Root cause

`EinsumDense.compute_output_shape` ignores its `input_shape` argument and returns the cached `self.full_output_shape`:

```python
def compute_output_shape(self, _):
    return self.full_output_shape
```

`self.full_output_shape` is only assigned in `build()`, so before build the attribute doesn't exist. Every other core layer's `compute_output_shape` derives the result from the passed `input_shape` and works pre-build.

### Fix

Compute the output shape from the `input_shape` argument using the same `_analyze_einsum_string` helper that `build` uses, so it works whether or not the layer has been built (and stays consistent with the actual call).

### Impact

Any code path that calls `EinsumDense.compute_output_shape` on an unbuilt layer crashes — e.g. shape inference utilities, model-introspection tools, or nesting an unbuilt `EinsumDense` where a parent needs its output shape.

rstar327:
## Description

Fixes #22999.

`EinsumDense.compute_output_shape` raised `AttributeError: 'EinsumDense' object has no attribute 'full_output_shape'` when called before the layer was built, because it returned the cached `self.full_output_shape` (only set in `build()`) and ignored its `input_shape` argument.

### Before

```python
import keras
from keras import layers

layer = layers.EinsumDense("ab,bc->ac", output_shape=(7,))
layer.compute_output_shape((None, 5))
# AttributeError: 'EinsumDense' object has no attribute 'full_output_shape'
```

### After

```python
layer.compute_output_shape((None, 5))   # (None, 7)
```

### Fix

`compute_output_shape` now derives the result from its `input_shape` argument via the same `_analyze_einsum_string` helper that `build()` uses, so it works whether or not the layer is built and stays consistent with the actual call (verified, including equations with an explicitly-unknown output axis).

Added `test_compute_output_shape_before_build`.

## Contributor Agreement

### DIFF
--- a/keras/src/layers/core/einsum_dense.py
+++ b/keras/src/layers/core/einsum_dense.py
@@ -303,8 +303,16 @@ class EinsumDense(Layer):
 
         return kernel
 
-    def compute_output_shape(self, _):
-        return self.full_output_shape
+    def compute_output_shape(self, input_shape):
+        # Derive the output shape from `input_shape` so this works before the
+        # layer is built (`self.full_output_shape` is only set in `build`).
+        _, _, full_output_shape, _, _ = _analyze_einsum_string(
+            self.equation,
+            self.bias_axes,
+            input_shape,
+            self.partial_output_shape,
+        )
+        return tuple(full_output_shape)
 
     def call(self, inputs, training=None):
         x = ops.einsum(self.equation, inputs, self.kernel)
--- a/keras/src/layers/core/einsum_dense.py
+++ b/keras/src/layers/core/einsum_dense.py
@@ -184,10 +184,7 @@ class EinsumDense(Layer):
             input_shape,
             self.partial_output_shape,
         )
-        kernel_shape, bias_shape, full_output_shape, input_axes, output_axes = (
-            shape_data
-        )
-        self.full_output_shape = tuple(full_output_shape)
+        kernel_shape, bias_shape, _, input_axes, output_axes = shape_data
         self.input_spec = InputSpec(ndim=len(input_shape))
 
         kernel_initializer = self.kernel_initializer
@@ -304,8 +301,6 @@ class EinsumDense(Layer):
         return kernel
 
     def compute_output_shape(self, input_shape):
-        # Derive the output shape from `input_shape` so this works before the
-        # layer is built (`self.full_output_shape` is only set in `build`).
         _, _, full_output_shape, _, _ = _analyze_einsum_string(
             self.equation,
             self.bias_axes,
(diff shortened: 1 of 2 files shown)


keras/src/layers/core/einsum_dense.py
hertschuh: Remove `self.full_output_shape = tuple(full_output_shape)`, this is no longer used.

keras/src/layers/core/einsum_dense.py
```diff
@@ -303,8 +303,16 @@ def kernel(self):
 
         return kernel
 
-    def compute_output_shape(self, _):
-        return self.full_output_shape
+    def compute_output_shape(self, input_shape):
+        # Derive the output shape from `input_shape` so this works before the
+        # layer is built (`self.full_output_shape` is only set in `build`).
```
hertschuh: Remove this comment, this won't make any sense outside of this PR and once `self.full_output_shape` is removed.

The new way is the normal way to do it.

keras/src/layers/core/einsum_dense_test.py
```diff
@@ -349,6 +349,26 @@ def test_einsum_dense_basics(
         if expected_bias_shape is not None:
             self.assertEqual(layer.bias.shape, expected_bias_shape)
 
+    def test_compute_output_shape_before_build(self):
+        # `compute_output_shape` must work without the layer being built
+        # first (it previously read `self.full_output_shape`, which only
+        # exists after `build`, and raised `AttributeError`).
```
hertschuh: Remove this part of the comment, this won't make any sense outside of this PR and once `self.full_output_shape` is removed.

hertschuh (changes requested):
Thanks for the fix!

rstar327:
Thanks! All three addressed:

1. Removed `self.full_output_shape = tuple(full_output_shape)` from `build` (and dropped it from the unpacking — it's no longer used anywhere).
2. Removed the PR-specific comment in `compute_output_shape`.
3. Removed the `self.full_output_shape`/AttributeError context from the test comment.

Confirmed `self.full_output_shape` has no other references in the codebase. All 99 EinsumDense tests pass.

hertschuh (approved):
Thanks!

PR #23010, opened 2026-05-30
Fix: reject decompression-bomb archive members when loading a `.keras` model

LinZiyuu:
## Summary

`_load_model_from_fileobj` (behind `keras.saving.load_model`) reads
`config.json`, the embedded `model.weights.h5`, and the shard map straight
into memory (or extracts them to disk) with no bound on the decompressed size.
The `config.json` read is the very first thing the loader does — before the
model is built, before weights are touched, and regardless of `safe_mode`.

A ZIP member can store almost nothing on disk yet declare an enormous
uncompressed size. A crafted `.keras` of a few MB can therefore declare a
multi-gigabyte member; reading it allocates memory for the **declared** size
(OOM-kill) or, on the extract-to-disk weights path, writes the full decompressed
member to disk (disk exhaustion) — CWE-409. No Lambda, no custom objects, and
`safe_mode=True` does not help (the read precedes deserialization).

Verified on master: a 1.5 MB `.keras` (config.json declaring 1.5 GiB) drives
`keras.saving.load_model(..., safe_mode=True)` toward a multi-GiB allocation;
scaling the member up reaches OOM.

## Relationship to CVE-2026-0897

CVE-2026-0897 / GHSA-mgx6 addressed the same class one layer down: an HDF5
*dataset* that declares far more data than it stores, guarded in
`safe_get_h5_dataset` by a declared-vs-stored ratio check. The `.keras`
*container* itself was not covered — the same "tiny on disk, huge in memory"
primitive is still reachable through the archive members the loader reads
before (and around) the HDF5 layer. This PR adds the equivalent guard at the
ZIP-member layer.

## Fix

`keras/src/saving/saving_lib.py`:

- Add `_reject_zip_bomb(archive, name)` — rejects a member whose declared size
  (`ZipInfo.file_size`) both exceeds a floor (4 GiB, matching the HDF5 dataset
  guard) and exceeds its on-disk size (`ZipInfo.compress_size`) by more than
  100x — and `_safe_zip_read` (checks, then reads into memory).
- Guard `config.json` and the sharded `weight_map` via `_safe_zip_read`.
- Guard `model.weights.h5` via `_reject_zip_bomb` placed **before** the
  in-memory / extract-to-disk / on-the-fly `try/except`. The bare `except`
  there falls back to reading the weights on the fly, so a check *inside* it
  would be swallowed and the bomb still read; checking up front covers all
  three weights-read paths at once.

Keras writes archive members uncompressed (ratio ~1:1) and DEFLATE cannot
exceed ~1032:1, so genuine artifacts — including arbitrarily large legitimate
models — stay well under the factor; the storage-ratio check (rather than a
flat cap) keeps large legitimate models loadable.

## Test plan

`keras/src/saving/saving_lib_test.py` — new `SafeZipReadTest`:

- `test_rejects_decompression_bomb`: a highly-compressible member is rejected
  by `_safe_zip_read`.
- `test_allows_incompressible_member`: an ordinary (stored) member is returned
  unchanged.
- `test_load_model_rejects_bomb_config`: end-to-end, `keras.saving.load_model`
  on a tiny `.keras` whose `config.json` decompresses huge is rejected.
- `test_load_model_rejects_bomb_weights`: end-to-end, a bomb `model.weights.h5`
  member is rejected up front (covering the extract/on-the-fly paths).

```
$ pytest keras/src/saving/saving_lib_test.py -k SafeZipRead
4 passed
```

`ruff check` and `ruff format` pass on the changed files.

## Disclosure

Reported via huntr (decompression-bomb sibling of CVE-2026-0897 /
GHSA-mgx6-5cf9-rr43 at the `.keras` archive-container layer).

## Contributor agreement

### DIFF
--- a/keras/src/saving/saving_lib.py
+++ b/keras/src/saving/saving_lib.py
@@ -431,10 +431,43 @@ def _model_from_config(config_json, custom_objects, compile, safe_mode):
     return model
 
 
+# Guard against ZIP "decompression bomb" archive members (CWE-409): a member
+# can store almost nothing on disk yet declare an enormous uncompressed size,
+# forcing a huge allocation when it is read into memory. Keras writes archive
+# members uncompressed (stored, ratio ~1:1) and DEFLATE cannot exceed ~1032:1,
+# so a member that both exceeds the floor and expands to more than
+# `_ZIP_MEMBER_MAX_EXPANSION`x its on-disk size is a bomb, not a genuine
+# artifact (this leaves ample headroom for an incidentally recompressed file).
+# The 4 GiB floor matches the HDF5 dataset guard for CVE-2026-0897.
+_ZIP_MEMBER_BOMB_FLOOR_BYTES = 1 << 32  # 4 GiB
+_ZIP_MEMBER_MAX_EXPANSION = 100
+
+
+def _reject_zip_bomb(archive, name):
+    """Raise if a ZIP member is a decompression bomb (see CWE-409)."""
+    info = archive.getinfo(name)
+    if (
+        info.file_size > _ZIP_MEMBER_BOMB_FLOOR_BYTES
+        and info.file_size > _ZIP_MEMBER_MAX_EXPANSION * info.compress_size
+    ):
+        raise ValueError(
+            f"Not allowed: ZIP member '{name}' declares "
+            f"{readable_memory_size(info.file_size)} but only "
+            f"{readable_memory_size(info.compress_size)} are stored on disk; "
+            "refusing to load a potential decompression bomb."
+        )
+
+
+def _safe_zip_read(archive, name):
+    """Read a ZIP member into memory, rejecting bombs (see CWE-409)."""
+    _reject_zip_bomb(archive, name)
+    with archive.open(name, "r") as f:
+        return f.read()
+
+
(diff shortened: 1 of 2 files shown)


keras/src/saving/saving_lib.py
```diff
@@ -446,6 +479,12 @@ def _load_model_from_fileobj(fileobj, custom_objects, compile, safe_mode):
         asset_store = None
         try:
             if _VARS_FNAME_H5 in all_filenames:
+                # Reject a decompression-bomb weights member up front, outside
+                # the try/except below: the bare `except` falls back to reading
+                # the weights on the fly, so a check inside it would be
+                # swallowed and the bomb still read. Checking here covers the
+                # in-memory, extract-to-disk, and on-the-fly paths at once.
```
hertschuh, line 486: Remove this comment.

hertschuh (approved):
Thanks!

LinZiyuu:
Thanks for merging this! This PR fixes the vulnerability I reported on huntr. Whenever you have a moment, could you validate the report on the platform? Link: https://huntr.com/bounties/952d310a-38df-4572-af12-6e56e4931e7b

PR #23007, opened 2026-05-29
Fix for flaky BatchNormalization renorm test.

hertschuh:
## Contributor Agreement

- [x] I am a human, and not a bot.
- [x] I will be responsible for responding to review comments in a timely manner.
- [x] I will work with the maintainers to push this PR forward until submission.

PR #23012, opened 2026-05-31
Optimize vectorized_map using list comprehension

andersendsa:
## Description

- Replaced the `for` loop and `.append()` call in `vectorized_map` with a list comprehension.
- This avoids attribute lookup overhead and leverages C-level list construction execution, improving performance.

## Contributor Agreement

### DIFF
--- a/keras/src/layers/core/einsum_dense_test.py
+++ b/keras/src/layers/core/einsum_dense_test.py
@@ -661,6 +661,7 @@ class EinsumDenseTest(testing.TestCase):
     ):
         layer = layers.EinsumDense(equation=equation, output_shape=output_shape)
         layer.build(input_shape)
+        random.seed(1337)
         x = ops.random.uniform(input_shape)
         y_float = layer(x)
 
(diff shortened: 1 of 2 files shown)


keras/src/layers/core/einsum_dense_test.py
```diff
@@ -679,6 +661,7 @@ def test_quantize_with_specific_equations(
     ):
         layer = layers.EinsumDense(equation=equation, output_shape=output_shape)
         layer.build(input_shape)
+        random.seed(1337)
```
hertschuh: This is unrelated so remove from this PR.

I do realize that this test has been flaky and something like this would help, however, it needs to move line 662 and it needs to be `utils.set_random_seed(1337)` before creating the layer so the random weights of the layers are also deterministic.

https://github.com/keras-team/keras/pull/23007/changes

hertschuh (changes requested):

keras/src/layers/core/einsum_dense_test.py
hertschuh: Oh, and in it's current form it's making the tests fail too.

andersendsa:
hi @hertschuh  i have done the changes and this pr is ready for review

hertschuh (approved):
Thanks!

PR #23013, opened 2026-05-31
Fix torch median to average the two middle values for even-length inputs

MarcosAsh:
## Summary

`keras.ops.median` returns the wrong value on the torch backend for any reduction over an even number of elements. `torch.median` returns the lower of the two middle values, whereas numpy (which `keras.ops` mirrors) and the jax/tensorflow backends average them.

```python
import numpy as np
from keras import ops
# torch backend
ops.median(ops.convert_to_tensor(np.array([1., 2., 3., 4.])))  # 2.0  (wrong)
# numpy / jax / tensorflow backends, and np.median: 2.5
```

This was wrong on every code path (full reduction, single axis, multiple axes, with and without `keepdims`).

The fix delegates to `quantile(x, q=0.5, ...)`, whose default `method="linear"` averages the two middle values, matching numpy. This mirrors how `nanmedian` in the same module is already implemented (it delegates to `nanquantile(q=0.5)`), and lets the multi-axis / `keepdims` handling reuse the tested `quantile` path instead of a separate implementation.

The existing correctness test only used `[[1, 2, 3], [3, 2, 1]]`, where the two middle values of the flattened array are both `2` (so the lower value coincidentally equals the average) and the `axis=1` reductions are odd-length. No assertion ever exercised an even-length reduction whose middle values differ. This PR adds an even-length case.

## Contributor Agreement

### DIFF
--- a/keras/src/backend/torch/numpy.py
+++ b/keras/src/backend/torch/numpy.py
@@ -1265,44 +1265,12 @@ def maximum(x1, x2):
 
 def median(x, axis=None, keepdims=False):
     x = convert_to_tensor(x)
-    compute_dtype = dtypes.result_type(x.dtype, "float32")
-    result_dtype = dtypes.result_type(x.dtype, float)
-    x = cast(x, compute_dtype)
 
-    if axis is None and keepdims is False:
-        return cast(torch.median(x), result_dtype)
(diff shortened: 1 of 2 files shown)


hertschuh (approved):
Thanks for the fix!

PR #23022, opened 2026-06-02
Suppress spurious 'Skipping nested container' warning on freshly-saved files

### LINKED ISSUE #23021: [Bug] Saving emits spurious 'Skipping nested container' warning on freshly-saved .keras files
### Summary

`keras.saving.load_model` emits multiple `UserWarning: Skipping nested container at '…': no matching group found in the saved file. This usually means the file was saved with a Keras version that did not serialize sublayers nested inside containers. Affected layers will retain their freshly-initialized weights.` even when loading a file that was **just saved with the current version**. The wording suggests weight loss, but in practice the weights round-trip correctly — the warning is a false positive.

The warning was introduced in #22362 to catch legacy files predating nested-container serialization, but it currently also fires for two benign cases:

1. **Containers of pure metadata** — e.g. `variable_serialization_spec`, which is a list of strings (axis/variable names). It holds no `KerasSaveable` and can't have anything to load.
2. **Containers that mirror already-loaded saveables** — e.g. a Functional model's `_operations_by_depth` / `_nodes_by_depth`, whose items are the same layers already loaded via the `layers` collection.

### Reproduction

```python
import os, warnings, tempfile
import keras
from keras import layers

inp = layers.Input(shape=(5,))
model = keras.Model(inp, layers.EinsumDense("ab,bc->ac", output_shape=(7,))(inp))
with tempfile.NamedTemporaryFile(suffix=".keras", delete=False) as f:
    path = f.name
model.save(path)
with warnings.catch_warnings(record=True) as w:
    warnings.simplefilter("always")
    keras.models.load_model(path)
print(sum("Skipping nested container" in str(x.message) for x in w))
# Prints 2+ even though the model is freshly saved.
```

Weights actually round-trip correctly — the warning text claiming "Affected layers will retain their freshly-initialized weights" is incorrect for this case.

### Expected

No `Skipping nested container` warning on a freshly-saved file. The warning should still fire for legacy files that genuinely lack a needed nested-container group (covered by `test_legacy_load_skips_missing_nested_containers`).

### Suggested fix

In `keras/src/saving/saving_lib._load_container_state`, before warning, only fire if the in-memory container transitively holds at least one **unvisited** `KerasSaveable`. Pure-metadata containers and containers that mirror already-loaded saveables don't satisfy this and shouldn't warn.

### Impact

Users see scary, repetitive warnings on every `load_model` call with common layers (`EinsumDense`, `Embedding`, `MultiHeadAttention`, every Functional model …) suggesting their weights weren't loaded — when in fact loading worked correctly. Hard to debug, noisy in production, undermines trust.

rstar327:
## Description

Fixes #23021.

`load_model` emits multiple `UserWarning: Skipping nested container at '...': no matching group found in the saved file. ... Affected layers will retain their freshly-initialized weights.` even when loading a file that was just saved with the current version. The wording suggests weight loss, but weights round-trip correctly — it's a false positive.

The warning was added in #22362 to catch legacy files predating nested-container serialization, but it also fires for two benign cases on current-version files:

1. **Containers of pure metadata** — e.g. `variable_serialization_spec`, a list of strings (axis/variable names). No `KerasSaveable` inside, nothing to load.
2. **Containers that mirror already-loaded saveables** — e.g. Functional's `_operations_by_depth` / `_nodes_by_depth`, whose items are the same layers already loaded via the `layers` collection (so they're already in `visited_saveables`).

### Before

```python
inp = layers.Input(shape=(5,))
model = keras.Model(inp, layers.EinsumDense("ab,bc->ac", output_shape=(7,))(inp))
model.save("m.keras")
keras.models.load_model("m.keras")
# UserWarning: Skipping nested container at 'layers/einsum_dense/variable_serialization_spec/container': ...
# UserWarning: Skipping nested container at '_operations_by_depth/container': ...
# (etc., 4+ warnings, all spurious)
```

### After

Zero spurious warnings on a freshly-saved file. Verified that weights still round-trip and outputs match for Dense, EinsumDense, Embedding+Dense, MultiHeadAttention.

### Implementation

In `_load_container_state`, replaced the unconditional warning with a check via a new `_container_has_unvisited_saveables(container, visited_saveables)`:

- Returns `True` only when the container transitively holds at least one `KerasSaveable` that hasn't been loaded yet via another path.
- Pure-metadata containers (e.g. `variable_serialization_spec`) return `False`.
- Containers like `_operations_by_depth` that mirror `layers` return `False` because every saveable in them is already in `visited_saveables`.

The legitimate legacy-file warning path is preserved — `test_legacy_load_skips_missing_nested_containers` still emits and asserts the warning correctly.

Added `test_fresh_save_load_emits_no_container_skip_warning` covering the false-positive case. All 63 saving tests pass.

## Contributor Agreement

### DIFF
--- a/keras/src/saving/saving_lib.py
+++ b/keras/src/saving/saving_lib.py
@@ -883,6 +883,37 @@ def _container_path_present(weights_store, assets_store, path):
     )
 
 
+def _container_has_unvisited_saveables(container, visited_saveables, seen=None):
+    """True if `container` transitively holds an unvisited `KerasSaveable`.
+
+    A container can show up in the in-memory traversal even when its state
+    has no corresponding group in the saved file — either because it holds
+    no `KerasSaveable` at all (e.g. lists of strings inside
+    `variable_serialization_spec`), or because every saveable in it was
+    already loaded via another path (e.g. `_operations_by_depth` mirrors
+    the `layers` collection that's loaded first). In neither case is a
+    missing group an error worth warning about.
+    """
+    from keras.src.saving.keras_saveable import KerasSaveable
+
+    if seen is None:
+        seen = set()
+    if id(container) in seen:
+        return False
+    seen.add(id(container))
+    if isinstance(container, dict):
+        container = container.values()
+    for item in container:
+        if isinstance(item, KerasSaveable):
+            if visited_saveables is None or id(item) not in visited_saveables:
+                return True
+        elif isinstance(
+            item, (list, dict, tuple)
+        ) and _container_has_unvisited_saveables(item, visited_saveables, seen):
+            return True
+    return False
+
+
 def _save_container_state(
     container,
     weights_store,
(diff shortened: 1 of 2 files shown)


keras/src/saving/saving_lib.py
```diff
@@ -1015,19 +1007,28 @@ def _load_container_state(
...
+                    # freshly-initialized weights).
+                    warnings.warn(
+                        f"Skipping nested container at '{nested_path}': no "
+                        "matching group found in the saved file. This "
+                        "usually means the file was saved with a Keras "
+                        "version that did not serialize sublayers nested "
+                        "inside containers. Affected layers will retain "
+                        "their freshly-initialized weights.",
+                        stacklevel=2,
+                    )
```
hertschuh: This new `_container_has_unvisited_saveables` function replicates quite a bit of the code in `_load_container_state`.

I'd like to reuse `_load_container_state` instead. I think one way it could work is:

```python
            if not _container_path_present(
                weights_store, assets_store, nested_path
            ):
                failed_before_count = len(failed_saveables)
                _load_container_state(..., skip_mismatch=True, ...)
                if len(failed_saveables) != failed_before_count:
                    warnings.warn(...)
            else:
                _load_container_state(...)
```
Let me know if this doesn't work.

hertschuh (changes requested):

rstar327:
Thanks — that's much cleaner. Refactored exactly as suggested:

- Dropped the standalone `_container_has_unvisited_saveables` helper.
- When `_container_path_present` returns False, recursively call `_load_container_state` with `skip_mismatch=True` and watch `failed_saveables` for new entries. If any appeared, the container genuinely held an unvisited `KerasSaveable` that needed loading → emit the legacy-skip warning. Otherwise stay silent.

Net diff is smaller (`+26 / -40`). Verified:
- Fresh save/load on Dense / EinsumDense / Embedding+Dense / MultiHeadAttn → 0 spurious warnings, weights still match (`test_fresh_save_load_emits_no_container_skip_warning`).
- Existing `test_legacy_load_skips_missing_nested_containers` still triggers the warning correctly.
- All 64 saving tests pass.

hertschuh (approved):
Great, thanks!

PR #23023, opened 2026-06-02
Save data_format in image preprocessing layers' get_config

MarcosAsh:
## Summary

`BaseImagePreprocessingLayer` stores `data_format` but its `get_config` never emits it. 22 of the 31 image preprocessing layers (RandomHue, RandomContrast, CutMix, MixUp, AugMix, RandAugment, Solarization, etc.) rely on the base class to serialize it, so a `from_config(get_config())` round-trip (e.g. `save` / `load_model`) silently resets `data_format` to the global image data format. A model built with `channels_first` augmentation layers reloads as `channels_last` and transforms the wrong axes.

The other 9 layers in the family (Resizing, CenterCrop, RandomFlip, ...) already emit `data_format` in their own `get_config`, and every other layer with a `data_format` argument (conv, pooling, adaptive pooling) saves it correctly - this family is the only gap.

The fix emits `data_format` from the base class `get_config`, covering all subclasses at once.

`run_layer_test` did not catch this because it compares the revived layer's re-serialized config against the original config, and since both omitted `data_format` they matched while the behavior silently changed. This PR adds a round-trip test that constructs each of the 31 layers with a non-default `data_format`, reserializes, and asserts it is preserved.

## Contributor Agreement

### DIFF
--- a/keras/src/layers/preprocessing/image_preprocessing/base_image_preprocessing_layer.py
+++ b/keras/src/layers/preprocessing/image_preprocessing/base_image_preprocessing_layer.py
@@ -256,6 +256,7 @@ class BaseImagePreprocessingLayer(DataLayer):
 
     def get_config(self):
         config = super().get_config()
+        config.update({"data_format": self.data_format})
         if self.bounding_box_format is not None:
             config.update(
                 {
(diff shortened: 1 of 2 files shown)


hertschuh (approved):
Thanks for the fix!

PR #23027, opened 2026-06-03
Fix TF backend depthwise and separable conv with stride and dilation

### LINKED ISSUE #22515: keras.ops.separable_conv returns incorrect static shape for channels_first with dilation and stride
### Summary
`keras.ops.separable_conv` computes a different output shape for symbolic tensors than for eager tensors when using `data_format="channels_first"` with `strides > 1` and `dilation_rate > 1`.

### Reproduction
```python
import keras
import tensorflow as tf

x = tf.random.normal((1, 3, 10, 10))
dw = tf.random.normal((3, 3, 3, 1))
pw = tf.random.normal((1, 1, 3, 6))

print(keras.ops.separable_conv(x, dw, pw, strides=2, padding="valid",
                               data_format="channels_first", dilation_rate=3).shape)

x_sym = keras.Input(shape=(3, 10, 10))
print(keras.ops.separable_conv(x_sym, dw, pw, strides=2, padding="valid",
                               data_format="channels_first", dilation_rate=3).shape)
```

### Expected
Both eager and symbolic calls should infer the same output shape.

### Actual
- Eager output shape: `(1, 6, 1, 1)`
- Symbolic output shape: `(None, 6, 2, 2)`

### Impact
This breaks shape inference for models using `keras.ops.separable_conv` in `channels_first` mode with dilation/stride combinations.

Lawson-Darrow:
## Description

Fixes #22515.

On the TensorFlow backend, `keras.ops.separable_conv` and `keras.ops.depthwise_conv` returned the wrong output when `strides > 1` and `dilation_rate > 1` were used together. The JAX and Torch backends were already correct.

Repro (channels_first, kernel 3, stride 2, dilation 3, input `(1, 3, 10, 10)`):

```python
import numpy as np, keras
from keras import ops

x = np.zeros((1, 3, 10, 10), "float32")
dw = np.zeros((3, 3, 3, 1), "float32")
pw = np.zeros((1, 1, 3, 6), "float32")
kw = dict(strides=2, padding="valid", data_format="channels_first", dilation_rate=3)

ops.separable_conv(ops.convert_to_tensor(x), dw, pw, **kw).shape       # (1, 6, 1, 1) wrong
ops.separable_conv(keras.KerasTensor((1, 3, 10, 10)), dw, pw, **kw).shape  # (1, 6, 2, 2) correct
```

The cause is in the TF ops, not in Keras. `tf.nn.depthwise_conv2d`, `tf.nn.separable_conv2d`, and `tf.nn.convolution` all reject `strides > 1` together with `dilation_rate > 1` (TF: "when dilation rate is greater than 1, then all values of strides must be 1") and fall back to a degenerate result. Wrapping the op in an XLA `tf.function` does not help. The limitation is enforced at trace time.

**Fix** (`keras/src/backend/tensorflow/nn.py`):

- When `strides > 1` and `dilation_rate > 1`, run the dilated conv at stride 1 (which TF allows) and subsample the output by the stride. For `valid` padding the subsample is exact. For `same` padding the input is explicitly padded first using TF's SAME sizing so the stride-1 valid conv matches.
- The fallback is gated to only the unsupported combination, so the common fused depthwise and separable paths are unchanged. It handles depthwise and separable, 1D and 2D, and both `channels_first` and `channels_last`.

**Tests** (`keras/src/ops/nn_test.py`):

- Removed the two `"This case is not supported by the TF backend"` skips in `test_depthwise_conv_2d` and `test_separable_conv_2d`. Those tests already exercise `strides=2` with `dilation_rate=(2, 2)` and compare against a numpy reference, so they now act as regression coverage for this fix on the TF backend (JAX and Torch already ran them).

**Validation** (tensorflow backend, CPU):

- The decomposition matches the native TF result exactly across 192 supported-combination checks (depthwise and separable, 1D and 2D, channels_first and channels_last, depth multiplier 1 and 2, valid and same padding).
- The previously skipped value tests now pass and the full conv test suite stays green.
- `ruff check` and `ruff format --check` are clean on the changed files.

**AI assistance disclosure (per the AI-Assisted Contribution Policy):** This PR was prepared with the assistance of Claude Code. The human author has reviewed and understands the change, takes full responsibility for it, and will respond to review feedback personally.

## Contributor Agreement

### DIFF
--- a/keras/src/ops/nn_test.py
+++ b/keras/src/ops/nn_test.py
@@ -1994,6 +1994,181 @@ class NNOpsCorrectnessTest(testing.TestCase):
         )
         self.assertAllClose(outputs, expected, tpu_atol=1e-2, tpu_rtol=1e-2)
 
+    @parameterized.product(
+        padding=("valid", "same"),
+        data_format=("channels_first", "channels_last"),
+    )
+    def test_depthwise_conv_2d_stride_dilation(self, padding, data_format):
+        # Regression coverage for #22515: on the TF backend, strides > 1
+        # together with dilation_rate > 1 is computed via a decomposition.
+        # Exercise both data formats so the channels_first transpose path in
+        # that decomposition is covered, not just the default layout.
+        strides, dilation_rate = 2, (2, 2)
+        if data_format == "channels_last":
+            input_shape = (2, 10, 10, 3)
+        else:
+            input_shape = (2, 3, 10, 10)
+        inputs_2d = np.arange(600, dtype=float).reshape(input_shape)
+        kernel = np.arange(24, dtype=float).reshape([2, 2, 3, 2])
+
+        outputs = knn.depthwise_conv(
+            inputs_2d,
+            kernel,
+            strides,
+            padding=padding,
+            data_format=data_format,
+            dilation_rate=dilation_rate,
+        )
+        expected = np_depthwise_conv2d(
+            inputs_2d,
+            kernel,
+            bias_weights=np.zeros((6,)),
+            strides=strides,
+            padding=padding,
+            data_format=data_format,
+            dilation_rate=dilation_rate,
+        )
+        self.assertAllClose(outputs, expected, tpu_atol=1e-2, tpu_rtol=1e-2)
+
+    @parameterized.product(
+        padding=("valid", "same"),
+        data_format=("channels_first", "channels_last"),
+    )
+    def test_separable_conv_2d_stride_dilation(self, padding, data_format):
+        # Regression coverage for #22515, separable variant, both layouts.
+        strides, dilation_rate = 2, (2, 2)
+        if data_format == "channels_last":
+            input_shape = (2, 10, 10, 3)
+        else:
+            input_shape = (2, 3, 10, 10)
+        inputs_2d = np.arange(600, dtype=float).reshape(input_shape)
+        depthwise_kernel = np.arange(24, dtype=float).reshape([2, 2, 3, 2])
+        pointwise_kernel = np.arange(72, dtype=float).reshape([1, 1, 6, 12])
+
+        outputs = knn.separable_conv(
+            inputs_2d,
+            depthwise_kernel,
+            pointwise_kernel,
+            strides,
+            padding=padding,
+            data_format=data_format,
+            dilation_rate=dilation_rate,
+        )
+        expected_depthwise = np_depthwise_conv2d(
+            inputs_2d,
+            depthwise_kernel,
+            np.zeros(6),
+            strides=strides,
+            padding=padding,
+            data_format=data_format,
+            dilation_rate=dilation_rate,
+        )
+        expected = np_conv2d(
+            expected_depthwise,
+            pointwise_kernel,
+            np.zeros(6 * 12),
+            strides=1,
+            padding=padding,
+            data_format=data_format,
+            dilation_rate=dilation_rate,
+            groups=1,
+        )
+        self.assertAllClose(outputs, expected, tpu_atol=1e-2, tpu_rtol=1e-2)
+
+    @parameterized.product(
+        padding=("valid", "same"),
+        data_format=("channels_first", "channels_last"),
+    )
+    def test_depthwise_conv_1d_stride_dilation(self, padding, data_format):
+        # Regression coverage for #22515, 1D variant. The TF decomposition
+        # lifts 1D to 2D internally; validate against the trusted 2D
+        # reference lifted the same way (singleton height axis).
+        strides, dilation_rate = 2, 2
+        if data_format == "channels_last":
+            input_shape = (2, 10, 3)
+        else:
+            input_shape = (2, 3, 10)
+        inputs_1d = np.arange(60, dtype=float).reshape(input_shape)
+        kernel = np.arange(12, dtype=float).reshape([2, 3, 2])
+        if data_format == "channels_last":
+            x2d = inputs_1d[:, None, :, :]
+            squeeze_axis = 1
+        else:
+            x2d = inputs_1d[:, :, None, :]
+            squeeze_axis = 2
+
+        outputs = knn.depthwise_conv(
+            inputs_1d,
+            kernel,
+            strides,
+            padding=padding,
+            data_format=data_format,
+            dilation_rate=dilation_rate,
+        )
+        expected_2d = np_depthwise_conv2d(
+            x2d,
+            kernel[None],
+            bias_weights=np.zeros((6,)),
+            strides=(1, strides),
+            padding=padding,
+            data_format=data_format,
+            dilation_rate=(1, dilation_rate),
+        )
+        expected = np.squeeze(expected_2d, axis=squeeze_axis)
+        self.assertAllClose(outputs, expected, tpu_atol=1e-2, tpu_rtol=1e-2)
+
+    @parameterized.product(
+        padding=("valid", "same"),
+        data_format=("channels_first", "channels_last"),
+    )
+    def test_separable_conv_1d_stride_dilation(self, padding, data_format):
+        # Regression coverage for #22515, 1D separable variant.
+        strides, dilation_rate = 2, 2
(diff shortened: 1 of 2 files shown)


Lawson-Darrow:
Thanks, good calls. Pushed an update: the SAME padding helper now does the integer math directly on the static strides and dilation instead of converting them to tensors, and the strided subsample is inlined with plain slicing, so the `_spatial_strided_slice_channels_last` helper is gone. Validation still passes, both the previously skipped TF value tests and the cross-check against the native ops.

keras/src/ops/nn_test.py
```diff
@@ -2009,6 +1994,181 @@ def test_separable_conv_2d(self, strides, padding, dilation_rate):
...
+        expected = np_depthwise_conv2d(
+            inputs_2d,
+            kernel,
+            bias_weights=np.zeros((6,)),
+            strides=strides,
+            padding=padding,
+            data_format=data_format,
+            dilation_rate=dilation_rate,
+        )
+        self.assertAllClose(outputs, expected, tpu_atol=1e-2, tpu_rtol=1e-2)
```
hertschuh: Instead of replicating all the code, simply add this line 1925 to augment the existing `test_depthwise_conv_2d`:
```python
 data_format=("channels_first", "channels_last"),
```

keras/src/ops/nn_test.py
```diff
@@ -2009,6 +1994,181 @@ def test_separable_conv_2d(self, strides, padding, dilation_rate):
...
+            expected_depthwise,
+            pointwise_kernel,
+            np.zeros(6 * 12),
+            strides=1,
+            padding=padding,
+            data_format=data_format,
+            dilation_rate=dilation_rate,
+            groups=1,
+        )
+        self.assertAllClose(outputs, expected, tpu_atol=1e-2, tpu_rtol=1e-2)
```
hertschuh, line 2023: Instead of replicating all the code, simply add this to augment the existing `test_separable_conv_2d`:
```python
 data_format=("channels_first", "channels_last"),
```

keras/src/ops/nn_test.py
```diff
@@ -2009,6 +1994,181 @@ def test_separable_conv_2d(self, strides, padding, dilation_rate):
...
+
+    @parameterized.product(
+        padding=("valid", "same"),
+        data_format=("channels_first", "channels_last"),
+    )
+    def test_depthwise_conv_1d_stride_dilation(self, padding, data_format):
```
hertschuh: Simply call this test `test_depthwise_conv_1d` and add this to the parameterization:
```python
        strides=(1, 2),
        dilation_rate=(1, 2),
```

keras/src/ops/nn_test.py
```diff
@@ -2009,6 +1994,181 @@ def test_separable_conv_2d(self, strides, padding, dilation_rate):
...
+            dilation_rate=(1, dilation_rate),
+        )
+        expected = np.squeeze(expected_2d, axis=squeeze_axis)
+        self.assertAllClose(outputs, expected, tpu_atol=1e-2, tpu_rtol=1e-2)
+
+    @parameterized.product(
+        padding=("valid", "same"),
+        data_format=("channels_first", "channels_last"),
+    )
+    def test_separable_conv_1d_stride_dilation(self, padding, data_format):
```
hertschuh: Simply call this test `test_separable_conv_1d` and add this to the parameterization:
```python
        strides=(1, 2),
        dilation_rate=(1, 2),
```

keras/src/ops/nn_test.py
```diff
@@ -2009,6 +1994,181 @@ def test_separable_conv_2d(self, strides, padding, dilation_rate):
...
+        self.assertAllClose(outputs, expected, tpu_atol=1e-2, tpu_rtol=1e-2)
+
+    @parameterized.product(
+        padding=("valid", "same"),
+        data_format=("channels_first", "channels_last"),
+    )
+    def test_depthwise_conv_1d_stride_dilation(self, padding, data_format):
+        # Regression coverage for #22515, 1D variant. The TF decomposition
+        # lifts 1D to 2D internally; validate against the trusted 2D
+        # reference lifted the same way (singleton height axis).
```
hertschuh: Remove this comment, it's only relevant within the context of this PR.

keras/src/ops/nn_test.py
```diff
@@ -2009,6 +1994,181 @@ def test_separable_conv_2d(self, strides, padding, dilation_rate):
...
+        )
+        expected = np.squeeze(expected_2d, axis=squeeze_axis)
+        self.assertAllClose(outputs, expected, tpu_atol=1e-2, tpu_rtol=1e-2)
+
+    @parameterized.product(
+        padding=("valid", "same"),
+        data_format=("channels_first", "channels_last"),
+    )
+    def test_separable_conv_1d_stride_dilation(self, padding, data_format):
+        # Regression coverage for #22515, 1D separable variant.
```
hertschuh: Remove this comment, it's only relevant within the context of this PR.

hertschuh (changes requested):
Thanks you for addressing this!

Some comments on the unit tests:

Lawson-Darrow:
Thanks! I tried folding `data_format=("channels_first", "channels_last")` into the existing `test_depthwise_conv_2d` / `test_separable_conv_2d` as suggested, but it turns the TF CPU job red. The regular (non stride+dilation) channels_first depthwise and separable paths pass NCHW straight to `tf.nn.depthwise_conv2d` / `separable_conv2d`, and TF CPU rejects that with `Depthwise convolution on CPU is only supported for NHWC format`. Only the 1D depthwise path and this PR's new decomposition transpose to NHWC first, so those are the channels_first cases that pass.

So channels_first coverage on these ops looks like a separate pre-existing gap from the stride+dilation fix here. Happy to either (1) keep channels_first in this PR only on the paths that support it on CPU (1D depthwise + the new stride+dilation cases), or (2) add the channels_first->NHWC transpose to the regular depthwise/separable paths too, though that feels like its own change. Which would you prefer?

hertschuh:
> Thanks! I tried folding `data_format=("channels_first", "channels_last")` into the existing `test_depthwise_conv_2d` / `test_separable_conv_2d` as suggested, but it turns the TF CPU job red. The regular (non stride+dilation) channels_first depthwise and separable paths pass NCHW straight to `tf.nn.depthwise_conv2d` / `separable_conv2d`, and TF CPU rejects that with `Depthwise convolution on CPU is only supported for NHWC format`. Only the 1D depthwise path and this PR's new decomposition transpose to NHWC first, so those are the channels_first cases that pass.
> 
> So channels_first coverage on these ops looks like a separate pre-existing gap from the stride+dilation fix here. Happy to either (1) keep channels_first in this PR only on the paths that support it on CPU (1D depthwise + the new stride+dilation cases)

Yes, let's do that. But can you please still have a single test with all the combination and then add back something like:
```python
        if (
            backend.backend() == "tensorflow"
            and data_format == "channels_first"
            and not testing.tensorflow_uses_gpu()
            and strides == ...
```

>, or (2) add the channels_first->NHWC transpose to the regular depthwise/separable paths too, though that feels like its own change. Which would you prefer?

Yes, let's make it a separate change.

Thanks!!

Lawson-Darrow:
Done, all addressed:

- Folded `data_format=("channels_first", "channels_last")` into the existing `test_depthwise_conv_2d` / `test_separable_conv_2d` instead of separate methods.
- Renamed the 1D tests to `test_depthwise_conv_1d` / `test_separable_conv_1d` and parameterized `strides=(1, 2)` / `dilation_rate=(1, 2)`.
- Removed the PR-only comments.
- Added the skip guard you sketched: the channels_first cases TF CPU can't run (everything off the transposing paths) are skipped, while jax/numpy and TF GPU still run all combos.

The regular-path channels_first->NHWC transpose I'll handle as a separate change.

hertschuh (approved):
Thanks for working on this!

Lawson-Darrow:
My pleasure! Love the project, look forward to contributing more!

PR #23029, opened 2026-06-03
Save seed in Attention layer get_config

MarcosAsh:
## Summary

`Attention` accepts a `seed` argument and uses it to seed the dropout `SeedGenerator`, but its `get_config` never emits it. A `from_config(get_config())` round-trip (e.g. `save` / `load_model`) silently resets `seed` to `None`, so a layer built with an explicit seed reloads with a different one and draws different dropout masks - the reloaded model is no longer reproducible.

Every other layer with a `seed` argument already saves it: `SimpleRNN`, `GRU`, `LSTM`, `MultiHeadAttention`, and `GroupedQueryAttention` - this layer is the only gap. `AdditiveAttention` subclasses `Attention` and was missing it for the same reason.

The fix stores `self.seed` and emits it from `get_config`, covering `AdditiveAttention` too through `super().get_config()`.

`run_layer_test` did not catch this because it compares the revived layer's re-serialized config against the original config, and since both omitted `seed` they matched while the behavior silently changed. This PR adds a round-trip test that constructs the layer with an explicit seed, reserializes, and asserts it is preserved.

## Contributor Agreement

### DIFF
--- a/keras/src/layers/attention/attention.py
+++ b/keras/src/layers/attention/attention.py
@@ -75,6 +75,7 @@ class Attention(Layer):
         self.use_scale = use_scale
         self.score_mode = score_mode
         self.dropout = dropout
+        self.seed = seed
         if self.dropout > 0:
             self.seed_generator = backend.random.SeedGenerator(seed=seed)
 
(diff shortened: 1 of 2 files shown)


hertschuh (approved):
Thanks for the fix!

PR #23030, opened 2026-06-03
Remove progress bar double averaging of metrics.

### LINKED ISSUE #21301: evaluate() shows incorrect metric values in progress bar vs returned results

hertschuh:
All metrics emitted to the progress bar in `model.fit` and `mode.evaluate` are stateful. For instance, average metrics maintain the running sum and denominator and output the average.

Therefore, the progress bar needs to consider all of the metrics stateless. To make this easier without having to constantly update the list of stateful metrics (like Keras 2 does), added a backwards compatible parameter `stateless_metrics` to list stateless metrics instead.

Fixes https://github.com/keras-team/keras/issues/21301

## Contributor Agreement

### DIFF
--- a/keras/src/callbacks/progbar_logger.py
+++ b/keras/src/callbacks/progbar_logger.py
@@ -82,7 +82,10 @@ class ProgbarLogger(Callback):
     def _maybe_init_progbar(self):
         if self.progbar is None:
             self.progbar = Progbar(
-                target=self.target, verbose=self.verbose, unit_name="step"
+                target=self.target,
+                verbose=self.verbose,
+                unit_name="step",
+                stateless_metrics=set(),
             )
 
     def _update_progbar(self, batch, logs=None):
--- a/keras/src/utils/progbar.py
+++ b/keras/src/utils/progbar.py
@@ -13,13 +13,21 @@ from keras.src.utils import io_utils
 class Progbar:
     """Displays a progress bar.
 
+    If neither of `stateful_metrics` and `stateless_metrics` is provided, all
+    metrics are assumed to be stateless (averaged by the progress bar before
+    display).
+
     Args:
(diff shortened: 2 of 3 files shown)


laxmareddyp (approved):
Thanks.

PR #23036, opened 2026-06-04
[Fix] ValueError when loading SAMImageSegmenter from preset sam_huge_sa1b in Keras Hub

### LINKED ISSUE #23018: ValueError when loading SAMImageSegmenter from preset sam_huge_sa1b in Keras Hub
**Description:**
I encountered a ValueError while running the official Keras Hub tutorial notebook for Segment Anything. The model fails to load weights for multiple layers (such as Dense, Embedding, etc.).

**Code to Reproduce:**
https://colab.research.google.com/github/keras-team/keras-io/blob/master/guides/ipynb/keras_hub/segment_anything_in_keras_hub.ipynb#scrollTo=lt6d9sU9b2PD&uniqifier=2

model = keras_hub.models.SAMImageSegmenter.from_preset("sam_huge_sa1b")

**Error Traceback:**
```
Downloading to /root/.cache/kagglehub/models/keras/sam/keras/sam_huge_sa1b/5/config.json...
100%|██████████| 3.06k/3.06k [00:00<00:00, 3.47MB/s]
Downloading to /root/.cache/kagglehub/models/keras/sam/keras/sam_huge_sa1b/5/task.json...
100%|██████████| 5.77k/5.77k [00:00<00:00, 4.14MB/s]
Downloading to /root/.cache/kagglehub/models/keras/sam/keras/sam_huge_sa1b/5/task.weights.h5...
100%|██████████| 2.39G/2.39G [00:56<00:00, 45.6MB/s]
ValueError                                Traceback (most recent call last)
/tmp/ipykernel_546/708357786.py in <cell line: 0>()
----> 1 model = keras_hub.models.SAMImageSegmenter.from_preset("sam_huge_sa1b")

5 frames
/usr/local/lib/python3.12/dist-packages/keras/src/saving/saving_lib.py in _raise_loading_failure(error_msgs, warn_only)
    640         warnings.warn(msg)
    641     else:
--> 642         raise ValueError(msg)
    643 
    644 

ValueError: A total of 61 objects could not be loaded. Example error message for object <Dense name=dense_172, built=True>:

Layer 'dense_172' expected 2 variables, but received 0 variables during loading. Expected: ['kernel', 'bias']

List of objects that could not be loaded:
[<Dense name=dense_172, built=True>, <Dense name=dense_173, built=True>, <Dense name=dense_174, built=True>, <Embedding name=embedding, built=True>, <Embedding name=embedding_1, built=True>, <Dense name=dense_160, built=True>, <Dense name=dense_161, built=True>, <Dense name=dense_162, built=True>, <Dense name=dense_163, built=True>, <Dense name=dense_164, built=True>, <Dense name=dense_165, built=True>, <Dense name=dense_166, built=True>, <Dense name=dense_167, built=True>, <Dense name=dense_168, built=True>, <Dense name=dense_169, built=True>, <Dense name=dense_170, built=True>, <Dense name=dense_171, built=True>, <Conv2DTranspose name=conv2d_transpose, built=True>, <LayerNormalization name=layer_normalization_77, built=True>, <Conv2DTranspose name=conv2d_transpose_1, built=True>, <Dense name=dense_157, built=True>, <Dense name=dense_159, built=True>, <Dense name=dense_156, built=True>, <Dense name=dense_158, built=True>, <LayerNormalization name=layer_normalization_76, built=True>, <Dense name=dense_139, built=True>, <Dense name=dense_141, built=True>, <Dense name=dense_138, built=True>, <Dense name=dense_140, built=True>, <Dense name=dense_133, built=True>, <Dense name=dense_135, built=True>, <Dense name=dense_132, built=True>, <Dense name=dense_134, built=True>, <LayerNormalization name=layer_normalization_68, built=True>, <LayerNormalization name=layer_normalization_69, built=True>, <LayerNormalization name=layer_normalization_70, built=True>, <LayerNormalization name...
``

```

ChiragSW:
## Description

Fixes: #23018 
I changed Keras so objects_to_skip expands recursively through the same saveable traversal used by saving/loading. Now when KerasHub passes the backbone in the skip list, Keras also skips backbone-owned children like SAM’s mask_decoder, even if _flatten_layers() misses them.

## Contributor Agreement

### DIFF
--- a/keras/src/models/functional.py
+++ b/keras/src/models/functional.py
@@ -99,7 +99,6 @@ class Functional(Function, Model):
     def __new__(cls, *args, **kwargs):
         return typing.cast(cls, super().__new__(cls))
 
-    @tracking.no_automatic_dependency_tracking
     def __init__(self, inputs, outputs, name=None, **kwargs):
         if isinstance(inputs, dict):
             for k, v in inputs.items():
@@ -133,21 +132,14 @@ class Functional(Function, Model):
         if not all(is_input_keras_tensor(t) for t in flat_inputs):
             inputs, outputs = clone_graph_nodes(inputs, outputs)
 
-        preexisting_layers = list(getattr(self, "_layers", []))
-
-        Function.__init__(self, inputs, outputs, name=name)
+        with tracking.DotNotTrackScope():
+            Function.__init__(self, inputs, outputs, name=name)
 
         if trainable is not None:
             self.trainable = trainable
 
-        graph_layers = self.layers
-        graph_layer_ids = {id(layer) for layer in graph_layers}
-        extra_layers = [
-            layer
-            for layer in preexisting_layers
-            if id(layer) not in graph_layer_ids
-        ]
-        self._layers = graph_layers + extra_layers
+        for layer in self.layers:
+            self._tracker.track(layer)
         self.build(None)
         # We will convert directly (to the correct dtype per input).
         self._convert_input_args = False
--- a/keras/src/models/functional.py
+++ b/keras/src/models/functional.py
@@ -133,12 +133,21 @@ class Functional(Function, Model):
         if not all(is_input_keras_tensor(t) for t in flat_inputs):
             inputs, outputs = clone_graph_nodes(inputs, outputs)
 
+        preexisting_layers = list(getattr(self, "_layers", []))
+
         Function.__init__(self, inputs, outputs, name=name)
 
         if trainable is not None:
             self.trainable = trainable
 
-        self._layers = self.layers
+        graph_layers = self.layers
+        graph_layer_ids = {id(layer) for layer in graph_layers}
+        extra_layers = [
+            layer
+            for layer in preexisting_layers
+            if id(layer) not in graph_layer_ids
+        ]
+        self._layers = graph_layers + extra_layers
         self.build(None)
         # We will convert directly (to the correct dtype per input).
         self._convert_input_args = False
--- a/keras/src/models/functional.py
+++ b/keras/src/models/functional.py
@@ -710,9 +710,12 @@ def operation_fn(operation, **call_context_args):
 
 
 def functional_like_constructor(cls):
-    init_args = inspect.getfullargspec(cls.__init__).args[1:]
+    init_spec = inspect.getfullargspec(cls.__init__)
+    init_args = init_spec.args[1:]
     functional_init_args = inspect.getfullargspec(Functional.__init__).args[1:]
     if init_args == functional_init_args:
+        return True
+    if init_spec.varargs is not None and init_spec.varkw is not None:
         return True
     return False
 
--- a/keras/src/saving/saving_api_test.py
+++ b/keras/src/saving/saving_api_test.py
@@ -9,6 +9,7 @@ from absl.testing import parameterized
 
 from keras.src import layers
 from keras.src.legacy.saving.legacy_h5_format import save_model_to_hdf5
+from keras.src.models import Model
 from keras.src.models import Sequential
 from keras.src.saving import saving_api
 from keras.src.testing import test_case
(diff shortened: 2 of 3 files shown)


hertschuh:
> even if _flatten_layers() misses them.

This changes the contract of `objects_to_skip`, which was not previously recursive (although KerasHub is probably the only user of this feature).

Do you understand why `_flatten_layers()` misses them? Why only for this model? Is that what should be fixed instead?

ChiragSW:
Yea actually, `_flatten_layers()` misses them because `SAMBackbone` is Functional and it's graph doesn't include `mask_decoder`, the decoder is an attribute serialized by the backbone but called later by `SAMImageSegmenter`. So it is in backbone structurally but unreachable from the backbone Functional graph.

That makes SAM unusual. While making changes I thought doing a  `objects_to_skip` contract fix would help but making it broad made it complicated too. I think it will be better if the model fix is done like you asked.

ChiragSW:
`objects_to_skip` is reverted and the fix is now why `_flatten_layers()` missed the layer, Functional subclasses were replacing `_layers` with graph layers only and dropping pre-tracked attribute layers outside Functional graph.

keras/src/models/functional.py
```diff
@@ -133,12 +133,21 @@ def __init__(self, inputs, outputs, name=None, **kwargs):
...
 
-        self._layers = self.layers
+        graph_layers = self.layers
+        graph_layer_ids = {id(layer) for layer in graph_layers}
+        extra_layers = [
+            layer
+            for layer in preexisting_layers
+            if id(layer) not in graph_layer_ids
+        ]
+        self._layers = graph_layers + extra_layers
```
hertschuh: I agree with the analysis. Thanks for figuring this out!

However, I prefer this approach for the fix instead of what you have. It's simpler and it uses the tracker instead of bypassing it (the tracker will dedupe).

```python
# Track layers from the graph.
for layer in self.layers:
    self._tracker.track(layer)
```

keras/src/saving/saving_api_test.py
```diff
@@ -116,6 +117,47 @@ def test_objects_to_skip(self):
             model.layers[0].get_weights()[1],
         )
 
+    def test_objects_to_skip_with_functional_subclass_attribute_layer(self):
+        class Backbone(Model):
```
hertschuh, line 121: You don't need to extend `Functional`?

keras/src/saving/saving_api_test.py
```diff
@@ -116,6 +117,47 @@ def test_objects_to_skip(self):
...
+            def __init__(self):
+                # Simulate a layer owned by a Functional subclass, but not
+                # part of the Functional graph. This matches e.g. a decoder
+                # attached to a backbone but called by a task model.
+                decoder = layers.Dense(2, name="decoder")
+                self._initialize_tracker()
+                self._tracker.track(decoder)
+                object.__setattr__(self, "decoder", decoder)
```
hertschuh: Why not just do this?
```python
self.decoder = decoder
```

hertschuh (changes requested):
Thanks for the update!

keras/src/saving/saving_api_test.py
ChiragSW: No need to extend Functional. I used Model subclass initialized with Functional args to mirror the SAM pattern but I can simplify the fixture if there’s a clearer version.

keras/src/saving/saving_api_test.py
ChiragSW: I used manual tracker assignment only to avoid the torch pre-`Module.__init__()` assignment error in an earlier version of the test. I can use self.decoder = decoder (thank you for the suggestion)

keras/src/models/functional.py
ChiragSW: will do, its cleaner

keras/src/saving/saving_api_test.py
```diff
@@ -116,6 +117,45 @@ def test_objects_to_skip(self):
...
 
+    def test_objects_to_skip_with_functional_subclass_attribute_layer(self):
+        class Backbone(Model):
+            def __init__(self):
+                # Simulate a layer owned by a Functional subclass, but not
+                # part of the Functional graph. This matches e.g. a decoder
+                # attached to a backbone but called by a task model.
+                decoder = layers.Dense(2, name="decoder")
+                self.decoder = decoder
+
```
hertschuh: It is indeed failing with Torch. Move this line after line 132 (after super().__init__)

keras/src/models/functional.py
```diff
@@ -133,12 +132,14 @@ def __init__(self, inputs, outputs, name=None, **kwargs):
         if not all(is_input_keras_tensor(t) for t in flat_inputs):
             inputs, outputs = clone_graph_nodes(inputs, outputs)
 
-        Function.__init__(self, inputs, outputs, name=name)
+        with tracking.DotNotTrackScope():
+            Function.__init__(self, inputs, outputs, name=name)
```
hertschuh, line 136: Why did you need the `DotNotTrackScope`?

hertschuh (changes requested):

keras/src/models/functional.py
hertschuh: Oh, I understand now, ignore.

keras/src/saving/saving_api_test.py
ChiragSW: Moved it

keras/src/models/functional.py
```diff
@@ -709,10 +710,13 @@ def call(*args, **kwargs):
...
 
 def functional_like_constructor(cls):
-    init_args = inspect.getfullargspec(cls.__init__).args[1:]
+    init_spec = inspect.getfullargspec(cls.__init__)
+    init_args = init_spec.args[1:]
     functional_init_args = inspect.getfullargspec(Functional.__init__).args[1:]
     if init_args == functional_init_args:
         return True
+    if init_spec.varargs is not None and init_spec.varkw is not None:
+        return True
```
hertschuh, line 719: Why did you need this? This seems dangerous.

So if you have an `__init__(self, *args, **kwargs)` you assume it's functional?

ChiragSW: Before `Functional.__init__` was wrapped so `inspect.getfullargspec(Functional.__init__)` exposed the forwarding `*args, **kwargs` . That meant a custom Functional subclass constructor like:
```
def __init__(self, *args, param=1, **kwargs):
    self.param = param
```
was treated as Functional-like and could be revived from config. After removing the decorator, `Functional.__init__` exposed its real `(inputs, outputs, name=None, **kwargs)` signature and that existing test started failing.

I did not intend to broaden the test but to preserve the previous behavior for forwarding constructors.
I will be happy to revise if there’s a preferred way to preserve the existing revive behavior without making the constructor test broader

hertschuh: Oh weird, I'm surprised we never hit this bug.

hertschuh (approved):
Thanks for the fix!

PR #23040, opened 2026-06-04
Apply causal mask in MHA/GQA manual attention path

### LINKED ISSUE #23038: MultiHeadAttention silently drops the causal mask when `use_causal_mask=True` and the fused SDPA path is disabled (dropout > 0, `return_attention_scores=True`, or non-4D inputs) — regression in #22910
# MultiHeadAttention silently drops the causal mask when `use_causal_mask=True` and the fused SDPA path is disabled (dropout > 0, `return_attention_scores=True`, or non-4D inputs) — regression in #22910

## Summary

Since #22910 ("Engage flash SDPA in MHA/GQA causal-only path"), `MultiHeadAttention`
(and `GroupedQueryAttention`) **silently apply no causal mask at all** whenever the
fused scaled-dot-product-attention path is not taken. Every query attends to every
key, including strictly-future positions. This is silent future-information leakage in
any causal transformer — there is no error, just wrong numbers (e.g. metrics like AUC
inflating toward ~1.0 because future timesteps become trivially predictable).

The bug triggers in three common, independent configurations of an MHA layer called
with `use_causal_mask=True` and no other mask source:

- `dropout > 0.0`
- `return_attention_scores=True`
- non-4D query/key/value inputs (`len(query.shape) != 4`)

Any one of these is sufficient. In particular, dropout=0 is **not** safe — a layer with
no dropout still leaks the moment it is called with `return_attention_scores=True`.

## Regression range

Bisected to a single commit (both endpoints verified by running the repro below):

- **First bad commit:** `5a524001dbc89cc9be4e034da0ee7c5f8064aafd` — PR #22910,
  "Engage flash SDPA in MHA/GQA causal-only path", 2026-05-27.
- **Last good commit:** `c42b5ae2476bcd7d1aa75c097d11fa879c518890` (the parent,
  `5a524001~1`) — PR #22992, 2026-05-27. Clean: future-key attention is `0.0`.
- **Releases affected:** none yet. Latest tag is `v3.14.0` (early April 2026), before the
  regression. This currently exists only on `master` / `keras-nightly`, but will ship in
  the next release if not fixed.

## Root cause

In `keras/src/layers/attention/multi_head_attention.py`, `call()` computes a
`causal_only` fast path and, when it applies, sets `attention_mask = None` and forwards
`use_causal_mask=True` into `_compute_attention`, expecting the backend's native causal
kernel to enforce masking:

```python
causal_only = (
    use_causal_mask
    and query_mask is None
    and value_mask is None
    and key_mask is None
    and attention_mask is None
    and self._sliding_window is None
    and type(self)._compute_attention is MultiHeadAttention._compute_attention
)
if causal_only:
    attention_mask = None   # <-- mask dropped here, on the assumption SDPA(is_causal=True) runs
```

But inside `_compute_attention`, the fused causal kernel only runs when
`use_dot_product_attention` is true:

```python
use_dot_product_attention = not (
    self._dropout > 0.0
    or return_attention_scores
    or (len(query.shape) != 4)
)

if use_dot_product_attention:
    if use_causal_mask and attention_mask is None:
        # fused path: ops.dot_product_attention(..., is_causal=True)  ✓ masks correctly
        ...
        return attention_output, None
    ...
# ---- fallthrough: manual attention ----
attention_scores = ops.einsum(self._dot_product_equation, key, query)
attention_scores = self._masked_softmax(attention_scores, attention_mask)  # attention_mask is None → NO MASK
```

The `causal_only` gate in `call()` checks the mask sources, sliding window, and subclass
override — but it never checks `dropout`, `return_attention_scores`, or input rank. When
any of those disables `use_dot_product_attention`, execution falls through to the manual
softmax path with `attention_mask=None`, so `_masked_softmax` applies no mask at all and
`is_causal=True` is never honored.

## Minimal reproduction

The physical invariant: under causal masking, attention onto strictly-future keys must
be exactly `0.0`.

```python
import os
os.environ["KERAS_BACKEND"] = "jax"  # reproduces on all backends
import numpy as np
import keras
from keras import layers

T = 6
future = np.triu(np.ones((T, T)), k=1).astype(bool)  # strictly-future key positions
x = np.random.RandomState(0).normal(size=(1, T, 8)).astype("float32")

# return_attention_scores=True is itself a trigger (dropout is 0 here):
mha = layers.MultiHeadAttention(num_heads=2, key_dim=4)
_, scores = mha(x, x, use_causal_mask=True, return_attention_scores=True)

leak = np.asarray(scores)[..., future].sum()
print("attention on future keys:", leak)   # must be 0.0; prints ~7.6 on master since #22910
assert leak < 1e-6, "causal mask was silently dropped"
```

### Output-equivalence check (covers the dropout>0, scores=False case)

When scores aren't returned, compare `use_causal_mask=True` against an explicit
lower-triangular `attention_mask` — they must be identical at inference:

```python
mha = layers.MultiHeadAttention(num_heads=2, key_dim=4, dropout=0.1)
out_causal   = np.asarray(mha(x, x, use_causal_mask=True, training=False))
explicit     = (~future)[None]  # (1, T, T) allowed-positions mask
out_explicit = np.asarray(mha(x, x, attention_mask=explicit, training=False))
print("max|causal - explicit|:", np.abs(out_causal - out_explicit).max())  # must be ~0
```

## Observed vs. expected

Verified by running the repro from source at each commit (jax backend; same on tf/torch):

| Case (`use_causal_mask=True`)                         | Good `c42b5ae24` | Bad `5a524001` |
|-------------------------------------------------------|------------------|----------------|
| dropout=0,   `return_attention_scores=True` (future-key attention) | `0.000000` | `7.613400` |
| dropout=0.1, `return_attention_scores=True` (future-key attention) | `0.000000` | `7.613400` |
| dropout=0.1, scores=False (`\|causal − explicit-mask\|` of output) | `0.000000` | `3.776578` |

**Expected:** all values `0.0` (no attention on future keys; `use_causal_mask=True`
identical to an explicit causal `attention_mask`).
**Actual on master:** future positions receive full attention weight; outputs diverge
from the explicit-mask reference.

## How it slipped through

PR #22910 added tests strengthening `extra_mask` coverage for the causal-only fallback,
but none exercise the fallback with `dropout > 0` or `return_attention_scores=True` — the
exact conditions that disable the fused kernel and expose the dropped mask.

## Workarounds (until fixed)

- On an affected build, pass an explicit causal mask via `attention_mask=...` instead of
  `use_causal_mask=True` — that path is unaffected.

## Environment

- Keras `master`; first bad commit `5a524001` (PR #22910), last good commit
  `c42b5ae24` (its parent).
- Backend: JAX (CPU). The defect is in backend-agnostic layer logic and reproduces on
  TensorFlow and PyTorch as well.
- Python 3.11.

MarcosAsh:
## Description

Regression from #22910: `use_causal_mask=True` silently applies no mask when the fused SDPA path is off (dropout>0, return_attention_scores=True, or non-4D), so queries attend to future keys. GQA has the same bug. Fixes #23038.

Fix materializes the causal mask in the manual attention fallthrough. Tests cover all three triggers on both layers.

## Contributor Agreement

### DIFF
--- a/keras/src/layers/attention/grouped_query_attention.py
+++ b/keras/src/layers/attention/grouped_query_attention.py
@@ -491,16 +491,6 @@ class GroupedQueryAttention(Layer):
                     flash_attention=self._flash_attention,
                 )
                 return attention_output, None
-            if use_causal_mask and attention_mask is not None:
-                # The native is_causal kernel above was skipped because an
-                # explicit mask was also provided. Fold the causal mask into it
-                # so both constraints apply. Backends disagree on whether
-                # is_causal composes with an explicit mask, so combine them here
-                # rather than forwarding the is_causal flag.
-                attention_mask = ops.logical_and(
-                    ops.cast(attention_mask, "bool"),
-                    self._compute_causal_mask(query, value),
-                )
             if attention_mask is not None:
                 # Ensure attention_mask has the correct shape for broadcasting
                 # Expected shape: [batch_size, num_heads, query_seq_len,
@@ -522,7 +512,7 @@ class GroupedQueryAttention(Layer):
                 bias=None,
                 mask=attention_mask,
                 scale=self._inverse_sqrt_head_dim,
-                is_causal=False,
+                is_causal=use_causal_mask,
                 flash_attention=self._flash_attention,
             )
             return attention_output, None
--- a/keras/src/layers/attention/grouped_query_attention.py
+++ b/keras/src/layers/attention/grouped_query_attention.py
@@ -491,6 +491,16 @@ class GroupedQueryAttention(Layer):
                     flash_attention=self._flash_attention,
                 )
                 return attention_output, None
+            if use_causal_mask and attention_mask is not None:
+                # The native is_causal kernel above was skipped because an
+                # explicit mask was also provided. Fold the causal mask into it
+                # so both constraints apply. Backends disagree on whether
+                # is_causal composes with an explicit mask, so combine them here
+                # rather than forwarding the is_causal flag.
+                attention_mask = ops.logical_and(
+                    ops.cast(attention_mask, "bool"),
+                    self._compute_causal_mask(query, value),
+                )
             if attention_mask is not None:
                 # Ensure attention_mask has the correct shape for broadcasting
                 # Expected shape: [batch_size, num_heads, query_seq_len,
@@ -521,9 +531,17 @@ class GroupedQueryAttention(Layer):
         # scores. We skipped the fused is_causal kernel above, so build the
         # causal mask here. `call` drops attention_mask when causal is the
         # only mask source, expecting that kernel to run, and this path never
-        # sees the is_causal flag otherwise.
-        if use_causal_mask and attention_mask is None:
-            attention_mask = self._compute_causal_mask(query, value)
+        # sees the is_causal flag otherwise. When an explicit mask is also
+        # present, combine the two so neither constraint is lost.
+        if use_causal_mask:
+            causal_mask = self._compute_causal_mask(query, value)
+            attention_mask = (
+                causal_mask
+                if attention_mask is None
+                else ops.logical_and(
+                    ops.cast(attention_mask, "bool"), causal_mask
+                )
+            )
         query = ops.multiply(
             query, ops.cast(self._inverse_sqrt_head_dim, query.dtype)
         )
--- a/keras/src/layers/attention/grouped_query_attention.py
+++ b/keras/src/layers/attention/grouped_query_attention.py
@@ -518,7 +518,12 @@ class GroupedQueryAttention(Layer):
             return attention_output, None
 
         # Default behavior without flash attention, with explicit attention
-        # scores
+        # scores. We skipped the fused is_causal kernel above, so build the
+        # causal mask here. `call` drops attention_mask when causal is the
+        # only mask source, expecting that kernel to run, and this path never
+        # sees the is_causal flag otherwise.
+        if use_causal_mask and attention_mask is None:
+            attention_mask = self._compute_causal_mask(query, value)
         query = ops.multiply(
             query, ops.cast(self._inverse_sqrt_head_dim, query.dtype)
         )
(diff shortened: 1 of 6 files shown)


keras/src/layers/attention/grouped_query_attention.py
hertschuh: Shouldn't this be `is_causal=use_causal_mask`?

If `use_causal_mask and attention_mask is None` it returns earlier, that's fine.

But if `use_causal_mask` is `True` and `attention_mask` is not None, it won't apply the causal mask.

keras/src/layers/attention/multi_head_attention.py
hertschuh: Same comment about changing this to `is_causal=use_causal_mask`.

hertschuh (commented):
Your change look good, but I think there is still a similar bug.

MarcosAsh:
Thanks for the review. You're right that this path drops the causal mask when an explicit `attention_mask` is also passed.

I tried `is_causal=use_causal_mask` first, but it only works on JAX and TF. The torch backend forces `is_causal=False` whenever a mask is present, since PyTorch's SDPA treats `attn_mask` and `is_causal` as mutually exclusive, so forwarding the flag passes on the default backends but silently leaks future keys on torch.

To stay correct everywhere I fold the causal mask into the explicit one instead:

```python
if use_causal_mask and attention_mask is not None:
    attention_mask = ops.logical_and(
        ops.cast(attention_mask, "bool"),
        self._compute_causal_mask(query, value),
    )
```

That also matches the manual softmax path, which already builds the mask explicitly. Did the same in both layers with a test covering both paths. Happy to push the fix into the torch backend instead if you'd prefer.

hertschuh:
> That also matches the manual softmax path, which already builds the mask explicitly. Did the same in both layers with a test covering both paths. Happy to push the fix into the torch backend instead if you'd prefer.

Well that means there is a bug in `dot_product_attention` on Torch then, whereby the `is_causal` flag is not really honored correctly.

Please move the fix to the Torch backend. Thanks a lot!

MarcosAsh:
Hi, just made the changes!

MarcosAsh:
The openvino test failed I just fixed it!

hertschuh (approved):
Thanks a lot!

PR #23041, opened 2026-06-04
Transpose channels_first to NHWC on CPU for depthwise/separable conv

### LINKED ISSUE #23031: TF backend: channels_first depthwise/separable conv fails on CPU (regular paths pass NCHW to tf.nn)
On the TensorFlow backend (CPU), `keras.ops.depthwise_conv` and `keras.ops.separable_conv` raise `UnimplementedError: Depthwise convolution on CPU is only supported for NHWC format` whenever `data_format="channels_first"`, except on the two paths that transpose to NHWC first: the 1D depthwise path and the `strides>1 & dilation>1` decomposition added in #23027.

Repro (TF backend, CPU; non-zero data so the op isn't constant-folded away):

```python
import numpy as np
from keras import ops

x = np.arange(600, dtype=float).reshape(2, 3, 10, 10)   # channels_first
dk = np.arange(24, dtype=float).reshape(2, 2, 3, 2)
ops.depthwise_conv(x, dk, strides=1, padding="valid", data_format="channels_first")
# UnimplementedError: Depthwise convolution on CPU is only supported for NHWC format
```

`separable_conv` fails the same way.

Root cause: the regular 2D depthwise path and the 1D/2D separable paths pass `tf_data_format="NCHW"` straight to `tf.nn.depthwise_conv2d` / `tf.nn.separable_conv2d` (`keras/src/backend/tensorflow/nn.py`, ~L1021 and ~L1082/1097). The 1D depthwise path (~L979-986) and the new decomposition already transpose `channels_first`->NHWC on CPU; the others don't.

Suggested fix: extend that same CPU transpose pattern to the regular 2D depthwise and 1D/2D separable paths. Once fixed, the `channels_first` cases that #23027 currently skips on TF CPU can be unskipped.

Split out from #23027 per discussion with @hertschuh. I'm happy to open a PR for this with tests.

Lawson-Darrow:
## Description

Fixes #23031. On the TensorFlow backend, channels_first `depthwise_conv` (2D) and `separable_conv` (1D and 2D) failed on CPU with "Depthwise convolution on CPU is only supported for NHWC format", because these paths passed NCHW straight to the `tf.nn` ops. The 1D depthwise path and the stride+dilation decomposition from #23027 already transpose channels_first to NHWC on CPU. This applies the same pattern to the three remaining paths. It transposes to channels_last on CPU, computes, transposes back, and keeps NCHW on GPU where it is supported.

It also removes the channels_first skip guards #23027 added to the conv tests, since those cases now pass. Verified on the TF, JAX, and NumPy backends (the previously skipped cases now run and pass) plus the depthwise and separable layer tests on TF.

Prepared with AI assistance from Claude Code and OpenAI Codex. I reviewed and verified the change, ran the conv op and layer tests locally across backends, and take responsibility for it.

## Contributor Agreement

### DIFF
--- a/keras/src/backend/tensorflow/nn.py
+++ b/keras/src/backend/tensorflow/nn.py
@@ -1,3 +1,4 @@
+import functools
 import math
 import warnings
 
@@ -16,6 +17,17 @@ from keras.src.backend.common.backend_utils import (
 )
 from keras.src.backend.tensorflow.core import cast
 from keras.src.backend.tensorflow.core import convert_to_tensor
+
+
+@functools.lru_cache(maxsize=None)
+def _cpu_only():
+    """Whether the TF runtime sees only CPU devices.
+
+    Cached because the device list does not change during a run and
+    `tf.config.list_logical_devices()` is comparatively expensive to call on
+    every conv invocation in eager mode.
+    """
+    return all(d.device_type == "CPU" for d in tf.config.list_logical_devices())
 
 
 def relu(x):
@@ -979,9 +991,7 @@ def depthwise_conv(
         # `tf.nn.depthwise_conv2d` does not support `channels_first` with
         # dilations on CPU. Transpose to `channels_last`, compute, and
         # transpose back to avoid the limitation.
-        need_transpose = data_format == "channels_first" and all(
-            d.device_type == "CPU" for d in tf.config.list_logical_devices()
-        )
+        need_transpose = data_format == "channels_first" and _cpu_only()
         if need_transpose:
             inputs = _transpose_spatial_inputs(inputs)
         if need_transpose or data_format == "channels_last":
@@ -1016,9 +1026,7 @@ def depthwise_conv(
     # Transpose to `channels_last`, compute, and transpose back to avoid the
     # limitation. On GPU, `channels_first` (NCHW) is supported natively, so we
     # keep it to avoid unnecessary copies.
-    need_transpose = data_format == "channels_first" and all(
-        d.device_type == "CPU" for d in tf.config.list_logical_devices()
-    )
+    need_transpose = data_format == "channels_first" and _cpu_only()
     if need_transpose:
         inputs = _transpose_spatial_inputs(inputs)
     if need_transpose or data_format == "channels_last":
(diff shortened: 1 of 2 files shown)


Lawson-Darrow:
Good call on the device check. I moved it into a cached helper (`_cpu_only`, `functools.lru_cache`) so `tf.config.list_logical_devices()` runs once instead of on every call, and pointed all four sites at it, including the existing 1D depthwise path.

On the codecov note, the remaining uncovered lines are the GPU branches that keep NCHW. They only run when a GPU is present, so the CPU test matrix cannot exercise them. This is the same as the channels_first decomposition path in #23027.

keras/src/backend/tensorflow/nn.py
hertschuh, line 854: Oh, I just thought about this.

Can you first check if a hack like this one would work for `separable_conv` and `depthwise_conv`? It might be faster.

hertschuh (commented):
Thanks for your work on this!

keras/src/backend/tensorflow/nn.py
Lawson-Darrow: Okay so I looked into it and wrapping the NCHW depthwise/separable op in `@tf.function(jit_compile=True)` does work on CPU and matches the transpose path within float32 rounding, but for this standalone backend op it was several times slower than the transpose approach.

For input 16x32x64x64 with a 3x3 kernel, steady state after warmup/compile:

- depthwise: transpose 1.2 ms vs XLA 7.4 ms
- separable: transpose 1.6 ms vs XLA 7.8 ms

So the native NHWC kernels plus two transposes are faster here than XLA's CPU NCHW path. 

This doesn't rule out XLA helping inside a larger jit-compiled graph though. I think for this backend fallback the transpose path is the better default but I'd love to hear your thoughts!

hertschuh: > This doesn't rule out XLA helping inside a larger jit-compiled graph though. I think for this backend fallback the transpose path is the better default but I'd love to hear your thoughts!

That was exactly my thought. The "eager" (non-compiled) path is typically not a real use case, it's just for testing. Once you're ready to train, you use `fit`, which is supposed to be compiled.

This is until I remembered that on TensorFlow, we turn off jit compilation on CPU by default because it turned out to be slower in most cases...

hertschuh (approved):
Thanks for the fix!

Lawson-Darrow:
Happy to help!

PR #23044, opened 2026-06-05
[Distribution] Implement sharding support for Keras PyDataset adapter

buildwithsuhana:
## Description
This PR implements native sharding support for keras.utils.PyDataset (and its underlying enqueuer), enabling parallel Python-based data pipelines in multi-process distributed training.

   * `PyDatasetAdapter` Sharding: Implements logic to shard indices across model replicas. Each process now only iterates over its designated data shard, significantly reducing redundant I/O and preprocessing in a distributed cluster.
   
[4th PR]: This PR should be rebased after merging of PR https://github.com/keras-team/keras/pull/23043

## Contributor Agreement

**Please review our [AI-Assisted Contribution Policy](../CONTRIBUTING.md#ai-assisted-contribution-policy) and check all boxes below before submitting your PR for review:**

- [x] I am a human, and not a bot.
- [x] I will be responsible for responding to review comments in a timely manner.
- [x] I will work with the maintainers to push this PR forward until submission.

**Note:** Failing to adhere to this agreement may result in your future PRs no longer being reviewed.

PR #23045, opened 2026-06-05
[Distribution] Implement multi-process sharding support for Torch DataLoaders

buildwithsuhana:
## Description
This PR completes the distributed data infrastructure by integrating Torch `DataLoader` into the Keras centralized sharding routing. It provides a seamless way to use native PyTorch data pipelines in multi-process distributed training without manual boilerplate.

  Key Changes:
   * `distribute_torch_dataloader`: Implements the final routing path in the Distribution base class, allowing Keras to automatically detect and shard Torch DataLoaders.
   * Automatic Sampler Injection: Introduces _add_torch_distributed_sampler, which inspects a user’s existing DataLoader and automatically injects a DistributedSampler or DistributedBatchSampler. This preserves user settings like batch_size, num_workers, collate_fn, and persistent_workers.
   * `ShardedIterableDataset`: A new wrapper for Torch IterableDataset instances. Since standard Torch DistributedSampler only supports map-style datasets, this provides the necessary interleaved sharding logic to support streaming data formats.

[5th PR]: This PR should be rebased after merging of PR https://github.com/keras-team/keras/pull/23044

## Contributor Agreement

### DIFF
--- a/keras/src/trainers/data_adapters/torch_data_loader_adapter.py
+++ b/keras/src/trainers/data_adapters/torch_data_loader_adapter.py
@@ -24,23 +24,10 @@ class TorchDataLoaderAdapter(DataAdapter):
         self._num_data_shards = 1
         self._data_shard_id = 0
         if dist is not None and getattr(dist, "auto_shard_dataset", False):
-            self._num_data_shards = min(
-                dist.num_model_replicas, dist.num_processes
-            )
+            self._num_data_shards = dist.num_data_shards
             self._data_shard_id = dist.data_shard_id
 
-        if (
-            dist is not None
-            and dist._is_multi_process
-            and getattr(dist, "auto_shard_dataset", False)
-        ):
-            # Detect shuffle from dataloader
-            shuffle = False
-            if hasattr(dataloader, "sampler") and isinstance(
-                dataloader.sampler, torch.utils.data.RandomSampler
-            ):
-                shuffle = True
-
+        if self._num_data_shards > 1:
             dataset = dataloader.dataset
             if isinstance(dataset, torch.utils.data.IterableDataset):
 
@@ -73,14 +60,24 @@ class TorchDataLoaderAdapter(DataAdapter):
                     ),
                     batch_size=dataloader.batch_size,
                     num_workers=dataloader.num_workers,
+                    collate_fn=dataloader.collate_fn,
                     pin_memory=dataloader.pin_memory,
                     drop_last=dataloader.drop_last,
                     timeout=dataloader.timeout,
                     worker_init_fn=dataloader.worker_init_fn,
+                    multiprocessing_context=dataloader.multiprocessing_context,
+                    generator=dataloader.generator,
                     prefetch_factor=dataloader.prefetch_factor,
                     persistent_workers=dataloader.persistent_workers,
                 )
             else:
+                # Detect shuffle from dataloader
+                shuffle = False
+                if hasattr(dataloader, "sampler") and isinstance(
+                    dataloader.sampler, torch.utils.data.RandomSampler
+                ):
+                    shuffle = True
+
                 sampler = torch.utils.data.distributed.DistributedSampler(
                     dataset,
                     num_replicas=self._num_data_shards,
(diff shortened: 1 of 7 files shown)


keras/src/trainers/data_adapters/torch_data_loader_adapter.py
```diff
@@ -19,18 +20,109 @@ def __init__(self, dataloader):
                 f"`torch.utils.data.DataLoader`. Received: {dataloader}"
             )
 
+        dist = distribution_lib.distribution()
+        self._num_data_shards = 1
+        self._data_shard_id = 0
+        if dist is not None and getattr(dist, "auto_shard_dataset", False):
+            self._num_data_shards = min(
+                dist.num_model_replicas, dist.num_processes
+            )
```
hertschuh: I didn't notice this in the previous CL, but I don't think this is correct, based on this line:
https://github.com/keras-team/keras/blob/master/keras/src/distribution/distribution_lib.py#L359

Specifically, if `num_model_replicas < num_processes`, we're using `processes_per_replica = self.num_processes // num_model_replicas` to compute the `data_shard_id`, so that's the number of data shards.

So two things:
- this needs to be fixed here and in `py_dataset_adapter.py`
- let's add a `dist.num_data_shards` property in `Distribution` to not replicate that logic.

keras/src/trainers/data_adapters/torch_data_loader_adapter.py
```diff
@@ -19,18 +20,109 @@ def __init__(self, dataloader):
...
+            self._data_shard_id = dist.data_shard_id
+
+        if (
+            dist is not None
+            and dist._is_multi_process
+            and getattr(dist, "auto_shard_dataset", False)
+        ):
```
hertschuh: I think this whole condition should be replaced with `if self._num_data_shards > 1`. Do you agree?

keras/src/trainers/data_adapters/torch_data_loader_adapter.py
```diff
@@ -19,18 +20,109 @@ def __init__(self, dataloader):
...
+            # Detect shuffle from dataloader
+            shuffle = False
+            if hasattr(dataloader, "sampler") and isinstance(
+                dataloader.sampler, torch.utils.data.RandomSampler
+            ):
+                shuffle = True
```
hertschuh: This is only used in the `else` case (map-style dataset), move it there to avoid confusion.

keras/src/trainers/data_adapters/torch_data_loader_adapter.py
```diff
@@ -19,18 +20,109 @@ def __init__(self, dataloader):
...
 
+    def on_epoch_begin(self):
+        if hasattr(self._dataloader, "sampler") and hasattr(
+            self._dataloader.sampler, "set_epoch"
+        ):
+            self._dataloader.sampler.set_epoch(self._epoch)
+
+    def on_epoch_end(self):
+        self._epoch += 1
+        self._output_signature = None
```
hertschuh: Why reset `_output_signature`? It's never supposed to change.

keras/src/trainers/data_adapters/torch_data_loader_adapter.py
```diff
@@ -19,18 +20,109 @@ def __init__(self, dataloader):
...
                 self._partial_batch_size = (
-                    len(dataloader.dataset) % self._batch_size
+                    len(self._dataloader.dataset) % self._batch_size
                 )
 
+    def on_epoch_begin(self):
+        if hasattr(self._dataloader, "sampler") and hasattr(
+            self._dataloader.sampler, "set_epoch"
+        ):
+            self._dataloader.sampler.set_epoch(self._epoch)
```
hertschuh, line 123: Is it only `torch.utils.data.distributed.DistributedSampler` that supports `set_epoch`?

If so, please add a comment that this is for `torch.utils.data.distributed.DistributedSampler`

keras/src/trainers/data_adapters/torch_data_loader_adapter.py
buildwithsuhana: Fixed

keras/src/trainers/data_adapters/torch_data_loader_adapter.py
buildwithsuhana: Yes, done

keras/src/trainers/data_adapters/torch_data_loader_adapter.py
buildwithsuhana: Done

keras/src/trainers/data_adapters/torch_data_loader_adapter.py
buildwithsuhana: Fixed

keras/src/trainers/data_adapters/torch_data_loader_adapter.py
buildwithsuhana: Added the comment

hertschuh (approved):
Thanks!

PR #23043, opened 2026-06-05
[Distribution] Implement centralized distribute_dataset

buildwithsuhana:
## Description
This PR refactors the distribute_dataset logic in the Distribution base class to establish a backend-agnostic routing infrastructure. It centralizes how Keras identifies and shards datasets in a multi-process environment, allowing for modular support of multiple data formats.

Key Changes:

   * Centralized Routing: Implemented a unified distribute_dataset method that serves as a central router. This PR includes the full implementation for TensorFlow (`distribute_tf_dataset`), while support for other formats (like Torch DataLoader) is architected to be added in subsequent PRs.
   * TF Dataset Refactor: Moves existing tf.data.Dataset distribution logic into a dedicated method. It introduces improved safety guards, such as checking for a static batch dimension to avoid IndexError and explicitly casting tensor batch sizes to Python integers.
   * Bug Fix (Error Messaging): Corrected a bug in the data adapter factory (get_data_adapter) where providing invalid arguments (like a separate y target) to a Python generator resulted in a confusing error message incorrectly citing the input as a "PyDataset". The error now correctly identifies the input as a "generator".

[2nd PR]: This PR should be rebased after merging of PR https://github.com/keras-team/keras/pull/23025
## Contributor Agreement

**Please review our [AI-Assisted Contribution Policy](../CONTRIBUTING.md#ai-assisted-contribution-policy) and check all boxes below before submitting your PR for review:**

- [x] I am a human, and not a bot.
- [x] I will be responsible for responding to review comments in a timely manner.
- [x] I will work with the maintainers to push this PR forward until submission.

**Note:** Failing to adhere to this agreement may result in your future PRs no longer being reviewed.

PR #23046, opened 2026-06-05
[Distribution] Implement native sharding support for Keras NumPy adapter

buildwithsuhana:
## Description
This PR introduces the core infrastructure for sharding Keras-native data formats, specifically focusing on NumPy and other array-like inputs handled by ArrayDataAdapter. It implements the logic required to slice global arrays into process-specific shards during distributed training.

Key Changes:

- DistributedBatchSampler: A new utility in data_adapter_utils.py that handles sharding at the batch level. It ensures that each process only sees a subset of the data while respecting drop_last settings and handling uneven batch distributions across replicas.

- ArrayDataAdapter Sharding: Implements sharding for get_numpy_iterator, get_jax_iterator, and get_torch_dataloader. The adapter now calculates the correct num_batches for the local shard, ensuring trainers perform the correct number of steps per epoch.

[3rd PR]: This PR will be rebased after merging of the PR https://github.com/keras-team/keras/pull/23043

## Contributor Agreement


keras/src/trainers/data_adapters/array_data_adapter.py
```diff
@@ -91,6 +92,27 @@ def __init__(
         self._batch_size = batch_size
         self._partial_batch_size = num_samples % batch_size
         self._shuffle = shuffle
+        self._epoch = 0
+        self._tf_epoch = None
```
hertschuh: I really don't think you need both. You can just convert `self._epoch` to a tensor when needed.

keras/src/trainers/data_adapters/array_data_adapter.py
```diff
@@ -122,6 +144,11 @@ def get_tf_dataset(self):
         # 3. pipelined permutation generation
         # 4. optimized permutation batching
         # 5. disabled static optimizations
+        if tf.available:
```
hertschuh: This block is all TF, remove the `if tf.available`.

keras/src/trainers/data_adapters/array_data_adapter.py
```diff
@@ -122,6 +144,11 @@ def get_tf_dataset(self):
...
+            if self._tf_epoch is None:
+                self._tf_epoch = tf.Variable(0, dtype=tf.int64, trainable=False)
+            self._tf_epoch.assign(self._epoch)
+        epoch = self._tf_epoch if self._tf_epoch is not None else self._epoch
```
hertschuh: Base on the removing `_tf_epoch` comment, just remove this code.

keras/src/trainers/data_adapters/array_data_adapter.py
```diff
@@ -91,6 +92,27 @@ def __init__(
...
+        dist = distribution_lib.distribution()
+        self._num_data_shards = 1
+        self._data_shard_id = 0
+        if dist is not None and getattr(dist, "auto_shard_dataset", False):
+            self._num_data_shards = min(
+                dist.num_model_replicas, dist.num_processes
+            )
+            self._data_shard_id = dist.data_shard_id
+
+    def _tf_shuffle(self, tensors, epoch_seed):
```
hertschuh: No need to pass `epoch_seed`, just use `self._epoch` directly.

keras/src/trainers/data_adapters/array_data_adapter.py
```diff
@@ -91,6 +92,27 @@ def __init__(
...
+        """Centralized helper for standard vs stateless cluster TF shuffling."""
+        from keras.src.utils.module_utils import tensorflow as tf
+
+        if self._num_data_shards > 1:
```
hertschuh: I'm not sure this distinction is useful. How about doing `tf.random.experimental.stateless_shuffle` always?

keras/src/trainers/data_adapters/array_data_adapter.py
```diff
@@ -295,21 +333,50 @@ def __len__(self):
...
+                    self.generator = generator
+                    self.adapter = adapter
+
+                def __iter__(self):
+                    if self.adapter._num_data_shards > 1:
+                        self.generator.manual_seed(self.adapter._epoch)
+                    yield from self.batch_sampler
+
+                def __len__(self):
+                    return len(self.batch_sampler)
```
hertschuh: I don't think we need both `RandomBatchSampler` and `SeededBatchSampler`. Can we combine them?

keras/src/trainers/data_adapters/array_data_adapter.py
```diff
@@ -276,15 +306,23 @@ def slice_and_convert(sliceable):
...
 
             def __len__(self):
-                return len(self.array[0])
+                return self.num_samples
 
         class RandomBatchSampler(torch.utils.data.Sampler):
-            def __init__(self, sampler):
+            def __init__(self, sampler, adapter):
                 self.sampler = sampler
+                self.adapter = adapter
```
hertschuh: I was confused by the name `adapter` until I realized it's the `ArrayDataAdapter` itself.

I would go one of two ways:
- either call it `seed_provider` and make it a callable that would be `lambda: self._epoch`
- or call it `array_data_adapter`.

keras/src/trainers/data_adapters/array_data_adapter.py
```diff
@@ -276,15 +306,23 @@ def slice_and_convert(sliceable):
...
 
             def __iter__(self):
+                generator = None
+                if self.adapter._num_data_shards > 1:
+                    generator = torch.Generator()
+                    generator.manual_seed(self.adapter._epoch)
```
hertschuh: Does it hurt if we always do what's in the `if`?

keras/src/trainers/data_adapters/data_adapter_utils.py
```diff
@@ -355,6 +355,49 @@ def is_torch_tensor(value):
     return False
 
 
+class DistributedBatchSampler:
+    def __init__(
+        self, batch_sampler, num_data_shards, data_shard_id, drop_last=False
+    ):
```
hertschuh, line 379: Is it just for torch or is it more general?

Can you add the docstring for `__init__` and in particular document the type of `batch_sampler`?

keras/src/trainers/data_adapters/data_adapter_utils.py
```diff
@@ -355,6 +355,49 @@ def is_torch_tensor(value):
...
+            (sampler_len // self.num_data_shards) * self.num_data_shards
+            if self.drop_last
+            else sampler_len
+        )
+
+        for i, batch in enumerate(self.batch_sampler):
+            if i >= limit:
+                break
+            if i % self.num_data_shards == self.data_shard_id:
+                yield batch
```
hertschuh, line 433: This approach works. But the thing with `ArrayDataAdapter` is that you can access the correct data directly by index.

What this code does is create all batches, shuffle them etc. and eventually drop most of them.

It would be nice to not even create those batches that are going to be dropped in the first place.

But see my high level comment first. Maybe the high level design will dictate whether or not you can add this optimization.

keras/src/trainers/data_adapters/array_data_adapter.py
hertschuh, line 272: I'm a bit confused here. For TensorFlow, you changed things to fix the shuffling, but the actual sharding is not done. Right?

hertschuh (commented):
High level comment:

The approach, until now, was that `Distribution.distribute_dataset` would be in charge of distributing the dataset.

With this PR, a new mechanism is introduced. The `DataAdapater` retrieves the current distribution and peeks into it to turn on distribution without using `Distribution.distribute_dataset`.

The top priority is to have one mechanism and use it consistently. And maybe the current approach doesn't make sense, and we should move the code from `Distribution.distribute_dataset` to `tf_dataset_adapter.py`.

keras/src/distribution/distribution_lib.py
```diff
@@ -450,76 +450,12 @@ def distribute_dataset(self, dataset):
         )
 
     def distribute_tf_dataset(self, dataset):
```
hertschuh: You can remove `distribute_dataset` and `distribute_tf_dataset` completely.

keras/src/trainers/data_adapters/array_data_adapter.py
```diff
@@ -231,11 +253,20 @@ def grab_one(x):
...
         indices_dataset = indices_dataset.flat_map(slice_batch_indices)
+
+        if self._num_data_shards > 1:
+            indices_dataset = indices_dataset.shard(
+                self._num_data_shards, self._data_shard_id
+            )
+
         if shuffle == "batch":
-            indices_dataset = indices_dataset.map(tf.random.shuffle)
+            indices_dataset = indices_dataset.map(lambda x: self._tf_shuffle(x))
```
hertschuh: `indices_dataset = indices_dataset.map(self._tf_shuffle)` should work.

keras/src/trainers/data_adapters/array_data_adapter.py
```diff
@@ -276,40 +308,91 @@ def slice_and_convert(sliceable):
...
+                for batch in self.batch_sampler:
+                    if self.shuffle_batch:
+                        yield [
+                            batch[i]
+                            for i in torch.randperm(
+                                len(batch), generator=batch_generator
+                            )
+                        ]
+                    else:
+                        yield batch
```
hertschuh: But this is a no-op if `shuffle_batch` is False, so we don't need it at all in this case.

So make `generator` mandatory, remove the `shuffle_batch` argument and rename `RandomBatchSampler`.

keras/src/trainers/data_adapters/array_data_adapter.py
```diff
@@ -276,40 +308,91 @@ def slice_and_convert(sliceable):
...
+                batch_sampler,
+                seed_provider=lambda: self._epoch,
+                shuffle_batch=True,
+            )
+        elif self._shuffle:
+            batch_sampler = SeededBatchSampler(
+                batch_sampler,
+                seed_provider=lambda: self._epoch,
+                generator=generator,
+            )
```
hertschuh: But with `shuffle_batch=False` (the default) this is a no-op.

If think this is missing the global shuffling logic. You have to draw indices once for all for the whole size of the array.

keras/src/trainers/data_adapters/tf_dataset_adapter.py
```diff
@@ -30,10 +31,69 @@ def __init__(self, dataset, class_weight=None, distribution=None):
...
                 make_class_weight_map_fn(class_weight)
             ).prefetch(tf.data.AUTOTUNE)
-        if distribution is not None:
-            dataset = distribution.distribute_dataset(dataset)
+
+        distribution = distribution or distribution_lib.distribution()
+        if (
+            distribution is not None
+            and distribution.auto_shard_dataset
+            and isinstance(dataset, tf.data.Dataset)
```
hertschuh: ` and isinstance(dataset, tf.data.Dataset)` this is not needed, this is checked line 23-24.

keras/src/trainers/data_adapters/tf_dataset_adapter.py
```diff
@@ -30,10 +31,69 @@ def __init__(self, dataset, class_weight=None, distribution=None):
             dataset = dataset.map(
                 make_class_weight_map_fn(class_weight)
             ).prefetch(tf.data.AUTOTUNE)
-        if distribution is not None:
-            dataset = distribution.distribute_dataset(dataset)
+
+        distribution = distribution or distribution_lib.distribution()
```
hertschuh: This is not needed, it's done in `get_data_adapter`.

hertschuh (changes requested):

keras/src/trainers/data_adapters/array_data_adapter.py
```diff
@@ -231,11 +253,20 @@ def grab_one(x):
...
+
         if shuffle == "batch":
-            indices_dataset = indices_dataset.map(tf.random.shuffle)
+            indices_dataset = indices_dataset.map(self._tf_shuffle)
 
         dataset = slice_inputs(indices_dataset, self._inputs)
 
+        # Note: sharding is also handled by `TFDatasetAdapter` if a distribution
+        # is active. We set `AutoShardPolicy.DATA` to ensure that
+        # `tf.distribute` will also shard the dataset by default.
```
hertschuh: I'm very confused by this comment.

Sure, `TFDatasetAdapter` handles distribution, but we're not using `TFDatasetAdapter` in any here, are we? There's no risk of interference, right?

keras/src/trainers/data_adapters/array_data_adapter.py
```diff
@@ -276,40 +308,84 @@ def slice_and_convert(sliceable):
                 )
 
             def __len__(self):
-                return len(self.array[0])
+                return self.num_samples
+
+        class SeededBatchSampler(torch.utils.data.Sampler):
+            def __init__(
```
hertschuh: It took me a while the understand the use and difference between `SeedBatchSampler` and `RandomBatchSampler` because:
- they both randomize
- they both are seeded
- they both yield batches

Let's rename them:
- `BatchShufflingSampler`
- `EpochShufflingSampler` or `GlobalShufflingSampler`.

keras/src/trainers/data_adapters/array_data_adapter.py
```diff
@@ -91,6 +93,27 @@ def __init__(
         self._batch_size = batch_size
         self._partial_batch_size = num_samples % batch_size
         self._shuffle = shuffle
+        self._epoch = backend.Variable(0, dtype="int64", trainable=False)
+
```
hertschuh: Why use a variable here? I really don't think we need this (rather than a plain int).

hertschuh (changes requested):

keras/src/trainers/data_adapters/array_data_adapter_test.py
```diff
@@ -328,3 +303,208 @@ def test_errors(self):
...
+        expected_rank,
+        mock_distribution,
+        mock_backend_dist_lib,
+        mock_get_rank,
+        mock_get_world_size,
+    ):
+        mock_get_world_size.return_value = world_size
+        mock_get_rank.return_value = rank
+        mock_backend_dist_lib.num_processes.return_value = world_size
+        mock_backend_dist_lib.process_id.return_value = rank
```
hertschuh: Why do you need to mock both the torch native one and the keras distribution one? Shouldn't one or the other be enough?

keras/src/trainers/data_adapters/array_data_adapter_test.py
```diff
@@ -328,3 +303,208 @@ def test_errors(self):
...
+                axis_names=("data", "model"),
+                devices=["cpu:0"] * np.prod(mesh_shape),
+            )
+            dist = dist_lib.ModelParallel(
+                device_mesh=device_mesh,
+                layout_map=dist_lib.LayoutMap(device_mesh),
+                batch_dim_name="data",
+            )
+        dist.auto_shard_dataset = True
+        mock_distribution.return_value = dist
```
hertschuh: I prefer mocking only what cannot be achieved without mocking.

Here you can just do `distribution_lib.set_distribution(...)`.

Then put a `tearDown` with `distribution_lib.set_distribution(None)`

keras/src/trainers/data_adapters/array_data_adapter_test.py
```diff
@@ -328,3 +303,208 @@ def test_errors(self):
...
+            new_dataloader.batch_sampler,
+            array_data_adapter.data_adapter_utils.DistributedBatchSampler,
+        )
+        self.assertEqual(
+            new_dataloader.batch_sampler.num_data_shards,
+            expected_num_replicas,
+        )
+        self.assertEqual(
+            new_dataloader.batch_sampler.data_shard_id, expected_rank
+        )
```
hertschuh: This is a bit of a weak verification. Can you instead use for instance use arrays that are a sequence (use `make_array`), and then verify that the returned values are the expected sequence, for instance `(1, 5, 9, 13)`.

See https://github.com/keras-team/keras/blob/master/keras/src/trainers/data_adapters/array_data_adapter_test.py#L130-L140

keras/src/trainers/data_adapters/array_data_adapter_test.py
```diff
@@ -328,3 +303,208 @@ def test_errors(self):
...
+        # rank 1 should get batches 1, 3
+        adapter = array_data_adapter.ArrayDataAdapter(
+            x, y, batch_size=10, shuffle=False
+        )
+
+        self.assertEqual(adapter._num_data_shards, 2)
+        self.assertEqual(adapter._data_shard_id, rank)
+
+        it = adapter.get_numpy_iterator()
+        batches = list(it)
```
hertschuh: Except for the `get_numpy_iterator` vs. `get_data_loader` there is no reason why this test is different from the previous one. Please combine them like `test_basic_flow` above.

keras/src/trainers/data_adapters/array_data_adapter_test.py
```diff
@@ -328,3 +303,208 @@ def test_errors(self):
...
+        mock_distribution.return_value = dist
+
+        x = np.arange(100).reshape((50, 2))
+        y = np.arange(50).reshape((50, 1))
+        adapter = array_data_adapter.ArrayDataAdapter(
+            x, y, batch_size=10, shuffle=False
+        )
+
+        it = adapter.get_jax_iterator()
+        batches = list(it)
```
hertschuh: Except for the `get_jax_iterator` vs. `get_data_loader` there is no reason why this test is different from the previous one. Please combine them like `test_basic_flow` above.

keras/src/trainers/data_adapters/array_data_adapter_test.py
```diff
@@ -328,3 +303,208 @@ def test_errors(self):
...
+
+        if rank == 0:
+            self.assertEqual(len(batches), 3)
+        else:
+            self.assertEqual(len(batches), 2)
+
+    def test_deterministic_shuffle(self):
+        x = np.arange(100).reshape((50, 2)).astype("float32")
+        y = np.arange(50).reshape((50, 1)).astype("float32")
+
```
hertschuh: Please try to combine with the 3 tests above like `test_basic_flow` above does. `shuffle` can be another parameter combined with the other options.

keras/src/trainers/data_adapters/data_adapter_utils.py
```diff
@@ -355,6 +355,88 @@ def is_torch_tensor(value):
     return False
 
 
+class DistributedBatchSampler:
+    """Sampler that shards a batch sampler.
```
hertschuh, line 359: Why can't you use `torch.utils.data.distributed.DistributedSampler`?

keras/src/trainers/data_adapters/tf_dataset_adapter_test.py
```diff
@@ -208,16 +209,17 @@ def test_class_weights_map_fn_nested_y(self):
             class_weights_map_fn(x, (y1, y2))
 
     def test_distribute_dataset(self):
-        x = tf.random.normal((34, 4))
-        y = tf.random.normal((34, 2))
+        x = tf.random.normal((32, 4))
+        y = tf.random.normal((32, 2))
```
hertschuh: Why was this test changed?

It was a great way to test that the last batch was handled correctly.

Doesn't this work anymore?

keras/src/trainers/data_adapters/tf_dataset_adapter_test.py
```diff
@@ -350,3 +344,248 @@ def dataset_fn(input_context):
...
+                distribution.__class__,
+                "num_processes",
+                new_callable=mock.PropertyMock,
+                return_value=2,
+            ),
+            mock.patch.object(
+                distribution.__class__,
+                "data_shard_id",
+                new_callable=mock.PropertyMock,
+                return_value=0,
```
hertschuh: Can you use a different value than 0 for "data_shard_id"?

You wouldn't be able to find out that "data_shard_id" is not implemented correctly if you only test with 0.

keras/src/trainers/data_adapters/tf_dataset_adapter_test.py
```diff
@@ -350,3 +344,248 @@ def dataset_fn(input_context):
...
+                new_callable=mock.PropertyMock,
+                return_value=2,
+            ),
+            mock.patch.object(
+                distribution.__class__,
+                "data_shard_id",
+                new_callable=mock.PropertyMock,
+                return_value=1,
+            ),
+        ):
```
hertschuh: In the test above, you just do this:
```
        data_distribution.num_model_replicas = 2
        data_distribution.num_processes = 2
        data_distribution.data_shard_id = 0
```
Which is so much easier and more readable. Can't you just do that here?

keras/src/trainers/data_adapters/tf_dataset_adapter_test.py
```diff
@@ -350,3 +344,248 @@ def dataset_fn(input_context):
...
+                new_callable=mock.PropertyMock,
+                return_value=4,
+            ),
+            mock.patch.object(
+                distribution.__class__,
+                "num_processes",
+                new_callable=mock.PropertyMock,
+                return_value=2,
+            ),
+        ):
```
hertschuh: Same comment.

keras/src/trainers/data_adapters/tf_dataset_adapter_test.py
```diff
@@ -350,3 +344,248 @@ def dataset_fn(input_context):
...
+        distribution = distribution_lib.ModelParallel(
+            layout_map=layout_map, batch_dim_name="data"
+        )
+        dataset = tf.data.Dataset.range(16)
+        with mock.patch.object(
+            distribution.__class__,
+            "num_processes",
+            new_callable=mock.PropertyMock,
+            return_value=2,
+        ):
```
hertschuh: Same comment.

keras/src/trainers/data_adapters/tf_dataset_adapter_test.py
```diff
@@ -350,3 +344,248 @@ def dataset_fn(input_context):
...
+                new_callable=mock.PropertyMock,
+                return_value=1,
+            ),
+            mock.patch.object(
+                distribution.__class__,
+                "num_processes",
+                new_callable=mock.PropertyMock,
+                return_value=2,
+            ),
+        ):
```
hertschuh: Same comment.

keras/src/trainers/data_adapters/tf_dataset_adapter_test.py
```diff
@@ -350,3 +344,248 @@ def dataset_fn(input_context):
...
+                        "num_processes",
+                        new_callable=mock.PropertyMock,
+                        return_value=num_devices,
+                    ),
+                    mock.patch.object(
+                        distribution.__class__,
+                        "data_shard_id",
+                        new_callable=mock.PropertyMock,
+                    ) as mock_shard_id,
+                ):
```
hertschuh: Same comment about not mocking.

hertschuh (changes requested):

keras/src/trainers/data_adapters/array_data_adapter_test.py
buildwithsuhana: Removed mocking

keras/src/trainers/data_adapters/array_data_adapter_test.py
buildwithsuhana: Done, used built-in with dist.scope(): functionality

keras/src/trainers/data_adapters/array_data_adapter_test.py
buildwithsuhana: Done

keras/src/trainers/data_adapters/array_data_adapter_test.py
buildwithsuhana: Done, refactored to the test_sharding test.

keras/src/trainers/data_adapters/array_data_adapter_test.py
buildwithsuhana: Done, refactored to the test_sharding test.

keras/src/trainers/data_adapters/array_data_adapter_test.py
buildwithsuhana: Done, Folded into test_sharding

keras/src/trainers/data_adapters/data_adapter_utils.py
buildwithsuhana: The standard torch.utils.data.distributed.DistributedSampler shards individual sample indices before they are batched. However, ArrayDataAdapter is architected around batch-level iteration where the underlying sampler already produces full batches (lists of indices).
DistributedBatchSampler was implemented as a batch-level sharder to allow us to distribute any existing batch level iterator directly. This preserves our current batch centric logic while keeping the implementation backend-agnostic.

keras/src/trainers/data_adapters/tf_dataset_adapter_test.py
buildwithsuhana: Restored the dataset size back to 34

keras/src/trainers/data_adapters/tf_dataset_adapter_test.py
buildwithsuhana: Fixed, now performs sequential checks for both distribution.data_shard_id = 0 and distribution.data_shard_id = 1

keras/src/trainers/data_adapters/tf_dataset_adapter_test.py
buildwithsuhana: Done

keras/src/trainers/data_adapters/tf_dataset_adapter_test.py
buildwithsuhana: Done, removed redundant mocks

keras/src/trainers/data_adapters/tf_dataset_adapter_test.py
buildwithsuhana: removed the mocking

keras/src/trainers/data_adapters/tf_dataset_adapter_test.py
buildwithsuhana: removed the mocking

keras/src/trainers/data_adapters/tf_dataset_adapter_test.py
buildwithsuhana: removed mocking here as well

keras/src/trainers/data_adapters/array_data_adapter_test.py
```diff
@@ -328,3 +329,132 @@ def test_errors(self):
...
+                for i, batch in enumerate(batches):
+                    bx, by = batch
+                    bx = backend.convert_to_numpy(bx)
+                    by = backend.convert_to_numpy(by)
+                    expected_batch_index = (
+                        expected_rank + i * expected_num_replicas
+                    )
+                    expected_first_sample_value = expected_batch_index * 10
+                    self.assertAllClose(bx[0, 0], expected_first_sample_value)
+                    self.assertAllClose(by[0, 0], expected_first_sample_value)
```
hertschuh: You can check the whole batch, not just the first value:
```
self.assertAllClose(bx[:, 0], np.arange(expected_first_sample_value, expected_first_sample_value + 10)
```

keras/src/trainers/data_adapters/array_data_adapter_test.py
```diff
@@ -328,3 +329,132 @@ def test_errors(self):
...
+                },
+            ],
+            shuffle=[False, True],
+        )
+    )
+    def test_sharding(
+        self,
+        dist_type,
+        world_size,
+        rank,
```
hertschuh: These are very torch specific names. Let's call them the same way in the rest of the code, that `num_processes` and `process_id`, right?

hertschuh (changes requested):
Last nitpicks:

keras/src/trainers/data_adapters/array_data_adapter_test.py
buildwithsuhana: Renamed

keras/src/trainers/data_adapters/array_data_adapter_test.py
buildwithsuhana: Done

hertschuh (approved):
Thanks!

PR #23034, opened 2026-06-04
🛡️ Sentinel: [ENHANCEMENT] Explicitly disable pickle in `np.load`

rni418:
🛡️ Sentinel: [ENHANCEMENT] Explicitly disable pickle in `np.load`

🚨 Severity: LOW / ENHANCEMENT
💡 Vulnerability: `np.load` was called without explicitly specifying `allow_pickle=False`. While modern NumPy versions default to `False`, relying on defaults can lead to insecure deserialization if the environment uses an older version or the default changes.
🎯 Impact: Insecure deserialization can lead to arbitrary code execution if a maliciously crafted `.npz` file is loaded.
🔧 Fix: Explicitly pass `allow_pickle=False` to `np.load` in `keras/src/saving/saving_lib.py`.
✅ Verification: Ran `pytest keras/src/saving/saving_lib_test.py` to ensure no regressions.

## Contributor Agreement

**Please review our [AI-Assisted Contribution Policy](../CONTRIBUTING.md#ai-assisted-contribution-policy) and check all boxes below before submitting your PR for review:**

- [x] I am a human, and not a bot.
- [x] I will be responsible for responding to review comments in a timely manner.
- [x] I will work with the maintainers to push this PR forward until submission.

**Note:** Failing to adhere to this agreement may result in your future PRs no longer being reviewed.

---
*PR created automatically by Jules for task [5579081950035332646](https://jules.google.com/task/5579081950035332646) started by @rni418*

PR #23026, opened 2026-06-03
🛡️ Sentinel: [HIGH] Fix insecure deserialization in dataset utilities

rni418:
🚨 Severity: HIGH
💡 Vulnerability: Insecure deserialization via `np.load(..., allow_pickle=True)`
🎯 Impact: A maliciously crafted `.npz` file could lead to arbitrary code execution if downloaded/loaded.
🔧 Fix: Explicitly set `allow_pickle=False` for standard numerical array datasets (`mnist`, `boston_housing`, `california_housing`), preventing potential insecure deserialization.
✅ Verification: Tested the datasets load correctly via `load_data()` with `allow_pickle=False`.

## Contributor Agreement
**Please review our [AI-Assisted Contribution Policy](../CONTRIBUTING.md#ai-assisted-contribution-policy) and check all boxes below before submitting your PR for review:**
- [X] I am a human, and not a bot.
- [X] I will be responsible for responding to review comments in a timely manner.
- [X] I will work with the maintainers to push this PR forward until submission.
**Note:** Failing to adhere to this agreement may result in your future PRs no longer being reviewed.

---
*PR created automatically by Jules for task [700360888984662221](https://jules.google.com/task/700360888984662221) started by @rni418*

PR #23047, opened 2026-06-06
Restrict unpickling when loading IMDB and Reuters datasets

LinZiyuu:
## Description

`keras.datasets.imdb.load_data` and `keras.datasets.reuters.load_data` load
their `.npz` files with `np.load(allow_pickle=True)`. Loading a crafted `.npz`
therefore unpickles arbitrary objects, which can execute code through a pickle
`__reduce__` gadget (insecure deserialization, CWE-502).

#23026 and #23034 removed `allow_pickle=True` from the numeric datasets
(`mnist`, `boston_housing`, `california_housing`) and `NpzIOStore`, but IMDB and
Reuters were left unchanged. These two datasets store ragged sequences as
`dtype=object` arrays, which numpy can only persist via pickle, so
`allow_pickle=False` cannot be used here without breaking the dataset format.

This PR adds `keras.src.datasets.npz_utils.load_npz`, used by both loaders:

- numeric members are read with pickling fully disabled;
- object members are read with a restricted unpickler that only permits numpy
  array reconstruction (`numpy.ndarray`, `numpy.dtype`, and the
  `numpy(._core).multiarray` `_reconstruct`/`scalar` globals).

This keeps the existing dataset format working while preventing a crafted
`.npz` from executing arbitrary code. A new `npz_utils_test.py` covers loading
ragged object arrays, loading numeric arrays, and rejecting a pickle gadget
without executing its payload.

## Contributor Agreement

### DIFF
--- a/keras/src/datasets/npz_utils.py
+++ b/keras/src/datasets/npz_utils.py
@@ -6,27 +6,29 @@ import zipfile
 
 import numpy as np
 
-# Globals that are legitimately required to reconstruct the numpy object arrays
-# stored inside the IMDB / Reuters `.npz` files (ragged sequences are saved as
-# `dtype=object` arrays, which numpy persists as pickle streams). Both the
-# numpy < 2 (`numpy.core`) and numpy >= 2 (`numpy._core`) module spellings are
-# listed so files written by either can be read. Anything outside this set
-# (e.g. `os.system`, `builtins.eval`, `subprocess.Popen`) is refused, which
-# prevents a tampered `.npz` from executing code through a pickle `__reduce__`
-# gadget.
+# Globals required to reconstruct the numpy object arrays stored inside the
+# IMDB / Reuters `.npz` files. These datasets store ragged sequences as
+# `dtype=object` arrays, which numpy can only persist as a pickle stream, so
+# `np.load(allow_pickle=False)` refuses them outright ("Object arrays cannot be
+# loaded when allow_pickle=False"). Rebuilding such an array uses exactly three
+# globals -- `numpy.ndarray`, `numpy.dtype` and `multiarray._reconstruct` -- and
+# nothing else (verified against the actual IMDB and Reuters files). Both the
+# numpy < 2 (`numpy.core`) and numpy >= 2 (`numpy._core`) spellings of
+# `multiarray` are listed so files written by either can be read. Anything
+# outside this set (e.g. `os.system`, `builtins.eval`, `subprocess.Popen`) is
+# refused, which prevents a tampered `.npz` from executing code through a pickle
+# `__reduce__` gadget.
 _ALLOWED_PICKLE_GLOBALS = frozenset(
     {
         ("numpy", "ndarray"),
         ("numpy", "dtype"),
         ("numpy.core.multiarray", "_reconstruct"),
-        ("numpy.core.multiarray", "scalar"),
         ("numpy._core.multiarray", "_reconstruct"),
-        ("numpy._core.multiarray", "scalar"),
     }
 )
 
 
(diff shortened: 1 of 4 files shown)


keras/src/datasets/npz_utils.py
```diff
@@ -0,0 +1,91 @@
...
+        ("numpy", "dtype"),
+        ("numpy.core.multiarray", "_reconstruct"),
+        ("numpy.core.multiarray", "scalar"),
+        ("numpy._core.multiarray", "_reconstruct"),
+        ("numpy._core.multiarray", "scalar"),
+    }
+)
+
+
+class _RestrictedUnpickler(pickle.Unpickler):
```
hertschuh: Rename `_RestrictedUnpickler` to `RestrictedUnpickler` for consistency with our coding style.

keras/src/datasets/npz_utils.py
```diff
@@ -0,0 +1,91 @@
...
+    def find_class(self, module, name):
+        if (module, name) in _ALLOWED_PICKLE_GLOBALS:
+            return super().find_class(module, name)
+        raise pickle.UnpicklingError(
+            f"Refusing to deserialize `{module}.{name}` while loading a Keras "
+            "dataset. The `.npz` file may be corrupted or malicious."
+        )
+
+
+def _load_npy_member(fp):
```
hertschuh: Rename `_load_npy_member` to `load_npy_member` for consistency with our coding style.

keras/src/datasets/npz_utils.py
```diff
@@ -0,0 +1,91 @@
...
+# prevents a tampered `.npz` from executing code through a pickle `__reduce__`
+# gadget.
+_ALLOWED_PICKLE_GLOBALS = frozenset(
+    {
+        ("numpy", "ndarray"),
+        ("numpy", "dtype"),
+        ("numpy.core.multiarray", "_reconstruct"),
+        ("numpy.core.multiarray", "scalar"),
+        ("numpy._core.multiarray", "_reconstruct"),
+        ("numpy._core.multiarray", "scalar"),
```
hertschuh: Just for my understanding, what in this list is used by Reuters and IMDB that is not allowed by `np.load(allow_pickle=False)`?

hertschuh (changes requested):

LinZiyuu:
Thanks for the review @hertschuh! Addressed in 290b22ef.

- Renamed _RestrictedUnpickler → RestrictedUnpickler and _load_npy_member → load_npy_member to match the naming convention.

Just for my understanding, what in this list is used by Reuters and IMDB that is not allowed by np.load(allow_pickle=False)?

Good question. The short answer is that none of them are allowed by allow_pickle=False. That flag refuses object arrays entirely (ValueError: Object arrays cannot be loaded when allow_pickle=False), so it can't load these datasets at all. IMDB and Reuters store their ragged sequences as dtype=object arrays, which numpy can only persist as a pickle stream.

I traced find_class against the actual imdb.npz / reuters.npz, and each object member only ever needs three globals:

  - numpy.ndarray
  - numpy.dtype
  - numpy.core.multiarray._reconstruct

The hosted files were written with numpy < 2, hence the numpy.core spelling; a numpy ≥ 2 re-save emits numpy._core.multiarray._reconstruct instead, which is why both spellings are listed. multiarray.scalar was never actually exercised by either dataset (the elements are plain Python lists of ints, not numpy scalars), so I've dropped those two entries — the allowlist is now exactly the set the datasets need, and anything else (os.system, builtins.eval, …) still raises UnpicklingError.

hertschuh (approved):
Thanks for implementing this!

PR #23048, opened 2026-06-06
Make Lambda and TorchModuleWrapper from_config fail closed when safe_mode is unset

LinZiyuu:
## Description

`Lambda.from_config` computed `safe_mode = safe_mode or in_safe_mode()`, which
evaluates to `None` (falsy) when a `Lambda` is deserialized outside of a
`SafeModeScope` and without an explicit `safe_mode` — for example a direct
`keras.Model.from_config(...)` / `keras.Sequential.from_config(...)` call, or a
layer config that omits the `"module"` key (which is routed through the legacy
deserializer). In that case the safe-mode guard did not fire and the marshalled
lambda bytecode was loaded via `func_load`, which can execute arbitrary code.
`TorchModuleWrapper.from_config` had the same problem with its `torch.load()`
pickle (`if in_safe_mode():`).

This makes both fail closed: an unset safe mode is treated as safe, matching the
handling already used by `TFSMLayer.from_config`. Deserialization is refused
unless safe mode is explicitly disabled via `safe_mode=False`,
`SafeModeScope(False)`, or `keras.config.enable_unsafe_deserialization()`.
`load_model` and `model_from_json` are unaffected — they already establish a
`SafeModeScope`, so the default loading paths behave exactly as before.

Adds regression tests for both layers (deserializing outside a `SafeModeScope`
is now refused) and updates the `TorchModuleWrapper` round-trip test to disable
safe mode explicitly.

## Contributor Agreement

### DIFF
--- a/keras/src/utils/torch_utils.py
+++ b/keras/src/utils/torch_utils.py
@@ -167,7 +167,11 @@ class TorchModuleWrapper(Layer):
         import torch
 
         if "module" in config:
-            if in_safe_mode():
+            # Fail closed: an unset ambient scope (`in_safe_mode()` is `None`,
+            # e.g. a direct `from_config` call) is treated as safe, so a
+            # `torch.load()` pickle is not deserialized unless safe mode was
+            # explicitly disabled.
+            if in_safe_mode() is not False:
                 raise ValueError(
                     "Requested the deserialization of a `torch.nn.Module` "
                     "object via `torch.load()`. This carries a potential risk "
--- a/keras/src/utils/torch_utils.py
+++ b/keras/src/utils/torch_utils.py
@@ -163,15 +163,19 @@ class TorchModuleWrapper(Layer):
         return {**base_config, **config}
 
     @classmethod
-    def from_config(cls, config):
+    def from_config(cls, config, custom_objects=None, safe_mode=None):
         import torch
 
         if "module" in config:
-            # Fail closed: an unset ambient scope (`in_safe_mode()` is `None`,
-            # e.g. a direct `from_config` call) is treated as safe, so a
-            # `torch.load()` pickle is not deserialized unless safe mode was
-            # explicitly disabled.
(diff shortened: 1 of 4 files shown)


keras/src/utils/torch_utils.py
```diff
@@ -167,7 +167,11 @@ def from_config(cls, config):
         import torch
 
         if "module" in config:
-            if in_safe_mode():
+            # Fail closed: an unset ambient scope (`in_safe_mode()` is `None`,
+            # e.g. a direct `from_config` call) is treated as safe, so a
+            # `torch.load()` pickle is not deserialized unless safe mode was
+            # explicitly disabled.
+            if in_safe_mode() is not False:
```
hertschuh: Can you add a `safe_mode` argument to `from_config` here to follow the pattern of the other ones.

hertschuh (changes requested):
Thanks for addressing this!

keras/src/utils/torch_utils.py
LinZiyuu: Done — `TorchModuleWrapper.from_config` now takes `custom_objects=None, safe_mode=None` to match `Lambda.from_config` / `TFSMLayer.from_config`, and falls back to the ambient scope (`safe_mode if safe_mode is not None else in_safe_mode()`), failing closed when it is unset. The test now also exercises passing `safe_mode=False` explicitly alongside the `SafeModeScope` path.

hertschuh (approved):
Thanks!

LinZiyuu:
Thanks @hertschuh,  just a gentle nudge on this. Since this PR is merged, validating the reports on Huntr should be quick and it needs to happen before Huntr locks OSS submissions on July 31, 2026. Would you  be able to take a look before then? Thanks so much!

https://huntr.com/bounties/98fde685-0b1d-47d0-a477-805715c3de62

PR #23049, opened 2026-06-07
Fix: do not follow symlinks when writing downloads in `get_file`

LinZiyuu:
## Description

`get_file` downloaded to the predictable path
`<cache>/<cache_subdir>/<fname>` by passing it directly to `urlretrieve`, which
opens the destination with `open(path, "wb")` and therefore **follows a symbolic
link** present at that path. If the cache directory is writable by another user —
for example the world-writable `/tmp/.keras` fallback that is used when
`~/.keras` is not writable, or a shared `KERAS_HOME` — an attacker can pre-create
a symlink at the (predictable) download path so that the downloaded bytes are
written **through** it to an arbitrary location outside the cache directory.
Pointing the symlink at e.g. a Python `*.pth` file or `~/.config/autostart` turns
this into code execution.

This downloads to a unique temporary file in the cache directory, validates the
hash, and then atomically `os.replace()`s it into place. `os.replace()` replaces
a symlink at the destination instead of writing through it, so a download can no
longer escape the cache directory. As a side benefit the hash is now checked
before the file is moved into place, so a corrupted download is never left in the
cache.

A regression test is added verifying that a pre-planted symlink at the download
path is not followed.

## Contributor Agreement

### DIFF
--- a/keras/src/utils/file_utils.py
+++ b/keras/src/utils/file_utils.py
@@ -384,24 +384,25 @@ def get_file(
                         self.finished = True
 
         error_msg = "URL fetch failure on {}: {} -- {}"
+        # Download to a temporary file in the cache directory and atomically
+        # move it into place. Writing directly to `download_target` would
+        # follow a pre-existing symlink at that path and could place the
+        # downloaded file outside the cache directory.
+        fd, tmp_download_target = tempfile.mkstemp(dir=datadir)
+        os.close(fd)
         try:
             try:
-                urlretrieve(origin, download_target, DLProgbar())
+                urlretrieve(origin, tmp_download_target, DLProgbar())
             except urllib.error.HTTPError as e:
                 raise Exception(error_msg.format(origin, e.code, e.msg))
             except urllib.error.URLError as e:
                 raise Exception(error_msg.format(origin, e.errno, e.reason))
-        except (Exception, KeyboardInterrupt):
-            if os.path.exists(download_target):
-                os.remove(download_target)
-            raise
 
-        # Validate download if succeeded and user provided an expected hash
-        # Security conscious users would get the hash of the file from a
-        # separate channel and pass it to this API to prevent MITM / corruption:
-        if os.path.exists(download_target) and file_hash is not None:
-            if not validate_file(
-                download_target, file_hash, algorithm=hash_algorithm
+            # Validate the download before moving it into place. Security
+            # conscious users would get the hash of the file from a separate
+            # channel and pass it to this API to prevent MITM / corruption:
+            if file_hash is not None and not validate_file(
+                tmp_download_target, file_hash, algorithm=hash_algorithm
             ):
                 raise ValueError(
                     "Incomplete or corrupted file detected. "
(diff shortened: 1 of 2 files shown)


LinZiyuu:
Hi @hertschuh, can you reiview this PR?

keras/src/utils/file_utils.py
```diff
@@ -384,31 +384,48 @@ def __call__(self, block_num, block_size, total_size):
...
+        )
+        os.close(fd)
+        # `mkstemp` creates the file with 0o600 permissions, but the previous
+        # `urlretrieve` wrote it through `open(..., "wb")`, which honors the
+        # umask (typically 0o644). Restore that so cached files stay readable
+        # by other users of a shared cache directory.
+        try:
+            os.chmod(tmp_download_target, 0o644)
+        except OSError:
+            pass
```
hertschuh, line 403: Why not just do
```python
tmp_download_target = os.path.join(datadir, tmp_prefix)
```
Instead of all of this:
- You're not using `fd`, you're just closing it
- You wouldn't have to deal with the `chmod`

I'm not sure what `mkstemp` provides here.

LinZiyuu: Good question — mkstemp is doing the security-relevant work here, and a predictable os.path.join(datadir, tmp_prefix) would re-open the exact hole this PR closes.
urlretrieve writes its destination with a plain open(filename, "wb") (CPython urllib/request.py), which follows a symlink at that path. The os.replace() only protects the final download_target; the bytes are still written to the temporary path first, so that temp path also has to be un-plantable. If it's predictable, an attacker who can write to the cache dir — the /tmp/.keras fallback or a shared KERAS_HOME, i.e. this PR's own threat model — just plants the symlink at the temp name instead:
predictable = os.path.join(datadir, ".tmp_test.txt_")     # attacker knows fname
os.symlink("/home/victim/.config/autostart/x.desktop", predictable) # later: urlretrieve(origin, predictable) -> open(predictable, "wb") follows it

I ran it to confirm: writing to a predictable temp path with a pre-planted symlink lands the bytes on the outside target; mkstemp never does, because it creates the file itself with O_CREAT | O_EXCL under an unpredictable name — so there's guaranteed to be a real (non-symlink) file there before urlretrieve re-opens it, and two concurrent downloads of the same fname can't collide either.

The os.close(fd) and chmod are just the cost of handing the path to urlretrieve (which re-opens by name): I don't write through fd because I'm letting urlretrieve do the fetch, and mkstemp hardcodes 0o600, so the chmod(0o644) restores the umask-respecting readability the old open("wb") gave shared-cache readers (os.replace keeps the temp file's mode).

 If you'd rather drop the fd/chmod dance, the alternative is to write the response through fd directly (os.fdopen(fd, "wb")) instead of calling urlretrieve — that also closes the tiny re-open window — but it means reimplementing the fetch/progress loop, so I kept the smaller change. Happy to go that route if you'd prefer.

hertschuh: So the part of this PR that I like is that it makes the update atomic in the cache.

However, the whole security aspect of this does not make sense to me. If an attacker can plant symlinks in the cache folder, they can also link the whole cache or `.keras` folder to wherever they want, or mount a different filesystem.

If the filesystem is compromised by an attacker, there is nothing Keras can do about it.

PR #23050, opened 2026-06-07
Fix: refuse vocabulary file paths when deserializing lookup layers in safe mode

LinZiyuu:
## Description

`IndexLookup` / `StringLookup` / `IntegerLookup` / `TextVectorization` accept a
`vocabulary` argument that may be a string **file path**. `set_vocabulary`
guards loading such a path with `if in_safe_mode():`, which does **not** fire
when the layer is deserialized outside of a `SafeModeScope` — e.g. a direct
`keras.Model.from_config(...)` / `Layer.from_config(...)` call — because
`in_safe_mode()` returns `None` there (falsy). As a result, deserializing an
untrusted config whose `vocabulary` is a file path reads an arbitrary local file
(and, via `tf.io.gfile`, a remote URI such as `gs://` / `http(s)://`) into the
layer's vocabulary, which is then recoverable via `get_vocabulary()`.

This adds a `from_config` that fails closed: a string `vocabulary` (file path)
is refused unless safe mode is explicitly disabled (`safe_mode=False`,
`SafeModeScope(False)`, or `keras.config.enable_unsafe_deserialization()`),
matching the fail-closed handling already used by `TFSMLayer.from_config`.

Behavior that is intentionally **unchanged**:

- Direct construction `StringLookup(vocabulary="vocab.txt")` (not a
  deserialization) still loads the file.
- `load_model` / `model_from_json` already establish a `SafeModeScope`, so the
  normal loading paths are unaffected.
- The normal round-trip of a list / asset vocabulary (the only forms
  `get_config` produces) is unaffected.

Adds a regression test.

## Contributor Agreement

### DIFF
--- a/keras/src/layers/preprocessing/index_lookup.py
+++ b/keras/src/layers/preprocessing/index_lookup.py
@@ -11,6 +11,31 @@ from keras.src.utils import tf_utils
 from keras.src.utils.module_utils import tensorflow as tf
 
 
+def raise_for_unsafe_vocabulary_load(vocabulary):
+    """Fail closed when a vocabulary file path is deserialized in safe mode.
+
+    A vocabulary supplied as a string is a file path loaded from outside the
+    model archive (``tf.io.gfile`` also resolves remote URIs such as ``gs://``).
+    When reconstructing a layer from a (potentially untrusted) config via
+    ``from_config``, only allow it if safe mode has been explicitly disabled;
+    fail closed when safe mode is unset, so that deserializing an untrusted
+    config cannot read arbitrary local files or reach remote URLs.
+    """
+    if isinstance(vocabulary, str) and (
+        serialization_lib.in_safe_mode() is not False
+    ):
+        raise ValueError(
+            "Requested the loading of a vocabulary file outside of the model "
+            "archive while deserializing a layer. This carries a potential "
+            "risk of loading arbitrary and sensitive files and thus it is "
+            "disallowed by default. If you trust the source of the artifact, "
+            "you can override this error by passing `safe_mode=False` to the "
+            "loading function, or calling "
+            "`keras.config.enable_unsafe_deserialization()`. "
+            f"Vocabulary file: '{vocabulary}'"
+        )
+
+
 class IndexLookup(Layer):
     """Maps values from a vocabulary to integer indices.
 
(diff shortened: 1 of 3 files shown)


keras/src/layers/preprocessing/index_lookup.py
hertschuh, line 475: But this is already caught here.

`from_config` calls `__init__`, which calls `set_vocabulary`, which checks here.

I don't think this is handling any new case that wasn't already handled.

LinZiyuu: The check at L473-475 is set_vocabulary's existing guard, which is if in_safe_mode(): raise — fail-open. My from_config uses in_safe_mode() is not False — fail-closed. They only agree when a SafeModeScope is active. The case I'm fixing is a top-level StringLookup.from_config({"vocabulary": "<path>"}) with no active scope: in_safe_mode() returns None, so set_vocabulary's if in_safe_mode(): does not fire and the file is loaded. 
Repro on master:
StringLookup.from_config({"name": "sl", "vocabulary": "/etc/passwd"})  # reads the file; secret recoverable via get_vocabulary()
The added test test_from_config_vocabulary_path_fails_closed fails on master and passes with this PR. This mirrors TFSMLayer.from_config, which already fails closed on the same threat for the same reason.

hertschuh: So, I did submit a number of PR adding `safe_mode` to `from_config` because there was a precendent (which I don't understand).

However, I don't think the concept makes sense. `from_config` is a API that you **implement** as a user, not an API that you **call**. Users call the constructor of classes, not `from_config`. `from_config` is a JSON deserialization concept, not a public API.

And the constructor doesn't have a `safe_mode` argument because it doesn't make sense. As a user, locally, you can do whatever you want, including loading vocabulary files from anywhere.

So even if you, as a user, call `from_config` manually, there is no particular reason to reject things that the constructor wouldn't reject, it's on you to pass valid arguments to `from_config`.

LinZiyuu: Thanks for the detailed explanation, that makes sense. The actual untrusted-load entry points (load_model / model_from_json) already establish a SafeModeScope, so the real path is covered; this PR only guarded a direct from_config call, which — as you say — is a deserialization internal and the caller's responsibility.

PR #23051, opened 2026-06-07
Fix: avoid out-of-memory crash when building an oversized Embedding table

LinZiyuu:
## Description

When an `Embedding` layer is built, the default (random) `embeddings_initializer`
materializes the full `(input_dim, output_dim)` table. When a model is
reconstructed from a config (for example via `keras.saving.load_model`),
`build()` runs during architecture reconstruction — before any stored weights
are read — so a config that declares an enormous `input_dim` allocates the whole
table up front. A tiny (a few-KB) `.keras` file can therefore request terabytes
of memory at load time and get the process OOM-killed, before the weight-loading
shape checks ever run.

When `psutil` is available, this raises a clear `ValueError` if the embeddings
table would not fit in available memory, instead of attempting the allocation
and crashing. Models that actually fit in memory are unaffected, and when
`psutil` is not installed the behavior is unchanged.

Adds a regression test (skipped when `psutil` is not installed).

## Contributor Agreement

### DIFF
--- a/keras/src/layers/core/embedding.py
+++ b/keras/src/layers/core/embedding.py
@@ -181,21 +181,32 @@ class Embedding(Layer):
         the full `(input_dim, output_dim)` table at build time -- before any
         stored weights are read -- which can exhaust memory and get the process
         OOM-killed. When `psutil` is available, raise a clear error instead of
-        attempting an allocation that cannot fit in available memory.
+        attempting an allocation that obviously cannot fit in system memory.
+
+        The bound is intentionally generous (a multiple of total system
+        memory rather than currently-available memory) so that legitimate
+        large models are not rejected under transient memory pressure or when
+        the table is sharded across hosts; it only blocks the absurd
+        allocations (terabytes from a tiny config) used to trigger an OOM.
         """
         if psutil is None:
             return
         num_params = math.prod(embeddings_shape)
-        bytes_per_param = dtype_utils.dtype_size(self.variable_dtype) / 8
+        try:
+            bytes_per_param = dtype_utils.dtype_size(self.variable_dtype) / 8
+        except ValueError:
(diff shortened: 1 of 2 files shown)


LinZiyuu:
Hi @hertschuh, can you reiview this PR?

hertschuh:
First, this won't work. This code compares the size of embedding with the free CPU memory. The embedding can be created directly on-device HBM so the CPU memory is irrelevant. Then, there is sharding. Embeddings can be sharded across thousands of TPUs, just because the embedding size is many GB, hundreds of time bigger than the HBM or CPU memory doesn't mean it won't fit.

More generally answering the question "will this fit in memory" is a very hard problem. We're considering this out of scope for Keras. Many (most) APIs allocate memory, some temporary (for intermediate computation), some not (variables). Whether the forward pass will work or run out of memory is not something we can accurately evaluate at the high-level that Keras operates at.

We also consider it out of scope from a security point of view. It's on you the user to verify that the model will not run out of memory. This has to be tested with a specific hardware and software combination. There is no way we, Keras, can correctly predict if the program will crash with an OOM before it happens.

LinZiyuu:
Hi @hertschuh, thanks for the detailed explanation on #23051, that fully makes sense. I agree #23164 is out of scope for the same reason.

My other open security PRs are a different class — file-read / decompression-bomb / ReDoS, not weight allocation, and the same kind Keras has already merged. Could you review these when you get a chance?

- **#23163** — arbitrary HDF5 file read via an intermediate `ExternalLink` 
- **#23159** — npz/ZIP shape bomb; 
- **#23160** — decompression bomb in the public `extract_archive` / `get_file(extract=True)` 
- **#23161** — HDF5 shape bomb in `KerasFileEditor` (same class as CVE-2026-0897)
- **#23162** — ReDoS via the config-serialized `Muon.exclude_layers` regex

Thanks
Ziyu

PR #23052, opened 2026-06-08
Implement PyTorch distribution utility functions and device management

buildwithsuhana:
## Description
This PR introduces the foundational distribution utilities for the PyTorch backend (distribution_lib.py) and comprehensive test coverage. These utilities provide a unified interface for device discovery, multi-process initialization, and Keras-to-Torch device mapping, enabling subsequent DataParallel and ModelParallel implementations.

Key Changes

   - Distributed Initialization: Implemented initialize() with support for multi-host setups by parsing job_addresses to configure the init_method (TCP).
   - Robust Device Mapping: Added to_backend_device() which maps Keras device strings (e.g., gpu:0, tpu:1) to native torch.device objects.

## Contributor Agreement

### DIFF
--- a/keras/src/backend/torch/distribution_lib.py
+++ b/keras/src/backend/torch/distribution_lib.py
@@ -1,164 +1,195 @@
 import os
 
+import numpy as np
 import torch
+import torch.distributed
 
-
-def _is_available(device_type):
-    """Check if a device type is available."""
-    if device_type in ("gpu", "cuda"):
-        return torch.cuda.is_available()
-    if device_type == "mps":
-        return torch.backends.mps.is_available()
-    if device_type == "xpu":
-        return hasattr(torch, "xpu") and torch.xpu.is_available()
-    if device_type == "tpu":
-        from keras.src.utils.module_utils import torch_xla
-
-        return torch_xla.available
-    return device_type == "cpu"
-
-
-def _get_default_device_type():
-    """Get the default device type for the current environment."""
-    for dt in ["gpu", "mps", "xpu", "tpu"]:
-        if _is_available(dt):
-            return dt
-    return "cpu"
-
-
-def get_device_count(device_type=None):
-    """Returns total device count across all hosts.
-
-    Args:
-        device_type: String of `"cpu"`, `"gpu"`, `"mps"`, `"xpu"`, or `"tpu"`.
-            Defaults to the best available device type.
-
-    Returns:
-        Int, total number of devices across all hosts.
-    """
-    device_type = (device_type or _get_default_device_type()).lower()
-    if device_type == "cuda":
-        device_type = "gpu"
-    if not _is_available(device_type):
-        return 0
-
-    if torch.distributed.is_initialized():
-        return torch.distributed.get_world_size()
-    if "WORLD_SIZE" in os.environ:
-        return int(os.environ["WORLD_SIZE"])
-
-    if device_type == "gpu":
-        return torch.cuda.device_count()
-    if device_type == "xpu":
-        return torch.xpu.device_count()
-    if device_type == "tpu":
-        from keras.src.utils.module_utils import torch_xla
-
-        return torch_xla.runtime.global_device_count()
-    return 1
+from keras.src.backend.torch.core import _parse_device_input
+from keras.src.backend.torch.core import get_device
 
 
 def list_devices(device_type=None):
-    """Returns Keras device strings representing global indices.
+    """Return all available devices as a list of strings.
 
     Args:
-        device_type: String of `"cpu"`, `"gpu"`, `"mps"`, `"xpu"`, or `"tpu"`.
-            Defaults to the best available device type.
+        device_type: Optional string, one of "cpu", "gpu", "cuda", or "tpu".
+            Defaults to the primary available device type.
 
     Returns:
-        List of strings like `["gpu:0", "gpu:1", ...]`.
+        A list of strings like ["gpu:0", "gpu:1"] or ["cpu:0"].
     """
-    device_type = (device_type or _get_default_device_type()).lower()
-    if device_type == "cuda":
-        device_type = "gpu"
-    return [f"{device_type}:{i}" for i in range(get_device_count(device_type))]
+    resolved_device_type = _parse_device_input(
+        device_type or get_device()
+    ).split(":")[0]
+    count = get_device_count(device_type)
+
+    display_type = (
+        "gpu" if resolved_device_type == "cuda" else resolved_device_type
+    )
+
+    return [f"{display_type}:{i}" for i in range(count)]
+
+
+def get_device_count(device_type=None):
+    """Return the total number of devices for a given type.
+
+    In a distributed setting, this returns the total number of processes
+    managing that device type across the cluster.
+
+    Args:
+        device_type: Optional string, one of "cpu", "gpu", "cuda", or "tpu".
+
+    Returns:
+        An integer representing the device count.
+    """
+    device_type = device_type.lower() if device_type else None
+
+    if torch.distributed.is_initialized() or "WORLD_SIZE" in os.environ:
+        actual_device_type = _parse_device_input(get_device()).split(":")[0]
+
+        if device_type in (None, "cpu", actual_device_type) or (
+            device_type == "gpu" and actual_device_type == "cuda"
+        ):
+            return num_processes()
+
+        return 0
+
+    resolved_device_type = _parse_device_input(
+        device_type or get_device()
+    ).split(":")[0]
+
+    if resolved_device_type == "cuda":
+        return torch.cuda.device_count()
+
+    if resolved_device_type == "mps":
+        return 1
+
+    if resolved_device_type == "xpu":
+        return torch.xpu.device_count()
+
+    if resolved_device_type == "tpu":
+        from keras.src.utils.module_utils import torch_xla
+
+        if torch_xla.available:
+            import torch_xla.core.xla_model as xm
+
+            return xm.xla_device_count()
+
+    return 1 if resolved_device_type == "cpu" else 0
 
 
 def initialize(job_addresses=None, num_processes=None, process_id=None):
-    """Initialize the current process for distributed training.
+    """Initialize the distributed process group.
 
     Args:
-        job_addresses: Address of the coordinator process (process 0).
-        num_processes: Total number of processes.
-        process_id: Rank of the current process.
+        job_addresses: Optional string, comma-separated list of host:port
+            addresses. The first address is used as the MASTER_ADDR/MASTER_PORT.
+        num_processes: Optional integer, the total number of processes
+            (WORLD_SIZE).
+        process_id: Optional integer, the rank of the current process.
     """
-    if torch.distributed.is_initialized():
-        return
+    if job_addresses:
+        address = job_addresses.split(",")[0]
+
+        if ":" in address:
+            master_addr, master_port = address.split(":")
+            os.environ.setdefault("MASTER_ADDR", master_addr)
+            os.environ.setdefault("MASTER_PORT", master_port)
+        else:
+            os.environ.setdefault("MASTER_ADDR", address)
 
     if num_processes is not None:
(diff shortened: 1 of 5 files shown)


keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,154 @@
...
+
+    if num_processes is not None:
+        os.environ["WORLD_SIZE"] = str(num_processes)
+    if process_id is not None:
+        os.environ["RANK"] = str(process_id)
+
+    world_size = int(os.environ.get("WORLD_SIZE", 1))
+    if world_size > 1:
+        rank = int(os.environ.get("RANK", 0))
+        local_rank = int(os.environ.get("LOCAL_RANK", rank))
```
hertschuh: What's the difference between RANK and LOCAL_RANK?

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,154 @@
...
+        if torch.cuda.is_available():
+            torch.cuda.set_device(local_rank)
+
+        init_method = None
+        if job_addresses:
+            coordinator = job_addresses.split(",")[0]
```
hertschuh: We should not allow comma separated values if all the other values are ignored.

keras/src/backend/torch/distribution_lib.py
buildwithsuhana: `RANK` is the global index of a process across the entire cluster. It is used to identify the process for global communication and coordination.
`LOCAL_RANK` is the local index of a process relative to the specific machine (node) it is running on.

`RANK` is required for connection: PyTorch's init_process_group needs the global RANK to know which process is which across the entire network.
`LOCAL_RANK` is required for GPU selection: On a second or third computer, the global RANK might be 8 or 16, but that computer only has GPUs numbered 0 to 7. If you try to call set_device(16), it will crash because that GPU doesn't exist on that specific machine. You need the LOCAL_RANK (e.g., 0) to pick the correct GPU on that local box.

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,198 @@
...
+        device_type = os.environ.get("KERAS_TORCH_DEVICE")
+
+    if device_type:
+        device_type = device_type.lower()
+        return "cuda" if device_type == "gpu" else device_type
+
+    if xm and xm.xla_device_count() > 0:
+        return "tpu"
+
+    return "cuda" if torch.cuda.is_available() else "cpu"
```
hertschuh: There is some overlap with this:
https://github.com/keras-team/keras/blob/a2f8eac8b244d8aa6a6fb85d6a05e6da9b71362c/keras/src/backend/torch/core.py#L31-L40

But they differ. Some code should be shared between these two things.

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,198 @@
...
+    """Convert the DeviceMesh to PyTorch backend specific Mesh.
+
+    Args:
+        device_mesh: DeviceMesh instance to convert.
+
+    Returns:
+        A `torch.distributed.DeviceMesh` instance.
+    """
+    import numpy as np
+    from torch.distributed.device_mesh import DeviceMesh
```
hertschuh: Move imports to the top of the file.

The second one is not necessarily needed, you could just do `torch.distributed.device_mesh.DeviceMesh`. But up to you.

keras/src/backend/torch/distribution_lib.py
```diff
@@ -0,0 +1,198 @@
...
+
+    device_type_str = parts[0]
+    device_index = (
+        parts[1] if len(parts) > 1 else os.environ.get("LOCAL_RANK", "0")
+    )
+
+    resolved_device_type = _get_device_type(device_type_str)
+
+    if resolved_device_type == "cpu":
+        return torch.device("cpu")
```
hertschuh, line 193: So if you run multiple processes on CPU, each process only sees one CPU, right? And it's called `"cpu"`. Does distribution still work in this case?

keras/src/backend/torch/distribution_lib_test.py
```diff
@@ -0,0 +1,256 @@
...
+        m_cuda = mock.patch(
+            "torch.cuda.is_available",
+            return_value=cuda["available"]
+            if cuda
+            else torch.cuda.is_available(),
+        )
+        m_cuda_c = mock.patch(
+            "torch.cuda.device_count",
+            return_value=cuda.get("count", 0) if cuda else 0,
+        )
```
hertschuh: Let's not mock. Instead, let's verify that on CUDA it works as expected. You can use `testing.torch_uses_gpu()`.

keras/src/backend/torch/distribution_lib_test.py
```diff
@@ -0,0 +1,256 @@
...
+            },
+            True,
+            None,
+            None,
+            1,
+        ),
+        ({}, False, "cuda", {"available": True, "count": 2}, 2),
+        ({}, False, "gpu", {"available": True, "count": 2}, 2),
+    )
+    def test_get_device_count(self, env, init, dtype, cuda, expected):
```
hertschuh: Calling it `dtype` is very confusing. Spell it out `device_type`.

keras/src/backend/torch/distribution_lib_test.py
```diff
@@ -0,0 +1,256 @@
...
+        ("cpu", {}, False, False, ["cpu:0"]),
+        (
+            "gpu",
+            {"WORLD_SIZE": "4", "KERAS_TORCH_DEVICE": "gpu"},
+            False,
+            False,
+            ["gpu:0", "gpu:1", "gpu:2", "gpu:3"],
+        ),
+    )
+    def test_list_devices(self, dtype, env, init, default, expected):
```
hertschuh: Same comment about `dtype`.

keras/src/backend/torch/distribution_lib_test.py
```diff
@@ -0,0 +1,256 @@
...
+        # Mock CUDA available for GPU tests to ensure they return world size
+        cuda_val = True if dtype == "gpu" or (env.get("WORLD_SIZE")) else False
+        with mock.patch("torch.cuda.is_available", return_value=cuda_val):
+            if default:
+                devices = distribution_lib.list_devices()
+                self.assertTrue(
+                    any(devices[0].startswith(s) for s in ["gpu:", "cpu:"])
+                )
+            else:
+                self.assertEqual(distribution_lib.list_devices(dtype), expected)
```
hertschuh: Same comment about mocking.

keras/src/backend/torch/distribution_lib_test.py
```diff
@@ -0,0 +1,256 @@
...
+                    distribution_lib.initialize(addr, nproc, pid)
+                    m_set.assert_called_once()
+                    m_init.assert_called_once()
+                    args, kwargs = m_init.call_args
+                    self.assertEqual(kwargs["backend"], "nccl")
+                    self.assertEqual(kwargs["rank"], 0)
+                    self.assertEqual(kwargs["world_size"], 1)
+            else:
+                distribution_lib.initialize(addr, nproc, pid)
+                m_init.assert_called_once()
```
hertschuh: Same comment about mocking. Can we not mock and instead run when it's available.

What makes `torch.distributed.init_process_group` available? Is it when CUDA is available or something else?

keras/src/backend/torch/distribution_lib_test.py
```diff
@@ -0,0 +1,256 @@
...
+
+    @parameterized.parameters(
+        ("cpu", False, "cpu"),
+        (None, True, "cuda"),
+        (None, False, "cpu"),  # fallback
+    )
+    def test_get_device_type(self, k_dev, cuda, expected):
+        self.set_env("KERAS_TORCH_DEVICE", k_dev)
+        with mock.patch("torch.cuda.is_available", return_value=cuda):
+            self.assertEqual(distribution_lib._get_device_type(), expected)
```
hertschuh: Let's not mock and instead verify that we get what we expect based on `testing.torch_uses_gpu()`.

keras/src/backend/torch/distribution_lib_test.py
```diff
@@ -0,0 +1,256 @@
...
+            isinstance(inp, torch.device)
+            and inp.type == "cuda"
+            and not torch.cuda.is_available()
+        ):
+            self.skipTest("No CUDA")
+        with mock.patch("torch.cuda.is_available", return_value=cuda):
+            dev = distribution_lib._to_backend_device(inp)
+            self.assertEqual(dev.type, etype)
+            if eidx is not None:
+                self.assertEqual(dev.index, eidx)
```
hertschuh: Same, let's not mock and verify either CPU or GPU/CUDA based on the real device.

keras/src/backend/torch/distribution_lib_test.py
```diff
@@ -0,0 +1,256 @@
...
+        mesh = DeviceMesh(shape=(1, 2), axis_names=["x", "y"], devices=devs)
+        with (
+            mock.patch("torch.cuda.is_available", return_value=cuda),
+            mock.patch("torch.distributed.device_mesh.DeviceMesh") as m_mesh,
+        ):
+            distribution_lib._to_backend_mesh(mesh)
+            args, kwargs = m_mesh.call_args
+            # The first argument is device_type
+            self.assertEqual(args[0], etype)
+            self.assertEqual(kwargs["mesh_dim_names"], ("x", "y"))
```
hertschuh: Same.

hertschuh (changes requested):
Thanks for implementing this.

Since we run tests on different devices, let's not mock and instead it works for the current device, like what was done for JAX.

keras/src/backend/torch/distribution_lib_test.py
buildwithsuhana: Done

keras/src/backend/torch/distribution_lib_test.py
buildwithsuhana: Done

keras/src/backend/torch/distribution_lib_test.py
buildwithsuhana: Done

keras/src/backend/torch/distribution_lib_test.py
buildwithsuhana: I've removed the mock, init_process_group is always available in PyTorch because it supports distribution on both CPU and GPU. The tests now perform real initialization with a fixed port and include cleanup to avoid interfering with other tests.

keras/src/backend/torch/distribution_lib_test.py
buildwithsuhana: Fixed

keras/src/backend/torch/distribution_lib_test.py
buildwithsuhana: Done

keras/src/backend/torch/distribution_lib_test.py
buildwithsuhana: Done

keras/src/backend/torch/distribution_lib_test.py
buildwithsuhana: Done

keras/src/backend/torch/distribution_lib.py
buildwithsuhana: Yes, it works. On CPU, all processes use the same cpu device. PyTorch handles the distribution using the process rank and the communication backend, rather than a device index.

keras/src/backend/torch/distribution_lib.py
buildwithsuhana: Done

keras/src/backend/torch/distribution_lib.py
buildwithsuhana: Done, distribution_lib now uses the core functions

hertschuh (approved):
Thanks!

Issue #1023, opened 2015-11-16
Combining different filter lengths in 1D convolutional layers


sebastianruder:
[1] show that combining filter lengths around the 'best' filter length achieves better sentence classification results than just using the 'best' filter length. Is there a possibility to specify a combination of filter lengths for 1D convolution? If not, is this a feature that should be implemented?

Zhang, Y., & Wallace, B. (2015). A Sensitivity Analysis of (and Practitioners’ Guide to) Convolutional Neural Networks for Sentence Classification, (1). Retrieved from http://arxiv.org/abs/1510.03820

dbonadiman:
This is kinda of topic but thanks a lot for the paper :) 
For the question itself i think that it should be up to the user to do that, on the other side the k-max pooling layer is missing from Keras and that is a big issue.

dbonadiman:
@sebastianruder Just a reminder do not exaggerate on the nb_filters because depending on the task you are modelling it might not converge (just experienced that problem).

sebastianruder:
@dbonadiman: Wow, that was fast indeed! Thanks a lot for the commented implementation snippet and the tip! :D

dbonadiman:
@sebastianruder It is fast to do these implementations if you have a working model to test them over ;)

MdAsifKhan:
It creates an identifier issue:

MdAsifKhan:
It creates an identifier issue:
  File "build/bdist.linux-x86_64/egg/keras/layers/containers.py", line 366, in add_node
Exception: Unknown identifier: 7input

Ambros94:
This graph model can still be created with keras 2.0 using sequential models?

hsiaoyi0504:
@Ambros94 They are different containers actually. You should check the [documentation](https://keras.io/layers/containers/) for further details.

PR #23053, opened 2026-06-08
Add regression test for dict input name preservation in LiteRT torch export

pctablet505:
Adds `test_dict_input_names_preserved` to `litert_torch_test.py` with a version-gated skip so it only runs when `litert-torch > 0.9.1` is installed.

This covers the upstream fix for:
- https://github.com/google-ai-edge/litert-torch/issues/1022

The test verifies that when a Functional model accepts dict inputs, the exported LiteRT model preserves the original dict keys in the TFLite signature (as fixed in `litert-torch` master via PR #1023).

**Test behavior:**
- `litert-torch <= 0.9.1` (released): test is **skipped**
- `litert-torch > 0.9.1` (master/upcoming release): test **passes**

### DIFF
--- a/keras/src/export/litert_torch_test.py
+++ b/keras/src/export/litert_torch_test.py
@@ -9,6 +9,7 @@ import numpy as np
 import pytest
 import tensorflow as tf
 import torch
+from packaging.version import parse as parse_version
 
 from keras.src import backend
 from keras.src import layers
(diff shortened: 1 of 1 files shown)


amitsrivastava78 (approved):
LGTM

hertschuh (approved):
Thanks!

PR #23056, opened 2026-06-08
Disable autocast in GroupNormalization for mixed precision stability

### LINKED ISSUE #22586: [Bug] Mixed Precision (float16) numerical instability in GroupNormalization with small epsilon

Vedag812:
Fixes #22586

## Contributor Agreement

## The problem

Under `mixed_float16`, `GroupNormalization` produces NaN outputs when inputs contain values above the float16 max (~65504). This is because `autocast=True` (the default) causes the layer's `__call__()` to cast inputs to float16 **before** `call()` runs. Values above 65504 overflow to `inf` at that point, and the internal float32 upcast inside `_apply_normalization` can't recover them since they are already `inf`.

This was discussed in #22586 and a previous PR (#22666) attempted a different approach (moving the upcast earlier in `call()`) but was closed after going stale without a signed CLA. The last comment on #22586 from @Lawson-Darrow identified the correct minimal fix.

## The fix

One line: `self.autocast = False` in `GroupNormalization.__init__()`.

This is the exact same approach `LayerNormalization` already uses (see [line 151](https://github.com/keras-team/keras/blob/master/keras/src/layers/normalization/layer_normalization.py#L151)). With autocast disabled, inputs keep their original dtype (float32) when entering `call()`, so the existing float32 upcast in `_apply_normalization` works as intended.

## Why this is the right approach

The Keras maintainer @hertschuh [commented](https://github.com/keras-team/keras/pull/22666#issuecomment-2825373637) that the fix should be to let users pass `autocast=False`. However, `LayerNormalization` already sets this unconditionally because normalization layers are inherently sensitive to precision. `GroupNormalization` should be consistent with that. Users shouldn't have to know about this footgun.

## What changed

| File | Change |
|------|--------|
| `group_normalization.py` | Added `self.autocast = False` in `__init__()` |
| `group_normalization_test.py` | Added regression test with large inputs under `mixed_float16` |

## Test

The new `test_mixed_float16_no_nan` creates a `GroupNormalization` layer with `dtype='mixed_float16'`, feeds it values in the 60000-70000 range (spanning the float16 max), and asserts no NaN in the output. Without the fix, this test fails.


Lawson-Darrow:
I offered to open this pr in the comment I laid out the fix in. Given the discussion around the issue and the contribution policy I though it best to wait for a maintainer to weigh in first before opening the pr.

Vedag812:
Hey @Lawson-Darrow, apologies for jumping ahead on this. Your analysis in the issue comment was spot on and directly informed this PR. I should have checked more carefully whether you were planning to submit it yourself.

If you'd prefer to open the PR from your end, I'm happy to close this one. Alternatively, if the maintainers are fine with this PR moving forward, I wanted to make sure you get proper credit for identifying the root cause and the correct fix (setting autocast=False to match LayerNormalization).

Either way works for me, let me know what you'd prefer.

Lawson-Darrow:
Thanks, no worries. 

Since you offered, I'll open the PR from my end so it ties back to the issue analysis and gives a maintainer the ability to weigh in first.  

Appreciate you catching it and feel free to review once it's up.

Lawson-Darrow:
Pr is posted.

#23070

hertschuh:
Closing in favor of https://github.com/keras-team/keras/pull/23070

PR #23069, opened 2026-06-10
[Bug-Fix] Audio resampling truncates before resampling

### LINKED ISSUE #23068: [Bug] Audio resampling truncates before resampling
`tf.audio.decode_wav(... desired_samples=output_sequence_length)` truncates in the source sampling rate before `tfio.audio.resample()` runs. A 48 kHz file with `sampling_rate=16000` and `output_sequence_length=16000` can produce about 5333 samples after resampling and not 16000.

repro:
```python
import os
import tempfile
import numpy as np
import tensorflow as tf
import keras

source_rate = 48000
target_rate = 16000
output_len = 16000

tmp = tempfile.mkdtemp()
wav_path = os.path.join(tmp, "sample.wav")

audio = np.random.uniform(-0.2, 0.2, size=(source_rate, 1)).astype("float32")
with open(wav_path, "wb") as f:
    f.write(tf.audio.encode_wav(audio, sample_rate=source_rate).numpy())

ds = keras.utils.audio_dataset_from_directory(
    tmp,
    labels=None,
    batch_size=None,
    shuffle=False,
    sampling_rate=target_rate,
    output_sequence_length=output_len,
)

sample = next(iter(ds))
print(sample.shape)
```

Observation:
For a 48 kHz WAV with sampling_rate=16000 and output_sequence_length=16000 keras first decodes only 16000 samples at 48 kHz which is ~0.333 sec audio. The partial audio is then resampled to 16 kHz producing roughly 5333 samples instead of 16000

Expected:
The dataset should output 16000 samples per ex after resampling assuming the input audio is long enough with shorter inputs padded with existing behavior.

ChiragSW:
## Description
Fixes: #23068 
When sampling_rate is set, WAV decode no longer truncates with desired_samples=output_sequence_length before resampling.
After resampling `_trim_and_pad_audio()` applies output_sequence_length to the returned audio. Updated the doc to account for this fix too and also added a regression test similar to the repro.

## Contributor Agreement

### DIFF
--- a/keras/src/utils/audio_dataset_utils.py
+++ b/keras/src/utils/audio_dataset_utils.py
@@ -76,7 +76,8 @@ def audio_dataset_from_directory(
             (the dataset will yield individual samples).
         sampling_rate: Audio sampling rate (in samples per second).
         output_sequence_length: Maximum length of an audio sequence. Audio files
-            longer than this will be truncated to `output_sequence_length`.
+            longer than this will be truncated to `output_sequence_length`,
+            after resampling if `sampling_rate` is set.
             If set to `None`, then all sequences in the same batch will
             be padded to the
             length of the longest sequence in the batch.
@@ -379,11 +380,13 @@ def read_and_decode_audio(
     """Reads and decodes audio file."""
     audio = tf.io.read_file(path)
 
-    if output_sequence_length is None:
-        output_sequence_length = -1
+    if output_sequence_length is not None and sampling_rate is None:
+        desired_samples = output_sequence_length
+    else:
+        desired_samples = -1
 
     audio, default_audio_rate = tf.audio.decode_wav(
-        contents=audio, desired_samples=output_sequence_length
+        contents=audio, desired_samples=desired_samples
     )
     if sampling_rate is not None:
         # default_audio_rate should have dtype=int64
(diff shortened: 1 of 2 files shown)


keras/src/utils/audio_dataset_utils.py
```diff
@@ -379,18 +380,30 @@ def read_and_decode_audio(
     """Reads and decodes audio file."""
     audio = tf.io.read_file(path)
 
-    if output_sequence_length is None:
-        output_sequence_length = -1
+    if output_sequence_length is not None and sampling_rate is None:
+        desired_samples = output_sequence_length
+    else:
+        desired_samples = -1
```
hertschuh, line 386: I think it would be easy to compute an estimate (rounded up) of `desired_samples` when the `sampling_rate` is set. Can you add this?

hertschuh (commented):
Thanks for fixing this!

ChiragSW:
Tried to do this locally but desired_samples needs to be a static/Python int in graph mode, so computing it from tf.audio.decode_wav() returned source rate doesn’t work inside dataset map.
The change is a bit larger if estimate is to be computed. I think current fix is good. Maybe the optimization part can be done later as I need some more directions for it.

hertschuh (approved):
Thanks for the fix!

PR #23070, opened 2026-06-10
Disable autocast in GroupNormalization for mixed precision stability

### LINKED ISSUE #22586: [Bug] Mixed Precision (float16) numerical instability in GroupNormalization with small epsilon

Lawson-Darrow:
## Description
GroupNormalization is numerically unstable under mixed precision because it does
not disable autocast the way LayerNormalization does. Under a `mixed_float16`
policy the layer's inputs and its `gamma`/`beta` reads get downcast to float16
before the normalization moments are computed. Inputs above the float16 max
(~65504) overflow to `inf` and produce NaN outputs and large affine weights
lose precision.

This sets `self.autocast = False` and passes `autocast=False` on the `gamma` and
`beta` weights matching LayerNormalization. The layer level flag keeps inputs in
their original precision through the moment computation. `autocast=False` on the
weights stops them being read as float16 inside an autocast scope (Keras can
still enter `AutocastScope(compute_dtype)` even with the layer-level flag off so
both pieces are needed to match LayerNormalization).

Added two regression tests:
- `test_mixed_precision_large_input_no_nan`: a large input under `mixed_float16`
  must not produce NaN (the reported symptom).
- `test_large_value_within_autocast_scope`: mirrors LayerNormalization's test,
  asserting the affine weights keep full precision inside an autocast scope.

Both tests fail on the current code and pass with this change. The full
`group_normalization_test.py` suite passes locally on the TensorFlow backend
(13 passed).

Fixes #22586

## Contributor Agreement

### DIFF
--- a/keras/src/layers/normalization/group_normalization.py
+++ b/keras/src/layers/normalization/group_normalization.py
@@ -1,7 +1,3 @@
-import warnings
-
-import ml_dtypes
-
 from keras.src import backend
 from keras.src import constraints
 from keras.src import initializers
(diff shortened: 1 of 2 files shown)


keras/src/layers/normalization/group_normalization.py
hertschuh: Can you remove this? This is no longer correct as the `compute_dtype` is no longer in the picture.

hertschuh (changes requested):
Thanks for addressing this!

keras/src/layers/normalization/group_normalization.py
Lawson-Darrow: Done!

hertschuh (approved):
Thanks!

PR #23073, opened 2026-06-10
[Bug-Fix] tf.data class weights fail for omitted classes

### LINKED ISSUE #23071: [Bug] `tf.data` class weights fail for omitted classes
`Model.fit(..., class_weight=...)` handles missing `class_weight` entries differently upon the input.
For `tf.data.Dataset` inputs Keras builds a dense `class_weight_tensor` from `range(max(class_weight.keys()) + 1)` and then calls `tf.gather(class_weight_tensor, y_classes)`. If the dataset contains a label whose class id is greater than the largest explicit key in `class_weight`, `tf.gather` raises an oor error instead of assigning the default weight `1.0`.
This makes array inputs and `tf.data` inputs different for the same labels and `class_weight`.

repro: 

```python
import os
import numpy as np
import tensorflow as tf
import keras
os.environ["TF_CPP_MIN_LOG_LEVEL"] = "2"

x = np.array([[0.0], [1.0], [2.0], [3.0]], dtype="float32")
y = np.array([0, 1, 2, 2], dtype="int32")

# Classes 1 and 2 are omitted to chcek
class_weight = {0: 2.0}

def make_model():
    inputs = keras.Input(shape=(1,))
    outputs = keras.layers.Dense(3, activation="softmax")(inputs)
    model.compile(optimizer="sgd", loss="sparse_categorical_crossentropy")
    return model

print("array input")
make_model().fit(
    x,
    y,
    class_weight=class_weight,
    batch_size=2,
    epochs=1,
    verbose=0,
)
print("array input completed")

print("tf.data input")
ds = tf.data.Dataset.from_tensor_slices((x, y)).batch(2)
make_model().fit(
    ds,
    class_weight=class_weight,
    epochs=1,
    verbose=0,
)
```
output:
The array input path completes successfully.

The tf.data.Dataset path fails with:

```
InvalidArgumentError: Error in user-defined function passed to MapDataset:
indices[1] = 1 is not in [0, 1)
```

ChiragSW:
## Description
Fixes: #23071 
`tf_dataset_adapter.py` uses a `tf.lookup.StaticHashTable` with default 1.0 instead of dense `tf.gather` and handles empty class_weight={} by returning ones. Also added a regression test for the same as repro.

## Contributor Agreement

### DIFF
--- a/keras/src/layers/preprocessing/image_preprocessing/random_perspective.py
+++ b/keras/src/layers/preprocessing/image_preprocessing/random_perspective.py
@@ -141,11 +141,6 @@ class RandomPerspective(BaseImagePreprocessingLayer):
             seed=seed,
             dtype=self.compute_dtype,
         )
-        perspective_factor = self.backend.numpy.where(
-            apply_perspective[:, None, None],
-            perspective_factor,
-            self.backend.numpy.zeros_like(perspective_factor),
-        )
 
         start_points = self.backend.convert_to_tensor(
             [
--- a/keras/src/layers/preprocessing/image_preprocessing/random_perspective.py
+++ b/keras/src/layers/preprocessing/image_preprocessing/random_perspective.py
@@ -141,6 +141,11 @@ class RandomPerspective(BaseImagePreprocessingLayer):
             seed=seed,
             dtype=self.compute_dtype,
         )
+        perspective_factor = self.backend.numpy.where(
+            apply_perspective[:, None, None],
+            perspective_factor,
+            self.backend.numpy.zeros_like(perspective_factor),
+        )
 
         start_points = self.backend.convert_to_tensor(
             [
--- a/keras/src/layers/preprocessing/image_preprocessing/random_perspective_test.py
+++ b/keras/src/layers/preprocessing/image_preprocessing/random_perspective_test.py
@@ -38,6 +38,10 @@ class RandomPerspectiveTest(testing.TestCase):
 
         np.random.seed(seed)
         inputs = np.random.randint(0, 255, size=(224, 224, 3))
+        transformation = layer.get_random_transformation(inputs)
+        self.assertAllClose(
+            transformation["start_points"], transformation["end_points"]
+        )
         output = layer(inputs)
         self.assertAllClose(inputs, output)
 
--- a/keras/src/layers/preprocessing/image_preprocessing/random_perspective_test.py
+++ b/keras/src/layers/preprocessing/image_preprocessing/random_perspective_test.py
@@ -38,10 +38,6 @@ class RandomPerspectiveTest(testing.TestCase):
 
         np.random.seed(seed)
         inputs = np.random.randint(0, 255, size=(224, 224, 3))
-        transformation = layer.get_random_transformation(inputs)
-        self.assertAllClose(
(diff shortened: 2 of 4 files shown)


ChiragSW:
Now did a stateless tensor + `tf.gather` implementation.

ChiragSW:
Applied fix for the failing test

keras/src/trainers/data_adapters/tf_dataset_adapter.py
```diff
@@ -106,12 +107,18 @@ def make_class_weight_map_fn(class_weight):
...
+        class_weight_clean = {
+            int(key): float(value) for key, value in class_weight.items()
+        }
+        max_class = max(class_weight_clean.keys())
+        class_weight_tensor = tf.convert_to_tensor(
+            [class_weight_clean.get(c, 1.0) for c in range(max_class + 1)],
+            dtype=class_weight_dtype,
+        )
+    else:
+        class_weight_tensor = None
```
hertschuh: This if / else is not needed, this is only called when `class_weight` exists.

But change line 29 from `if class_weight is not None:` to `if class_weight:`.

keras/src/trainers/data_adapters/tf_dataset_adapter.py
```diff
@@ -137,7 +144,16 @@ def class_weights_map_fn(*data):
             # Special casing for rank 1, where we can guarantee sparse encoding.
             y_classes = tf.cast(tf.round(y), tf.int32)
 
-        cw = tf.gather(class_weight_tensor, y_classes)
+        if class_weight_tensor is None:
+            cw = tf.ones(tf.shape(y_classes), dtype=class_weight_dtype)
```
hertschuh: This if / else is not needed, this is only called when `class_weight` exists

keras/src/layers/preprocessing/image_preprocessing/random_perspective.py
```diff
@@ -141,6 +141,11 @@ def get_random_transformation(self, data, training=True, seed=None):
             seed=seed,
             dtype=self.compute_dtype,
         )
+        perspective_factor = self.backend.numpy.where(
+            apply_perspective[:, None, None],
+            perspective_factor,
+            self.backend.numpy.zeros_like(perspective_factor),
+        )
```
hertschuh: Unrelated, move to a separate PR.

keras/src/layers/preprocessing/image_preprocessing/random_perspective_test.py
```diff
@@ -38,6 +38,10 @@ def test_random_perspective_no_op(self):
 
         np.random.seed(seed)
         inputs = np.random.randint(0, 255, size=(224, 224, 3))
+        transformation = layer.get_random_transformation(inputs)
+        self.assertAllClose(
+            transformation["start_points"], transformation["end_points"]
+        )
```
hertschuh: Unrelated, move to a separate PR.

hertschuh (changes requested):
Thanks for the fix!

ChiragSW:
The changes are done. I apologize for the unrelated changes being pushed, there was a branch mixup

hertschuh (approved):
Thank you for the fix!

